PR #515, opened 2023-07-28
Add first batch of qsharp language samples

sezna:


samples/language/Comments.qs
```diff
@@ -0,0 +1,21 @@
+/// # Sample
+/// Comments
+///
+/// # Description
+/// Comments begin with two forward slashes (//) and continue until the
+/// end of line. Such end-of-line comments may appear anywhere in the
+/// source code. Q# does not currently support block comments.
+/// Documentation comments, or doc comments, are denoted with three
+/// forward slashes instead of two.
```
cesarzc, line 9:
```suggestion
/// Comments begin with two forward slashes (`//`) and continue until the
/// end of line. Comments may appear anywhere in the source code. 
/// Q# does not currently support block comments.
/// Documentation comments, or doc comments, are denoted with three
/// forward slashes (`///`) instead of two.
```

cesarzc: Some notes on the suggestions:
- Text that is meant to be Q# source code can be enclosed between backticks ` `.
- Simplified one sentence to make it clearer (IMO).

samples/language/Comments.qs
```diff
@@ -0,0 +1,21 @@
...
+
```
cesarzc, line 10: nit: I think the blank line between doc comments and the namespace declaration is not needed.

samples/language/EntryPoint.qs
```diff
@@ -0,0 +1,17 @@
+/// # Sample
+/// EntryPoint
+///
+/// # Description
+/// When defining an entry point into a Q# program, the Q# compiler recognizes
+/// the @EntryPoint() attribute rather than requiring that entry points have a
+/// particular name, for example, main, Main, or __main__. That is, from the
+/// perspective of a Q# developer, entry points are ordinary operations annotated
+/// with @EntryPoint().
```
cesarzc, line 9:
```suggestion
/// The `@EntryPoint()` attribute is used to designate a particular operation as
/// the entry point of a Q# program rather than requiring entry points to have 
// a particular name such as `main`, `Main`, or `__main__`.
```

cesarzc: I've changed the description quite a bit to hopefully make it clearer IMO.

This is just a suggestion, please feel free to ignore it if you think the previous description was better :)

samples/language/Comments.qs
cesarzc: The same applies to the other files in this PR.

samples/language/EntryPoint.qs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace MyQuantumApp {
+
+    @EntryPoint()
```
cesarzc, line 10:
```suggestion
    // The Q# compiler identifies `Main` as the entry point operation because it is marked with the `@EntryPoint()` attribute.
    @EntryPoint()
```

cesarzc: Including additional comments in the code sections that are relevant to the sample topic would be helpful for AI training even if it seems a bit redundant.

samples/language/Functions.qs
```diff
@@ -0,0 +1,15 @@
+/// # Sample
+/// Functions  
+///
+/// # Description
+/// A function's execution always produces the same output, given the same input.
+/// Q# allows you to explicitly split out such purely deterministic computations into functions. 
+/// Functions in Q#, contrasted with _operations_, are pure, i.e. lacking side-effects.
+/// Thus, functions can only call other functions, while operations can call both functions and operations.
+namespace MyQuantumApp {
```
cesarzc, line 9: In this case I would make `Main` an operation and define another function that contains calls into other functions and commented calls into operations. This way the sample code shows what is allowed and what not:
```qsharp
function MyFunction(qubits: Qubit[]) : Int {
    // Functions can call other functions.
    let length = Length(qubits);

    // Functions cannot call operations.
    // Uncommenting the following line would cause a compilation error.
    // CNOT(qubits[0], qubits[1]);

    return length;
}

File: samples/language/Namespaces.qs [hunk 6]
```diff
@@ -0,0 +1,14 @@
+/// # Sample
+/// Namespaces
+///
+/// # Description
```
cesarzc, line 4: Could you please mention how to open namespaces in the description?

Also, could you add opened namespaces to the code, maybe using a function/operation from an opened namespace and add comments explaining the code itself. E.g.:
```qsharp
// The following program declares a namespace block named `MyQuantumApp`.
namespace MyQuantumApp {
    // The following `open` directive is used to import all types and callables declared in the
    // Microsoft.Quantum.Arrays namespace.

    // Rest of the sample...
}
```

samples/language/Operations.qs
```diff
@@ -0,0 +1,25 @@
+/// # Sample
+/// Operations   
+///
+/// # Description
+/// Operations are the basic building blocks of a Q# program. A Q#
+/// operation is a quantum subroutine. That is, it's a callable routine
+/// that contains quantum operations that modify the state of the
+/// qubit register.
```
cesarzc, line 8:
```suggestion
/// that contains quantum operations that modify the state of qubits.
```

samples/language/Operations.qs
```diff
@@ -0,0 +1,25 @@
...
+namespace MyQuantumApp {
+	
+    @EntryPoint()
+    operation MeasureOneQubit() : Result {
+        // Allocate a qubit, by default it is in zero state      
+        use q = Qubit();  
+        // We apply a Hadamard operation H to the state
+        // It now has a 50% chance of being measured 0 or 1
```
cesarzc, line 16:
```suggestion
        // We apply a Hadamard transformation to qubit `q` using the H operation
```

cesarzc: Removed the second line of the comment since it is not very relevant for the purposes of this specific sample.

samples/language/Operations.qs
```diff
@@ -0,0 +1,25 @@
...
+namespace MyQuantumApp {
+	
+    @EntryPoint()
+    operation MeasureOneQubit() : Result {
+        // Allocate a qubit, by default it is in zero state      
+        use q = Qubit();  
+        // We apply a Hadamard operation H to the state
+        // It now has a 50% chance of being measured 0 or 1  
+        H(q);      
+        // Now we measure the qubit in Z-basis.
```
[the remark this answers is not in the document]
cesarzc: Emphasizing the use of operations.

samples/language/Operations.qs
```diff
@@ -0,0 +1,25 @@
...
+        let result = M(q);
+        // We reset the qubit before releasing it.
+        if result == One { X(q); }
```
cesarzc, line 21:
```suggestion
        // We reset the qubit before releasing it using the `Reset` operation.
        Reset(q);
```

cesarzc: Resetting the qubit using the measurement result is logically correct but could not be executed on some hardware platforms (it implies branching based on measurement capabilities). It is more portable to use the `Reset` operations which works on all platforms. In general, when possible, our samples should aim to be as portable as possible.

samples/language/Specializations.qs
```diff
@@ -0,0 +1,33 @@
+/// # Sample
+/// Specializations
+///
+/// # Description
+/// Q# employs symbolic computation that can automatically generate
+/// the corresponding adjoint implementation for a particular body implementation.
+namespace MyQuantumApp {
+
+    /// Implicit
```
cesarzc, line 9:
```suggestion
    // The adjoint, controlled and adjoint-controlled specializations are implicitly
    // generated for the `DoNothing` operation that declares supports for these
    // specializations using the `is` keyword followed by the union of the supported
    // specializations (`Adj + Ctl`).
```

samples/language/Specializations.qs
```diff
@@ -0,0 +1,33 @@
...
+    operation DoNothing() : Unit 
+        is Adj + Ctl { }
+
+    /// Explicit
```
cesarzc, line 13: Could you please add a similar level of detail in the comment regarding explicitly defined specializations?

cesarzc: Adding a brief comment on each specialization implementation would also be helpful.

samples/language/Comments.qs
```diff
@@ -0,0 +1,21 @@
...
+namespace MyQuantumApp {
+    open Microsoft.Quantum.Diagnostics;
+	
+	/// This is a doc-comment for the `Main` operation.
```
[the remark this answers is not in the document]
cesarzc: Seems like a stubborn tab 😆 
GH's diff still showing an off indentation.

PR #517, opened 2023-07-28
Fix katas using package type setting

### LINKED ISSUE #516: Language server should have configurable documentation type
At the moment, the language server treats every document as an executable, meaning it is required to have exactly one `@EntryPoint()` attributed callable or a valid entry expression. Eventually, we need a way to configure the document type so the language service understands when a library is being edited and no entry is required. This is especially important for the editing experience for kata exercises, which are not exe documents.

swernli:
This is a workaround to fix kata behavior in the playground, where they were incorrectly treated as exe documents and failed compilation due to missing entry point. The katas need to be treated as if they are lib document. This change adds a new parameter to `updateDocument` on the language server so that the caller can indicate what type the document is.

Fixes #516

### DIFF
--- a/language_service/src/lib.rs
+++ b/language_service/src/lib.rs
@@ -15,6 +15,7 @@ use crate::{
     completion::CompletionList, definition::Definition, hover::Hover, qsc_utils::compile_document,
 };
 use log::trace;
+use qsc::PackageType;
 use qsc_utils::Compilation;
 use std::collections::HashMap;
 
@@ -50,9 +51,15 @@ impl<'a> LanguageService<'a> {
     /// This should be called before any language service requests have been made
     /// for the document, typically when the document is first opened in the editor.
     /// It should also be called whenever the source code is updated.
-    pub fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+    pub fn update_document(
+        &mut self,
+        uri: &str,
+        version: u32,
+        text: &str,
+        package_type: PackageType,
+    ) {
         trace!("update_document: {uri:?} {version:?}");
-        let compilation = compile_document(uri, text);
+        let compilation = compile_document(uri, text, package_type);
         let errors = compilation.errors.clone();
 
         // insert() will update the value if the key already exists
--- a/npm/src/browser.ts
+++ b/npm/src/browser.ts
@@ -154,12 +154,18 @@ export { type CompilerState } from "./compiler/compiler.js";
 export { QscEventTarget } from "./compiler/events.js";
 export {
   getAllKatas,
-  getExerciseDependencies,
+  getExerciseSources,
   getKata,
-  type Kata,
-  type KataSection,
+  type ContentItem,
   type Example,
   type Exercise,
+  type ExplainedSolution,
+  type ExplainedSolutionItem,
+  type Kata,
+  type KataSection,
+  type Lesson,
+  type LessonItem,
+  type Question,
 } from "./katas.js";
 export { default as samples } from "./samples.generated.js";
 export { type VSDiagnostic } from "./vsdiagnostic.js";
--- a/npm/src/browser.ts
+++ b/npm/src/browser.ts
@@ -167,3 +167,4 @@ export { log, type LogLevel };
 export type { ICompilerWorker, ICompiler };
 export type { ILanguageServiceWorker, ILanguageService };
 export { type LanguageServiceEvent } from "./language-service/language-service.js";
+export { PackageType } from "../lib/web/qsc_wasm.js";
--- a/npm/src/browser.ts
+++ b/npm/src/browser.ts
@@ -164,7 +164,11 @@ export {
 export { default as samples } from "./samples.generated.js";
 export { type VSDiagnostic } from "./vsdiagnostic.js";
 export { log, type LogLevel };
-export type { ICompilerWorker, ICompiler };
-export type { ILanguageServiceWorker, ILanguageService };
+export type {
+  ICompilerWorker,
+  ICompiler,
+  ILanguageServiceWorker,
+  ILanguageService,
+};
+export const PackageType = wasm.PackageType;
 export { type LanguageServiceEvent } from "./language-service/language-service.js";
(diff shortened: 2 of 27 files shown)


language_service/src/lib.rs
```diff
@@ -50,9 +51,15 @@ impl<'a> LanguageService<'a> {
     /// This should be called before any language service requests have been made
     /// for the document, typically when the document is first opened in the editor.
     /// It should also be called whenever the source code is updated.
-    pub fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+    pub fn update_document(
+        &mut self,
+        uri: &str,
+        version: u32,
+        text: &str,
+        package_type: PackageType,
```
billti, line 59: Seems a little odd that it's per-document, rather than a setting on the entire package/compilation, but I guess until we have multi-file projects this can be a workaround for now.

npm/src/compiler/compiler.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { IDiagnostic } from "../../lib/node/qsc_wasm.cjs";
+import { PackageType } from "../../lib/web/qsc_wasm.js";
```
billti, line 5: Can you not just add the extra import on the line above? The node and web packages should be the same (we use the node one as it has a little less noise in it).

npm/src/language-service/language-service.ts
```diff
@@ -8,6 +8,7 @@ import type {
   IDefinition,
   LanguageService,
 } from "../../lib/node/qsc_wasm.cjs";
+import { PackageType } from "../../lib/web/qsc_wasm.js";
```
billti, line 11: See prior comment. Add PackageType to the above imports.

npm/src/browser.ts
```diff
@@ -167,3 +167,4 @@ export { log, type LogLevel };
 export type { ICompilerWorker, ICompiler };
 export type { ILanguageServiceWorker, ILanguageService };
 export { type LanguageServiceEvent } from "./language-service/language-service.js";
+export { PackageType } from "../lib/web/qsc_wasm.js";
```
billti, line 170: Being we import all of the wasm package as 'wasm' in line 7 above, you should just be able to export `wasm.PackageType as PackageType` in the exports on line 166.

language_service/src/lib.rs
swernli: Yeah, this is definitely a workarond until we have packages.

npm/test/basics.js
```diff
@@ -19,6 +19,7 @@ import {
   getKata,
 } from "../dist/katas.js";
 import samples from "../dist/samples.generated.js";
+import { PackageType } from "../lib/node/qsc_wasm.cjs";
```
billti, line 22: Being that you're exporting this from `browser.ts` for the browser environments, you should probably also export from `main.ts` for the node environments, then you'd import from main.js in the list on line 9.

billti:
Just some import/export tweaks, else looks good. Thanks!

billti (approved):

cesarzc (approved):

npm/src/compiler/compiler.ts
swernli: When I try to make that change, I start getting failures when compiling the playground that look like this:
```
Running esbuild
X [ERROR] Could not resolve "util"

    ../npm/lib/node/qsc_wasm.cjs:4:45:
      4 │ const { TextDecoder, TextEncoder } = require(`util`);
        ╵                                              ~~~~~~

  The package "util" wasn't found on the file system but is built into node. 
  Are you trying to bundle for node? You can use "platform: 'node'" to do    
  that, which will remove this error.
```
If I leave it as importing from the .js instead of adding to the .cjs, then it works. I'm hesitant to change any platform settings for how esbuild runs just to try and address this error, so I think I'm better off leaving this as a separate import (though it's possible this is an issue with my local configuration, I suppose).

npm/src/language-service/language-service.ts
swernli: Oddly enough, making the suggested change here does work, while the previoius location does not. I'm wondering if there is something strange about the settings on my machine...

npm/src/browser.ts
swernli: I can't get that syntax to work; I can export `wasm` itself but then callers have to import that and use `wasm.PackageType.Lib` in the code.

billti:
Yeah, I just played around with it and it is a little tricky. Previously all of the compiler and language service only used types from the generated wasm, and the actual JavaScript (non-type code) was only loaded by the platform specific entry point (browser.ts or main.ts) as the wasm JavaScript is platform specific.

An enum is a value not a type. So by using an enum from the generate wasm in the compiler/language service code, we now need to suck the runtime specific JavaScript into those supposedly platform-agnostic abstractions, and not just types.

I'll try and think of a cleaner way to do this. My instinct is to just stay away from enums (they're a TypeScript construct. There are no enums in JavaScript).

billti:
I just added one commit which avoids the loading or platform specific runtime JavaScript when not desired. It does mean using the underlying value rather than the enum in a couple places. We can look at options there later, but this fixes the issue without undesirable side-effects for now.

Note: I had to build locally with '--no-check'. Seems there's quite  a bit of code in this branch failing cargo fmt or clippy for me.

billti:
Actually, I wonder if just a flag/bool `isExe` might be much simpler for that arg? I don't see us supporting more than type types (exe and lib) any time soon, and as I say, avoiding enums in TypeScript is best. They're kind of a minefield.

swernli:
> Actually, I wonder if just a flag/bool `isExe` might be much simpler for that arg? I don't see us supporting more than two types of projects (exe and lib) any time soon, and as I say, avoiding enums in TypeScript is best. They're kind of a minefield.

Yeah, at this point I'm ready to give in and just use a Boolean. I was applying Rust sensibility to this part, which prefers named enums for clarity. We can keep the enum in the Rust implementation, but just use a bool for the wasm boundary and ts parts.

PR #519, opened 2023-07-31
Namespace Info on Hover over Callables

ScottCarda-MS:
Adds what namespace a callable is in to the callable's hover info card.

### DIFF
--- a/language_service/src/hover/tests.rs
+++ b/language_service/src/hover/tests.rs
@@ -50,6 +50,7 @@ fn hover_callable_unit_types() {
             Test
             operation Bar() : Unit
             ```
+            ---
             Doc comment
             with multiple lines!
         "#]],
@@ -70,6 +71,7 @@ fn hover_callable_with_callable_types() {
             Test
             operation Foo(x: (Int => Int)) : (Int => Int)
             ```
+            ---
             Doc comment!
         "#]],
     );
@@ -108,6 +110,7 @@ fn hover_callable_unit_types_functors() {
             Test
             operation Foo() : Unit is Ctl
             ```
+            ---
             Doc comment!
         "#]],
     );
@@ -127,6 +130,7 @@ fn hover_callable_with_callable_types_functors() {
             Test
             operation Foo(x: (Int => Int is Adj + Ctl)) : (Int => Int is Adj) is Adj
             ```
+            ---
             Doc comment!
         "#]],
     );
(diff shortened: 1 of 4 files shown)


language_service/src/hover/tests.rs
```diff
@@ -46,7 +46,7 @@ fn hover_callable_unit_types() {
         }
     "#},
         &expect![[r#"
-            "```qsharp\noperation Bar() : Unit\n```\nDoc comment\nwith multiple lines!\n"
+            "```qsharp\nTest\noperation Bar() : Unit\n```\nDoc comment\nwith multiple lines!\n"
```
minestarks, line 49: I just noticed you're using the `assert_debug_eq` function instead of the `assert_eq` function. The former gets the debug representation of the string, which adds unnecessary `"` and `\n` characters.... you don't really need the debug representation since you're just comparing strings. I suggest changing to `assert_eq` and updating all the expect strings, I suspect it'll look much prettier.

language_service/src/qsc_utils.rs
```diff
@@ -59,5 +64,15 @@ pub(crate) fn find_item<'a>(compilation: &'a Compilation, id: &ItemId) -> Option
...
+        (
+            item,
+            item.parent
+                .and_then(|parent_id| package.items.get(parent_id))
+                .and_then(|parent| match &parent.kind {
+                    qsc::hir::ItemKind::Namespace(namespace, _) => Some(namespace.name.clone()),
+                    _ => None,
+                }),
+        )
+    })
```
minestarks, line 77: IMO this logic looks a bit one-off and belongs more in hover.rs (where you call `find_item`). You don't even use the namespace return value at the other callsite in display.rs.

minestarks (approved):
LGTM with a couple of suggestions for refactoring

language_service/src/hover/tests.rs
ScottCarda-MS: That's a big improvement! Thanks!

language_service/src/qsc_utils.rs
ScottCarda-MS: Since I'd need access to the `package` to do this logic, and that is most of what this function does, I'd have to have a hover-specific version of this function that is exactly what's in the PR now. Maybe an alternative is for the `find_item` to return the item and the `package` in a tuple?

ScottCarda-MS: I've tried the approach to return the package with the item. Let me know if this is better.

minestarks: Oh, I see. Yes I like what you came up with - somehow it seems more reusable for the future (though somewhat hypothetical at the moment)

PR #520, opened 2023-07-31
Adding main def in package.json

idavis:
Extension wasn't loading when installing locally without `"main"` entry defined.


ScottCarda-MS (approved):
Verified on my machine

minestarks (approved):

billti:
This shouldn't be needed. The extension should only need the `browser` entry point as we're a web extension. I'll look into what the issue was here.

PR #522, opened 2023-07-31
Add second batch of language samples

sezna:


samples/language/Array.qs
```diff
@@ -0,0 +1,39 @@
...
+    @EntryPoint()
+    operation Main() : Int[] {
+
+        // A basic Int Array literal
+        let intArray: Int[] = [1, 2, 3, 4];
+        // A basic String Array literal
+        let stringArray: String[] = ["a", "string", "array"];
+
+        // A new array expression creating the array `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`
+        let intArray = [0, size = 10];
```
cesarzc, line 30: We are not sure yet if we want to keep this way of creating arrays of arbitrary size.

Could you please use `let intArray = Repeated(0, 10);` instead? And update doc comments correspondingly?

Maybe we should also create a language issue evaluating this syntax.

samples/language/DataTypes.qs
```diff
@@ -0,0 +1,72 @@
...
+/// `Pauli`, and `Range`. In addition to these primitive types, Q# offers primitive aggregate types
+/// as well, `Array` and `Tuple`; composite types (a.k.a. user-defined types or UDTs); and function
+/// and operation types.
+namespace MyQuantumApp {
+    
+    /// In the below code, all varibles have type annotations to showcase their type.
+    @EntryPoint()
+    operation MeasureOneQubit() : Unit {
+        // Notably, Qubits are allocated with the `use` keyword instead of declared with the `let`
+		// keyword.
```
cesarzc, line 16: nit: there are some indentation issues in this file.

samples/language/Qubit.qs
```diff
@@ -0,0 +1,40 @@
...
+///
+/// # Description
+/// Q# uses the `Qubit` primitive type to represent quantum state. A qubit represents 
+/// the smallest addressable physical unit in a quantum computer. Qubits are long-lived,
+/// and accumulates transformations to quantum states. A Q# program has no ability to
+/// introspect into the state of a qubit, and thus is entirely agnostic about what a
+/// quantum state is or on how it is realized. Rather, a program can call operations
+/// such as Measure to learn information about the quantum state of the computation.
+namespace MyQuantumApp {
+
```
cesarzc, line 12: nit: I think the pattern so far has been to not have blank lines between the start of a namespace and `open` directives.

We need a linter to do all this for us!

samples/language/Qubit.qs
```diff
@@ -0,0 +1,40 @@
...
+/// # Description
+/// Q# uses the `Qubit` primitive type to represent quantum state. A qubit represents 
+/// the smallest addressable physical unit in a quantum computer. Qubits are long-lived,
+/// and accumulates transformations to quantum states. A Q# program has no ability to
+/// introspect into the state of a qubit, and thus is entirely agnostic about what a
+/// quantum state is or on how it is realized. Rather, a program can call operations
+/// such as Measure to learn information about the quantum state of the computation.
+namespace MyQuantumApp {
+    
+    open Microsoft.Quantum.Intrinsic;
```
cesarzc, line 13:
```suggestion
```

cesarzc: We can safely avoid opening the intrinsic namespace since it is automatically opened by the compiler.

samples/language/Result.qs
```diff
@@ -0,0 +1,30 @@
+/// # Sample
+/// Result
+///
+/// # Description
+/// Represents the result of a projective measurement onto the eigenspaces
+/// of a quantum operator with eigenvalues ±1. Possible values are `Zero` or `One`.
+namespace MyQuantumApp {
+
+    // Open the `Microsoft.Quantum.Measurement` namespace to import `MResetZ`.
+    open Microsoft.Quantum.Measurement;
```
cesarzc, line 10: Since the focus of this sample is the `Result` type, I would use `M` instead of `MResetZ` and do an explicit `Reset` on the qubit.

samples/language/Qubit.qs
```diff
@@ -0,0 +1,40 @@
...
+    open Microsoft.Quantum.Diagnostics;
+    /// In the below code, all varibles have type annotations to showcase their type.
+    @EntryPoint()
+    operation MeasureOneQubit() : Unit {
+        // The following statement allocates three qubits.
+        use qs = Qubit[3];
```
cesarzc, line 19: Could you also add an example of a single qubit allocation?

samples/language/Unit.qs
```diff
@@ -0,0 +1,22 @@
...
+
+    @EntryPoint()
+    operation ExplicitReturn() : Unit {
+        // Explicitly return `Unit`.
+        return ();
+    }
+    operation ImplicitReturn() : Unit {
+        // Implicitly return `Unit`.
+        ()
+    }
```
cesarzc, line 18:
```suggestion
```

cesarzc: In general, for samples we are avoiding implicit returns for now.

samples/language/Result.qs
```diff
@@ -0,0 +1,32 @@
...
+        H(q);
+
+		// Measure the qubit.
+		let measurement = M(q);
+		
+        // Reset the qubit.
+		Reset(q);
+
+		// Return the measurement.
+		return measurement;
```
minestarks, line 30: Wonky indentation

samples/language/Qubit.qs
```diff
@@ -0,0 +1,41 @@
...
+        Y(qs[2]);
+
+        // Apply the SWAP gate to qubits 3 and 1.
+        SWAP(qs[2], qs[0]);
+
+        // Display the quantum state of the qubits.
+        DumpMachine();
+
+        // Free the allocated qubits.
+        ResetAll(qs);
```
minestarks, line 38: For consistency you'll want to reset `qubit` as well, I think.

Also, I don't think this technically "frees" the qubit, at least in the sense that I understand from other languages/runtimes. It resets the qubit to the |0> state. The qubit is "freed" just by virtue of it going out of scope.

FWIW I found this mental model pretty confusing when it was first explained to me. @swernli or someone should probably fact-check me.

cesarzc: That's right, "Free" is not the precise word to use in this comment, "Reset" should be used instead.

cesarzc (approved):

samples/language/Pauli.qs
```diff
@@ -0,0 +1,20 @@
+/// # Sample
+/// Pauli
+///
+/// # Description
+/// Represents a single-qubit Pauli matrix. Possible values are `PauliI`, `PauliX`, `PauliY`, or `PauliZ`.
+namespace MyQuantumApp {
+
+    @EntryPoint()
+    operation MeasureSuperposition() : Result {
```
cesarzc, line 9:
```suggestion
    operation Main() : Result {
```

samples/language/Result.qs
```diff
@@ -0,0 +1,32 @@
+/// # Sample
+/// Result
+///
+/// # Description
+/// Represents the result of a projective measurement onto the eigenspaces
+/// of a quantum operator with eigenvalues ±1. Possible values are `Zero` or `One`.
+namespace MyQuantumApp {
+    @EntryPoint()
+    operation MeasureSuperposition() : Result {
```
cesarzc, line 9:
```suggestion
    operation Main() : Result {
```

samples/language/DataTypes.qs
```diff
@@ -0,0 +1,72 @@
...
+
+        // A user-defined-type (UDT) consisting of two named parameters, `Real` and `Imaginary`,
+        // and one anonymous parameter of Boolean type.
+        newtype ComplexBool = (Real: Double, Imaginary : Double, Bool);
+        // Instantiation of the above UDT.
+        let complex: ComplexBool = ComplexBool(42.0, 0.0, false);
+
+        // A function that takes an integer and returns a boolean. This variable declaration
+        // uses a Lambda function as its right hand side.
+        let function_type: Int => Bool = (int) => int == 0;
```
cesarzc, line 69: nit: pascalCase is preferred

cesarzc (approved):

PR #525, opened 2023-08-02
Implement new operations in the katas library

cesarzc:
Implement new operations in the katas library that allow to check the equivalence of unitary operations across multiple inputs.


katas/content/multi_qubit_gates/preparing_bell_state/Verification.qs
```diff
@@ -9,14 +9,19 @@ namespace Kata.Verification {
...
     operation CheckSolution() : Bool {
-        let isCorrect = VerifyMultiQubitOperation(Kata.BellState, BellState);
+        let solution = Kata.BellState;
+        let reference = BellState;
+        let isCorrect = CheckOperationsEquivalenceOnZeroState(solution, reference, 2);
 
-        // Output different feedback to the user depending on whether the exercise was correct.
-        use target = Qubit[2];
+        // Output different feedback to the user depending on whether the solution was correct.
+        use target = Qubit[2]; // |00〉
```
swernli, line 17: Now that `target` is only used in the else branch below, this use can be moved there.

katas/content/single_qubit_gates/prepare_arbitrary_state/Verification.qs
```diff
@@ -16,16 +16,17 @@ namespace Kata.Verification {
...
                 Message("The solution was incorrect for at least one test case.");
-                use target = Qubit[1];
-                let opOnRegister = register => op(register[0]);
-                let referenceOnRegister = register => reference(register[0]);
-                ShowQuantumStateComparison(target, opOnRegister, referenceOnRegister);
+                Message("Hint: examine the effect your solution has on the |0〉 state and compare it with the effect " +
+                    "it is expected to have.");
+                use target = Qubit[1]; // |0〉
+                ShowQuantumStateComparison(target, solution, reference);
+                ResetAll(target);
```
swernli, line 29: Most other `CheckSolution()` implementations reset the target qubits used to show the difference in behavior, but a few do not. Is that something that should be added to all of the checks for correctness or is there a reason some do not reset before release?

katas/content/single_qubit_gates/global_phase_i/Verification.qs
```diff
@@ -9,21 +9,21 @@ namespace Kata.Verification {
...
-
-        // Output different feedback to the user depending on whether the exercise was correct.
-        use target = Qubit[1];
-        let op = register => Kata.GlobalPhaseI(register[0]);
+        let solution = register => Kata.GlobalPhaseI(register[0]);
         let reference = register => GlobalPhaseI(register[0]);
+        let isCorrect = CheckOperationsEquivalenceStrict(solution, reference, 1);
+
+        // Output different feedback to the user depending on whether the solution was correct.
+        use target = Qubit[1]; // |0〉
```
swernli, line 17: Same here, `target` and the corresponding `ResetAll` call can be moved into the else-block.

katas/content/KatasLibrary.qs
```diff
@@ -2,9 +2,93 @@
...
+    /// # Summary
+    /// Given two operations, checks whether they act identically for all input states.
+    /// This operation is implemented by using the Choi–Jamiołkowski isomorphism.
+    operation CheckOperationsEquivalence(
+        op : (Qubit[] => Unit is Adj + Ctl),
+        reference : (Qubit[] => Unit is Adj + Ctl),
+        inputSize: Int)
+    : Bool {
+        Fact(inputSize > 0, "`inputSize` must be positive");
+        use (control, target) = (Qubit[inputSize], Qubit[inputSize]);
```
swernli, line 18: Why do we need `inputSize` number of control qubits here? That makes the resulting operation much more expensive, and I don't think it gives us any additional validation compared to using a single qubit. Is there some corner case I'm missing?

katas/content/single_qubit_gates/prepare_arbitrary_state/Verification.qs
cesarzc: I missed a few verifications that don't `Reset` the allocated qubits. I'll add them for correctness. Thanks for the catch!

katas/content/multi_qubit_gates/preparing_bell_state/Verification.qs
cesarzc: Updated.

katas/content/single_qubit_gates/prepare_arbitrary_state/Verification.qs
cesarzc: Added resets to all verifications that did not have it.

katas/content/KatasLibrary.qs
cesarzc: This is the same implementation as [AssertOperationsEqualReferenced](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertoperationsequalreferenced).

This is a generalization of the check for equivalence for n-qubits instead of just one, hence the input size parameter. Since `main` doesn't have many katas yet, we are using this to check operations that act on just 1 qubit, but it is/will be also used to check operations that act on multiple qubits. I think it's ok to have just one general implementation instead of two implementations, one for 1-qubit operations and another one for n-qubit operations.

The check could also be done using the approach of [AssertOperationsEqualInPlace](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertoperationsequalinplace) which does not need twice as many qubits, but it requires multiple calls to the operations being compared.

I believe the current implementation is the appropriate one for katas because most (maybe all) the katas just require a few qubits, so we generally won't be using more than 6 qubits to check solutions. In my opinion, this implementation is also easier to understand since it directly maps to the Choi–Jamiołkowski isomorphism.

cesarzc: @swernli based on our offline conversation, I created #535 and referenced in the doc comments of the operation.

swernli (approved):

PR #527, opened 2023-08-02
LS Auto Opens for Completions

ScottCarda-MS:
When a completion for a callable is selected by a user, checks to see if there is an `open` directive for the appropriate namespaces. If not, automatically add one. If there is an `open`, but the namespace is aliased, qualifies the inserted callable with the alias.


language_service/src/completion/tests.rs
```diff
@@ -40,31 +36,135 @@ fn in_block_contains_std_functions() {
         &expect![[r#"
             [
                 Some(
-                    (
-                        "Fake",
-                        Function,
-                        Some(
+                    CompletionItem {
```
ScottCarda-MS, line 39: I'd like to have these expects be a bit more succinct, but I've found it difficult to have something like a test-specific Display/Debug impl for the data types involved. @samarsha @swernli, any suggestions or Rust-tips to make this readout simpler?

vscode/src/completion.ts
```diff
@@ -45,6 +45,15 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
       const item = new CompletionItem(c.label, kind);
       item.sortText = c.sortText;
       item.detail = c.detail;
+      item.additionalTextEdits = c.additionalTextEdits?.map((edit) => {
+        return new vscode.TextEdit(
+          new vscode.Range(
+            document.positionAt(edit.range.start),
```
minestarks, line 51: Sadly there's another mapping that needs to happen in the JS layer every time we return an offset from the language service. It happens in `npm/src/language-service/language-service.ts`. As an example see how we handle the hover results:

https://github.com/microsoft/qsharp/blob/67bf92acb7fa92ddbdbea05f4c2b66fad92ed985/npm/src/language-service/language-service.ts#L109C13-L109C13

It's easy to miss because the utf8-utf16 conversion only matters when we're using multibyte Unicode characters in the source. Some of our samples (Bernstein-Vazirani I think?) have formulas in the sample that manifest this problem. If you're returning text edits to a position that occurs after these Unicode characters, your edits will be shifted unless you've added the mapping.

language_service/src/qsc_utils.rs
```diff
@@ -7,6 +7,13 @@ use qsc::{
     CompileUnit, PackageStore, PackageType, SourceMap, Span,
 };
 
+/// Represents a span of text used by the Language Server API
+#[derive(Debug, PartialEq)]
+pub struct LsSpan {
```
minestarks, line 12: I like this new common type, but feel that `qsc_utils.rs` is getting a bit mish-mashy. I would appreciate a new file `types.rs` - especially since I think we're likely to add more common data types soon.

Oh also I prefer just `Span` even though I know overloading names can get a bit frustrating sometimes :) It's just that we're _already_ in the language service module so I'm not a fan of the `Ls` prefix.

language_service/src/completion.rs
```diff
@@ -156,27 +205,27 @@ impl CompletionListBuilder {
...
-        );
-        self.push_sorted_completions(
-            Self::get_callables(core, &display),
-            CompletionItemKind::Function,
-        );
-        self.push_completions(Self::get_namespaces(current), CompletionItemKind::Module);
-        self.push_completions(Self::get_namespaces(std), CompletionItemKind::Module);
-        self.push_completions(Self::get_namespaces(core), CompletionItemKind::Module);
+        self.push_sorted_completions(Self::get_callables(current, &display));
+        self.push_sorted_completions(Self::get_std_callables(
```
minestarks, line 209: I see you only special-case the standard library. What if I declare two namespaces in the same file and reference one from the other? Is there a way to make the function more generic so that it's applicable to all `Package`s in the compilation, including the current one?

This would also set us up nicely for the future when we support multiple files.

language_service/src/completion.rs
```diff
@@ -185,75 +234,150 @@ impl CompletionListBuilder {
...
+                }
+
+                let additional_text_edits = if additional_edits.is_empty() {
+                    None
+                } else {
+                    Some(additional_edits)
+                };
+
+                let label = if let Some(qualification) = qualification {
+                    format!("{qualification}.{name}")
```
minestarks, line 331: Nice work handling aliases, I would have forgotten they existed...

minestarks (changes requested):

language_service/src/qsc_utils.rs
ScottCarda-MS: I actually find calling it `Span` hugely confusing as there is already a commonly referenced type `Span` from the compiler, and just seeing `Span` somewhere makes me think that we are referencing the compiler `Span`. This is why I wanted the span that is specific to the Language Server API to be called something different or prefaced in someway to distinguish it from the other span.

language_service/src/completion.rs
ScottCarda-MS: I've made the change to support auto-opens for callables in other namespaces. While doing so I discovered that it is possible in Q# to have nested operations in Q#, which are difficult to support in the current implementation of completions. I've had to disallow these in this PR to avoid completion items that users don't actually have access to. We'll have to work on supporting this in the future.

vscode/src/completion.ts
ScottCarda-MS: Nice catch! Unicode can be tricky! Interestingly the implementation for the getHover seemed to be incorrect, as when I copied from that, I got incorrect behavior. After reading the doc for the `mapUtf16UnitsToUtf8Units` I found that the return object is actually a map from the original offset to the new offset value, not an array that takes indices. I've fixed that one as well.

language_service/src/qsc_utils.rs
ScottCarda-MS: Discussed offline. Resolved to create a separate file for defining the protocol surface for the language server that will define the data structures required by the language server protocol. The `Span` type will be defined there.

minestarks (approved):

swernli (approved):

PR #530, opened 2023-08-03
Fix WASM logging for VS Code language service

minestarks:
Fix wasm logging initialization and correct log levels.

### DIFF
--- a/npm/src/browser.ts
+++ b/npm/src/browser.ts
@@ -111,10 +111,7 @@ export function getCompilerWorker(workerArg: string | Worker): ICompilerWorker {
 }
 
 export async function getLanguageService(): Promise<ILanguageService> {
-  if (!wasmModule) throw "Wasm module must be loaded first";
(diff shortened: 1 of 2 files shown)


npm/src/browser.ts
```diff
@@ -111,10 +111,7 @@ export function getCompilerWorker(workerArg: string | Worker): ICompilerWorker {
 }
 
 export async function getLanguageService(): Promise<ILanguageService> {
-  if (!wasmModule) throw "Wasm module must be loaded first";
-  if (!wasmInstancePromise) wasmInstancePromise = initWasm(wasmModule);
-  await wasmInstancePromise;
-
+  await instantiateWasm();
```
billti, line 114: Is this just a drive by fix-up to match the compiler change? (i.e. did I miss that in my change?)

billti (approved):

npm/src/browser.ts
minestarks: It's to match the compiler change, yes. It _was_ necessary to get logging working though (since logging is initialized within `instantiateWasm`).

PR #531, opened 2023-08-04
Create quantum "Hello world!" sample

cesarzc:
This change creates a quantum "Hello World!" sample which is intended to be a first example of Q# code for people so it is very thoroughly commented to explain each element of the program.


viktorveis (approved):

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,24 @@
+/// # Quantum Hello World!
+///
+/// This Q# code generates a random bit by setting a qubit in a superposition of the
+/// computational basis states |0〉 and |1〉, and returning the measurement result.
+namespace QuantumHelloWorld {
+    // Import the `Message` function and the `H` operation used in the program.
+    open Microsoft.Quantum.Intrinsic;
```
swernli, line 7: Technically this import is no longer required, though it doesn't hurt. The Intrinsic namespace is now part of the default prelude.

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,24 @@
...
+
+    @EntryPoint()
+    operation RandomBit() : Result {
+        Message("Hello world!");
```
swernli, line 11: In the spirit of commenting on everything, it might be worth an explanatory comment here that `Message` is how Q# programs print to stdout.

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,24 @@
+/// # Quantum Hello World!
+///
+/// This Q# code generates a random bit by setting a qubit in a superposition of the
+/// computational basis states |0〉 and |1〉, and returning the measurement result.
+namespace QuantumHelloWorld {
+    // Import the `Message` function and the `H` operation used in the program.
```
swernli, line 6: If we keep the import, the comment would benefit from additional info, since this is also where `M` and `Reset` come from.

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,24 @@
...
+    open Microsoft.Quantum.Intrinsic;
+
+    @EntryPoint()
+    operation RandomBit() : Result {
+        Message("Hello world!");
+        use qubit = Qubit();
```
swernli, line 12: A comment about qubit allocation and automatic scoping could be helpful here.

samples/algorithms/QuantumHelloWorld.qs
viktorveis: If it's no longer required, I think that it would be better to remove it.

samples/algorithms/QuantumHelloWorld.qs
viktorveis: I suggested to Cesar to make comments more succinct to keep Hello World as lean as possible. We plan to make this sample as default code on quantum.microsoft.com. The goal here is to show how easy it is to start with Q#, so we can skip commenting on code that is intuitive for non-quantum developers.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: Removed the `open` statement.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: Ended up removing the `open` statement based on a previous suggestion.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: Added a brief comment on qubit allocation and scoping.

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,27 @@
+/// # Quantum Hello World!
+///
+/// This Q# code generates a random bit by setting a qubit in a superposition of
+/// the computational basis states |0〉 and |1〉, and returning the measurement
+/// result.
```
viktorveis, line 5: Nit: this comment can be more compact and scannable by moving "result." to the previous line. We don't have to be religious with not exceeding 80 characters per line.

samples/algorithms/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,27 @@
...
+
+        // Qubits are only accesible for the duration of the scope where they
+        // are allocated and are automatically released at the end of the scope.
+        use qubit = Qubit();
+
+        // Set the qubit in superposition by applying a Hadamard transformation.
+        H(qubit);
+
+        // Measure the qubit. There is a 50% probability of measuring either
+        // `Zero` or `One`.
```
viktorveis, line 20: Consider making this one-line comment.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: There will be a "Random Bit" specific sample later so I'll include this comment there.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: Updated.

samples/algorithms/QuantumHelloWorld.qs
cesarzc: Updated.

swernli (approved):

PR #537, opened 2023-08-04
Looping Samples

ScottCarda-MS:
Adds samples for describing looping syntax in Q#.


samples/language/RepeatUntilLoops.qs
```diff
@@ -0,0 +1,38 @@
+/// # Sample
+/// Repeat-Until Loops
+///
+/// # Description
+/// The repeat loop runs a block of code before evaluating a condition.
```
cesarzc, line 5: nit: if we are referring to this as a "repeat-until loop", it should always be consistent, i.e. this sentence should be "The repeat loop runs a block of code before evaluating a condition.".

samples/language/WhileLoops.qs
```diff
@@ -0,0 +1,18 @@
+/// # Sample
+/// While Loops
+///
+/// # Description
+/// While-loops are a basic looping looping structure in Q# for repeatedly
```
cesarzc, line 5: nit: similar to Alex's comment, here it should probably be "While loops" instead of "While-loops".

cesarzc (approved):

samples/language/ForLoops.qs
```diff
@@ -0,0 +1,22 @@
+/// # Sample
+/// For Loops
+///
+/// # Description
+/// For-loops are a basic looping structure in Q# for looping
```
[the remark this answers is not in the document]
ScottCarda-MS: done

samples/language/RepeatUntilLoops.qs
ScottCarda-MS: done

PR #538, opened 2023-08-04
Mapping debugger utf8 unit offsets to utf16 units

idavis:


wasm/src/debug_service.rs
```diff
@@ -163,8 +169,15 @@ pub struct BreakpointSpan {
 
 #[wasm_bindgen(typescript_custom_section)]
 const IStackFrameList: &'static str = r#"
+export interface IStackFrame {
```
billti, line 172: If you're going to export these as explicit interfaces in these custom sections, do the actual structs (StackFrame and BreakpointSpan) still need to be exposed via wasm_bindgen and pub?

npm/src/debug-service/debug-service.ts
```diff
@@ -30,18 +31,40 @@ export type IDebugServiceWorker = IDebugService & IServiceProxy;
...
-    const stack_frames: StackFrame[] = stack_frame_list.frames;
+
+    const stack_frames: IStackFrame[] = stack_frame_list.frames.map(
+      (frame: IStackFrame) => {
+        if (frame.path in this.code) {
+          const mappedSpan = mapUtf8UnitsToUtf16Units(
+            [frame.lo, frame.hi],
+            this.code[frame.path]
+          );
+          const result = {} as IStackFrame;
```
billti, line 57: A more canonical way to do lines 57 - 62 in JavaScript would be to use the object spread operator (https://stackoverflow.com/a/49491435). i.e. replace all those lines with `return {...frame, lo: mappedSpan[frame.lo], hi: mappedSpan[frame.hi]};`

billti (approved):

wasm/src/debug_service.rs
idavis: Good call, removing them.

PR #539, opened 2023-08-07
Fix Python wheel being overwritten by local install

minestarks:
The editable install (`pip install -e .`) step was somehow overwriting the `.whl` file that was built in the previous step. I changed it to do a wheel install instead of a local install.

Also removed the JupyterLab extension integration test since it can fail in some environments, due to the specific way `build.py` sets up the virtual environment. The JupyterLab application directory is based on `sys.prefix` which can change depending on where/how the script is run. I could not find a reliable way to get the test working, so I decided to take it out entirely.

In the future, it would be nice to have an integration test run that sets up a proper environment and does more thorough validation. Until then, I think we can live without this one test, which was of dubious value anyway.


swernli (approved):

kuzminrobin (approved):

kuzminrobin:
Feel free to merge changes from main to this branch and enable auto-merge.

billti (approved):

build.py
```diff
@@ -185,10 +185,21 @@
...
         subprocess.run(pip_install_args, check=True, text=True, cwd=pip_src)
-        pip_install_args = [python_bin, "-m", "pip", "install", "-e", "."]
+        pip_install_args = [
+            python_bin,
+            "-m",
+            "pip",
+            "install",
+            "--force-reinstall",
+            "--no-index",
+            "--find-links=" + wheels_dir,
```
idavis, line 195: This will stop working when the wheel is published.

https://pip.pypa.io/en/stable/cli/pip_install/#finding-packages

The path to the wheel file needs to be used:
```
pip install --force-reinstall ./path/to/package.whl
```

idavis (changes requested):

build.py
```diff
@@ -185,10 +185,21 @@
...
+            "qsharp",
```
kuzminrobin, line 196: For me the `python build.py --pip` fails without the following:
```suggestion
            "qsharp_python",
```

minestarks: That's odd. I wonder if the package name is getting picked up from Cargo.toml (instead of pyproject.toml) somehow. I'll take another look.

build.py
minestarks: Hmmm. I think I'll take a look again tomorrow. The problem is that I can't figure out how to get the full path to the .whl, since the filename is generated based on the version/platform. `--no-index` is supposed to ignore PyPi but at this point I don't trust my understanding.

build.py
kuzminrobin: Oh, that was my local change. I forgot to mention that I rebased my local changes on top of yours. And in my changes I made the name identical in Cargo.toml and pyproject.toml. 
My bad, disregard.

build.py
kuzminrobin: For the time being is it possible to use the wildcards instead of exact file names? E.g., "qsharp_python-*.*.*-cp37-abi3-*.whl".
Sometime later we plan to have a single tool to specify/update the version synchronously in all the .toml, .yml, etc. (likely including this "build.py").

build.py
minestarks: Thanks for clarifying. The [documentation for the build backend](https://github.com/PyO3/maturin#python-metadata) (Maturin) says that it merges metadata from Cargo.toml and pyproject.toml, with pyproject.toml taking precedence. So, for the purposes of the Python package name, the name in Cargo.toml should be getting ignored.

I believe the reason I ended up changing the name in Cargo.toml to `qsharp_python` is that this package is also referenced from the Rust workspace, and we didn't want a generic name like `qsharp` in that context. ([see original PR](https://github.com/microsoft/qsharp/pull/183/files/ea8def5e1c3a5faf4a554bc47e3625bebe27c65d#r1178521419)). In practice, I don't think the name in Cargo.toml is going to matter too much since we're never going to publish this Rust crate by itself.

PR #540, opened 2023-08-07
Add Step Next/In/Out to Debugger

idavis:
Debug stepping ~has introduced enough overhead that it~ may be worthwhile to split eval_continue into two calls. There is a lot going on in the eval for debugging. We will probably want to take a look after scopes support is added. Step granularity on expressions may also introduce noticeable overhead, but that is speculation at the moment.

```
Teleport evaluation     time:   [103.03 µs 103.08 µs 103.14 µs]
                        change: [+0.3841% +0.6847% +0.9592%] (p = 0.00 < 0.05)

Deutsch-Jozsa evaluation
                        time:   [22.510 ms 22.517 ms 22.525 ms]
                        change: [+0.6672% +0.8058% +0.8961%] (p = 0.00 < 0.05)
```

### DIFF
--- a/compiler/qsc/src/interpret/stateful/stepping_tests.rs
+++ b/compiler/qsc/src/interpret/stateful/stepping_tests.rs
@@ -141,7 +141,7 @@ fn step_in(
     Result<StepResult, Vec<crate::interpret::stateful::Error>>,
     String,
 ) {
-    step(interpreter, breakpoints, &qsc_eval::StepAction::In)
+    step(interpreter, breakpoints, qsc_eval::StepAction::In)
 }
 
 fn step_next(
@@ -151,7 +151,7 @@ fn step_next(
     Result<StepResult, Vec<crate::interpret::stateful::Error>>,
     String,
 ) {
-    step(interpreter, breakpoints, &qsc_eval::StepAction::Next)
+    step(interpreter, breakpoints, qsc_eval::StepAction::Next)
 }
 
 fn step_out(
@@ -161,13 +161,13 @@ fn step_out(
     Result<StepResult, Vec<crate::interpret::stateful::Error>>,
     String,
 ) {
-    step(interpreter, breakpoints, &qsc_eval::StepAction::Out)
+    step(interpreter, breakpoints, qsc_eval::StepAction::Out)
 }
 
 fn step(
     interpreter: &mut Interpreter,
     breakpoints: &[StmtId],
-    step: &StepAction,
+    step: StepAction,
 ) -> (
     Result<StepResult, Vec<crate::interpret::stateful::Error>>,
     String,
--- a/compiler/qsc/src/interpret/stateful/stepping_tests.rs
+++ b/compiler/qsc/src/interpret/stateful/stepping_tests.rs
@@ -1,0 +1,208 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::interpret::stateful::Interpreter;
+use qsc_eval::{output::CursorReceiver, StepAction, StepResult};
+use qsc_fir::fir::StmtId;
+use qsc_frontend::compile::SourceMap;
+use qsc_passes::PackageType;
+use std::io::Cursor;
+
+#[cfg(test)]
+mod given_interpreter_with_sources {
+    use super::*;
+
+    static STEPPING_SOURCE: &str = r#"
+        namespace Test {
+            @EntryPoint()
+            operation A() : Int {
+                let d = B();
+                let e = d / 1;
+                e
+            }
+            operation B() : Int {
+                let g = 10;
+                let h = 20;
+                let l = C(g, h);
+                42
+            }
+            operation C(m: Int, n: Int) : Int {
+                let o = 42 - (m + n);
+                let p = (m + n) + o;
+                p
+            }
+        }"#;
+    #[cfg(test)]
+    mod step {
+        use super::*;
+
+        #[test]
+        fn in_one_level_operation_works() -> Result<(), Vec<crate::interpret::stateful::Error>> {
+            let sources = SourceMap::new([("test".into(), STEPPING_SOURCE.into())], None);
+            let mut interpreter = Interpreter::new(true, sources, PackageType::Exe)?;
+            interpreter.set_entry()?;
+            let ids = get_breakpoint_ids(&interpreter, "test");
+            let expected_id = ids[0];
+            expect_bp(&mut interpreter, &ids, expected_id);
+            expect_in(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            let expected = "42";
+            expect_return(interpreter, expected);
+            Ok(())
+        }
+
+        #[test]
+        fn next_crosses_operation_works() -> Result<(), Vec<crate::interpret::stateful::Error>> {
+            let sources = SourceMap::new([("test".into(), STEPPING_SOURCE.into())], None);
+            let mut interpreter = Interpreter::new(true, sources, PackageType::Exe)?;
+            interpreter.set_entry()?;
+            let ids = get_breakpoint_ids(&interpreter, "test");
+            let expected_id = ids[0];
+            expect_bp(&mut interpreter, &ids, expected_id);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            let expected = "42";
+            expect_return(interpreter, expected);
+            Ok(())
+        }
+
+        #[test]
+        fn in_multiple_operations_works() -> Result<(), Vec<crate::interpret::stateful::Error>> {
+            let sources = SourceMap::new([("test".into(), STEPPING_SOURCE.into())], None);
+            let mut interpreter = Interpreter::new(true, sources, PackageType::Exe)?;
+            interpreter.set_entry()?;
+            let ids = get_breakpoint_ids(&interpreter, "test");
+            let expected_id = ids[0];
+            expect_bp(&mut interpreter, &ids, expected_id);
+            expect_in(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_in(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            let expected = "42";
+            expect_return(interpreter, expected);
+            Ok(())
+        }
+
+        #[test]
+        fn out_multiple_operations_works() -> Result<(), Vec<crate::interpret::stateful::Error>> {
+            let sources = SourceMap::new([("test".into(), STEPPING_SOURCE.into())], None);
+            let mut interpreter = Interpreter::new(true, sources, PackageType::Exe)?;
+            interpreter.set_entry()?;
+            let ids = get_breakpoint_ids(&interpreter, "test");
+            let expected_id = ids[0];
+            expect_bp(&mut interpreter, &ids, expected_id);
+            expect_in(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_next(&mut interpreter);
+            expect_in(&mut interpreter);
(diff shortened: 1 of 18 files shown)


compiler/qsc_eval/src/lib.rs
```diff
@@ -424,44 +491,70 @@ impl State {
...
+                            && current_frame == self.call_stack.len()
+                        {
+                            StepResult::Next
+                        } else if matches!(step, StepAction::Out)
+                            && current_frame > self.call_stack.len()
+                        {
+                            StepResult::StepOut
+                        } else {
+                            continue;
+                        }
```
swernli, line 545: Since `StepResult` is a simple enum, these could use equality comparison instead of matches if you derive `PartialEq`, ie:

```suggestion
                        if step == StepAction::In {
                            StepResult::StepIn
                        } else if step == StepAction::Next
                            && current_frame == self.call_stack.len()
                        {
                            StepResult::Next
                        } else if step == StepAction::Out
                            && current_frame > self.call_stack.len()
                        {
                            StepResult::StepOut
                        } else {
                            continue;
                        }
```

compiler/qsc_eval/src/lib.rs
```diff
@@ -154,26 +139,55 @@ pub fn eval_expr(
...
+    state.push_stmt(stmt);
+    let res = state.eval(globals, env, sim, receiver, &[], &StepAction::Continue)?;
+    let StepResult::Return(value) = res else {
+        panic!("eval_stmt should always return a value");
+    };
+    Ok(value)
+}
+
+/// The type of step action to take during evaluation
+pub enum StepAction {
```
swernli, line 172: This could have a derive to make other code simpler:

```suggestion
/// The type of step action to take during evaluation
#[derive(PartialEq)]
pub enum StepAction {
```

compiler/qsc_eval/src/tests.rs
```diff
@@ -30,8 +30,11 @@ pub(super) fn eval_expr(
     let mut env = Env::with_empty_scope();
     let mut sim = SparseSim::new();
     state.push_expr(expr);
-    state.resume(globals, &mut env, &mut sim, out, &[])?;
-    Ok(state.pop_val())
+    let res = state.eval(globals, &mut env, &mut sim, out, &[], &StepAction::Continue)?;
+    match res {
+        StepResult::Return(value) => Ok(value),
+        _ => unreachable!("eval_expr should always return a value"),
+    }
```
swernli, line 37: minor nit: this could be written with a let-else pattern.

compiler/qsc_eval/src/lib.rs
```diff
@@ -1108,8 +1201,10 @@ impl State {
                 let scope = env.0.last_mut().expect("binding should have a scope");
                 match scope.bindings.entry(variable.id) {
                     Entry::Vacant(entry) => entry.insert(Variable {
+                        name: variable.name.clone(),
```
swernli, line 1204: Food for thought: right now the type string displayed in the debugger comes from the value, which will just give you the top level type (ie: `Tuple` instead of `(Int, Qubit[])`). If you wanted the full type you could get it here with something like `ty_name: pat.ty.to_string(),` and then bubbling that up from the locals instead of using the `value.ty_name()`. That would cause an allocation of the string on each binding, which would have a perf impact, so not sure if there is a better way to accomplish that of if the benefit would be worth the cost.

wasm/src/debug_service.rs
```diff
@@ -195,3 +313,43 @@ struct StackFrame {
     pub lo: u32,
     pub hi: u32,
 }
+
+#[wasm_bindgen(typescript_custom_section)]
+const IVariableList: &'static str = r#"
+export interface IVariable {
+    name: string;
+    value: string;
+    var_type: "Array"
```
swernli, line 322: As mentioned in the other comment, it's possible this could just become `string` if you allow the full type to propagate out of the evaluator.

swernli (approved):
A few minor comments, but otherwise the Rust code looks good! Signing off on those components.

vscode/src/common.ts
```diff
@@ -3,6 +3,7 @@
 
 export const qsharpLanguageId = "qsharp";
 export const qsharpExtensionId = "qsharp-vscode";
+export const QsLibraryUriScheme = "qsharp-source-request";
```
minestarks, line 6:
```suggestion
export const qsharpLibraryUriScheme = "qsharp-library-source";
```

Since you said you were looking for naming suggestions :)

wasm/src/lib.rs
```diff
@@ -28,6 +28,31 @@ pub fn git_hash() -> JsValue {
     JsValue::from_str(git_hash)
 }
 
+#[wasm_bindgen]
+pub fn provide_text_document_content(name: &str) -> JsValue {
```
minestarks, line 32:
```suggestion
pub fn get_library_source_content(name: &str) -> JsValue {
```

Suggest name to parallel my uri scheme name suggestion.

wasm/src/lib.rs
```diff
@@ -28,6 +28,31 @@ pub fn git_hash() -> JsValue {
...
+    thread_local! {
+        static STORE_CORE_STD: (PackageStore, PackageId) = {
+            let mut store = PackageStore::new(compile::core());
+            let std = store.insert(compile::std(&store));
+            (store, std)
+        };
+    }
```
minestarks, line 39: Since this is the same thing that the `compile` function does below, can we just pull this out into the global scope in this file so that the two functions can share the instance?

compiler/qsc_eval/src/lib.rs
```diff
@@ -267,20 +294,61 @@ impl Env {
...
+            .collect();
+
+        let variables_by_scope: Vec<Vec<VariableInfo>> = candidate_scopes
+            .into_iter()
+            .map(|bindings| {
+                bindings
+                    .map(|(id, var)| VariableInfo {
+                        id: *id,
+                        name: var.name.clone(),
+                        type_name: var.value.type_name().to_string(),
```
minestarks, line 336: For complex types, how is the type name formatted? We have custom formatters for UDTs and other complex types in `display.rs` in the language service. They are used in Hover today, and the code should probably be shared so that we get consistent formatting across the editor features.

wasm/src/debug_service.rs
idavis: We'll take another look at this later. Straightforward solutions slow down execution 2-7%.

compiler/qsc/src/interpret/stateful/stepping_tests.rs
```diff
@@ -0,0 +1,208 @@
...
+                let h = 20;
+                let l = C(g, h);
+                42
+            }
+            operation C(m: Int, n: Int) : Int {
+                let o = 42 - (m + n);
+                let p = (m + n) + o;
+                p
+            }
+        }"#;
```
billti, line 34: Do we have any test-cases that include non-ascii chars - being that we had an issue with those that required the mapping stuff be added. Might be a good safety net.

vscode/src/debugger/session.ts
```diff
@@ -190,64 +209,90 @@ export class QscDebugSession extends LoggingDebugSession {
...
+      async () => await this.debugService.evalStepOut(bps, this.eventTarget)
+    );
+  }
+
+  private async endSession(message: string, exitCode: number): Promise<void> {
     log.trace(message);
-    vscode.debug.activeDebugConsole.appendLine("");
-    vscode.debug.activeDebugConsole.appendLine(SimulationCompleted);
+    this.writeToStdOut("");
+    this.writeToStdOut(SimulationCompleted);
```
minestarks, line 283: May be me being picky, but I wouldn't expect anything printed to stdout that isn't coming from the program itself (like `Message()` output.) Maybe this could go to the debug console?

npm/src/main.ts
```diff
@@ -31,6 +31,15 @@ type Wasm = typeof import("../lib/node/qsc_wasm.cjs");
 let wasm: Wasm | null = null;
 const require = createRequire(import.meta.url);
 
+export async function provideTextDocumentContent(
```
billti, line 34: Minor nit - but you don't need the `async` keyword here. The function body doesn't `await` anything.

vscode/src/debugger/session.ts
```diff
@@ -472,52 +540,82 @@ export class QscDebugSession extends LoggingDebugSession {
...
+            );
+            return sf;
+          } else {
+            try {
+              // This file isn't part of the workspace, so we'll
+              // create a URI which can try to load it from the core and std lib
+              // There is a custom content provider subscribed to this scheme.
+              // Opening the text document by that uri will use the content
+              // provider to look for the source code.
+              const uri = vscode.Uri.parse(QsLibraryUriScheme + ":" + f.path);
```
minestarks, line 580:
```suggestion
              const uri = vscode.Uri.from({ scheme: qsharpLibraryUriScheme, path: f.path });
```

minestarks (approved):
Mostly LGTM with a few suggestions

npm/src/main.ts
billti: (Though I guess that makes it easier to return the result as a Promise, so ignore this)

wasm/src/debug_service.rs
```diff
@@ -195,3 +313,43 @@ struct StackFrame {
...
+        | "Double"
+        | "Global"
+        | "Int"
+        | "Pauli"
+        | "Qubit"
+        | "Range"
+        | "Result"
+        | "String"
+        | "Tuple";
+}
```
billti, line 335: Is there any issue with UDTs here?

idavis: They show up as Tuple. We can enhance this later. Stefan and I tried some stuff but the perf hit was significant.

vscode/src/extension.ts
```diff
@@ -142,3 +149,16 @@ async function loadLanguageService(baseUri: vscode.Uri) {
...
+export class QsTextDocumentContentProvider
+  implements vscode.TextDocumentContentProvider
+{
+  onDidChange?: vscode.Event<vscode.Uri> | undefined;
+  provideTextDocumentContent(
+    uri: vscode.Uri,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    token: vscode.CancellationToken
+  ): vscode.ProviderResult<string> {
+    return provideTextDocumentContent(uri.path);
```
billti, line 162: Do we only store docs internally via the path, not the full Uri? This could be an issue later if we support referencing docs via other protocols (e.g. I have in my project "src/foo.qs" and "http://somesite.com/src/foo.qs" then do they have the same path). Is it possible for the path part of the std lib files to conflict with user files?

wasm/src/debug_service.rs
```diff
@@ -141,6 +201,64 @@ impl Default for DebugService {
...
+                id: StepResultId::Return.into(),
+                value: 0,
+            },
+        }
+    }
+}
+
+#[wasm_bindgen]
+#[derive(Copy, Clone, Serialize, Deserialize)]
+pub enum StepResultId {
```
billti, line 233: In general, we should avoid exporting enums at the wasm layer. There's no such thing in JavaScript, and TypeScript does it by exporting a value (a class that maps names to numbers) in the .d.ts, which then adds a bunch of footguns with importing platform specific JS in the platform agnostic code. Here I see you're only using this value in the platform specific 'browser.ts', so not currently an issue. A note for the future though, as we likely want guidance NOT to do this in general.

billti (approved):

vscode/src/extension.ts
idavis: This provider is only used when resolving core/std library files as a fallback for having missing source. Any local files should load just fine.

compiler/qsc/src/interpret/stateful/stepping_tests.rs
idavis: The offsets are never used when running. We use distinct int ids for breakpoints. The utf8=>utf16 is only done in the JS layer for stack frames and breakpoint ui location mapping. The spans are never sent back into the compiler.

PR #555, opened 2023-08-09
Unified tutorial overviews to include TOC and prerequisites

DmitryVasilevsky:
All tutorials now include overview with the list of covered subjects and the list of prerequisites. I.e. the following sections:
**This tutorial covers the following topics**
**What you should know to start working on this tutorial**

(Didn't touch Multi-Qubit gates to avoid conflicts)

### DIFF
--- a/katas/content/multi_qubit_measurements/index.md
+++ b/katas/content/multi_qubit_measurements/index.md
@@ -9,18 +9,22 @@ In the previous kata, we discussed the concept of measurements done on single-qu
 Building upon those ideas, this kata will introduce you to measurements done on multi-qubit systems, and how to implement such measurements in Q#.
 This will include measuring a single qubit in a multi-qubit system, as well as measuring multiple qubits simultaneously.
 
-We recommend to go through the kata that introduces single-qubit system measurements before starting this one.
+**This tutorial covers the following topics:**
+
+- Measuring a single qubit in a multi-qubit system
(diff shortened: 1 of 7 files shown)


katas/content/multi_qubit_measurements/index.md
```diff
@@ -9,18 +9,22 @@ In the previous kata, we discussed the concept of measurements done on single-qu
 Building upon those ideas, this kata will introduce you to measurements done on multi-qubit systems, and how to implement such measurements in Q#.
 This will include measuring a single qubit in a multi-qubit system, as well as measuring multiple qubits simultaneously.
 
-We recommend to go through the kata that introduces single-qubit system measurements before starting this one.
-$\renewcommand{\ket}[1]{\left\lvert#1\right\rangle}$
-$\renewcommand{\bra}[1]{\left\langle#1\right\rvert}$
+**This tutorial covers the following topics:**
```
cesarzc, line 12: nit: Matt has mentioned that we should use the term "kata" instead of "tutorial" throughout the content. This is not something that has to be done in this PR though. It's already captured in the feedback for each kata so we can address it separately.

cesarzc (approved):

katas/content/multi_qubit_measurements/index.md
DmitryVasilevsky: Let's do it as a separate PR.

PR #557, opened 2023-08-10
Run new samples in CI

### LINKED ISSUE #554: Some language samples do not compile
The new language samples have a few compilation errors:

- Namespaces.qs uses `ConstantArray` from the `Microsoft.Quantum.Arrays` to demonstrate usage of namespaces, but `ConstantArray` doesn't exist. The equivalent is `Repeated` but that is in the canon namespace which doesn't need to be opened. We should pick a different example, perhaps something like `Microsoft.Quantum.Math.PI()` since that is a commonly used function.
- RepeatUntilLoops.qs does not declare the variable `x` before using it, and the second loop example tries to use variables outside of the scope in which they are declared (`result` and `qubit` need to be declared before the loop).
- Specializations.qs is the only one that doesn't include an entry point, which can be a bit confusing since the language server treats that as an error. If it were compiled as a library it would be fine, but we don't have that setting yet and it's a noticeable deviation from the others, so it probably makes sense to just include one that calls the defined operations as an example of using specializations.

Some of this would be helped by having a test somewhere that includes compiling these files. As it is, we are at risk of missing typos or changes that break compilation.

sezna:
Closes #554

### DIFF
--- a/katas/content/multi_qubit_measurements/index.md
+++ b/katas/content/multi_qubit_measurements/index.md
@@ -9,18 +9,22 @@ In the previous kata, we discussed the concept of measurements done on single-qu
 Building upon those ideas, this kata will introduce you to measurements done on multi-qubit systems, and how to implement such measurements in Q#.
 This will include measuring a single qubit in a multi-qubit system, as well as measuring multiple qubits simultaneously.
 
-We recommend to go through the kata that introduces single-qubit system measurements before starting this one.
+**This tutorial covers the following topics:**
+
+- Measuring a single qubit in a multi-qubit system
+- Measuring multiple qubits simultaneously
+- how to implement such measurements in Q#
+
+**What you should know to start working on this tutorial:**
+
+- Basic linear algebra
+- Single and multi-qubit systems
+- Single and multi-qubit gates
+- Single-qubit system measurements
+
 $\renewcommand{\ket}[1]{\left\lvert#1\right\rangle}$
 $\renewcommand{\bra}[1]{\left\langle#1\right\rvert}$
-
-You should be familiar with the following concepts before tackling the Single-Qubit System Measurements tutorial (and this workbook):
-
-1. Basic linear algebra
-2. Single and multi-qubit systems
-3. Single and multi-qubit gates
-   $\renewcommand{\ket}[1]{\left\lvert#1\right\rangle}$
-   $\renewcommand{\bra}[1]{\left\langle#1\right\rvert}$
-
+ 
 @[section]({
     "id": "multi_qubit_measurements_types_of_measurements",
     "title": "Types of measurements"
--- a/katas/content/multi_qubit_systems/index.md
+++ b/katas/content/multi_qubit_systems/index.md
@@ -9,16 +9,16 @@ This kata introduces you to multi-qubit systems - their representation in mathem
 
 If you are not familiar with single-qubit systems, we recommend that you complete the qubit kata first.
 
-**This kata covers the following topics:**
+**This tutorial covers the following topics:**
 
-* Vector representation of multi-qubit systems
-* Entangled and separable states
-* Dirac notation
+- Vector representation of multi-qubit systems
+- Entangled and separable states
+- Dirac notation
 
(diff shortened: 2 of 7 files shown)


samples/language/Namespaces.qs
```diff
@@ -8,13 +8,13 @@
...
-    open Microsoft.Quantum.Arrays;
+    // Microsoft.Quantum.Diagnostics namespace.
+    open Microsoft.Quantum.Diagnostics;
 	
     @EntryPoint()
     operation Main() : Result[] {
-        // `ConstantArray` is in the Microsoft.Quantum.Arrays namespace
-        let array = ConstantArray(3, true);
+        // `Message` is in the Microsoft.Quantum.Diagnostics namespace
+        Message("Hello World!");
```
swernli, line 17: This is a weird one, because `Message` is technically in the `Microsoft.Quantum.Intrinsic` namespace and is another callable that is part of the prelude and always available. If you want something from `Microsoft.Quantum.Diagnostics` you should pick `DumpMachine` instead.

build.py
```diff
@@ -293,3 +296,12 @@
...
     subprocess.run(pip_build_args, check=True, text=True, cwd=jupyterlab_src)
+
+if compile_samples:
+    print("Compiling qsharp samples")
+    files = [os.path.join(dp, f) for dp, dn, filenames in os.walk("samples") for f in filenames if os.path.splitext(f)[1] == '.qs']
+    args = ["cargo", "run", "--bin", "qsc"]
+    if build_type == "release":
+        args.append("--release")
+    for file in files:
+        subprocess.run((args + [file]), check=True, text=True, cwd=root_dir)
```
swernli, line 307: Do you need one more `--` between the args and the file list?

swernli (changes requested):
The namespaces sample compiles but is still not quite correct.

swernli (approved):
Changes look good and CI passes!

billti:
As I mentioned in person, the nodejs unit tests already have a method to test all the algorithm samples compile. (https://github.com/microsoft/qsharp/blob/main/npm/test/basics.js#L307). Probably would have been more consistent and faster to execute to expand on that, rather than add even more build script logic and spin up a new process for every sample. Done now I guess...

billti:
@sezna Also, I just noticed that because you default to running this on `args.test`, which is on by default, even if I'm just compiling with `--wasm` or building the playground, this causes the native CLI binaries to be built and the samples tests to be run (due to `cargo run --bin`). Please fix this up.

cesarzc:
> As I mentioned in person, the nodejs unit tests already have a method to test all the algorithm samples compile. (https://github.com/microsoft/qsharp/blob/main/npm/test/basics.js#L307). Probably would have been more consistent and faster to execute to expand on that, rather than add even more build script logic and spin up a new process for every sample. Done now I guess...

@sezna and I talked offline about this, and I made the suggestion to build the samples from build.py. The reasoning for that is the following:
- IMO, tests in basics.js are meant to test the APIs and content in the npm package. Some of the samples are part of that content, but not all of them. We don't necessarily want to have all the 70+ samples (that we will soon have) to be part of the package and showing in the playground.
    - Assuming that not all samples are part of the npm package, we need to validate that all of them build somewhere and build.py looks like a good place to do that. Building the samples is part of building the repo after all.
    - Alternatively, we could have done it in basics.js by traversing the samples folder from there. However, it seemed out of place because it's "testing" something not directly related to the npm package. It's still valid to test the samples that are part of the npm package from there since it also validates that they were correctly "packaged".
- Ideally, we would like to automatically validate a new sample in the folder without having to add it to samples.mjs, which we could forget to do, and we would end up merging samples that do not build to `main`.

cesarzc:
> @sezna Also, I just noticed that because you default to running this on `args.test`, which is on by default, even if I'm just compiling with `--wasm` or building the playground, this causes the native CLI binaries to be built and the samples tests to be run (due to `cargo run --bin`). Please fix this up.

Noticed the same and opened this PR: https://github.com/microsoft/qsharp/pull/569

PR #494, opened 2023-07-20
Add QIR base profile validation pass

swernli:
This change adds an analysis pass that will check whether a given program conforms to the QIR base profile. Specifically, this disallows:

 - Comparison of `Result` values
 - `Result` literals
 - Operation intrinsics that don't return Result or Unit
 - Entry point callables or expressions that return types other than `Result` or aggregates of `Result`

The logic for the pass with validate the entire package provided, but then only validate callables from dependent packages that are called somewhere in the given package. This allows for use of a subset of the standard library without producing extraneous errors.

The new pass is not run by default, but will be triggered if the compiler is invoked with qir emission, ie: `qsc <file> --emit qir`.

PR #559, opened 2023-08-10
Python code generation and submit to Azure

minestarks:
Sample showing QIR generation and submission to Azure quantum

This still depends on a local `qat.exe` , so obviously not even close to ready for merging

This PR should be updated once #494 is merged.

### DIFF
--- a/compiler/qsc/src/interpret/stateful.rs
+++ b/compiler/qsc/src/interpret/stateful.rs
@@ -14,6 +14,7 @@ use crate::{
 use miette::Diagnostic;
 use num_bigint::BigUint;
 use num_complex::Complex;
+use qsc_codegen::qir_base::BaseProfGen;
 use qsc_data_structures::index_map::IndexMap;
 use qsc_eval::backend::Backend;
 use qsc_eval::{
@@ -92,14 +93,19 @@ impl LineError {
 }
 
 #[derive(Clone, Debug, Diagnostic, Error)]
-#[diagnostic(transparent)]
 pub enum LineErrorKind {
     #[error(transparent)]
+    #[diagnostic(transparent)]
     Compile(#[from] incremental::Error),
     #[error(transparent)]
+    #[diagnostic(transparent)]
     Pass(#[from] qsc_passes::Error),
     #[error("runtime error")]
+    #[diagnostic(transparent)]
     Eval(#[from] qsc_eval::Error),
+    #[error("code generation target mismatch")]
+    #[diagnostic(code("Qsc.Interpret.TargetMismatch"))]
+    TargetMismatch,
 }
 
 struct Lookup<'a> {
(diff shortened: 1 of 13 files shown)


compiler/qsc_passes/src/baseprofck.rs
```diff
@@ -0,0 +1,132 @@
...
+    hir::{
+        BinOp, CallableDecl, CallableKind, Expr, ExprKind, Item, ItemKind, Lit, Package, SpecBody,
+        SpecGen, Stmt,
+    },
+    ty::{Prim, Ty},
+    visit::{walk_expr, walk_item, Visitor},
+};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
```
billti, line 19: There's a lot more changes to internals than I was expecting. Or is this just work from Stefan's  branch that isn't merged into main in those PRs yet?

pip/qsharp/_qsharp.py
```diff
@@ -34,3 +52,51 @@ def interpret_file(path) -> None:
...
+def compile(entry_expr):
+    """
+    Compiles Q# into a job for submission to Azure
+    """
+    ll_str = _interpreter.qir(entry_expr)
+    return QirWrapper("main", ll_str)
+
+
+interim_ll_path = "C:\\temp\\interim.ll"
+interim_bc_path = "C:\\temp\\interim.bc"
```
billti, line 66: This is gonna work great on my mac :-p

compiler/qsc_passes/src/baseprofck.rs
billti: Ah, I see you mention 494 in the description. Never mind...

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -414,6 +422,90 @@ impl Interpreter {
...
+            return Err(vec![LineError(WithSource::new(
+                line.into(),
+                LineErrorKind::TargetMismatch,
+                None,
+            ))]);
+        }
+
+        let mut codegen = BaseProfGen::new();
+
+        let mut fragments = self.compiler.compile_fragments(line).map_err(|errors| {
```
minestarks, line 440: I think we should specifically use the expr parser and reject anything that's not an expression here. IMO it makes for a really weird user experience if we allow the user to pass in statements / namespace declarations inside `qsharp.compile(xxx)`. The signature at the Python layer currently defines this as an "entry expression" which makes sense to me from a user's perspective.

If the user wants to declare additional operations or whatever, they can do that outside of `compile()` , in a regular `%%qsharp` cell or `qhsarp.eval()` call.

I'm also motivated by the fact that we plan to add `qsharp.eval_with_shots(xxx)`, and that will also be restricted to expressions (to keep things sane). It would make a coherent api if these two functions had similar calling patterns.

swernli: That sounds like a good, cohesive story. It should be an easy change to make in the code that will simplify things quite a bit, too. Let me try that in the other branch, and we can see if that one is clean enough to warrant a PR to supersede this one.

minestarks:
Superseded by #588

PR #560, opened 2023-08-11
Qubit Kata - Updated content, fixed code

DmitryVasilevsky:
Removed sections on functionality that we don't support.
Simplified and re-arranged content.
Updated verification code to pass proper states to user code.
Added hints on how to complete exercises.


cesarzc (approved):
The updated content with the hints looks great!

PR #570, opened 2023-08-12
Add more basics of quantum computing samples

cesarzc:


samples/algorithms/BellState.qs
```diff
@@ -7,14 +7,12 @@
...
 namespace Sample {
-    open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Diagnostics;
 
     @EntryPoint()
     operation BellState() : (Result, Result) {
         // Allocate the two qubits that will be used to create a Bell state.
-        use q1 = Qubit();
-        use q2 = Qubit();
+        use (q1, q2) = (Qubit(), Qubit());
```
DmitryVasilevsky, line 15: BellState is one of the first samples we have. Do we want to use tuples right from the start? I think it may be easier to understand the very first sample without tuples. On the other hand, they don't seem to be too complex...

cesarzc: I think they're not too complex and seem easy to understand. They work the same way that tuples in other languages so it doesn't seem like a big deal IMO.

samples/algorithms/Entanglement.qs
```diff
@@ -0,0 +1,32 @@
+/// # Sample
+/// Entanglement
```
DmitryVasilevsky, line 2: I find these samples so nice and simple - I think we should include them with the other samples in our playground (in samples.mjs).
Superposition should probably be included before the BellState, and RandomBit before the Random Number Generator. In that case BellState may have tuples.
Entanglement, however, seems to be exactly the same as the BellState. Do we need it at all?

DmitryVasilevsky (approved):
Minor comments.

samples/algorithms/Entanglement.qs
cesarzc: I'll add them to the playground in the order that you suggested :)

They are the same. I could expand the Bell state sample to cover all four bell states.

samples/algorithms/BellState.qs
DmitryVasilevsky: I don't mind using tuples here especially if you add Superposition sample as the first sample.

samples/algorithms/Entanglement.qs
DmitryVasilevsky: Yes, this is interesting. We may add Entanglement after Superposition, but before the BellState. In this case we can indeed show all four Bell states and we don't need as many comments. Great idea!
The only question is how to do this so that the histogram shows the results nicely. Maybe let user uncomment one of the four preparation routines?

DmitryVasilevsky: We may have something like this:
```qsharp
    @EntryPoint()
    operation Main() : (Result, Result) {
        // Allocate the two qubits that will be used to create a Bell state.
        use q1 = Qubit();
        use q2 = Qubit();

        // Uncomment any or both of the X gates to obtain different Bell states.
        // X(q1);
        // X(q2);
    
        BellState(q1, q2)
    }
```

cesarzc: I ended up returning measurement results for all Bell states and doing a per-state preparation operation.

PR #573, opened 2023-08-13
Better error for symbols skipped due to target

swernli:
This change tracks which symbols are dropped due to conditional compilation and provides a more specific error message to help the user understand why they are not available.


swernli:
Here is an example of what the updated error looks like:
MicrosoftTeams-image
The first error is the normal "not found" error, and the second is the new "not available" error that calls attention to the fact that the name matches a known item that was conditionally compiled out due to the selected target.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -207,7 +220,14 @@ impl Resolver {
         let namespace = &path.namespace;
         match resolve(kind, &self.globals, &self.scopes, name, namespace) {
             Ok(id) => self.names.insert(path.id, id),
-            Err(err) => self.errors.push(err),
+            Err(err) => match err {
+                Error::NotFound(name, span)
+                    if self.dropped_names.iter().any(|n| n.as_ref() == name) =>
```
minestarks, line 225: Just checking - do the `dropped_names` contain namespaces or not? My expectation is that we have to match the namespace + name exactly before we report the target error, but I can't tell whether we do that here.

minestarks (approved):

compiler/qsc_frontend/src/resolve.rs
swernli: @idavis and I were discussing exactly this question just as you wrote the comment. It does NOT include namespace, and factoring that in is actually rather hard for interesting reasons. We should sync up on the trade-offs to see what you think, but the approach I took for now was to update the error to be the same as `NotFound` but with additional help text, which mitigates corner cases (see https://github.com/microsoft/qsharp/pull/573/commits/b54a6094434b35ef259fd27b055b39862b14180c). Curious to know your thoughts!

minestarks: Could we go the extra mile and at least print out the fully qualified name of _what_ we found in the error message? Even if it's not a "guaranteed" match?

I'm OK with not doing this if it's too complicated.

swernli: hmm... I hadn't thought of that... I think I can make that work!

swernli:
Here is an updated version of what the error message looks like now:
image
Both versions are now reported as `NotFound` but the second shows additional help text with the fully qualified name of the item that is conditionally compiled out.

compiler/qsc_frontend/src/resolve.rs
swernli: Ok, I was able to get the fully qualified name into the error help text (see https://github.com/microsoft/qsharp/pull/573#issuecomment-1698457736). I think this is a good compromise that gives the user more information!

PR #581, opened 2023-08-14
Add optional jupyterlab dependency to qsharp

minestarks:
- Remove the existing "test" dependency (we don't use it anywhere - and I don't recall why I initially added it)
- Add optional dependency to `qsharp_jupyterlab`. The way to install it would be `pip install qsharp[jupyterlab]`.


billti:
Looks good to me, but I'm not that familiar with those files and how dependencies work, so I'll let Ian give the thumbs up. Thanks!

idavis (approved):

PR #583, opened 2023-08-15
Test that Q# examples in katas build and run successfully

cesarzc:
This change adds validation for Q# examples that are part of katas.


billti (approved):

npm/test/basics.js
```diff
@@ -241,42 +282,42 @@ test("all katas work", async () => {
 
 test("qubit kata is valid", async () => {
```
DmitryVasilevsky, line 283: We have individual tests for Katas? Can we get the list from Content/Kata.json

PR #584, opened 2023-08-15
Unify file access in debugger

idavis:
This is a workaround for windows which doesn't play well with URIs. The DAP `setDebuggerPathFormat`/`convertClientPathToDebugger`/`convertDebuggerPathToClient` can't be used as there is a [bug requiring nodejs](https://github.com/microsoft/vscode-debugadapter-node/issues/298) which prevents usage in web extensions. Once the bug is fixed we can update the debugger to use URIs without this PR's retry conversion.

This PR also fixes an issue where if the user closed the file being debugged or it was open as 'temporarily open' then we couldn't step back into the code.

### DIFF
--- a/vscode/src/debugger/activate.ts
+++ b/vscode/src/debugger/activate.ts
@@ -128,11 +128,14 @@ export const workspaceFileAccessor: FileAccessor = {
   normalizePath(path: string): string {
     return path.replace(/\\/g, "/");
   },
+  convertToWindowsPathSeparator(path: string): string {
+    return path.replace(/\//g, "\\");
+  },
   resolvePathToUri(path: string): vscode.Uri {
     const normalizedPath = this.normalizePath(path);
     return vscode.Uri.parse(normalizedPath, false);
   },
-  async openFile(path: string): Promise<vscode.TextDocument> {
+  async openPath(path: string): Promise<vscode.TextDocument> {
     const uri: vscode.Uri = this.resolvePathToUri(path);
     return this.openUri(uri);
   },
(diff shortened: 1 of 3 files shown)


vscode/src/debugger/activate.ts
```diff
@@ -125,15 +125,27 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
...
+  },
+  async openPath(path: string): Promise<vscode.TextDocument> {
+    const uri: vscode.Uri = this.resolvePathToUri(path);
+    return this.openUri(uri);
+  },
+  async openUri(uri: vscode.Uri): Promise<vscode.TextDocument> {
+    try {
+      return await vscode.workspace.openTextDocument(uri);
+    } catch {
+      const path = this.convertToWindowsPathSeparator(uri.toString());
```
billti, line 146: It's weird this is needed, as Windows typically handles forward slashes in paths fine as well. What was failing without the conversion?

vscode/src/debugger/session.ts
```diff
@@ -62,7 +62,7 @@ export class QscDebugSession extends LoggingDebugSession {
   ) {
     super();
 
-    this.program = vscode.Uri.parse(this.config.program).path;
+    this.program = fileAccessor.resolvePathToUri(this.config.program);
```
billti, line 65: I had to explicitly change this from a Uri to a string to fix github.dev. Are you sure this won't break opening a repo on GitHub again?

idavis: It all works as far as I can test on mac, local web, vscode.dev side loading, wsl, and windows.

vscode/src/debugger/activate.ts
idavis:
```
[trace] breakpointLocationsRequest: %O {"source":{"name":"BernsteinVazirani.qs","path":"c:\\src\\qsharp\\samples\\algorithms\\BernsteinVazirani.qs"},"line":34}

[error] Failed to open file: CodeExpectedError: cannot open c:/src/qsharp/samples/algorithms/BernsteinVazirani.qs. Detail: Unable to resolve resource c:/src/qsharp/samples/algorithms/BernsteinVazirani.qs
```

billti (approved):

PR #586, opened 2023-08-15
Add three quantum samples

sezna:
I am not particularly confident in the correctness of my quantum explanations, so please thoroughly inspect them. Particularly the superdense one.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,9 +1,9 @@
-* @billti @idavis @minestarks @samarsha @swernli
+* @billti @idavis @minestarks @swernli
 
-/.github/ISSUE_TEMPLATE/fuzz_bug_report.md @billti @kuzminrobin @swernli
-/.github/workflows/fuzz.yml @billti @kuzminrobin @swernli
-/compiler @idavis @samarsha @swernli
-/fuzz  @billti @kuzminrobin @swernli
+/.github/ISSUE_TEMPLATE/fuzz_bug_report.md @billti @idavis @swernli
+/.github/workflows/fuzz.yml @billti @idavis @swernli
+/compiler @idavis @minestarks @swernli
+/fuzz  @billti @idavis @swernli
 /katas @billti @cesarzc @DmitryVasilevsky @sezna @swernli
 /jupyterlab @billti @idavis @minestarks
 /language_service @billti @idavis @minestarks @ScottCarda-MS
@@ -15,7 +15,7 @@
 /vscode @billti @idavis @minestarks
 /wasm @billti @cesarzc @swernli
 
-/Cargo.lock @idavis @samarsha @swernli
-/Cargo.toml @idavis @samarsha @swernli
+/Cargo.lock @idavis @minestarks @swernli
+/Cargo.toml @idavis @minestarks @swernli
 /package-lock.json @billti @cesarzc @minestarks
 /package.json @billti @cesarzc @minestarks
--- a/compiler/qsc/src/interpret/debug/tests.rs
+++ b/compiler/qsc/src/interpret/debug/tests.rs
@@ -137,7 +137,7 @@ fn stack_traces_can_cross_file_and_entry_boundaries() {
             ("1.qs".into(), source1.into()),
             ("2.qs".into(), source2.into()),
         ],
-        Some("Adjoint Test2.A(0);".into()),
+        Some("Adjoint Test2.A(0)".into()),
     );
     let interpreter =
         stateless::Interpreter::new(true, source_map).expect("Failed to compile base environment.");
--- a/compiler/qsc/src/interpret/stateful.rs
+++ b/compiler/qsc/src/interpret/stateful.rs
@@ -14,6 +14,7 @@ use crate::{
 use miette::Diagnostic;
 use num_bigint::BigUint;
 use num_complex::Complex;
+use qsc_codegen::qir_base::generate_qir_for_stmt;
 use qsc_data_structures::index_map::IndexMap;
 use qsc_eval::backend::Backend;
 use qsc_eval::{
@@ -92,14 +93,19 @@ impl LineError {
 }
 
 #[derive(Clone, Debug, Diagnostic, Error)]
-#[diagnostic(transparent)]
 pub enum LineErrorKind {
     #[error(transparent)]
+    #[diagnostic(transparent)]
     Compile(#[from] incremental::Error),
     #[error(transparent)]
+    #[diagnostic(transparent)]
     Pass(#[from] qsc_passes::Error),
     #[error("runtime error")]
+    #[diagnostic(transparent)]
     Eval(#[from] qsc_eval::Error),
+    #[error("code generation target mismatch")]
+    #[diagnostic(code("Qsc.Interpret.TargetMismatch"))]
+    TargetMismatch,
 }
 
 struct Lookup<'a> {
@@ -160,6 +166,7 @@ pub struct Interpreter {
     fir_store: IndexMap<PackageId, qsc_fir::fir::Package>,
     state: State,
     source_package: PackageId,
+    target: TargetProfile,
 }
 
 impl Interpreter {
@@ -221,6 +228,7 @@ impl Interpreter {
             state: State::new(map_hir_package_to_fir(package)),
             lowerer,
             fir_store,
+            target,
         })
     }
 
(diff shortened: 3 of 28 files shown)


samples/algorithms/JointMeasurement.qs
```diff
@@ -0,0 +1,21 @@
...
+/// The basis array and qubit array must be of the same length.
+namespace Sample {
+    open Microsoft.Quantum.Measurement;
+    @EntryPoint()
+    operation Main () : Result {
+        // `Measure` takes two arguments: first, an array `bases` of Pauli values, and
+        // an array `qubits` of qubits to measure.
+
+        // The below code shows joint measurement of a register of qubits.
+        use qs = Qubit[2];
```
cesarzc, line 16:
```suggestion
        // In Q#, the `Measure` operation performs a joint measurement of one or
        // more qubits in the specified Pauli bases. It takes two arguments: an
        // array of single-qubit Pauli values indicating the tensor product
        // factors on each qubit, and a register of qubits to be measured.

        // The below code uses a joint measurement as a way to check the parity
        // of two qubits.
        use qs = Qubit[2];
        H(qs[0]);
        CNOT(qs[0], qs[1]);
```

samples/algorithms/JointMeasurement.qs
```diff
@@ -0,0 +1,21 @@
+/// # Sample
+/// Joint Measurement
+///
+/// # Description
+/// The `Measure` operation performs a Pauli measurement. It performs a
+/// joint measurement of one or more qubits in the specified Pauli bases.
+/// The basis array and qubit array must be of the same length.
```
cesarzc, line 7:
```suggestion
/// Joint measurements, also known as Pauli measurements, are a generalization
/// of 2-outcome measurements to multiple qubits and other bases.
```

samples/algorithms/JointMeasurement.qs
```diff
@@ -0,0 +1,21 @@
...
+namespace Sample {
+    open Microsoft.Quantum.Measurement;
```
cesarzc, line 9:
```suggestion
```

cesarzc: This is not needed since `Measure` is in the `Microsoft.Quantum.Intrinsic` namespace, which is automatically opened by default.

samples/algorithms/Measurement.qs
```diff
@@ -0,0 +1,35 @@
...
+
+        use q = Qubit();
+        let result: Result = Measure([PauliZ], [q]);
+
+        // The below code shows joint measurement of a register of qubits.
+        use qs = Qubit[2];
+        let result = Measure([PauliZ, PauliZ], qs);
+
+        return result;
+    }
```
cesarzc, line 34:
```suggestion
    operation Main () : (Result, Result[], Result) {
        // The `M` operation performs a measurement of a single qubit in the
        // computational basis, also known as the Pauli Z basis.
        use q = Qubit();
        let result = M(q);
        Reset(q);

        // The `MeasureEachZ` operation measures each qubit in an array in the
        // computational basis and returns an array of `Result` values.
        use qs = Qubit[2];
        let results = MeasureEachZ(qs);

        // The `Measure` operation performs a joint measurement of one or more
        // qubits in the specified Pauli bases.
        H(qs[0]);
        CNOT(qs[0], qs[1]);
        let jointResult = Measure([PauliZ, PauliZ], qs);
        ResetAll(qs);

        return (result, results, jointResult);
    }
```

samples/algorithms/Measurement.qs
```diff
@@ -0,0 +1,35 @@
+/// # Sample
+/// Measurement
+///
+/// # Description
+/// Measurement is a critical component of quantum algorithm design. Measurement refers to
+/// measurements in the computational basis, unless otherwise specified. The result of
+/// a measurement is a value of the type `Result`, that is, `One` or `Zero`. This appears
+/// similar to classical computation, but the measurement itself is a probabilistic process.
+/// The process of measuring a qubit can change the state of the qubit.
```
cesarzc, line 9:
```suggestion
/// Quantum measurement is an irreversible operation in which a quantum system
/// is manipulated to yield a numerical result. Measuring a quantum system
/// generally changes the quantum state that describes that system.
///
/// In Q#, the result of a measurement is a value of the type `Result`, that is,
/// `One` or `Zero`.
///
/// This Q# program exemplifies different types of measurements.
```

samples/algorithms/JointMeasurement.qs
```diff
@@ -0,0 +1,24 @@
...
+        // more qubits in the specified Pauli bases. It takes two arguments: an
+        // array of single-qubit Pauli values indicating the tensor product
+        // factors on each qubit, and a register of qubits to be measured.
+
+        // The below code uses a joint measurement as a way to check the parity
+        // of two qubits.
+        use qs = Qubit[2];
+        H(qs[0]);
+        CNOT(qs[0], qs[1]);
+        let result = Measure([PauliZ, PauliZ], qs);
```
ScottCarda-MS, line 20: Might be useful to mention that Joint Measurements can be used to get information about a collection of qubits without necessarily collapsing the state of any one qubit.

ScottCarda-MS: For example here, the result will always be even parity (I think that is `Zero`) and the state of the qubits are unchanged, as we would have already known the parity must be zero from the previous code. If you want to emphasize how the joint measurement can change the state of the system of qubits, you could show the joint measurement doing the entangling instead of the `CNOT`, performing an `X(qs[1])` if the result is odd parity (`One`), though that example uses a measurement result in a condition.

samples/algorithms/Measurement.qs
```diff
@@ -0,0 +1,37 @@
...
+        use q = Qubit();
+        let result = M(q);
+        Reset(q);
+
+        // The `MeasureEachZ` operation measures each qubit in an array in the
+        // computational basis and returns an array of `Result` values.
+        use qs = Qubit[2];
+        let results = MeasureEachZ(qs);
+
+        // The `Measure` operation performs a joint measurement of one or more
```
ScottCarda-MS, line 28: This seems like it should be reserved for the Joint Measurement sample. Maybe just show how `Measure` can be used to measure a single qubit in a different base that `PauliZ`, and maybe put a reference to the Joint Measurement sample as a "see more on Joint Measurement".

cesarzc: I think probably removing `Measure` from here is a good since each sample should be self-contained and we should not reference other samples.

samples/algorithms/JointMeasurement.qs
cesarzc: I think using the joint measurement to perform an entanglement might be too advanced for a simple sample and is a topic that belongs in the Katas.

However, the example of getting information about a collection of qubits makes sense. We could modify the example to be something like this:
```qsharp
    operation Main() : (Result, Result[]) {
        // Prepare an entangled state.
        use qs = Qubit[2];  // |00〉
        H(qs[0]);           // 1/sqrt(2)(|00〉 + |10〉)
        CNOT(qs[0], qs[1]); // 1/sqrt(2)(|00〉 + |11〉)

        // Show the quantum state before performing the joint measurement.
        DumpMachine();

        // The below code uses a joint measurement as a way to check the parity
        // of the first two qubits. In this case, the parity measurement result
        // will always be `Zero`.
        // Notice how the state was not collapsed by the joint measurement.
        let parityResult = Measure([PauliZ, PauliZ], qs[...1]); 
        DumpMachine();

        // However, if we perform a measurement just on the first qubit, we can
        // see how the state collapses.
        let firstQubitResult = M(qs[0]);
        DumpMachine();

        // Measuring the last qubit does not change the state quantum state
        // since the state of the second qubit collapsed when the first qubit 
        // was measured because they were entangled.
        let secondQubitResult = M(qs[1]);
        DumpMachine();

        ResetAll(qs);
        return (parityResult, [firstQubitResult, secondQubitResult]);
    }
```

samples/algorithms/JointMeasurement.qs
```diff
@@ -0,0 +1,40 @@
...
+        // Notice how the state was not collapsed by the joint measurement.
+        let parityResult = Measure([PauliZ, PauliZ], qs[...1]); 
+        DumpMachine();
+
+        // However, if we perform a measurement just on the first qubit, we can
+        // see how the state collapses.
+        let firstQubitResult = M(qs[0]);
+        DumpMachine();
+
+        // Measuring the last qubit does not change the state quantum state
```
ScottCarda-MS, line 31: I think there is an extra "state" word here.

ScottCarda-MS (approved):

cesarzc (approved):

PR #587, opened 2023-08-15
Support for shots in Python

minestarks:
Adds a `run()` method to the Python library that takes a `shots=` parameter.


pip/samples/sample.ipynb
minestarks, line 1: Review note: You can select "view file" to view the formatted notebook on GitHub

compiler/qsc_ast/src/assigner.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
     mut_visit::{self, MutVisitor},
 };
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
swernli, line 12: Does this still need to be clone-able? I don't see anywhere else in the changes that adds cloning for it, but I might be missing something. Cloning the assigner can be risky; it's safe as long as all AST handled by the cloned assigner is thrown away, but if it co-exists with AST handled by the original assigner it can open up the possibility of node id collisions.

minestarks: Hmm, this seems to be an accident, thanks for the catch!

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -358,14 +357,38 @@ impl Interpreter {
     ) -> Result<Value, Vec<LineError>> {
         let mut result = Value::unit();
 
+        let stmt_ids = self.compile_to_stmts(line)?;
```
minestarks, line 360: This refactoring changes behavior in a way that we may not intend. I'm going to revert most of this and keep my changes contained to the new method I added.

minestarks:
@swernli @billti this is now ready for review

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -439,6 +437,122 @@ impl Interpreter {
...
+        for _i in 0..shots {
+            results.push(
+                match eval_stmt(
+                    stmt_id,
+                    &globals,
+                    &mut Env::with_empty_scope(),
+                    &mut SparseSim::new(),
+                    self.package,
+                    receiver,
+                ) {
```
swernli, line 535: I'm still torn on the environment part of it... I think what you have here is probably the right place to start, but putting an alternative here just for future consideration. If we want to allow the user to provide expressions that make use of variables they have defined in the interactive environment, we could add `Clone` derives to `Scope` and `Env` structs, and then this code would work:

```suggestion
            let mut env = self.env.clone();
            results.push(
                match eval_stmt(
                    stmt_id,
                    &globals,
                    &mut env,
                    &mut SparseSim::new(),
                    self.package,
                    receiver,
                ) {
```
However, then we open up the door to reuse of globally allocated qubits, which would cause problems with the simulator (today it would panic if it recieves a qubit ID that that it has not allocated). To really make this work without the corner case crash, we would have to have an additional check of the input expression to confirm it doesn't include any free variables of qubit type.

I don't think this work would fit into the current PR, but it is something we can consider for later. Note that `qirgen` below reuses the environment rather than getting a fresh one, so it already has this corner case issue.

pip/qsharp/_qsharp.py
```diff
@@ -59,6 +59,25 @@ def eval_file(path) -> None:
...
+
+    :returns values: A list of results or runtime errors.
+
+    :raises QSharpError: If there is an error interpreting the input.
+    """
+
+    def callback(output):
+        print(output)
+
+    return _interpreter.run(entry_expr, shots, callback)
```
swernli, line 78: This should use `get_interpreter()` to ensure it will initialize the value if it hasn't been yet:

```suggestion
    return get_interpreter().run(entry_expr, shots, callback)
```

compiler/qsc/src/interpret/stateful.rs
minestarks: Yeah, I don't think there's a quick ideal solution here, so let's reconsider after this PR. In my ideal world, I think everything - including the compilation state! - would be clone-able, and both `qirgen` and `run` would operate on a cloned state. No interference at all. But that's a big change.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -439,6 +437,122 @@ impl Interpreter {
...
+            }
+        }
+
+        let stmt_id = stmt_ids
+            .pop()
+            .expect("expression should yield exactly one statement");
+        assert!(
+            stmt_ids.is_empty(),
+            "expression should yield exactly one statement"
+        );
```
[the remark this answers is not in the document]
minestarks: Thanks! I was tearing my hair out trying to make this cleaner.

swernli (approved):

minestarks:
@billti / @idavis  do you mind giving a signoff for the Python changes please? Thank you!

billti (approved):

PR #588, opened 2023-08-15
Python `compile` for QIR codegen

swernli:
This adds a `compile` method to the Python `qsharp` module and introduces an `init` pattern to be used when starting up a session. The expected usage is that for full compilation and simulation, `qsharp.init()` or `qsharp.init(target=qsharp.Target.Full)` are called at the outset to configure compilation (NOTE: latest allows the user to skip this and just takes the default on first use). If `qsharp.init(target=qsharp.Target.Base)` is used instead then local simulation will follow the same restrictions as base profile code generation. When in this mode, `qsharp.compile(<expr>)` will produce output that can be used as part of submitting to QIR capable targets.

### DIFF
--- a/compiler/qsc/src/interpret/stateful.rs
+++ b/compiler/qsc/src/interpret/stateful.rs
@@ -14,7 +14,7 @@ use crate::{
 use miette::Diagnostic;
 use num_bigint::BigUint;
 use num_complex::Complex;
-use qsc_codegen::qir_base::BaseProfGen;
+use qsc_codegen::qir_base::generate_qir_for_stmt;
 use qsc_data_structures::index_map::IndexMap;
 use qsc_eval::backend::Backend;
 use qsc_eval::{
@@ -438,8 +438,6 @@ impl Interpreter {
             ))]);
         }
 
-        let mut codegen = BaseProfGen::new();
-
         let mut fragment = self.compiler.compile_expr(expr).map_err(|errors| {
             let source = expr.into();
             errors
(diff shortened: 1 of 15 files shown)


compiler/qsc/src/interpret/stateful.rs
```diff
@@ -414,6 +422,76 @@ impl Interpreter {
...
+        if !pass_errors.is_empty() {
+            let source = expr.into();
+            return Err(pass_errors
+                .into_iter()
+                .map(|error| LineError(WithSource::new(Arc::clone(&source), error.into(), None)))
+                .collect());
+        }
+
+        let Fragment::Stmt(stmt) = fragment else {
+            panic!("only a stmt fragment should reach here.")
```
minestarks, line 465: Can we make `compile_expr` return the `Stmt` instead? (We can panic in `compile_expr` function if the `Fragment` isn't a statement, but the intent will be much clearer since we created the fragment in that same function)

minestarks (approved):

pip/src/interpreter.rs
```diff
@@ -28,24 +29,39 @@
...
+/// A Q# measurement result.
+pub(crate) enum Target {
+    Full,
+    Base,
+}
+
 #[pyclass(unsendable)]
 pub(crate) struct Interpreter {
     pub(crate) interpreter: stateful::Interpreter,
+    // TODO: Smoke and mirrors.
```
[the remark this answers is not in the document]
swernli: Unused code removed.

compiler/qsc/src/interpret/stateful.rs
swernli: Yeah, I had the same thought at one point, but the desire to reuse the pass code without having to introduce a different, codegen-specific variant led me to keep it as-is.

pip/qsharp/_qsharp.py
```diff
@@ -1,10 +1,33 @@
...
+
+def get_interpreter() -> Interpreter:
+    """
+    Returns the Q# interpreter.
+
+    :returns: The Q# interpreter.
+    """
+    global _interpreter
+    if _interpreter is None:
+        init()
```
swernli, line 29: @minestarks I changed this to auto-init, so it's not required for folks who just want the default, non-hardware compatible settings. If we want to go further and force init to be called only once, we could have `init` raise an error if `_interpreter` is not `None`, which would force the kernel restart.

idavis (approved):

PR #599, opened 2023-08-17
Adding ability to create sdist for pip install.

idavis:
```
cd pip
python -m maturin sdist
pip install --force-reinstall ../target/wheels/qsharp-0.0.0.tar.gz
```

### DIFF
--- a/library/src/lib.rs
+++ b/library/src/lib.rs
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub const CORE_LIB: &[(&str, &str)] = &[
+    ("core.qs", include_str!("../core/core.qs")),
+    ("qir.qs", include_str!("../core/qir.qs")),
+];
+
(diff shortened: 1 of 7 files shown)


library/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub const CORE_LIB: &[(&str, &str)] = &[
```
minestarks, line 4: This seems like an opportunity to simplify the lookup in [`wasm/src/lib.rs`](https://github.com/microsoft/qsharp/blob/1aae38c481d14a50567ca8e73b4e8cea8318f9df/wasm/src/lib.rs#L40) . Would it be worth the effort to refactor?

idavis: I think we could do this as another PR, but I think it is beyond what we're trying to accomplish here.

idavis:
Closing for now. Can revisit if we like. Would still likely benefit from pulling `stdlib` into a crate in a separate PR.

PR #601, opened 2023-08-17
Doc-comments confuse `%%qsharp` cells

### LINKED ISSUE #593: Doc-comments confuse `%%qsharp` cells
If an item in a Q# magic cells includes a doc-comment and is not part of a namespace, it causes the fragment parser to expect namespace style Q#:

image

We don't really need doc-comment support on top-level items, but not allowing them can make the exercise of copying code out of a documented Q# file and into a notebook very annoying (see for example trying to copy content from our samples into a cell).

Of note, doc-comments work fine if they appear inside of a namespace:

image

swernli:
Fixes #593

### DIFF
--- a/compiler/qsc_parse/src/item.rs
+++ b/compiler/qsc_parse/src/item.rs
@@ -21,8 +21,8 @@ use crate::{
 };
 use qsc_ast::ast::{
     Attr, Block, CallableBody, CallableDecl, CallableKind, Ident, Item, ItemKind, Namespace,
-    NodeId, Pat, PatKind, Path, Spec, SpecBody, SpecDecl, SpecGen, Stmt, Ty, TyDef, TyDefKind,
-    TyKind, Visibility, VisibilityKind,
+    NodeId, Pat, PatKind, Path, Spec, SpecBody, SpecDecl, SpecGen, Stmt, StmtKind, Ty, TyDef,
+    TyDefKind, TyKind, Visibility, VisibilityKind,
 };
 use qsc_data_structures::span::Span;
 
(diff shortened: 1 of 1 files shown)


compiler/qsc_parse/src/item.rs
```diff
@@ -98,10 +98,22 @@ pub(super) fn parse_fragments(s: &mut Scanner) -> Result<Vec<Fragment>> {
...
+    let doc = parse_doc(s);
+    if let Some(mut namespace) = opt(s, parse_namespace)? {
+        namespace.doc = doc.into();
         Ok(Fragment::Namespace(namespace))
     } else {
-        stmt::parse(s).map(Fragment::Stmt)
+        let kind = s.peek().kind;
+        let span = s.peek().span;
+        let mut fragment = stmt::parse(s).map(Fragment::Stmt)?;
+        if let Fragment::Stmt(stmt) = &mut fragment {
```
samarsha, line 109: This is always true.

swernli: Good point. I thought about using a let else pattern but then I realized I could just skip the mapping on the line above. I'll fix that.

compiler/qsc_parse/src/item.rs
```diff
@@ -98,10 +98,20 @@ pub(super) fn parse_fragments(s: &mut Scanner) -> Result<Vec<Fragment>> {
 }
 
 fn parse_fragment(s: &mut Scanner) -> Result<Fragment> {
-    if let Some(namespace) = opt(s, parse_namespace)? {
+    let doc = parse_doc(s);
```
samarsha, line 101: Maybe a short comment explaining why this is necessary and that we hope it can be cleaned up in the future? (since the fact that this will hopefully be obsoleted by modules-as-items made us more OK with a hacky solution for now)

samarsha (approved):

compiler/qsc_parse/src/item.rs
swernli: Added.

PR #602, opened 2023-08-18
Display errors on failure in katas

DmitryVasilevsky:
Displays an error under the editor in case kata exercise fails to execute. For example, if `fail` keyword is encountered in user code or a library code.
Thanks @billti for helping with this.
Thanks @minestarks for providing cleaner solution!


playground/src/editor.tsx
```diff
@@ -127,13 +127,37 @@ export function Editor(props: {
 
     try {
       if (props.kataExercise) {
+        // Clear any shot errors before running verification
```
minestarks, line 130: Hmm, I'm not convinced this is the right fix. We already have a mechanism for the regular editor to update the UI, as long as events are coming through the event target. The kata editors should be wired up exactly the same way. I'm not sure what is going wrong (looking at the logs, I see the error events coming through the WASM just fine, they're just not updating the UI).

But I feel like this TODO comment may be the more likely suspect? https://github.com/microsoft/qsharp/blob/64538ed62c13643842a7b84ac278d27f7a1b1cd3/playground/src/main.tsx#L107

minestarks: I hope you don't mind I poked around a bit, the TODO comment was too suspicious :) It looks like error handling just wasn't wired up for the `Kata` component. This is the alternate fix I had in mind: https://github.com/microsoft/qsharp/commit/516a58b6e51b3707e3bbdd1f42d177ca8df553ae

DmitryVasilevsky: This is an interesting suggestion. Yes, when I started to work on this, I was trying to figure out how to use the events, but I wasn't able to figure it out. What we have here is a program failure during execution, not messages from the program. As far as I understand messages are communicated via events while checkExcerciseSolution is still in progress. The situation in question happens after the program ends. So, Bill suggested this approach that we look at the result and add message accordingly as we are already in the typescript on the page and not in WASM.
This is different from the normal editor in the playground because that editor may have multiple shots. If only a few shots fail, the errors are shown when user navigates to the corresponding shot. In other words, they are stored in a structure and are displayed only as a result of user selecting appropriate shot. This selection of shot is not present in Katas as it is not needed. That's why the handling is different.
My understanding of this code is minimal - we can chat more if you feel there is a better solution.

DmitryVasilevsky: Your fix seems a lot cleaner!

minestarks: As discussed. Pushed my changes!

billti (approved):

PR #609, opened 2023-08-21
Add devcontainer

swernli:
This change adds a devcontainer.json file (ignored by Prettier) with set up for VSCode and GitHub Codespaces compatible container. It includes all the required components for building the repo with `python build.py` as well as the set of workspace recommended extensions.


billti:
OK.... I guess. Do you or others use DevContainers then? If we're going to actively use this in the team then it seems fine to check in. But I don't want it to bit-rot if it's not something we actively need/use.

billti (approved):

swernli:
I use devcontainers as my primary way of testing Linux from Windows and for running in codespaces from my iPad, and I know other repos have benefitted from having this configured. It’s also handy that all our dependencies can be configured from official Microsoft devcontainer images (and one simple post-create command), so there is no need for complex configuration or even a Dockerfile.

PR #615, opened 2023-08-23
Measurement Kata: Newlines in Equation Blocks

ScottCarda-MS:


katas/content/single_qubit_measurements/a_b_basis_measurements/solution.md
```diff
@@ -1,8 +1,8 @@
 For this problem, we follow the procedure given in the "Measurements in Arbitrary Orthogonal Bases" section of the Kata.
 As noted in the solution of the "Distinguishing Orthogonal States: 2" exercise, the gate $R_x(-\theta)$, with $\theta = 2\alpha$ transforms the $\ket A/\ket B$ states to the $\ket 0/\ket 1$ states:
 \begin{align}
-R_x(-\theta) \ket A &= \ket 0 \\
-R_X(-\theta) \ket B &= \ket 1
+R_x(-\theta) \ket A &= \ket 0, \\\\
+R_x(-\theta) \ket B &= \ket 1.
 \end{align}
```
cesarzc, line 6: The way that I've been fixing these issues is by not using `\begin{align}` and having each one of the equations be its own LaTex "line", e.g.:
```
$$R_x(-\theta) \ket A = \ket 0$$
$$R_x(-\theta) \ket B = \ket 1$$
```

Since punctuation doesn't really add value in these cases, and requires additional LaTex notation (e.g. `&=` instead of simply `=`), I believe doing separation of lines this way is better.

ScottCarda-MS: As discussed offline, I've change it to use the `$$` syntax instead of `align`, but will keep the punctuation for grammatical purposes.

cesarzc (approved):
Thanks for fixing this @ScottCarda-MS!

PR #619, opened 2023-08-23
Add launch.json and stopOnEntry support.

idavis:


vscode/src/debugger/activate.ts
```diff
@@ -106,8 +107,26 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
...
         config.noDebug = "noDebug" in config ? config.noDebug : false;
+        config.stopOnEntry = !config.noDebug;
+      }
+    } else {
+      // we have a launch config, resolve the program path
+
+      // ensure we have the program uri correctly formatted
+      // this is a user specified path.
+      if (config.program && config.program !== "${file}") {
+        const uri = workspaceFileAccessor.resolvePathToUri(config.program);
```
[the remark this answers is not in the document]
minestarks: Or do we want to implement [`resolveDebugConfigurationWithSubstitutedVariables`](https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider) instead? There are [a number of variables](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables) we may want to support.

vscode/src/debugger/activate.ts
```diff
@@ -106,8 +107,26 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
...
+        config.program = uri.toString();
+      } else {
+        // Use the active editor if no program or ${file} is specified.
+        const editor = vscode.window.activeTextEditor;
+        if (
+          editor &&
+          vscode.languages.match(qsharpDocumentFilter, editor.document)
+        ) {
+          config.program = editor.document.uri.toString();
+        }
```
minestarks, line 129: Do we also want implement `provideDebugConfiguration` so that we can recommend a default configuration to the user? I'm not sure how that would work exactly, perhaps we would need discover .qs files in the workspace? I'm not sure if it's worth it.

minestarks (changes requested):
See comment about variable substitution - LGTM otherwise

vscode/src/debugger/activate.ts
idavis: Not sure, we can look at it in the future.

vscode/src/debugger/activate.ts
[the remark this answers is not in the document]
idavis: Great idea, done.

npm/test/basics.js
```diff
@@ -560,9 +560,10 @@ test("debug service loading source without entry point attr fails - web worker",
         let m1 = M(q1);
         return [m1];
     }
-}`
+}`,
+      undefined
     );
-    assert.equal(false, result);
+    assert.notEqual("", result);
```
billti, line 566: Seems a little fragile. If it returns null or undefined, this will pass then? Maybe verify it is a non-zero length string if that is the expectation.

npm/test/basics.js
```diff
@@ -576,9 +577,38 @@ test("debug service loading source with syntax error fails - web worker", async
...
+test("debug service loading source with bad entry expr fails - web worker", async () => {
+  const debugService = getDebugServiceWorker();
+  try {
+    const result = await debugService.loadSource(
+      "test.qs",
+      `namespace Sample { operation main() : Unit { } }`,
+      "SomeBadExpr()"
     );
-    assert.equal(false, result);
+    assert.notEqual("", result);
```
billti, line 597: Again, seems very prone to false positives. Maybe a more specific check.

billti (approved):
Couple minor comments. Thanks!

minestarks (approved):

PR #621, opened 2023-08-23
VS Code editor codegen and Azure submission

billti:
A work in progress, but with this you can:

 - Run the "Get QIR for current Q# program" to see the LL codegen for the currently active editor - if base profile compliant. (Note: May not leave this in, but it's handy for testing and those who are QIR-curious).
 - Go into settings and enable the Azure Integration (need to opt in currently due to experimental nature and use of temp service for processing)
 - Sign-in to Azure via you AAD or MSA account and see your workspaces (ignore the 'Connection String' option for now which is mock data for quick validation)
 - Click on the submit icon (looks like a Play button) to the right of a hardware target to submit the currently active editor code.
 - Refresh the workspace view to see latest job status.
 - Click on the download icon to see the raw histogram data for completed jobs.

It's spaghetti code currently from much experimentation, but sharing for early feedback on design and for a peek into service integration.

Still TODO for this PR:

- [x] Wire up the QIR Profile selector to switch the language service between base and full profiles and show corresponding errors.
- [x] Disable job submission if active program has errors
- [x] Improve display/filtering of providers and only enable submission on QIR capable ones.
- [x] Better code structure, error handling, and diagnostics logging.

TODO in a follow-up PR:

- Auto-refresh the workspace tree periodically if any jobs are in the waiting or executing state, or newly submitted.
- Remember and automatically reload workspaces in the tree view across editor sessions. (And ability to remove)
- Ability to filter, sort, and manage (cancel, delete, etc.) jobs.
- Ability to fetch the raw result data, display as histogram, etc.


wasm/src/lib.rs
```diff
@@ -37,6 +38,34 @@ pub fn git_hash() -> String {
...
+    let sources = SourceMap::new([("test".into(), code.into())], None);
+
+    let unit = qsc::compile::compile(
+        &store,
+        &[std],
+        sources,
+        PackageType::Exe,
+        TargetProfile::Base,
+    );
+    let package = store.insert(unit.0);
```
swernli, line 55: This should check for errors before inserting into the store. Putting a CompileUnit into the store when it has errors is unsafe. I vaguely recall there being a panic in the PackageStore insert method that would have prevented this but it might have been removed for some reason (perhaps for the language server?).

vscode/src/qirGeneration.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (diagnostics?.length > 0) {
+    throw new QirGenerationError(
+      "The current program contains errors that must be fixed before submitting to Azure"
+    );
+  }
+
+  const code = editor.document.getText();
+
+  // Create a temporary worker just to get the QIR, as it may loop/panic during codegen.
+  // Let it run for max 10 seconds, then terminate it if not complete.
```
swernli, line 50: 10 seconds seems like a rather aggresive timeout. Eventually I think we want a status popup that says we are compiling and allows cancellation. In the meantime I think something like a 30 second timeout might be more appropriate.

billti: You think? I actually had it at 5 seconds right until I pushed and decided to be more 'lenient'. With the speed we can run even complex shots at, what do you think might legitimately take up to 30 seconds to produce base profile code?

One we have telemetry I'll add a metric on compilation time (and timeouts). It would be an interesting value to observe.

vscode/src/azure/REST API tests.http
minestarks, line 1: How is this used exactly? If it's a test file should it be under `test` and/or excluded from publishing?

vscode/src/azure/commands.ts
```diff
@@ -0,0 +1,141 @@
...
+  queryWorkspace,
+  queryWorkspaces,
+  submitJob,
+} from "./workspaceActions";
+import { QuantumUris, compileToBitcode } from "./networkRequests";
+import { getQirForActiveWindow } from "../qirGeneration";
+
+// TODO: Proper docs on the qsharp wiki
+const corsDocsUri =
+  "https://gist.github.com/billti/09637269db4bae86c0e3a552dd20eb9b";
```
minestarks, line 25: Shall I make an `aka.ms` real quick so we can redirect this after we ship if need be?

vscode/package.json
```diff
@@ -59,6 +121,41 @@
         "category": "Debug",
         "enablement": "!inDebugMode",
         "icon": "$(play)"
+      },
+      {
+        "command": "qsharp-vscode.workspacesRefresh",
```
minestarks, line 126: Should these commands be enabled conditional on `AzureQuantumIntegration` as well?

vscode/src/azure/commands.ts
```diff
@@ -0,0 +1,141 @@
...
+          vscode.window.showErrorMessage(e.message);
+          return;
+        }
+      }
+      if (!qir) return;
+
+      // Note: Below compilation to be removed when all regions support .ll text directly
+      const compilerService: string | undefined = vscode.workspace
+        .getConfiguration("Q#")
+        .get("compilerService"); // e.g. in settings.json: "Q#.compilerService": "https://qsx-proxy.azurewebsites.net/api/compile"
```
minestarks, line 53: Was there supposed to be a separate consent setting for this service or is it covered by the `AzureQuantumIntegration` setting?

vscode/src/azure/networkRequests.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+    log.debug(`Got response ${response.status} ${response.statusText}`);
+    return response;
+  } catch (e) {
+    log.error(`Failed to fetch ${uri}: ${e}`);
+    throw e;
+  }
+}
+
+export class AzureUris {
```
minestarks, line 80: Wouldn't it be easier to use our JavaScript client library for this? https://learn.microsoft.com/en-us/javascript/api/overview/azure/quantum-jobs-readme?view=azure-node-preview

I guess I wouldn't be surprised to learn that one doesn't work in the browser, or something...

minestarks (approved):

vscode/src/azure/REST API tests.http
billti: Yeah, I debated if it should be checked in or not, but being that I put hours or working into it and I couldn't have figured out the request/networking/cors issues without it, seemed like a valuable asset. As the comment alludes to at the top, it's used with a VS Code extension to send/inspect REST requests. I can move it elsewhere, or add more comments to the top, or just delete. Thoughts?

vscode/src/azure/commands.ts
billti: This is now an 'undocumented' setting. It will also be deleted soon also once the latest service changes finish rolling out to production.

vscode/src/azure/networkRequests.ts
billti: I spend hours digging through all the various JavaScript libraries for Azure authentication, ARM subscriptions, blob storage APIs, quantum requests, etc. They all were problematic is some ways, such as not working in Web Workers, using now deprecated libraries, having dependencies on other VS Code extensions being installed, etc.

Down the road we could look at adopting a dependency again to make this easier, but right now this was a confusing and problematic mess.

vscode/package.json
billti: I should probably just remove that setting and have Azure integration always on, now that we plan to have it ready by preview.

vscode/src/azure/networkRequests.ts
```diff
@@ -0,0 +1,291 @@
...
+    tenantId: string;
+    displayName: string;
+  };
+
+  export type SubscriptionList = {
+    value: Array<Subscription>;
+  };
+
+  export type Provider = {
+    providerId: string; // e.g., 'ionq', 'quantinuum', 'rigetti'
```
idavis, line 210: Remove partner IDs from code
```suggestion
    providerId: string;
```

vscode/src/azure/networkRequests.ts
```diff
@@ -0,0 +1,291 @@
...
+    limit: number;
+  };
+
+  export type QuotaList = {
+    nextLink: string | null;
+    value: Array<Quota>;
+  };
+
+  export type Target = {
+    id: string; // ionq.qpu, ionq.simulator, rigetti.sim.qvm, quantinuum.sim.h1-2e, etc.
```
idavis, line 247: Remove partner ids
```suggestion
    id: string;
```

vscode/src/azure/networkRequests.ts
```diff
@@ -0,0 +1,291 @@
...
+    currentAvailability: "Available" | "Degraded" | "Unavailable";
+    averageQueueTime: number; // minutes
+    statusPage: string; // url
+  };
+
+  export type ProviderStatus = {
+    id: string; // ionq, quantinuum, rigetti, etc.
```
idavis, line 254: Remove partner ids

```suggestion
    id: string;
```

vscode/src/azure/treeView.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+export class WorkspaceTreeProvider
+  implements vscode.TreeDataProvider<WorkspaceTreeItem>
+{
+  private treeState: Map<string, WorkspaceConnection> = new Map();
+
+  private _onDidChangeData = new vscode.EventEmitter<
+    WorkspaceTreeItem | undefined
+  >();
```
idavis, line 22: nit: why are private members using different naming conventions?

vscode/src/azure/treeView.ts
```diff
@@ -0,0 +1,277 @@
...
+  storageAccount: string;
+  endpointUri: string;
+  tenantId: string;
+  quota?: any;
+  providers: Provider[];
+  jobs: Job[];
+};
+
+export type Provider = {
+  providerId: string; // ionq, quantinuum, rigetti, etc.
```
idavis, line 82:
```suggestion
  providerId: string;
```

vscode/src/azure/treeView.ts
```diff
@@ -0,0 +1,277 @@
...
+  currentAvailability: "Available" | "Degraded" | "Unavailable";
+  targets: Target[];
+};
+
+export type Target = {
+  id: string; // ionq.qpu, ionq.simulator, rigetti.sim.qvm, quantinuum.sim.h1-2e, etc.
```
idavis, line 88:
```suggestion
  id: string;
```

vscode/src/azure/workspaceActions.ts
```diff
@@ -0,0 +1,374 @@
...
+    "quantinuum.sim.h2-1e",
+    "quantinuum.qpu.h2-1",
+    "rigetti.sim.qvm",
+    "rigetti.qpu.aspen-m-3",
+    "ionq.qpu-preview",
+    "ionq.qpu.aria-1-preview",
+    "ionq.qpu.aria-2-preview",
+    "ionq.simulator-preview",
+  ];
+
```
swernli, line 228: We shouldn't be hard-coding any target ids like this. I thought there was a way we could filter to targets that accept the input type `qir.v1`?

vscode/src/qirGeneration.ts
```diff
@@ -0,0 +1,94 @@
...
+  const worker = getCompilerWorker(compilerWorkerScriptPath);
+  const compilerTimeout = setTimeout(() => {
+    worker.terminate();
+  }, 10000);
```
idavis, line 54: nit: magic numbers should be const members

vscode/package.json
```diff
@@ -43,8 +66,47 @@
           "command": "qsharp-vscode.runEditorContents",
           "when": "resourceLangId == qsharp"
         }
+      ],
+      "view/title": [
+        {
+          "command": "qsharp-vscode.workspacesRefresh",
```
idavis, line 72: should we update the command prefixes to match the package ID? I'm not sure on the convention here.

idavis (changes requested):

vscode/package.json
billti: Thought about that... but we'll be publishing to two different packages with the dev and stable channels, so which one? :-)

I think that's a convention rather than a requirement. Unless someone else publishes a package called 'qsharp-vscode' (and follows the convention), it shouldn't be an issue.

vscode/src/azure/treeView.ts
billti: I was just following the docs on this one (https://code.visualstudio.com/api/extension-guides/tree-view#updating-tree-view-content). I can rename if need-be.

vscode/src/azure/workspaceActions.ts
billti: I don't see anywhere in the targets properties where it lists the input formats supported. (https://github.com/Azure/azure-rest-api-specs/blob/main/specification/quantum/data-plane/Microsoft.Quantum/preview/2022-09-12-preview/examples/providerStatus.json), and the only place inputDataFormat is used is in the job submission (https://github.com/Azure/azure-rest-api-specs/blob/main/specification/quantum/data-plane/Microsoft.Quantum/preview/2022-09-12-preview/quantum.json#L819).

We should request a service change so we can get this info back, and add a work item to make this dynamic once available. But in the meantime, I'd rather not allow jobs to be submitted to targets that we know won't work.

vscode/src/azure/REST API tests.http
billti: I've moved it to a dir at ./src/azure/test. I think it makes sense to keep it under the 'azure' directory, as it's very specific to that component/feature.

vscode/src/qirGeneration.ts
swernli: If the program is very heavy on classical computation and they are running on a lighter weight client, like VSCode on an iOS device via Safari + WASM, then it's possible they could bump into this. I agree that telemetry could make this an easier call, but for now I'd err on the side of more permissive of slower/longer computation.

vscode/src/azure/workspaceActions.ts
swernli: > But in the meantime, I'd rather not allow jobs to be submitted to targets that we know won't work.

I understand the sentiment, but that's not consistent with how our service interactions work anywhere else. The user can pick any target and try to submit any content to it, it may fail (and fail quickly) if that input type is not compatible with that target. Almost of all of our other APIs work this way. Even though we can publish quickly, keeping a whitelist of targets in the extension itself means we have to update the extension (and the user has to install it) any time the allowed targets changes. I'd much rather leave the door open to users submitting to targets that reject the input than maintain this list here, even for an initial preview.

billti: Again, we'll remove once the service can tell us. But fow now, it's not like the Azure integration code is generic, it only submits to Azure as written, and we know what targets support QIR. I'd rather hardcode for now and give the same desired end-user experience. It seems counter-intuitive to let the user try and eventually get some error back if we know in advance which targets will and won't work and can prevent that up front.

vscode/src/qirGeneration.ts
billti: Done.

vscode/src/azure/commands.ts
billti: Updated to point to the wiki. We can update with a aka.ms redirect later if needed (hopefully the service will fix the need for CORS changes well before we ship anyway)

vscode/package.json
billti: Removed the AzureQuantumIntegration option (it's not optional now)

idavis (approved):

wasm/src/lib.rs
```diff
@@ -37,6 +38,48 @@ pub fn git_hash() -> String {
...
+            Error::OutputFail(Span {
+                lo: 0,
+                hi: code.len() as u32,
+            }),
+            vec![],
+        ));
+    }
+
+    let package = store.insert(unit.0);
+
```
swernli, line 70: It's clearer to unwrap the tuple here.
```suggestion
    let (unit, errors) = qsc::compile::compile(
        &store,
        &[std],
        sources,
        PackageType::Exe,
        TargetProfile::Base,
    );

    // Ensure it compiles before trying to add it to the store.
    if !errors.is_empty() {
        // This should never happen, as the program should be checked for errors before trying to
        // generate code for it. But just in case, simply report the entire source as 'fail to generate output'
        return Err((
            Error::OutputFail(Span {
                lo: 0,
                hi: code.len() as u32,
            }),
            vec![],
        ));
    }

    let package = store.insert(unit);

```

wasm/src/lib.rs
```diff
@@ -37,6 +38,48 @@ pub fn git_hash() -> String {
     git_hash.into()
 }
 
+fn get_qir_internal(code: &str) -> std::result::Result<String, (Error, Vec<Frame>)> {
```
swernli, line 41: I was starting to make a suggestion that involved following the pattern of `run_internal` and using `stateless::Error` to return either an eval failure or compilation failure (since it has an `ErrorKind` for each of those), but changing the return type and marshalling the errors is too big of a change for a git-diff suggestion. I think we should circle back to that for a follow-up PR that cleans up the pattern here; yes, as the comment points out below, the only caller we have currently verifies the document state before calling into this, but since we have to call compile we might as well use the error kinds we have to return at least the first compilation error like `run_internal`.

swernli (approved):

swernli (approved):
One final suggestion and comment for future refactoring work, but otherwise I think we are ready to merge and iterate in follow up PRs.

wasm/src/lib.rs
billti: Yeah, I meant to make that change for readability. (Though I do note we often use the tuple.index pattern a bit in some places). Will do. Thanks.

Issue #596, opened 2023-08-16
Interpreter panic when error message contains a span from a previous line


minestarks:
FIR structs have a `Span` associated with them ([example](https://github.com/microsoft/qsharp/blob/4f7e49ea6a48bd83ee1e7ee9fd8afeaf53962c84/compiler/qsc_fir/src/fir.rs#L663)). However we don't save any context about which `SourceMap`s these `Span`s actually came from. If a runtime error contains a `Span` ([example](https://github.com/microsoft/qsharp/blob/4f7e49ea6a48bd83ee1e7ee9fd8afeaf53962c84/compiler/qsc_eval/src/intrinsic.rs#L67C49-L67C49)), we assume that the span is associated with the _current_ line ([here](https://github.com/microsoft/qsharp/blob/4f7e49ea6a48bd83ee1e7ee9fd8afeaf53962c84/compiler/qsc/src/interpret/stateful.rs#L404C38-L404C38)). This assumption isn't always correct.

**Repro Steps**

1. `cargo build` to build the interpreter binary
2. Run `qsi` and run the following in *two separate lines*
```qsharp
operation Main() : Unit { Microsoft.Quantum.Random.DrawRandomInt(2,1); }
```
```qsharp
Main()
```

**Expected**
Properly formatted `EmptyRange` message about `DrawRandomInt`

**Actual**
Panic:

```
qsi$ operation Main() : Unit { Microsoft.Quantum.Random.DrawRandomInt(2,1); }
qsi$ Main()
Error: empty range
Call stack:
    at Microsoft.Quantum.Random.DrawRandomInt in random.qs
    at Main in <expression>

error: Qsc.Eval.EmptyRange

  × runtime error
  ╰─▶ empty range
thread 'main' panicked at 'failed printing to stderr: formatter error', library/std/src/io/stdio.rs:1019:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: std::io::stdio::print_to
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/io/stdio.rs:1019:9
   3: std::io::stdio::_eprint
[...]

PR #622, opened 2023-08-24
Fix reporting of interpreter errors with spans across multiple lines.

### LINKED ISSUE #595: Some execution failures can cause Python interpreter to panic
(I don't have a small repro for this, so I've shared a notebook directly with a few folks, but can also add it to a gist somewhere if that is helpful for sharing more broadly.)

Under certain circumstances, the Python interpreter can end up in a state that causes a panic on an unwrap of an `Err`:

image

Near as I can tell, it seems like it would be coming from `format_errors` in interpreter.rs, but I'm not sure how that is possible. Updating these to `expect` with a message could help determine if that is indeed the source.

minestarks:
Fixes most of #596. The one case it doesn't address is when runtime errors have spans that should point to a different package than the current one, e.g. stdlib. I will tackle that in a different PR.

Also fixes #595 , minus the parsing of doc comments.

Lots of file changes (sorry, mostly tests!), but there are really only two key points:
- We now preserve already-executed interpreter lines in a `SourceMap` object that we keep appending to.
- The `WithSource` error wrapper can now contain more than one source file, which is handy when the error happens to contain multiple spans that come from different files.


compiler/qsc/src/bin/qsc.rs
```diff
@@ -98,11 +101,10 @@ fn main() -> miette::Result<ExitCode> {
...
     } else {
         for error in errors {
-            if let Some(source) = unit.sources.find_by_diagnostic(&error) {
-                eprintln!("{:?}", Report::new(error).with_source_code(source.clone()));
-            } else {
-                eprintln!("{:?}", Report::new(error));
-            }
+            eprintln!(
+                "{:?}",
+                Report::new(WithSource::from_map(&unit.sources, error))
```
minestarks, line 106: Essentially I removed all uses of `with_source_code` from the code base. To properly map spans to sources, you must always construct a `WithSource` wrapper, passing the whole `SourceMap` .

compiler/qsc/src/error.rs
```diff
@@ -1,56 +1,40 @@
...
     error::Error,
     fmt::{self, Debug, Display, Formatter},
 };
+use thiserror::Error;
 
-#[derive(Clone, Debug)]
-pub(super) struct WithSource<S, E> {
-    source: Option<S>,
+#[derive(Clone, Debug, Error)]
+pub struct WithStack<E>
```
minestarks, line 12: This is just a refactoring. `WithStack` adds a `stack_trace` to an object, which is something that used to be part of the old `WithSource` struct. Only runtime errors can have stacks, so I didn't feel the need to carry around an optional stack trace field with all `WithSource` types which are usually compiler errors without traces.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -358,13 +357,23 @@ impl Interpreter {
...
-            let source = line.into();
-            errors
-                .into_iter()
-                .map(|error| LineError(WithSource::new(Arc::clone(&source), error.into(), None)))
-                .collect::<Vec<_>>()
-        })?;
+        let line_label = format!("line_{}", self.lines);
+        self.lines += 1;
+
+        let mut fragments =
```
minestarks, line 363: This is where we "name" each line as we append it to the source map. This name can get surfaced later in error messages, such as:

```
Error: empty range
Call stack:
    at Microsoft.Quantum.Random.DrawRandomInt in random.qs
    at Main in <expression>
Qsc.Eval.EmptyRange

  × runtime error
  ╰─▶ empty range
   ╭─[line_0:1:1]
 1 │ operation Main() : Unit { Microsoft.Quantum.Random.DrawRandomInt(2,1); }
   ·                                                                 ──┬──
   ·                                                                   ╰── the range cannot be empty
   ╰────
```

compiler/qsc/src/interpret/stateful/tests.rs
```diff
@@ -48,7 +50,16 @@ mod given_interpreter {
                 .expect("interpreter should be created");
 
                 let (result, output) = line(&mut interpreter, "Message(\"_\")");
-                is_only_error(&result, &output, "name error: `Message` not found");
+                is_only_error(
```
minestarks, line 53: I added some additional validation to the existing test cases to make sure source spans were getting formatted correctly.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -94,50 +93,51 @@ impl SourceMap {
...
         let mut offset_sources = Vec::new();
-        for (name, contents) in sources {
-            offset_sources.push(Source {
-                name,
-                contents,
-                offset: next_offset(&offset_sources),
-            });
-        }
 
         let entry_source = entry.map(|contents| Source {
```
minestarks, line 97: The entry expression (if it exists) now comes first in the source map. This way we can append more sources to the end without the offsets conflicting with the entry expression's offset.

compiler/qsc_frontend/src/error.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use crate::compile::{Source, SourceMap};
+use miette::{Diagnostic, MietteError, MietteSpanContents, SourceCode, SourceSpan, SpanContents};
+use std::{
+    error::Error,
+    fmt::{self, Debug, Display, Formatter},
+};
+
+#[derive(Clone, Debug)]
+pub struct WithSource<E>
```
minestarks, line 15: `WithSource` is pre-existing, I just reworked it quite a bit. Now, the only way to instantiate it is with `from_map` which takes a `SourceMap`.

I also removed the optional stack trace field and moved it to a new struct definition (`WithStack` above).

compiler/qsc/src/bin/qsc.rs
DmitryVasilevsky: I think what you have in comments for the PR you should have in comments to the source code. For example, you could put a comment saying that we pass the whole map to make sure we report multi-file spans properly...

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -358,13 +357,23 @@ impl Interpreter {
...
 
-        let mut fragments = self.compiler.compile_fragments(line).map_err(|errors| {
-            let source = line.into();
-            errors
-                .into_iter()
-                .map(|error| LineError(WithSource::new(Arc::clone(&source), error.into(), None)))
-                .collect::<Vec<_>>()
-        })?;
+        let line_label = format!("line_{}", self.lines);
+        self.lines += 1;
```
swernli, line 361: nitpick: it would be nice if this were wrapped in a function that returned the string and performed the increment all together, something like `self.next_line_label()`. Even though this is only done one place, it makes clear that the increment and construction of a label should be done together. Not a high priority though.

swernli (approved):
One minor nitpick, otherwise looks good!

compiler/qsc/src/interpret/stateful.rs
minestarks: I like this a lot, thanks!

minestarks:
@billti and/or @idavis could you please sign off as codeowner?

idavis (approved):

PR #625, opened 2023-08-24
VS Code extension tests

minestarks:
Minimal setup for integration tests for the VS Code extension. 

Runs the Chrome browser in "headless" mode so it can (in theory) be run in CI.


vscode/test/runTest.ts
minestarks, line 1: Copied from https://github.com/microsoft/vscode-test-web/tree/3f0f858ab15cb65ef3c19564b0f5a6910ea9414e/sample/src/web/test

vscode/test/suite/index.ts
minestarks, line 1: Copied from https://github.com/microsoft/vscode-test-web/blob/3f0f858ab15cb65ef3c19564b0f5a6910ea9414e/sample/src/web/test/suite/index.ts , removing the webpack gobbledygook.

vscode/package.json
```diff
@@ -132,7 +132,9 @@
   "scripts": {
     "tsc:check": "node ../node_modules/typescript/bin/tsc -p ./tsconfig.json",
     "build": "npm run tsc:check && node build.mjs",
-    "run:web": "npx @vscode/test-web --extensionDevelopmentPath . ../samples"
+    "run:web": "npx --no-install @vscode/test-web --extensionDevelopmentPath . ../samples",
```
billti, line 135: Interesting note which may bite us in the future, but apparently 'npx' is deprecated and an alias for 'npm exec' now, and the `--no-install` option is also deprecated (replaced with `--no`)

See https://docs.npmjs.com/cli/v9/commands/npm-exec?v=true#compatibility-with-older-npx-versions

vscode/build.mjs
```diff
@@ -41,10 +53,19 @@ function copyWasm() {
...
+  build(extensionOptions).then(() =>
+    console.log(`Built bundle to ${extensionOptions.outdir}`)
+  );
+}
+
+function buildTests() {
+  console.log("Running esbuild");
+
+  build(testOptions).then(() =>
+    console.log(`Built bundle to ${testOptions.outdir}`)
```
billti, line 65: Copy/paste error. You didn't build the bundle, you built the tests.

vscode/package.json
```diff
@@ -132,7 +132,9 @@
...
+    "test": "node ./test/runTest.js",
```
billti, line 136: I was a little confused at first by the fact this file will be side-by-side with the .ts file, but the test suite files are placed in an ./out directory. After a bit of digging around I realized they are built separately - with one building the by a package.json task, and the other being built by the build.mjs script with different options. Any way we can unify build mechanisms and locations here to make this more intuitive?

minestarks:
Putting back into draft as I work to address some of the feedback. 

Will also add a folder structure for the test suite to separate debugger/language service/etc tests from each other.

vscode/package.json
minestarks: I think I'll change this to `runTests.js` - plain JS. If I move the build to `build.mjs` it's overkill as it doesn't need to be bundled or built with ESBuild. If I make it a `tsc` build step then I have to introduce a custom step that's different than anything else in the repo.

This is just a script that's just going to run on the dev machine, so it's more along the lines of `build.mjs` or `generate_katas_content.js`. We chose plain JavaScript for those scripts, so I think I'll just follow that lead and make this JS as well.

minestarks:
@billti this is ready for review again.

idavis (approved):

minestarks:
Looks like this _still_ needs approval from @billti (since @cesarzc is out).

build.py
minestarks, line 1: Reviewer: To save your sanity please hide whitespace diffs when reviewing this file.

image

vscode/build.mjs
minestarks, line 1: Upon closer look, I'm realizing building the tests here along with the product code isn't such a great idea. I think I'll pull tests out into their own `build.mjs` script.

1. We have a desire to run each test suite (e.g. language service, debugger) in a separate VS Code instance, which means that each suite needs to built into an individual bundle. Adding all of these build steps to the product build will clutter the script and add unnecessary overhead to CI builds. It's probably best that the test builds are contained under `vscode/test`.

2. When developing integration tests, it's nice to be able to just build the tests, without building the product, iterating quickly. Vice versa when working on product code. So it makes sense to separate these builds.

minestarks: This is done now.

vscode/package.json
```diff
@@ -249,10 +249,13 @@
     ]
   },
   "scripts": {
-    "tsc:check": "node ../node_modules/typescript/bin/tsc -p ./tsconfig.json",
     "build": "npm run tsc:check && node build.mjs",
     "build:watch": "node build.mjs --watch",
-    "run:web": "npx @vscode/test-web --extensionDevelopmentPath . ../samples"
+    "run:web": "npm exec --no @vscode/test-web --extensionDevelopmentPath . ../samples",
+    "start": "npm exec --no @vscode/test-web --extensionDevelopmentPath . ../samples",
```
billti, line 255: If this is identical to the line above, we should probably deduplicate and have just one.

vscode/package.json
```diff
@@ -249,10 +249,13 @@
...
+    "test": "node ./test/runTests.mjs",
+    "pretest": "npm exec --no playwright install && node ./test/buildTests.mjs",
```
billti, line 257: So is this the only time the tests get built or checked - when the tests are run?

test.py
```diff
@@ -0,0 +1,63 @@
...
+    step_end()
+
+npm_cmd = "npm.cmd" if platform.system() == "Windows" else "npm"
+
+root_dir = os.path.dirname(os.path.abspath(__file__))
+vscode_src = os.path.join(root_dir, "vscode")
+    
+step_start("Running the VS Code integration tests")
+vscode_args = [npm_cmd, "test"]
+subprocess.run(vscode_args, check=True, text=True, cwd=vscode_src)
```
billti, line 62: Seems a little excessive to add a new top level script with ~60 lines of Python just to run 'npm test' in the VS Code dir. Could this be an option on the main build script?

test.py
```diff
@@ -0,0 +1,63 @@
...
+
+parser.add_argument(
+    "build_args",
+    nargs="*",
+    help="When --build is used, arguments passed to build.py. Example: `test.py --build -- --debug`"
+)
+
+args = parser.parse_args()
+
+def step_start(description):
```
billti, line 38: Being that these are being duplicated in ./build.py as well, maybe import one from the other, or just de-duplicate the scripts into one as per the later comment.

vscode/test/runTests.mjs
```diff
@@ -0,0 +1,64 @@
...
+// download the latest Insiders build of VS Code for the Web and launch
+// it in a headless instance of Chromium to run the integration test suite.
+
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { runTests } from "@vscode/test-web";
+
+const thisDir = dirname(fileURLToPath(import.meta.url));
+
+async function main() {
```
billti, line 19: If this isn't exported and is just going to be invoked immediately after, why not just write it as top level statements? (Top-level await has been supported since Node 14).

vscode/test/buildTests.mjs
```diff
@@ -0,0 +1,38 @@
...
+  external: ["vscode"],
+  format: "cjs",
+  platform: "browser",
+  target: ["es2020"],
+  sourcemap: "linked",
+  //logLevel: "debug",
+  define: { "import.meta.url": "undefined" },
+};
+
+function buildTests() {
```
billti, line 30: Likewise, this top level function adds no value if it's not exported and just called once immediately after definition. Just use top-level statements.

vscode/test/runTests.mjs
minestarks: Good point, I was copying the sample but they probably wrote it when top-level await didn't exist.

vscode/package.json
minestarks: Any attachment to the naming of `run:web`? I liked `start` because it's idiomatic but I agree having both seems silly

vscode/package.json
minestarks: Yes. I can add `buildTests.mjs` to the regular build, but I didn't see a need to. 

They do get checked by eslint. But not by TypeScript, at least not in the build. You get TypeScript errors in the editor of course.

I can add in the checking steps for the test code, but again, I didn't feel it necessary. (I can be talked into it).

test.py
minestarks: Yes..... that would be so much better, actually.

billti (approved):
Couple minor comments, else looks good. Thanks!

minestarks:
I think this PR is stuck in an unmergeable state, because I'm changing the names of the required CI checks in this PR. Since all of the new checks _did_ pass, I will go ahead and bypass branch protections, and go double-check the branch protections [here](https://github.com/microsoft/qsharp/settings/branch_protection_rules/33737095) and [here](https://github.com/microsoft/qsharp/settings/branch_protection_rules/35454654) to make sure they're still valid.

image

PR #626, opened 2023-08-25
Removed or Moved LaTex Syntax from STD Doc Strings

ScottCarda-MS:
Limits the use of LaTex syntax to the Remarks second of doc strings in the SDT library files.

### DIFF
--- a/library/std/canon.qs
+++ b/library/std/canon.qs
@@ -21,7 +21,7 @@ namespace Microsoft.Quantum.Canon {
     /// The target on which the operation acts.
     ///
     /// # Example
-    /// Prepare a three-qubit $\ket{+}$ state:
+    /// Prepare a three-qubit |+⟩ state:
     /// ```qsharp
     /// use register = Qubit[3];
     /// ApplyToEach(H, register);
@@ -47,7 +47,7 @@ namespace Microsoft.Quantum.Canon {
     /// The target on which the operation acts.
     ///
     /// # Example
-    /// Prepare a three-qubit $\ket{+}$ state:
+    /// Prepare a three-qubit |+⟩ state:
     /// ```qsharp
     /// use register = Qubit[3];
     /// ApplyToEach(H, register);
@@ -77,7 +77,7 @@ namespace Microsoft.Quantum.Canon {
     /// The target on which the operation acts.
     ///
     /// # Example
-    /// Prepare a three-qubit $\ket{+}$ state:
+    /// Prepare a three-qubit |+⟩ state:
     /// ```qsharp
     /// use register = Qubit[3];
     /// ApplyToEach(H, register);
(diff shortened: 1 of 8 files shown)


library/std/intrinsic.qs
```diff
@@ -396,6 +377,16 @@ namespace Microsoft.Quantum.Intrinsic {
...
     /// # Remarks
+    /// \begin{align}
+    ///     R_1(n, k) \mathrel{:=}
+    ///     \operatorname{diag}(1, e^{i \pi k / 2^n}).
+    /// \end{align}
+    ///
+    /// > [!WARNING]
+    /// > This operation uses the **opposite** sign convention from
+    /// > @"microsoft.quantum.intrinsic.r", and does not include the
+    /// > factor of $1/ 2$ included by @"microsoft.quantum.intrinsic.r1".
```
swernli, line 388: I think it would be good to leave the warning in the summary so that it does show up in hover, since it alerts people to an important fact about how to use this operation.

library/std/intrinsic.qs
```diff
@@ -458,6 +438,16 @@ namespace Microsoft.Quantum.Intrinsic {
...
     /// # Remarks
+    /// \begin{align}
+    ///     R_{\mu}(n, k) \mathrel{:=}
+    ///     e^{i \pi n \sigma_{\mu} / 2^k},
+    /// \end{align}
+    /// where $\mu \in \{I, X, Y, Z\}$.
+    ///
+    /// > [!WARNING]
+    /// > This operation uses the **opposite** sign convention from
+    /// > @"microsoft.quantum.intrinsic.r".
```
swernli, line 449: Same here, probably worth including this in the summary (minus LaTex)

DmitryVasilevsky (approved):

library/std/intrinsic.qs
ScottCarda-MS: I've moved these to the summary sections as suggested, but I'm not sure about leaving in the `@` link syntax. It doesn't create links in the hover cards and may be confusing to users. Does the syntax create links in some other context?

ScottCarda-MS: I've removed the `@` syntax.

swernli: Sounds good, thanks!

library/std/canon.qs
```diff
@@ -125,9 +125,14 @@ namespace Microsoft.Quantum.Canon {
...
-    /// # Description
+    /// # Input
+    /// ## control
+    /// Control qubit for the CX gate.
+    /// ## target
+    /// Target qubit for the CX gate.
+    ///
+    /// # Remarks
     /// This operation can be simulated by the unitary matrix
-    /// $$
```
cesarzc, line 130: Is there a reason to remove the LaTex delimiters? How would automated and parsing tools be able to know where LaTex begins and ends?

cesarzc: For consistency and expressiveness, I would suggest enclosing LaTex between `$` and `$$` delimiters even if the tools that we have used so far do not require it.

For search purposes, if you want to look at where LaTex is used across the codebase, it's easier to look for the delimiters than to look for the delimiters and `\begin`.

cesarzc (approved):
Approving with a suggestion.

library/std/canon.qs
ScottCarda-MS: I've put the `$$` back in the canon.qs and for consistency added them to intrinsic.qs as well.

PR #627, opened 2023-08-26
Better breakpoint granularity and diagnostic information in debug adapter

idavis:
This PR adds

- metadata to the debug breakpoint information stored in the debugger extension.
- enables inline breakpoints and fixes issues with multi-line statements where breakpoints weren't being set correctly.
- breakpoints for if/while loops
  - more work needs to be done to support stepping for the conditional checks (while, elif) 
- filters bp requests to ensure we are processing Q# files
- trace instead of error logging when failing to load files. Almost all failures are false positives for cached BPs from previous sessions.

### DIFF
--- a/compiler/qsc_data_structures/src/span.rs
+++ b/compiler/qsc_data_structures/src/span.rs
@@ -16,6 +16,16 @@ pub struct Span {
     pub hi: u32,
 }
 
+impl Span {
+    #[must_use]
+    pub fn subtract(&self, offset: u32) -> Span {
+        Span {
+            lo: self.lo - offset,
+            hi: self.hi - offset,
+        }
+    }
+}
+
 impl Add<u32> for Span {
     type Output = Self;
 
--- a/compiler/qsc_data_structures/src/span.rs
+++ b/compiler/qsc_data_structures/src/span.rs
@@ -4,7 +4,7 @@
 use miette::SourceSpan;
 use std::{
     fmt::{self, Display, Formatter},
-    ops::{Add, Index},
+    ops::{Add, Index, Sub},
 };
 
 /// A region between two offsets in an array. Spans are the half-open interval `[lo, hi)`.
@@ -16,16 +16,6 @@ pub struct Span {
     pub hi: u32,
 }
 
-impl Span {
-    #[must_use]
-    pub fn subtract(&self, offset: u32) -> Span {
-        Span {
-            lo: self.lo - offset,
-            hi: self.hi - offset,
-        }
-    }
-}
-
 impl Add<u32> for Span {
     type Output = Self;
 
@@ -33,6 +23,17 @@ impl Add<u32> for Span {
         Self {
             lo: self.lo + rhs,
             hi: self.hi + rhs,
+        }
+    }
+}
+
+impl Sub<u32> for Span {
+    type Output = Self;
+
+    fn sub(self, rhs: u32) -> Self::Output {
+        Self {
+            lo: self.lo - rhs,
+            hi: self.hi - rhs,
         }
     }
 }
--- a/compiler/qsc_passes/src/replace_qubit_allocation/tests.rs
+++ b/compiler/qsc_passes/src/replace_qubit_allocation/tests.rs
@@ -47,7 +47,7 @@ fn test_single_qubit() {
                                 Stmt 9 [80-90]: Local (Immutable):
                                     Pat 10 [84-85] [Type Int]: Bind: Ident 11 [84-85] "x"
                                     Expr 12 [88-89] [Type Int]: Lit: Int(3)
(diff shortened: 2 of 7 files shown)


swernli (approved):

package.json
```diff
@@ -13,7 +13,8 @@
     "eslint:check": "eslint -c ./.eslintrc.cjs --max-warnings 0 ./",
     "prettier:check": "prettier -c --no-config ./",
     "check": "npm run eslint:check && npm run prettier:check",
-    "prettier:fix": "prettier -w --no-config ./"
+    "prettier:fix": "prettier -w --no-config ./",
+    "vscode:dev": "cd vscode && npx serve --cors -l 5000 --ssl-cert $HOME/certs/localhost.pem --ssl-key $HOME/certs/localhost-key.pem"
```
minestarks, line 17: This won't work on Windows. Also, if we're keeping it, I think this should go in `vscode/package.json` along with the other command for launching VS Code.

minestarks: It's also not entirely clear to me what advantage this has over the `run:web` command we already have in the other package.json?

`"run:web": "npx @vscode/test-web --extensionDevelopmentPath . ../samples"`

compiler/qsc_data_structures/src/span.rs
```diff
@@ -16,6 +16,16 @@ pub struct Span {
     pub hi: u32,
 }
 
+impl Span {
+    #[must_use]
+    pub fn subtract(&self, offset: u32) -> Span {
```
minestarks, line 21: Since we already impl `Add<u32>` for `Span`, I suggest implementing `Sub<u32>` for symmetry (and also for the convenience of using the `-` operator)

compiler/qsc_passes/src/replace_qubit_allocation/tests.rs
```diff
@@ -39,17 +39,17 @@ fn test_single_qubit() {
...
+                                    Expr 15 [55-71] [Type Qubit]: Call:
+                                        Expr 14 [55-71] [Type (Unit => Qubit)]: Var: Item 4 (Package 0)
+                                        Expr 16 [55-71] [Type Unit]: Unit
                                 Stmt 9 [80-90]: Local (Immutable):
                                     Pat 10 [84-85] [Type Int]: Bind: Ident 11 [84-85] "x"
                                     Expr 12 [88-89] [Type Int]: Lit: Int(3)
-                                Stmt 20 [59-60]: Semi: Expr 21 [59-60] [Type Unit]: Call:
-                                    Expr 19 [59-60] [Type (Qubit => Unit)]: Var: Item 5 (Package 0)
-                                    Expr 22 [59-60] [Type Qubit]: Var: Local 7
+                                Stmt 20 [0-0]: Semi: Expr 21 [0-0] [Type Unit]: Call:
```
minestarks, line 50: Is it expected that this went to `0-0`?

minestarks: Just checking - qubit deallocation _can_ raise an error, will the call stack be formatted properly when the statement / expr spans are `0-0`?

minestarks (changes requested):
Overall looks good, just a couple of questions

compiler/qsc_passes/src/replace_qubit_allocation/tests.rs
idavis: The only spans considered in stack traces are for items. The span above is for the hidden qubit deallocation which isn't user code.

package.json
idavis: It allows one to test in vscode.dev which is different than the web ext host. For windows, use WSL and this will work if you set up the environment. For which file it goes in, I'm really tired of moving into the vscode directory to run tasks. Do you have any suggestions?

minestarks: `npm -w vscode run ___` at the top level will let you run a script from `vscode/package.json` . 

I'm not opposed to adding tasks to the top-level package.json for convenience, I just think if we're going to do that, we should move the existing tasks up there too. Not a fan of having one-off scripts sprinkled across different files.

The script should be usable on Windows without WSL - more than a couple devs use Windows on the team, and our other build scripts account for this.

Maybe this can just go in your personal `tasks.json`?

minestarks: > vscode.dev which is different than the web ext host

I'm not sure about this - doesn't vscode/test-web serve a is a locally hosted vscode.dev ? I think they're the same app.

billti: Yes, I believe they're the same app.

Regarding the task, besides not working on Windows, is has options such as `--ssl-cert $HOME/certs/localhost.pem --ssl-key $HOME/certs/localhost-key.pem"`, so that's not going to work unless you do some prior setup to create certs in the correct location and with the correct names - none of which is documented in this PR.

Tasks we check-in should work out of the box or have docs on how to use them, and work on all supported platforms.

minestarks (approved):

PR #628, opened 2023-08-28
Items shouldn't need to be defined before use in `%%qsharp` cells

### LINKED ISSUE #594: Items shouldn't need to be defined before use in `%%qsharp` cells
For a Q# magic cell, the whole cell should be treated as a unit and allow type inference and definitions to appear anywhere within. Instead, a name cannot be used before it is defined, even for functions:
image
This does not match expectation for how item definition should work, which is different than the convention for bound variables.

swernli:
This change updates the way fragments are compiled to ensure that each stage of compilation (ie: parsing, item discovery, name resolution, type checking, etc) are performed on all fragments before progressing to the next stage. This allows for the expected Q# patterns to work as expected in cells, namely that namespaces can be mutually recursive and items can be used before they are defined (as long as they are defined later in the same cell). This also adds some additional test cases for non-incremental compilation to verify these behaviors there too.
Fixes #594


minestarks (approved):
It's what we discussed, so LGTM!

PR #630, opened 2023-08-28
Test ID uniqueness and IDs pattern unification in katas

cesarzc:


ScottCarda-MS (approved):

npm/test/basics.js
```diff
@@ -280,6 +280,39 @@ test("all katas work", async () => {
   );
 });
 
+test("IDs used in katas are unique", async () => {
```
minestarks, line 283: I'd kind of expect this to be part of `generateKatas.js` so that we can fail quickly at build time. Would that be easy?

minestarks (approved):

npm/test/basics.js
cesarzc: I considered that option too, but my initial thought was that it would be "safer" to check using the APIs that the frontend uses. At the end, it would use the same JS objects, so I guess it makes sense to move this check at build time. I'll change it.

cesarzc: Updated.

npm/generate_katas_content.js
```diff
@@ -566,6 +566,42 @@ function generateKataContent(path, globalCodeSources) {
   return kata;
 }
 
+function validateIdsUniqueness(katas) {
+  console.log("Validating IDs uniqueness across all katas");
+  const allIds = new Set();
+  const assertUniqueness = (id) => {
+    const isIdUnique = !allIds.has(id);
+    if (!isIdUnique) {
```
DmitryVasilevsky, line 574: This double negation makes my head spin :) May be name it alreadySeen or similar... Up to you.

npm/generate_katas_content.js
```diff
@@ -566,6 +566,42 @@ function generateKataContent(path, globalCodeSources) {
...
+    allIds.add(id);
+  };
+
+  for (const kata of katas) {
+    // Check kata IDs are unique.
+    assertUniqueness(kata.id);
+    for (const section of kata.sections) {
+      // Check section IDs are unique.
+      assertUniqueness(section.id);
+      if (section.type === "exercise") {
```
DmitryVasilevsky, line 586: If we change Kata structure and forget to update this code, we will run into a risk of duplicate ids. But I guess this is OK.

DmitryVasilevsky (approved):

npm/generate_katas_content.js
cesarzc: Renamed the variable to `idAlreadyExists` :)

npm/generate_katas_content.js
cesarzc: I'm not too worried about this because if the structure of katas changes, this code will probably fail anyway and would have to be updated.

PR #631, opened 2023-08-28
Use PyQIR to validate base profile output

### LINKED ISSUE #572: Test QIR output via PyQIR
To verify behavior of generated QIR and corresponding decompositions, we should compare the execution of Q# code in the evaluator to QIR with each of the decompositions linked in. One way to accomplish this would be to use [PyQIR](https://github.com/qir-alliance/pyqir) to consume the output .ll and any decompositions, link and generate assembly/bitcode, and then execute that bitcode via [qir-runner](https://github.com/qir-alliance/qir-runner). This would not need to be run across all platforms, so could be a Linux-specific workflow that runs as part of CI.

swernli:
This change adds new test logic to the Python tests to verify output QIR is valid and contains the expected gate sequence.
Fixes #572

### DIFF
--- a/build.py
+++ b/build.py
@@ -216,7 +216,7 @@ if build_pip:
             "test_requirements.txt",
         ]
         subprocess.run(qir_install_args, check=True, text=True, cwd=qir_test_dir)
-        pyqir_check_args = [python_bin, "-c", '"import pyqir"']
+        pyqir_check_args = [python_bin, "-c", "import pyqir"]
         if (
             subprocess.run(
                 pyqir_check_args, check=False, text=True, cwd=qir_test_dir
--- a/build.py
+++ b/build.py
@@ -215,15 +215,19 @@ if build_pip:
             "-r",
             "test_requirements.txt",
         ]
-        pyqir_install = subprocess.run(
-            qir_install_args, check=False, text=True, cwd=qir_test_dir
-        )
-        if pyqir_install.returncode == 0:
+        subprocess.run(qir_install_args, check=True, text=True, cwd=qir_test_dir)
+        pyqir_check_args = [python_bin, "-c", '"import pyqir"']
+        if (
+            subprocess.run(
+                pyqir_check_args, check=False, text=True, cwd=qir_test_dir
+            ).returncode
+            == 0
+        ):
             print("Running tests for the pip package with PyQIR")
             pytest_args = [python_bin, "-m", "pytest"]
             subprocess.run(pytest_args, check=True, text=True, cwd=qir_test_dir)
         else:
-            print("Skipping PyQIR tests")
+            print("Could not import PyQIR, skipping tests")
 
 if build_wasm:
     print("Building the wasm crate")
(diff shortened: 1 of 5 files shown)


swernli:
@billti, you brought up a good point about platform compatibility. If we want to use the build script to produce and test pip packages for platforms that PyQIR doesn't currently support (specifically, Windows and Linux arm64), then this will break that process. It seems like we have a couple options here and I'd like to get your preference:

1. Leave this change as is, and force build processes for Windows and Linux arm64 to run with `--no-test` when building pip packages. This would work, but would mean we aren't testing the Python module functionality (loading, running simulation) and that seems wrong.
2. Split the PyQIR-based tests into their own file that is not run by default in the build script, then update the CI to opt-in so the tests still run. This would make it easy for someone to run those tests locally (just give the option to build.py) but keeps them out of the normal path for publishing.
3. Split the PyQIR-based tests into their own file completely independent from the build script and have an extra step/task in the CI yaml that runs that file. Local execution would require running that separate file after making sure the environment is properly configured.

I'm in favor of number 2 myself.

billti:
2) Seems preferable. (Or just have the build script itself run the test by default ONLY on platforms that support PyQIR, and leave the CI and options as-is).

Will test_requirements.txt need to change? I assume we only want to try to install even on platforms where it will succeed. (And not expect/ignore errors on installing packages).

swernli:
> 2. Seems preferable. (Or just have the build script itself run the test by default ONLY on platforms that support PyQIR, and leave the CI and options as-is).
> 
> Will test_requirements.txt need to change? I assume we only want to try to install even on platforms where it will succeed. (And not expect/ignore errors on installing packages).

Yes, I'll make it an entirely separate test script that does not use the same test_requirements.txt. That way we don't get spurious failures due to failed dependency installation.

swernli:
Since it will take me a little while longer to get the independent test set up, I'm going to split out the bug fix from the this PR so that can get reviewed and merged sooner.

swernli:
Ok, I have a solution I think is viable. The PyQIR-based tests are separated out into their own `tests-qir` folder, and the build script has been updated with logic that will perform a speculative `import pyqir` call first to check whether the installation was successful. The result is that on platforms that PyQIR can run on the tests do run:

image

and on platforms PyQIR doesn't support it skips the tests:

image

build.py
```diff
@@ -205,6 +205,30 @@
...
+        # Try to install PyQIR and if successful, run additional tests.
+        qir_install_args = [
+            python_bin,
+            "-m",
+            "pip",
+            "install",
+            "-r",
+            "test_requirements.txt",
+        ]
+        subprocess.run(qir_install_args, check=True, text=True, cwd=qir_test_dir)
```
billti, line 218: Interesting.... so this has `check=True`, but if the above fails to install all the packages in `test_requirements.txt` it doesn't fail? :-/

swernli: PyQIR has an "any" package that acts as a fallback for installation. So installation will always succeed on all platforms, but if it installed the fallback then it will fail at import time instead. @idavis was explaining this to me yesterday when I saw other Python packages that use the same fallback pattern.

billti (approved):

PR #634, opened 2023-08-30
fixed wasm-pack version detection

filipw:
The change introduced in https://github.com/microsoft/qsharp/pull/412 broke `wasm-pack` version detection for `wasm-pack` 0.11+. It was meant to address the disappearance of `--version` command in 0.12.0 of `wasm-pack`, but the introduced workaround does not work for version 0.12 and higher anyway, because in the same release the version was also removed from the `--help` command 🙂 

But, as it turns out, `--version` command  in `wasm-pack` was only removed by mistake and [0.12.1](https://github.com/rustwasm/wasm-pack/releases/tag/v0.12.1) release restored it, so this PR re-introduces the proper check.


swernli (approved):

prereqs.py
```diff
@@ -164,15 +162,14 @@ def check_prereqs(install=False):
...
-            wasm_pack_version = wasm_pack_help.decode().splitlines()[0]
+            wasm_pack_version = subprocess.check_output(["wasm-pack", "--version"])
         else:
             print(
                 "wasm-pack not found. Please install from https://rustwasm.github.io/wasm-pack/installer/"
             )
             exit(1)
 
-    version_match = re.search(r"wasm-pack (\d+)\.(\d+).\d+", wasm_pack_version)
+    version_match = re.search(r"wasm-pack (\d+)\.(\d+).\d+", wasm_pack_version.decode())
```
billti, line 172: You should update the `wasmpack_ver` tuple on line 21 if this required 0.12.1 now.

swernli: Should it be updated though? 0.11.0 works, as does 0.12.1. It was just 0.12.0 that had the mistakenly removed `--version`.

filipw: correct 0.11 and lower all work too. But I updated it to 0.12.1 now. 

I didn't want to arbitrarily decide on raising min versions myself, but it makes sense to periodically require latest anyway, and it's better this way than some special case check for 0.12.0 only.

billti: Right. With this change, if you have 0.12.0 it will throw an error. It would be weird to allow 0.11.0, and 0.12.1, but error on 0.12.0. I think just requiring 0.12.1 or later is the right fix. Thanks!

billti (approved):

PR #636, opened 2023-08-30
Fix panic when QIR entry expr defines items

swernli:
This fixes a missed case after the changes in #628 that meant `compile_expr` could panic if the given expression was a block that included defining new items.

### DIFF
--- a/compiler/qsc/src/interpret/stateful/tests.rs
+++ b/compiler/qsc/src/interpret/stateful/tests.rs
@@ -29,6 +29,7 @@ mod given_interpreter {
     }
 
     mod without_sources {
+        use expect_test::expect;
         use indoc::indoc;
 
         use super::*;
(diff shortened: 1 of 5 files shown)


compiler/qsc/src/interpret/stateful/tests.rs
```diff
@@ -375,6 +376,403 @@ mod given_interpreter {
...
+                !1 = !{i32 7, !"qir_minor_version", i32 0}
+                !2 = !{i32 1, !"dynamic_qubit_management", i1 false}
+                !3 = !{i32 1, !"dynamic_result_management", i1 false}
+            "#]].assert_eq(&res);
+
+            // Can't validate here in this test, but note that the following call will cause
+            // `lower_fragments` (https://github.com/microsoft/qsharp/blob/e0c4dd334a81a0b0be82f4fbff0d850acc0a5fc8/compiler/qsc_frontend/src/incremental.rs#L235)
+            // to yield items that were previously declared in the qirgen call.
+            // this is because `lowerer.drain_items` is being called (it wasn't called
+            // previously as part of `qirgen()`).
```
minestarks, line 703: This comment can probably be removed

minestarks (approved):

compiler/qsc/src/interpret/stateful/tests.rs
swernli: Ah, good catch, thanks!

PR #640, opened 2023-08-31
Better Hover for Callable Parameters

ScottCarda-MS:
Adds a line to the top of the hover cards for locals to tell users if it is a parameter and if so, of what callable. Adds parsing of callable doc strings to get descriptions for parameters that will show up in the parameter's hover cards.


language_service/src/qsc_utils.rs
```diff
@@ -69,3 +70,27 @@ pub(crate) fn find_item<'a>(
     };
     (package.items.get(id.item), Some(package))
 }
+
+pub(crate) struct AstIdentFinder<'a> {
```
minestarks, line 74: I think it would be a cleaner interface if you exposed a `find_ident()` function instead of this struct. I see you use `AstIdentFinder` in two places above in the same way. Would make the call site cleaner if they called a simple function instead.

Something like:

```rust
fn find_ident(decl: &CallableDecl, node_id: NodeId) -> Option<Ident> {
  let mut finder = AstIdentFinder {
      node_id,
      ident: None,
  };
  finder.visit_callable_decl(decl);
  finder.ident
}
```

(Not sure if the above compiles, wrote it blind)

language_service/src/hover.rs
```diff
@@ -204,26 +300,68 @@ impl Visitor<'_> for HoverVisitor<'_> {
     }
 }
 
-fn markdown_with_doc(doc: &str, namespace: Option<Rc<str>>, code: impl Display) -> String {
-    let parsed_doc = parse_doc(doc);
+fn protocol_span(span: qsc::Span) -> protocol::Span {
```
minestarks, line 303: This doesn't matter in practice today since we only deal with single files, but... technically, `span.lo` and `span.hi` are offsets into the `SourceMap`, whereas `protocol::Span` expects an offset into the current source. Once we move to the "multiple file world" we're going to notice the offsets start being off. To get ahead of that, I'd suggest you take the `SourceMap` here and use `find_by_offset` to get the correct source offset and subtract it. (definition.rs does this correctly today, as an example).

language_service/src/hover.rs
```diff
@@ -204,26 +300,68 @@ impl Visitor<'_> for HoverVisitor<'_> {
...
+    local_name: &str,
+    callable_name: &str,
+    callable_doc: &str,
+) -> String {
+    match param_kind {
+        LocalKind::Param => {
+            let param_doc = parse_doc_for_param(callable_doc, local_name);
+            with_doc(
+                &param_doc,
+                format!("param of `{callable_name}`\n{markdown}",),
```
minestarks, line 342: "parameter of" would sound more human-friendly IMO 🙂 

Also, does the single line break `\n` work well visually? Markdown sometimes requires double line breaks for proper visual separation. (I can't tell from reading this code whether this will look OK, just worth checking)

language_service/src/hover.rs
```diff
@@ -204,26 +300,68 @@ impl Visitor<'_> for HoverVisitor<'_> {
...
+            )
+        }
+        LocalKind::LambdaParam => format!("lambda param\n{markdown}"),
+        LocalKind::Local => format!("local\n{markdown}"),
```
minestarks, line 346: Ehh, I think we can skip the "local" descriptor for a more uncluttered look. I don't think it adds any new information for the user.

language_service/src/hover/tests.rs
```diff
@@ -831,3 +844,40 @@ fn callable_empty_summary() {
...
+    "#},
+        &expect![[r#"
+            param of `Foo`
+            ```qsharp
+            x: Int
+            ```
+            ---
+            Doc string for `x`
+            ### Note
+            note for `x`
```
minestarks, line 880: Same point as my comment about line breaks above. I suggest you check this output with a couple of different markdown renderers. I would expect double line breaks between each block here for proper formatting, but I may be wrong.

minestarks (approved):

language_service/src/hover/tests.rs
ScottCarda-MS: I've looked at it on both the playground and in VS Code, and it looks fine in both to me. What other renderers would you recommend I check it against?

language_service/src/hover.rs
ScottCarda-MS: Changed to "parameter of". The single line break seems to work well. Let's follow up with that in the other comment.

language_service/src/hover.rs
ScottCarda-MS: Oh that is a good call out. That would have been tricky to debug down the line. I've updated the code to use source offsets. Thanks!

language_service/src/qsc_utils.rs
ScottCarda-MS: Added `find_ident` and make the `AstIdentFinder` private.

language_service/src/hover.rs
ScottCarda-MS: My end-goal for these is to have the hover for each thing tell the user the _kind_ of object they are hovering over in a consistent way, i.e. `parameter`, `local`, `operation`, `function`, `newtype`, `field`, and any additional kinds. These would appear left of the name in blue (as is currently the case for `operation`, `function`, and `newtype`), with the container name above (namespace/module name, callable name, UDT name, "lambda"). It would be even better if we could add links to the hover card for people to click through to definition for those callables and UDTs.

ScottCarda-MS: I'm not currently sure how to make the text appear blue in the same way as the `operation` or `function` though, as that is done by the fact that the markdown is using qsharp keyword recognition.

language_service/src/hover/tests.rs
minestarks: Oh that sounds good then. I think my previous markdown mistakes made me paranoid. I just pasted it into this GitHub comment and checked "Preview" too and I saw no problem. Thanks for checking!

language_service/src/hover.rs
minestarks: I don't think there's an easy way for us to provide formatting besides Q# syntax highlighting. Taking a page from TypeScript, instead of adding a word like "local", they synthesize some code like `let foo = 3;` to show how the local was declared. ([See this example ](https://www.typescriptlang.org/play?#code/PQKhCgAIUgBAHAhgJ0QW0mgnk1GAKK6ApgC7HKQAmA9gMYCuaxAdqYqQJY0tQjDgAZgxZ0uPTFmGjxLABTZc6AFyQAyqWScWAcwCUkAN5RIpujwDOpSeZZXIAXkgAiABadnAbhOmANmUkANxRHF1diX18aLx9IW3tfdAAjKkRQuUS0FMQlNGUNLV1IAwcAPiNITOzcyABfb1jFIjRvU1NseNJWtuxg5G9a8HBsaTFuFiA) - hover over the names). I checked Rust and they basically do the same thing. That looks similar to how we format `operation` and `function`s - we make up some hover text that looks just like declaration itself. What do you think of doing this for locals too?

For parameters, I see TypeScript just tacks on `(parameter)` and Rust doesn't bother calling them out. But I like your solution the best of all three.

ScottCarda-MS: Discussed design offline. First line of hover card will be location information for the symbol, i.e. "local", "parameter of ...", namespace, etc. Second line will be signature. Subsequent lines will be doc comments.

PR #466, opened 2023-07-17
Adds Publishing Py Packages

kuzminrobin:

PR #641, opened 2023-08-31
Set up package publishing

idavis:
Replaces https://github.com/microsoft/qsharp/pull/466/


Cargo.toml
```diff
@@ -52,4 +53,5 @@ pyo3 = { version = "0.18", features = ["abi3-py37", "extension-module"] }
 quantum-sparse-sim = { git = "https://github.com/qir-alliance/qir-runner", rev = "e96b9a12a43e98f83f2d4d499f46d651d0c6c341", default-features = false }
 
 [profile.release]
+debug = 1
```
billti, line 56: Does this add value if the next line is 'strip = true'? Doesn't that strip any debug info anyway?

pip/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
-name = "qsharp_python"
-version = "0.0.0"
+name = "qsharp"
```
billti, line 2: I thought we were going with `qsharp-preview` to avoid any risk with the existing `qsharp` package for now. CC @minestarks

Cargo.toml
swernli: I think this may have been a weird merge side effect. I removed `debug = 1` in a different change but I think these same lines were edited in the original PR this branch was made from.

idavis: `strip=true` removes the symbols but not the debug information. There were multiple edits across the branches, so I reduced the set here so we can test the actual effect. The way it was applied before had serious performance implications.

pip/Cargo.toml
idavis: I set up a script which can apply the nightly or preview suffix to the package names. Unless you want to default to qsharp-preview for all packages, and then change the whole name to qsharp-nigthly and qsharp.

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
-      python build.py --wasm --npm --vscode
+      python build.py --wasm --npm --vscode --no-check --no-test --no-check-prereqs
     displayName: Build VSCode Extension
-      
+    condition: and(eq(variables['Agent.OS'], '${{ variables.VSCodeExtensionBuildPlatform }}'), eq(variables['arch'], 'x86_64'))
+
   - script: |
-      cp ../LICENSE.txt .
-      vsce package $(Ver.Major).$(Ver.Minor).$(Ver.Patch)
+      python -c "import shutil; shutil.copy2('../LICENSE.txt', '.')"
```
minestarks, line 59: Can we add this copy step as a [pre-publish step](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prepublish-step) into the `package.json` file? Without it, when I do `vsce package` from the command-line for testing, I always get a complaint about a missing LICENSE.txt which I have to interactively dismiss. If we added the pre-publish step, then `vsce` would just run it automatically.

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
+        --env ARCH=$ARCH --rm -i -v `pwd`:/io \
+        quay.io/pypa/manylinux_2_28_$ARCH \
+        /io/pip/build_manylinux.sh
+
+      ls target/wheels
+    displayName: Build Platform-Dependent Py Packages, Linux
+    condition: eq(variables['Agent.OS'], 'Linux')
+
+  - script: |
+      python build.py --jupyterlab --no-test --no-check-prereqs
```
minestarks, line 99: Add `--no-check` here as well? Just for symmetry with the other commands

build.py
```diff
@@ -95,6 +105,33 @@
...
+if os.environ.get("QSHARP_PACKAGE_VERSION") is not None:
+    version = os.environ["QSHARP_PACKAGE_VERSION"]
+
+    update_package_version_args = [
+        sys.executable,  # use the current python, we aren't installing anything
+        "update_package_version.py",
+        version,
+    ]
+    print(f"Updating packages version to {version}")
+    subprocess.run(update_package_version_args, check=True, text=True, cwd=root_dir)
```
minestarks, line 134: I prefer the way we call out to `prereqs.py` above, where it's just a normal function import. `prereqs.py` is also runnable as a standalone script (the `if __name__ == "__main__":` check below). 

Since these are both helper scripts that get called from `build.py`, it'd be nice if we followed the same pattern in both. I like symmetry.

build.py
```diff
@@ -95,6 +105,33 @@
 wheels_dir = os.path.join(root_dir, "target", "wheels")
 vscode_src = os.path.join(root_dir, "vscode")
 jupyterlab_src = os.path.join(root_dir, "jupyterlab")
+package_name = "qsharp-nightly"
+
+if os.environ.get("QSHARP_PACKAGE_SUFFIX") is not None:
```
minestarks, line 110: Could you add a comment explaining _who_ might define these variables (this and `QSHARP_PACKAGE_VERSION`)? A build pipeline?

pip/pyproject.toml
```diff
@@ -19,8 +19,11 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-jupyterlab = ["qsharp_jupyterlab"]
+jupyterlab = ["qsharp_jupyterlab_nightly"]
```
minestarks, line 22: Change to dash? I think?

minestarks (approved):

pip/Cargo.toml
billti: I see in the npm package.json and jupyterlabs package.json you have the `-nightly` suffix on the name. Why the difference here?

minestarks: This name is ignored when the package is built for publishing (through `pip` and not Cargo). It's only ever "seen" when you run `cargo build`. It could be anything that doesn't conflict with an existing Cargo package name in this repo. 

We have questions about this name literally every time this file is touched in a PR, and it's already been changed back-and-forth between `qsharp` and `qsharp_python` a couple times. I don't think we'll be able to find the magic name to eliminate all confusion :D

build.py
billti: Also, being that they are largely doing similar work on roughtly the same set of files, would seem optimial do have one script for updating name and version, rather than two that duplicate a lot of code/work.

build.py
```diff
@@ -166,6 +203,12 @@
         # Already in a virtual environment, use current Python
         python_bin = sys.executable
 
+    # copy the process env vars
+    pip_env: dict[str, str] = {**os.environ}
```
billti, line 207: Looking around, seems `pip-env = os.environ.copy()` might be more canonical (and a little more intuitive and readable)

pip/build_manylinux.sh
```diff
@@ -0,0 +1,36 @@
...
+#       ARCH: Architecture: x86_64 or aarch64, etc.
+# Output:
+#   "<this repo>/target/wheels/qsharp-preview-*-cp3*-abi3-manylinux*_$ARCH.whl", 
+#       e.g., "<this repo>/target/wheels/qsharp-preview-0.0.7-cp37-abi3-manylinux_2_28_x86_64.whl".
+
+set -x -e
+
+# Install Rust:
+curl --proto '=https' --tlsv1.2 -sS https://sh.rustup.rs > rustup-init.sh
+chmod +x rustup-init.sh
```
billti, line 19: Would we be able to easily switch between msrustup and public rust here?

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
+        arch: x86_64
+      linux_ARM64:
+        imageName: 'ubuntu-latest'      # Manylinux Py (ARM64). Takes 45 - 65 minutes, hence is in a separate job.
+        arch: aarch64
+      mac:
+        imageName: 'macOS-latest'       # MacOS-specific Py (Mac is usually quite limited).
+        arch: x86_64
+      windows:
+        imageName: 'windows-latest'     # Win-specific Py + Platform-independent Py.
+        arch: x86_64
```
billti, line 32: Should we add a comment/todo about Windows on ARM64?

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
+  - script: |
       npm install -g @vscode/vsce
-    displayName: Install VSCE Pub Prereqs
-    
+    displayName: Install Prereqs for VSCode Extension
+    condition: and(eq(variables['Agent.OS'], '${{ variables.VSCodeExtensionBuildPlatform }}'), eq(variables['arch'], 'x86_64'))
+   
   - script: |
-      python build.py --wasm --npm --vscode
+      python build.py --wasm --npm --vscode --no-check --no-test --no-check-prereqs
```
billti, line 54: Are we skipping the checks/tests/prereqs here purely for time saving? They should all pass and only take a min or so, right?

.ado/publish.yml
billti: That would be useful (I always ended up copying manually myself also when running local). In reality, there's probably no harm in just also checking in LICENSE.txt there, and avoiding any need for additional steps - manual or otherwise.

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends \
+          qemu qemu-system-misc qemu-user-static qemu-user
+      fi
+
+      chmod +x pip/build_manylinux.sh
+
+      docker run \
+        --env ARCH=$ARCH --rm -i -v `pwd`:/io \
+        quay.io/pypa/manylinux_2_28_$ARCH \
```
billti, line 91: I had to search around to find what this is. Maybe add a comment linking to `https://github.com/pypa/manylinux#docker-images` for the newbies like me.

.ado/publish.yml
```diff
@@ -1,45 +1,193 @@
...
+        /io/pip/build_manylinux.sh
+
+      ls target/wheels
+    displayName: Build Platform-Dependent Py Packages, Linux
+    condition: eq(variables['Agent.OS'], 'Linux')
+
+  - script: |
+      python build.py --jupyterlab --no-test --no-check-prereqs
+    displayName: Build Platform-Independent Py Packages
+    condition: eq(variables['Agent.OS'], '${{ variables.PlatfIndepPyPackBuildPlatf }}')
```
billti, line 101: Does this need to build on x64 and ARM64, or should it limit to one arch here also?

.ado/publish.yml
idavis: I like the duplicate file idea. no extra build steps, and we won't likely be ever changing the license.

.ado/publish.yml
idavis: I assume so, but will have to run to see.

.ado/publish.yml
idavis: One arch should be fine, the jupyterlab package is x-plat

build.py
idavis: `{**os.environ}` is what we used in the qdk, but `copy()` works too.

pip/build_manylinux.sh
idavis: We only can use msrustup in ado, which is where this is being used.

billti:
Now done in #660 .

PR #645, opened 2023-09-01
Add telemetry to VSCode extension

sezna:
This PR adds telemetry to the VScode extension. It uses the extension-telemetry library, meaning it respects VScode telemetry settings. 

This PR is intended to only add the framework and some initial events. My hope is that we can agree on this structure, and add more events in a lightweight follow-up PR. Event data is fully typed and located [here](https://github.com/microsoft/qsharp/pull/645/files#diff-32ed3aba5fd5923bc613377d34d89f319123f812affb2cac846d5e416c58d55eR17). The events added by this PR are   `DebugSessionStart`, `Initialize`, `LoadLanguageService`, and `QSharpJupyterCellInitialized`.

### DIFF
--- a/vscode/src/telemetry.ts
+++ b/vscode/src/telemetry.ts
@@ -62,4 +62,9 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
     return;
   }
   reporter.sendTelemetryEvent(event, properties, measurements);
+  log.debug(
+    `Sent telemetry: ${event} ${JSON.stringify(properties)} ${JSON.stringify(
+      measurements
+    )}`
+  );
 }


vscode/src/extension.ts
```diff
@@ -156,10 +158,18 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
  * Loads the Q# compiler including the WASM module
  */
 async function loadLanguageService(baseUri: vscode.Uri) {
+  const start = performance.now();
   const wasmUri = vscode.Uri.joinPath(baseUri, "./wasm/qsc_wasm_bg.wasm");
   const wasmBytes = await vscode.workspace.fs.readFile(wasmUri);
   await loadWasmModule(wasmBytes);
-  return await getLanguageService();
+  const end = performance.now();
```
billti, line 165: Shouldn't this be after you await getting the language service (line below) if you're timing how long the language service takes to load?

vscode/src/telemetry.ts
```diff
@@ -0,0 +1,56 @@
+import * as vscode from "vscode";
+import TelemetryReporter from "@vscode/extension-telemetry";
+import { log } from "qsharp";
+// the application insights key (also known as instrumentation key)
+const key = "175861b7-3a41-4015-9571-1d930b8b0722";
```
billti, line 5: I think the common pattern is to put the key in the package.json and fetch it from there, rather than embed it into the code.

billti (changes requested):
See comments. The threading/worker issues is the main concern.

vscode/src/telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+  [EventType.DebugSessionStart]: {
+    properties: Empty;
+    measurements: {
+      timeToStart: number;
+    };
+  };
+  [EventType.LoadLanguageService]: {
+    properties: Empty;
+    measurements: {
+      timeToStart: number;
```
minestarks, line 32: In general when dealing with durations I like putting the units in the property name e.g. `timeToStartMs`

vscode/src/telemetry.ts
```diff
@@ -4,9 +4,7 @@ import TelemetryReporter, {
   TelemetryEventProperties,
 } from "@vscode/extension-telemetry";
 import { log } from "qsharp";
-// the application insights key (also known as instrumentation key)
-const key = "175861b7-3a41-4015-9571-1d930b8b0722";
-
+import { getPackageInfo } from "./packageInfo.js";
```
billti, line 7: There is no packageInfo.js. Should it be in this PR?

vscode/src/telemetry.ts
```diff
@@ -48,10 +46,18 @@ type WrappedTelemetryEvent = {
...
+  const packageInfo = getPackageInfo(context);
+  if (!packageInfo) {
+    return;
+  }
   // see issue here: https://github.com/microsoft/vscode-extension-telemetry/issues/183
   // we cannot use the latest version of extension-telemetry until this is fixed
-  const reporter = new TelemetryReporter("qsharp-vscode", "0.0.0", key);
+  const reporter = new TelemetryReporter(
+    "qsharp-vscode",
+    "0.0.0",
```
billti, line 58: What is the "0.0.0" here for? Looking at the constructor, this doesn't look like a 'ReplacementOptions[]' (https://github.com/microsoft/vscode-extension-telemetry/blob/main/dist/telemetryReporter.d.ts#L34)

vscode/src/telemetry.ts
```diff
@@ -0,0 +1,77 @@
...
+type WrappedTelemetryEvent = {
+  id: string;
+  data?: {
+    measurements: TelemetryEventMeasurements;
+    properties: TelemetryEventProperties;
+  };
+};
+
+export function initTelemetry() {
+  const extension = vscode.extensions.getExtension("quantum.qsharp-vscode");
```
billti, line 49: Pass the context in from the extension activation and get the current extension from the context (see https://code.visualstudio.com/api/references/vscode-api#ExtensionContext). Especially as we'll have two different extensions (stable and 'dev'), we want to avoid hardcoding the id in anywhere.

vscode/src/telemetry.ts
```diff
@@ -0,0 +1,77 @@
...
+    ({
+      id,
+      data: { properties, measurements } = { properties: {}, measurements: {} },
+    }: WrappedTelemetryEvent) =>
+      reporter.sendTelemetryEvent(id, properties, measurements)
+  );
+  sendTelemetryEvent(EventType.Initialize, {}, {});
+}
+
+export function sendTelemetryEvent<E extends keyof EventTypes>(
```
billti, line 71: The intent was that code, be it in npm, playground, or VS Code, would call log.logTelemetry to send events, and then whatever telemetry consumer had been wired-up would be hooked into that. Seems odd here that we're exposing a sendTelemetryEvent API for use in VS Code, and then that calls log.logTelemetry. Can we not just wire up so the VS Code source also calls log.logTelemetry? Happy to refactor that API if you think it's too loose.

billti: > Since, if I'm not mistaken, the VSCode extension is the only consumer of the API

You are mistaken :-)  A search of the playground code will show one place where we call it (mostly to prove it worked, happy for you to refactor that also)

vscode/src/telemetry.ts
```diff
@@ -0,0 +1,65 @@
...
+  event: E,
+  properties: EventTypes[E]["properties"] = {},
+  measurements: EventTypes[E]["measurements"] = {}
+) {
+  if (reporter === undefined) {
+    log.trace(`No telemetry reporter. Omitting telemetry event ${event}`);
+    return;
+  }
+  reporter.sendTelemetryEvent(event, properties, measurements);
```
billti, line 64: One final nit: It might be good to add a log.debug here also to log any telemetry we're sending, so we can easily hook into it and troubleshoot if need be.

billti (approved):

PR #646, opened 2023-09-02
Parsing Error with UDTs based on Arrays

### LINKED ISSUE #437: Parsing Error with UDTs based on Arrays
The following code should be valid:
```qsharp
namespace Foo { newtype Bar = (Int => Double)[]; }
```

However, when compiling, it claims that it expects a `;` where it encountered the `[`. This is especially confusing because simpler array types _do_ work as expected:
```qsharp
namespace Foo { newtype Bar = Double[]; }
```

Interestingly, in contexts where UDT field names are not allowed it seems to work correctly:
```qsharp
namespace Foo {
    newtype Bar1 = x: (Int => Double)[];
    newtype Bar2 = (Int => (String, Double)[]);
}
```

swernli:
This updates the logic for parsing UDT field definitions to treat any `TyDef` with no named fields as a single field rather than a tuple of fields, and in that case allow that field's type to be an array or arrow type. This ends up being simpler and more consistent, since individual unnamed field entries are not directly addressable. All user-facing functionality thus remains the same: you access unnamed fields by using the unwrap (`!`) unary operator, while named fields are treated as separate entities tht are individually addressable.
Fixes #437

### DIFF
--- a/compiler/qsc_frontend/src/lower/tests.rs
+++ b/compiler/qsc_frontend/src/lower/tests.rs
@@ -372,11 +372,8 @@ fn lift_newtype_tuple() {
                 Item 1 [18-46] (Public):
                     Parent: 0
                     Type (Ident 0 [26-29] "Foo"): UDT [18-46]:
-                        TyDef [32-45]: Tuple:
-                            TyDef [33-36]: Field:
-                                type: Int
-                            TyDef [38-44]: Field:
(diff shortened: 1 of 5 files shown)


billti (approved):

ScottCarda-MS (approved):
Thanks for fixing this issue!

PR #647, opened 2023-09-02
Shared settings should include tests for clippy scans

swernli:
The shared settings file correctly configures rust-analyzer to use clippy as the default check command, but should also include the setting that passes `--all-targets` as an additional command. This ensures that test code also gets clippy lints in the editor.


swernli:
Looks like this doesn't work the way I thought, and the all targets flag is already included. My issue seeing clippy results for test files seems to have been an intermittent rust-analyzer sync issue.

PR #648, opened 2023-09-02
Two errors at EOF when parser error recovery occurs

### LINKED ISSUE #431: Two errors at EOF when parser error recovery occurs
Some syntax errors cause two error messages to be emitted with the same location (EOF). The messages seem to contradict each other because they mention expecting different tokens.

Code:

[Playground link](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACstLzE0tLkhMTlXwrQwsTcwrKc11LChQqOZSAIL8gtQ8Bd%2FM5KL84vy0Ej2ovJ5LZmJ6Xn5xSWZysTUXWKGDa15JUWVAfmZeiYYmTGtRYklmPlB%2FYmaehqaClUJQanFpTkl0rAIALT3ts3QAAAA%3D)

```qsharp
namespace MyQuantumApp {

    @EntryPoint()
    operation Main() : Result[] 
```

Produces errors:

```
main.qs@(5,33): syntax error: expected `{`, found EOF
main.qs@(5,33): syntax error: expected `}`, found EOF
```

**Expected behavior**

A single syntax error

`git bisect` narrowed this down to #380.

swernli:
Avoids duplicate EOF errors by ignoring additonal EOFs when recovery has already happened.
Fixes #431

### DIFF
--- a/compiler/qsc_parse/src/scan.rs
+++ b/compiler/qsc_parse/src/scan.rs
@@ -20,6 +20,7 @@ pub(super) struct Scanner<'a> {
     tokens: Lexer<'a>,
     barriers: Vec<&'a [TokenKind]>,
     errors: Vec<Error>,
+    recovered_eof: bool,
     peek: Token,
     offset: u32,
 }
@@ -36,6 +37,7 @@ impl<'a> Scanner<'a> {
                 .into_iter()
                 .map(|e| Error(ErrorKind::Lex(e)))
                 .collect(),
+            recovered_eof: false,
             peek: peek.unwrap_or_else(|| eof(input.len())),
             offset: 0,
         }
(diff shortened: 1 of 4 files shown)


billti (approved):

compiler/qsc_parse/src/scan.rs
```diff
@@ -98,7 +100,11 @@ impl<'a> Scanner<'a> {
     }
 
     pub(super) fn push_error(&mut self, error: Error) {
-        self.errors.push(error);
+        let is_eof_err = matches!(error.0, ErrorKind::Token(_, TokenKind::Eof, _));
```
minestarks, line 103: What about when the error is not an `ErrorKind::Token`, such as https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACstLzE0tLkhMTlXwrQwsTcwrKc11LChQqOZSAIL8gtQ8Bd%2FM5KL84vy0Ej2ovJ5LZmJ6Xn5xSWZysTUXWKGDa15JUWVAfmZeiYYmTGtRYklmfp4CAJN6eMhiAAAA ?

minestarks: I think it may make sense to make this check less strict - only allow one parser error at EOF, regardless of the exact error kind.

swernli: After some discussion, I took an alternate approach: stop recovering at EOF. It changes some behavior in other tests, but seems appropriate: if we fail to parse an item because we hit EOF we will also, always fail to parse the namespace due to EOF. The difference is that now that fails the whole namespace parse such that other fully formed items in that namespace will no longer be in the output HIR (see for example the test renamed from `recover_unclosed_namespace` to `no_recover_unclosed_namespace`). This will affect hover and completion logic on the file until there is enough text in the file to allow recover at a non-EOF token. Is that acceptable behavior? To my mind it could be, since modern editors tend to always create `{}` pairs when writing and authoring in a missing closing delimiter at EOF state is very uncommon.

swernli: Ok, reverted the "no recovery at EOF" strategy and when back to what I had here, with the added logic to include `ErrorKind::Rule` since that is the other kind of error that can occur at EOF. I think overall this gets us the right middle ground between recovery to support language service featuers and avoiding spamming the user with EOF-related errors.

minestarks (approved):

PR #649, opened 2023-09-03
Return statements in lambdas are checked against the surrounding scope's expected return type

### LINKED ISSUE #510: Return statements in lambdas are checked against the surrounding scope's expected return type
**Describe the bug**

Return statements within lambda functions unify with the wrong type.

**To Reproduce**

```qsharp
namespace MyQuantumApp {
    @EntryPoint()
    operation Main() : Unit {
        let foo = () => {
            return 42;
        };
        return ();
    }
}
```

The more information you can provide, the more likely someone will be successful at reproducing the issue and finding a fix.

**Expected behavior**

There should not be an error for the line `return 42;` -- the lambda's return type should be inferred as a fresh type variable.

Note that this can be worked around with implicit returns, so the below works:

```qsharp
namespace MyQuantumApp {
    @EntryPoint()
    operation Main() : Unit {
        let foo = () => {
            42
        };
        return ();
    }
}
```

swernli:
Fixes #510

### DIFF
--- a/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/compiler/qsc_frontend/src/typeck/rules.rs
@@ -329,11 +329,13 @@ impl<'a> Context<'a> {
             ExprKind::Lambda(kind, input, body) => {
                 let input = self.infer_pat(input);
                 let prev_ret_ty = self.return_ty.take();
-                let output_ty =
-                    self.inferrer.fresh_ty(TySource::not_divergent(body.span));
+                let output_ty = self.inferrer.fresh_ty(TySource::not_divergent(body.span));
                 self.return_ty = Some(output_ty);
                 let body_ty = self.infer_expr(body).ty;
-                let output_ty = self.return_ty.take().expect("return type should be present");
+                let output_ty = self
+                    .return_ty
+                    .take()
+                    .expect("return type should be present");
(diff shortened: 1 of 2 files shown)


compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -328,11 +328,22 @@ impl<'a> Context<'a> {
...
+                let output_ty = self.inferrer.fresh_ty(TySource::not_divergent(body.span));
+                self.return_ty = Some(output_ty);
+                let body_ty = self.infer_expr(body).ty;
+                let output_ty = self
+                    .return_ty
+                    .take()
+                    .expect("return type should be present");
+                self.return_ty = prev_ret_ty;
+                if body_ty != Ty::UNIT {
+                    self.inferrer.eq(body.span, output_ty.clone(), body_ty);
```
billti, line 341: Why is if block needed? What does it do exactly? (Maybe add a comment for the noobs like me).

billti (approved):

compiler/qsc_frontend/src/typeck/rules.rs
swernli: Added clarifying comment

PR #653, opened 2023-09-06
Refactor how custom types are defined for wasm boundary

minestarks:
I noticed it was getting to be a chore every time we had to add a custom data type to be serialized across the WASM boundary. Before we add even more types, I wanted to clean them up a bit and standardize how we declare these types. I also realized there were a few spots we weren't taking advantage of `wasm-bindgen`'s ability to work with native Rust types and we were converting to/from `JsValue` manually, which isn't necessary. There were some other excessive conversions/copies that I was able to clean up.


npm/src/debug-service/debug-service.ts
```diff
@@ -164,42 +151,23 @@ export class QSharpDebugService implements IDebugService {
...
     const positionMap = mapUtf8UnitsToUtf16Units(positions, this.code[path]);
     const breakpoint_spans: IBreakpointSpan[] = breakpoint_list.spans.map(
-      (span: IBreakpointSpan) => {
-        const result = {} as IBreakpointSpan;
-        result.id = span.id;
-        result.lo = positionMap[span.lo];
-        result.hi = positionMap[span.hi];
-        return result;
-      }
+      (span) => ({
```
ScottCarda-MS, line 154: Do you need the parens around the braces?

npm/src/language-service/language-service.ts
```diff
@@ -151,6 +163,10 @@ export class QSharpLanguageService implements ILanguageService {
       if (url.protocol === qsharpLibraryUriScheme + ":") {
         code = wasm.get_library_source_content(url.pathname);
       }
+      if (!code) {
+        log.error(`expected ${url} to be in the library`);
+        return undefined;
+      }
```
ScottCarda-MS, line 169: Seems like this should be inside of the previous `if` as it is specific to that branch.

wasm/src/debug_service.rs
```diff
@@ -267,137 +264,130 @@ impl From<StepResultId> for usize {
...
-export interface IBreakpointSpanList {
-    spans: Array<IBreakpointSpan>
-}
-"#;
-
-#[derive(Serialize, Deserialize)]
-struct BreakpointSpanList {
-    pub spans: Vec<BreakpointSpan>,
+serializable_type! {
+    pub struct StructStepResult {
```
ScottCarda-MS, line 268: `StructStepResult` seems to be the only `pub struct`, while the others are just `struct`. Is this on purpose?

npm/src/debug-service/debug-service.ts
minestarks: Yes! This is kind of an annoying JavaScript syntax ambiguity. `() => {}` and `() => ({})` mean different things - in the first one, `{}` is parsed as a block, so the lambda returns `void`. In the second one `{}` is parsed as an object declaration, so the lambda returns an `object`.

Worse, `() => { foo: 3 }` is valid code, but instead of returning an object like you would expect, it's actually declaring a code block with the labeled statement `foo` and a statement with just `3` in it, which is a valid expression therefore a valid statement. Then the block returns `void`. JavaScript will happily execute this, but at least TypeScript will complain that the label is unused. It's very common to write code like this and then stare at it wondering what TypeScript is complaining about :D

wasm/src/debug_service.rs
minestarks: Good call out. I was in the middle of trying to figure out whether we actually needed `pub`, then I forgot. I think they can all be made private.

ScottCarda-MS (approved):

wasm/src/debug_service.rs
ScottCarda-MS: I see now that a couple of the types under wasm/src/language_service.rs also have the `pub`.

swernli (approved):
I really like how this macro helps us clean up the Rust code!

npm/src/language-service/language-service.ts
```diff
@@ -97,11 +97,15 @@ export class QSharpLanguageService implements ILanguageService {
     offset: number
   ): Promise<ICompletionList> {
     const code = this.code[documentUri];
+    if (!code) {
```
billti, line 100: Due to "" being 'falsy', this will be false here if the document is an empty string.

npm/src/language-service/language-service.ts
```diff
@@ -115,13 +119,17 @@ export class QSharpLanguageService implements ILanguageService {
     return result;
   }
 
-  async getHover(documentUri: string, offset: number): Promise<IHover | null> {
+  async getHover(
+    documentUri: string,
+    offset: number
+  ): Promise<IHover | undefined> {
     const code = this.code[documentUri];
+    if (!code) {
```
billti, line 127: Same as above. I expect it's valid for a document to be empty (e.g. a new file, or just deleted all the contents to start again).

npm/src/tsconfig.json
```diff
@@ -2,11 +2,11 @@
   "compilerOptions": {
     "module": "node16",
     "target": "ES2020",
-    "lib": ["DOM", "WebWorker", "ES2020"],
```
billti, line 5: Was this breaking anything, or just clean up? I think it was added as the npm package also targets browser, and we may at some point use a WebView in VS Code for visualizations. Harmless to remove it if not needed now I guess, as long as anybody who needs it later knows where to go put it back :-)

wasm/src/debug_service.rs
```diff
@@ -267,137 +262,130 @@ impl From<StepResultId> for usize {
...
+        pub id: usize,
+        pub value: usize,
+    },
+    r#"export interface IStructStepResult {
+        id: number;
+        value: number;
+    }"#,
+    StructStepResult,
+    IStructStepResult,
+    "IStructStepResult"
```
billti, line 276: It's a little unfortunate you still need to repeat the name 'StructStepResult' 5 times when using the macro. No way to avoid this?

wasm/src/language_service.rs
```diff
@@ -80,147 +81,142 @@ impl LanguageService {
...
+serializable_type! {
+    pub(crate) struct VSDiagnostic {
+        pub start_pos: usize,
+        pub end_pos: usize,
+        pub message: String,
+        pub severity: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub code: Option<VSDiagnosticCode>,
+    },
+    r#"export interface IDiagnostic {
```
billti, line 202: Interesting we lose the 'VS' from VSDiagnostic on this one when exporting to TypeScript. Is that deliberate or for a reason?

billti (approved):
Some nice changes. Could have some subtle issues. If you have high confidence it doesn't regress anything (being that it doesn't 'fix' anything either), and you want to merge this close to Preview release, go ahead :-)

npm/src/tsconfig.json
minestarks: Yes, as soon as I turned off `skipLibCheck` I started getting duplicate definition errors across `DOM` and `WebWorker`. 

I decided to turn off `skipLibCheck` because I wanted to get errors from `qsc_wasm.d.ts` to catch some obvious mistakes I was making in the wasm TypeScript definitions.

I think this is "technically correct" today, since we run in the extension host worker, we don't have access to the DOM? If we added DOM manipulation would we add it in this package, or a different one? I'm content to leave that unanswered for now...

wasm/src/debug_service.rs
minestarks: I thought I couldn't, but now that I'm looking with fresh eyes I think I can eliminate at least one! Revising...

wasm/src/language_service.rs
minestarks: Oh, that must have been a personal choice I made a while back - I really didn't like exposing the name as `IVSDiagnostic` to TypeScript, since it's really not a "VS" diagnostic - it has differences from both [`vscode.Diagnostic`](https://github.com/microsoft/qsharp/blob/76e91f14a9d330a29358b45692e7530a3ccae10d/vscode/src/diagnostics.ts#L4) and [`monaco.Diagnostic`](https://github.com/microsoft/qsharp/blob/76e91f14a9d330a29358b45692e7530a3ccae10d/playground/src/editor.tsx#L26). It's our own internal type.

But I left it as "VS" in Rust code since it differentiates it from `miette::Diagnostic`. And I guess the same type gets redefined as `VSDiagnostic` further up the stack (compare [here](https://github.com/microsoft/qsharp/blob/76e91f14a9d330a29358b45692e7530a3ccae10d/npm/src/vsdiagnostic.ts#L4) and [here](https://github.com/microsoft/qsharp/blob/76e91f14a9d330a29358b45692e7530a3ccae10d/wasm/src/language_service.rs#L196)) anyway. I don't know, it's a bit of a mess right now, but the current PR isn't making it _worse_, I think.

npm/src/language-service/language-service.ts
minestarks: Ooh, good catch. Thank you.

wasm/src/language_service.rs
```diff
@@ -80,147 +81,142 @@ impl LanguageService {
...
+        start_pos: number;
```
ScottCarda-MS, line 203: Some of the fields are snake case while others are camel case. I think the API actually wants camel case, but Rust wants snake case. Could we make them consistently all one way or the other? My preference is to do camel case as that is more true to the API, if it is not too much hassle to get Rust to not complain about it.

minestarks: I would also prefer all camel case, since these names inevitably propagate up to JS code. I actually added a warning suppression _inside_ the macro definition, so we should be able to use camel case freely for any types we define in the future.

I just... didn't want to touch this Diagnostic type for reasons I called out above... it's used in a bunch of places and I'm not super happy about how much we repeat ourselves. It would have made this PR even larger... so let's leave that cleanup for later.

PR #657, opened 2023-09-06
Add ability to set target profile in language service

minestarks:
Introducing a `WorkspaceConfiguration` concept to the language service which provides a central means of setting compiler options, including `TargetProfile` as well as `PackageType`, which we were previously passing in `update_document`.

In a future PR, these settings can be wired up to VS Code settings or to a VS Code command. (Draft PR #663)

I'm keeping compatibility with LSP in mind. The new `update_configuration` method carries the semantics of the [`workspace/didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration) notification in LSP. 

We don't currently support per-document settings. I'm not sure if we would want to, given our current user scenarios (where would we persist such a setting in VS Code?). As we flesh out Q# "projects" and other ways of passing compiler settings, we can work those into the language service as well.

### DIFF
--- a/language_service/src/lib.rs
+++ b/language_service/src/lib.rs
@@ -3,14 +3,12 @@
 
 #![warn(clippy::mod_module_files, clippy::pedantic, clippy::unwrap_used)]
 
-use std::collections::HashMap;
-
-use log::trace;
-use protocol::{CompletionList, Definition, Hover};
-use qsc::PackageType;
-use qsc_utils::Compilation;
-
 use crate::qsc_utils::compile_document;
+use log::trace;
+use protocol::{CompletionList, Definition, Hover, WorkspaceConfigurationUpdate};
+use qsc::{PackageType, TargetProfile};
+use qsc_utils::Compilation;
+use std::collections::HashMap;
 
 pub mod completion;
 pub mod definition;
@@ -20,13 +18,34 @@ pub mod protocol;
 mod qsc_utils;
 #[cfg(test)]
 mod test_utils;
+#[cfg(test)]
+mod tests;
 
 pub struct LanguageService<'a> {
+    /// Workspace configuration can include compiler settings
+    /// that affect error checking and other language server behavior.
+    /// Currently these settings apply to all documents in the
+    /// workspace. Per-document configurations are not supported.
+    configuration: WorkspaceConfiguration,
     /// Associate each known document with a separate compilation.
     document_map: HashMap<String, DocumentState>,
     /// Callback which will receive diagnostics (compilation errors)
     /// whenever a (re-)compilation occurs.
     diagnostics_receiver: Box<DiagnosticsReceiver<'a>>,
+}
+
+struct WorkspaceConfiguration {
+    pub target_profile: TargetProfile,
+    pub package_type: PackageType,
+}
+
+impl Default for WorkspaceConfiguration {
+    fn default() -> Self {
+        Self {
+            target_profile: TargetProfile::Full,
+            package_type: PackageType::Exe,
+        }
+    }
 }
 
 struct DocumentState {
@@ -44,8 +63,23 @@ type DiagnosticsReceiver<'a> = dyn FnMut(&str, u32, &[qsc::compile::Error]) + 'a
 impl<'a> LanguageService<'a> {
     pub fn new(diagnostics_receiver: impl FnMut(&str, u32, &[qsc::compile::Error]) + 'a) -> Self {
         LanguageService {
+            configuration: WorkspaceConfiguration::default(),
             document_map: HashMap::new(),
             diagnostics_receiver: Box::new(diagnostics_receiver),
+        }
+    }
+
+    /// Updates the workspace configuration. If any compiler settings are updated,
+    /// a recompilation may be triggered, which will results in a new set of diagnostics
+    /// being published.
+    pub fn update_configuration(&mut self, configuration: &WorkspaceConfigurationUpdate) {
+        trace!("update_configuration: {configuration:?}");
+
+        let need_recompile = self.apply_configuration(configuration);
+
+        // Some configuration options require a recompilation as they impact error checking
+        if need_recompile {
+            self.recompile_all();
         }
     }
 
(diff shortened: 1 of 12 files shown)


swernli (approved):
Rust updates look good!

language_service/src/lib.rs
```diff
@@ -145,4 +179,50 @@ impl<'a> LanguageService<'a> {
...
+                    .sources
+                    .find_by_name(uri)
+                    .expect("source should be found")
+                    .contents
+                    .clone(),
+            ));
+        }
+
+        // Now recompile everything with current settings
+        for source in sources {
```
billti, line 219: Are you not able to call update_document as you iterate over the document_map above? The interim sources vector seems unnecessary.

Also, does this also recompile the std lib for use with the new settings? With the `@Config` annotations there it behaves a little differently for base or full profile.

language_service/src/tests.rs
```diff
@@ -0,0 +1,213 @@
...
+                        ),
+                    ),
+                ),
+            ],
+        ),
+    ]
+"#]],
+    );
+
+    ls.update_configuration(&WorkspaceConfigurationUpdate {
```
billti, line 182: Per other comment, it would be a good test to use a stdlib method that only available in Full (e.g. ResultAsBool), then switch to Base and see if you get an error for it.

npm/src/language-service/language-service.ts
```diff
@@ -195,15 +199,16 @@ export class QSharpLanguageService implements ILanguageService {
   onDiagnostics(uri: string, version: number, diagnostics: IDiagnostic[]) {
     try {
       const code = this.code[uri];
-      if (code === undefined) {
-        log.error(`expected ${uri} to be in the document map`);
+      const empty = diagnostics.length === 0;
+      if (code === undefined && !empty) {
```
billti, line 203: Why does this error depend on there being some diagnostics now? A comment to clarify would help.

npm/test/basics.js
```diff
@@ -496,12 +495,49 @@ test("language service diagnostics - web worker", async () => {
...
+  // Above document should have generated a missing entrypoint error
+  assert(gotDiagnostics);
+
+  // Reset expectations
+  gotDiagnostics = false;
+  expectedMessages = [];
+
+  await languageService.updateConfiguration({ packageType: "lib" });
+
+  // Updating the config should cause another diagnostics event clearing the errors
```
billti, line 535: Should you be verifying the expectedMessages is emtpy here?

npm/package.json
```diff
@@ -25,7 +25,8 @@
     "build": "npm run generate && npm run build:tsc",
     "generate": "node generate_katas_content.js && node generate_samples_content.js",
     "build:tsc": "node ../node_modules/typescript/bin/tsc -p ./src/tsconfig.json",
-    "tsc:watch": "node ../node_modules/typescript/bin/tsc -p ./src/tsconfig.json --watch"
+    "tsc:watch": "node ../node_modules/typescript/bin/tsc -p ./src/tsconfig.json --watch",
+    "test": "node --test"
```
billti, line 29: Is typing `npm run test` easier than `node --test`? :-)

wasm/src/language_service.rs
```diff
@@ -33,17 +33,25 @@ impl LanguageService {
...
+            .update_configuration(&qsls::protocol::WorkspaceConfigurationUpdate {
+                target_profile: config.targetProfile.map(|s| match s.as_str() {
+                    "base" => qsc::TargetProfile::Base,
+                    "full" => qsc::TargetProfile::Full,
+                    _ => panic!("invalid target profile"),
+                }),
+                package_type: config.packageType.map(|s| match s.as_str() {
+                    "lib" => qsc::PackageType::Lib,
+                    "exe" => qsc::PackageType::Exe,
+                    _ => panic!("invalid package type"),
```
billti, line 48: Being the string is caller provided data, is a panic on an unexpected string the right behavior here? Or would returning an Err be more appropriate (which wasm turns into an exception on the JavaScript side).

wasm/src/language_service.rs
```diff
@@ -33,17 +33,25 @@ impl LanguageService {
...
-            } else {
-                qsc::PackageType::Lib
-            },
-        );
+    pub fn update_configuration(&mut self, config: IWorkspaceConfiguration) {
+        let config: WorkspaceConfiguration = config.into();
+        self.0
+            .update_configuration(&qsls::protocol::WorkspaceConfigurationUpdate {
+                target_profile: config.targetProfile.map(|s| match s.as_str() {
+                    "base" => qsc::TargetProfile::Base,
```
billti, line 41: Being in the struct both TargetProfile and PackageType are optional, does this handle the case where those fields might be None/undefined?

language_service/src/lib.rs
minestarks: Rust won't let me call `update_document` while holding another reference to the `document_map`, but I suppose I could refactor to use `iter_mut()` instead (and basically reuse the contents of the method `update_method` instead of calling it). There was some annoying code duplication when I did that, but I can give it another try.

Yes stdlib gets compiled here as well!

npm/test/basics.js
minestarks: It's done in the callback above. (I always find it a challenge to test callbacks, if you have any refactoring tricks I'd welcome them).

npm/package.json
minestarks: lol, it's actually `npm test` so I really did find it easier!! If there was a reason it was omitted, I can take it out, but I figured it was harmless.

wasm/src/language_service.rs
minestarks: Hmmm good question. The TypeScript type of `IWorkspaceConfiguration.packageType` is defined as `"lib" | "exe"`, so the correct value will be "enforced" by the .d.ts. And the wasm library is only ever meant to be used from our own npm package. In the VS Code extension, I sanitize the user-provided setting before passing it into this method. So I feel like the developer would be _really_ trying to shoot themselves in the foot if a wrong value slips in, and I feel that the panic suits the fail-fast approach I tend to prefer.

I like it this way now, but I can be talked out of it if it's important.

wasm/src/language_service.rs
minestarks: Yep! The underlying language service is always initialized with the defaults (see the very top of this PR.). If `update_configuration` is never called, or any of the fields are null, those values are simply left alone.

billti (approved):

language_service/src/tests.rs
minestarks: omg. So glad you asked me to add a test - I wasn't even passing in the `target_profile` to the stdlib compilation. Did you already notice that and were just subtly hinting at it??

PR #659, opened 2023-09-07
Recover from missing semicolons

minestarks:
This is to lay the groundwork for signature help.

We should be able to parse just enough to know that the below is a function call, even before the user has typed the trailing semicolon. This way signature help can provide accurate parameter info when invoked at the cursor shown below.

```qsharp
namespace MyQuantumProgram {
    open Microsoft.Quantum.Intrinsic;

    operation Foo(x: Int, y: Double, z: String) : Unit {}

    @EntryPoint()
    operation Main() : Unit {
        Foo(↘️)
        let x = 3;
    }
}

```

cc @ScottCarda-MS

### DIFF
--- a/compiler/qsc_parse/src/prim.rs
+++ b/compiler/qsc_parse/src/prim.rs
@@ -203,6 +203,17 @@ pub(super) fn recovering<T>(
     }
 }
 
+pub(super) fn recovering_semi(s: &mut Scanner) -> Result<()> {
+    match token(s, TokenKind::Semi) {
+        Ok(()) => Ok(()),
+        Err(error) => {
+            s.push_error(error);
+            // no recovery, just move on to the next token
+            Ok(())
+        }
+    }
+}
+
 pub(super) fn recovering_token(s: &mut Scanner, t: TokenKind) -> Result<()> {
     match token(s, t) {
         Ok(()) => Ok(()),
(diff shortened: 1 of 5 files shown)


compiler/qsc_parse/src/prim.rs
```diff
@@ -203,6 +203,17 @@ pub(super) fn recovering<T>(
     }
 }
 
+pub(super) fn recovering_semi(s: &mut Scanner) -> Result<()> {
```
swernli, line 206: nitpick: I wonder if there is a better name for this, because on the one hand it is effectively recovering from a failure to find a token, but on the other hand it does not call `s.recover` so it doesn't actually perform the recovery action. `expect_semi` is not quite right, since that sounds like `expect` in Rust which panics. Maybe something like `check_semi`?

swernli: There is already a `check_semis` in stmt.rs, and this function is only called from stmt.rs, so maybe it could move there to be next to each other?

minestarks: Yeah naming was tricky. Its semantics are the same as `recovering<T>` and `recovering_token`, in that it attempts to parse an X, and if it can't, it appends an error, and leaves the scanner in a state where we can continue parsing. I agree that it doesn't perform "recovery" per se by eating tokens, but that seems like an implementation detail?

I think `check_semi` sounds too similar to the existing `check_semis` which has totally different behavior. And it does more than checking - it actually parses a semicolon.

I thought about `opt_semi`, but that would evoke the existing `opt()` function which has different semantics as well (doesn't push errors).

I'm not sure, right now my choice is still my favorite but I'm happy to riff on it more.

minestarks: I really think this belongs more next to the other two `recovering*` functions... it's really just a special case of `recovering_token`.

PR #660, opened 2023-09-07
Publish artifacts via ADO pipeline

billti:
This is a variation on Ian's #641 that deals with naming and versioning a little differently.

Still some minor follow up TODOs, but want to get this in now.


vscode/src/extension.ts
```diff
@@ -8,7 +8,7 @@ import {
   log,
   getLibrarySourceContent,
   qsharpLibraryUriScheme,
-} from "qsharp";
+} from "qsharp-lang";
```
minestarks, line 11: So no package aliasing? We're just committing to the new name?

vscode/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
...
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE
```
minestarks, line 21: This seems to be missing a period at the end.

vscode/.artifactignore
minestarks, line 1: Same question as https://github.com/microsoft/qsharp/pull/641/files#r1313293998 , but now to you :)

> What is a .artifactignore file for? (To go along with the .gitignore and .vscodeignore already in this dir)? Can you add a comment in the file as to what purpose it serves? Thanks.

billti: https://learn.microsoft.com/en-us/azure/devops/artifacts/reference/artifactignore?view=azure-devops

vscode/LICENSE.txt
billti: It was a file copy from the one in the root dir (which is identical), so I won't worry about it in this PR.

vscode/src/extension.ts
billti: Yep. Being it will be 'qsharp-lang' on npm, I think aliases would just be confusing. We should use it as anyone installing from npm will.

vscode/.artifactignore
minestarks: I saw that, but I was scratching my head wondering why we'd want to exclude the one thing we want to publish. I _just_ noticed the `!`.

build_version.py
```diff
@@ -0,0 +1,113 @@
...
+from datetime import datetime, timezone
+
+# To be updated every time we start a new major.minor version.
+major_minor = "1.0"
+
+# Default to 'dev' builds with a YYMMDD build number
+BUILD_TYPE = os.environ.get("BUILD_TYPE") or "dev"
+BUILD_NUMBER = os.environ.get("BUILD_NUMBER") or "YYMMDD"
+
+build_ver = datetime.now(timezone.utc).strftime("%y%m%d")
```
swernli, line 16: This does mean we can't publish more than once in a day, which we should hopefully never need, but if we do, what would be the approach?

swernli (approved):

jupyterlab/README.md
```diff
@@ -46,11 +46,11 @@ Then refresh the browser.
 Remove the `pip` package by running:
 
 ```bash
-pip uninstall qsharp_jupyterlab
+pip uninstall qsharp-lang-jupyterlab
```
minestarks, line 49: A few files above refer to the package as `qsharp-jupyterlab`.  Which one is it?

minestarks (approved):

build_version.py
billti: When you run the pipeline, you can specify a build number for the patch part (in fact, you have to for non dev builds).

jupyterlab/README.md
billti: I was hoping you could confirm all that! I think the issue is that the package name is qsharp-lang-jupyterlab, but the dir we build into is qsharp-jupyterlab. Might want to clean that up later.

minestarks: Oh I can take a closer look then. (I also prefer the package name `qsharp-jupyterlab` but it's OK if that case is closed now.)

build_version.py
minestarks, line 1: Sorry last minute request! I just noticed this makes tab-completion really annoying when I type `./build` at the command line, is it possible to name it to `version.py` or something?

jupyterlab/README.md
billti: I think we can decide to change the package name in a follow up. Agree that `qsharp-lang-jupyterlab` is a bit of a mouthful.

build_version.py
billti: I noticed that too. Renamed.

PR #663, opened 2023-09-07
Add target profile setting in VS Code

minestarks:
- A new configuration setting for target profile.
- Status bar item to show this setting.
- Command to set this setting.
- Language service update to consume this setting.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -59,6 +59,12 @@
         "category": "Debug",
         "enablement": "!inDebugMode",
         "icon": "$(play)"
+      },
+      {
+        "command": "quantum-set-target",
+        "category": "Q#",
+        "title": "Set the Azure Quantum QIR target profile",
+        "enablement": "resourceLangId == qsharp"
       }
     ],
     "breakpoints": [
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -125,7 +125,25 @@
           }
         }
       }
-    ]
+    ],
+    "configuration": {
+      "title": "Q#",
+      "properties": {
+        "qsharp.targetProfile": {
+          "type": "string",
+          "default": "full",
+          "enum": [
+            "full",
+            "base"
+          ],
+          "enumDescriptions": [
+            "The full set of capabilities required to run any Q# program. This option maps to the Full Profile as defined by the QIR specification.",
+            "The minimal set of capabilities required to run a quantum program. This option maps to the Base Profile as defined by the QIR specification."
+          ],
+          "description": "Setting the target profile allows the Q# extension to generate programs that are compatible with a specific target. The target is the hardware or simulator which will be used to run the Q# program. The target profile is a description of a target's capabilities."
+        }
+      }
+    }
   },
   "scripts": {
     "tsc:check": "node ../node_modules/typescript/bin/tsc -p ./tsconfig.json",
--- a/vscode/src/common.ts
+++ b/vscode/src/common.ts
@@ -1,18 +1,29 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { DocumentFilter, TextDocument, Uri } from "vscode";
+import { DocumentFilter, TextDocument, Uri, languages } from "vscode";
 
 export const qsharpLanguageId = "qsharp";
+
 // Matches all Q# documents, including unsaved files, notebook cells, etc.
 export const qsharpDocumentFilter: DocumentFilter = {
   language: qsharpLanguageId,
 };
+
 // Matches only Q# notebook cell documents.
 export const qsharpNotebookCellDocumentFilter: DocumentFilter = {
   language: qsharpLanguageId,
   notebookType: "jupyter-notebook",
 };
+
+export function isQsharpDocument(document: TextDocument): boolean {
+  return languages.match(qsharpDocumentFilter, document) > 0;
+}
+
+export function isQsharpNotebookCell(document: TextDocument) {
+  return languages.match(qsharpNotebookCellDocumentFilter, document);
+}
+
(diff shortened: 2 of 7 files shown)


vscode/package.json
```diff
@@ -125,7 +131,25 @@
           }
         }
       }
-    ]
+    ],
+    "configuration": {
+      "title": "Q#",
```
minestarks, line 136: Reviewer: If you're going to nitpick on the documentation here, please also nitpick on the documentation for the Python library as it's lifted directly from there.

https://github.com/microsoft/qsharp/blob/f2529e17d5ff21b189e846d456702ea077e989b8/pip/qsharp/_native.pyi#L7

vscode/src/common.ts
```diff
@@ -1,18 +1,29 @@
...
   language: qsharpLanguageId,
   notebookType: "jupyter-notebook",
 };
+
+export function isQsharpDocument(document: TextDocument): boolean {
+  return languages.match(qsharpDocumentFilter, document) > 0;
+}
+
+export function isQsharpNotebookCell(document: TextDocument) {
+  return languages.match(qsharpNotebookCellDocumentFilter, document);
```
billti, line 24: Why doesn't think return a bool if result if > 0 like the method above?

vscode/src/extension.ts
```diff
@@ -124,21 +101,15 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
...
       }
     })
   );
 
   function updateIfQsharpDocument(document: vscode.TextDocument) {
-    if (
-      vscode.languages.match(qsharpDocumentFilter, document) &&
-      !vscode.languages.match(qsharpNotebookCellDocumentFilter, document)
-    ) {
+    if (isQsharpDocument(document) && !isQsharpNotebookCell(document)) {
```
billti, line 112: It is treated more like a bool here (per other comment)

vscode/src/extension.ts
```diff
@@ -151,16 +122,81 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
...
+        .getConfiguration("qsharp")
+        .get<string>("targetProfile");
+      log.info(`Target profile changed: ${targetProfile}`);
+
+      switch (targetProfile) {
+        case "base":
+        case "full":
+          break;
+        default:
+          log.warn(`Invalid value for target profile: ${targetProfile}`);
```
billti, line 189: They may well have removed the setting, which is totally valid and shouldn't be a warning, so may want to not warn on 'undefined'

vscode/src/statusbar.ts
```diff
@@ -0,0 +1,116 @@
...
+    refreshStatusBarItemValue();
+  }
+
+  function refreshStatusBarItemValue() {
+    // The target profile setting is a "window" scoped setting,
+    // meaning it can't be set on a per-folder basis. So we don't
+    // need to pass a specific scope here to retrieve the document
+    // value - we just use the workspace setting.
+    // VS Code will return the default value defined by the extension
+    // if none was set by the user, so targetProfile should always
```
billti, line 57: This is true if you have the configuration in the package.json and thus it contributes to the VS Code settings UX. I wonder if we need that UX though, being we have a command and status bar item to change it with anyway. Once Azure is on by default, we'd have no other settings in the UX (yet).

So maybe remove it from package.json for a cleaner UX? (And thus default value would be undefined).

vscode/package.json
```diff
@@ -125,7 +131,25 @@
...
+      "properties": {
+        "qsharp.targetProfile": {
```
billti, line 138: I ended up using settings of the pattern "Q#.targetProfile", as the initial string shows up in the UX, and customers know the language as 'Q#', not 'Qsharp' (e.g. see below)

image

billti (approved):
Couple little things, else looks great. Thanks!

vscode/src/statusbar.ts
minestarks: I'm not sure where we would persist it then. Can we just persist arbitrary configurations not defined in package.json? I wasn't sure of that.

vscode/package.json
minestarks: Ohh, I was wondering where that string was coming from. Thanks for pointing this out.

vscode/src/statusbar.ts
billti: Actually, I've been testing, and if you remove the 'configuration' contribution from package.json, while you can still read any arbitrary setting, you can't actually write one that's not defined!! So we need to leave this in package.json.

I renamed to "Q#.targetProfile" and integrated with my branch. If you check out my azure PR, this with changes is in there and works great!

billti:
Just noticed a couple bugs in testing, the first to do with this PR, the second a more general one:

1) It doesn't set the profile on starting the language service, only when you change it. So if I have base profile set in my settings already, it doesn't take effect. I added this commit to my branch to fix it <https://github.com/microsoft/qsharp/pull/621/commits/0077b387b24720c93058112a24fe2c39a9eb9294>

2) In general, if you launch VS Code with a .qs file already loaded, no errors show up until you make an edit - be they profile errors, syntax errors, etc. Looks like we need to force a diagnostic pass if there are documents already open on launch. I can start looking into that too.

billti (changes requested):
See latest feedback on issues. (Some already fixed in my branch).

billti:
OK. The 2nd issue was pretty easy to fix (I think). Just needed to start listening for diagnostics earlier (before it adds the open diagnostics). Take a look at the below commit in my Azure branch and see if that looks right.

https://github.com/microsoft/qsharp/pull/621/commits/335d917c0e8a292dda9922bcfeccce0a679101fe

billti:
This work was merged in as part of #621 . Thanks!

PR #664, opened 2023-09-08
Language Server: Basic Signature Help

ScottCarda-MS:
Adds basic signature help to the language server.


wasm/src/language_service.rs
```diff
@@ -178,6 +208,47 @@ serializable_type! {
...
+
+serializable_type! {
+    ParameterInformation,
+    {
+        label: Span,
+        documentation: Option<String>,
+    },
+    r#"export interface IParameterInformation {
+        label: { start: number; end: number };
+        documentation: string | undefined;
```
minestarks, line 248: You can use `ISpan` here which is now defined in this file. Also `?` for an optional parameter looks more consistent in style (though it's basically the same as `string | undefined`)

```suggestion
        label: ISpan;
        documentation?: string;
```

wasm/src/language_service.rs
```diff
@@ -178,6 +208,47 @@ serializable_type! {
...
+serializable_type! {
+    SignatureInformation,
+    {
+        label: String,
+        documentation: Option<String>,
+        parameters: Vec<ParameterInformation>,
+    },
+    r#"export interface ISignatureInformation {
+        label: string;
+        documentation: string | undefined;
```
minestarks, line 235: For consistency 

```suggestion
        documentation?: string;
```

wasm/src/language_service.rs
```diff
@@ -178,6 +208,47 @@ serializable_type! {
     IDefinition
 }
 
+serializable_type! {
+    SignatureHelp,
+    {
+        signatures: Vec<SignatureInformation>,
+        active_signature: u32,
+        active_parameter: u32,
```
minestarks, line 216: We should be able to use camelCase now without clippy yelling at us. (Another style consistency thing for this file)

vscode/src/extension.ts
```diff
@@ -65,6 +66,16 @@ export async function activate(context: vscode.ExtensionContext) {
     )
   );
 
+  // signature help
+  context.subscriptions.push(
+    vscode.languages.registerSignatureHelpProvider(
+      "qsharp",
```
minestarks, line 72: Could you use `qsharpDocumentFilter` like we do for hover above? (And if you could fix go-to-def as well that would be cherry on top). 

(No practical difference today, but I have plans for this later when I work on notebook cells. Would be nice to have everything consistent now).

language_service/src/signature_help/tests.rs
```diff
@@ -0,0 +1,623 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::get_signature_help;
+use crate::test_utils::{compile_with_fake_stdlib, get_source_and_marker_offsets};
+use expect_test::{expect, Expect};
+use indoc::indoc;
+
+/// Asserts that the hover text at the given cursor position matches the expected hover text.
```
minestarks, line 9: Comment seems copy-pasted from hover, update the wording?

language_service/src/signature_help.rs
```diff
@@ -0,0 +1,156 @@
...
+        match &decl.input.kind {
+            hir::PatKind::Discard | hir::PatKind::Bind(_) => {
+                vec![self.make_params_with_offset(offset, &decl.input)]
+            }
+            hir::PatKind::Tuple(items) => {
+                let mut cumulative_offset = offset;
+                items
+                    .iter()
+                    .map(|item| {
+                        let info = self.make_params_with_offset(cumulative_offset, item);
```
minestarks, line 119: Extreme nitpicking here, but I feel like this should be called `make_param_with_offset`, singular.

language_service/src/signature_help.rs
```diff
@@ -0,0 +1,156 @@
...
+fn unwrap_parens(expr: &ast::Expr) -> &ast::Expr {
+    match &*expr.kind {
+        ast::ExprKind::Paren(inner) => unwrap_parens(inner),
+        _ => expr,
+    }
+}
+
+impl SignatureHelpFinder<'_> {
+    fn get_params(&self, decl: &hir::CallableDecl) -> Vec<ParameterInformation> {
+        let offset = self.display.get_param_offset(decl);
```
minestarks, line 108: Actually it'd be really helpful if you could add a comment on top that describes what we're doing with a code example. Such as:

```
// operation Foo(bar: Int, baz: Int)
//               ^param 1 start
//                         ^parameter 2 start
```

Feel free to modify my suggestion per your taste. I find that my eyes kind of glaze over all the Rust code but a visual example really gets across what the function is trying to accomplish.

minestarks (approved):
Looks great, left a bunch of style suggestions.

language_service/src/signature_help.rs
ScottCarda-MS: I left a similar example as a comment along with a small description of what the function is doing.

language_service/src/signature_help/tests.rs
ScottCarda-MS: Oh good catch! Fixed

language_service/src/signature_help.rs
ScottCarda-MS: done

vscode/src/extension.ts
ScottCarda-MS: done

wasm/src/language_service.rs
ScottCarda-MS: Thank goodness!

wasm/src/language_service.rs
ScottCarda-MS: done

wasm/src/language_service.rs
ScottCarda-MS: done

swernli (approved):

PR #666, opened 2023-09-08
Use Node 18 for GitHub workflows

minestarks:
Node.js seems to have fixed a bunch of test runner issues v18.

Bumping the prerequisite to v18.

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "license": "MIT",
   "engines": {
-    "node": ">=16.17.0"
+    "node": ">=18.17.1"
   },
   "repository": {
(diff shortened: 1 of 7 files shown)


npm/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "license": "MIT",
   "engines": {
-    "node": ">=16.17.0"
+    "node": ">=18.17.1"
```
billti, line 7: I don't think we should bump this here. This restricts anyone who uses our package from npm to require them to be on 18.17.1 or later.

minestarks: OK, will revert

billti (approved):

PR #674, opened 2023-09-12
Better target filtering and command handling

billti:
Improvements to the tree view in terms of what we do/don't show, how this is determined, and when commands to submit jobs or download results are available.

Built on top of the pending Azure integration branch.


swernli:
Looks like this is out of date since the other PR created a merge commit. This needs a rebase and force push.

Issue #149, opened 2023-04-12
Improve WASM layer error handling


swernli:
The `qsc_wasm` layer does not fully handle the errors reported by `check_internal`, as some information gets lost in translation (secondary labels, help text, etc). In addition, `run_internal` API currently only supports returning a single error rather than a list, which might not work for errors from passes like extracting entry point. We need to update the error handling to support multiple, detailed errors, and work on the compilation/execution context caching semantics to better provide actionable errors to the user via WASM.

### Tasks
+ - [x] Include additional info from compiler errors (#194)
+ - [ ] Cleanup object vs json diagnostic return, with consistent control character escaping

minestarks:
[...]

PR #675, opened 2023-09-12
Handle runtime error spans from different packages

### LINKED ISSUE #654: Interpreter panic when runtime error contains span from stdlib
In a notebook or `qsi`:

```qsharp
use q = Qubit(); CNOT(q,q)
```

```
Error: qubits in gate invocation are not unique
Call stack:
    at QIR.Intrinsic.__quantum__qis__cx__body in qir.qs
    at Microsoft.Quantum.Intrinsic.CNOT in intrinsic.qs

error: Qsc.Eval.QubitUniqueness

  × runtime error
  ╰─▶ qubits in gate invocation are not unique
thread 'main' panicked at 'failed printing to stderr: formatter error', library/std/src/io/stdio.rs:1019:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: std::io::stdio::print_to
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/io/stdio.rs:1019:9
   3: std::io::stdio::_eprint
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/io/stdio.rs:1106:5
   4: qsi::print_interpret_result
             at ./src/bin/qsi.rs:189:17
   5: qsi::repl
             at ./src/bin/qsi.rs:149:13
   6: qsi::main
             at ./src/bin/qsi.rs:125:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

_Originally posted by @minestarks in https://github.com/microsoft/qsharp/issues/596#issuecomment-1681387708_

minestarks:
Fixes #654.

Evaluator errors contain source spans. These source spans sometimes come from a package that is different than the one being currently executed. Previously, to report errors with source code context, we would always use the `SourceMap` associated with the currently executing package. Now, to track which package the source span actually belongs to, we use a new `PackageSpan` type (package ID / span pair) instead of a `Span` type.

When we later try to resolve the source code for reporting, we look up the `PackageId` from the error in the package store to locate the correct `SourceMap`.

I also had to change in the interpreter to make the lookup work: Previously, we weren't actually storing the _current_ package (the "user package" as it's sometimes referred to in the code) in the package store. So the `PackageId` lookup would succeed for std/core libs, but fail for the current package! With these changes, we now insert the `CompileUnit` that backs the current package into the package store. As we append sources to the compilation, we update the `CompileUnit` in the store with the new sources.

### DIFF
--- a/compiler/qsc/src/error.rs
+++ b/compiler/qsc/src/error.rs
@@ -73,20 +73,12 @@ impl<E: Diagnostic> Diagnostic for WithStack<E> {
 pub type Eval = WithStack<WithSource<qsc_eval::Error>>;
 
 pub fn eval(error: qsc_eval::Error, store: &PackageStore, stack_trace: Option<String>) -> Eval {
-    // TODO: handle more than one span at some point...
-    let span = error.spans().next();
+    let span = error.span();
 
-    let error = match span {
-        Some(span) => {
-            let sources = &store
-                .get(map_fir_package_to_hir(span.package))
-                .expect("expected to find package id in store")
-                .sources;
+    let sources = &store
+        .get(map_fir_package_to_hir(span.package))
+        .expect("expected to find package id in store")
+        .sources;
 
-            WithSource::from_map(sources, error)
-        }
-        None => WithSource::no_sources(error),
-    };
-
-    WithStack::new(error, stack_trace)
+    WithStack::new(WithSource::from_map(sources, error), stack_trace)
 }
--- a/compiler/qsc/src/error.rs
+++ b/compiler/qsc/src/error.rs
@@ -2,6 +2,12 @@
 // Licensed under the MIT License.
 
 use miette::Diagnostic;
+use qsc_eval::debug::map_fir_package_to_hir;
+use qsc_frontend::{
+    compile::{PackageStore, SourceMap},
+    error::WithSource,
+};
+use qsc_hir::hir::PackageId;
 use std::{
     error::Error,
     fmt::{self, Debug, Display, Formatter},
@@ -66,4 +72,47 @@ impl<E: Diagnostic> Diagnostic for WithStack<E> {
     fn diagnostic_source(&self) -> Option<&dyn Diagnostic> {
         self.error.diagnostic_source()
     }
+}
+
+pub type Eval = WithStack<WithSource<qsc_eval::Error>>;
+
+pub fn eval(
+    error: qsc_eval::Error,
+    source_map: Option<&SourceMap>,
+    store: &PackageStore,
+    stack_trace: Option<String>,
+) -> Eval {
+    // TODO: handle more than one span at some point...
+    let span = error.spans().next();
+
+    let error = match span {
+        Some(span) => {
+            let package_id = map_fir_package_to_hir(span.package);
+            // TODO: Workaround: the user package in the interpreter is never updated
+            // in the package store with each line. So if the package_id
+            // is not std or core, we assume the package_id is associated with
+            // the user package, but we return the up-to-date source map
+            // instead of the one from the package store.
+            let sources =
+                if package_id == PackageId::CORE || package_id == PackageId::CORE.successor() {
+                    &store
+                        .get(package_id)
+                        .expect("expected to find std or core package")
+                        .sources
+                } else {
+                    source_map.unwrap_or_else(|| {
+                        // stateless interpreter doesn't update the source map, instead it passes it in
+                        &store
+                            .get(package_id)
+                            .expect("expected to find package id in store")
+                            .sources
+                    })
+                };
+
+            WithSource::from_map(sources, error)
+        }
+        None => WithSource::no_sources(error),
+    };
+
+    WithStack::new(error, stack_trace)
 }
--- a/compiler/qsc/src/error.rs
+++ b/compiler/qsc/src/error.rs
@@ -3,11 +3,7 @@
 
 use miette::Diagnostic;
 use qsc_eval::debug::map_fir_package_to_hir;
-use qsc_frontend::{
-    compile::{PackageStore, SourceMap},
-    error::WithSource,
-};
-use qsc_hir::hir::PackageId;
+use qsc_frontend::{compile::PackageStore, error::WithSource};
 use std::{
     error::Error,
     fmt::{self, Debug, Display, Formatter},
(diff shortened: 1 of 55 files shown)


compiler/qsc/src/interpret/stateful.rs
```diff
@@ -60,49 +61,19 @@ impl Error {
...
+    Eval(#[from] WithStack<WithSource<qsc_eval::Error>>),
     #[error("entry point not found")]
     #[diagnostic(code("Qsc.Interpret.NoEntryPoint"))]
     NoEntryPoint,
-}
-
-#[derive(Clone, Debug, Diagnostic, Error)]
-#[diagnostic(transparent)]
-#[error(transparent)]
-pub struct LineError(WithSource<LineErrorKind>);
```
minestarks, line 78: This is just a cosmetic refactoring: `LineError` and `Error` were extremely similar enums that could be combined into one enum. I find this a bit easier to read.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -154,23 +125,27 @@ impl<'a> NodeLookup for Lookup<'a> {
 }
 
 pub struct Interpreter {
```
minestarks, line 127: Another cosmetic change: I rearranged and roughly grouped the fields here because they were getting real hard to follow. In the future, I hope to separate out the "compilation state" group of fields to a different struct that encapsulate the incremental compilation bits.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -361,86 +313,91 @@ impl Interpreter {
...
-                            self.compiler.source_map(),
-                            error.into(),
-                        ))
-                    })
-                    .collect::<Vec<_>>()
-            })?;
+        let label = &self.next_line_label();
+
+        let (items, stmts) =
+            self.compile_incremental(|c, p| c.compile_fragments(p, label, line))?;
```
minestarks, line 319: I extracted a helper function here `compile_incremental` for code repeated in three functions (`interpret_line`, `run`, `qirgen`). In the future I'm hoping that bit of code will become part of the incremental compiler I separate out from this file.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -361,86 +313,91 @@ impl Interpreter {
...
-                            };
-
-                            return Err(vec![LineError(WithSource::from_map(
-                                self.compiler.source_map(),
-                                WithStack::new(error, stack_trace).into(),
-                            ))]);
-                        }
-                    }
-                }
+        // Partition list into items and statements since items are always
```
minestarks, line 388: Previously, we would do `sort_by_key` to process items before statements... I took this opportunity to just partition the list into two different ones since we're always going to process items and statements separately....

compiler/qsc_eval/src/lib.rs
```diff
@@ -39,67 +41,97 @@ use val::GlobalId;
...
 
     #[error("range with step size of zero")]
     #[diagnostic(code("Qsc.Eval.RangeStepZero"))]
-    RangeStepZero(#[label("invalid range")] Span),
+    RangeStepZero(#[label("invalid range")] PackageSpan),
 
     #[error("Qubit{0} released while not in |0⟩ state")]
     #[diagnostic(code("Qsc.Eval.ReleasedQubitNotZero"))]
-    ReleasedQubitNotZero(usize),
+    ReleasedQubitNotZero(usize, #[label("Qubit{0}")] PackageSpan),
```
minestarks, line 95: This is a new "feature": I added this span based on a suggestion from the bug bash. It'd be helpful to show the user _which_ qubit is being released.

wasm/src/lib.rs
```diff
@@ -56,28 +57,13 @@ fn get_qir_internal(code: &str) -> std::result::Result<String, (Error, Vec<Frame
     // Ensure it compiles before trying to add it to the store.
     if !errors.is_empty() {
         // This should never happen, as the program should be checked for errors before trying to
-        // generate code for it. But just in case, simply report the entire source as 'fail to generate output'
-        return Err((
-            Error::OutputFail(Span {
```
minestarks, line 61: This line was added _after_ I made my changes and was causing a build break because I now require `PackageSpan` 😭 . Given that the error is immediately thrown out and converted to a string in the caller, I changed it to be a hardcoded string...

compiler/qsc_eval/src/lib.rs
swernli: It used to show them, and it was removed before because it was just a the location in the stdlib (the id was still there). Now that we have nice stack traces, I agree it makes sense to put this back in. Thanks for doing that!

compiler/qsc/src/error.rs
```diff
@@ -67,3 +68,18 @@ impl<E: Diagnostic> Diagnostic for WithStack<E> {
         self.error.diagnostic_source()
     }
 }
+
+pub fn eval(
```
swernli, line 72: nitpick: the name `eval` makes it feel like this should be evaluating something, but really it's a conversion helper. Perhaps something like `from_eval`?

compiler/qsc/src/interpret/stateful.rs
swernli: Agreed; I think it could make sense to pull that up into either a new `incremental.rs` or the existing `compile.rs` that sits one folder up in qsc/src

wasm/src/lib.rs
swernli: I mentioned this on Bill's PR, but I think we want to refactor some here to make sure that there is common behavior between this code and `run_internal` in terms of how compile errors from the provided code string are handled. There is an existing backlog issue for it in the commont on `run_internal` ([#149](https://github.com/microsoft/qsharp/issues/149))

swernli (approved):

swernli (approved):

compiler/qsc/src/error.rs
minestarks: I really wanted this to be `eval_error`, because, you know, it's creating an eval error, but Clippy shamed me for using the name of the module in the function. 

I really don't get along with Clippy.

billti (approved):

idavis (approved):

idavis (approved):

PR #676, opened 2023-09-12
Better target filtering and command handling

billti:
Show all targets and providers (if not explicitly excluded), and only allow job submission on QIR-supporting targets, and downloads for jobs with results.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -60,6 +60,22 @@
         {
           "command": "qsharp-vscode.runEditorContents",
           "when": "resourceLangId == qsharp"
+        },
+        {
+          "command": "qsharp-vscode.targetSubmit",
+          "when": "qsharp-vscode.treeItemSupportsQir"
+        },
+        {
+          "command": "qsharp-vscode.downloadResults",
+          "when": "qsharp-vscode.treeItemSupportsDownload"
+        },
+        {
+          "command": "qsharp-vscode.getQir",
+          "when": "resourceLangId == qsharp"
+        },
+        {
+          "command": "qsharp-vscode.setTargetProfile",
+          "when": "resourceLangId == qsharp"
         }
       ],
       "view/title": [
@@ -78,12 +94,12 @@
         {
           "command": "qsharp-vscode.targetSubmit",
           "group": "inline",
-          "when": "view == quantum-workspaces && viewItem == target"
+          "when": "view == quantum-workspaces && viewItem == qir-target"
         },
         {
           "command": "qsharp-vscode.downloadResults",
           "group": "inline",
-          "when": "view == quantum-workspaces && viewItem == result"
+          "when": "view == quantum-workspaces && viewItem == result-download"
         }
       ]
     },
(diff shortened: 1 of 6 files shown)


vscode/src/azure/commands.ts
```diff
@@ -19,19 +19,62 @@ import {
...
+
 export function initAzureWorkspaces() {
   const workspaceTreeProvider = new WorkspaceTreeProvider();
-  vscode.window.createTreeView("quantum-workspaces", {
+  const treeView = vscode.window.createTreeView("quantum-workspaces", {
     treeDataProvider: workspaceTreeProvider,
   });
 
+  treeView.onDidChangeSelection((e) => {
+    // Capture the selected item, and set context if if supports job submission or results download.
```
idavis, line 35:
```suggestion
    // Capture the selected item and set context if the supports job submission or results download.
```

vscode/src/azure/commands.ts
```diff
@@ -19,19 +19,62 @@ import {
...
+import { targetSupportQir } from "./providerProperties";
 
 const corsDocsUri = "https://github.com/microsoft/qsharp/wiki/Enabling-CORS";
 
+let currentTreeItem: WorkspaceTreeItem | undefined = undefined;
+
 export function initAzureWorkspaces() {
   const workspaceTreeProvider = new WorkspaceTreeProvider();
-  vscode.window.createTreeView("quantum-workspaces", {
+  const treeView = vscode.window.createTreeView("quantum-workspaces", {
```
minestarks, line 30: Didn't catch this in the first go around, but this (and other resources) should probably be added to `context.subscriptions` to be disposed of when the extension unloads.

Same with the return value of `onDidChangeSelection`.

vscode/src/azure/commands.ts
```diff
@@ -19,19 +19,62 @@ import {
...
+      supportsQir
+    );
+    vscode.commands.executeCommand(
+      "setContext",
+      "qsharp-vscode.treeItemSupportsDownload",
+      supportsDownload
+    );
+  });
+
   vscode.commands.registerCommand(
```
minestarks, line 69: `registerCommand` also returns `Disposable`.

vscode/package.json
```diff
@@ -142,14 +158,13 @@
...
-        "title": "Download full Azure Quantum job results",
+        "title": "Download Azure Quantum job results",
         "icon": "$(cloud-download)"
       },
       {
         "command": "qsharp-vscode.setTargetProfile",
         "category": "Q#",
-        "title": "Set the Azure Quantum QIR target profile",
-        "enablement": "resourceLangId == qsharp"
+        "title": "Set the Azure Quantum QIR target profile"
```
minestarks, line 167: Why did we remove the enablement filter on this?

minestarks (approved):

idavis (approved):

vscode/package.json
billti: It's a global setting (saved in your user settings), not document specific. I'm not sure you should have to have a Q# file open to be able to change it from the menu. I could be persuaded either way though.

PR #677, opened 2023-09-12
Save the workspace list across restarts

billti:


vscode/src/azure/commands.ts
```diff
@@ -21,13 +22,27 @@ import { QuantumUris, compileToBitcode } from "./networkRequests";
 import { getQirForActiveWindow } from "../qirGeneration";
 
 const corsDocsUri = "https://github.com/microsoft/qsharp/wiki/Enabling-CORS";
+const workspacesSecret = "qsharp-vscode.workspaces";
 
-export function initAzureWorkspaces() {
+export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
```
idavis, line 27: Should we add the return type now that this is async?

vscode/src/azure/commands.ts
```diff
@@ -21,13 +22,27 @@ import { QuantumUris, compileToBitcode } from "./networkRequests";
...
   vscode.window.createTreeView("quantum-workspaces", {
     treeDataProvider: workspaceTreeProvider,
   });
 
+  // Add any previously saved workspaces
+  const savedWorkspaces = await context.secrets.get(workspacesSecret);
+  if (savedWorkspaces) {
+    log.debug("Loading workspaces: ", savedWorkspaces);
+    const workspaces: WorkspaceConnection[] = JSON.parse(savedWorkspaces);
+    for (const workspace of workspaces) {
```
idavis, line 38: Should we use `Promise.all()`?

vscode/src/azure/commands.ts
```diff
@@ -75,25 +90,57 @@ export function initAzureWorkspaces() {
         return;
       }
 
-      setTimeout(async () => {
-        await queryWorkspace(arg.workspace);
-        workspaceTreeProvider.updateWorkspace(arg.workspace);
-      }, 1000);
+      setTimeout(() => {
+        workspaceTreeProvider.refresh();
+      }, 2000);
```
idavis, line 95: nit: magic numbers should be const vars.

minestarks (approved):

idavis (approved):

vscode/src/azure/commands.ts
billti: It should still be inferred correctly, as it was before.

vscode/src/azure/commands.ts
billti: I actually re-wrote this a couple times after hitting some async hazards, and this was the safest and simplest code. There are some subtle bugs and gotchas when you iterate through a list and kick of async work for each item. (I expect the vast majority of users will only have the one workspace in the tree anyway).

If you have a strong opinion though, I could be convinced.

vscode/src/azure/commands.ts
```diff
@@ -75,25 +90,57 @@
         return;
       }
 
-      setTimeout(async () => {
-        await queryWorkspace(arg.workspace);
-        workspaceTreeProvider.updateWorkspace(arg.workspace);
-      }, 1000);
+      setTimeout(() => {
+        workspaceTreeProvider.refresh();
```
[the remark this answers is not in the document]
billti: It doesn't handle untrusted code.

PR #686, opened 2023-09-12
Ask user for number of shots when submitting job to Azure

### LINKED ISSUE #682: Submission to Azure in VS Code should ask for number of shots
As per the title, when we submit a job to Azure via the extension command, it should prompt for number of shots.

sezna:
Asks user for number of shots when submitting to azure. Defaults to 100 if nothing is specified or if the dialog is closed.

Closes #682

### DIFF
--- a/vscode/src/azure/workspaceActions.ts
+++ b/vscode/src/azure/workspaceActions.ts
@@ -316,6 +316,11 @@ export async function submitJob(
       validateInput: validateShotsInput,
     })) || "100";
 
+  // abort if the user hits <Esc> during shots entry
+  if (numberOfShots === undefined) {
+    return;
+  }
+
   // Get a sasUri for the container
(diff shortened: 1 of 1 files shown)


vscode/src/azure/workspaceActions.ts
```diff
@@ -301,6 +301,21 @@ export async function submitJob(
 
   const jobName = await vscode.window.showInputBox({ prompt: "Job name" });
 
+  // validator for the user-provided number of shots input
+  const validateShotsInput = (input: string) => {
+    const result = parseFloat(input);
```
billti, line 306: Why not use parseInt if you're going to check for that? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

billti (approved):
The parseInt thing, plus exit on 'esc'. Thanks!

PR #688, opened 2023-09-13
Update error handling of floating doc comments

### LINKED ISSUE #652: Floating doc comment consumes to next semicolon or close-curly token as part of recovery
When recovering from failing to parse an item, the parser will consume until it hits EOF or a recovery token, namely semicolon or close-curly-bracket. This can trigger weird looking errors when the doc comment is the only text or last text in the namespace, as it consumes the close-curly for the namespace as if it were the malformed item and then causes namespace parsing to fail at EOF:

image

[Playground link](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACi2NwQrCMBBE7%2FmKgd7NTxQ9CRb9gTVd7EI3CckWEfHfTaPDXuaxM%2BO9x4AraV7Z%2BWZObCbxgWpUjOeddT5g5BqKZJMUO7ktUtGOoBJFacU0IJf0KKSwhQyBIu6MrfIMS79KPIv0gfYc0swHF0m5ZgqM82vaKNqml3%2FL26FpHzuuiXpsTKHlVDma%2B3wBLgLdT74AAAA%3D)

Not sure how fixable this is, but it does create some confusion during editing while a doc comment is being written before an item is added below it.

eureka-cpu:
Closes #652 

Checks if doc comments on an `Item` that fails to parse are non-empty. Pushes an error, `ErrorKind::FloatingDocComment(span)`, and returns an error `Item` so that the parser will continue. This returns only a single, more helpful error message for floating doc comments and doesn't consume the closing curly brace.

Notes: If compilation _shouldn't fail_ because of floating doc comments, there could be a recovery step for the error that also removes the error `Item`.


eureka-cpu:
current output:
```sh
Qsc.Parse.Rule

  × syntax error
  ╰─▶ expected item, found `}`
   ╭─[test.qs:7:1]
 7 │     /// Floating Doc comment
 8 │ }
   · ─
   ╰────

Qsc.Parse.EmptyNamespace

  × syntax error
  ╰─▶ expected namespace item, found EOF
   ╭─[test.qs:7:1]
 7 │     /// Floating Doc comment
 8 │ }
   ╰────
   
```
Ideally it would just be the latter, and point to the doc comment itself.

eureka-cpu:
> 

eureka-cpu:
There are some tests failing, please let me know if the code currently present is acceptable and I can update the test data.

compiler/qsc_parse/src/item.rs
```diff
@@ -44,6 +44,12 @@ pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
     } else if let Some(callable) = opt(s, parse_callable_decl)? {
         Box::new(ItemKind::Callable(callable))
     } else {
+        if !doc.is_empty() {
+            return Err(Error(ErrorKind::FloatingDocComment(Span {
```
minestarks, line 48: The `Span` used here seems irrelevant to the error message - it might be more helpful if we used the doc comment's span instead of the next token's span.

compiler/qsc_parse/src/item.rs
```diff
@@ -127,6 +133,7 @@ fn parse_namespace(s: &mut Scanner) -> Result<Namespace> {
     let name = dot_ident(s)?;
     token(s, TokenKind::Open(Delim::Brace))?;
     let items = barrier(s, &[TokenKind::Close(Delim::Brace)], parse_many)?;
+    check_empty_namespace(s, items.as_ref());
```
minestarks, line 136: I'm not sure I agree that empty namespace should be a parse error, it's perfectly valid syntax.

minestarks:
Really appreciate the contribution! I'm not entirely convinced that this is the right approach to fix this particular error, but I think one of us will have to do more due diligence to be sure. To me, it seems that the correct approach would be to look closer at the error recovery code, as hinted in the original issue description. But it's not like I have a clear fix in mind, so I'm just not sure.

compiler/qsc_parse/src/item.rs
eureka-cpu: Hey @minestarks thanks for clarifying.

The example code in the issue threw me off a bit. Since removing the doc comment also produced an error, I assumed errors for empty namespaces was intentional. Working on a new solution now with your feedback in mind.

compiler/qsc_parse/src/item.rs
eureka-cpu: I'll have to see if there is a way to directly get the span of the doc comment. What I have here _is_ getting the span (at least for the examples I've ran it on), just indirectly.

eureka-cpu:
Reverting this back to a draft for the time being.

compiler/qsc_parse/src/item.rs
eureka-cpu: Update: I was able to get the span directly thanks again

compiler/qsc_parse/src/item.rs
```diff
@@ -44,12 +44,17 @@ pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
     } else if let Some(callable) = opt(s, parse_callable_decl)? {
         Box::new(ItemKind::Callable(callable))
     } else {
-        return Err(Error(ErrorKind::Rule("item", s.peek().kind, s.peek().span)));
+        let p = s.peek();
+        if !doc.is_empty() {
+            s.push_error(Error(ErrorKind::FloatingDocComment(t.span)));
+            return Ok(Box::default());
```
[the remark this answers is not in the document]
eureka-cpu: I agree with all of the above, thanks for the feedback.

Although you're right that `Item::default` isn't explicitly used anywhere, there may be instances of `unwrap_or_default` elsewhere, where it might be returning the default `Item` to prevent consuming tokens unnecessarily. I could be wrong though (have yet to check), and in this case an error recovery item sounds appealing. 

A guard in this case makes sense, since the error and the item are a pair.

compiler/qsc_parse/src/item.rs
```diff
@@ -44,12 +44,17 @@ pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
     } else if let Some(callable) = opt(s, parse_callable_decl)? {
         Box::new(ItemKind::Callable(callable))
     } else {
-        return Err(Error(ErrorKind::Rule("item", s.peek().kind, s.peek().span)));
+        let p = s.peek();
```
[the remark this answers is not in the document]
eureka-cpu: That's not a bad idea, depends on if doc comments without an item should throw an error. I assumed it should based on the tests but let me know and I can explore that path as well.

compiler/qsc_parse/src/item.rs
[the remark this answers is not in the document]
minestarks: We have precedent for using `ItemKind::Err` for creating error nodes, but they follow a different pattern. See the `default` function in this file and how it's used. It should take a valid span (which would fix the `0-0` spans you're getting in the tests). See `default` in `stmt.rs` for a similar pattern for statements. I think you can return `Ok(default(s.span(lo)))` here to follow that pattern.

I kind of agree that creating error recovery nodes should always go together with calling `push_error`.. Currently this is ensured by only ever creating error nodes in `recover` (in `prim.rs`). This will be the only new place we're creating error nodes. I like the idea of extracting a helper method like @sezna suggested but I don't know how we would do it without looking really awkward, so I can take it or leave it.

compiler/qsc_parse/src/item.rs
```diff
@@ -33,7 +33,7 @@ pub enum Fragment {
 }
 
 pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
-    let lo = s.peek().span.lo;
+    let t = s.peek();
```
minestarks, line 36: I think the fact that you're using `t` to create the error span means that you'll only report the error for the first token, not necessarily the whole doc comment.

Try using `s.span(lo)` below instead, to create a span that starts from `lo` (where we started parsing) to the current offset (at the time you create the error node).

Try adding a test case for this and see what the error span covers... It should cover the whole doc comment (both lines) not just the first token. Because the whole thing would be treated as a doc comment, were an item to follow it.

```qsharp
namespace Foo {
  /// doc comment token
  /// another doc comment token
}
```

compiler/qsc_parse/src/item.rs
```diff
@@ -44,12 +44,17 @@ pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
     } else if let Some(callable) = opt(s, parse_callable_decl)? {
         Box::new(ItemKind::Callable(callable))
     } else {
-        return Err(Error(ErrorKind::Rule("item", s.peek().kind, s.peek().span)));
+        let p = s.peek();
+        if !doc.is_empty() {
```
minestarks, line 48: Could you please make this an `else if`? The nested `if` breaks the established pattern here of `if ... {} else if  ... {} else {}` which made for easier reading IMO.

compiler/qsc_parse/src/item.rs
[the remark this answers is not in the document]
swernli: I agree with Mine. `ItemKind::Err` is a good pattern to use here, even if it's the first occurrence of that outside of the recovery code. We knew parser recovery was going to be along tail of improvements, so it's likely this won't be be the last time we need this pattern, and the refactoring Alex recommends would be a good improvement to introduce once we have a second instance to incorporate.

compiler/qsc_parse/src/stmt/tests.rs
```diff
@@ -603,18 +600,19 @@ fn doc_followed_by_non_item() {
         }",
         &expect![[r#"
             Block _id_ [0-72]:
-                Stmt _id_ [14-62]: Err
+                Stmt _id_ [14-39]: Item: Item _id_ [0-0]:
+                    Err
+                Stmt _id_ [52-62]: Local (Immutable):
+                    Pat _id_ [56-57]: Bind:
+                        Ident _id_ [56-57] "x"
+                    Expr _id_ [60-61]: Lit: Int(2)
```
minestarks, line 608: This is kind of a nice side effect of your fix :) . We recover enough to be able to parse the `let` statement now. That should improve both error messages and language server features for imperfect code.

minestarks (changes requested):

compiler/qsc_parse/src/item.rs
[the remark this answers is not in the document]
eureka-cpu: Thanks for all the feedback, I should have some time near the end of the week to update this branch.

compiler/qsc_parse/src/lib.rs
```diff
@@ -51,6 +51,9 @@ enum ErrorKind {
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.Parse.Token"))]
     Token(TokenKind, TokenKind, #[label] Span),
+    #[error("expected item after doc comment")]
```
swernli, line 54: I'm on the fence regarding the error text. While "doc comment" is a very common term of art, I'm tempted to say the abbreviate should be skipped to help readability. However, "documentation comment" looks super weird to my eyes. Perhaps it's fine to leave as is and see if we get any other feedback on it?

compiler/qsc_parse/src/item/tests.rs
```diff
@@ -975,31 +1011,15 @@ fn doc_without_item() {
         }",
         &expect![[r#"
             Namespace _id_ [0-62] (Ident _id_ [10-11] "A"):
-                Item _id_ [26-62]:
+                Item _id_ [0-0]:
```
minestarks, line 1014: This `0-0` span seems off - it should match the span of the doc comment (see my comment above about using `s.span(lo)` for how to achieve that)

compiler/qsc_parse/src/item.rs
```diff
@@ -33,7 +33,7 @@ pub enum Fragment {
 }
 
 pub(super) fn parse(s: &mut Scanner) -> Result<Box<Item>> {
-    let lo = s.peek().span.lo;
+    let t = s.peek();
     let doc = parse_doc(s);
     let attrs = many(s, parse_attr)?;
     let visibility = opt(s, parse_visibility)?;
```
swernli, line 39: I know this is getting outside of the scope of the original fix, but we can hit the same recovery issue of floating doc comments if there is a floating attribute (like `@EntryPoint()`) or floating visibility (like `internal`) that is not followed by an item. Perhaps a way to address all of these would be to update `parse_doc` to return an `Option<String>` (and update the usage below in the success case to use `unwrap_or_else` to treat `None` and an empty string), then have logic that checks how far in processing an item we got before failing to find an actual item. If `visibility.is_some()` then we can indicate that an item is expected to follow visibility keyword, else if `!attrs.is_empty()` then we can indicate that an item is expected after one or more attributes, then finally if `doc.is_some()` indicate that an item is expected after a doc comment.

eureka-cpu: I think this is a reasonable follow up issue. It would be much cleaner to use `Option` there, but it would require changing the `doc` field on types that can be documented. Unless you're okay with just providing an empty string slice, which I personally think looks a bit out of place.

compiler/qsc_parse/src/lib.rs
eureka-cpu: Maybe you'd prefer "docstring"?

compiler/qsc_parse/src/item.rs
eureka-cpu: I sent ahead and updated the function signature of `parse_doc` and used `unwrap_or_default` where applicable. I can open a separate issue and open a follow up PR for this if that's okay?

compiler/qsc_parse/src/item.rs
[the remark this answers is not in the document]
eureka-cpu: Okay, I think it's sitting in a decent spot now. WRT the helper method, would that belong in `prim.rs`?

The current logic I have could be expressed as:
```rust
pub fn recovering_doc(s: &mut Scanner, span: Span) -> Box<Item> {
    let err_item = default(s.span(lo));
    s.push_error(Error(ErrorKind::FloatingDocComment(err_item.span)));
    err_item
}
```
Though it doesn't look much like the other recovering functions since it doesn't make use of the `token` function, nor `recover`. It could just be grouped in `item.rs` with `parse_doc`, since it really would just wrap the logic together, returning `default`.

Otherwise, I believe I've hammered out the rest of the requested changes. Let me know if there's anything else I can do, and thanks again for all the great feedback.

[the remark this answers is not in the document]
minestarks: I like the way you have it now. Perhaps we can keep an eye on it, and if more duplication crops up as we add more error recovery, we'll come up with a better design.

minestarks (approved):
Thanks for the fix!

compiler/qsc_parse/src/lib.rs
swernli: I think leaving it as-is should be fine, I'm probably overthinking it!

compiler/qsc_parse/src/item.rs
swernli: Thanks! Filing another issue for the floating attribute/visibility should be fine, I'd appreciate if you can add that and link to this comment.

swernli (approved):

swernli (approved):
Thanks for the work on this, I think it's in a good state to merge. I'm sure this is just the beginning of the tweaks and improvements to parser recovery we'll tackle in the coming months!

compiler/qsc_parse/src/item.rs
eureka-cpu: Ref #757

PR #691, opened 2023-09-13
Type Annotation Style Consistency

ScottCarda-MS:
Updates LS displays, Katas, Libraries, and Samples to consistently use `<name> : <type>` type annotation style, over `<name>: <type>` style, as per our style guidelines.

### DIFF
--- a/katas/content/random_numbers/weighted_random_bit/verification.qs
+++ b/katas/content/random_numbers/weighted_random_bit/verification.qs
@@ -3,7 +3,7 @@ namespace Kata.Verification {
     open Microsoft.Quantum.Convert;
 
     @EntryPoint()
-    operation CheckSolution(): Bool {
+    operation CheckSolution() : Bool {
         for x in [0.0, 0.25, 0.5, 0.75, 1.0] {
             Message($"Testing generating zero with {x*100.0}% probability...");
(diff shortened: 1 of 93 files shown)


katas/content/random_numbers/weighted_random_bit/verification.qs
```diff
@@ -46,10 +46,10 @@ namespace Kata.Verification {
             }
         } else {
             if zeroCount < goalZeroCount - 4 * nRuns / 100 {
-                Message($"Unexpectedly low number of 0's generated: expected around {x * IntAsDouble(nRuns)} 0's, got {zeroCount} out of {nRuns}");
+                Message($"Unexpectedly low number of 0's generated : expected around {x * IntAsDouble(nRuns)} 0's, got {zeroCount} out of {nRuns}");
```
DmitryVasilevsky, line 49: We even do this in messages?

ScottCarda-MS: I tried to avoid changing messages and comments. Looks like I missed these two.

ScottCarda-MS: should be fixed now

swernli (approved):

minestarks (approved):

PR #710, opened 2023-09-14
Show warning to user if the old QDK extension is installed

### LINKED ISSUE #696: Warn on presence of prior QDK
If the prior QDK is installed, this can conflict and provide an odd experience. We should detect and warn on this, informing the user how to disable one of them. (May also want to do this is they have the 'dev' channel and 'stable' channel of our new QDK installed).

sezna:
Closes #696

### DIFF
--- a/compiler/qsc/src/interpret/stateful.rs
+++ b/compiler/qsc/src/interpret/stateful.rs
@@ -192,7 +192,7 @@ impl Interpreter {
         let user_fir = lowerer.lower_package(&unit.package);
         let user_package = store.insert(unit);
         fir_store.insert(map_hir_package_to_fir(user_package), user_fir);
-        let compiler = Compiler::new(&store, dependencies);
+        let compiler = Compiler::new(&store, dependencies, target);
         Ok(Self {
             store,
             compiler,
--- a/compiler/qsc_frontend/src/compile/preprocess.rs
+++ b/compiler/qsc_frontend/src/compile/preprocess.rs
@@ -3,7 +3,7 @@
 
 use core::str::FromStr;
 use qsc_ast::{
-    ast::{Attr, ExprKind, ItemKind, Namespace},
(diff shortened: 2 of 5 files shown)


vscode/src/extension.ts
```diff
@@ -238,3 +240,17 @@ export class QsTextDocumentContentProvider
...
+  const extension = vscode.extensions.getExtension(
+    "quantum.quantum-devkit-vscode"
+  );
+
+  const warningMessage =
+    "Previous QDK Extension `quantum.quantum-devkit-vscode` found. We recommend uninstalling the prior QDK before using this release.";
+
+  if (extension) {
+    log.warn(warningMessage);
+    vscode.window.showErrorMessage(warningMessage);
```
minestarks, line 254: `showWarningMessage` feels like a more appropriate level for this one.

minestarks: Since we're not `await`ing here, could you add a clarifying comment saying we don't wait for the dialog to be dismissed before continuing? I think it's not 100% clear this is intentional.

minestarks (approved):

vscode/src/extension.ts
```diff
@@ -238,3 +240,18 @@ export class QsTextDocumentContentProvider
...
   }
 }
+
+function checkForOldQdk() {
+  const extension = vscode.extensions.getExtension(
+    "quantum.quantum-devkit-vscode"
+  );
+
+  const warningMessage =
+    "Previous QDK Extension `quantum.quantum-devkit-vscode` found. We recommend uninstalling the prior QDK before using this release.";
```
billti, line 250: Is any user going to know what `quantum.quantum-devkit-vscode` is? Can we add something more friendly like "Two Q# extensions are currently active - the new Azure Quantum Development Kit Preview, and the prior Microsoft Quantum Development Kit. Please disable or remove one of them to avoid conflicts".

Also, we discussed detected if they have both the 'dev' channel and 'stable' channel of the new extension installed and warning on that also.

PR #711, opened 2023-09-14
Show warning when QIR base profile is not set when submitting a job

### LINKED ISSUE #694: Make the need to switch to QIR:Base more obvious
Users are confused when they get the error message on trying to submit to hardware that they need to target QIR base profile.

One potential option, show a model popup (rather than a little info-box in the lower right) that offers to switch to target base profile for them when this is encountered. It should also include a link to the docs/wiki page for more information on what the profiles are and when they are important (e.g. targeting hardware).

minestarks:
Fixes #694

### DIFF
--- a/vscode/src/qirGeneration.ts
+++ b/vscode/src/qirGeneration.ts
@@ -25,15 +25,28 @@ export async function getQirForActiveWindow(): Promise<string> {
     );
   }
 
+  const configuration = vscode.workspace.getConfiguration("Q#");
   // Check that the current target is base profile, and current doc has no errors.
-  const targetProfile = vscode.workspace
-    .getConfiguration("Q#")
-    .get<string>("targetProfile", "full");
+  const targetProfile = configuration.get<string>("targetProfile", "full");
   if (targetProfile !== "base") {
-    throw new QirGenerationError(
-      "Submitting to Azure is only supported when targeting the QIR base profile. " +
-        "Please update the QIR target via the status bar selector or extension settings."
+    const result = await vscode.window.showWarningMessage(
+      "Submitting to Azure is only supported when targeting the QIR base profile.",
+      { modal: true },
+      { title: "Change the QIR target profile and continue", action: "set" },
+      { title: "Cancel", action: "cancel", isCloseAffordance: true }
     );
+    if (result?.action !== "set") {
+      throw new QirGenerationError(
+        "Submitting to Azure is only supported when targeting the QIR base profile. " +
+          "Please update the QIR target via the status bar selector or extension settings."
+      );
+    } else {
+      await configuration.update(
+        "targetProfile",
+        "base",
+        vscode.ConfigurationTarget.Global
+      );
+    }
   }
 
   // Get the diagnostics for the current document.


vscode/src/qirGeneration.ts
```diff
@@ -25,15 +25,28 @@ export async function getQirForActiveWindow(): Promise<string> {
...
   // Check that the current target is base profile, and current doc has no errors.
-  const targetProfile = vscode.workspace
-    .getConfiguration("Q#")
-    .get<string>("targetProfile", "full");
+  const targetProfile = configuration.get<string>("targetProfile", "full");
   if (targetProfile !== "base") {
-    throw new QirGenerationError(
-      "Submitting to Azure is only supported when targeting the QIR base profile. " +
-        "Please update the QIR target via the status bar selector or extension settings."
+    const result = await vscode.window.showWarningMessage(
```
billti, line 32: I explicitly avoided showing UX in this method as it can be called in multiple ways. For example, this is also wired into the command to "Get QIR for current Q# program", which doesn't submit to Azure, so the error message here doesn't make sense.

I think it would be preferable to return an error from this helper and let the top-level command determine how to handle them and display any UX.

billti: (Though interestingly it probably does just display the same "Submitting to Azure..." error today :-/ ).

vscode/src/qirGeneration.ts
```diff
@@ -25,15 +25,28 @@ export async function getQirForActiveWindow(): Promise<string> {
...
+      { title: "Change the QIR target profile and continue", action: "set" },
+      { title: "Cancel", action: "cancel", isCloseAffordance: true }
     );
+    if (result?.action !== "set") {
+      throw new QirGenerationError(
+        "Submitting to Azure is only supported when targeting the QIR base profile. " +
+          "Please update the QIR target via the status bar selector or extension settings."
+      );
+    } else {
+      await configuration.update(
```
billti, line 44: Are we sure this change to configuration will be done, the language service updated, the documents revalidated for base profile, and the diagnostics up to date for Base profile before this 'await' returns and it moves on to check the diagnostics are empty in the following code?

minestarks: Nope, not at all sure :/ I found it very tricky to guarantee this, so I just verified manually that the events seem to get fired in the correct order and we move on. But I can't be sure it will always work. It would be fairly complex to wire up the language service to force a diagnostics "refresh" at this point. 

Worst case, we set the base profile, but still sometimes fail out of this function due to compilation errors further down this function. In that case, the user can retry.

I don't love this, so if I missed a simpler solution please let me know.

minestarks: Relying on editor squigglies to validate the document was kind of a race-y approach to begin with. This works today since our language service is synchronous (it's not even running in a separate worker), but as projects grow, you can imagine a TypeScript-like approach where we batch/delay computing diagnostics for performance reasons. So you can't be 100% sure that the diagnostics are perfectly in sync with the document contents.

vscode/src/qirGeneration.ts
minestarks: Besides the reference to Azure in the message text, is there anything wrong with prompting the user to set the base profile in the "Get QIR" command? This seems like a desirable side effect since we'd be solving the exact same UX problem in the "Get QIR" scenario.

billti: Yeah, I think this can go in as-is for now. Just in general, for library/utility type functions that may get called from different commands, probably best to return an error or exception and let the consuming code decide on the UX experience. I think this is fine for now though.

vscode/src/qirGeneration.ts
billti: We could add a `setTimeout`?... I kid, I kid. OK, let's go with it for now, and figure out any 'cleaner' approach later.

I did initially want to call 'compile' directly and see if there were any errors, but as we removed that from the compiler and it's only on the language service now 🙄 , and I didn't want to create a new language service, this "checking if the doc has errors currently" was kind of a hack anyway.

minestarks: Well, at the end of the day, `getQir` will fail if there are compilation errors. I think that's all we need. You shouldn't need to call the compiler twice. Maybe we could refine the error messaging, but the behavior is correct. 

I do _like_ the squiggly check, as race-y as it is. It's a nice, best-effort check to fail fast in case something looks suspicious. But in general, the language service should just be there to "decorate" the editor, not to interact with codegen or the runtime.

vscode/src/qirGeneration.ts
minestarks: I'm thinking we can refactor out the UX from the "smarts" as we build the integration tests. (Which are happening any day now, I promise...

billti (approved):

PR #712, opened 2023-09-14
Renaming library files to have unique names

### LINKED ISSUE #702: Can't step into core/core.qs
Trying to step into a function in core/core.qs doesn't work because there's a file named core.qs in the current std directory and std/core.qs is the first file in the search path

We can change the core/qir file names in each lib to avoid this or we have to enhance the std lib stepping file lookup to understand the package that the file comes from.

idavis:
Fixes #702


swernli (approved):

swernli (approved):
I like the new solution even better; nice, targeted fix!

PR #716, opened 2023-09-15
Refresh the tree view on new job submission and pending jobs

### LINKED ISSUE #705: Workspace tree view should auto-refresh
Perhaps only when jobs are in the pending/running state to avoid spamming the service. Or could do periodically with a back-off timer to some limit (e.g., fall-back to once every 10 mins over time).

billti:
Fixes #705 

Whenever a new job is submitted, or a workspace is added to the tree that has not-yet-completed jobs, the workspace will refresh in the tree for a period using a back-off latency up to a maximum time period. It will stop refreshing a workspace once the submitted jobs are present and all jobs are in a completed state in the tree.

If the user manually refreshes the tree via the refresh icon, the process outlined above will start again.

### DIFF
--- a/vscode/src/azure/treeRefresher.ts
+++ b/vscode/src/azure/treeRefresher.ts
@@ -73,8 +73,8 @@ export async function refreshUntilJobsAreFinished(
   // Stop any other refreshes for this workspace
   stopTimeout(workspace.id);
 
-  // Initially refresh every 5 seconds, backing off up to 5 minutes, and only
-  // keep refreshing for 1 hour, or until all jobs are completed
+  // Initial refresh at 5 seconds, backing off up to 5 minutes between requests (doubling the latency each time),
+  // and only keep refreshing for up to 1 hour, or until all jobs report as completed
   const iter = refreshRangeIterator(5000, 5 * 60 * 1000, 2, 60 * 60 * 1000);
   const predicate = () => {
     if (!treeProvider.hasWorkspace(workspace.id)) return true;
--- a/vscode/src/azure/treeRefresher.ts
+++ b/vscode/src/azure/treeRefresher.ts
@@ -1,0 +1,103 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { log } from "qsharp-lang";
+import { WorkspaceConnection, WorkspaceTreeProvider } from "./treeView";
+
+// To ensure that for a given workspace, we don't have multiple refreshes running
+const runningTimeouts = new Map<string, any>();
+
+function stopTimeout(workspaceId: string) {
+  if (runningTimeouts.has(workspaceId)) {
+    clearTimeout(runningTimeouts.get(workspaceId));
+    runningTimeouts.delete(workspaceId);
+  }
+}
+
+/**
+ * @param minMs - Initial refresh delay
+ * @param maxMs - Maximum refresh delay
+ * @param stepFactor - Refresh delay multiplier
+ * @param limit - Maximum total time to keep refreshing
+ * @returns A generator that yields the next refresh delay
+ */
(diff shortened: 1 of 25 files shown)


vscode/src/azure/workspaceActions.ts
```diff
@@ -370,9 +369,12 @@ export async function submitJob(
       entryPoint: "ENTRYPOINT__main",
       arguments: [],
       count: parseInt(numberOfShots),
+      shots: parseInt(numberOfShots),
```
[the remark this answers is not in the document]
billti: Yes. Some providers expect 'count', some expect 'shots' :shrug:

vscode/src/azure/treeRefresher.ts
```diff
@@ -0,0 +1,103 @@
...
+  // Stop any other refreshes for this workspace
+  stopTimeout(workspace.id);
+
+  // Initial refresh at 5 seconds, backing off up to 5 minutes between requests (doubling the latency each time),
+  // and only keep refreshing for up to 1 hour, or until all jobs report as completed
+  const iter = refreshRangeIterator(5000, 5 * 60 * 1000, 2, 60 * 60 * 1000);
+  const predicate = () => {
+    if (!treeProvider.hasWorkspace(workspace.id)) return true;
+
+    treeProvider.refreshWorkspace(workspace);
```
minestarks, line 82: Should you `await` this, or do we just catch the results of the refresh "next time"?

vscode/src/azure/treeRefresher.ts
```diff
@@ -0,0 +1,103 @@
...
+      log.debug("Reached time limit for workspace refresh");
+    } else {
+      log.debug("Predicate is true for workspace refresh. Done refreshing");
+    }
+    runningTimeouts.delete(id);
+    if (onDone) onDone();
+  }
+}
+
+export async function refreshUntilJobsAreFinished(
```
minestarks, line 66: Nit: since we never `await` this (as we shouldn't), and this doesn't `await` anything, the `Promise` is kind of worthless here... I think this doesn't need to be `async` and could be called something like `queueRefreshUntilJobsAreFinished` to convey the intent.

vscode/src/azure/treeRefresher.ts
```diff
@@ -0,0 +1,103 @@
...
+) {
+  log.debug("In refreshUntilJobsAreFinished for workspace: ", workspace.id);
+  // Stop any other refreshes for this workspace
+  stopTimeout(workspace.id);
+
+  // Initial refresh at 5 seconds, backing off up to 5 minutes between requests (doubling the latency each time),
+  // and only keep refreshing for up to 1 hour, or until all jobs report as completed
+  const iter = refreshRangeIterator(5000, 5 * 60 * 1000, 2, 60 * 60 * 1000);
+  const predicate = () => {
+    if (!treeProvider.hasWorkspace(workspace.id)) return true;
```
minestarks, line 80: If you removed the last workspace, this will keep iterating until the very end. I'm assuming that's an acceptable edge case.

minestarks (approved):
Seems great, just a couple things stuck out as suspicious.

vscode/src/azure/treeRefresher.ts
billti: I'm not sure I follow. If the workspace is gone (last or not), this will return true, ending the refresh checks for this workspace.

minestarks: Ignore me! My brain just got flipped around reading the logic.

vscode/src/azure/treeRefresher.ts
billti: Good call. Probably need to tidy this up a bit.

PR #726, opened 2023-09-16
Update language for job results

swernli:
This change updates the text document language used when opening downloaded job results to json. This enables syntax highlighting and formatting.

### DIFF
--- a/vscode/src/azure/commands.ts
+++ b/vscode/src/azure/commands.ts
@@ -289,7 +289,7 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
           if (file) {
             const doc = await vscode.workspace.openTextDocument({
               content: file,
-              language: "plaintext",
+              language: "json",
             });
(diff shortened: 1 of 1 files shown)


minestarks (approved):
I think this looks good, but I'll be more comfortable waiting for @billti to weigh in and/or local validation.

vscode/src/azure/commands.ts
```diff
@@ -289,7 +289,7 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
           if (file) {
             const doc = await vscode.workspace.openTextDocument({
               content: file,
-              language: "plaintext",
+              language: "json",
```
swernli, line 292: I believe this is the right value based on strings from the selector UI, but as @minestarks mentioned we will merge once we have more manual validation done.

swernli:
Confirmed after local build that the downloaded results show up as json now:
image

billti (approved):

PR #728, opened 2023-09-17
fixed the build script crashes for python versions < 3.9

deepbrain:
the build.py was crashing with an ambiguous error with python versions < 3.9. I added a check of the python version to display a message that a newer version of python is needed to run the build.py.

### DIFF
--- a/build.py
+++ b/build.py
@@ -14,6 +14,9 @@ import functools
 
 from prereqs import check_prereqs
 
+if sys.version_info < (3, 9):
+    raise SystemError("You need Python 3.9 or later to use this script.")
+
 # Disable buffered output so that the log statements and subprocess output get interleaved in proper order
 print = functools.partial(print, flush=True)
 


billti:
What was the ambiguous error? `build.py` already runs the `prereqs.py` script (at line 66) which checks the Python version is 3.11 or later.

deepbrain:
> What was the ambiguous error? `build.py` already runs the `prereqs.py` script (at line 66) which checks the Python version is 3.11 or later.

Here is the error:

(envpython) (base) XPSHDD:~/qsharp/qsharp$ python ./build.py 
Traceback (most recent call last):
  File "./build.py", line 43, in <module>
    action=argparse.BooleanOptionalAction,
AttributeError: module 'argparse' has no attribute 'BooleanOptionalAction'
(envpython) (base) XPSHDD:~/qsharp/qsharp$ python --version
Python 3.8.10

minestarks:
@deepbrain sorry for the delay on this. The bug is worth fixing (thanks for catching!), but we suggest a slightly different way of handling this:

- Check for 3.11 instead of 3.9 since that's our actual minimum version (see prereqs.py)
- Remove the redundant check in prereqs.py.
- Put the check in prereqs.py, but outside of `check_prereqs()` so that it runs both when running prereqs.py and build.py (but before argument parsing)

If you'd like to make these changes to the PR, I'll go ahead and merge. Thanks!

PR #729, opened 2023-09-17
fixed early termination in the Shor algorithm sample

deepbrain:
fixed early termination in the Shor algorithm sample in case when trivial factors like 1:n are found:
Screenshot from 2023-09-16 20-54-49


swernli (changes requested):
Unfortunately, this seemingly reasonable change has an unintended side effect: the program will now hang if you give it a prime number. The corner case of trivial factors is definitely annoying but a well-known limitation of Shor's algorithm. As annoying as it is, I think it's not unreasonable to expect the user to rerun the program in this case.

DmitryVasilevsky:
The check should probably be combined with `if halfPower != modulus - 1`.

deepbrain:

deepbrain:
> Unfortunately, this seemingly reasonable change has an unintended side effect: the program will now hang if you give it a prime number. The corner case of trivial factors is definitely annoying but a well-known limitation of Shor's algorithm. As annoying as it is, I think it's not unreasonable to expect the user to rerun the program in this case.

Thank you, I moved the check to the "if halfPower != modulus - 1" where now it looks cleaner and less dangerous.

swernli (approved):
While I'm not thrilled about the behavior of using all 100 retries on a prime number, at least it's not going to hang. For large prime numbers this could take a while, but that's the cost of simulating a large quantum program. Otherwise the changes seem reasonable to me!

PR #732, opened 2023-09-18
Minor TypeScript changes to get ready for TS 5.2

minestarks:
I noticed that upgrading to TypeScript 5.2 would introduce a couple of build errors.

We're not upgrading yet, but I thought I'd get these fixes in now to be ready.


playground/src/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
-    "moduleResolution": "nodenext",
+    "moduleResolution": "NodeNext",
```
minestarks, line 3: Fixes: `error TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.`

jupyterlab/tsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
-    "moduleResolution": "nodenext",
+    // To ensure compatibility with jupyterlab, options impacting emit (target,module) are copied from
+    // https://github.com/jupyterlab/jupyterlab/blob/c6744e548c7b2790e4c6a4d41a60aae03f22c62a/tsconfigbase.json
+    "target": "ES2018",
+    "module": "esnext",
+    "moduleResolution": "node",
     "outDir": "lib",
```
minestarks, line 8: Fixes: `error TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.`

Setting `module` to `NodeNext` changed how code was emitted, which made me nervous. I decided instead to follow JupyterLab's settings to ensure compatibility.

jupyterlab/src/index.ts
```diff
@@ -4,9 +4,7 @@ import {
 } from "@jupyterlab/application";
 
 import { IEditorLanguageRegistry } from "@jupyterlab/codemirror";
-// @ts-expect-error No typings avaiable for @codemirror/legacy-modes
 import { simpleMode } from "@codemirror/legacy-modes/mode/simple-mode";
-// @ts-expect-error No typings avaiable for @codemirror/language
 import { LanguageSupport, StreamLanguage } from "@codemirror/language";
```
minestarks, line 8: New settings made these errors go away. Typings _do_ exist for these modules, maybe TypeScript was failing to discover them before?

jupyterlab/package.json
```diff
@@ -36,7 +36,7 @@
     "typescript": "^5.0.2"
   },
   "dependencies": {
-    "@codemirror/language": "6.6.0",
+    "@codemirror/language": "6.7.0",
     "@codemirror/legacy-modes": "6.3.2",
```
minestarks, line 40: Updated because `@jupyterlab/codemirror` also has a dependency on `@codemirror/language`, and when the versions don't match, the same types end up getting defined twice. Occasionally, they subtly conflict and TypeScript complains.

There were no errors before, because we were failing to resolve typings for these modules so everything was `any`.

billti (approved):

PR #738, opened 2023-09-19
Sig Help Improvements

ScottCarda-MS:
More accurate handling of tuple parameters. Adds documentation to signature and parameters.


language_service/src/signature_help.rs
```diff
@@ -65,23 +67,31 @@ impl<'a> Visitor<'a> for SignatureHelpFinder<'a> {
...
                                         {
+                                            let documentation = parse_doc_for_summary(&item.doc);
+                                            let documentation = (!documentation.is_empty())
+                                                .then_some(documentation);
+
                                             let sig_info = SignatureInformation {
                                                 label: self
                                                     .display
                                                     .hir_callable_decl(callee)
                                                     .to_string(),
```
minestarks, line 82: Style: the indentation here is getting pretty wild... this is how the diff shows up on my laptop:

image

Would it be possible to extract at least part of this function to a helper function?

ScottCarda-MS: I broke out the processing of direct callee expressions in anticipation of eventually supporting indirect callee expressions (and how that will likely look very different).

PR #743, opened 2023-09-20
Publish aarch64 wheels for Linux and Windows

idavis:
- Added `--samples/--no-samples` so that we can disable the compilation of samples during builds. This is needed when cross-compiling as we can't execute the compiler.
- Adding aarch64 linux release target. Cross-compiling on x86_64 and then running tests in an emulated aarch64 container along with auditwheel
- Adding aarch64 windows release target. This is built but not tested.

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -161,7 +161,7 @@ jobs:
       cratesIoFeedOverride: $(cratesIoFeedOverride)
       toolchainFeed: $(toolchainFeed)
     displayName: Install Rust toolchain
-    condition: ne(additionalRustTargets, '')
+    condition: ne(variables['additionalRustTargets'], '')
 
   # otherwise just install the default toolchain
   - task: RustInstaller@1
@@ -170,7 +170,7 @@ jobs:
       cratesIoFeedOverride: $(cratesIoFeedOverride)
       toolchainFeed: $(toolchainFeed)
     displayName: Install Rust toolchain
-    condition: eq(additionalRustTargets, '')
+    condition: eq(variables['additionalRustTargets'], '')
 
   - task: UsePythonVersion@0
     inputs:
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -176,6 +176,11 @@ jobs:
     displayName: Install Rust toolchain
     condition: eq(variables['additionalRustTargets'], '')
 
+  - script: |
+      rustc --version
+      rustc --print target-list
+    displayName: View rust target info
+
   - task: UsePythonVersion@0
     inputs:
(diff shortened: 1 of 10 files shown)


swernli (approved):

pip/Cargo.toml
```diff
@@ -12,10 +12,15 @@ license.workspace = true
...
-pyo3 = { workspace = true }
 qsc = { path = "../compiler/qsc" }
 miette = { workspace = true, features = ["fancy"] }
 
+[target.'cfg(not(any(target_os = "windows")))'.dependencies]
+pyo3 = { workspace = true, features = ["abi3-py37", "extension-module"] }
+
+[target.'cfg(any(target_os = "windows"))'.dependencies]
+pyo3 = { workspace = true, features = ["abi3-py37", "extension-module", "generate-import-lib"] }
+
```
minestarks, line 23: Could you add a comment so we later remember why we had to add the windows specific setting?

build.py
```diff
@@ -29,7 +29,12 @@
 parser.add_argument("--wasm", action="store_true", help="Build the WebAssembly files")
 parser.add_argument("--npm", action="store_true", help="Build the npm package")
 parser.add_argument("--play", action="store_true", help="Build the web playground")
-parser.add_argument("--samples", action="store_true", help="Compile the Q# samples")
+parser.add_argument(
+    "--samples",
+    action=argparse.BooleanOptionalAction,
+    default=True,
+    help="Build the samples (default is --samples)",
+)
```
minestarks, line 37: I'm being super nitpicky here but I'd prefer if this was moved below next to the other `BooleanOptionalAction` args.

Also I kinda liked the original description better... ("Compile the Q# samples (default is --samples)")

.ado/publish.yml
```diff
@@ -178,7 +208,33 @@ jobs:
       rm target/wheels/*-linux_x86_64.whl
       ls target/wheels
     displayName: Run auditwheel for Linux Wheels
-    condition: eq(variables['Agent.OS'], 'Linux')
+    condition: and(eq(variables['Agent.OS'], 'Linux'), eq(variables['arch'], 'x86_64'))
+
+  # for linux aarch64 cross build we want to skip tests as we can't run the code.
+  # and we can't run the samples as qsc is currently cross compiled.
+  - script: |
+      env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu python build.py --pip --no-check-prereqs --no-test --no-samples
```
minestarks, line 216: Do you want to do `--no-check` as well here like you did for the Windows one below?

minestarks (approved):

PR #744, opened 2023-09-21
Switch from deprecated `checkCode` to `update_document` from the language service in `basics.js`.

### LINKED ISSUE #469: Switch from deprecated `checkCode` to `update_document` from the language service in `basics.js`.
In `basics.js`, we currently use a deprecated test function. We should move off of it.

swernli:
Fixes #469

### DIFF
--- a/npm/test/basics.js
+++ b/npm/test/basics.js
@@ -71,33 +71,45 @@ namespace Test {
 });
 
 test("one syntax error", async () => {
-  const compiler = getCompiler();
-
-  const diags = await compiler.checkCode("namespace Foo []");
-  assert.equal(diags.length, 1);
-  assert.equal(diags[0].start_pos, 14);
-  assert.equal(diags[0].end_pos, 15);
+  const languageService = getLanguageService();
+  let gotDiagnostics = false;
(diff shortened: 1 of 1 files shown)


billti:
Do we still have coverage for that API anywhere, being that's what QCOM is still using currently. If we are removing it, we'll need to document migration steps for them. I believe the plan was they should move first, then we'll remove it after.

swernli:
> Do we still have coverage for that API anywhere, being that's what QCOM is still using currently. If we are removing it, we'll need to document migration steps for them. I believe the plan was they should move first, then we'll remove it after.

The API hasn't been removed, so we aren't breaking anyone yet. The `checkCode` implementation already just wraps calls into the language server, so the only special thing it does is map positions in the returned diagnostics with `mapUtf8UnitsToUtf16Units`. Do we want coverage specifically for that? Because the previous tests really didn't verify any corner cases of that mapping.

billti:
If QCOM is still using checkCode, (which arguably gets a lot more usage than our VS Code extension currently), then I think we should still have the tests exercise checkCode.

swernli:
In that case, I'll drop this for now.

PR #746, opened 2023-09-21
Empty `Result[]` array should be rejected by QIR codegen

### LINKED ISSUE #725: Empty `Result[]` array should be rejected by QIR codegen
Empty arrays can cause problems in processing for providers and tools. Compilation analysis doesn't always know the size of an array, but QIR codegen can catch this at "runtime" when generating the QIR. We'll need to add a new error case to catch this and fail codegen when it is detected.

swernli:
This adds a new error case for evaluation that is only used by codegen to identify times where an empty array is returned as part the final result of execution. This must be a "runtime" error because the size of the array is not known at compile time but can be known once the program has been evaluated as part of codegen.
Fixes #725

### DIFF
--- a/compiler/qsc_codegen/src/qir_base.rs
+++ b/compiler/qsc_codegen/src/qir_base.rs
@@ -9,11 +9,11 @@ use num_complex::Complex;
 use qsc_data_structures::index_map::IndexMap;
 use qsc_eval::{
     backend::Backend,
-    debug::{map_hir_package_to_fir, Frame},
+    debug::{map_fir_package_to_hir, map_hir_package_to_fir, Frame},
     eval_expr, eval_stmt,
     output::GenericReceiver,
     val::{GlobalId, Value},
-    Env, Error, Global, NodeLookup, State,
+    Env, Error, Global, NodeLookup, PackageSpan, State,
 };
 use qsc_fir::fir::{BlockId, ExprId, ItemKind, PackageId, PatId, StmtId};
 use qsc_frontend::compile::PackageStore;
@@ -52,6 +52,14 @@ pub fn generate_qir(
 
     let mut stdout = std::io::sink();
     let mut out = GenericReceiver::new(&mut stdout);
+    let entry_span = PackageSpan {
+        package: map_fir_package_to_hir(package),
+        span: unit
+            .exprs
+            .get(entry_expr)
+            .expect("entry expr should be present in map")
+            .span,
+    };
     let result = eval_expr(
         &mut State::new(package),
         entry_expr,
@@ -64,8 +72,7 @@ pub fn generate_qir(
     );
     match result {
         Ok(val) => {
-            sim.write_output_recording(&val)
-                .expect("writing to string should succeed");
+            sim.write_output_recording(&val, entry_span)?;
             sim.instrs.push_str(POSTFIX);
             Ok(sim.instrs)
         }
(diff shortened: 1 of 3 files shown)


compiler/qsc_codegen/src/qir_base.rs
```diff
@@ -64,8 +72,7 @@ pub fn generate_qir(
     );
     match result {
```
minestarks, line 73: Not your change, but since you're returning the error directly, wouldn't it simplify the code to do `let val = eval_expr(...)?` above instead of having the `match` here?

minestarks: Same comment for the `generate_qir_for_stmt` change below.

compiler/qsc_codegen/src/qir_base.rs
```diff
@@ -169,21 +179,28 @@ impl BaseProfSim {
...
+    fn write_output_recording(
+        &mut self,
+        val: &Value,
+        entry_span: PackageSpan,
+    ) -> std::result::Result<(), (Error, Vec<Frame>)> {
         match val {
             Value::Array(arr) => {
-                self.write_array_recording(arr.len())?;
+                if arr.is_empty() {
+                    return Err((qsc_eval::Error::EmptyArrayOutput(entry_span), Vec::new()));
```
minestarks, line 190: It would be nice to return a real call stack here, no? Is there any way you can use `get_stack_frames` from here?

compiler/qsc_codegen/src/qir_base.rs
swernli: Good call!

compiler/qsc_eval/src/lib.rs
```diff
@@ -51,6 +51,13 @@ pub enum Error {
     #[diagnostic(code("Qsc.Eval.DivZero"))]
     DivZero(#[label("cannot divide by zero")] PackageSpan),
 
+    #[error("cannot record output for empty array")]
+    #[diagnostic(help(
+        "returning an empty array in the final result of a program is not supported"
+    ))]
+    #[diagnostic(code("Qsc.Eval.EmptyArrayOutput"))]
+    EmptyArrayOutput(PackageSpan),
```
minestarks, line 59: Nit: I don't know if this would count as way too much work just to add one new error, but I think it would be nice if this was under a different error enum specific to QIR generation.

minestarks (approved):

compiler/qsc_codegen/src/qir_base.rs
swernli: That's the tricky thing... we aren't inside of execution anymore, we are in the post-execution output processing. This code doesn't run "inside" any callable or even in the entry expression, it's something we do with the result after execution is done. So empty call stack is technically correct.

compiler/qsc_eval/src/lib.rs
swernli: Yeah, I considered that, and I think you are right. It sort of belongs as a QIR-only error, and then QIR generation would return an aggregate error that is either evaluation error or something QIR specific. That would require updating all the callers of QIR generation, which definitely makes the change bigger but might be worth it, especially if we expect more QIR codegen specific errors later (and I do).

swernli:
Marking as draft for now. We might want this to be a service-side check, like the one for qubit and result usage that already exists today to prevent empty programs from going to partners.

PR #753, opened 2023-09-25
Samples are only run when specified or during build_all

idavis:


swernli (approved):
Thanks for fixing this!

PR #754, opened 2023-09-25
Relax clippy docs requirements

swernli:
We've discussed relaxing the clippy requirements for doc comments on public functions given that our crates are not published for use outside of the codebase. If we ever decide to publish we can turn these back on and follow more formal commenting requirements.


minestarks (approved):
I've never approved of a PR more :)

cesarzc (approved):

PR #756, opened 2023-09-25
Refactor to eliminate stateless interpreter

swernli:
This change eliminates the stateless interpreter and refactors the internals of the stateful interpreter to clean up some code repetition. For now this leaves the stateful interpreter as a separate `stateful` module to make other in-flight refactoring a little easier to merge, but a follow up can and should move the code from `stateful` into `interpret` since now there is only one interpreter implementation.

Performance benchmarks compared against main show no negative impact:
```
Teleport evaluation
                        time: [167.24 µs 167.90 µs 168.63 µs]
                        change: [-1.7025% -0.6191% +0.3523%] (p = 0.25 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe

Deutsch-Jozsa evaluation
                        time: [35.189 ms 35.275 ms 35.374 ms]
                        change: [-1.4248% -0.9964% -0.5813%] (p = 0.00 < 0.05)
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

Large file parity evaluation
                        time: [37.790 ms 37.896 ms 38.014 ms]
                        change: [-1.2062% -0.6223% -0.0448%] (p = 0.04 < 0.05)
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe

Large input file
                        time: [32.822 ms 33.113 ms 33.481 ms]
                        change: [-1.5965% -0.2764% +1.1202%] (p = 0.70 > 0.05)
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  9 (9.00%) high mild
  4 (4.00%) high severe

Standard library
                        time: [12.261 ms 12.371 ms 12.500 ms]
                        change: [-5.1162% -3.3155% -1.5653%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
```


compiler/qsc/src/interpret/stateful.rs
```diff
@@ -438,111 +490,54 @@ impl Interpreter {
...
-
-                    vec![Error(
-                        error::from_eval(error, &self.store, stack_trace).into(),
-                    )]
-                },
-            );
+            if i != 0 {
+                // If running more than one shot, re-initialize the simulator to start the next shot
+                // from a clean state.
+                sim.reinit();
```
minestarks, line 536: Perhaps instead of requiring a `reinit` method, we can have this function take a function that creates a new simulator instance? I kind of liked the old approach of using `SparseSim::new()`. IMO, it makes it clearer that absolutely no state will be shared between instances.

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -406,22 +423,57 @@ impl Interpreter {
...
+
+        let mut sim = BaseProfSim::new();
+        let mut stdout = std::io::sink();
+        let mut out = GenericReceiver::new(&mut stdout);
+
+        let val = self
+            .run_with_sim(&mut sim, &mut out, expr, 1)?
+            .into_iter()
+            .next()
+            .expect("execution should have exactly one result")?;
```
minestarks, line 450: Nit: This is really checking for "at least" one result, not "exactly". Not sure if it's worth adding the extra validation, but the expect message could be updated.

compiler/qsc/src/interpret/stateful.rs
swernli: That approach would work for `run` but for other callers that need to interact with the simulator after execution it gets harder. The function would need to return an owned instance, which means the caller wouldn't get to keep it after execution. The `qirgen` function, for example, needs to call `finish` on the simulator get the constructed QIR out.

compiler/qsc/benches/eval.rs
```diff
@@ -13,38 +13,41 @@ const LARGE: &str = include_str!("./large.qs");
...
-        let evaluator = stateless::Interpreter::new(true, sources).expect("code should compile");
+        let mut evaluator =
+            stateful::Interpreter::new(true, sources, PackageType::Exe, TargetProfile::Full)
+                .expect("code should compile");
         b.iter(move || {
             let mut out = Vec::new();
             let mut rec = GenericReceiver::new(&mut out);
-            let mut eval_ctx = evaluator.new_eval_context();
-            assert!(eval_ctx.eval_entry(&mut rec).is_ok());
+            assert!(evaluator.eval_entry(&mut rec).is_ok());
```
minestarks, line 22: Hmm, there's a behavior difference here (and in the below two benchmarks) because we're not using a fresh evaluation context each time, meaning the iterations aren't as independent as they used to be. Intentional?

compiler/qsc/src/interpret/stateful.rs
swernli: Whoops, yes! I had fixed this comment at one point, but then needed to revert some changes and lost this. One option I explored that got lost in the shuffle is using `.last()` instead, since technically I only want the output from the last statement. However, the difference is negligible because we know `run_with_sim` uses the expression parser, so there should only ever be one result from an expression. Still, I think updating the wording and using `.last()` do make the intent clearer.

minestarks (approved):
Looks good, just had a couple of minor callouts.

DmitryVasilevsky (approved):
Reviewed katas and libraries.

compiler/qsc/benches/eval.rs
swernli: Yes, that's technically true. In this case we know it's safe to repeatedly run the entry point callable without fully re-initializing the state for a couple of reasons:
- The execution succeeds, so no early termination leaves the environment in any intermediate states.
- The entry point is being invoked directly without any top-level statements, so there are no global variables to contend with and all qubits have been released.
- Released qubits must have been in the |0⟩ state or they would have triggered an error, so there is no leftover global quantum state.
- Insider knowledge of the simulator tells us that when all qubits are released, all accumulated phase is cleared because the internal data structures are reinitialized (see https://github.com/qir-alliance/qir-runner/blob/main/sparsesim/src/lib.rs#L165-L169)

minestarks: You've convinced me it's safe, but is it a valid performance benchmark? Are the iterations sharing any cached state?

compiler/qsc/src/interpret/stateful.rs
swernli: Updated!

compiler/qsc/benches/eval.rs
swernli: Fair question. The intent of the benchmark separating out the construction of the interpreter from the evaluation, even in the old API pattern, was to perform compilation outside of the benchmarked loop and focus on measuring just execution time. I don't think we have any shared state between iterations that invalidates that goal; technically the environment object and simulator object are preserved but empty, so the only thing we aren't measuring is the time needed to construct those. That lines up with my intent, at least, of trying as much as possible to measure the perf of the evaluator internals. And the fact that the benchmark numbers stayed the same supports the idea that we are still measuring those internals which are unchanged.

All that said, I think what I outlined above undercuts the use of these benchmarks to measure any change in behavior for interpreter elements that surround the evaluator, because those aren't being measured here. The benchmarks in this file measure just the evaluator, the benchmarks in large.rs and library.rs measure the compiler, but we don't actually have benchmarks that measure the end-to-end of combining the two. Perhaps the wasm folder would be the right place to add those?

cesarzc (approved):

idavis (approved):

PR #758, opened 2023-09-26
Cleaning up samples execution in build script

idavis:
Reverting workaround for avoiding samples as they now run when they're supposed to.


minestarks (approved):
Thank you!

PR #759, opened 2023-09-26
Language service for notebook cells

minestarks:
Start here: [language_service/src/lib.rs](https://github.com/microsoft/qsharp/pull/759/files#diff-cc9d1999e065d2e7a12c284a2fa5453b8b019e63ae6a77772d9e96a0636a6d73), [language_service/src/compilation.rs](https://github.com/microsoft/qsharp/pull/759/files#diff-d98d007bbe0a3fce75f8d49278e0b566c9f8d91731f3e451296db14bebff001f), [vscode/src/notebook.ts](https://github.com/microsoft/qsharp/pull/759/files#diff-dcb5937f0acbe118194e3aa0e1a3960aa8ee986e5f6b71d6f438e876be8d05a9)

Added support for multiple documents per `Compilation` in the language service. This has implications on how we keep state and how we publish diagnostics.

Added two `CompilationKind`s to the language service, `OpenDocument` and `Notebook`. Notebooks are made up of cells, with each cell corresponding to a single document, and all compiled together using the incremental compiler. The `OpenDocument` compilation kind represents the existing, single-Q#-document type of compilation.

Added `update_notebook_document`/`close_notebook_document` and wired it up to VS Code to support notebook cells.

Moved a bunch of loose code out of `qsc_utils.rs` into a new file, `compilation.rs` and attempted to group the functions into coherent traits.

Changes for language service features to support incremental syntax, e.g. completion list now inserts item keywords & globals at the top-level.


language_service/src/test_utils.rs
```diff
@@ -66,10 +69,99 @@ pub(crate) fn compile_with_fake_stdlib(source_name: &str, source_contents: &str)
...
+                );
+                assert!(
+                    offset.replace(cursor_offsets[0]).is_none(),
+                    "only one cell can have a cursor marker"
+                );
+            }
+            if !targets.is_empty() {
+                assert!(
+                    target_cell_uri.replace(c.0).is_none(),
+                    "only one cell can have a target marker"
```
ScottCarda-MS, line 105: Why can only one cell contain target markers? I can imagine a rename or find-all-references operation that targets things in multiple cells.

vscode/src/definition.ts
```diff
@@ -23,10 +24,28 @@ class QSharpDefinitionProvider implements vscode.DefinitionProvider {
...
-    ).positionAt(definition.offset);
+
+    const definitionPosition = (await openDocument(uri)).positionAt(
+      definition.offset,
+    );
     return new vscode.Location(uri, definitionPosition);
   }
 }
+
+async function openDocument(uri: vscode.Uri) {
```
ScottCarda-MS, line 35: I'm wondering if this is going to need to be used more broadly across several different LS features like rename and FAR.

npm/src/language-service/language-service.ts
```diff
@@ -165,33 +196,46 @@ export class QSharpLanguageService implements ILanguageService {
...
-      // Inspect the URL protocol (equivalent to the URI scheme + ":").
-      // If the scheme is our library scheme, we need to call the wasm to
-      // provide the library file's contents to do the utf8->utf16 mapping.
-      const url = new URL(result.source);
-      if (url.protocol === qsharpLibraryUriScheme + ":") {
-        code = wasm.get_library_source_content(url.pathname);
-        if (code === undefined) {
-          log.error(`getDefinition: expected ${url} to be in the library`);
-          return undefined;
+      let targetCode = this.code[result.source];
```
ScottCarda-MS, line 214: Do we need to be concerned about the edge case where there is a user source with the same name as a library source? Would they have the same URI except the library source would have the scheme prepending it?

language_service/src/compilation.rs
```diff
@@ -0,0 +1,258 @@
...
+                        Some(s)
+                    }
+                })
+        })
+        .map(|s| (s.name.as_ref(), s.contents.as_ref()))
+        .collect()
+    }
+}
+
+pub(crate) trait Lookup {
```
ScottCarda-MS, line 169: I didn't see where this trait is used beyond `Compilation`. Is it used somewhere else?

language_service/src/rename.rs
```diff
@@ -63,9 +63,11 @@ impl<'a> Rename<'a> {
     }
 
     fn get_spans_for_item_rename(&mut self, item_id: &hir::ItemId, ast_name: &ast::Ident) {
```
minestarks, line 65: Bug: Through testing we discovered that we return the wrong span here when the rename is across cells.

language_service/src/compilation.rs
minestarks: No, it's not used anywhere else. It's only hear to provide a bit of code organization - I feel that all these methods belong together and this felt like a nice way to group them together. 

I expect `Compilation` and its impl to grow and grow in the future, and I'm slightly concerned about it veering into [God object](https://en.wikipedia.org/wiki/God_object) territory. I wanted to take a step towards breaking it up without necessarily doing all the work to separate out the data yet. I might perhaps pull out `package_store` and `user_package` into their own struct definition one day, and on that day `Lookup` can just become an impl for _that_ struct.

npm/src/language-service/language-service.ts
minestarks: They would both be full URIs, with different schemes, so I believe we're safe.

Example:
`vscode-test-web://mount/intrinsic.qs`
`qsharp-library-source:intrinsic.qs`

vscode/src/notebook.ts
```diff
@@ -65,6 +72,98 @@ export function registerQSharpNotebookHandlers() {
...
+        vscode.languages.match(qsharpDocumentFilter, cell.document),
+      );
+  }
+
+  function getQSharpText(document: vscode.TextDocument) {
+    // Erase the %%qsharp magic line if it's there
+    // Replace it with whitespace so that document offsets remain the same.
+    // This will save us from having to map offsets later when
+    // communicating with the language service.
+    if (document.lineAt(0).text.startsWith(qsharpCellMagic)) {
```
idavis, line 154: Is it common practice to require language magic to be the very first line in a a cell?

minestarks: Are you asking about allowing whitespace/empty lines before the cell magic? Technically, IPython requires the cell magic to be the first thing in the cell, it does allow empty lines / whitespace before it. Though I've never seen any user code that doesn't just put the cell magic on the first line.

We don't skip whitespace here, so we're being more restrictive than IPython, but there's no real reason for it... I can update our code to allow whitespace.

vscode/src/definition.ts
minestarks: Yes, this will probably be used for any language feature that could return spans from different files. I can move it to a common location.

language_service/src/test_utils.rs
minestarks: Well, this was convenient right now since other unit tests don't use target markers. You could totally update those unit tests to use target markers :) . Then we can relax this restriction.

vscode/src/definition.ts
```diff
@@ -23,10 +24,28 @@ class QSharpDefinitionProvider implements vscode.DefinitionProvider {
...
+          .filter((cell) =>
+            vscode.languages.match(qsharpDocumentFilter, cell.document),
+          )
+          .map((cell) => cell.document),
+      ),
+    )
+    .find((doc) => doc.uri.toString() === uri.toString());
+
+  return doc || (await vscode.workspace.openTextDocument(uri));
+}
```
idavis, line 51: Does it make sense to cache the uri.toString() so that we aren't calling it for every comparison?

wasm/src/language_service.rs
```diff
@@ -59,6 +62,27 @@ impl LanguageService {
         self.0.close_document(uri);
     }
 
+    pub fn update_notebook_document(&mut self, notebook_uri: &str, cells: Vec<ICell>) {
+        let cells: Vec<Cell> = cells.into_iter().map(|c| c.into()).collect();
```
idavis, line 66: We are double materializing (collect) the cells here. Does it make sense to continue mapping so that we only iterate once and pass the result into the update call?

minestarks: Yeah I'm glad you brought it up, because this one drove me crazy. But I don't think it's possible to make this work without double-materializing. If I try to chain the `map`s like below, then the `.as_ref()` calls create a problem as we end up with a reference to a temporary value (created by `Cell::from(c)`). In order for the strings in `Cell` _not_ to be temporary, I need to materialize once after I convert from `JsValue` to `Cell`. Then, I need to materialize again since the callee `update_notebook_document` expects a slice, not an iterator.

```rs
let cells = cells
            .into_iter()
            .map(|c| Cell::from(c))
            .map(|s| (s.uri.as_ref(), s.version, s.code.as_ref()))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot return value referencing local data
                                      `s.uri` returns a value referencing data owned by the current function
            .collect::<Vec<(&str, u32, &str)>>();
```

minestarks: I think I'll have the inner `update_notebook_document` take an iterator instead of a slice. That should fix it.

vscode/src/definition.ts
minestarks: It does. Thanks!

language_service/src/test_utils.rs
ScottCarda-MS: I've updated the other unit tests in the language server to use target markers where possible.

idavis (approved):

language_service/src/compilation.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    /// Gets the `CompileUnit` associated with user (non-library) code.
+    pub fn user_unit(&self) -> &CompileUnit {
+        self.package_store
+            .get(self.user)
+            .expect("expected to find user package")
+    }
+
+    /// Returns the package and `SourceMap` offset associated with a code location.
+    pub(crate) fn resolve_offset(&self, source_name: &str, offset: u32) -> (&AstPackage, u32) {
```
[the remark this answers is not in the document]
minestarks: Yeah I don't like how this refactoring came out. I took your naming suggestion, and dropped the `AstPackage` from the return value since it only ever returns the user package and the caller already has a reference to that. Thanks.

language_service/src/rename.rs
minestarks: Fixed

language_service/src/lib.rs
```diff
@@ -92,125 +120,219 @@ impl<'a> LanguageService<'a> {
...
+            trace!("update_notebook_document: cell: {cell_uri} {version}");
+            self.open_documents.insert(
+                (*cell_uri).into(),
+                OpenDocument {
+                    version: *version,
+                    compilation: compilation_uri.clone(),
+                },
+            );
+        }
+
```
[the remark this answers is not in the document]
minestarks: Fixed, though this was totally harmless, as we don't ever use the presence of a document in `open_documents` to determine anything.

language_service/src/compilation.rs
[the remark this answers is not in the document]
ScottCarda-MS: Sorry I'm not seeing the change you mentioned. Is there a commit that hasn't been pushed yet?

language_service/src/compilation.rs
```diff
@@ -0,0 +1,258 @@
...
+use std::iter::successors;
+
+/// Represents an immutable compilation state that can be used
+/// to implement language service features.
+pub(crate) struct Compilation {
+    /// Package store, containing the current package and all its dependencies.
+    pub package_store: PackageStore,
+    /// The `PackageId` of the user package. User code
+    /// is non-library code, i.e. all code except the std and core libs.
+    pub user: PackageId,
```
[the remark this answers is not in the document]
ScottCarda-MS: Sorry I'm not seeing a change here. Is there a commit that hasn't been pushed yet?

language_service/src/compilation.rs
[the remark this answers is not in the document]
ScottCarda-MS: Looks like the PR got out of sync with your branch somehow. Maybe it is just a matter of waiting for it to get back in sync.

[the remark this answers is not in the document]
ScottCarda-MS: I see the sync fixed the confusion. Thanks for looking into this!

language_service/src/compilation.rs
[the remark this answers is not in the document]
ScottCarda-MS: I see the sync fixed the confusion. Thanks for looking into this!

ScottCarda-MS (approved):

swernli (approved):

vscode/test/suites/language-service/notebook.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    await activate();
+    const notebook = await vscode.workspace.openNotebookDocument(
+      vscode.Uri.joinPath(workspaceFolderUri, "test.ipynb"),
+    );
+
+    const firstQSharpCellUri = notebook.cellAt(1).document.uri;
+    const secondQSharpCellUri = notebook.cellAt(2).document.uri;
+
+    const pos = new vscode.Position(2, 0); // cursor on the usage of Test()
+    console.error("please what");
```
billti, line 84: ??

billti (approved):

vscode/test/suites/language-service/notebook.test.ts
minestarks: Wow... that's an interesting one. No idea what I was in the middle of doing there but glad you caught it.

cesarzc (approved):

PR #762, opened 2023-09-28
Refactor the incremental compiler

minestarks:
The motivation for this refactoring is to be able to implement language service support for notebook cells. (You can preview how that's going to fit together at #759.) But it achieves more than that, by eliminating a lot of code duplicated between the batch compilation and incremental compilation paths, devising a coherent model for the incremental compiler, and clearly separating the evaluation parts of the `Interpreter` from the compilation parts.

**1. New top-level nodes in the AST and HIR**

Previously, the incremental compilation was made up of loose AST and HIR nodes stored in `stateful::Interpreter`, and an empty `CompileUnit` in the package store. This packages in this `CompileUnit` were not updated as fragments got compiled in.

Every feature in the language service is implemented using an AST visitor and uses the HIR package for lookups. This imposes the requirement that notebook cells are able to be represented as AST and HIR packages. They should be kept up to date with each cell added to the compilation.

Q# notebook cells were previously thought of as "fragments" (not "real" packages). This PR makes fragments syntax official, so to speak, by adding the ability for `ast::Package` and `hir::Package` to contain top-level statements.

When we are able to represent each incremental unit (notebook cell) as a whole package, we can eliminate a lot of lowering/resolution/checking code that was designed to work with loose statement/item nodes (e.g. `check_stmt_fragment`).. We can now use their package counterparts (e.g. `check_package`), making for more code shared between the batch and incremental compilers.

**2. Layering and `qsc::incremental`**

Previously, the incremental compiler implementation was spread out between `stateful` and `qsc_frontend::incremental`. The `stateful::Interpret` struct contained too much state, making it really hard to write code that mutates logically separate parts of the struct (e.g. compiler state vs. evaluator state) without angering the Rust borrow checker.

The new `qsc::incremental` module is mainly the compiler-y parts of `stateful` pulled out into their own module, bridging the gap between `stateful` and `qsc_frontend::incremental`. The layering is meant to parallel that of batch compilation:

`qsi` -> `qsc::interpret::stateful` -> `qsc::incremental` -> `qsc_frontend::incremental`
`qsc` -> `qsc::compiler` -> `qsc_frontend::compiler`

**3. Mutable `CompileUnit` in the `PackageStore`**

#675 required that the current compilation be available to be looked up in the `PackageStore`, since errors may contain the `PackageId` of the current package. But the current compilation also needs to be updated with incremental changes. Adding a `get_mut` to the `PackageStore` felt wrong since almost all use cases of the `PackageStore` require packages to remain immutable. To address the awkwardness, the concept of an `OpenPackageStore` is introduced. A `PackageStore` can be "opened" for modifications, making it possible to obtain a mutable reference to only the last `CompileUnit` in the store (the open package). The incremental compiler uses the open package store to keep updating the compilation, while the rest of the components (evaluator, error span lookups, etc) have access to the regular, immutable, `PackageStore` for lookups.

### DIFF
--- a/compiler/qsc/src/incremental.rs
+++ b/compiler/qsc/src/incremental.rs
@@ -139,15 +139,7 @@ impl Compiler {
     pub fn update(&mut self, new: Increment) {
         let (_, unit) = self.store.get_open_mut();
 
-        // Extend the AST and HIR packages
-        unit.ast.package.extend(new.ast.package);
-        unit.package.extend(new.hir);
-
-        // The new `Increment` will contain the names and tys
-        // from the original package as well, so just
-        // replace the current tables instead of extending.
-        unit.ast.names = new.ast.names;
-        unit.ast.tys = new.ast.tys;
+        self.frontend.update(unit, new);
     }
 
     /// Returns a reference to the underlying package store.
--- a/compiler/qsc/src/incremental.rs
+++ b/compiler/qsc/src/incremental.rs
@@ -6,7 +6,7 @@ use miette::Diagnostic;
 use qsc_frontend::{
     compile::{OpenPackageStore, PackageStore, SourceMap, TargetProfile},
     error::WithSource,
-    incremental::CompileUnitAddition,
+    incremental::Increment,
 };
 use qsc_hir::hir::PackageId;
 use qsc_passes::{PackageType, PassContext};
@@ -81,16 +81,16 @@ impl Compiler {
         &mut self,
         source_name: &str,
         source_contents: &str,
-    ) -> Result<CompileUnitAddition, Errors> {
+    ) -> Result<Increment, Errors> {
         let (core, unit) = self.store.get_open_mut();
 
-        let mut unit_addition = self
+        let mut increment = self
             .frontend
             .compile_fragments(unit, source_name, source_contents)
             .map_err(into_errors)?;
 
         let pass_errors = self.passes.run_default_passes(
-            &mut unit_addition.hir,
+            &mut increment.hir,
             &mut unit.assigner,
             core,
             PackageType::Lib,
@@ -100,7 +100,7 @@ impl Compiler {
             return Err(into_errors_with_source(pass_errors, &unit.sources));
         }
 
-        Ok(unit_addition)
+        Ok(increment)
     }
 
     /// Compiles an entry expression.
@@ -112,16 +112,16 @@ impl Compiler {
     /// get information about the newly added items, or do other modifications.
     /// It is then the caller's responsibility to merge
     /// these packages into the current `CompileUnit` using the `update()` method.
-    pub fn compile_expr(&mut self, expr: &str) -> Result<CompileUnitAddition, Errors> {
+    pub fn compile_expr(&mut self, expr: &str) -> Result<Increment, Errors> {
         let (core, unit) = self.store.get_open_mut();
 
-        let mut unit_addition = self
+        let mut increment = self
             .frontend
             .compile_expr(unit, "<entry>", expr)
             .map_err(into_errors)?;
 
         let pass_errors = self.passes.run_default_passes(
-            &mut unit_addition.hir,
+            &mut increment.hir,
             &mut unit.assigner,
             core,
             PackageType::Lib,
@@ -131,23 +131,23 @@ impl Compiler {
             return Err(into_errors_with_source(pass_errors, &unit.sources));
         }
 
-        Ok(unit_addition)
+        Ok(increment)
     }
 
     /// Updates the current compilation with the AST and HIR packages,
     /// and any associated context, returned from a previous incremental compilation.
-    pub fn update(&mut self, addition: CompileUnitAddition) {
+    pub fn update(&mut self, new: Increment) {
         let (_, unit) = self.store.get_open_mut();
 
         // Extend the AST and HIR packages
-        unit.ast.package.extend(addition.ast.package);
-        unit.package.extend(addition.hir);
+        unit.ast.package.extend(new.ast.package);
+        unit.package.extend(new.hir);
 
-        // The other package will contain the names and tys
+        // The new `Increment` will contain the names and tys
         // from the original package as well, so just
         // replace the current tables instead of extending.
-        unit.ast.names = addition.ast.names;
-        unit.ast.tys = addition.ast.tys;
+        unit.ast.names = new.ast.names;
+        unit.ast.tys = new.ast.tys;
     }
 
     /// Returns a reference to the underlying package store.
--- a/compiler/qsc/src/incremental.rs
+++ b/compiler/qsc/src/incremental.rs
@@ -1,0 +1,199 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::compile::{self, compile, core, std};
+use miette::Diagnostic;
+use qsc_frontend::{
+    compile::{OpenPackageStore, PackageStore, SourceMap, TargetProfile},
+    error::WithSource,
+    incremental::CompileUnitAddition,
+};
+use qsc_hir::hir::PackageId;
+use qsc_passes::{PackageType, PassContext};
+
+/// An incremental Q# compiler.
+pub struct Compiler {
+    /// A package store that contains the current, mutable, `CompileUnit`
+    /// as well as all its immutable dependencies.
+    store: OpenPackageStore,
+    /// The ID of the source package. The source package
+    /// is made up of the initial sources passed in when creating the compiler.
+    source_package_id: PackageId,
+    /// Context for passes that is reused across incremental compilations.
+    passes: PassContext,
+    /// The frontend incremental compiler.
+    frontend: qsc_frontend::incremental::Compiler,
+}
+
+/// An incremental compiler error.
+pub type Errors = Vec<WithSource<compile::Error>>;
+
+impl Compiler {
+    /// Creates a new incremental compiler, compiling the passed in sources.
+    /// # Errors
+    /// If compiling the sources fails, compiler errors are returned.
+    pub fn new(
+        include_std: bool,
+        sources: SourceMap,
+        package_type: PackageType,
+        target: TargetProfile,
+    ) -> Result<Self, Errors> {
+        let core = core();
+        let mut store = PackageStore::new(core);
+        let mut dependencies = Vec::new();
+        if include_std {
+            let std = std(&store, target);
+            let id = store.insert(std);
+            dependencies.push(id);
+        }
+
+        let (unit, errors) = compile(&store, &dependencies, sources, package_type, target);
+        if !errors.is_empty() {
+            return Err(into_errors_with_source(errors, &unit.sources));
+        }
+
+        let source_package_id = store.insert(unit);
+        dependencies.push(source_package_id);
+
+        let frontend = qsc_frontend::incremental::Compiler::new(&store, dependencies, target);
+        let store = store.open();
+
+        Ok(Self {
+            store,
+            source_package_id,
+            frontend,
+            passes: PassContext::new(target),
+        })
+    }
+
+    /// Compiles Q# fragments. Fragments are Q# code that can contain
+    /// top-level statements as well as namespaces. A notebook cell
+    /// or an interpreter entry is an example of fragments.
+    ///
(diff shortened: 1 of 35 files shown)


compiler/qsc/src/interpret/debug/tests.rs
```diff
@@ -87,7 +87,7 @@ fn stack_traces_can_cross_eval_session_and_file_boundaries() {
                              at Adjoint Test.C in 1.qs
                              at Adjoint Test.B in 1.qs
                              at Adjoint Test2.A in 2.qs
-                             at Z in <expression>
+                             at Z in line_0
```
minestarks, line 90: This behavior change results from the fact that we're now keeping the HIR updated in the package store, allowing [`get_item_file_name`](https://github.com/microsoft/qsharp/blob/30f11ca118ba39a59f76be2adecf57f08d80ff90/compiler/qsc/src/interpret/debug.rs#L74) to work for the current package.

compiler/qsc_ast/src/ast.rs
```diff
@@ -109,12 +115,28 @@ impl Hash for NodeId {
...
+    let mut v = std::mem::take(left).into_vec();
+    v.reserve_exact(right.len());
+    v.extend(std::mem::take(right).into_vec());
+    v.into_boxed_slice()
+}
+
+impl Package {
+    /// Extends the `Package` with the contents of another `Package`.
+    /// `other` should not contain any `NodeId`s
+    /// that conflict with the current `Package`.
```
swernli, line 134: You can turn this into a check using the `Validator` from qsc_ast/src/validate.rs. It will walk the tree and panic if it finds any duplicate ids. I think that would be worth doing in this function after the call to `take_and_concat`.

minestarks: Ooh that would work perfectly, thanks!

compiler/qsc_hir/src/hir.rs
```diff
@@ -215,10 +215,25 @@ impl Display for Res {
...
+    /// The top-level statements in the package.
+    pub stmts: Vec<Stmt>,
     /// The entry expression for an executable package.
     pub entry: Option<Expr>,
 }
 
+impl Package {
+    /// Extends the `Package` with the contents of another `Package`.
+    /// `other` should not contain any `LocalItemId`s
+    /// that conflict with the current `Package`.
```
swernli, line 227: Same thing here, it's worth using the `Validator` (the hir variant) after the packages have been merged.

compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -11,40 +11,72 @@ use miette::Diagnostic;
...
+                Stmt 1 [0-16]: Qubit (Fresh)
+                    Pat 2 [4-5]: Bind:
+                        Ident 3 [4-5] "q"
+                    QubitInit 4 [8-15] Single
+            names:
+            node_id:3,
+            terms:
+            node_id:1,node_id:2,node_id:3,node_id:4,
+            hir:
+            Package:"#]],
```
swernli, line 78: Looks like this test output is missing it's HIR package content for some reason? The fact that it has `Package: ` means it is getting an HIR object, but it seems to be missing the top-level qubit allocation statement. Is the display for top-level stmts missing?

compiler/qsc_hir/src/hir.rs
```diff
@@ -215,10 +215,25 @@ impl Display for Res {
 pub struct Package {
     /// The items in the package.
     pub items: IndexMap<LocalItemId, Item>,
+    /// The top-level statements in the package.
+    pub stmts: Vec<Stmt>,
```
swernli, line 219: Aha! As mentioned in the comment on the test file, top-level statments are in the package struct but not included in the display logic below.

compiler/qsc_frontend/src/incremental/tests.rs
minestarks: Good catch, I thought I went over the new test output but I missed this. Investigating...

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -550,26 +550,31 @@ impl GlobalTable {
...
+                            &namespace.name.name,
+                            || intrapackage(assigner.next_item()),
+                            item,
+                        ) {
+                            Ok(()) => {}
+                            Err(error) => errors.push(error),
+                        }
+                    }
                 }
+                TopLevelNode::Stmt(_) => todo!("Not yet supported for top-level statements"),
```
swernli, line 577: nitpick: I think we'd want to make this `unimplemented!()` instead of `todo!()` because right now we have no reason to believe we will come back to support this (top-level statements aren't something we add here, they are handled during incremental compilation). There is a subtle distinction between the two macros as called out by the docs at https://doc.rust-lang.org/std/macro.unimplemented.html:

> The difference between unimplemented! and todo! is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims. Its message is “not implemented”. Also some IDEs will mark todo!s.

compiler/qsc_frontend/src/error.rs
```diff
@@ -24,15 +21,17 @@ impl<E: Diagnostic> WithSource<E> {
...
+    pub fn into_error(self) -> E {
+        self.error
+    }
 }
 
 impl<E: Diagnostic> WithSource<E> {
     /// Construct a diagnostic with source information from a source map.
     /// Since errors may contain labeled spans from any source file in the
     /// compilation, the entire source map is needed to resolve offsets.
     /// # Panics
```
swernli, line 34: niptick: Looks like the panic text was removed but not the `# Panic` part of the doc comment. That can be dropped now too.

compiler/qsc/src/incremental.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn source_package_id(&self) -> PackageId {
+        self.source_package_id
+    }
+
+    /// Consumes the incremental compiler and returns an immutable package store.
+    /// This method can be used to finalize the compilation.
+    #[must_use]
+    pub fn into_package_store(self) -> (PackageStore, PackageId) {
+        self.store.into_package_store()
+    }
```
swernli, line 177: nitpick: I like this from an API completeness perspective, but I wonder when we'd actually use it (right now it's not called, so it's nearly dead code but is not flagged as such because it's marked as `pub` and could technically be called by something outside the crate). I think I'm fine keeping it, but some part of me still flags this as unused code that can be safely deleted. All our incremental scenarios are interactive (Python, Jupyter, REPL) and don't really need a concept of finalizing the compilation, even if it makes sense on paper. Perhaps there is an anticipated scenario I'm not seeing?

compiler/qsc_passes/src/lib.rs
```diff
@@ -59,60 +56,80 @@ pub struct PassContext {
...
+            .map(Error::CallableLimits)
+            .chain(borrow_errors.drain(..).map(Error::BorrowCk))
+            .chain(spec_errors.into_iter().map(Error::SpecGen))
+            .chain(conjugate_errors.into_iter().map(Error::ConjInvert))
+            .chain(entry_point_errors)
+            .chain(base_prof_errors.into_iter().map(Error::BaseProfCk))
+            .collect()
+    }
+}
 /// Run the default set of passes required for evaluation.
```
swernli, line 120: nipick: missing newline

compiler/qsc_frontend/src/resolve.rs
minestarks: I didn't know about the difference, so thanks for that! But I also just realized I don't quite remember what I was going for with this change. It looks unfinished, honestly. This method is only called for batch compilation, and it's suspiciously similar to `bind_namespace_items` which is called for the incremental path. I was trying to consolidate these two paths but I think I just missed this one. I'll take a closer look to see if I can clean it up further.

compiler/qsc_frontend/src/error.rs
minestarks: It panics! What more do you wanna know? 😆

compiler/qsc/src/incremental.rs
minestarks: Ah! It's used for the language service. The language service requires a "finished" compilation. I could certainly leave it out until _that_ PR, but I figured showing the "complete" api up front would be helpful .

swernli: That sounds good enough to me! Glad to know a usage is coming down the pipe.

compiler/qsc_passes/src/baseprofck.rs
```diff
@@ -60,27 +57,6 @@ pub fn check_base_profile_compliance(package: &Package) -> Vec<Error> {
     checker.errors
```
swernli, line 57: The `check_base_profile_compliance_for_stmt` function used to have this extra check to make sure that the statment being used to trigger QIR codegen didn't return anything that was incompatible with base profile, but now nothing checks that. It can caus a panic in codegen from Python when the provided expr returns something invalid:
```python
import qsharp
qsharp.init(qsharp.TargetProfile.Base)
qir = qsharp.compile("1")
```
Used to return `Qsc.BaseProfCk.ReturnNonResult` and now hits `PanicException: unexpected value type: Int(1)`. I think you might be able to handle this by checking the last top-level statement under the assumption that QIR codegen would be using that as the entry statement, which I think holds, but is not quit obvious. Something like this:
```rust
    if let Some(StmtKind::Expr(expr)) = &package.stmts.last().map(|stmt| &stmt.kind) {
        if any_non_result_ty(&expr.ty) {
            checker.errors.push(Error::ReturnNonResult(expr.span));
        }
    }
```
added into the new unified `check_base_profile_compliance` after the similar check on the entry expression.

swernli: for what it's worth, I confirmed the above works to resolve the issue. Still worth thinking it through (and adding a test case).

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -307,10 +256,10 @@ impl Interpreter {
         if let Some(entry) = unit.entry {
             return Ok(entry);
         };
-        Err(vec![Error(ErrorKind::NoEntryPoint)])
+        Err(vec![Error::NoEntryPoint])
     }
 
-    /// # Errors
+    /// # Errors
```
swernli, line 262: nitpick: extra spaces showed up here.

swernli (commented):
There's one bug fix I think needs to be applied, and few other nitpicks to review. Overall really great refactor, looking forward to getting this in!

compiler/qsc_ast/src/ast.rs
minestarks: Huh. So I added the AST validator, which immediately caught the fact that the `NodeId` on the AST `Package` is never set! This got me thinking a little deeper about the semantics of `ast::Package::extend()`. Should it retain the current package ID? Should it adopt the Package ID of `other`? Come to think of it, shouldn't we be creating a new unique ID for the resulting package, since the result of `extend()` is a new, unique, AST? I really liked that idea. The same `NodeId` should never refer to two different "states" of a node. Those two "states" are actually two different trees, and trees should be immutable. Intuitively, I feel like this principle sets us up for better debuggability, and perhaps even things like partial recompilation, down the line.

So instead of keeping this `extend()` method that takes a mutable reference to self, I think I'll create a function with a name like `concat()`, which takes two `Package`s and creates a new, merged, one, with a new unique ID.

minestarks: Pushed commit to fix this: https://github.com/microsoft/qsharp/pull/762/commits/ea50f1b6f7bcab763269b83ec2573f4e1518e9da

I ended up pulling the implementation of `extend()`/`concat()` out of ast.rs and hir.rs. The layering seemed problematic - these functions are very specific to incremental compilation, and they now take a dependency on the validators.... so it seemed wrong that they would live alongside the AST/HIR struct definitions. They live with the incremental compiler now.

minestarks:
Made a round of updates - this is ready to review again

swernli (approved):
Updates look great!

minestarks:
@ScottCarda-MS , @idavis and/or @billti would you mind signing off on the one line change in `language_service`? Thanks!

ScottCarda-MS (approved):
signing off for the language-server changes

PR #764, opened 2023-10-03
Added Complex and ComplexPolar to Math namespace

DmitryVasilevsky:
- Added Complex, ComplexPolar types and related functions. Mosty based on previous implementation.
- Cleaned up doc comments, replaced formulas with Unicode.
- Simplified implementations, especially around multiplication and division.
- Simplified internal PowCAsCP function. Added proof of approach instead of links to Wolfram Alpha computation result.
- Clarified that we are taking one value of a multi-valued function.
- Added simple tests for all functions. Intrinsic functions already have test cases in evaluator, but I added extra tests nevertheless.
- Floating point numbers (as sole values and in tuples) are compared approximately.


minestarks:
@DmitryVasilevsky  Heads up - you'll need to merge from `main` for this PR to pass checks. I just changed the names of the checks in ci.yml #625 which is causing issues with currently open PRs. Sorry about that!

library/tests/src/lib.rs
```diff
@@ -34,5 +38,37 @@ pub fn test_expression(expr: &str, expected: &Value) {
...
         .expect("test should run successfully");
 
-    assert_eq!(expected, &result);
+    match (&expected, result) {
+        (&Value::Tuple(tup1), Value::Tuple(tup2)) if tup1.len() == tup2.len() => {
+            // If both values are tuples of the same length, we crack them open and compare elements
+            for (value1, value2) in tup1.iter().zip(tup2.iter()) {
+                if let (Value::Double(double1), Value::Double(double2)) = (value1, value2) {
+                    // If both elements are doubles, we use approximate comparison
+                    assert_doubles_almost_equal(*double1, *double2);
```
swernli, line 47: One minor note that I hadn't thought of before: this will only work one "layer" deep, so if the data structure under test is a tuple of tuple of double, as in `((Double, Double), Double)`, the inner tuple will not get caught here and still go through direct comparison. That isn't a problem right now with how this is used, but something we might want to keep in mind for the future.

swernli (approved):

library/tests/src/lib.rs
DmitryVasilevsky: True. Also we don't handle array elements, etc. I guess we will exert the effort when we need it :)

library/std/math.qs
```diff
@@ -442,6 +441,7 @@ namespace Microsoft.Quantum.Math {
     /// Returns a tuple (u,v) such that u*a+v*b=GCD(a,b)
     /// Note: GCD is always positive except that GCD(0,0)=0.
     function ExtendedGreatestCommonDivisorL(a : BigInt, b : BigInt) : (BigInt, BigInt) {
+        open Microsoft.Quantum.Convert;
```
cesarzc, line 444: Interesting that the open directive was moved here, I didn't know it could be done. Was there any particular reason to move it here?

cesarzc (approved):

library/std/math.qs
DmitryVasilevsky: Conversions are only used in three functions, so I decided to move it to those functions.

DmitryVasilevsky: (I believe the ability to put open directive in any block is the feature of the new compiler)

PR #766, opened 2023-10-03
Fix cross-unit UDT dependency

swernli:
This fixes a subtle bug in usage of UDTs across compilation units. Because a UDT is captured as a type that includes a `Res` in it, it wil have the placeholder `None` for it's package id when initially compiled (package ids are only known after insertion into the store). Later, a type mismatch could occur because the caller has the package id in the UDT type but the signature stored in the scheme uses UDT types with `None`. This fix adds a notion of updating a scheme to include the package id at type checking time by using `with_package` such that type checking for both internal/`None` resolutions and dependencies from other compilation units with known package ids can all be properly used.

### DIFF
--- a/compiler/qsc_frontend/src/compile/tests.rs
+++ b/compiler/qsc_frontend/src/compile/tests.rs
@@ -484,34 +484,26 @@ fn package_dependency() {
     let unit2 = compile(&store, &[package1], sources2, TargetProfile::Full);
     assert!(unit2.errors.is_empty(), "{:#?}", unit2.errors);
 
-    let ItemKind::Callable(callable) = &unit2
-        .package
-        .items
-        .get(LocalItemId::from(1))
-        .expect("package should have item")
-        .kind
-    else {
-        panic!("item should be a callable");
-    };
-    let SpecBody::Impl(_, block) = &callable.body.body else {
-        panic!("callable body have a block")
-    };
-    let StmtKind::Expr(expr) = &block.stmts[0].kind else {
-        panic!("statement should be an expression")
-    };
-    let ExprKind::Call(callee, _) = &expr.kind else {
-        panic!("expression should be a call")
-    };
-    let ExprKind::Var(res, _) = &callee.kind else {
-        panic!("callee should be a variable")
-    };
-    assert_eq!(
-        &Res::Item(ItemId {
-            package: Some(package1),
-            item: LocalItemId::from(1),
-        }),
-        res
-    );
+    expect![[r#"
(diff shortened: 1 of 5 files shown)


compiler/qsc_frontend/src/compile/tests.rs
```diff
@@ -478,12 +482,12 @@ fn package_dependency() {
         None,
     );
     let unit2 = compile(&store, &[package1], sources2, TargetProfile::Full);
+    assert!(unit2.errors.is_empty(), "{:#?}", unit2.errors);
 
-    let foo_id = LocalItemId::from(1);
```
DmitryVasilevsky, line 482: Is this change needed just to uncouple two blocks or LocalItemId::from(1) actually changes between two uses?

swernli: Nope, the id is the same in both places. I was just removing it because I found it confusing more than helpful and it matches the pattern of the new test I added, but could revert if you think it's worth keeping. The ID does indeed stay the same between the two uses and could be a variable.

compiler/qsc_hir/src/ty.rs
```diff
@@ -50,6 +50,25 @@ pub enum Ty {
...
     pub const UNIT: Self = Self::Tuple(Vec::new());
+
+    #[must_use]
+    pub fn with_package(&self, package: Option<PackageId>) -> Self {
+        match self {
+            Ty::Array(item) => Ty::Array(Box::new(item.with_package(package))),
+            Ty::Arrow(arrow) => Ty::Arrow(Box::new(arrow.with_package(package))),
+            Ty::Infer(infer) => Ty::Infer(*infer),
+            Ty::Param(param) => Ty::Param(*param),
+            Ty::Prim(prim) => Ty::Prim(*prim),
```
DmitryVasilevsky, line 61: Don't know much Rust, hence the question. Can you use the same trick for the three cases as you used before ("_ => *self,")

swernli: Yes, actually I could!

swernli: Actually, no, because each case needs to construct a different type.

idavis (approved):

compiler/qsc_hir/src/ty.rs
swernli: Ok, never mind. Came back again, and you are right. Each case was created an instance of the same type, so those can all be captured by a single case that clones. Good eye!

PR #769, opened 2023-10-04
Update TextMate grammar

swernli:
This change removes some older parts of the grammar used for syntax highlighting Q# source. We had previously removed the old disallowed keywords based on C#/.NET compatibility, and this comes back and removes some keywords that are no longer supported and the outdated list of built-in functions. Rather than fight to keep that list up-to-date and continually have some callables from the stdlib highlighted differently than others or those defined in user code, we should defer callable highlighting until we decide whether we want to do semantic highlighting.

For a quick visual of the differences, here is the before:
image
and here is the after:
image
Note that callables like `Message` and `X` used be highlighted as built-ins while others like `MResetZ` were not. Likewise, the deprecated keyword `using` was still highlighted as a keyword even though it is no longer supported. After the update, callables all are highlighted the same way as other local and global symbols, and `using` is no longer highlighted as a keyword.

One other note: the syntax highlighting for `w/` or copy-update never worked because it seems TextMate grammar does not support highlights that include special characters like that. I tried finding any supporting documentation that would point to a way to do it and only found docs that say "use regular expression syntax." I tried several variations of JavaScript regex syntax including specific unicode identifiers, and nothing worked. Given that, I just removed `w/` from the list.


minestarks:
Sadly we have some duplication of our syntax highlighting rules... do you think you could update the Codemirror definitions as well, please?

https://github.com/microsoft/qsharp/blob/main/jupyterlab/src/index.ts#L32
https://github.com/microsoft/qsharp/blob/main/pip/qsharp/.data/qsharp_codemirror.js#L4

vscode/syntaxes/qsharp.tmLanguage.json
```diff
@@ -42,11 +39,11 @@
       "patterns": [
         {
           "name": "keyword.control.qsharp",
-          "match": "\\b(use|using|borrow|borrowing|mutable|let|set|if|elif|else|repeat|until|fixup|for|in|while|return|fail|within|apply)\\b"
+          "match": "\\b(use|borrow|mutable|let|set|if|elif|else|repeat|until|fixup|for|in|while|return|fail|within|apply)\\b"
         },
         {
           "name": "keyword.other.qsharp",
-          "match": "\\b(new|not|and|or|w/)\\b"
+          "match": "\\b(not|and|or)\\b"
```
minestarks, line 46: `\b` is a word boundary, which is why the regex will never match the "boundary" between the `/` and the following whitespace. Try:

```suggestion
          "match": "\\b(not|and|or)\\b|\\b(w/)"
```

Note this also matches the `w/` in something like `w/i`, but as far as I can tell, that is valid Q# (parsed as two separate tokens, `w/` and `i`). If we ever wanted to disallow adjacent identifiers like that, we could put a whitespace character in the regex.

swernli: Ooo, that does work! But oddly, doesn't seem to work for `w/=` where only the `w/` gets highlighted despite adding `|\\b(w/=)` to the string. That said, `=` is never highlighted when it's part of a combined operation and assignment, so in `or=` only the `or` is highlighted. So maybe just highlighting the `w/` is sufficient and consistent?

swernli: Ok, with the latest update it will highlight both the `w/` and the `<-` as keywords:
image

swernli:
> Sadly we have some duplication of our syntax highlighting rules... do you think you could update the Codemirror definitions as well, please?
> 
> https://github.com/microsoft/qsharp/blob/main/jupyterlab/src/index.ts#L32 https://github.com/microsoft/qsharp/blob/main/pip/qsharp/.data/qsharp_codemirror.js#L4

Good call out, didn't realize we had those too!

swernli:
@minestarks I made the changes to the other files, but I'm not sure of how to test them. I tried running Jupyter and JupyterLab locally but the language syntax highlighting options were grayed out. It could be that I'm missing something in my setup though.

filipw:
Last year I have started a project for Q# TM grammar here https://github.com/qsharp-community/qsharp-tmLanguage. It comes with tests and a build process and such.

The latest generated file is here https://github.com/qsharp-community/qsharp-tmLanguage/blob/main/grammars/qsharp.tmLanguage.json

It's still not complete but it offers a lot of improvements over the built-in one. If you are interested we could perhaps include that one instead (my ambition was to actually add it to VS Code too, as it ships a number of prebuilt TM files in the box, as long as they have their own repos and proper build scripts)

filipw:
by the way, here is your example using the grammar I linked to (in Monokai Theme):

Bildschirmfoto 2023-10-05 um 09 49 46

vscode/syntaxes/qsharp.tmLanguage.json
minestarks: Wait, but if we're marking `<-` as a keyword, why not all the other operators? Why not, say, `->` or even `==` at that point?

Also, now that I'm checking the docs, `keyword.operator` seems to be the more appropriate scope for these. https://github.com/dunstontc/textmate/blob/master/scopes.md#keyword

jupyterlab/src/index.ts
```diff
@@ -60,7 +60,7 @@ function registerQSharpLanguage(
     },
     {
       token: "keyword",
-      regex: String.raw`(let|set|w\/|new|not|and|or|use|borrow|using|borrowing|newtype|mutable)\b`,
+      regex: String.raw`(let|set|w\/|\<-|not|and|or|use|borrow|mutable)\b`,
```
minestarks, line 63: This still has the same word boundary issue that you already fixed in the tmLanguage file.

vscode/syntaxes/qsharp.tmLanguage.json
swernli: Honestly, I think we should, especially with the further distinction as `keyword.operator`. I'll look through the parser and try to round up all the operators we support.

swernli:
> Last year I have started a project for Q# TM grammar here https://github.com/qsharp-community/qsharp-tmLanguage. It comes with tests and a build process and such.
> 
> The latest generated file is here https://github.com/qsharp-community/qsharp-tmLanguage/blob/main/grammars/qsharp.tmLanguage.json
> 
> It's still not complete but it offers a lot of improvements over the built-in one. If you are interested we could perhaps include that one instead (my ambition was to actually add it to VS Code too, as it ships a number of prebuilt TM files in the box, as long as they have their own repos and proper build scripts)

This sounds great, thanks Filip! I think what I'd like to do is at least get this one cleaned up and merged to start, and then we can coordinate on how to make use of your infrastructure checked into the qsharp-community org. Looks like there is some good work you've done there that the ecosystem could benefit from!

vscode/syntaxes/qsharp.tmLanguage.json
swernli: Ok, I've added more operators and they seem to be doing the right thing. I still need to add a few more, specifically some of the triple symbol binary operators, and then I should have a full set.

jupyterlab/src/index.ts
swernli: Made some tweaks that I think address this, will be adding more operators soon...

swernli (commented):
I think this now ready for re-review.

minestarks:
@swernli you'll need to merge from `main` for this PR to pass checks. I just changed the names of the checks in ci.yml #625 which is causing issues with currently open PRs. Sorry about that!

minestarks (approved):

PR #771, opened 2023-10-05
Language Server: Rename

ScottCarda-MS:
Support for renaming local variables, parameters, callables, UDT's, and UDT fields. Does not support renaming of elements defined in packages other than the current package, such as STD callables.


playground/src/tsconfig.json
minestarks, line 1: Nice, I had the same change in #732 .

language_service/src/rename.rs
```diff
@@ -0,0 +1,375 @@
...
+    let package = &compilation.unit.ast.package;
+
+    let mut prepare_rename = Rename::new(compilation, offset, true);
+    prepare_rename.visit_package(package);
+    prepare_rename.prepare.map(|p| {
+        (
+            protocol::Span {
+                start: p.0.lo,
+                end: p.0.hi,
+            },
```
minestarks, line 39: I think it's worth creating a tiny helper function for this (and the same pattern below in the lambda inside `get_rename`). 

Also, do we map `SourceMap` offsets to source offsets somewhere, like [`protocol_span` in hover.rs](https://github.com/microsoft/qsharp/blob/598b939260c6da90e0d8b66ffda35bd94c51aee2/language_service/src/hover.rs#L310C16-L310C16)?

language_service/src/rename.rs
```diff
@@ -0,0 +1,375 @@
...
+                    locations: vec![],
+                };
+                rename.visit_callable_decl(curr);
+                self.locations = rename.locations;
+            }
+        }
+    }
+}
+
+fn span_contains(span: Span, offset: u32) -> bool {
```
minestarks, line 158: Don't we already have this in qsc_utils.rs?

language_service/src/rename.rs
```diff
@@ -0,0 +1,375 @@
...
+    offset >= span.lo && offset <= span.hi
+}
+
+fn ast_item_id_to_hir_item_id(
```
minestarks, line 162: This seems like a pretty general-purpose function, and indeed I see the same logic in hover.rs, definition.rs and signature_help.rs. I think now is a good refactoring opportunity: it's worth putting this in qsc_utils.rs, and going back to those other implementations and refactoring them to use this function.

I think by refactoring a little bit at a time like this, we'll start to see some more common patterns emerge between these features which should make them all easier to reason about.

language_service/src/rename.rs
```diff
@@ -0,0 +1,375 @@
...
+                        walk_callable_decl(self, decl);
+                        self.current_callable = context;
+                    }
+                    // Note: the `item.span` can cover things like doc
+                    // comment, attributes, and visibility keywords, which aren't
+                    // things we want to have logic for, while the `decl.span` is
+                    // specific to the contents of the callable decl, which we do want
+                    // logic for. If the `if` or `else if` above is not met, then
+                    // the cursor is at one of these non-decl parts of the item,
+                    // and we want to do nothing.
```
minestarks, line 197: Oh, this comment is giving me some real deja-vu. Doesn't have to be this PR, but I really do think we should look at refactoring the visitor between definition, hover, and rename. (I did agree with your offline point about not all language features necessarily fitting in the same mold. But I think these three features are siblings, in a way, because they all deal with locating definitions and references. Find-All-References will be another one that's going to look very similar).

language_service/src/rename/tests.rs
minestarks, line 1: I didn't find a test case for it, but can we reject a rename that contains non-identifier characters (whitespace, punctutation etc)? I don't know how other languages handle this, but as a user I'd expect the editor to prevent me from doing this.

minestarks (approved):
Looks great, mostly a few requests for refactoring.

language_service/src/rename/tests.rs
ScottCarda-MS: added

language_service/src/rename.rs
ScottCarda-MS: This one is slightly different in that it includes the `span.hi` location. I've added a comment clarifying this distinction to the code.

language_service/src/rename.rs
ScottCarda-MS: I hadn't thought to do the source mapping for this because there's no other files to test against yet, but it is a good idea to add it now to make it future compatible with multi-file. I've added a helper to the utils that does this mapping.

language_service/src/lib.rs
```diff
@@ -197,6 +198,36 @@ impl<'a> LanguageService<'a> {
         res
     }
 
+    /// # Panics
+    ///
+    /// This function will panic if compiler state is invalid or in out-of-memory conditions.
```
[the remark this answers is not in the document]
ScottCarda-MS: Looks like it can be removed. Nice!

language_service/src/rename.rs
ScottCarda-MS: Yah it is tough because all three are doing a similar step in that they need to find where the user cursor is located in the AST. The problem is that there is not a way to generalize over locations like that, and each of those features need to grab different information from those locations and transform them in different ways. Maybe this could be accomplished by using a set of callback functions, but that seems like a lot of callback functions. Maybe this could be accomplished with inheritance? Not sure right now.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -32,6 +32,7 @@ const PRELUDE: &[&str] = &[
 ];
 
 // All AST Path nodes get mapped
+// Ident Ids from Callable and Ty Items are mapped
```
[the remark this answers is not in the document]
ScottCarda-MS: Removing this comment for now to avoid confusion.

language_service/src/rename.rs
ScottCarda-MS: Discussed offline: decided to add a `source_touches` function to the utils for the difference cases involved.

language_service/src/rename.rs
ScottCarda-MS: Discussed offline: decided to remove this function and inline its logic.

language_service/src/rename.rs
ScottCarda-MS: Discussed offline: design for new component discussed.

billti (approved):
Thanks!

PR #781, opened 2023-10-10
Fix wasm panic at 'ExprId should have been lowered'

### LINKED ISSUE #777: Wasm panic occurred: panicked at 'ExprId should have been lowered'
Run the following code in the playground and observe the JavaScript console log:

```qsharp
namespace MyQuantumProgram {
    @EntryPoint()
    operation Main() : Result[] {
        use q = Qubit();
        CNOT(q,q);
        return [];
    }
}
```

Playground link: https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACk2LQQrCMBAA73nFHlPwBYogiMdqK95KD6ssEjCbZrN7KNK%2FG1TEOc4wjJHKhDeCdu4NWS12ku6CEZ7OQWV3YJW5S4HVN2%2BTJhLUkBhaDOwbWMOZij10GOsEX6wQZNhCb9dQz80v7I%2Bni8%2Br%2FKeE1IRhGD9qcYt7Af5w7yKZAAAA

```
[wasm] Wasm panic occurred: panicked at 'ExprId should have been lowered', compiler/qsc/src/interpret/stateful.rs:95:14

Stack:

Error
    at imports.wbg.__wbg_new_0232637a3cb0b1a7 (https://microsoft.github.io/qsharp/libs/compiler-worker.js:682:19)
    at wasm://wasm/00759a0e:wasm-function[2263]:0x183986
    at wasm://wasm/00759a0e:wasm-function[1140]:0x14882d
    at wasm://wasm/00759a0e:wasm-function[1823]:0x16fbdb
    at wasm://wasm/00759a0e:wasm-function[1644]:0x168cab
    at wasm://wasm/00759a0e:wasm-function[1392]:0x15db75
    at wasm://wasm/00759a0e:wasm-function[147]:0x78867
    at wasm://wasm/00759a0e:wasm-function[1352]:0x158daf
    at Object.run (https://microsoft.github.io/qsharp/libs/compiler-worker.js:260:12)
    at Compiler.run (https://microsoft.github.io/qsharp/libs/compiler-worker.js:1187:17)
```

Git bisect narrowed this down to #756 (cc @swernli)

swernli:
When executing the entry expression for a compilation in the stateful interpreter, it should be treated the same way as running shots for Python or executing for code generation: use a fresh execution state and environment. State reuse is only meaningful in steppable debugging, and environment reuse is only for fragment execution in the contex of a REPL or Python/Jupyter. The fix used here is very small (just the change in stateful.rs), the rest is introducing test infra and cases for the wasm crate.

Fixes #777


wasm/src/tests.rs
minestarks, line 1: Code organization: if you're creating a new test file, perhaps you could move the tests that live inside lib.rs here as well?

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -231,10 +231,10 @@ impl Interpreter {
         };
 
         eval_expr(
```
minestarks, line 233: At this point, isn't `eval_entry` almost functionally identical to `run()`? I wonder if it's worth consolidating those to functions into one as well? Especially since the wasm layer even has its own "run with shots" implementation.

That being said I understand if you want to keep this PR scoped only to this fix.

minestarks (approved):

compiler/qsc/src/interpret/stateful.rs
swernli: I think you are right, we are very close to eliminating this function in favor of using `run`... I'll see if this PR can accomodate it and if it gets too messy I can move that refactor to a follow up.

wasm/src/tests.rs
swernli: I had forgotten this crate didn't follow the same pattern as the others, good reminder! I'll move these tests here and drop the old ones.

compiler/qsc/src/interpret/stateful.rs
swernli: Unfortunately it does end up touching more than I'd like for inclusion here, so I'll do a separate refactor.

cesarzc (approved):

PR #784, opened 2023-10-11
Update to Rust 1.73

swernli:
This change updates the build scripts and yaml to use Rust 1.73, which introduced some new clippy lints that the code was in violation of. The changes have the version update and the code updates for compatibility, and to keep the new clippy lints in check the minimum required version is updated in prereqs.py.

This also updates the devcontainer to pin Python 3.11 because latest is now pointing to 3.12 which fails to build during container creation.


compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -125,7 +125,7 @@ fn matches_target(attrs: &[Box<Attr>], target: TargetProfile) -> bool {
                     ExprKind::Path(path) => {
                         match TargetProfile::from_str(path.name.name.as_ref()) {
                             Ok(t) => t == target,
-                            Err(_) => true,
+                            Err(()) => true,
```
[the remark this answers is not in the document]
swernli: Me too... it's definitely pedantic to the max, but not without reason: https://rust-lang.github.io/rust-clippy/master/index.html#/ignored_unit_patterns.

prereqs.py
```diff
@@ -13,13 +13,13 @@
 import functools
 
 python_ver = (3, 11)  # Python support for Windows on ARM64 requires v3.11 or later
-rust_ver = (1, 72)  # Ensure Rust version 1.69 or later is installed
+rust_ver = (1, 73)  # Ensure Rust version 1.69 or later is installed
```
billti, line 16: Probably should update the comment on the line also.

.devcontainer/devcontainer.json
```diff
@@ -11,7 +11,11 @@
 		"ghcr.io/devcontainers/features/python:1": {
 			"installTools": true,
 			"installJupyterlab": true,
-			"version": "latest"
+			"version": "3.11"
+		},
+		"ghcr.io/devcontainers/features/azure-cli:1": {
+			"version": "latest",
+			"extensions": "quantum"
```
billti, line 18: Why is this here? The azure-cli is pretty huge. Does anything depend on this? Or just for the convenience of anyone spinning up a container who wants to run `az login`?

billti (approved):

prereqs.py
swernli: Done

.devcontainer/devcontainer.json
swernli: Yeah, for login and job monitoring. Based on local testing it makes the container around 800MB bigger, maybe a bit more. I thought it would be useful to have by default, but if we don't think it's central enough to our workflow I can take it out.

swernli: maybe it would be worth commenting it out so folks can see this is how you can add it to a devcontainer definition?

billti: It's fine. I guess you need it (currently) for auth in Python.

idavis (approved):

PR #786, opened 2023-10-12
Added remaining functions to math namespace

DmitryVasilevsky:
- Ported Min, Max, DivRemL, ContinuedFractionConvergentL, BitSizeL, ApproximateFactorial, LogGammaD, LogFactorialD, Binom, SquaredNorm, PNorm, PNormalized, SmallestFixedPoint, LargestFixedPoint.
- IsCoprimeL, IsCoprimeL are now implemented via GreatestCommonDivisor rather than ExtendedGreatestCommonDivisor.
- RealMod is now implemented via simple truncation of a ratio rather than sin, cos and atan functions. Unlike previous implementation it passes the case that is mentioned in documentation (RealMod(3.6, 2.4, -1.2) = -1.2, BTW it wasn't among test cases in the old QDK).
- FactorialI has slightly simpler implementation.
- FactorialL has classical implementation. Previous recursive decomposision was complex but didn't gain anything.
- Added new function DivRemI to match DivRemL.
- Added tests.
- Replaced TeX with unicode.


library/std/math.qs
```diff
@@ -2,8 +2,49 @@
...
+    }
+
+    /// # Summary
+    /// Returns the natural logarithm of 2.
+    ///
+    /// # Output
+    /// Returns a `Double` equal to 0.6931471805599453.
+    function LogOf2 () : Double
+    {
+        return 0.6931471805599453;
```
cesarzc, line 45:
```suggestion
        0.6931471805599453
```

cesarzc: nit: since in general in the standard library we are now using implicit returns.

cesarzc: This applies to other functions as well.

library/std/math.qs
```diff
@@ -85,36 +126,52 @@ namespace Microsoft.Quantum.Math {
...
+    /// # Output
+    /// The largest element of `values`.
+    function Max (values : Int[]) : Int {
+        Fact(Length(values) > 0, "Array must contain at least one element.");
+        mutable max = values[0];
+        for idx in 1 .. Length(values) - 1 {
+            if values[idx] > max {
+                set max = values[idx];
+            }
+        }
```
cesarzc, line 145:
```suggestion
        for element in values[1...] {
            if element > max {
                set max = element;
            }
        }
```

cesarzc: This can be done without the need to use indexes through slices.

library/std/math.qs
```diff
@@ -85,36 +126,52 @@ namespace Microsoft.Quantum.Math {
...
+    /// # Output
+    /// The smallest element of `values`.
+    function Min (values : Int[]) : Int {
+        Fact(Length(values) > 0, "Array must contain at least one element.");
+        mutable min = values[0];
+        for idx in 1 .. Length(values) - 1 {
+            if values[idx] < min {
+                set min = values[idx];
+            }
+        }
```
cesarzc, line 166:
```suggestion
        for element in values[1...] {
            if element < min {
                set min = element;
            }
        }
```

cesarzc: Similar to my other comment.

library/std/math.qs
DmitryVasilevsky: Removed all unnecessary return statements.

library/std/math.qs
DmitryVasilevsky: Neat!

library/std/math.qs
```diff
@@ -479,9 +587,67 @@ namespace Microsoft.Quantum.Math {
...
+    /// Finds the continued fraction convergent closest to `fraction`
+    /// with the denominator less or equal to `denominatorBound`
+    /// Using process similar to this: https://nrich.maths.org/1397
+    function ContinuedFractionConvergentL(
+        fraction : (BigInt, BigInt),
+        denominatorBound : BigInt) : (BigInt, BigInt) {
+        Fact(denominatorBound > 0L, "Denominator bound must be positive");
+
+        let (a, b) = fraction;
+        let signA = IntAsBigInt(SignL(a));
```
billti, line 607: Why is this necessary. a and b are already BigInt. Does SignL not return a BigInt?

library/std/math.qs
```diff
@@ -499,7 +665,288 @@ namespace Microsoft.Quantum.Math {
...
+            set result *= i;
+        }
+        result
+    }
+
+    /// # Summary
+    /// Returns the factorial of a given number.
+    ///
+    /// # Description
+    /// Returns the factorial of a given nonnegative integer n, where 0 ≤ n ≤ 20.
```
billti, line 725: Cut & paste error? I thought the L version was good for n > 20.

library/std/math.qs
```diff
@@ -499,7 +665,288 @@ namespace Microsoft.Quantum.Math {
...
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Math.FactorialI
+    /// - Microsoft.Quantum.Math.FactorialL
+    function ApproximateFactorial(n : Int) : Double {
+        Fact(n >= 0, "The factorial is not defined for negative inputs.");
+        Fact(n <= 169, "The largest approximate factorial that be stored as an Double is 169!. Use FactorialL.");
+
+        // For small enough n, use the exact factorial instead.
+        if n < 10 {
```
billti, line 773: Why not 20 if that's what FactorialI supports?

library/std/math.qs
DmitryVasilevsky: SignL returns Int, not BigInt. We may consider changing it, but IntAsBigInt is required until then.

library/std/math.qs
DmitryVasilevsky: Cut & paste error indeed.

cesarzc (approved):

library/std/math.qs
DmitryVasilevsky: I updated FactorialI implementation and changed this function to use it when possible.
I think that the point was to use fewer operations if the result can be approximate. I changed factorial to use precomputed array of 21 element so that the execution time is constant. Now this function can use FactorialI whenever possible.

PR #787, opened 2023-10-12
Error squiggle span fixes

### LINKED ISSUE #750: Runtime Error underline is misplaced in the playground editor
Use this code:

https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACk1OMW7DMAzc9YoDvDiLvSco0KFFJwNx26ErY7MJgUhyKApFUOTvlV036IELj3fHa9sWFd7IT2d2bVle2EzCEclIjceZW%2FgKT5wGlckkhoV5P0lCGYKXIJ7O6CtMGo9KHnYiw0ABB0ZOPMLibyS%2BVJYHRTzEkRsXyHOaaGB01z5TsOz3a8q3Q0GcOKCTQWOKn9asmqZjSlnZc7CdW4SPz8H0uo8SrN78WZXmxuhIQr3BFq%2Bc8tnW6BmlHi54QJ8PUny7%2B%2BGjvvzblC1rybmTN3dzP4C0HAc%2FAQAA

Run once.
You'll see the red squiggly not underlining the first 'u' but rather underlining last semicolon.
Start editing text. For example, start commenting out 'open' statement.
You'll see red squiggly changing its position.

I expect the red squiggly to stay in the right place. Alternatively, it may disappear as I edit the text.

minestarks:
Fixes #750 . Addresses some of #149 .

Runtime error squiggles in the playground were off by one. The root cause was that we actually use `SourceMap` offsets, not `Source` offsets, to report the error. This sometimes used to "just work" since we always have a single file in the compilation. However, this really broke down when #614 inserted an additional EOF character between sources... and then when #622 moved the entry expression (`""` in this case) to the _beginning_ of the source map. 

Moreover, when runtime error spans came from stdlib, the offsets were just wildly off (since the spans were actually coming from stdlib). This probably wasn't noticeable since the UI "covered up" the problem by mapping the squiggles to the very end of the document.

So, the bulk of this PR is properly resolving `SourceMap` offsets to `Source` offsets in the wasm layer. When the span doesn't belong in the current document, we just show a squiggle at the beginning of the document. But we also take advantage of the "related information" feature of Monaco to still include the relevant stdlib labels in the hover text (see screenshots below).

This brings spans to the realm of "technically correct", but I'm still not happy with the experience in the playground UI for runtime errors. For a better experienece, we should consider:

- Either, grabbing spans from the call stack and squiggle the statements in the stack.
- Or, not bothering with runtime error squiggles at all....

However, this change was already getting too big, and the callstack changes would have been pretty complicated. I decided to just go with this as an incremental improvement, and debate the rest later.

### Playground runtime error changes

**Case 1: The runtime error has a label that belongs in the current file**

This had the off-by-one error, now fixed.

**Before**

image

**After**

image

**Case 2: The runtime error has a label that belongs in stdlib**

This used to map to the end of the file (just by luck), now it maps to the beginning of the file and shows the label from the original location in the hover text.

**Before**

image

**After**

image

**Case 3: The runtime error has a label that belongs in stdlib, _and_ the label doesn't contain a message**

This is the worst experience - we could show the stdlib location in the hover, but there's no message in the label so it would be entirely pointless to display. So we omit it. The location of the squiggle is now technically correct but just as unhelpful.

**Before**
image

**After**
image

### Other goodness:

- In the playground, we clear the last runtime squiggle when the user starts typing (or click on a different sample).

- Properly reporting compiler errors with multiple spans (e.g. [`Ambiguous`](https://github.com/microsoft/qsharp/blob/c75940913518763000a3b18b053ef683cc0e3ad7/compiler/qsc_frontend/src/resolve.rs#L57)) using VSCode/Monaco's "related information" feature. These related spans show up in the error hover.

image

- `qsc::compile::compile()` now just returns a `WithSource<Error>` instead of relying on the caller to use the `SourceMap` to resolve error spans to source files. Every caller has to do this mapping at _some_ point to report errors, so it makes sense for the errors to just contain that information from the start. Using `WithSource` as the error type also mirrors the error types that `Interpreter` and `Incremental` already return.

- Somehow `VSDiagnostic` was defined in two places. One with the name `IDiagnostic` and the other `VSDiagnostic`. They describe the same thing. Deleted one of the definitions.

### DIFF
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -90,13 +90,8 @@ pub fn get_library_source_content(name: &str) -> Option<String> {
 
 #[wasm_bindgen]
 pub fn get_hir(code: &str) -> String {
-    let package = compile(code);
-    package.to_string()
-}
-
-fn compile(code: &str) -> qsc::hir::Package {
-    STORE_CORE_STD.with(|(store, std)| {
-        let sources = SourceMap::new([("code".into(), code.into())], None);
+    let sources = SourceMap::new([("code".into(), code.into())], None);
+    let package = STORE_CORE_STD.with(|(store, std)| {
         let (unit, _) = compile::compile(
             store,
             &[*std],
@@ -105,7 +100,8 @@ fn compile(code: &str) -> qsc::hir::Package {
             TargetProfile::Full,
         );
         unit.package
-    })
+    });
+    package.to_string()
 }
 
 struct CallbackReceiver<F>
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use crate::language_service::VSDiagnostic;
+use diagnostic::VSDiagnostic;
 use katas::check_solution;
 use num_bigint::BigUint;
 use num_complex::Complex64;
@@ -20,6 +20,7 @@ use std::fmt::Write;
 use wasm_bindgen::prelude::*;
 
 mod debug_service;
+mod diagnostic;
 mod language_service;
 mod logging;
 mod serializable_type;
@@ -89,24 +90,21 @@ pub fn get_library_source_content(name: &str) -> Option<String> {
 
 #[wasm_bindgen]
 pub fn get_hir(code: &str) -> String {
-    let (package, _) = compile(code);
+    let package = compile(code);
     package.to_string()
 }
 
-fn compile(code: &str) -> (qsc::hir::Package, Vec<VSDiagnostic>) {
+fn compile(code: &str) -> qsc::hir::Package {
     STORE_CORE_STD.with(|(store, std)| {
         let sources = SourceMap::new([("code".into(), code.into())], None);
-        let (unit, errors) = compile::compile(
+        let (unit, _) = compile::compile(
             store,
             &[*std],
             sources,
             PackageType::Exe,
             TargetProfile::Full,
         );
-        (
-            unit.package,
-            errors.into_iter().map(|error| (&error).into()).collect(),
-        )
+        unit.package
     })
 }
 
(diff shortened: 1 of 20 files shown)


idavis (approved):

wasm/src/lib.rs
```diff
@@ -89,24 +90,21 @@ pub fn get_library_source_content(name: &str) -> Option<String> {
 
 #[wasm_bindgen]
 pub fn get_hir(code: &str) -> String {
-    let (package, _) = compile(code);
+    let package = compile(code);
```
minestarks, line 93: I'm going to inline `compile()` and delete the function definition since it's now only called from `get_hir()` and is not as useful as the name makes it sound.

swernli (approved):
Changes in WASM crate look good, the inlining you suggest is a good nice-to-have too.

ScottCarda-MS (approved):

wasm/src/lib.rs
```diff
@@ -89,25 +90,18 @@ pub fn get_library_source_content(name: &str) -> Option<String> {
...
-    package.to_string()
-}
-
-fn compile(code: &str) -> (qsc::hir::Package, Vec<VSDiagnostic>) {
-    STORE_CORE_STD.with(|(store, std)| {
-        let sources = SourceMap::new([("code".into(), code.into())], None);
-        let (unit, errors) = compile::compile(
+    let sources = SourceMap::new([("code".into(), code.into())], None);
+    let package = STORE_CORE_STD.with(|(store, std)| {
+        let (unit, _) = compile::compile(
```
cesarzc, line 95: I'm curious, what was the reason for dropping errors here?

minestarks: The only (remaining) purpose of this function is to populate the HIR view in the playground . We don't need the compilation errors when formatting the HIR - they're already shown in the UI via the editor.

image

cesarzc (approved):

PR #788, opened 2023-10-13
Add attributes to completion list

minestarks:


ScottCarda-MS (approved):

idavis (approved):

vscode/src/extension.ts
```diff
@@ -151,7 +151,7 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     vscode.languages.registerCompletionItemProvider(
       qsharpDocumentFilter,
       createCompletionItemProvider(languageService),
-      "."
```
billti, line 154: Do you not still want . also?

billti (approved):

vscode/src/extension.ts
minestarks: The only case we'd want to trigger on `.` is if you typed a complete namespace and needed an item from that namespace, like:

```
Microsoft.Quantum.Diagnostics.DumpMachine
                             ^
```

But sadly this type of completion isn't currently supported... I figured I would add this and other trigger characters like `:` when proper support is implemented.

PR #792, opened 2023-10-14
Command to create a sample notebook in the workspace

billti:
Some refactoring to add the workspace connection info if connected.


vscode/src/notebook.ts
```diff
@@ -63,3 +65,114 @@ export function registerQSharpNotebookHandlers() {
...
+        );
+        const template = await vscode.workspace.fs.readFile(templatePath);
+        let content = new TextDecoder().decode(template);
+
+        // Update the workspace connection info in the notebook if workspaces are already connected to
+        const workspaces =
+          WorkspaceTreeProvider.instance?.getWorkspaceIds() || [];
+        let choice = workspaces[0] || undefined;
+        if (workspaces.length > 1) {
+          choice = await vscode.window.showQuickPick(workspaces);
```
billti, line 147: Oops. Should of put a title on this quick pick. Will add it next push.

vscode/resources/notebookTemplate.ipynb
```diff
@@ -0,0 +1,237 @@
...
+        "To use Jupyter Notebooks with Q#, the `qsharp-lang` and `jupyterlab` Python packages should be installed.\n",
+        "\n",
+        "To render charts as shown in this notebook, Matplotlib with notebook widgets support should be installed via `ipympl`.\n",
+        "\n",
+        "To submit to Azure Quantum as shown in this notebook, the `azure-quantum` package needs to be installed.\n",
+        "\n",
+        "You can install all the above packages in your Python environment by running the below in your terminal:\n",
+        "\n",
+        "```bash\n",
+        "pip install jupyterlab qsharp-lang ipympl azure-quantum\n",
```
minestarks, line 19: I don't know how necessary this is, but this can actually be a code cell that uses the `%pip install` magic so that it can be executed in the notebook. Though... if we install all the packages at once like this, the output ends up being pretty unreadable in the notebook... so maybe the bash script is a better experience after all.

vscode/src/notebook.ts
```diff
@@ -63,3 +65,114 @@ export function registerQSharpNotebookHandlers() {
...
+            }
+          );
+          if (!choice) {
+            // User cancelled
+            return;
+          }
+          workspaceFolder = choice.workspace;
+        }
+
+        let notebookName = await vscode.window.showInputBox({
```
minestarks, line 105: It seems like a better experience to use `workspace.openNotebookDocument` to create an untitled notebook instead of having the user pick a filename before they can see anything. Saves us from having to do file name validation, etc. Seems more aligned with the "create: new file" or "create: new jupyter notebook" commands, too.

Implementation of the latter is here for reference: https://github.com/microsoft/vscode/blob/482d5ba393c2014c40535ec1b39ec743af3ffecc/extensions/ipynb/src/ipynbMain.ts#L73-L85

minestarks (approved):

vscode/src/notebook.ts
```diff
@@ -63,3 +65,114 @@ export function registerQSharpNotebookHandlers() {
...
+        const template = await vscode.workspace.fs.readFile(templatePath);
+        let content = new TextDecoder().decode(template);
+
+        // Update the workspace connection info in the notebook if workspaces are already connected to
+        const workspaces =
+          WorkspaceTreeProvider.instance?.getWorkspaceIds() || [];
+        let choice = workspaces[0] || undefined;
+        if (workspaces.length > 1) {
+          choice = await vscode.window.showQuickPick(workspaces);
+        }
```
idavis, line 148:
```suggestion
        let choice = undefined;
        if (workspaces.length > 1) {
          choice = await vscode.window.showQuickPick(workspaces) ?? workspaces[0];
        }
```

idavis (commented):
Left a couple of questions and maybe suggestion.

vscode/resources/notebookTemplate.ipynb
billti: Yeah, I didn't want to break them all out into separate pip installs, but didn't want to omit any either. Plus, I forget which but it asks you to restart the kernel afterwards, so figured it was easier to just tell the user that these prereqs should be in the environment (many of them will be if using something like Conda, but wont if using a clean Python install)

vscode/src/notebook.ts
billti: Done. Thanks!

vscode/src/azure/commands.ts
```diff
@@ -33,6 +33,8 @@ const workspacesSecret = "qsharp-vscode.workspaces";
 
 export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
   const workspaceTreeProvider = new WorkspaceTreeProvider();
+  WorkspaceTreeProvider.instance = workspaceTreeProvider;
```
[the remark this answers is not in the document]
billti: Yeah, that or top level exports from the module (effectively a scoped global). This is more typical though.

vscode/src/notebook.ts
billti: It would be null here if they escaped out (i.e. didn't choose a workbook), so would then later just put the placeholder code in instead of a specific workbook. That said, I did notice I was showing the long id here, so updating to show a friendly name.

PR #794, opened 2023-10-16
Introduce Q# project structure

sezna:
# Q# Projects

This PR introduces a Q# manifest file, `qsharp.json`, and a corresponding notion of _Q# projects_.

## Feature Overview
When the compiler is invoked, it looks upward in the directory structure for a manifest file. Upon finding a valid manifest, we set that directory as the root directory of the project. We then include all `*.qs` files in that directory or subdirectories of 
that directory as sources, unless they are explicitly excluded in the manifest. This implementation *does not follow symlinks of any kind*. 

All FS requirements are expressed as the `FileSystem` interface. This will provide compatibility when we integrate projects with the language service and other virtual file systems. `qsc_project` only utilizes `std::fs` if the `fs` feature is enabled.

## API Changes

There are some notable differences in the behavior of `qsc` and `qsi`:
* `qsc` still accepts sources as input, but now, if no sources are specified, `qsc`  will look for a directory in the current directory.
* Currently, `qsc` requires a list of source names on the command line. With this PR, if `qsc` is invoked within a project with
  no sources specified, it will include project sources.
* `qsi` will include sources within its project scope if it is invoked within a project. A flag `--no-init-project`/`-n` 
   has been added to skip this.

## Testing
`qsc_project` is tested using our normal `expect_test` workflow. There are a series of test projects in `qsc_project/tests` that represent various project structures and manifests.  We expect certain directory structures and manifests to deterministically correspond to a set of source names and source contents, and this is the interface that is tested.

### DIFF
--- a/compiler/qsc_project/src/manifest.rs
+++ b/compiler/qsc_project/src/manifest.rs
@@ -1,14 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#[cfg(feature = "fs")]
 use crate::Error;
-use regex_lite::Regex;
-use serde::Deserialize;
-use std::path::PathBuf;
+#[cfg(feature = "fs")]
 use std::{
     env::current_dir,
     fs::{self, DirEntry, FileType},
 };
+
+use regex_lite::Regex;
+use serde::Deserialize;
+use std::path::PathBuf;
 
 pub const MANIFEST_FILE_NAME: &str = "qsharp.json";
 
@@ -31,13 +34,6 @@ pub struct ManifestDescriptor {
 }
 
 impl ManifestDescriptor {
-    pub(crate) fn new(manifest: Manifest, manifest_dir: PathBuf) -> Self {
-        Self {
-            manifest,
-            manifest_dir,
-        }
-    }
-
     pub(crate) fn exclude_regexes(&self) -> Result<Vec<Regex>, crate::Error> {
         self.manifest
             .exclude_regexes
@@ -52,6 +48,7 @@ impl ManifestDescriptor {
     }
 }
 
+#[cfg(feature = "fs")]
 impl Manifest {
     /// Starting from the current directory, traverse ancestors until
     /// a manifest is found.
@@ -80,7 +77,10 @@ impl Manifest {
 
                     let manifest = fs::read_to_string(item.path())?;
                     let manifest = serde_json::from_str(&manifest)?;
-                    return Ok(Some(ManifestDescriptor::new(manifest, manifest_dir)));
+                    return Ok(Some(ManifestDescriptor {
+                        manifest,
+                        manifest_dir,
+                    }));
                 }
             }
         }
@@ -90,6 +90,7 @@ impl Manifest {
 
 /// Utility function which filters out any [DirEntry] which is not a valid file or
 /// was unable to be read.
+#[cfg(feature = "fs")]
 fn only_valid_files(item: std::result::Result<DirEntry, std::io::Error>) -> Option<DirEntry> {
     match item {
         Ok(item)


Cargo.toml
```diff
@@ -45,7 +47,7 @@ num-complex = "0.4"
 num-traits = "0.2.15"
 indenter = "0.2"
 regex-lite = "0.1.0"
-serde = "1.0"
+serde = { version = "1.0", features = ["derive"] }
```
billti, line 50: We only list the dependency in the top level Cargo.toml, add the features in the crate where it's needed (e.g. see the Cargo.toml in the wasm crate where it does what this line does).

Cargo.toml
```diff
@@ -34,6 +35,7 @@ enum-iterator = "1.2.0"
 env_logger = "0.10.0"
 expect-test = "1.4.0"
 getrandom = { version = "0.2" }
+globwalk = "0.8"
```
billti, line 38: Is this of the 'quality' we want to depend on? Looks like it hasn't been updated in 3 years, and has open issues since about then (e.g. https://github.com/Gilnaa/globwalk/issues/29).

compiler/qsc_project/src/manifest.rs
```diff
@@ -0,0 +1,81 @@
+use crate::Error;
+use serde::Deserialize;
+use std::path::PathBuf;
+use std::{
+    env::current_dir,
+    fs::{self, DirEntry, FileType},
```
billti, line 6: Is this crate going to be what understands the qsharp.json and builds up the structure/context for general use by the compiler and language service? If so, it will need to run in wasm and the browser sandbox also. Be sure not to use anything from places like std::env and std::fs which will cause issues in that environment. It should use an interface with the host to do things like walk directories, read files, get the current dir/workspace, etc.

compiler/qsc_project/src/manifest.rs
```diff
@@ -0,0 +1,81 @@
...
+};
+
+pub const MANIFEST_FILE_NAME: &str = "qsharp.json";
+
+#[derive(Deserialize, Debug, Default)]
+pub struct Manifest {
+    pub author: Option<String>,
+    pub license: Option<String>,
+    #[serde(default)]
+    pub exclude_files: Vec<String>,
```
minestarks, line 16: We do want the ability to include files from anywhere on the filesystem, right? (At least, that's what I recall from the group discussion - hope I didn't get mixed up). In that case we probably need an include pattern as well.

compiler/qsc/src/bin/qsc.rs
```diff
@@ -74,12 +75,21 @@ fn main() -> miette::Result<ExitCode> {
...
         .map(read_source)
         .collect::<miette::Result<Vec<_>>>()?;
 
+    let fs = StdFs;
+    let manifest = Manifest::load()?;
+    if let Some(manifest) = manifest {
+        let project = fs.load_project(manifest)?;
+        let mut project_sources = project.sources;
+
+        sources.append(&mut project_sources);
```
swernli, line 90: In the description for the PR (which seems to be cut off for some reason?) the expected behavior is that a manifest is loaded only when no input sources are provided, but this code path looks like it will execute no matter what. Should it check first to ensure the `sources` vector is empty?

compiler/qsc/src/bin/qsi.rs
```diff
@@ -72,12 +73,20 @@ impl Receiver for TerminalReceiver {
...
         .map(read_source)
         .collect::<miette::Result<Vec<_>>>()?;
 
+    let fs = StdFs;
+    let manifest = Manifest::load()?;
+    if let Some(manifest) = manifest {
+        let project = fs.load_project(manifest)?;
+        let mut project_sources = project.sources;
+
+        sources.append(&mut project_sources);
```
swernli, line 88: Same here, looks like sources from the project manifest are appended regardless of whether other sources were provided as input.

compiler/qsc_project/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! This module handles the logic that constitutes the Q# project system.
```
swernli, line 1: This and other new files must have the boilerplate Microsoft copyright comment at the top, ie:
```rust
```

compiler/qsc_project/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+mod error;
+#[cfg(feature = "fs")]
+mod fs;
+mod manifest;
+mod project;
+
+pub use error::Error;
+#[cfg(feature = "fs")]
+pub use fs::StdFs;
+pub use manifest::{Manifest, ManifestDescriptor, MANIFEST_FILE_NAME};
```
swernli, line 14: Does `MANIFEST_FILE_NAME` need to be pub from the module or is it an internal detail to the crate? Just curious.

compiler/qsc_project/src/manifest.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl ManifestDescriptor {
+    pub(crate) fn new(manifest: Manifest, manifest_dir: PathBuf) -> Self {
+        Self {
+            manifest,
+            manifest_dir,
+        }
+    }
+}
```
swernli, line 34: nitpick: this may just be a style thing, but I don't see the need for a `new` constructor when the fields are all public and the only place it is constructed is inside the crate. The one call location could just create the struct directly, which is what any callers outside the crate would do since the struct is pub and `new` is only pub(crate).

compiler/qsc_project/tests/projects/basic_manifest/Dependency1.qs
```diff
@@ -0,0 +1,5 @@
+namespace Dependency1 {
+	function First() : String {
```
swernli, line 2: nitpick: Tabs?!? We use spaces everywhere else for our indentation.

compiler/qsc_project/tests/projects/exclude_list/qsharp.json
```diff
@@ -0,0 +1,3 @@
+{
+  "exclude_files": [".*Excluded.qs"]
```
swernli, line 2: I'm a little torn on one design element here. On the one hand I like that the regex here can match any part of the path, which is nice for doing the folder-based excludes you show in other tests. On the ohter hand, it seems annoying for the user who just wants to have a flat list of excluded files to have to remember to prepend each file name with `.*` to match the first part of the path. Is there a resonable way we could include the wildcard automatically to lessen the burden for users while still getting the behavior we want?

swernli (commented):
I really like the overall design and clean code, some minor feedback and a couple questions left along the way.

compiler/qsc_project/src/manifest.rs
swernli: I don't feel strongly about it, so it's fine to leave as-is. I think reading through the code I was surprised to see a constructor that "wasn't doing anything" paired with public fields, but it's not something I feel strongly enough to advocate one way or the other.

compiler/qsc_project/tests/projects/basic_manifest/Dependency1.qs
billti: Does clippy or rustfmt not automatically fix that for us?

compiler/qsc_project/tests/projects/exclude_list/qsharp.json
swernli: After discussing this some more and realizing even the `.` in `.qs` needs to be escaped, there isn't an easy way to have one field pull double duty. Instead, let's split this up into two fields, one that takes only file names and one that takes regex. That way the simple and advanced cases are both covered without stepping on each other's toes.

compiler/qsc_project/tests/projects/basic_manifest/Dependency1.qs
swernli: Not in .qs files ;)

compiler/qsc/src/bin/qsc.rs
```diff
@@ -74,12 +75,23 @@ fn main() -> miette::Result<ExitCode> {
...
-    let sources = cli
+    let mut sources = cli
         .sources
         .iter()
         .map(read_source)
         .collect::<miette::Result<Vec<_>>>()?;
 
+    if sources.is_empty() {
+        let fs = StdFs;
+        let manifest = Manifest::load()?;
```
billti, line 86: Typically you can provide the manifest file itself, or the directory containing it, as the argument to the compiler. Is the expectation here that you would always set current directory first and then that's all that is used to try to locate a manifest?

compiler/qsc_project/src/manifest.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl Manifest {
+    /// Starting from the current directory, traverse ancestors until
+    /// a manifest is found.
+    /// Returns an error if there are any filesystem errors, or if
+    /// a manifest file exists but is the wrong format.
+    /// Returns `Ok(None)` if there is no file matching the manifest file
+    /// name.
+    pub fn load() -> std::result::Result<Option<ManifestDescriptor>, Error> {
+        let current_dir = current_dir()?;
```
billti, line 47: What does 'current_dir' even mean in the context of wasm or the browser? Looking at the implementation, it appears to call os_imp::getcwd(), so I'm guess it ain't gonna work (https://doc.rust-lang.org/src/std/env.rs.html#56)

compiler/qsc_project/src/manifest.rs
```diff
@@ -0,0 +1,90 @@
...
+        let ancestors = path.ancestors();
+        for ancestor in ancestors {
+            let listing = ancestor.read_dir()?;
+            for item in listing.into_iter().filter_map(only_valid_files) {
+                if item.file_name().to_str() == Some(MANIFEST_FILE_NAME) {
+                    let mut manifest_dir = item.path();
+                    // pop off the file name itself
+                    manifest_dir.pop();
+
+                    let manifest = fs::read_to_string(item.path())?;
```
billti, line 65: I'm going to go out on a limb here and guess that fs::read_to_string doesn't work in the browser or VS Code.

compiler/qsc_project/src/manifest.rs
billti: This module isn't conditional being the 'fs' feature if I'm reading correctly, so it's always compiled in - is that right?. Why would you want to put a 'default fs-based implemention' in this project if it isn't usable in any of our supported scenarios? (I'm guessing Jupyter Notebooks would usually want to point to the manifest/dir also, and not just use current directory).

I don't think we should have default implementation of what the host should provide in the project, especially if most of our scenarios won't use it.

compiler/qsc_project/src/manifest.rs
billti: But this is always compiled in and part of the public API, yet on wasm this results in calls to this module:

image

..which simply errors.

We shouldn't expose APIs by default we know are just going to unconditiionally fail on our main platforms.

swernli (approved):

PR #798, opened 2023-10-17
Update npm dependencies to latest

minestarks:
Updated all direct npm dependencies to `@latest`.

Observed changes:
- Prettier: now adds trailing commas to all lists. https://github.com/prettier/prettier/pull/11479
- Katas: `marked` now omits `id` attributes for header tags (confirmed benign) https://github.com/markedjs/marked/pull/2890
- Other trivial differences in JS emit through `esbuild`.

I was really hoping this would be a trivial process, but sadly it wasn't. For posterity, here's what the process ended up being:

```js
const { execSync } = require("node:child_process");

function run(command) {
  console.log(`running command ${command}`);
  return execSync(command);
}

const deps = JSON.parse(run(`npm ls --package-lock-only --json`));

// Do these in this order first, otherwise @typescript-eslint/eslint-plugin
// gets really upset about the peer dependency missing
run("npm install @typescript-eslint/parser@latest");
run("npm install @typescript-eslint/eslint-plugin@latest");

// Update all direct deps to @latest
for (const depName of Object.keys(deps.dependencies)) {
  if (!depName.startsWith("qsharp")) {
    run(`npm install ${depName}@latest`);
  }
}

// Update indirect deps in package-lock.json
run("npm update");

// This fixes the workspace names in package-lock.json,
// which for some reason get erased above
run("npm install");

```

Ran `./build.py` to confirm the repo still built, discovered formatting changes.

To reformat with new Prettier defaults:

```bash
npm run prettier:fix
```

Since we pulled a lot of JS tooling updates, I ran a quick script to save the built artifacts so I could diff the JS output:

```bash
#!/bin/bash

set -e

# clean repo
git clean -fdx
rm -rf baseline-compare/
mkdir baseline-compare

# build
./build.py --wasm --npm --vscode --no-check --no-test

# npm
mkdir -p baseline-compare/npm
pushd npm
npm pack --pack-destination .
popd
tar -xzf npm/qsharp-lang-0.0.0.tgz -C baseline-compare/npm

# vsix
npm install -g @vscode/vsce
pushd vscode
vsce package --pre-release
popd
unzip vscode/qsharp-lang-vscode-dev-0.0.0.vsix -d baseline-compare/vscode

# commit baseline
cp package-lock.json baseline-compare/
git add baseline-compare/
git commit -m "Baseline for $(git rev-parse HEAD)"
```

Ran this script first on `main`, then with the updated `package.json` & `package-lock.json`. Used `git` to diff the `baseline-compare` changes. Confirmed they looked harmless.


package.json
```diff
@@ -17,28 +17,28 @@
...
-    "@vscode/debugadapter": "^1.61.0",
-    "@vscode/debugprotocol": "1.61.0",
-    "@vscode/extension-telemetry": "0.8.4",
-    "@vscode/test-web": "^0.0.45",
-    "chai": "^4.3.8",
-    "esbuild": "^0.17.12",
-    "eslint": "^8.39.0",
+    "@types/chai": "^4.3.8",
+    "@types/mocha": "^10.0.2",
+    "@types/node": "^20.8.6",
```
minestarks, line 22: I probably shouldn't have bumped the node.js typings. Will downgrade this to 18 to match the `engines`  requirement.

package.json
```diff
@@ -17,28 +17,28 @@
   },
   "private": true,
   "devDependencies": {
```
minestarks, line 19: @billti  I wonder if we should just be pinning these devDependencies to `latest`, instead of specific major versions. I ended up running `npm install foo@latest` on everything anyway, and we have a lockfile, so the version spec in `package.json` ends up being meaningless. (The exception would be the node.js typings, which we update along with node.js.) Thoughts?

minestarks:
@sezna yeah just formatting changes in the code. I called out everything I could spot in the PR description.

This is blocked on codeowner approval and has already accrued a bunch of merge conflicts :/ 

Will update and resolve

minestarks:
@billti would you mind taking a look? Merge conflicts freshly resolved.

package.json
billti: That seems a little dangerous. I know eslint has some quite significant changes, and I though we stayed on an older version of marked for a reason (I see it went from 4.x to 9.x now - we'll def. want to check the rendering is all looking good still).

What benefit would making all 'latest' construe?

billti (approved):

package.json
minestarks: If all of these said `"latest"`, then instead of running the script I pasted in the PR description, I could just run `npm update` next time.

> I see it went from 4.x to 9.x now - we'll def. want to check the rendering is all looking good still

Yes, exactly, the bulk of my time was spent doing this validation. And also writing the PR description.

Considering that either way, I'm running `npm install foo@latest` on all the dependencies, and validating if something broke, what's the harm in just putting `latest` in package.json? If a major upgrade caused severe issues and we didn't want to take it, we would just pin to a downlevel version.

billti: OK. We can give it a shot here if you like, or just try it out as a follow-up.

minestarks: All these updates are already making me a bit nervous, I think I'd rather get this in and tackle that later! I just wanted to bring it up for discussion - thanks for thinking through it with me.

PR #803, opened 2023-10-19
Conform to QIR Base Profile Spec

### LINKED ISSUE #802: Fully conform to QIR Base Profile specification
Currently the QIR produced from the compiler is technically a "custom" profile as it does not adhere to every element of the [Base Profile specification](https://github.com/qir-alliance/qir-spec/blob/main/specification/under_development/profiles/Base_Profile.md) and depends on post processing (at the moment performed in the service) to finish conforming to the spec. Instead, the frontend compiler should produce compliant code directly, and requires a few tweaks to do so. Specifically,

1. Measurements should be deferred to the end of the program.
2. Qubits should not be reused after measurement.
3. `required_num_qubits` and `required_num_results` attributes should be set to the appropriate counts.
4. `qir_profiles` attribute should be set to `base_profile` instead of custom.

With these changes, further transformation by the service should be unnecessary.

swernli:
This change updates base profile QIR codegen to conform to the specification by following additional requirements as outlined in #802.

Fixes #802.

### DIFF
--- a/library/std/intrinsic.qs
+++ b/library/std/intrinsic.qs
@@ -226,8 +226,40 @@ namespace Microsoft.Quantum.Intrinsic {
     /// ```qsharp
     /// Measure([PauliZ], [qubit]);
     /// ```
+    @Config(Full)
     operation M(qubit : Qubit) : Result {
         __quantum__qis__m__body(qubit)
+    }
+
+    /// # Summary
+    /// Performs a measurement of a single qubit in the
+    /// Pauli _Z_ basis.
+    ///
+    /// # Input
+    /// ## qubit
+    /// Qubit to be measured.
+    ///
+    /// # Output
+    /// `Zero` if the +1 eigenvalue is observed, and `One` if
+    /// the -1 eigenvalue is observed.
+    ///
+    /// # Remarks
+    /// The output result is given by
+    /// the distribution
+    /// $$
+    /// \begin{align}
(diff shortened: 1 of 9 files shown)


idavis (approved):

library/std/intrinsic.qs
```diff
@@ -226,10 +226,42 @@ namespace Microsoft.Quantum.Intrinsic {
...
+    ///         \braket{\psi | 0} \braket{0 | \psi}.
+    /// \end{align}
+    /// $$
+    ///
+    /// Equivalent to:
+    /// ```qsharp
+    /// Measure([PauliZ], [qubit]);
+    /// ```
+    @Config(Base)
+    operation M(qubit : Qubit) : Result {
```
cesarzc, line 261: Why is a different version of `M` needed for base profile? How does it help to defer measurements?

swernli: We had the chance to discuss this, but capturing the output here. The idea here is that because measurement needs to leave the qubit in the measured state, the only way to do that while remaining base profile compliant (defering measurements and avoiding qubit reuse) is to use an additional qubit and entanglement to perform the measurement. This lets us ensure that the program will still be correct when measurements are "pulled down" to the end of the program. We should ensure that as we extend the values allowed by `@Config` we don't force ourselves to duplicate an identical function definitions, which implies some version of additive `@Config` options.

cesarzc (approved):

PR #804, opened 2023-10-19
Update devcontainer for Python 3.12

swernli:
This change updates the devcontainer definition to use a Python base image and install Rust as a feature rather than the other way around. This avoids an issue where the Python feature fails building 3.12 from source, and generally makes perf better for fresh containers since Rust and npm features install quickly. Of note, this also adds `cargo update --dry-run` as a post-build command to ensure the crates.io index is updated on new container creation, preventing an issue where crates would get downgraded on first build due to old index cache.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,19 +2,15 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/rust
 {
 	"name": "qsharp",
-	"image": "mcr.microsoft.com/devcontainers/rust:dev-bullseye",
+	"image": "mcr.microsoft.com/devcontainers/python:3",
 	"features": {
 		"ghcr.io/devcontainers/features/node:1": {
 			"nodeGypDependencies": true,
 			"version": "lts"
 		},
-		"ghcr.io/devcontainers/features/python:1": {
-			"installTools": true,
-			"installJupyterlab": true,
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -2,19 +2,15 @@
...
 		"ghcr.io/devcontainers/features/node:1": {
 			"nodeGypDependencies": true,
 			"version": "lts"
 		},
-		"ghcr.io/devcontainers/features/python:1": {
-			"installTools": true,
-			"installJupyterlab": true,
-			"version": "3.11"
-		}
+		"ghcr.io/devcontainers/features/rust:1": {}
```
minestarks, line 11: Now that you start with a non-Rust base image, will this step not fetch the latest crates.io index as part of bootstrapping Rust? Do we still need `cargo update`?

minestarks (approved):

.devcontainer/devcontainer.json
swernli: Sadly no, from what I observed rustup will install cargo and pull down a matching cached index of crates.io, so it does not come with the latest index by default. If I'm wrong and the behavior I observed was caused by something else, then the `cargo update --dry-run` will be mostly benign and just around for people in weird corner cases, but either way it's a nice to have with low cost.

PR #807, opened 2023-10-21
Expand Telemetry Coverage

sezna:
This PR expands telemetry to provide actionable insights, as opposed to general usage data. All added telemetry serves to either identify performance baselines and subsequent regressions, usage data, or failure rates of various feature.s

For fallible events like user flows and generating QIR, we use a _Start_ and a _Stop_ event. This allows for tracking failures as well as performance. These events carry an `associationId` that allows us to reconstruct user flows and see where and how they terminate.

No telemetry contains PII, or even any user-provided strings at all.

### DIFF
--- a/vscode/src/azure/networkRequests.ts
+++ b/vscode/src/azure/networkRequests.ts
@@ -254,7 +254,7 @@ export async function checkCorsConfig(token: string, quantumUris: QuantumUris) {
   log.debug("Pre-flighted GET request didn't throw, so CORS seems good");
   sendTelemetryEvent(
     EventType.CheckCorsEnd,
-    { correlationId, flowStatus: UserFlowStatus.CompletedSuccessfully },
+    { correlationId, flowStatus: UserFlowStatus.Succeeded },
     {}
   );
 }
--- a/vscode/src/telemetry.ts
+++ b/vscode/src/telemetry.ts
@@ -187,9 +187,9 @@ export enum QsharpDocumentType {
 export enum UserFlowStatus {
   // "Aborted" means the flow was intentionally canceled or left, either by us or the user
   Aborted = "Aborted",
-  CompletedSuccessfully = "CompletedSuccessfully",
+  Succeeded = "Succeeded",
   // "CompletedWithFailure" means something that we can action -- service request failure, exceptions, etc.
-  CompletedWithFailure = "CompletedWithFailure",
+  Failed = "Failed",
 }
 
 export enum DebugEvent {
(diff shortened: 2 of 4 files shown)


vscode/src/extension.ts
```diff
@@ -103,6 +108,16 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
   const subscriptions = [];
   subscriptions.push(
     vscode.workspace.onDidOpenTextDocument((document) => {
+      const documentType = isQsharpDocument(document)
```
billti, line 111: Doesn't this fire for every document opened (not just Q# or Jupyter documents)? (CC @minestarks). Do we really want to send an event for every document opened with a type of 'Other'?

vscode/src/utils.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Guid format such as "00000000-1111-2222-3333-444444444444"
+export function getRandomGuid(): string {
```
minestarks, line 5: I think a simple sequence number (1, 2, 3...) would work here. No need to deal with GUIDs.

The [common properties](https://github.com/Microsoft/vscode-extension-telemetry#common-properties) already include the VSCode session ID. So you don't have to worry about _global_ uniqueness, just uniqueness within a VS Code session.

The extension can be deactivated and reactivated within a VS Code session, so you probably want some sort of ExtensionActivationId to differentiate between activations. For that, I think "start time" could work. (We can - and probably should - include that in `sendTelemetryEvent` as a common property for all events).

Once you have those two in each event, then you don't need to worry about global uniqueness in any of the other properties. A monotonically increasing integer should suffice.

vscode/src/telemetry.ts
```diff
@@ -32,17 +49,161 @@ type EventTypes = {
...
+  Other = "Other",
+}
+
+export enum UserFlowStatus {
+  // "Aborted" means the flow was intentionally canceled or left, either by us or the user
+  Aborted = "Aborted",
+  CompletedSuccessfully = "CompletedSuccessfully",
+  // "CompletedWithFailure" means something that we can action -- service request failure, exceptions, etc.
+  CompletedWithFailure = "CompletedWithFailure",
+}
```
minestarks, line 193: Just my personal taste but `Success` and `Failure` are shorter...

vscode/src/azure/networkRequests.ts
```diff
@@ -27,6 +30,16 @@ export async function azureRequest(
 
     if (!response.ok) {
       log.error("Azure request failed", response);
+      if (associationId) {
```
minestarks, line 33: Do you happen to know if there is some kind of correlation ID that we can send _to_ the service that will help them correlate our events with theirs?

minestarks (approved):

PR #810, opened 2023-10-23
Find Identifier for LS Refactor

ScottCarda-MS:
Generalizes the logic to find the location of the cursor on an identifier for several language server features by adding the `IdentifierLocator` visitor that wraps language server feature structs.


language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,318 @@
...
+        context: &LocatorContext<'package>,
+        decl: &'package ast::CallableDecl,
+    ) {
+    }
+
+    fn at_callable_ref(
+        &mut self,
+        path: &'package ast::Path,
+        item_id: &'package hir::ItemId,
+        item: &'package hir::Item,
```
minestarks, line 24: Hmm, it looks like the `item` is only used to get the containing namespace. Personally I think it would look cleaner to just resolve the namespace in the locator and pass it here. Even though only hover uses it.

language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,318 @@
...
+use crate::qsc_utils::{find_ident, find_item, span_contains, span_touches, Compilation};
+use qsc::ast::visit::{walk_expr, walk_namespace, walk_pat, walk_ty_def, Visitor};
+use qsc::{ast, hir, resolve};
+
+#[allow(unused_variables)]
+pub(crate) trait LocatorAPI<'package> {
+    fn at_callable_def(
+        &mut self,
+        context: &LocatorContext<'package>,
+        decl: &'package ast::CallableDecl,
```
minestarks, line 16: Symmetry:

I like the idea of all the `*_def` methods having the `Ident` node that contains the name as the first parameter (after `self` and `context` of course).

So they would all start with:

```rust
&mut self,
context: &LocatorContext<'package>,
name: &ast::Ident
```

I think that would make a very nice and clear pattern. The first parameter is consistently "the thing that the cursor is on."

So for `at_callable_def`, I think you should get `decl.name` in the locator and pass it in here explicitly. You're getting `decl.name` on all three implementations, so may as well do it in the locator.

```rs
 fn at_callable_def(
        &mut self,
        context: &LocatorContext<'package>,
        name: &'package ast::Ident,
        decl: &'package ast::CallableDecl,
    )
```

`at_field_def` and `at_new_type_def` already follow this pattern so great.

For `at_local_def`, you can reorder the parameters so that the name comes first just like the others:

```rs
    fn at_local_def(
        &mut self,
        context: &LocatorContext<'package>,
        ident: &'package ast::Ident,
        pat: &'package ast::Pat,
    )```

File: language_service/src/identifier_locator.rs [hunk 3]
```diff
@@ -0,0 +1,318 @@
...
+    ) {
+    }
+
+    fn at_new_type_def(&mut self, type_name: &'package ast::Ident, def: &'package ast::TyDef) {}
+
+    fn at_new_type_ref(
+        &mut self,
+        path: &'package ast::Path,
+        item_id: &'package hir::ItemId,
+        item: &'package hir::Item,
```
minestarks, line 36: My comments about `ItemId` and `Item` above for `at_callable_ref` also apply here.

language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,318 @@
...
+    ) {
+    }
+
+    fn at_field_ref(
+        &mut self,
+        expr_id: &'package ast::NodeId,
+        field_ref: &'package ast::Ident,
+        item_id: &'package hir::ItemId,
+        field_def: &'package hir::ty::UdtField,
+    ) {
```
minestarks, line 57: In the spirit of keeping the "the first parameter is the thing that the cursor is on", I suggest putting the `field_ref` parameter at the top. All the other `*_ref` methods already follow this pattern

language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,327 @@
...
+        &mut self,
+        context: &LocatorContext<'package>,
+        path: &'package ast::Path,
+        node_id: &'package ast::NodeId,
+        ident: &'package ast::Ident,
+    ) {
+    }
+}
+
+pub(crate) struct LocatorContext<'package> {
```
minestarks, line 81: To align with my `IdentVisitor` rename, maybe `IdentContext` for this? Or something else that makes sense to you.

language_service/src/identifier_locator.rs
ScottCarda-MS: done

language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,318 @@
...
+
+use std::mem::replace;
+use std::rc::Rc;
+
+use crate::qsc_utils::{find_ident, find_item, span_contains, span_touches, Compilation};
+use qsc::ast::visit::{walk_expr, walk_namespace, walk_pat, walk_ty_def, Visitor};
+use qsc::{ast, hir, resolve};
+
+#[allow(unused_variables)]
+pub(crate) trait LocatorAPI<'package> {
```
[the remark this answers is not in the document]
ScottCarda-MS: Discussed offline: went with `name_locator::Handler`.

language_service/src/identifier_locator.rs
ScottCarda-MS: Discussed offline: the name makes more sense now with the renaming of `IdentifierLocator` to just `Locator`.

language_service/src/identifier_locator.rs
ScottCarda-MS: Discussed offline: Since Hover is the only one to use this, let's wait to see if any other features added in the future need this info. In the meantime it can live in the hover.

language_service/src/identifier_locator.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn at_callable_def(
+        &mut self,
+        context: &LocatorContext<'package>,
+        decl: &'package ast::CallableDecl,
+    ) {
+    }
```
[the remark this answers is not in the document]
ScottCarda-MS: Done

minestarks (approved):

PR #815, opened 2023-10-24
Integrate the project system into the VSCode Extension

sezna:
This is a draft/staging branch for the wasm interop version of the qsharp project system.

### DIFF
--- a/compiler/qsc_project/src/js.rs
+++ b/compiler/qsc_project/src/js.rs
@@ -1,0 +1,62 @@
+use std::path::PathBuf;
+
+use crate::{DirEntry, EntryType};
+
+pub struct JsFileEntry {
+    ty: EntryType,
+    extension: String,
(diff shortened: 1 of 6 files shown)


compiler/qsc/src/incremental.rs
```diff
@@ -77,27 +77,43 @@ impl Compiler {
...
-        if !pass_errors.is_empty() {
-            return Err(into_errors_with_source(pass_errors, &unit.sources));
+        let mut errors = false;
+        let mut increment =
+            self.frontend
+                .compile_fragments(unit, source_name, source_contents, |e| {
+                    errors = errors || !e.is_empty();
+                    accumulate_errors(into_errors(e))
+                })?;
+
```
idavis, line 106: Why are we stopping default passes? Some provide valuable feedback when there are errors.

PR #816, opened 2023-10-25
Using FxHash* instead of Hash*

idavis:
Using `rustc`'s `FxHashSet` and `FxHashMap` and disabling the use of `std::collections::Hash(Set|Map)`

```console
Teleport evaluation     time:   [80.908 µs 81.012 µs 81.138 µs]
                        change: [-9.0801% -8.7843% -8.4988%] (p = 0.00 < 0.05)
                        Performance has improved.

Deutsch-Jozsa evaluation
                        time:   [21.520 ms 21.534 ms 21.548 ms]
                        change: [-2.6201% -2.4535% -2.3096%] (p = 0.00 < 0.05)
                        Performance has improved.

Large file parity evaluation
                        time:   [33.821 ms 33.876 ms 33.942 ms]
                        change: [-0.8022% -0.4683% -0.1777%] (p = 0.00 < 0.05)

Large input file        time:   [16.596 ms 16.614 ms 16.632 ms]
                        change: [-6.6542% -6.4932% -6.3307%] (p = 0.00 < 0.05)
                        Performance has improved.

Standard library        time:   [7.3145 ms 7.3215 ms 7.3294 ms]
                        change: [-4.6977% -4.5701% -4.4439%] (p = 0.00 < 0.05)
                        Performance has improved.
```


compiler/qsc_eval/src/lib.rs
```diff
@@ -386,7 +387,7 @@ impl Env {
 
 #[derive(Default)]
 struct Scope {
-    bindings: HashMap<NodeId, Variable>,
+    bindings: FxHashMap<NodeId, Variable>,
```
minestarks, line 390: Why a hash map instead of an `IndexMap` in the first place? Looks like in a lot of places we use `IndexMap` for `NodeId`s and other sequentially generated integers.

clippy.toml
```diff
@@ -0,0 +1 @@
+disallowed-types = ["std::collections::HashMap", "std::collections::HashSet"]
```
minestarks, line 1: Clippy lets you provide a reason here, something like `{ path = "std::collections::HashMap", reason = "use FxHashMap instead" }`. Might be helpful for our future selves...

compiler/qsc_frontend/src/closure.rs
```diff
@@ -63,7 +61,7 @@ impl Visitor<'_> for VarFinder {
 }
 
 struct VarReplacer<'a> {
-    substitutions: &'a HashMap<NodeId, NodeId>,
+    substitutions: &'a FxHashMap<NodeId, NodeId>,
```
minestarks, line 64: Same question here about `IndexMap<NodeId, T>` vs. a hash map.

Basically I have the same question everywhere a `FxHashSet<FooId>` or `FxHashMap<FooId, T>` occurs in the PR... I'll refrain from commenting on every single one :)

minestarks (approved):
Thank you for this!

Going forward I think it'll be useful to have a benchmark for WASM as well as that's our primary runtime and I can imagine it having surprisingly different performance characteristics than native.

compiler/qsc_frontend/src/closure.rs
idavis: Great question @minestarks - I had hoped to do a follow up and see where we can use `IndexMap`, but there is a density issue. If we don't have a dense map, then `IndexMap` could allocate a lot of resources for very few entries. For example, if the `IndexMap` is given `NodeId` `22358`, then we just allocated a buffer with 22k+ slots for a single value. This PR is just a starting point.

compiler/qsc_eval/src/lib.rs
idavis: See other comment

clippy.toml
idavis: Updated

compiler/qsc_frontend/src/closure.rs
minestarks: I did have that similar concern. But then my question becomes, how are we so sure that the _other_ `IndexMap<NodeId>`s in the code won't start at `22358`? I can see a few spots where it's almost guaranteed the map will contain most of the nodes in the program, but I'm not so sure about other spots. 

That being said, I agree this is out of scope for this PR, and I'm happy to hear you plan to keep digging!

DmitryVasilevsky:
This is probably worth mentioning to show tradeoffs of switching.
https://nnethercote.github.io/perf-book/hashing.html

PR #819, opened 2023-10-25
Ripple-carry addition - arithmetic namespace functions

DmitryVasilevsky:
First portion of the arithmetic namespace functionality:
- Added Ripple-carry little-endian addition operations to Arithmetic namespace: `RippleCarryIncByL`, `RippleCarryIncByLE`, and `RippleCarryAddLE`, and a few auxiliary internal operations.
- Also added wrapper functions IncByL, IncByLE, and AddLE.
- Added functions `TrailingZeroCountI` and `TrailingZeroCountL` to Math namespace and `ApplyXorInPlaceL` to Arithmetic namespace.
- Added exhaustive and branching case tests. Large test functions are moved to separate .qs files.

Acknowledgements:
- Thanks to @msoeken for providing most code and tests
- Thanks to @swernli for help with fitting tests into our testing approach


library/std/arithmetic.qs
```diff
@@ -33,6 +34,32 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// Applies a bitwise-XOR operation between a classical integer and an
+    /// integer represented by a register of qubits.
+    ///
+    /// # Description
+    /// Applies `X` operations to qubits in a little-endian register based on
+    /// 1 bits in an integer.
+    ///
+    /// Let us denote `value` by a and let y be an unsigned integer encoded in `target`,
+    /// then `ApplyXorInPlace` performs an operation given by the following map:
+    /// |y⟩ -> |y ⊕ a⟩, where ⊕ is the bitwise exclusive OR operator.
```
msoeken, line 47:
```suggestion
    /// |y⟩ ↦ |y ⊕ a⟩, where ⊕ is the bitwise exclusive OR operator.
```
We might want to use this unicode character. We are already using unicode (for XOR) and it's more accurate than ->

library/std/arithmetic.qs
```diff
@@ -33,6 +34,32 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    operation ApplyXorInPlaceL(value : BigInt, target : Qubit[]) : Unit is Adj+Ctl {
+        body(...) {
```
msoeken, line 49:
```suggestion
        body (...) {
```

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    //
+
+    /// # Summary
+    /// Increment a little-endian ys register by a BigInt number c
+    ///
+    /// # Description
+    /// Compute ys += c modulo 2ⁿ, where ys is a little-endian register
+    /// of length n > 0, c >= 0 is a BigInt constant, and c < 2ⁿ.
+    /// This operation can be used when the choice of implementation
+    /// is not important.
```
msoeken, line 344: I think in most cases, the user will choose this operation. We could phrase this sentence more positively by writing something like "If the choice of implementation is of importance, use ... or ..." with references to the possible implementations.

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// n = Length(ys), and Length(xs) <= Length(ys).
+    /// This operation can be used when the choice of implementation
+    /// is not important.
+    @Config(Full)
+    operation IncByLE (xs : Qubit[], ys : Qubit[]) : Unit is Adj + Ctl {
+        RippleCarryIncByLE(xs, ys);
+    }
+
+    /// # Summary
+    /// Set a zero-initialized little-endian register zs to the sum of
```
msoeken, line 364:
```suggestion
    /// Sets a zero-initialized little-endian register zs to the sum of
```

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// using the ripple-carry algorithm.
+    ///
+    /// # Description
+    /// Compute ys += c modulo 2ⁿ, where ys is a little-endian register
+    /// of length n > 0, c >= 0 is a BigInt constant, and c < 2ⁿ.
+    /// This operation uses the ripple-carry algorithm.
+    @Config(Full)
+    operation RippleCarryIncByL (c : BigInt, ys : Qubit[]) : Unit is Adj + Ctl {
+        let ysLen = Length(ys);
+        Fact(ysLen > 0, "Length of `ys` must be at least 1");
```
msoeken, line 389: There is inconsistency in whether diagnostic messages start lower case (see above) or upper case (here).

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// # Summary
+    /// Increment a little-endian ys register by a BigInt number c
+    /// using the ripple-carry algorithm.
+    ///
+    /// # Description
+    /// Compute ys += c modulo 2ⁿ, where ys is a little-endian register
+    /// of length n > 0, c >= 0 is a BigInt constant, and c < 2ⁿ.
+    /// This operation uses the ripple-carry algorithm.
+    @Config(Full)
+    operation RippleCarryIncByL (c : BigInt, ys : Qubit[]) : Unit is Adj + Ctl {
```
msoeken, line 387: One could create a new internal operation based on this operation for constant increment that takes as argument an adder implementation. Then this operation can call it with a ripple carry adder and one can have increment by addition also for carry lookahead.

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// using the ripple-carry algorithm.
+    ///
+    /// # Description
+    /// Compute ys += xs modulo 2ⁿ, where xs and ys are little-endian registers,
+    /// n = Length(ys), and Length(xs) <= Length(ys).
+    /// Note that ff Length(xs) != Length(ys), xs is padded with 0-initialized
+    /// qubits to match ys's length
+    ///
+    /// # Reference:
+    ///     - [arXiv:1709.06648](https://arxiv.org/pdf/1709.06648.pdf)
```
msoeken, line 420: Maybe we can list authors and title.

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+
+
+    /// # Summary
+    /// Increments a little-endian register ys by a little-endian register xs
+    /// using the ripple-carry algorithm.
+    ///
+    /// # Description
+    /// Compute ys += xs modulo 2ⁿ, where xs and ys are little-endian registers,
+    /// n = Length(ys), and Length(xs) <= Length(ys).
+    /// Note that ff Length(xs) != Length(ys), xs is padded with 0-initialized
```
msoeken, line 416:
```suggestion
    /// Note that if Length(xs) != Length(ys), xs is padded with 0-initialized
```

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// qubits to match ys's length
+    ///
+    /// # Reference:
```
msoeken, line 419:
```suggestion
    /// # Reference
```

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+        Fact(xsLen >= 1, "Registers xs and ys must contain at least one qubit");
+
+        if ysLen - xsLen >= 2 {
+            use padding = Qubit[ysLen - xsLen - 1];
+            RippleCarryIncByLE(xs + padding, ys);
+        } elif xsLen == 1 {
+            if ysLen == 1 {
+                CNOT(xs[0], ys[0]);
+            } elif ysLen == 2 {
+                HalfAdderInc(xs[0], ys[0], ys[1]); // TODO: Can we pass ys[1] here?
```
msoeken, line 436: Is the TODO still required?

library/std/arithmetic_internal.qs
```diff
@@ -0,0 +1,163 @@
...
+        }
+        controlled adjoint auto;
+    }
+
+    /// # Summary
+    /// Applies AND gate between `control1` and `control2` and stores the result
+    /// in `target` assuming `target` is in |0> state. This allows for an
+    /// otimized adjoint implementation.
+    @Config(Full)
+    internal operation ApplyAndWith0Target(control1 : Qubit, control2 : Qubit, target: Qubit) : Unit is Adj {
```
msoeken, line 149: This should eventually be a public operation (CC @swernli).

library/std/arithmetic_internal.qs
```diff
@@ -0,0 +1,163 @@
...
+            CCNOT(ctl, x, y); // Controlled X(ctls + [x], y);
+            CNOT(carryIn, x);
+            CNOT(carryIn, y);
+        }
+        controlled adjoint auto;
+    }
+
+    /// # Summary
+    /// Applies AND gate between `control1` and `control2` and stores the result
+    /// in `target` assuming `target` is in |0> state. This allows for an
```
msoeken, line 146: Can we check that target is in |0>?

library/std/arithmetic_internal.qs
```diff
@@ -0,0 +1,163 @@
...
+    internal operation Carry(carryIn : Qubit, x : Qubit, y : Qubit, carryOut : Qubit) : Unit is Adj + Ctl {
+        body (...) {
+            CNOT(carryIn, x);
+            CNOT(carryIn, y);
+            ApplyAndWith0Target(x, y, carryOut);
+            CNOT(carryIn, carryOut);
+        }
+        adjoint auto;
+        controlled (ctls, ...) {
+            // TODO: Is it not controlled actually?
```
msoeken, line 113: It's worth to point this out in fact. The `Carry` and `Uncarry` operations are meant to be used only in an in-place ripple-carry adder implementation. In the controlled ripple-carry adder it is sufficient to only control the `Uncarry` operation. By defining the `Carry` operation like this (i.e., having the controlled version being the same as the body), we can simplify the implementation of the ripple-carry adder, by automatically generating the controlled variant.

msoeken (commented):
That looks great! Thanks @DmitryVasilevsky

library/std/arithmetic_internal.qs
DmitryVasilevsky: I named it this way, but don't like this name much... It kind of sounds like you apply AND, and one of the arguments is "0 target", which is not what this operation does.

library/std/arithmetic.qs
DmitryVasilevsky: Updated in both places.

library/std/arithmetic.qs
```diff
@@ -33,6 +34,32 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// integer represented by a register of qubits.
+    ///
+    /// # Description
+    /// Applies `X` operations to qubits in a little-endian register based on
+    /// 1 bits in an integer.
+    ///
+    /// Let us denote `value` by a and let y be an unsigned integer encoded in `target`,
+    /// then `ApplyXorInPlace` performs an operation given by the following map:
+    /// |y⟩ -> |y ⊕ a⟩, where ⊕ is the bitwise exclusive OR operator.
+    operation ApplyXorInPlaceL(value : BigInt, target : Qubit[]) : Unit is Adj+Ctl {
```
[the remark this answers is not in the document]
DmitryVasilevsky: Updated to use spaces. Eventually we'll have a formatter to take case of this.

library/std/arithmetic.qs
DmitryVasilevsky: Added space in both places.

library/std/arithmetic.qs
DmitryVasilevsky: Updated text.

library/std/arithmetic.qs
DmitryVasilevsky: Rather than describing what a function does, I actually prefer the imperative form with simple present tense - same form when we give instructions and directions. Many of our functions are commented that way. So when you look at your program you have something like this:
- Allocate qubit
- Apply H gate
- Measure result
- Print the result
rather than description of what each function does.
We have a mix, so I update a few other places.

DmitryVasilevsky: I checked and most places use this 'descriptive' form rather than imperative form. I'll update a few place to use s-es.

library/std/arithmetic_internal.qs
msoeken: It would be great to have this as an intrinsic operation (like `CCNOT`) and then have a simple name such as `And`.

library/std/arithmetic.qs
msoeken: I should have commented on consistency. It's also okay to change the other occurrences.

library/std/arithmetic.qs
DmitryVasilevsky: We could do it this way when I port another adder.

library/std/arithmetic.qs
DmitryVasilevsky: Typically error messages start with capital letter. They start with the small letter if the first word is the reference to one of the arguments. I'll add quotations if it helps.

library/std/arithmetic.qs
DmitryVasilevsky: Updated.

library/std/arithmetic.qs
DmitryVasilevsky: Updated in a few places.

library/std/arithmetic.qs
DmitryVasilevsky: Added.

library/std/arithmetic.qs
DmitryVasilevsky: This is the place where we need to use CCNOT because ys may be 11₂? So we still need to fix this code?

library/std/arithmetic.qs
```diff
@@ -289,4 +316,167 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// little-endian registers xs and ys using the ripple-carry algorithm.
+    ///
+    /// # Description 
+    /// Compute zs := xs + ys + zs[0], where xs, ys, and zs are little-endian
+    /// registers, Length(xs) = Length(ys), Length(xs) <= Length(zs) <= Length(xs)+1,
+    /// assuming zs is 0-initialized except for maybe zs[0], which can be in |0> or |1> state.
+    /// Use `zs[0]` as carry-in, and use `zs[n]` as carry-out, if `zs` is
+    /// longer than `xs`.
+    ///
+    /// # Reference:
```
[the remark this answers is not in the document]
DmitryVasilevsky: Updated here and in a few other places.

library/std/arithmetic_internal.qs
DmitryVasilevsky: I added the comment.

library/std/arithmetic_internal.qs
swernli: @msoeken Agreed. I think we want that to be the case eventually, and use that intrinsic both here and in the decompositions invoked in internal.qs (today called `PhaseCCX`). I'll include mention of this in our upcoming specifications so that it can be a clear part of the roadmap.

library/std/arithmetic_internal.qs
DmitryVasilevsky: Added check.

library/std/arithmetic.qs
DmitryVasilevsky: I replaced AND with CCNOT in HalfAdderInc and removed the TODO. It was indeed possible to pass 1 in that ys[1]. Added test.

library/tests/src/test_arithmetic.rs
```diff
@@ -0,0 +1,401 @@
...
+#[test]
+fn check_add_i_1_1() {
+    test_expression(
+        {
+            "{  // Shortest case
+                use x = Qubit[1];
+                use y = Qubit[1];
+                open Microsoft.Quantum.Arithmetic;
+                X(x[0]);
+                AddI(x,y);
```
msoeken, line 129: Will AddI point to IncByLE eventually?

library/tests/src/test_arithmetic.rs
```diff
@@ -0,0 +1,401 @@
...
+                ResetAll(q);
+                return result;
+            }"
+        },
+        &Value::Int(45967),
+    );
+}
+
+#[test]
+fn check_add_i_nc() {
```
msoeken, line 64: I suggest to exhaustively test the adders for all possible input value combinations up to some bit width. There are a lot of corner cases to consider.

msoeken (approved):

library/tests/src/test_arithmetic.rs
DmitryVasilevsky: I also need to add tests for controlled versions.

library/tests/src/test_arithmetic.rs
DmitryVasilevsky: We should remove AddI along with the backing code. We have a way of pointing users to the new function if they use the old one.

swernli:
Thanks for updating the tests to be more granular! As we discussed, I think we may want to do a similar refactoring to break up tests in other places in the libraries, but that is out of scope for this PR. I'll file an issue to cover that and we can continue that part of the discussion there.

swernli (approved):

swernli:
> Thanks for updating the tests to be more granular! As we discussed, I think we may want to do a similar refactoring to break up tests in other places in the libraries, but that is out of scope for this PR. I'll file an issue to cover that and we can continue that part of the discussion there.

Filed as #832

PR #820, opened 2023-10-26
Add ability to continue on error for incremental compiler

minestarks:
This change is required for the notebook language service (#759). The new public method is not used in this PR yet. It will be used in #759.

This change adds two flavors of `compile_fragments` to the incremental compiler: one that fails fast and one that accumulates errors but continues compilation.

When calling from the interpreter, we want to fail fast when parsing fails, so that we don't pollute the resolver/checker/lowerer state with flawed data when we already _know_ that cell execution will fail.

On the other hand, when calling from the language service, we want to continue compilation even in the case of errors, so we can finish building our AST & HIR we will use for editor features. Also, this way we can report resolve/check errors even if there are syntax error elsewhere in the program (e.g. in another cell).

### DIFF
--- a/compiler/qsc/src/lib.rs
+++ b/compiler/qsc/src/lib.rs
@@ -10,7 +10,7 @@ pub mod incremental;
 pub mod interpret;
 
 pub use qsc_frontend::compile::{
-    AstPackage, CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
+    CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
 };
 
 pub mod resolve {
--- a/compiler/qsc/src/lib.rs
+++ b/compiler/qsc/src/lib.rs
@@ -10,7 +10,7 @@ pub mod incremental;
 pub mod interpret;
 
 pub use qsc_frontend::compile::{
-    CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
+    AstPackage, CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
 };
 
 pub mod resolve {
--- a/compiler/qsc/src/incremental.rs
+++ b/compiler/qsc/src/incremental.rs
@@ -77,27 +77,50 @@ impl Compiler {
     /// get information about the newly added items, or do other modifications.
     /// It is then the caller's responsibility to merge
     /// these packages into the current `CompileUnit` using the `update()` method.
-    pub fn compile_fragments(
+    pub fn compile_fragments_fail_fast(
         &mut self,
         source_name: &str,
         source_contents: &str,
     ) -> Result<Increment, Errors> {
+        self.compile_fragments(source_name, source_contents, fail_on_error)
+    }
+
+    /// Compiles Q# fragments. See [`compile_fragments_fail_fast`] for more details.
+    ///
+    /// This method calls an accumulator function with any errors returned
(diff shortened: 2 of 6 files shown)


minestarks:
@sezna closures in Rust are super common though? My mental model was something like the closure that goes into [`take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while)  . It's a closure that gets called with each element, and iteration stops if the closure returns `false`. Except in my code, we're "iterating" over compiler phases, the "elements" are errors, and the closure returns an error instead of a boolean.

I'm not sure if the trait example you gave would work for this code, because `accumulate_errors` is supposed to be implemented by the caller.

Maybe I should have linked to the site where `compile_fragments` will actually be called with an accumulator so that my intention was clearer.

https://github.com/microsoft/qsharp/pull/759/files#diff-d98d007bbe0a3fce75f8d49278e0b566c9f8d91731f3e451296db14bebff001fR81-R92

compiler/qsc/src/lib.rs
```diff
@@ -10,7 +10,7 @@ pub mod incremental;
 pub mod interpret;
 
 pub use qsc_frontend::compile::{
-    CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
+    AstPackage, CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
```
swernli, line 13: Why is `AstPackage` becoming pub here? Is this a precursor of usage somewhere else?

swernli:
> On the other hand, when calling from the language service, we want to continue compilation even in the case of errors, so we can finish building our AST & HIR we will use for editor features. Also, this way we can report resolve/check errors even if there are syntax error elsewhere in the program (e.g. in another cell).

Just to make sure I understand, when does the LS know to throw away AST/HIR and associated incremental compilation that might be "poisoned" by errors? Does it recreate the whole incremental compiler each time? This may be a question that is deferred until #759?

minestarks:
> Just to make sure I understand, when does the LS know to throw away AST/HIR and associated incremental compilation that might be "poisoned" by errors? Does it recreate the whole incremental compiler each time? This may be a question that is deferred until #759?

@swernli So, to make sure I'm interpreting your question correctly. Are you asking about updating the documents (i.e. user typing) or incrementally compiling each cell?

As the user edits a cell, the entire program is thrown away and recreated with each keystroke. So that part's simple.

Let's think about the read-only case. When you open a notebook, each cell is incrementally compiled. The AST and HIR for each cell are kept, even if that cell happens to contain errors. There are error nodes, sure, but the tree isn't "poisoned" per se.


Example:


cell 1:
```qsharp
operation Foo() : Unit  {}

operation Bar() : some_undefined_type {}
                  ^^^^^^^^^^^^^^^^^^^

some_bad_syntax
^^^^^^^^^^^^^^^
```

cell 2:
```qsharp
Foo()
```

There's really no reason that `Foo` should fail to resolve just because cell 1 contains unrelated errors. It shouldn't contain squiggles. If I go-to-definition on it, it should work.

compiler/qsc/src/lib.rs
minestarks: That snuck in from a different change, will pull it out!

swernli (approved):

minestarks:
@sezna thank you for this, I learned something new! As discussed, I envision this closure being a one-off thing, but I'll refactor later if we find ourselves adding more closures and need to group them together into a coherent trait.

PR #821, opened 2023-10-26
Remove superfluous and flawed telemetry event

sezna:
Thanks @minestarks for pointing this out. This event only fires when cells are set to q# for the first
time, not on subsequent cell accesses. This isn't particularly useful telemetry. Additionally, this event
is superseded by [this event](https://github.com/microsoft/qsharp/blob/main/vscode/src/extension.ts#L121).

### DIFF
--- a/vscode/src/telemetry.ts
+++ b/vscode/src/telemetry.ts
@@ -5,7 +5,6 @@ import { log } from "qsharp-lang";
 export enum EventType {
   InitializePlugin = "Qsharp.InitializePlugin",
   LoadLanguageService = "Qsharp.LoadLanguageService",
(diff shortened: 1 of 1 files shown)


minestarks (approved):

vscode/src/notebook.ts
```diff
@@ -58,7 +57,6 @@ export function registerQSharpNotebookHandlers() {
             cell.document,
             qsharpLanguageId
           );
-          sendTelemetryEvent(EventType.QSharpJupyterCellInitialized);
```
billti, line 61: If this is the only place this eventtype was used, shouldn't you be removing the (now unused) definition from vscode/src/telemetry.ts also?

PR #823, opened 2023-10-27
Un-deprecate `checkCode`

minestarks:
Apparently, the website intends to use this to check the validity of generated Q# snippets. This seems like a legitimate use case. It's a perfectly good API :)

My original intention for deprecating `checkCode` was to guide people towards using the language service for editor squiggles, and that mission has already been accomplished...

### DIFF
--- a/npm/src/compiler/compiler.ts
+++ b/npm/src/compiler/compiler.ts
@@ -14,9 +14,6 @@ type Wasm = typeof import("../../lib/node/qsc_wasm.cjs");
 // These need to be async/promise results for when communicating across a WebWorker, however
 // for running the compiler in the same thread the result will be synchronous (a resolved promise).
 export interface ICompiler {
-  /**
-   * @deprecated use the language service for errors and other editor features.
-   */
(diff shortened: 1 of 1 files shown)


billti:
I believe as part of deprecating it, you also made the 'deprecated' implementation spin up a language service and update a document to run 'checkCode'. Is that still an efficient and valid way to implement this API if we intend to keep it? (Maybe it's fine - just asking).

billti (approved):

minestarks:
> I believe as part of deprecating it, you also made the 'deprecated' implementation spin up a language service and update a document to run 'checkCode'. Is that still an efficient and valid way to implement this API if we intend to keep it? (Maybe it's fine - just asking).

Yeah, good question. I think it's OK. There's nothing "spinning up", this is exactly the same amount of work as the old `check_code` (compile a snippet and report errors) except we don't cache the std/core libs. Perhaps we'll want to bring back the dedicated `check_code` in the wasm layer if we want more out of this API, but for now I think this fine.

Issue #699, opened 2023-09-14
Consider more informative guidance for customers trying to migrate code


billti:
Rather than just 'identifier not found' (such as for trying to use an API from the old library that no longer exists), or 'syntax error' (such as for interpolated string changes), we could perhaps detect these cases and have a more helpful error message, perhaps redirecting to a page that lists the changes and how to migrate such code (or open an issue to request an API/feature).

billti:
Lots of fun debate on this one. I think we still need to 'triage' further to come to agreement.

filipw:
I vote for a compatibility layer (for missing library functions)

swernli:
The `@Unimplemented` attribute was added to the language to help provide guidance on this, but does not allow for adding any extra text to the error. After discussion with @DmitryVasilevsky we landed at wanting support for an optional string that gets added to the error output. I'll work on that, and that should enable any additional libraries we want to add with guidance for users on what to use.

swernli:
Diggin into this further, we hit some complications: 
 - Adding support of a string in the `@Unimplemented` attribute parameters would mean copying that string across AST->HIR, which makes some of the data structures non-`Copy` and interferes with the way they are being used. This can be fixed, but becomes a larger refactor that ideally would use `Copy` identifiers (like integers) to index into a cache of strings. At that point it just becomes a small implementation of string interning, which we have wanted to get across the compiler anyway, but shouldn't be done quickly for just one part of the code.
 - APIs marked as unimplemented should be suppressed in completion lists, which will require an update to the language server to make it attribute aware when putting together the completion list.

PR #826, opened 2023-10-30
Support Unimplemented attribute

swernli:
This change introduces the `@Unimplemented()` attributed which can be used to mark an item as something that should not be used. The motivation behind including this and not just treating it as a `Not Found` error is to give the user more information that this API may have previously existing and is not currently supported (see #699 for context).

### DIFF
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -629,7 +629,7 @@ fn bind_global_items(
 ) {
     names.insert(
         namespace.name.id,
-        Res::Item(intrapackage(assigner.next_item()), ItemStatus::Normal),
+        Res::Item(intrapackage(assigner.next_item()), ItemStatus::Available),
     );
     scope.namespaces.insert(Rc::clone(&namespace.name.name));
 
@@ -817,7 +817,7 @@ fn resolve(
 
     if candidates.len() > 1 {
         // If there are multiple candidates, remove unimplemented items. This allows resolution to
-        // succeed in cases where both an older, unimplimented API and newer, implemented API with the
+        // succeed in cases where both an older, unimplemented API and newer, implemented API with the
         // same name are both in scope without forcing the user to fully qualify the name.
         let mut removals = Vec::new();
         for res in candidates.keys() {
@@ -878,7 +878,7 @@ fn resolve_scope_locals(
     }
 
     if let Some(&id) = scope.item(kind, name) {
-        return Some(Res::Item(id, ItemStatus::Normal));
+        return Some(Res::Item(id, ItemStatus::Available));
     }
 
     if let ScopeKind::Namespace(namespace) = &scope.kind {
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -420,11 +420,10 @@ impl AstVisitor<'_> for With<'_> {
     }
 
     fn visit_attr(&mut self, attr: &ast::Attr) {
-        if hir::Attr::from_str(attr.name.name.as_ref()) == Ok(hir::Attr::Config) {
-            // The Config attribute arguments do not go through name resolution.
-            return;
+        // The Config attribute arguments do not go through name resolution.
+        if hir::Attr::from_str(attr.name.name.as_ref()) != Ok(hir::Attr::Config) {
+            walk_attr(self, attr);
         }
-        walk_attr(self, attr);
     }
 
     fn visit_callable_decl(&mut self, decl: &ast::CallableDecl) {
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -36,7 +36,7 @@ pub(super) type Names = IndexMap<NodeId, Res>;
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum Res {
     /// A global item.
-    Item(ItemId),
+    Item(ItemId, ItemStatus),
     /// A local variable.
     Local(NodeId),
     /// A type/functor parameter in the generics section of the parent callable decl.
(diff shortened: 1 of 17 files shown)


swernli:
whoops... looks like I somehow merged @DmitryVasilevsky 's changes with my own here. I'll try to fix this up...

compiler/qsc_hir/src/hir.rs
```diff
@@ -174,6 +174,8 @@ pub struct ItemId {
     pub package: Option<PackageId>,
     /// The item ID.
     pub item: LocalItemId,
+    /// The item status.
+    pub status: ItemStatus,
```
minestarks, line 178: It's very strange to me that this is part of the `ItemId`. It's... not an ID. It's part of the item data. There are a _lot_ of places in the code that use `ItemId` as hashmap keys and now we're adding another field to hash for no good reason. Putting it in the `ItemId` also leaks this field into a bunch of places that don't need it, like the language service.

If this attr _has_ to be checked in the resolver, maybe it can just be another field in `resolve::Res::Item`? 

(Though you already know that my real preference is for this logic not to be in the resolver at all... the HIR `Item` will already have the attributes on it, it could be so easy just to make this check a pass. )

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -771,6 +807,21 @@ fn resolve(
         }
     }
 
+    if candidates.len() > 1 {
+        // If there are multiple candidates, remove unimplemented items.
```
minestarks, line 811: Why?

compiler/qsc_hir/src/hir.rs
swernli: I can try out what the code looks like by having it be a part of the `hir::Res::Item` and `resolve::Res::Item`. I think you are right that the latter would affect fewer places.

As for making it a pass, I think if we did that it would have to go in the HIR somewhere. While you are correct that the original item has the attributes on it, right now there are no passes that use the global table to look up resolved references from other parts of the package or dependent packages (note that `run_default_passes` doesn't take any global lookup tables, nor a reference to the compilation store). To have a pass check all resolutions by following the link and looking up the corresponding package from the store would be very different and upend the nice pattern we have today where passes only run on the package being compiled. Like type info, whether or not a resolution is pointing to a unimplemented item is information we have earlier in the compilation that we can put into the HIR itself to save later phases lookup.

All that said, right now I'm not using this info in later phases, so that's where your first suggestion of keeping this scoped to the resolver by using `resolve::Res::Item` fits with the current approach.

compiler/qsc_frontend/src/resolve.rs
swernli: Ah, the downsides of switching PRs... this question came up on the previous PR here: https://github.com/microsoft/qsharp/pull/785#discussion_r1365964792. Since it's come up twice, I think that's telling me I need to expand the comment to explain the scenario this is covering.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -390,8 +419,13 @@ impl AstVisitor<'_> for With<'_> {
...
     }
 
-    // We do not perform name resolution on attributes, as those are checking during lowering.
-    fn visit_attr(&mut self, _: &ast::Attr) {}
+    fn visit_attr(&mut self, attr: &ast::Attr) {
+        if hir::Attr::from_str(attr.name.name.as_ref()) == Ok(hir::Attr::Config) {
+            // The Config attribute arguments do not go through name resolution.
+            return;
+        }
+        walk_attr(self, attr);
```
ScottCarda-MS, line 427: nit: 
```suggestion
        // The Config attribute arguments do not go through name resolution.
        if hir::Attr::from_str(attr.name.name.as_ref()) != Ok(hir::Attr::Config) {
            walk_attr(self, attr);
        }
```

ScottCarda-MS (approved):

language_service/src/rename.rs
```diff
@@ -182,3 +187,87 @@ impl<'a> Handler<'a> for Rename<'a> {
         }
     }
 }
+
```
minestarks, line 190: Does this belong to this PR or did it sneak in from a merge?

compiler/qsc_hir/src/hir.rs
```diff
@@ -185,6 +185,28 @@ impl Display for ItemId {
     }
 }
 
+/// The status of an item.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum ItemStatus {
+    /// The item is defined normally.
+    Normal,
```
minestarks, line 192: Personal preference: `Normal` sounds very vague, maybe something more descriptive like `Available`?

language_service/src/rename.rs
swernli: hmm... it showed up after doing a merge from main. I think I might need to rebase on main instead.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -771,6 +815,21 @@ fn resolve(
         }
     }
 
+    if candidates.len() > 1 {
+        // If there are multiple candidates, remove unimplemented items. This allows resolution to
+        // succeed in cases where both an older, unimplimented API and newer, implemented API with the
```
minestarks, line 820:
```suggestion
        // succeed in cases where both an older, unimplemented API and newer, implemented API with the
```

minestarks (approved):
This implementation LGTM. Thank you!

Since we had that discussion about the fate of `@Config()` recently, looking at this PR with fresh eyes, this feature seems more-or-less equivalent to the `@Config()` attribute. We could define a special Config expression that should _never_ be true, like `@Config(Unimplemented)`. Would reduce complexity (since it would just get handled along with all the other `@Config`ged items in the preprocessing step) and I think the behavior would be ... the same? Minus some very specific error reporting, which may be worth the complexity I think.

That's my current take on the feature itself, but I'm happy whichever way you decide to go. Thanks for all the revisions and discussion!

compiler/qsc_hir/src/hir.rs
swernli: That sounds reasonable to me! I struggled with what to call this, and forgot to circle back and reevaluate. I like your suggestion better than what I had.

language_service/src/rename.rs
swernli: Ah, looks like these got moved to another file at the same time that I updated them, and the conflict resolution ended up duplicating them. Should be fixed now.

compiler/qsc_hir/src/hir.rs
swernli: I think we landed at a better place for this. If we do end up deciding to move this (and any upcoming logic) to a pass, we can reevaluate.

PR #831, opened 2023-11-02
Look-ahead and Fourier addition - arithmetic namespace functions

DmitryVasilevsky:
Adding Look-ahead and Fourier addition. Improved test error messages.
Two versions of ApplyAndAssuming0Target allows adders to support base and full profile.
Added HammingWeightI to Math namespace and tests for TrailingZeroCountI, TrailingZeroCountL.
Added ApplyQFT and SwapReverseRegister to Canon namespace.


library/std/arithmetic.qs
```diff
@@ -488,4 +488,74 @@ namespace Microsoft.Quantum.Arithmetic {
         }
     }
 
+    /// # Summary
+    /// Sets a zero-initialized little-endian register zs to the sum of
+    /// little-endian registers xs and ys using the carry-lookahead algorithm.
+    ///
+    /// # Description
+    /// Computes zs := xs + ys + zs[0] modulo 2ⁿ, where xs, ys, and zs are
+    /// little-endian registers, Length(xs) = Length(ys) ≤ Length(zs) = n (!!!),
```
msoeken, line 497: Why the (!!!)?

library/std/arithmetic.qs
```diff
@@ -488,4 +488,74 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    operation LookAheadAddLE(xs : Qubit[], ys : Qubit[], zs : Qubit[]) : Unit is Adj {
+        let xsLen = Length(xs);
+        let zsLen = Length(zs);
+        Fact(Length(ys) == xsLen, "Registers `xs` and `ys` must be of same length.");
+        Fact(zsLen == xsLen or zsLen == xsLen + 1, "Register `zs` must be same length as `xs` or one bit longer.");
+
+        if zsLen == xsLen + 1 { // with carry-out
+            // compute initial generate values
+            for k in 0..xsLen - 1 {
+                ApplyAndAssuming0Target(xs[k], ys[k], zs[k + 1]);
```
msoeken, line 516: A quantum AND gate always assumes that the target qubit is |0>, therefore the name is somewhat redundant. I suggest to use `ApplyAnd` or `CCNOTAssuming0Target`.

library/std/math.qs
```diff
@@ -716,6 +716,14 @@ namespace Microsoft.Quantum.Math {
         count
     }
 
+    /// # Summary
+    /// Returns the number of 1 bits in the binary representation of integer `n`.
+    function PopulationCountI (n : Int) : Int {
+        let i1 = n - ((n >>> 1) &&& 0x5555555555555555);
+        let i2 = (i1 &&& 0x3333333333333333) + ((i1 >>> 2) &&& 0x3333333333333333);
+        // Multiplication may overlow. See https://github.com/microsoft/qsharp/issues/828
```
msoeken, line 724:
```suggestion
        // Multiplication may overflow. See https://github.com/microsoft/qsharp/issues/828
```

library/std/math.qs
```diff
@@ -716,6 +716,14 @@ namespace Microsoft.Quantum.Math {
         count
     }
 
+    /// # Summary
+    /// Returns the number of 1 bits in the binary representation of integer `n`.
+    function PopulationCountI (n : Int) : Int {
```
msoeken, line 721: Alternative name could be `HammingWeightI`.

msoeken (approved):

library/std/arithmetic.qs
DmitryVasilevsky: There's a discrepancy between this doc comment and the code. I want code to accept longer zs but haven't done it yet.

DmitryVasilevsky: Relaxed condition.

library/std/arithmetic.qs
DmitryVasilevsky: We will fix this and similar places when we have proper intrinsic for this function. We'll probably move it to Canon and just call it AND gate.

library/std/canon.qs
```diff
@@ -476,4 +476,51 @@ namespace Microsoft.Quantum.Canon {
...
+            for j in 0..i-1 {
+                Controlled R1Frac([qs[i]], (1, j+1, qs[i-j-1]));
+            }
+        }
+    }
+
+    /// # Summary
+    /// Applies Quantum Fourier Transform (QFT) to a quantum register
+    /// and reverses the result to match the big-endian format.
+    operation QFTLE (qs : Qubit[]) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 508: Note that in this form QFTLE matches the legacy QDK operation with the same name.

swernli, line 508: This seems odd... is it intentional that the comment calls out the operation "reverses the result to match the big-endian format" but the operation is called `QFTLE`? Is this what you meant about maintaining the name or is it just a bug in the comment?

DmitryVasilevsky: This is very odd indeed and I'm glad you've asked. I spent a lot of time trying to figure out how I should comment this...
First, the truth. I only want to match the legacy QDK function with the same name. It would be weird if the legacy QDK function and the modern QDK function with the same name produced different outputs.
The legacy QDK operation [QFTLE](https://github.com/microsoft/QuantumLibraries/blob/a943e7087a45bbfcac57aea916d34c619d69325e/Standard/src/Canon/CommonGates.qs#L158C8-L158C8), through the series of transformations, was calling [ApproximateQFT](https://github.com/microsoft/QuantumLibraries/blob/a943e7087a45bbfcac57aea916d34c619d69325e/Standard/src/Canon/QFT.qs#L36), which was BE to BE. At least this is what [this comment](https://github.com/microsoft/QuantumLibraries/blob/a943e7087a45bbfcac57aea916d34c619d69325e/Standard/src/Canon/QFT.qs#L51) was saying. As a result, the input passed to QFTLE was transformed into BE format and the output of ApproximateQFT was passed back as is. That's why I decided to replicate this comment.
Frankly, I don't see the meaning of LE or BE on the output of QFT - I wouldn't call some bit "least significant" of "most significant". It's the exponent that's different from bit to bit. So I added a comment to ApplyQFT explaining where the corresponding bit is. ApplyQFT doesn't do any swaps, therefore it's faster (and is often acceptable). But it doesn't match any of the legacy QDK functions. So I added the QFTLE with this comment.
Alternatively I can add a different comment just explaining in which order it returns bits. I should also note that couldn't find any tests for QFTLE in the legacy QDK, only QFTBE, and one doesn't call the other. So it may very well be some bug...

DmitryVasilevsky: We decided to remove QFTLE entirely.

library/std/canon.qs
```diff
@@ -476,4 +476,43 @@ namespace Microsoft.Quantum.Canon {
...
+    ///
+    /// # Input
+    /// ## register
+    /// The qubits order of which should be reversed using SWAP gates
+    operation SwapReverseRegister (register : Qubit[]) : Unit is Adj + Ctl {
+        let length = Length(register);
+        for i in 0 .. length/2 - 1 {
+            SWAP(register[i], register[(length - i) - 1]);
+        }
+    }
```
swernli, line 516: One more question, out of curiosity: why would someone use this to reverse the order of a qubit array, costing actual `SWAP` invocations, when they could just reverse the order via slicing with `register[...-1...]`?

swernli (approved):

library/std/canon.qs
DmitryVasilevsky: Slicing won't change the state of qubits, only reindex them. So if you want to actually set the state of specific qubit, you'll have to use SwapReverseRegister.
An example with two variables is probably easier to understand. If you have qubits named a and b, and after transformation X you got the state you wanted for a in b and vice versa, one option you have is to use the name 'a' in your program where you wanted the state of 'b'. But such code re-write isn't always possible. For example, if the code that runs before X and after X is the same.
This could be optimized on a different level. I expect actual quantum computer or simulator to have a mapping between qubits in the program to underlying qubits on the device. SWAP would just update that mapping performing no actual quantum operations.

PR #835, opened 2023-11-03
Add rotation angle validation in qsc_eval intrinsic functions

### LINKED ISSUE #834: Rotating by `NaN` value causes crash when used with controlled gates
[Playground repro](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACn2QTU%2FDMAyG7%2F0V1sahk6YWcWRCAjHEqdI%2B4IQ4eKlpLTVOl7hC07T%2FTtZt0F2wckj8Pn7tOM9zGMMabdtQksfHK6myVBAUvVJ5zPX5McwpGM%2BtspM%2B81ZzgHgQLAtbbGA5hta7yqMFrVHBoMCGoAtUgrqTJXx77htE2LiSskTQUmjREBS7ZYeinV2cXfZJAjFcSwIFG%2B%2BC%2B9LsDGVzxkpcUDZh9h9XoNYn4PFF1O8WjkXTyaXE4%2FFLUCBLOoF7eBfW2BnOEacHAw%2Bw7DYcq2ZXwnYo%2FCoNKaBUDUX1yZt1NL7LbodEQSFgRenN6MLt%2B8thNKTikle7tBemsB20fnbxH65p4l4j8GE%2Bp%2FCHDR3mnW0LNDULXU24okCaXiwPySH5AZ8jkVQJAgAA)

When a `NaN` value is passed to a controlled rotation, something in the evaluator crashes, causing no reporting to come back. The code from the playground link above:
```qsharp
namespace MyQuantumProgram {

    open Microsoft.Quantum.Math;
    @EntryPoint()
    operation Main() : Unit {
        use c = Qubit();
        use q = Qubit();

        let angle = ArcSin(2.0);

        Message($"angle = {angle}");

        // Ry(angle, q);
        Controlled Ry([c], (angle, q));

        DumpMachine();

        Reset(q);
    }
}
```
will only report one shot no matter how many are requested because of the crash. Of note, it seems to only affect the controlled operation, so if you comment out `Controlled Ry([c], (angle, q));` in favor of `Ry(angle, q);` the program will actually successfully run all shots. The behavior between controlled and uncontrolled rotation should be the same, and should either succeed predictably or fail gracefully.

swernli:
Fixes #834

### DIFF
--- a/compiler/qsc_eval/src/intrinsic/tests.rs
+++ b/compiler/qsc_eval/src/intrinsic/tests.rs
@@ -1173,3 +1173,75 @@ fn qubit_not_unique_three_qubit_error_second_third() {
         &expect!["qubits in gate invocation are not unique"],
     );
 }
+
+#[test]
+fn single_qubit_rotation_nan_error() {
+    check_intrinsic_output(
+        "",
+        indoc! {"{
+            use q = Qubit();
+            Rx(Microsoft.Quantum.Math.ArcSin(2.0), q);
+        }"},
+        &expect!["invalid rotation angle: NaN"],
+    );
+}
+
+#[test]
+fn two_qubit_rotation_nan_error() {
+    check_intrinsic_output(
+        "",
+        indoc! {"{
+            use (q1, q2) = (Qubit(), Qubit());
+            Rxx(Microsoft.Quantum.Math.ArcSin(2.0), q1, q2);
+        }"},
+        &expect!["invalid rotation angle: NaN"],
+    );
+}
+
+#[test]
+fn single_qubit_rotation_inf_error() {
+    check_intrinsic_output(
+        "",
+        indoc! {"{
+            use q = Qubit();
+            Rx(-Microsoft.Quantum.Math.Log(0.0), q);
+        }"},
+        &expect!["invalid rotation angle: inf"],
+    );
+}
+
+#[test]
+fn two_qubit_rotation_inf_error() {
+    check_intrinsic_output(
+        "",
+        indoc! {"{
+            use (q1, q2) = (Qubit(), Qubit());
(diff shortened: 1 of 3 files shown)


compiler/qsc_eval/src/intrinsic/tests.rs
```diff
@@ -1173,3 +1173,75 @@ fn qubit_not_unique_three_qubit_error_second_third() {
...
     );
 }
+
+#[test]
+fn single_qubit_rotation_nan_error() {
+    check_intrinsic_output(
+        "",
+        indoc! {"{
+            use q = Qubit();
+            Rx(Microsoft.Quantum.Math.ArcSin(2.0), q);
```
minestarks, line 1183: Should `ArcSin` do any parameter validation so it fails _before_ it returns NaN?

minestarks (approved):

compiler/qsc_eval/src/intrinsic/tests.rs
swernli: I don't think so, because `NaN` is the defined result of `ArcSin` when the input value is outside of the accepted range (apparently that is common for trig libraries). A user used to be able to guard against these kinds of values in their own code with utility functions like `IsNaN` and `IsInfinite` (see the old math library [here](https://github.com/microsoft/qsharp-runtime/blob/3af588c549aa7d3d6d512e00e61ab697597b2fed/src/Simulation/QSharpFoundation/Math/FloatingPoint.qs#L46)), but we don't have equivalents for those currently. I'm not sure we actually need or want those functions back, I defer to @DmitryVasilevsky on that one.

minestarks: Makes sense to me. Thanks

Issue #749, opened 2023-09-21
Recover from Parsing Partial Expressions


ScottCarda-MS:
In order for more sophisticated Signature Help in the Language Server, the Q# parser needs to be able to better recover from dealing with partially typed expressions, for example `Foo(1 + )`. An AST and HIR should still be generated for this, maybe with some kind of error value for the incomplete expression.

PR #836, opened 2023-11-04
Use `Default` to handle missing entry in sequence

### LINKED ISSUE #672: Recover from Parsing Erroneous Tuple with Commas
In order for more sophisticated Signature Help in the Language Server, the Q# parser needs to be able to better recover from dealing with tuples (specifically argument tuples) that have commas separating no elements, for example `Foo(1, , 3)`. An AST and HIR should still be generated for this, maybe with some kind of error value where there are missing entries.

See also #749

swernli:
This uses a more complex but ultimately more correct approach that fixes #672. By creating default error kinds for several of the AST and HIR types that can appear in sequences, sequence parsing recovery now inserts error entries to correspond to places where commas appear at the start of a sequence or back-to-back in the sequence. The end result is better signature help due to recovered HIR including placeholder error entries.

### DIFF
--- a/compiler/qsc_data_structures/src/span.rs
+++ b/compiler/qsc_data_structures/src/span.rs
@@ -74,3 +74,18 @@ impl From<Span> for SourceSpan {
         Self::from((value.lo as usize)..(value.hi as usize))
     }
 }
+
+#[allow(clippy::module_name_repetitions)]
+pub trait WithSpan {
+    #[must_use]
+    fn with_span(self, span: Span) -> Self;
+}
+
+impl<T> WithSpan for Box<T>
+where
+    T: WithSpan,
+{
+    fn with_span(self, span: Span) -> Self {
+        Box::new((*self).with_span(span))
+    }
+}
--- a/compiler/qsc_data_structures/src/span.rs
+++ b/compiler/qsc_data_structures/src/span.rs
@@ -80,12 +80,3 @@ pub trait WithSpan {
     #[must_use]
     fn with_span(self, span: Span) -> Self;
 }
-
-impl<T> WithSpan for Box<T>
-where
-    T: WithSpan,
-{
-    fn with_span(self, span: Span) -> Self {
-        Box::new((*self).with_span(span))
-    }
-}
--- a/compiler/qsc_frontend/src/typeck/convert.rs
+++ b/compiler/qsc_frontend/src/typeck/convert.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use core::panic;
-
 use crate::resolve::{self, Names};
 use qsc_ast::ast::{
     self, CallableBody, CallableDecl, CallableKind, FunctorExpr, FunctorExprKind, Ident, Pat,
@@ -150,7 +148,11 @@ pub(super) fn ast_ty_def(names: &Names, def: &TyDef) -> (UdtDef, Vec<MissingTyEr
                     })
                     .collect(),
             ),
-            TyDefKind::Err => panic!("tydef error not expected here"),
+            TyDefKind::Err => UdtDefKind::Field(UdtField {
+                name_span: None,
+                name: None,
+                ty: Ty::Err,
+            }),
         },
     };
 
--- a/compiler/qsc_frontend/src/typeck/convert.rs
+++ b/compiler/qsc_frontend/src/typeck/convert.rs
@@ -75,6 +75,7 @@ pub(crate) fn ty_from_ast(names: &Names, ty: &ast::Ty) -> (Ty, Vec<MissingTyErro
             }
             (Ty::Tuple(tys), errors)
         }
+        TyKind::Err => (Ty::Err, Vec::new()),
     }
 }
 
@@ -109,6 +110,7 @@ fn ast_ty_def_base(names: &Names, def: &TyDef) -> (Ty, Vec<MissingTyError>) {
 
             (Ty::Tuple(tys), errors)
         }
+        TyDefKind::Err => (Ty::Err, Vec::new()),
     }
 }
 
@@ -146,6 +148,11 @@ pub(super) fn ast_ty_def(names: &Names, def: &TyDef) -> (UdtDef, Vec<MissingTyEr
                     })
                     .collect(),
             ),
+            TyDefKind::Err => UdtDefKind::Field(UdtField {
+                name_span: None,
+                name: None,
+                ty: Ty::Err,
+            }),
         },
     };
 
@@ -210,7 +217,7 @@ fn synthesize_functor_params_in_pat(
     pat: &mut hir::Pat,
 ) -> Vec<GenericParam> {
     match &mut pat.kind {
-        hir::PatKind::Discard | hir::PatKind::Bind(_) => {
+        hir::PatKind::Discard | hir::PatKind::Err | hir::PatKind::Bind(_) => {
             synthesize_functor_params(next_param, &mut pat.ty)
         }
         hir::PatKind::Tuple(items) => {
@@ -247,6 +254,7 @@ pub(crate) fn ast_pat_ty(names: &Names, pat: &Pat) -> (Ty, Vec<MissingTyError>) 
             }
             (Ty::Tuple(tys), errors)
         }
+        PatKind::Err => (Ty::Err, Vec::new()),
     }
 }
 
--- a/compiler/qsc_frontend/src/typeck/convert.rs
+++ b/compiler/qsc_frontend/src/typeck/convert.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use core::panic;
+
 use crate::resolve::{self, Names};
 use qsc_ast::ast::{
     self, CallableBody, CallableDecl, CallableKind, FunctorExpr, FunctorExprKind, Ident, Pat,
@@ -148,11 +150,7 @@ pub(super) fn ast_ty_def(names: &Names, def: &TyDef) -> (UdtDef, Vec<MissingTyEr
                     })
                     .collect(),
             ),
-            TyDefKind::Err => UdtDefKind::Field(UdtField {
-                name_span: None,
-                name: None,
-                ty: Ty::Err,
-            }),
+            TyDefKind::Err => panic!("tydef error not expected here"),
         },
     };
 
--- a/compiler/qsc_parse/src/expr/tests.rs
+++ b/compiler/qsc_parse/src/expr/tests.rs
@@ -2181,3 +2181,197 @@ fn nested_interpolated_string_with_exprs() {
                 Lit: " baz""#]],
     );
 }
+
+#[test]
+fn duplicate_commas_in_tuple() {
+    check(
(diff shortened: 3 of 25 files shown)


swernli:
Need to add some new tests for sequences, as well as for signature help.

language_service/src/signature_help/tests.rs
```diff
@@ -257,7 +256,48 @@ fn insert_second_argument() {
...
+                                    start: 35,
+                                    end: 45,
+                                },
+                                documentation: None,
+                            },
+                        ],
+                    },
+                ],
+                active_signature: 0,
+                active_parameter: 3,
```
swernli, line 298: This seems like a bug... shouldn't this be `2` and not `3`? Playing with it in the UI, it does seem to give help for the next item until I start actually typing and then it jumps back to the current one. This may be a bug in how I did `Err` handling in the signature help...

compiler/qsc_parse/src/expr/tests.rs
```diff
@@ -2181,3 +2181,197 @@ fn nested_interpolated_string_with_exprs() {
...
+
+#[test]
+fn duplicate_commas_in_tuple() {
+    check(
+        expr,
+        "(x,, y)",
+        &expect![[r#"
+            Expr _id_ [0-7]: Tuple:
+                Expr _id_ [1-2]: Path: Path _id_ [1-2] (Ident _id_ [1-2] "x")
+                Expr _id_ [0-0]: Err
```
minestarks, line 2193: Haven't looked at the rest of the code yet but I really think these error nodes should have accurate spans. They can be 0-length spans if necessary. That would be consistent with the behavior of other recovering parsers. You also need accurate spans in language service features to determine which AST node the cursor falls into.

swernli: Ah, that was exactly the source of the issue I was seeing with signature help! Thanks for pointing that out. I was able to add a trait that updates the spans such that the right zero-length span shows up in the Err AST members and now signature help behaves as expected.

language_service/src/signature_help/tests.rs
swernli: @minestarks comment on the spans ended up being the key here, and now this is working as expected.

ScottCarda-MS:
Looks great! I noticed two strange behaviors while playing around with this:
 - `let temp = (1,);` shows as having type `(Int,)` in the HIR and `(Int)` in the LS display. I expected it to have type `(Int, ?)`.
 - `let temp = (1,,);` shows as having type `(Int, ?)` in the HIR and LS display. I expecte dit to have type `(Int, ?, ?)`.
I think both of these situations arise from the fact that I think a comma should imply the existence of the following item, and when the item is not given, it should be the new `Err` item.

ScottCarda-MS:
It doesn't need to be part of this PR, but this solution could be scaled to address #749 as well in a later PR.

ScottCarda-MS:
> Looks great! I noticed two strange behaviors while playing around with this:
> 
> * `let temp = (1,);` shows as having type `(Int,)` in the HIR and `(Int)` in the LS display. I expected it to have type `(Int, ?)`.
> * `let temp = (1,,);` shows as having type `(Int, ?)` in the HIR and LS display. I expected it to have type `(Int, ?, ?)`.
>   I think both of these situations arise from the fact that I think a comma should imply the existence of the following item, and when the item is not given, it should be the new `Err` item.

Discussed offline: Decided to keep the notion that trailing commas do not imply a following item. Decided to keep the use of the `(Int,)` syntax for arity-1 tuple type displays to be consistent with the convention seen in other languages. This means that `(Int)` is not a valid display for a type.
There still needs to be a fix for how the language server displays arity-1 tuples to be consistent with this decision.

compiler/qsc_data_structures/src/span.rs
```diff
@@ -74,3 +74,18 @@ impl From<Span> for SourceSpan {
...
+    fn with_span(self, span: Span) -> Self;
+}
+
+impl<T> WithSpan for Box<T>
+where
+    T: WithSpan,
+{
+    fn with_span(self, span: Span) -> Self {
+        Box::new((*self).with_span(span))
+    }
```
minestarks, line 90: Super nit: I think you can avoid a reallocation this way:
```suggestion
    fn with_span(mut self, span: Span) -> Self {
        *self = (*self).with_span(span);
        self
    }
```
You should check my work though, I'm like 90% sure.

compiler/qsc_frontend/src/typeck/convert.rs
```diff
@@ -146,6 +148,11 @@ pub(super) fn ast_ty_def(names: &Names, def: &TyDef) -> (UdtDef, Vec<MissingTyEr
                     })
                     .collect(),
             ),
+            TyDefKind::Err => UdtDefKind::Field(UdtField {
+                name_span: None,
+                name: None,
+                ty: Ty::Err,
+            }),
```
minestarks, line 155: I have no idea how this code is invoked, but this one makes me nervous because the handling looks nontrivial compared to the other `Err` cases. Can we have a unit test case that exercises this specifically?

Also, should we be pushing an error into `errors` when we encounter this node? Or is the idea that the error should have been previously reported for this node anyway?

compiler/qsc_parse/src/prim.rs
```diff
@@ -170,12 +170,29 @@ pub(super) fn many<T>(s: &mut Scanner, mut p: impl Parser<T>) -> Result<Vec<T>>
     Ok(xs)
 }
 
-pub(super) fn seq<T>(s: &mut Scanner, mut p: impl Parser<T>) -> Result<(Vec<T>, FinalSep)> {
+pub(super) fn seq<T>(s: &mut Scanner, mut p: impl Parser<T>) -> Result<(Vec<T>, FinalSep)>
```
minestarks, line 173: Okay, every time I see an iteration of this fix, I ask for more, but...

Would it be possible to also recover from broken syntax, not just _missing_ entries? Like in `foo( , . , -- ,)` . Can `.` and `--` be turned into error expression nodes?

The scenario I have in mind is you start typing `DrawRandomInt(-)` , starting to type a negative integer literal, and signature help disappears.

I feel like this _could_ be achievable by using the recovering parser and `,` as a barrier token. In my experiments, I did manage to do it for expression sequences, but failed to generalize it for other kinds of sequences.

language_service/src/signature_help/tests.rs
```diff
@@ -244,7 +244,6 @@ fn last_argument() {
     );
 }
 
-#[ignore = "Parser needs updating to handle `(1,, \"Four\")`"]
```
minestarks, line 247: I love that there was already a test for this!

minestarks (approved):
I'm loving the commitment to using Err nodes everywhere, I think this sets up nicely to do even richer parser recovery in the future. I added one wishlist item and a couple of nits.

compiler/qsc_parse/src/prim.rs
minestarks: I now see that Scott already pointed to an issue he filed on this above.

compiler/qsc_data_structures/src/span.rs
ScottCarda-MS: I think this adds a requirement that the objects using this method must be mutable, which is not desirable.

swernli: The other option I could explore here is to not use `Default` at all (I know, flies in the face of my PR name). If I instead introduce a new trait called something like `Error` or `AsError` then I can have the option to pass the span as an argument to the contruction of the error instance. And likewise implement it as a kind of constructor here.

compiler/qsc_parse/src/prim.rs
swernli: Yeah, I think these are great suggestions for us to continue down this road, and would prefer to address them in a follow up PR focused on expr recovery.

language_service/src/signature_help/tests.rs
swernli: Thank @ScottCarda-MS for working ahead ;)

compiler/qsc_frontend/src/typeck/convert.rs
swernli: I can add a test for this one to confirm my understanding. And yes, the expectation is that error entries in the AST/HIR should already correspond to errors that were filed during parsing, so there is no need to file an error here. This spot just needs to be able to handle an error in the type definition without failing to produce a type so that we can eventually get the right HIR representing the partially defined type.

swernli: Actually, try as I might, I cannot cause the code to go down this path. I will turn this into a panic rather than leave unused branch in here.

ScottCarda-MS: This should be reachable with `newtype Bar = (, y : Int);`.

ScottCarda-MS: I just confirmed that `newtype Bar = (, y : Int);` throws the added panic, and in the commit before the panic was added, it gave the expected behavior of yielding `newtype Bar = (?, y : Int)`.

swernli: Ah, you're right. I see where my mistake was: I was counting on the tests in the parser which already have a case for this, but they just show AST. This code path produces a `qsc_hir::ty::UdtDef` so it only gets invoked when we are starting to calculate the relevant HIR structures. What this tells me is that I'm missing tests on the HIR error kinds that we expect to see, so I'll add those cases to lower/tests.rs as part of putting back this logic. Good catch Scott!

compiler/qsc_data_structures/src/span.rs
swernli: Ok! I think I found a solution, which ironically involved adding one more trait: `DefaultWithSpan`. I was able to implement that in a way that avoids the extra allocation for `Box` and doesn't require special extra implementation for any of the other types. Turns out only `qir_ast::Ty` was used directly by `seq` without a box, so just that one implementation was needed after imlementing it for `Box`.

minestarks: > I think this adds a requirement that the objects using this method must be mutable, which is not desirable.

@ScottCarda-MS  `self` and `mut self` both consume `self`, so it shouldn't affect anything from the caller's perspective. The `mut` only matters to the implementation - it's there so that you can assign to `*self` (which you already own) without the compiler yelling at you. (I did find this a bit perplexing at first. I couldn't find a good documentation link for it, but the OP in this thread sums up my reaction to it. https://users.rust-lang.org/t/whats-the-point-of-arguments-with-mutable-ownership/23315)

@swernli did `DefaultWithSpan` have to be a trait? Why not have a standalone function like

```rs
fn default_with_span<T>(span: Span) -> T where T: Default + WithSpan {
    Box::new(T::default().with_span(span))
}

// call site
default_with_span::<T>(span);
```

Is one of those more idiomatic than the other?


I know the PR is merged (yay!) but just responding for the sake of closure.

swernli: Oh, I misunderstood what you were suggesting before, and you are right, this did not need to be a trait. The impl for `Box<T>` can have `mut self` in the signature without requiring the signature for the trait itself to change. I've put this up in a small PR #852. Thanks for following up and asking more about this!

PR #843, opened 2023-11-07
Allow reuse of unmeasured qubits in QIR codegen

swernli:
The previous change to disallow qubit reuse was too aggresive with how it used fresh qubit ids, causing programs to get too big too fast. The problem was the assumption that a qubit that is released can never be used again. Instead, this updates the logic to ensure that qubits that are released without ever having been measured or reset can be reused in later allocations. As an example, this takes the Grover's Search sample program from using 29 qubits to using 9 by allowing reuse of unmeasing auxiliary qubits used during decomposition of multi-controlled gates.

### DIFF
--- a/compiler/qsc_codegen/src/qir_base.rs
+++ b/compiler/qsc_codegen/src/qir_base.rs
@@ -121,6 +121,7 @@ pub(super) fn get_global(
 pub struct BaseProfSim {
     next_meas_id: usize,
     next_qubit_id: usize,
+    next_qubit_hardware_id: usize,
     qubit_map: IndexMap<usize, usize>,
     instrs: String,
     measurements: String,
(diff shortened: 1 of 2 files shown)


idavis (approved):

idavis (approved):

idavis (approved):

compiler/qsc_codegen/src/qir_base.rs
```diff
@@ -474,12 +475,12 @@ impl Backend for BaseProfSim {
...
         let id = self.next_qubit_id;
         self.next_qubit_id += 1;
-        self.qubit_map.insert(id, id);
+        let _ = self.map(id);
         id
     }
 
     fn qubit_release(&mut self, _q: usize) {
-        // Base Profile qubits are never released, since they cannot be reused.
+        self.next_qubit_id -= 1;
```
minestarks, line 483: Are qubits guaranteed to be released in reverse order to allocation?

minestarks (approved):

compiler/qsc_codegen/src/qir_base.rs
swernli: Yes, as long as the caller is using qubit initialization statements, ie `use q = Qubit();`. We guarantee this by the way the `replace_qubit_allocation` pass works, since it creates an implied scope for each qubit init statement.

PR #844, opened 2023-11-08
The configured target is now passed to the debugger for compilation

### LINKED ISSUE #824: Debugger should respect selected execution profile
**Describe the bug**

When debugging a Q# program, the execution profile stays configured to `Full` even when `Base` is selected.

idavis:
Fixes #824

### DIFF
--- a/npm/src/debug-service/debug-service.ts
+++ b/npm/src/debug-service/debug-service.ts
@@ -22,6 +22,7 @@ export interface IDebugService {
   loadSource(
     path: string,
     source: string,
+    target: string,
     entry: string | undefined,
   ): Promise<string>;
   getBreakpoints(path: string): Promise<IBreakpointSpan[]>;
@@ -65,10 +66,11 @@ export class QSharpDebugService implements IDebugService {
   async loadSource(
     path: string,
     source: string,
+    target: string,
     entry: string | undefined,
   ): Promise<string> {
     this.code[path] = source;
-    return this.debugService.load_source(path, source, entry);
+    return this.debugService.load_source(path, source, target, entry);
   }
 
   async getStackFrames(): Promise<IStackFrame[]> {
--- a/npm/src/debug-service/debug-service.ts
+++ b/npm/src/debug-service/debug-service.ts
@@ -22,7 +22,7 @@ export interface IDebugService {
   loadSource(
     path: string,
     source: string,
-    target: string,
+    target: "base" | "full",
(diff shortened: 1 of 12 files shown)


swernli (approved):

wasm/src/debug_service.rs
```diff
@@ -31,12 +31,27 @@ impl DebugService {
...
             [(path.into(), source.into())],
             entry.as_deref().map(|value| value.into()),
         );
-        match Interpreter::new(true, source_map, qsc::PackageType::Exe, TargetProfile::Full) {
+        let target = match target_profile.as_str() {
+            "base" => TargetProfile::Base,
+            "full" => TargetProfile::Full,
+            _ => {
+                return render_error(qsc::interpret::stateful::Error::UnknownTarget(
+                    target_profile.to_string(),
```
minestarks, line 50: I think I'd rather sanitize this value before we call into WASM, and panic here if targetprofile is an invalid value. It doesn't feel worth adding a specific Interpreter error for, since getting the wrong value here would indicate a bug in _our_ code.

That being said, there are already 3 places where we read this value from settings and they all handle invalid values differently...

For display in the status bar, Defaults to "QIR:Invalid" for invalid value.
https://github.com/microsoft/qsharp/blob/b74867ceb788f490b95e6bb8a552faabc05c19ee/vscode/src/statusbar.ts#L58-L70

For configuring the language service, emits a warning but doesn't sanitize (this seems like a bug).
https://github.com/microsoft/qsharp/blob/b74867ceb788f490b95e6bb8a552faabc05c19ee/vscode/src/extension.ts#L228-L242

For QIR generation - just checks for "base" and ignores invalid values
https://github.com/microsoft/qsharp/blob/b74867ceb788f490b95e6bb8a552faabc05c19ee/vscode/src/qirGeneration.ts#L32-L33C13

This may be a good opportunity to refactor these into a utility function that sanitizes the value so we don't have to deal with handling invalid values everywhere?

npm/src/debug-service/debug-service.ts
```diff
@@ -22,6 +22,7 @@ export interface IDebugService {
   loadSource(
     path: string,
     source: string,
+    target: string,
```
minestarks, line 25: It would be nice to strongly type this as "base" | "full" so that you force the caller to sanitize the value before calling. (See my other comment about this)

minestarks (approved):

Issue #797, opened 2023-10-16
Multi-File: Debugging Support


sezna:
#794 introduces Q# projects. We need to update the debugger to understand projects and include the correct source files on compilation, and ensure jumping between files in the debugger works correctly.

PR #849, opened 2023-11-09
Project Support in the VS Code Extension

### LINKED ISSUE #796: Multi-File: Project Support in the Extension
#794 introduces Q# projects. We need to update the extension to understand projects and include the correct source files on compilation.

sezna:
# Project Support for the Language Service

This PR adds multi-file projects to the language service. Closes #796 

## Overview

`qsc_project` defines a project system that sits on top of a given file system-like interface. This PR is an implementation of that interface for the language service. This is a non-trivial implementation, as it has cascading effects. The most prominent of these effects is the asyncification of the language service, both in WASM and in TS.

## Async Information

There is some key behavior related to the asynchronous functions that needs to be highlighted.
- We only trigger one compilation at a time, storing an optional "next" compilation to catch up to the editor state. This prevents race conditions when we fire off a compile event every keystroke.  
- The WASM now uses `async`/`await` in Rust. It is important to know that these Rust `Future`s are equivalent to Javscript `Promise`s. They share the same event loop scheduler. `wasm_bindgen` handles the sharing of the async runtime between v8 and Rust, and it is indeed magical.
- `tokio` was added to the workspace dependencies. This is a bummer, but it is only used for running the unit tests in the language service. We do not use tokio's runtime anywhere in our non-test code.
- I've done my best to architect this in such a way that it is easy to debug. That being said, there are some rough edges. I will schedule some time to go over those rough edges.

## Debugger
This PR does not introduce debugger support for projects. We will add debugger support in a future PR. For now, this PR disallows running the debugger if the current context is a Q# project. See #797.
image

## Follow-On Tasks
Below are the follow-on tasks that will come in future PRs:
- Using VS Code's FS watcher API to notify us of file updates, and caching files.
- Redesigning the manifest format and `excludeFiles`/`excludeRegexes`
- Adding project support to the Python API (#861)

### DIFF
--- a/wasm/src/language_service.rs
+++ b/wasm/src/language_service.rs
@@ -7,7 +7,7 @@ use crate::{diagnostic::VSDiagnostic, serializable_type};
 use js_sys::JsString;
 use qsc::{self};
 use qsc_project::{EntryType, Manifest, ManifestDescriptor};
-use qsls::JSFileEntry;
+use qsls::{protocol::DiagnosticUpdate, JSFileEntry};
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
@@ -168,23 +168,25 @@ impl LanguageService {
             .expect("expected a valid JS function")
             .clone();
 
+        let diagnostics_callback = move |update: DiagnosticUpdate| {
+            let diags = update
+                .errors
+                .iter()
+                .map(|err| VSDiagnostic::from_compile_error(&update.uri, err))
+                .collect::<Vec<_>>();
+            let _ = diagnostics_callback
+                .call3(
+                    &JsValue::NULL,
+                    &update.uri.into(),
+                    &update.version.into(),
+                    &serde_wasm_bindgen::to_value(&diags)
+                        .expect("conversion to VSDiagnostic should succeed"),
+                )
+                .expect("callback should succeed");
+        };
+
         let inner = qsls::LanguageService::new(
-            move |update| {
-                let diags = update
-                    .errors
-                    .iter()
-                    .map(|err| VSDiagnostic::from_compile_error(&update.uri, err))
-                    .collect::<Vec<_>>();
-                let _ = diagnostics_callback
-                    .call3(
-                        &JsValue::NULL,
-                        &update.uri.into(),
-                        &update.version.into(),
-                        &serde_wasm_bindgen::to_value(&diags)
-                            .expect("conversion to VSDiagnostic should succeed"),
-                    )
-                    .expect("callback should succeed");
-            },
+            diagnostics_callback,
             read_file,
             list_directory,
             get_manifest,


package.json
```diff
@@ -44,6 +44,7 @@
     "preact": "^10.18.1",
     "prettier": "^3.0.3",
     "typescript": "^5.2.2",
-    "url": "^0.11.3"
+    "url": "^0.11.3",
+    "vscode-uri": "^3.0.8"
```
minestarks, line 48: You shouldn't need this dependency - `Uri` is already part of the [VS Code API](https://code.visualstudio.com/api/references/vscode-api#Uri). You can just import it from `vscode`

npm/src/language-service/language-service.ts
```diff
@@ -130,28 +161,29 @@ export class QSharpLanguageService implements ILanguageService {
...
+    await this.languageService.close_notebook_document(documentUri, cellUris);
   }
 
   async getCompletions(
     documentUri: string,
     offset: number,
   ): Promise<ICompletionList> {
-    const code = this.code[documentUri];
-    if (code === undefined) {
+    const code = await this.loadFile(documentUri);
```
minestarks, line 184: Was there a specific reason to change these to file reads? I don't see why it should be necessary - the source code contents should have already been populated in a prior `updateDocument` call. I'd prefer to go back to the existing map lookup. 


You can, however, add these `loadFile` calls to https://github.com/microsoft/qsharp/blob/17cc2d253522f0d3d57f5b51f1a21ac89c414823/npm/src/language-service/language-service.ts#L260-L264 and similar locations in `getReferences` and `getRename`, since those are the points where we're about to navigate to an *unopened* file and we need the file contents.

minestarks: Oh, I just noticed `loadFile` does the map lookup first! I think that's alright, but then you'll want to update the remaining spots in this file that do the `this.code[uri]` lookups. You missed a few (probably added after you made the changes)...

npm/src/language-service/language-service.ts
```diff
@@ -99,15 +99,46 @@ export class QSharpLanguageService implements ILanguageService {
...
+      listDir,
+      getManifest,
     );
+    this.readFile = readFile;
+  }
+
+  async loadFile(uri: string): Promise<string | null> {
+    const result = this.code[uri];
+    if (result === undefined || result === null) {
+      await this.readFile(uri);
```
minestarks, line 127: Are you actually returning the result from `readFile`?

npm/src/language-service/worker-browser.ts
```diff
@@ -23,7 +23,13 @@ export function messageHandler(e: MessageEvent) {
       {
         log.setLogLevel(data.qscLogLevel);
         wasm.initSync(data.wasmModule);
-        const languageService = new QSharpLanguageService(wasm);
+        // we pass in dummy arguments for `readFile` and `listDir` because the browser worker doesn't support the project system right now
+        const languageService = new QSharpLanguageService(
+          wasm,
+          () => Promise.resolve(null),
+          () => Promise.resolve([]),
+          () => Promise.resolve(null),
```
minestarks, line 31: I wonder if we should have `getManifest` return `null` like you did, but have `readFile` and `listDir` fail instead. If we ever try to do file i/o from the website/playground, it's a bug, and failing fast would help us catch the error quicker.

npm/src/main.ts
```diff
@@ -96,9 +96,19 @@ export function getDebugServiceWorker(): IDebugServiceWorker {
   return proxy;
 }
 
-export function getLanguageService(): ILanguageService {
+export function getLanguageService(
+  readFile: (uri: string) => Promise<string | null> = () =>
+    Promise.resolve(null),
```
minestarks, line 101: I see you declare these defaults in a bunch of places - how about moving the default arguments to the `QSharpLanguageService` constructor (which ends up being the common code path), declaring the arguments as optional (i.e. using `?` in the argument list) here, and just omitting them in `worker-browser` and `worker-node`? That way you'll only end up defining the defaults in one place.

npm/src/compiler/compiler.ts
```diff
@@ -43,14 +53,32 @@ export class Compiler implements ICompiler {
     globalThis.qscGitHash = this.wasm.git_hash();
   }
 
-  async checkCode(code: string): Promise<VSDiagnostic[]> {
+  async checkCode(
```
minestarks, line 56: I think I'd leave these out of the `checkCode` signature and just pass the defaults to the constructor below... This function is used to quickly check a piece of Q# code, and doesn't really make sense to use in a project scenario. (I may be proven wrong in the future when we use this with projects, but maybe we can just add the parameters then?)

vscode/src/extension.ts
```diff
@@ -145,14 +150,33 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
     }),
   );
 
-  function updateIfQsharpDocument(document: vscode.TextDocument) {
+  let currentlyCompiling = false;
+  let pendingCompile: vscode.TextDocument | null = null;
```
minestarks, line 154: Hmm, I'd definitely want to push this logic down to the Rust layer, leaving this code mostly as is. Basically, these bools can live in `language_service/src/lib.rs` instead. When the code is pending a compile, the language service requests (getCompletions etc) could yield empty results. The "recompile if there are pending changes" behavior can also be handled in the Rust layer.

This would keep the complexity in the language service itself, where it's more appropriate. It would also make the service behave correctly without requiring each client (vscode, playground, and maybe jupyterlab/lsp clients one day) to each implement their own logic.

vscode/src/projectSystem.ts
```diff
@@ -0,0 +1,124 @@
...
+async function findManifestDocument(
+  currentDocumentUriString: string,
+): Promise<{ uri: vscode.Uri; content: string } | null> {
+  // /home/foo/bar/document.qs
+  const currentDocumentUri = URI.parse(currentDocumentUriString);
+
+  // /home/foo/bar
+  let uriToQuery = Utils.dirname(currentDocumentUri);
+
+  let attempts = 100;
```
minestarks, line 66: We should stop at the VSCode workspace root instead of going all the way up to the drive root. Users do silly things -- forgetting a qsharp.json at the drive/source root and then wondering why we're crawling their entire disk as soon as they open a Q# document, is definitely going to be one of those things.

vscode/src/projectSystem.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (oldUriToQuery === uriToQuery) {
+      return null;
+    }
+  }
+  return null;
+}
+
+// this function currently assumes that `directoryQuery` will be a relative path from
+// the root of the workspace
+export async function directoryListingCallback(
```
minestarks, line 94: Nit: I don't love the name "callback" here (and in the below function). Maybe they can just be called `listDir` and `readFile` to match the names you use everywhere else? (And you could rename the other `readFile` below to something unique, like, I dunno `readFileWithUri`. Not too fussed about that one, since it's private to the module)

wasm/src/language_service.rs
minestarks, line 1: Holy moly, this file. Is there anything in here that can be extracted into helper functions? Just pulling the tiniest bit of repetitive code into utility functions would help readability a lot.

language_service/src/lib.rs
```diff
@@ -21,18 +22,21 @@ mod test_utils;
...
 use qsc::{compile::Error, PackageType, TargetProfile};
+use qsc_project::FileSystemAsync;
 use rustc_hash::{FxHashMap, FxHashSet};
-use std::{mem::take, sync::Arc};
+use std::{future::Future, mem::take, path::PathBuf, pin::Pin, sync::Arc};
 
 type CompilationUri = Arc<str>;
 type DocumentUri = Arc<str>;
 
+#[allow(clippy::type_complexity)]
```
minestarks, line 39: I agree with clippy here :D Surely some of these cmoplex types can be aliased?

language_service/src/lib.rs
```diff
@@ -336,24 +397,25 @@ impl<'a> LanguageService<'a> {
...
+    async fn publish_diagnostics_for_doc(&self, uri: &str, errors: Vec<Error>) {
         let version = self.open_documents.get(uri).map(|d| d.version);
         trace!("publishing diagnostics for {uri} {version:?}): {errors:?}");
         (self.diagnostics_receiver)(DiagnosticUpdate {
             uri: uri.into(),
             version,
             errors,
-        });
+        })
+        .await;
```
minestarks, line 418: Was there a reason to make the diagnostics callback async, or is it just for overall consistency? This is essentially just an event - we don't care what the receiver does with it and there is no return value. There is no reason to await it. Unless you needed it for some reason I'd rather it stay a synchronous function.

language_service/src/lib.rs
```diff
@@ -304,7 +365,7 @@ impl<'a> LanguageService<'a> {
             .open_documents
             .get(uri)
             .unwrap_or_else(|| {
-                panic!("{op_name} should not be called for a document that has not been opened",)
+                panic!("{op_name} (called on {uri} should not be called for a document that has not been opened",)
```
minestarks, line 368: Missed a close paren here

language_service/src/lib.rs
```diff
@@ -121,40 +145,77 @@ impl<'a> LanguageService<'a> {
...
+    /// This is the "entry point" for the language service's logic, after its constructor.
+    ///
     /// LSP: textDocument/didOpen, textDocument/didChange
-    pub fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+    pub async fn update_document(&mut self, uri: &str, version: u32, text: &str) {
         trace!("update_document: {uri} {version}");
-        let compilation = Compilation::new_open_document(
-            uri,
-            text,
+        let manifest = (self.get_manifest)(uri.to_string()).await;
```
minestarks, line 153: Have you thought about whether we might want to cache the `get_manifest` call to avoid paying the penalty of walking up the directory structure on every keystroke, even when we're not using projects? It seems like it would be a good idea, though it might require some care to make sure it doesn't get out of date. Maybe you're already doing this somewhere else and I missed it.

wasm/src/language_service.rs
```diff
@@ -1,44 +1,231 @@
...
+                )
+            })
+            .clone();
+
+        let read_file = move |path_buf: PathBuf| {
+            let path_buf_string = &path_buf.to_string_lossy().to_string();
+            let path = JsValue::from_str(path_buf_string);
+            let res: js_sys::Promise = read_file
+                .call1(&JsValue::NULL, &path)
+                .expect("callback should succeed")
```
minestarks, line 51: Hmm, we really shouldn't panic on file i/o errors - we should have a strategy to recover gracefully. One easy way to hit i/o errors is if I'm deleting files in the workspace while a project recompilation is going on in the background. Especially once you hookup file watchers - bulk deletes will trigger project updates, project update will cause a mass of file reads, those file reads will fail, etc...

language_service/src/name_locator.rs
```diff
@@ -127,15 +127,15 @@ impl<'inner, 'package, T> Locator<'inner, 'package, T> {
 impl<'inner, 'package, T: Handler<'package>> Visitor<'package> for Locator<'inner, 'package, T> {
     fn visit_namespace(&mut self, namespace: &'package ast::Namespace) {
         if span_contains(namespace.span, self.offset) {
-            self.context.current_namespace = namespace.name.name.clone();
+            self.context.current_namespace = Rc::from(&*namespace.name.name);
```
minestarks, line 130: Did this changes unintentionally sneak in?

language_service/src/compilation.rs
```diff
@@ -125,13 +128,14 @@ impl Compilation {
...
+            .into_iter()
+            .map(|(a, b)| (Arc::from(a), Arc::from(b)))
+            .collect();
 
         let new = match self.kind {
-            CompilationKind::OpenDocument => {
-                assert!(sources.len() == 1);
-                Self::new_open_document(sources[0].0, sources[0].1, package_type, target_profile)
-            }
+            CompilationKind::OpenDocument => Self::new(&sources, package_type, target_profile),
```
minestarks, line 138: I think you'll want to update the doc comment (and possibly name) for `OpenDocument` to reflect the fact that it applies to more than single-file projects

minestarks: Maybe just `Project`?

language_service/src/project_system.rs
```diff
@@ -0,0 +1,51 @@
...
+            .unwrap_or_default()
+    }
+
+    fn entry_name(&self) -> String {
+        self.name.clone()
+    }
+
+    fn path(&self) -> PathBuf {
+        PathBuf::from(&self.name)
+    }
```
minestarks, line 35: Isn't `entry_name` supposed to return just the filename portion while the `path` returns the full path including the directory query (at least that's what I see [here](https://github.com/microsoft/qsharp/blob/04e551beed2c89c9475be4bb5d6ff3907bae5039/compiler/qsc_project/src/fs.rs#L29-L35))? It seems inconsistent that these two implementations return the same data.

compiler/qsc_project/src/fs.rs
```diff
@@ -13,13 +16,46 @@ use std::{path::PathBuf, sync::Arc};
...
+            .unwrap_or_default()
+    }
+
+    fn entry_name(&self) -> String {
+        self.file_name()
+            .expect("canonicalized symlink cannot end in relative path")
+            .to_string_lossy()
+            .to_string()
+    }
+
```
minestarks, line 46: Do we actually need `entry_extension` and `entry_name` on the trait? Looks like all three implementations do the same thing - call the corresponding method on `PathBuf`. The caller can just retrieve the `PathBuf` and do this work itself, instead of us having 3 duplicate implementations.

vscode/src/projectSystem.ts
billti: Can confirm. Have been bitten by this a few times in other ecosystems (e.g. wrestling with weirdness in a 'manifest-free' project in ~/scratch/foo for hours, only to later realize I had a manifest in ~/scratch from a prior experiment that was getting found and pulling in other stuff).

compiler/qsc_project/src/manifest.rs
```diff
@@ -16,16 +16,20 @@ use std::path::PathBuf;
...
     pub license: Option<String>,
-    #[serde(default)]
+    #[serde(default = "default_exclude_regexes")]
     pub exclude_regexes: Vec<String>,
     #[serde(default)]
     pub exclude_files: Vec<String>,
 }
 
+fn default_exclude_regexes() -> Vec<String> {
+    vec![".*node_modules.*".into(), ".*\\.git.*".into()]
```
minestarks, line 30: These shouldn't just be the default but they should be always added, IMO. Otherwise as soon as the user adds an exclude regex we'll start crawling .git again which isn't an ideal user experience.

language_service/src/project_system.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[async_trait(?Send)]
+impl qsc_project::FileSystemAsync for LanguageService<'_> {
+    type Entry = JSFileEntry;
+    async fn read_file(
+        &self,
+        path: &std::path::Path,
+    ) -> miette::Result<(std::sync::Arc<str>, std::sync::Arc<str>)> {
+        Ok((self.read_file_callback)(path.into()).await)
```
minestarks, line 45: I think I added a comment elsewhere about this, but I think these should propagate file i/o errors.

minestarks (changes requested):
Looking good! Changes I'd like to see are in the comments - mainly I think we need a lot more tests. Pushing the `pendingCompile` etc handling down to the rust layer will allow you to thoroughly test the finicky async parts in the Rust unit tests, which will be nice!

vscode/src/projectSystem.ts
billti: One thing to be aware of, is when opening stuff in VS Code online, such as a virtual file system (e.g. our playground) or a Git repo, then the root of that opened workspace can appear as the path "/".

language_service/src/lib.rs
minestarks: Still, as far as the _language service_ is concerned, it doesn't care about the async-ness. The client can queue up a bunch of work or do whatever it wants in the background. We shouldn't need to await its results. If we do, a bunch of things in the LS get gated behind work we don't care about the results of - we can do better than that.

Did this cause issues in the product or just the tests? If it's a test-only concern I think we can address it in the tests itself (I can actually look into this, since I wrote the diagnostics callbacks in the tests, which I remember being a tearing-my-hair-out process to get the synchronization just right).

minestarks: > This was primarily for test failures. You're suggesting we just call the promise but never await it, and let it execute when the event loop gets around to it?

Yep that's what I'm suggesting.

language_service/src/lib.rs
```diff
@@ -121,27 +162,81 @@ impl<'a> LanguageService<'a> {
...
     /// It should also be called whenever the source code is updated.
     ///
+    /// This is the "entry point" for the language service's logic, after its constructor.
+    ///
     /// LSP: textDocument/didOpen, textDocument/didChange
-    pub fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+    #[async_recursion(?Send)]
+    pub async fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+        if self.currently_updating {
+            self.pending_update = Some(PendingUpdate {
```
minestarks, line 171: I think you need this to be a queue, not just a single item, otherwise you may miss edits. We could skip multiple updates to the same `uri` though.

minestarks:
Do we ever unload a project when you close the last document?

language_service/src/lib.rs
```diff
@@ -121,27 +162,81 @@ impl<'a> LanguageService<'a> {
     /// for the document, typically when the document is first opened in the editor.
     /// It should also be called whenever the source code is updated.
     ///
+    /// This is the "entry point" for the language service's logic, after its constructor.
+    ///
     /// LSP: textDocument/didOpen, textDocument/didChange
-    pub fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+    #[async_recursion(?Send)]
+    pub async fn update_document(&mut self, uri: &str, version: u32, text: &str) {
+        if self.currently_updating {
```
minestarks, line 170: I can't leave a comment near `document_op`, but you'll want to check this flag within `document_op` and return a default result if we're currently compiling. See: https://github.com/microsoft/qsharp/commit/89e2248316dea51756d353d8e47fd79169e6bd37

Alternatively, instead of checking this flag, you could return a default if the document doesn't exist in the compilation map. The former is more strict about our assumptions, the latter is more likely to gloss over bugs and "just work".

```rs
 trace!("{op_name}: uri: {uri}, offset: {offset}");
        let Some(compilation_uri) = &self
            .open_documents
            .get(uri)
            .as_ref()
            .map(|x| x.compilation.clone())
        else {
            trace!("Skipping {op_name} since compilation is in progress");
            return T::default();
        };
```

minestarks:
> @minestarks
> 
> > Do we ever unload a project when you close the last document?
> 
> No. Do you think we should?

_I_ think we should, but it's a design decision we can defer. My concerns are: 

- User annoyance: they can't make errors go away from the problems view no matter how many documents they close. 
- Perf: Memory usage and unnecessary file watcher activity (once we have file watchers) for a project that's not being used.

I don't think these are real concerns for now, so okay to defer (maybe track with an issue).

minestarks (approved):

compiler/qsc_project/src/manifest.rs
billti: We decided to get rid of excludes (and RegExes), right?

compiler/qsc_project/src/project.rs
```diff
@@ -1,15 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use crate::manifest::ManifestDescriptor;
+use regex_lite::Regex;
```
billti, line 5: Still needed in light of recent discussions?

compiler/qsc_project/src/project.rs
```diff
@@ -18,20 +16,107 @@ pub struct Project {
...
+        exclude_patterns: &[Regex],
+        exclude_files: &[String],
+        initial_path: &Path,
+    ) -> miette::Result<Vec<Self::Entry>> {
+        let listing = self.list_directory(initial_path).await?;
+        let mut files = vec![];
+        for item in listing {
+            let file_name = item.entry_name();
+            let name = item.path().to_string_lossy().to_string();
+            if regex_matches(exclude_patterns, &name) || exclude_files.contains(&file_name) {
```
billti, line 78: Exclude patterns?

compiler/qsc_project/tests/tests.rs
```diff
@@ -121,7 +121,10 @@ fn empty_manifest() {
                 manifest: Manifest {
                     author: None,
                     license: None,
-                    exclude_regexes: [],
+                    exclude_regexes: [
```
billti, line 124: At the risk of repeating myself....

vscode/src/projectSystem.ts
```diff
@@ -0,0 +1,143 @@
...
+ * returns null otherwise
+ */
+async function findManifestDocument(
+  currentDocumentUriString: string,
+): Promise<{ uri: vscode.Uri; content: string } | null> {
+  // /home/foo/bar/document.qs
+  const currentDocumentUri = URI.parse(currentDocumentUriString);
+
+  // /home/foo/bar
+  let uriToQuery = Utils.dirname(currentDocumentUri);
```
billti, line 64: Does this return JUST the directory portion? Do you lose the scheme and authority from the Uri in the process? Being the result of this is what you build the URIs to query later on, are you sure this works when the scheme and authority aren't important information? (Such as in a vscode.dev virtual file system, when opening a github repo by pressing '.', etc.)

vscode/src/projectSystem.ts
```diff
@@ -0,0 +1,143 @@
...
+  };
+}
+
+/** Returns the manifest document if one is found
+ * returns null otherwise
+ */
+async function findManifestDocument(
+  currentDocumentUriString: string,
+): Promise<{ uri: vscode.Uri; content: string } | null> {
+  // /home/foo/bar/document.qs
```
billti, line 60: It may be worth noting here some of the more complex URIs you'll need to deal with beyond just a file system path. For example, in github.dev when looking at files that URI can be something like `vscode-vfs://github%2B7b2276223a312c22726566223a7b2274797065223a332c226964223a22383439227d7d/microsoft/qsharp/samples/shor.qs`. The scheme, authority, and path are all critical to maintain in future queries.

wasm/src/language_service.rs
```diff
@@ -1,40 +1,194 @@
...
+        };
+        $js_async_fn
+    }};
+}
+
+async fn map_js_promise<F, T>(res: JsFuture, func: F) -> T
+where
+    F: Fn(JsValue) -> T,
+{
+    let res = res.await.expect("js future shouldn't throw an exception");
```
minestarks, line 46: Doesn't have to happen in this PR but IMO it would be more appropriate to return the failure here and have the _caller_  do the `.expect()` if it wants to. Since this is a generic wrapper function it makes me nervous that it makes assumptions about the fallibility of the wrapped JS function.

wasm/src/language_service.rs
```diff
@@ -1,40 +1,194 @@
...
 impl LanguageService {
     #[wasm_bindgen(constructor)]
-    pub fn new(diagnostics_callback: DiagnosticsCallback) -> Self {
+    pub fn new(
+        diagnostics_callback: DiagnosticsCallback,
+        read_file: ReadFileCallback,
+        list_directory: ListDirectoryCallback,
+        get_manifest: GetManifestCallback,
+    ) -> Self {
+        let transformer = move |js_val: JsValue, path_buf_string: String| match js_val.as_string() {
```
cesarzc, line 71: It would be nice to have a comment for each transformer function briefly describing what they do.

cesarzc: Thinking about it a bit more, it could help readability to have each transformer function as a separate utility function rather than a closure (even if they're only used inside `new`). That way, readers can focus on what `new` does rather than on understanding what the transformer functions do. It would also help by giving the transformer functions meaningful names.

vscode/src/projectSystem.ts
billti: And for example in our https://vscode.dev/quantum/playground/ experience, where document URIs are like `qsharp-vfs://playground/samples/Entanglement.qs`), which is our own scheme/file-system ("qsharp-vfs") and authority ("playground") before the path ("/samples/Entanglement.qs").

wasm/src/language_service.rs
```diff
@@ -1,40 +1,194 @@
...
 use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::JsFuture;
 
 #[wasm_bindgen]
 pub struct LanguageService(qsls::LanguageService<'static>);
 
+/// This macro calls an async JS function, awaits it, and then applies a transformer function to it. Ultimately, it returns a function that accepts a String
+/// and returns a JS promise that is represented by a Rust future. (I know. Ouch. My head.)
+macro_rules! call_async_js_fn {
+    ($js_async_fn: ident, $transformer: expr) => {{
```
minestarks, line 22: The name and doc comment are throwing me off here. This code doesn't actually _call_ the function, it just wraps it in a new function that calls it (and does other stuff), right? 

I think the correct name for this would be `async_js_fn_into_async_rust_fn`, but that sounds ridiculously long, so maybe just `into_async_rust_fn`?

`transformer` can be called `map_result` . Since that's basically doing. It's mapping the result value to another value.

wasm/src/language_service.rs
```diff
@@ -1,40 +1,194 @@
...
+        let $js_async_fn = move |input: String| {
+            let path = JsValue::from_str(&input);
+            let res: js_sys::Promise = $js_async_fn
+                .call1(&JsValue::NULL, &path)
+                .expect("callback should succeed")
+                .into();
+
+            let res: JsFuture = res.into();
+
+            let input = input.clone();
```
minestarks, line 34: Why clone the input here? This behavior feels very specific to `read_file`, since `read_file` basically returns its input argument as output (forgot why), but for a generic wrapper macro, it seems arbitrary.

If you feel that the use of this macro will be limited to this file, it's probably ok. If we want to make it a shared utility I'd prefer to clean it up a bit.

minestarks: Oh also I believe you're calling `.clone()` twice? What happens if you remove the `.clone()` here and let the closure below own `input`?

wasm/src/language_service.rs
```diff
@@ -1,40 +1,194 @@
...
-                    &update.uri.into(),
-                    &update.version.into(),
-                    &serde_wasm_bindgen::to_value(&diags)
-                        .expect("conversion to VSDiagnostic should succeed"),
-                )
-                .expect("callback should succeed");
-        });
+
+        let inner = qsls::LanguageService::new(
+            move |update| {
```
cesarzc, line 172: I believe it would also help readability to move the diagnostics callback closure out of the call to `new` so that it's super clear to see how `inner` gets constructed and what the final `diagnostics_callback` does e.g.:
```rust
let diagnostics_callback = move |update| {
    let diags = update
        .errors
        .iter()
        .map(|err| VSDiagnostic::from_compile_error(&update.uri, err))
        .collect::<Vec<_>>();
    let _ = diagnostics_callback
        .call3(
            &JsValue::NULL,
            &update.uri.into(),
            &update.version.into(),
            &serde_wasm_bindgen::to_value(&diags)
                .expect("conversion to VSDiagnostic should succeed"),
        )
        .expect("callback should succeed");
};

let inner = qsls::LanguageService::new(
    diagnostics_callback,
    read_file,
    list_directory,
    get_manifest,
);
```

language_service/src/tests.rs
```diff
@@ -636,13 +636,50 @@ fn new_language_service(received: &RefCell<Vec<ErrorInfo>>) -> LanguageService<'
...
+    // a document hasn't fully loaded
+    let _ = ls.update_document(
+        "foo.qs",
+        1,
+        "namespace Foo { open Microsoft.Quantum.Diagnostics; @EntryPoint() operation Main() : Unit { DumpMachine() } }",
+    );
+
+    // this should be empty, because the doc hasn't loaded
+    assert!(ls.get_completions("foo.qs", 76).items.is_empty())
+}
```
minestarks, line 665: You may want to hold on to the future from `update_document` and `await` on it here, just to make sure the _test_ waits for its completion.

npm/src/browser.ts
```diff
@@ -162,9 +162,17 @@ export function getCompilerWorker(workerArg: string | Worker): ICompilerWorker {
   return proxy;
 }
 
-export async function getLanguageService(): Promise<ILanguageService> {
+export async function getLanguageService(
+  readFile?: (uri: string) => Promise<string | null>,
+  listDir?: (uri: string) => Promise<[string, number][]>,
+  getManifest?: (uri: string) => Promise<{
+    excludeFiles: string[];
```
cesarzc, line 169: I was under the impression that excluding files and regexes was not something we would do during the first version. Should this just get the manifest directory?

PR #850, opened 2023-11-10
LS Support for Type Parameters

ScottCarda-MS:
Adds LS support for handling type parameters: go to definition, find all references, rename, and on-hover info.


language_service/src/rename/tests.rs
```diff
@@ -326,3 +326,25 @@ fn no_rename_std_udt_return_type() {
     "#,
     );
 }
+
+#[test]
+fn ty_param_def() {
+    check(
+        r#"
+        namespace Test {
+            operation Foo<'◉↘T◉>(x : '◉T◉) : '◉T◉ { x }
```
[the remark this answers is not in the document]
ScottCarda-MS: That's part of the testing infrastructure. The ↘ tells the test where the user's cursor is, and the ◉ tells the test where the expected spans are for renaming.

language_service/src/definition.rs
```diff
@@ -59,6 +59,28 @@ impl<'a> Handler<'a> for DefinitionFinder<'a> {
...
+        &mut self,
+        context: &LocatorContext<'a>,
+        _: &'a ast::Ident,
+        param_id: hir::ty::ParamId,
+    ) {
+        if let Some(curr) = context.current_callable {
+            if let Some(def_name) = curr.generics.get(usize::from(param_id)) {
+                self.definition = Some(protocol_location(self.compilation, def_name.span, None));
+            }
+        }
```
minestarks, line 81: Symmetry: Somehow it bothers me that the `at_local_ref` and `at_type_param_ref` signatures (and their go-to-def implementations) don't look more alike. They're sooo similar in how they're treated that I think the method parameters should also map exactly.

We have:

```rs
fn at_local_ref(
    &mut self,
    context: &LocatorContext<'package>, // context
    path: &'package ast::Path, // reference node
    node_id: &'package ast::NodeId, // id (for finding references etc)
    ident: &'package ast::Ident, // declaration node
);
```

We have
```rs
fn at_type_param_ref(
    &mut self,
    context: &LocatorContext<'a>, // context
    _: &'a ast::Ident, // reference node
    param_id: hir::ty::ParamId, // id (for finding references etc)
// no declaration node parameter
);
```

_Almost_ the same. All it's missing is the declaration node at the end. You're looking that up here in `definition.rs`. But for locals, you do the equivalent lookup in `name_locator.rs`. I feel like either they should _both_ pass in the `ast::Ident` for the declaration, or they should _both_ omit it and let `definition.rs` look it up.

minestarks (approved):
Looks great overall, refactoring/renaming suggestions as usual.

compiler/qsc_hir/src/ty.rs
```diff
@@ -376,7 +376,7 @@ impl Display for FunctorSet {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         match self {
             Self::Value(value) => Display::fmt(value, f),
-            Self::Param(param) => Display::fmt(param, f),
+            Self::Param(param) => write!(f, "Param<{param}>"),
```
swernli, line 379: nitpick: Maybe a slightly different name in the string here would help different these from type parameters? Looking at the tests where it shows up, initially I was confused why a type paramter would appear after a return type, ie `Unit is Param<0>` but then eventually realized it was a functor set. Perhaps `FuncSet<{param}>`?

That said, I don't feel strongly about it, so if you want to leave it as-is for consistency with the Rust enum name and/or the other params, I'm fine with it.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -29,6 +29,8 @@ const PRELUDE: &[&str] = &[
 ];
 
 // All AST Path nodes get mapped
+// Ident Ids from Callable and Ty Items are mapped
+// Ident Ids of AST TyKind::Param names are mapped
```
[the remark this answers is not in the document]
ScottCarda-MS: Yes I talked with Stefan about this as well. I've updated the comments to talk about including all Idents (except those under Paths).

language_service/src/display.rs
```diff
@@ -453,23 +453,36 @@ impl Display for UdtDef<'_> {
...
-                    write!(f, ")")?;
-                } else {
-                    write!(f, "Unit")?;
-                }
+            UdtDefKind::TupleTy(defs) => display_tuple(f, defs, |def| def),
+        }
+    }
+}
+
+fn display_tuple<'a, 'b, D, I>(
```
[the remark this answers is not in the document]
ScottCarda-MS: done

language_service/src/definition.rs
ScottCarda-MS: I've opted to add the parameter to the `at_type_para_ref` signature as you suggested.

compiler/qsc_hir/src/ty.rs
ScottCarda-MS: Do you know why the ids for functor sets are lumped in with the ids for generics. Seems like a strange pairing to me.

ScottCarda-MS: Discussed offline: because the functor sets and type parameters share the same sequencing for their ids, it is confusing to label them differently. Sticking with `Param<0>` for now.

PR #853, opened 2023-11-14
Add `dump_machine` to Python API

swernli:
This adds a `dump_machine` API that allows Python code to get the allocated qubit count and dictionary of sparse state amplitudes in Python. This will enable more test scenarios in Python for Q# code.

This also includes an update to how `BigInt` values are returned in Python by using the arbitrary bitwidth Python integer type instead of formatting them as a string.

image

### DIFF
--- a/pip/src/interpreter.rs
+++ b/pip/src/interpreter.rs
@@ -101,6 +101,9 @@ impl Interpreter {
         }
     }
 
+    /// Dumps the quantum state of the interpreter.
+    /// Returns a tuple of a dictionary mapping bitstrings to complex amplitudes and the
+    /// total number of allocated qubits.
     fn dump_machine(&mut self, py: Python) -> (Py<PyDict>, usize) {
         let (state, num_qubits) = self.interpreter.get_quantum_state();
         (
--- a/pip/src/interpreter.rs
+++ b/pip/src/interpreter.rs
@@ -102,8 +102,8 @@ impl Interpreter {
     }
 
     /// Dumps the quantum state of the interpreter.
-    /// Returns a tuple of a dictionary mapping bitstrings to complex amplitudes and the
-    /// total number of allocated qubits.
+    /// Returns a tuple of (amplitudes, num_qubits), where amplitudes is a dictionary from integer indices to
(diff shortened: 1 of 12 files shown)


pip/tests/test_qsharp.py
```diff
@@ -32,6 +32,25 @@ def test_stdout_multiple_lines() -> None:
 
     assert f.getvalue() == "STATE:\n|0⟩: 1.0000+0.0000𝑖\nHello!\n"
 
+def test_dump_machine() -> None:
```
swernli, line 35: @tcNickolas This test case shows how the API could be used to verify expected states between executions of qsharp snippets (these could also be `%%qsharp` magic cells with Python validation interspersed). Let me know what you think of the API, happy to iterate.

pip/src/interpreter.rs
```diff
@@ -97,6 +101,31 @@ impl Interpreter {
         }
     }
 
+    /// Dumps the quantum state of the interpreter.
+    /// Returns a tuple of (amplitudes, num_qubits), where amplitudes is a dictionary from integer indices to
+    /// pairs of real and imaginary amplitudes.
+    fn dump_machine(&mut self, py: Python) -> (Py<PyDict>, usize) {
+        let (state, num_qubits) = self.interpreter.get_quantum_state();
```
minestarks, line 108: I think it'd be great if we went the extra mile to return a Python object here, maybe with a `.dict()` method on it to return the raw data as a dictionary (as you're doing here).

Because then we could implement the `__repr__` and `_repr_html_` methods to get bonus HTML formatting for notebooks. See: https://github.com/microsoft/qsharp/blob/34357178ece371eaa242bc7006130ec7b0b3f536/pip/src/interpreter.rs#L151-L175

minestarks (approved):
Looks great. Just left a comment asking for more :)

pip/src/interpreter.rs
swernli: More great suggestions, thanks! I tried what you said and I think it worked well. It now returns a `StateDump` object that can be pretty printed, turned into a dict, or even indexed into. Here's an example:
image

PR #854, opened 2023-11-14
LS Type Param Display

ScottCarda-MS:
Changes the display logic for type parameters to allow them to be displayed as their Q# name in language server features such as hover and signature help.

To do this, the ParamId in the HIR is paired with ItemId. This allows the references to identify their source in a context-free way.


language_service/src/test_utils.rs
```diff
@@ -48,10 +48,10 @@ pub(crate) fn compile_with_fake_stdlib(source_name: &str, source_contents: &str)
             "<std>".into(),
             r#"namespace FakeStdLib {
                 operation Fake() : Unit {}
-                operation FakeWithParam(x: Int) : Unit {}
+                operation FakeWithParam(x : Int) : Unit {}
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm open to updating them, but I like having the preface of `Fake` because it tells me in-line exactly where it came from, instead of having to remember `WithParam` comes from the `FakeStdLib` namespace. It is a little thing, but I find it useful.

compiler/qsc_eval/src/lower.rs
```diff
@@ -498,8 +498,9 @@ impl Lowerer {
             qsc_hir::ty::Ty::Infer(id) => {
                 qsc_fir::ty::Ty::Infer(qsc_fir::ty::InferTyId::from(usize::from(*id)))
             }
-            qsc_hir::ty::Ty::Param(id) => {
-                qsc_fir::ty::Ty::Param(qsc_fir::ty::ParamId::from(usize::from(*id)))
+            qsc_hir::ty::Ty::Param(_, param_id) => {
+                qsc_fir::ty::Ty::Param(qsc_fir::ty::ParamId::from(usize::from(*param_id)))
+                // ToDo: does the FIR also need the item_id?
```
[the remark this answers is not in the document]
swernli: I'm of the opinion that the FIR does not need it and this todo can be removed. If for some reason in the future we do end up needing it, we can add it to lowering once there is reason.

compiler/qsc_frontend/src/typeck/infer.rs
```diff
@@ -590,7 +590,7 @@ impl Solver {
             (&Ty::Infer(infer), ty) | (ty, &Ty::Infer(infer)) if !contains_infer_ty(infer, ty) => {
                 self.bind_ty(infer, ty.clone(), span)
             }
-            (Ty::Param(name1), Ty::Param(name2)) if name1 == name2 => Vec::new(),
+            (Ty::Param(_, name1), Ty::Param(_, name2)) if name1 == name2 => Vec::new(), // ToDo: do we need to test if the item ids are the same?
```
ScottCarda-MS, line 593: Just wanted to double-check on this. It seems like a risk if the type-checker doesn't compare the item ids as well as the param ids. It seems like there'd be a risk it would equate two different type params from two different callables. Is this a case we need to cover?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -360,13 +360,22 @@ impl Resolver {
     }
 
     fn bind_type_parameters(&mut self, decl: &CallableDecl) {
+        // ToDo: This seems like the wrong way to get the item_id
```
ScottCarda-MS, line 363: The reason this seems wrong is that we are grabbing items from the names map while we are also populating that same map. More specifically, this function is only ever called while using the With visitor on a package, and this logic requires that the [bind_global_items](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_frontend/src/resolve.rs#L624) function has already been called.
I can see that that is the case for both the [incremental compiler](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_frontend/src/incremental.rs#L198) and for the [full compiler](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_frontend/src/compile.rs#L511).
In short - it seems risky to get the id like this.

compiler/qsc_eval/src/lower.rs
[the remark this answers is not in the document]
ScottCarda-MS: removed

compiler/qsc_frontend/src/typeck/infer.rs
swernli: I don't think so. Remember, you've added the item ids solely for display purposes, so they aren't functional from the perspective of the compiler. A mismatch could indicate a bug in the logic that figures out what item to link for display, but the param id is fully sufficient for describing the item for type inferences purposes.

compiler/qsc_frontend/src/resolve.rs
swernli: Hmm... I think this is a safe assumption to make, and will be verified by the panic you have in the else case below. The reason why this is safe is that it's the same assumption that the resolver itself makes. If a callable name is not in the names list then any use of that callable cannot be resolved to an item id, and name resolution breaks down. I don't think we have a stronger way to enforce this at the moment besides exactly the way you are doing it now. If this assumption ever changes, it would have to be accompanied by a pretty significatn refactor of name resolution as a whole, which would include fixing up this function too.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -360,13 +360,22 @@ impl Resolver {
...
+        let item_id = if let Some(Res::Item(item_id, _)) = self.names.get(decl.name.id) {
+            *item_id
+        } else {
+            panic!("Callable decl expected to have item_id in names map.");
+        };
+
```
swernli, line 369: You can simplify this a bit by using the let-else pattern:

```suggestion
        let Some(Res::Item(item_id, _)) = self.names.get(decl.name.id).copied() else {
            panic!("Callable decl expected to have item_id in names map.");
        };

```

compiler/qsc_frontend/src/resolve.rs
ScottCarda-MS: removed todo

compiler/qsc_frontend/src/typeck/infer.rs
ScottCarda-MS: removed todo

compiler/qsc_frontend/src/resolve.rs
ScottCarda-MS: done

ScottCarda-MS:
Closed in favor of #862

PR #855, opened 2023-11-14
Added MAJ and ReflectAboutInteger + aux functions

DmitryVasilevsky:
* Added MAJ and ReflectAboutInteger to Arithmetic namespace
* Added CheckOperationsAreEqual to Diagnostics namespace
* Added "Ctl" modifier to FourierIncByLE because controlled version can be autogenerated.

### DIFF
--- a/library/std/arithmetic.qs
+++ b/library/std/arithmetic.qs
@@ -91,6 +91,56 @@ namespace Microsoft.Quantum.Arithmetic {
     }
 
     /// # Summary
+    /// This applies the in-place majority operation to 3 qubits.
+    ///
+    /// # Description
+    /// Assuming the state of the input qubits are |x⟩, |y⟩ and |z⟩, then
+    /// this operation performs the following transformation:
+    /// |x⟩|y⟩|z⟩ ↦ |x ⊕ z⟩|y ⊕ z⟩MAJ(x, y, z).
+    ///
+    /// # Input
+    /// ## x
+    /// The first input qubit.
+    /// ## y
+    /// The second input qubit.
+    /// ## z
+    /// A qubit onto which the majority function will be applied.
+    operation MAJ (x : Qubit, y : Qubit, z : Qubit) : Unit is Adj + Ctl {
+        CNOT(z, y);
+        CNOT(z, x);
+        CCNOT(y, x, z);
+    }
+
+	/// # Summary
+    /// Reflects a quantum register about a given classical integer.
+    ///
+    /// # Description
+    /// Given a quantum register initially in the state ∑ᵢ(αᵢ|i⟩),
+    /// where each |i⟩ is a basis state representing an integer i,
+    /// reflects the state of the register about the basis state |j⟩
+    /// for a given integer j: ∑ᵢ(-1)^(δᵢⱼ)(αᵢ|i⟩) 
+    ///
+    /// # Input
+    /// ## index
+    /// The classical integer j indexing the basis state about which to reflect.
+    /// ## reg
+    /// Little-endian quantum register to reflect.
+    ///
+    /// # Remarks
+    /// This operation is implemented in-place, without explicit allocation of
+    /// additional auxiliary qubits.
+	operation ReflectAboutInteger (index : Int, reg : Qubit[]) : Unit is Adj + Ctl {
+        within {
+            // We want to reduce to the problem of reflecting about the all-ones
+            // state. To do that, we apply our reflection within an application
+            // of X instructions that flip all the zeros in our index.
+            ApplyPauliFromInt(PauliX, false, index, reg);
+        } apply {
+            Controlled Z(Most(reg), Tail(reg));
+        }
+	}
+
+    /// # Summary
     /// Automatically chooses between addition with
     /// carry and without, depending on the register size of `ys`,
     /// which holds the result after operation is complete.
@@ -542,7 +592,7 @@ namespace Microsoft.Quantum.Arithmetic {
     /// # Reference
     ///     - [arXiv:quant-ph/0008033](https://arxiv.org/abs/quant-ph/0008033)
     ///      "Addition on a Quantum Computer" by Thomas G. Draper
-    operation FourierIncByLE (xs : Qubit[], ys : Qubit[]) : Unit is Adj {
+    operation FourierIncByLE (xs : Qubit[], ys : Qubit[]) : Unit is Adj + Ctl {
         within {
             ApplyQFT(ys);
         } apply {
--- a/library/std/diagnostics.qs
+++ b/library/std/diagnostics.qs
@@ -31,4 +31,77 @@ namespace Microsoft.Quantum.Diagnostics {
         }
     }
 
+    /// # Summary
+    /// Given two registers, prepares the maximally entangled state
+    /// between each pair of qubits on the respective registers.
+    /// All qubits must start in the |0⟩ state.
+    ///
+    /// # Input
+    /// ## left
+    /// A qubit array in the |0...0⟩ state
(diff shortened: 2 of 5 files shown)


library/std/diagnostics.qs
```diff
@@ -31,4 +31,77 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    is Adj + Ctl {
+        Fact( Length(left) == Length(right), "PrepareEntangledState: Lengths of qubits registers must be equal.");
+        for idxQubit in 0 .. Length(left) - 1  {
+            H(left[idxQubit]);
+            CNOT(left[idxQubit], right[idxQubit]);
+        }
+    }
+    
+    /// # Summary
+    /// Given two operations, asserts that they act identically for all input states.
```
swernli, line 54: I like the updated name to match the behavior of returning a Boolean (and the fact that it matches `CheckZero` and `CheckAllZero`)! Looks like the doc comment needs updating to remove references to "assert" and "assertion" so that it's consistent.

library/std/diagnostics.qs
```diff
@@ -31,4 +31,77 @@ namespace Microsoft.Quantum.Diagnostics {
...
+        } apply {
+            actual(target);
+            Adjoint expected(target);
+        }
+
+        // Check and resturn result.
+        let areEqual = CheckAllZero(reference + target);
+        ResetAll(target);
+        ResetAll(reference);
+        areEqual
```
swernli, line 103: Since all we care about is whether the operations are equal and not which qubit they are unequal for, we don't need 2 * nQubits to do it, just nQubits + 1. This significantly reduces load on the simulator. It also has the nice side benefit of removing the dependency on `PrepareEntangledState` such that the internal operation can be removed.

```suggestion
        // Prepare a reference register entangled with the target register.
        use reference = Qubit();
        use target = Qubit[nQubits];

        // Apply operations.
        within {
            H(reference);
            for qubit in target {
                CNOT(reference, qubit);
            }
        } apply {
            actual(target);
            Adjoint expected(target);
        }

        // Check and resturn result.
        let areEqual = CheckAllZero([reference] + target);
        ResetAll(target);
        Reset(reference);
        areEqual
```

library/tests/src/test_diagnostics.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::test_expression;
+use qsc::interpret::Value;
+
+#[test]
+fn check_operations_are_equal() {
+    test_expression(
+        "{
+            open Microsoft.Quantum.Diagnostics;
+            open Microsoft.Quantum.Arrays;
+            operation op1(xs: Qubit[]): Unit is Adj+Ctl {
```
swernli, line 13: nitpick: this doesn't need to be `Ctl`

library/std/diagnostics.qs
DmitryVasilevsky: Checking Array and the extra qubit separately would probably be faster than concatenating arrays.

library/tests/src/test_diagnostics.rs
```diff
@@ -0,0 +1,27 @@
...
+                CCNOT(xs[0], xs[1], xs[2]);
+            }
+            operation op2(xs: Qubit[]): Unit is Adj {
+                Controlled X(Most(xs), Tail(xs));
+            }
+            operation op3(xs: Qubit[]): Unit is Adj {
+                Controlled X(Rest(xs), Head(xs));
+            }
+            [CheckOperationsAreEqual(3, op1, op2), CheckOperationsAreEqual(3, op1, op3)]
```
swernli, line 22: Since the operations under test here are all adjointable, it would be good to confirm equality works both ways, ie: op1 == op2 and op2 == op1, as well as op1 != op3 and op3 != op1.

library/std/diagnostics.qs
swernli: Yeah, good point... avoids two memory allocations. So something like `CheckAllZero(target) and CheckZero(reference)`.

library/std/arithmetic.qs
```diff
@@ -90,6 +90,56 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// The second input qubit.
+    /// ## z
+    /// A qubit onto which the majority function will be applied.
+    operation MAJ (x : Qubit, y : Qubit, z : Qubit) : Unit is Adj + Ctl {
+        CNOT(z, y);
+        CNOT(z, x);
+        CCNOT(y, x, z);
+    }
+
+	/// # Summary
```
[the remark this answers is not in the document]
ScottCarda-MS: Oh looks like that happened in a couple other places too. Can you check to make sure you remove tabs everywhere?

library/std/diagnostics.qs
DmitryVasilevsky: Unfortunately, this won't work. If you take an n-qubit operation, it can be represented by 2ⁿ x 2ⁿ matrix, or 2²ⁿ coefficients. When two n-qubit registers are used as in the original algorithm, it creates a state with 2²ⁿ coefficients. The verification procedure maps matrix coefficients of AB† into the coefficients this state. Therefore, verifying that the matrix is an identity matrix. When we use n-bit register and an additional qubit, we only have 2*2ⁿ coefficients - not sufficient to map and verify all coefficients. So, the verification procedure no longer works.

Here's a short example ([or in playground](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACn2QQUvDQBCF7%2FkVc9yAtOrRIlis4CXagl4sPYzpJB0xs3EyQYLkv8vaJG2D%2Bg7LDu%2B9b5YVLKgqMSVImlWNYnWxVJ8rFvAVAQD4kgQSTtVXPrNJl5ksGHPxlXFazf7LzVWx6SI%2Fx82dmDZLz2Iu7puKxl4gQRYXwxU8C1u3P6iuCJQyUpKU4BpW9Subi2cnAUPNyXp3fbk52J9sO5YjYNC9G5BHpKDMK3wECLD02NNu0O3D49MBcbZvjEjtMLWAZfnejDiLuigTTHcs5EbVF7ffvD7f%2FOVcjJ3pFObbt%2FC1YB54S2JszeENw%2B2XvW3URt%2BHc%2ByhDQIAAA%3D%3D)):

```qsharp
namespace MyQuantumProgram {
    
    @EntryPoint()
    operation Main() : Unit {
        use reference = Qubit();
        use target = Qubit[2];
        within {
            H(reference);
            for qubit in target {
                CNOT(reference, qubit);
            }
        } apply {
            DumpMachine();
            Z(target[0]);
            Z(target[1]);
            // Adjoint to identity
        }
        DumpMachine();
    }
}
```

The state created is only |000⟩+|111⟩ (normalized). Any unitary with eigenvectors |00⟩+|11⟩ will leave this state unchanged. Here we use Z⊗Z. And this procedure cannot distinguish it from the identity gate.

library/std/diagnostics.qs
DmitryVasilevsky: Updated.

library/tests/src/test_diagnostics.rs
DmitryVasilevsky: Updated.

library/tests/src/test_diagnostics.rs
DmitryVasilevsky: Compared all the way.

cesarzc (approved):
LGTM after you've addressed Scott's and Stefan's comments.

library/std/diagnostics.qs
swernli: Thanks for the detailed explanation and example!

PR #858, opened 2023-11-16
Add samples that are compatible with NISQ devices

cesarzc:
This change adds a few samples that are compatible with NISQ devices.


samples/algorithms/HiddenShiftNISQ.qs
```diff
@@ -0,0 +1,168 @@
...
+    /// flat Fourier (Walsh–Hadamard) spectra.
+    ///
+    /// In this case, the Roetteler algorithm (see References, below) uses
+    /// black-box oracles for 𝑓^* and 𝑔, where 𝑓^* is the dual bent function to
+    /// 𝑓, and computes the hidden shift 𝑠 between 𝑓 and 𝑔.
+    ///
+    /// # Input
+    /// ## Ufstar
+    /// A quantum operation that implements
+    /// $U_f^*: |𝑥〉 ↦ (-1)^{f^*(x)} |𝑥〉$,
```
DmitryVasilevsky, line 58: Should this be converted to Unicode?

cesarzc: I'm not sure this can completely be expressed in unicode.

samples/algorithms/BernsteinVaziraniNISQ.qs
```diff
@@ -0,0 +1,151 @@
...
+    ///   (https://doi.org/10.1137/S0097539796300921)
+    operation BernsteinVazirani(Uf : ((Qubit[], Qubit) => Unit), n : Int)
+    : Result[] {
+        // We allocate n + 1 clean qubits. Note that the function Uf is defined
+        // on inputs of the form (x, y), where x has n bits and y has 1 bit.
+        use queryRegister = Qubit[n];
+        use target = Qubit();
+
+        // The last qubit needs to be flipped so that the function will actually
+        // be computed into the phase when Uf is applied.
```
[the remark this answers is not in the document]
cesarzc: Updated the comment to hopefully make it clearer.

samples/samples.mjs
```diff
@@ -7,16 +7,20 @@
...
+    { title: "Minimal", file: "./language/GettingStarted.qs", shots: 100 },
+    { title: "Superposition", file: "./algorithms/Superposition.qs", shots: 100 },
+    { title: "Entanglement", file: "./algorithms/Entanglement.qs", shots: 100 },
+    { title: "Bell States", file: "./algorithms/BellState.qs", shots: 100 },
+    { title: "Teleportation", file: "./algorithms/Teleportation.qs", shots: 1 },
+    { title: "Random Bit", file: "./algorithms/RandomBit.qs", shots: 100 },
+    { title: "Random Number Generator", file: "./algorithms/QRNGNISQ.qs", shots: 1000 },
+    { title: "Random Number Generator (Advanced)", file: "./algorithms/QRNG.qs", shots: 1000 },
+    { title: "Deutsch-Jozsa", file: "./algorithms/DeutschJozsaNISQ.qs", shots: 1 },
+    { title: "Deutsch-Jozsa (Advanced)", file: "./algorithms/DeutschJozsa.qs", shots: 1 },
```
DmitryVasilevsky, line 19: Feels a bit strange to have "advanced" designation for a normal case and no designation for "dumbed down" case. Of course, this is all subjective, but to me "advanced" is a normal case. But I cannot come up with a better way to name these so it's up to you. You can keep it this way.

DmitryVasilevsky (approved):

samples/samples.mjs
cesarzc: Yes, it is a bit strange to have duplicates of the same samples and naming it is hard. I'll keep it this way for now.

PR #861, opened 2023-11-17
qsharp.json support in Python

minestarks:
Allow `qsharp.init` to take an optional project root. We'll look for a qsharp.json at the project root and compile the project sources at initialization.

This PR originally removed `qsharp.eval_file` but I decided to relegate that to a separate PR to keep it manageable

----- 

image

-----


pip/src/fs.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug)]
+struct Entry {
+    entry_type: EntryType,
+    path: String,
+    entry_name: String,
+    extension: String,
+}
+
+impl DirEntry for Entry {
+    type Error = pyo3::PyErr;
```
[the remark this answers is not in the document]
minestarks: I had to pick an error type because `entry_type` on the trait returns a `Result` . It's true none of the implementations can actually fail. What would be an error type to pick in this case?

pip/qsharp/_qsharp.py
```diff
@@ -6,28 +6,40 @@
...
-    if _interpreter is None:
-        init()
-    return _interpreter
+        import json
+
+        manifest_descriptor = {}
+        manifest_descriptor["manifest_dir"] = project_root
+        manifest_descriptor["manifest"] = json.loads(
+            open(qsharp_json, mode="r", encoding="utf-8").read()
+        )
```
idavis, line 38: 1. Wihout a corresponding close() call the file is left open. The `with open(...) as f:` construct handles this closing the file automatically.
2. Do we want to provide a custom error message to the user if the project file can't be read?
3. There is a util function at the bottom of the file that does the same read logic, do we want to consolidate?

pip/qsharp/_fs.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import os
+
+
```
idavis, line 6: All functions missing return and param types

pip/qsharp/_native.pyi
```diff
@@ -28,11 +28,14 @@ class TargetProfile:
 class Interpreter:
     """A Q# interpreter."""
 
-    def __init__(self, target_profile: TargetProfile) -> None:
+    def __init__(
+        self, target_profile: TargetProfile, manifest_descriptor: Optional[dict]
```
idavis, line 32: Dict missing type parameters

pip/qsharp/_qsharp.py
```diff
@@ -37,16 +37,54 @@ def _repr_mimebundle_(
         return {"application/x.qsharp-config": self._config}
 
 
-def init(target_profile: TargetProfile = TargetProfile.Unrestricted) -> Config:
+def init(
+    *, target_profile: TargetProfile = TargetProfile.Unrestricted, project_root=None
```
idavis, line 41: Project root is missing type and has an implict optional type. Specity Optional[T] for what path join needs

pip/qsharp/_fs.py
minestarks: Thanks for cleaning this up!

pip/qsharp/_qsharp.py
```diff
@@ -37,16 +37,54 @@ def _repr_mimebundle_(
...
+) -> Config:
     """
     Initializes the Q# interpreter.
 
     :param target_profile: Setting the target profile allows the Q#
         interpreter to generate programs that are compatible
         with a specific target. See :py:class: `qsharp.TargetProfile`.
+
+    :param project_root: The root directory of the Q# project. It must
+        contain a qsharp.json project manifest.
```
swernli, line 53: nitpick: the comment makes it seem like this is required, even the parameter is optional. Maybe something like:


```suggestion
    :param project_root: An optional path to a root directory with a Q# project to include.
        It mustcontain a qsharp.json project manifest.
```

pip/qsharp/_qsharp.py
```diff
@@ -85,13 +123,13 @@ def callback(output: Output) -> None:
 def eval_file(path: str) -> Any:
```
swernli, line 123: I think it's worth removing this now, as discussed offline.

swernli (approved):

pip/qsharp/_qsharp.py
minestarks: OK! Will do in separate PR since I'm rushing to merge this now.

PR #862, opened 2023-11-17
LS and Error Display Refactor

ScottCarda-MS:
Refactors the language server and error display of type parameters and UDT's. These types now have `Rc<str>` for their names, avoiding us to drop a lot of complex lookup logic. The display logic for the types was also moved from the language server's display.rs to the `.display()` on the `Ty` struct.
As a result of these changes, error messages can now print a user-friendly display of types like type parameters and UDT's.


compiler/qsc_frontend/src/typeck.rs
```diff
@@ -39,7 +39,7 @@ pub(super) struct Error(ErrorKind);
 enum ErrorKind {
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.TypeCk.TyMismatch"))]
-    TyMismatch(Ty, Ty, #[label] Span),
+    TyMismatch(String, String, #[label] Span),
```
minestarks, line 42: I like this. Strings instead of HIR objects in the error makes a lot of sense to me!

wasm/src/lib.rs
```diff
@@ -155,7 +155,12 @@ where
     }
 }
 
-fn run_internal<F>(code: &str, expr: &str, event_cb: F, shots: u32) -> Result<(), stateful::Error>
+fn run_internal<F>(
+    code: &str,
+    expr: &str,
+    event_cb: F,
+    shots: u32,
+) -> Result<(), Box<stateful::Error>>
```
minestarks, line 163: Why did this end up needing to be boxed?

minestarks (approved):

compiler/qsc_frontend/src/typeck.rs
ScottCarda-MS: They are smaller in size too 😊

compiler/qsc_hir/src/ty.rs
```diff
@@ -220,11 +238,26 @@ impl Display for GenericParam {
 #[derive(Clone, Debug, PartialEq)]
 pub enum GenericParam {
     /// A type parameter.
-    Ty,
+    Ty(TypeParamName),
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm concerned that giving them NodeIds and calling them Idents implies that they need to be mappable in the type map.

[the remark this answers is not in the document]
minestarks: Hmm, that's a concern I hadn't thought of. But generic params are already represented by `Ident`s in the AST, and the `NodeId`s used for the types table are for the _AST_ types. The types table already references generic parameters by their `NodeId`. So I think that's all consistent:

https://github.com/microsoft/qsharp/blob/81873fe2709f05d8b7b9671b53dd72a79d3d6596/compiler/qsc_frontend/src/typeck.rs#L28-L30

Similarly, the names table in the resolver also uses AST `NodeId`s. Except maybe the comment may be incorrect, since all `Ident` nodes may not get mapped? I'm guessing generic parameter `Ident`s are excluded?

https://github.com/microsoft/qsharp/blob/81873fe2709f05d8b7b9671b53dd72a79d3d6596/compiler/qsc_frontend/src/resolve.rs#L31-L33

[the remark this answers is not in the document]
ScottCarda-MS: From poking around a bit, I found that the ids of the AST Idents are mappable in the names map, but not in the type map, which is fine and consistent with the comments describing those mappings.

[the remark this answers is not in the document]
ScottCarda-MS: Another issue I'm running into is that `Ty`'s don't have Idents on them, and adding an Ident to the ty.rs would ruin the separation between the hir.rs and the ty.rs modules.

wasm/src/lib.rs
ScottCarda-MS: I started getting a warning that the Err part of this Result was too large. I'm unsure why I was getting that waring considering it was also telling me the size hadn't change since before my changes, _and_ `String` objects are smaller than `Ty` objects. In any case, it was directing me to box the Err, so I did to alleviate the warning.

compiler/qsc_hir/src/ty.rs
```diff
@@ -61,28 +61,68 @@ impl Ty {
...
+                } else {
+                    format!(" is {}", arrow.functors)
+                };
+                format!(
+                    "({} {arrow_symbol} {}{functors})",
+                    arrow.input.display(),
+                    arrow.output.display()
+                )
+            }
+            Ty::Infer(_) | Ty::Err => "?".to_string(),
```
swernli, line 89: One minor advantage of including the inference integer in the error is it lets the user know what types are distinct even if they are unknown. I don't know if that's worth the effort of figuring out how to display them consistently, but it is some information that user previous could get from error messages that will be lost with this pattern.

compiler/qsc_hir/src/ty.rs
```diff
@@ -61,28 +61,68 @@ impl Ty {
...
+                } else {
+                    let items = items.iter().map(Ty::display).collect::<Vec<_>>().join(", ");
+                    format!("({items})")
+                }
+            }
         }
     }
 }
 
 impl Display for Ty {
```
swernli, line 107: I wonder if it's worth unifying this implementation with the `display` such that only `fmt` is needed, and then the places where `.display()` were called become `.to_string()` which internally calls `fmt`. Right now there are a few differences, but if we could reconcile them then there could be a single implementation rather than two that we need to remember to choose between.

ScottCarda-MS: I considered trying to do this, but realized we really have separate needs for printing depending on the context. The HIR printing really needs to print all components of a struct in as compact a form as possible, while the error message/language server messages don't need to tell the user all the internal fields of the struct and needs to prioritize easy to read displays. Because they have such different needs, I'm not sure we can have one single way of printing these data structs that can satisfy both.

ScottCarda-MS: As an aside though, I did briefly look into using formatting flags with the `fmt` impl to get the two different ways of printing to be used with `format!` macro, but that turned out to be a bit more cryptic than simply having a separate `.display` function.

wasm/src/lib.rs
swernli: I see the same thing on my end... with the changes or without them, `stateful::Error` is 128 bytes, but only with the changes does clippy complain about the size. Very odd, but if a `Box` at the top level is enough to satisfy it then I suppose it's not a big concern.

swernli (approved):

compiler/qsc_hir/src/ty.rs
ScottCarda-MS: That's true, and I think it is a balancing act of trying to figure out what is most helpful to the user. I lean slightly on the side of thinking that that information is not usually very relevant to a user and the single `?` is prettier. I could be convinced that that information is very useful to a user though. I'd also like to get @minestarks to weigh in on this change, whether or not to go forward with it.

wasm/src/lib.rs
minestarks: Wow, that is a large error type... I guess we normally store these in a `Vec<>` instead of returning directly, so they're already heap-allocated in other places. Clippy's fickleness aside, the reasoning makes sense to me.

compiler/qsc_hir/src/ty.rs
minestarks: Well, if my opinion is the tiebreaker, I still cast my vote with the `?` display. While the numbers are helpful to people who "know what they're doing" (i.e. us), I don't think the average programmer knows enough about type inference to understand what the numbers are for, and the confusion they create negates any clarification they bring. Again, my opinion and not a very strong one.

minestarks: Oh and other mature languages (TypeScript, Rust) which do pretty sophisticated type inference also just show "unknown" without exposing any other inference info. This tells me that there just isn't much user demand for this type of thing.

compiler/qsc_hir/src/ty.rs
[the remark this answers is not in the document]
minestarks: But `ty.rs` already imports a ton of stuff from `hir.rs`. Either way I'm okay with this - having a more specific type is not such a bad thing after all.

compiler/qsc_hir/src/ty.rs
swernli: That makes sense to me. I don't feel strongly about it either, so just "?" seems good!

ScottCarda-MS: Okay, I'm going forward with the single `?` then. This can easily be changed in the future if the demand for more specific information surfaces.

PR #863, opened 2023-11-18
set target profile in the notebook language service via `qsharp.init()`

minestarks:
This feature makes `qsharp.init()` generate some hidden output that provides a hint to the language service to set the correct target profile. No magic commands needed!

target-profile


pip/qsharp/_qsharp.py
```diff
@@ -116,3 +120,24 @@ def __init__(self, name: str, ll_str: str):
...
+            case TargetProfile.Full:
+                target_profile = "full"
+            case TargetProfile.Base:
+                target_profile = "base"
+        self._config = {"targetProfile": target_profile}
+
+    def __repr__(self):
+        return "Q# initialized with configuration: " + str(self._config)
+
+    def _repr_mimebundle_(self, include=None, exclude=None):
```
idavis, line 142: Could you please add a comment as to what this is used for?

idavis (approved):

cesarzc (approved):

PR #864, opened 2023-11-18
Table lookup with select networks

msoeken:
This adds an operation `Select` in a new namespace `Microsoft.Quantum.TableLookup`. It implements a table lookup based on a recursive implementation of the select network algorithm (see references inside the doc string for more details).

The behavior of this operation is similar to [MultiplexOperations](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.canon.multiplexoperations), with the following differences:

- It uses an optimized implementation
- It has a name more commonly used in the quantum algorithms community
- It is less general as it writes bit strings instead of allowing an arbitrary set of unitary operations

**NOTE:** Some clippy checks are failing in code that is not changed by this PR. I did not fix these, since they are not related to this PR.

### DIFF
--- a/compiler/qsc_frontend/src/compile.rs
+++ b/compiler/qsc_frontend/src/compile.rs
@@ -445,6 +445,10 @@ pub fn std(store: &PackageStore, target: TargetProfile) -> CompileUnit {
                 "random.qs".into(),
                 include_str!("../../../library/std/random.qs").into(),
             ),
+            (
+                "table_lookup.qs".into(),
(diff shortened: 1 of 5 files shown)


billti:
We want to keep the core library built into the product small and focused where possible (back-compat concerns notwithstanding). Especially as this adds a new namespace as well, we'd want to have clear justification, discussion, and agreement that this is broadly useful and durable 'core functionality' before adding. Let's set up a call to discuss. I'll ping you.

DmitryVasilevsky (approved):

PR #865, opened 2023-11-18
Refactor UX components and add WebViews to VS Code

billti:
There is more I want to do, but this is big enough already, and want to get the major pieces in.

This refactors some of the Preact UX components into the NPM package and then shares them between the playground and VS Code.

It adds histograms, re, and a help page rendering as WebViews in VS Code.

The help page (content still to be fleshed out) will open to the side when the Help command is run, and this will be done by default when opening a code sharing link that takes you to vscode.dev/quantum.


billti:
You can play around with the UX in vscode.dev by opening [this code share link](https://insiders.vscode.dev/+quantum.qdk-test?code=H4sIAAAAAAAAE51VwW4TMRC95ytGLUiJQJvkwKVRgUAR7SEVTbghpHWyk8Ri15vY3rahzR%2FwJXwWX8KM7WS9adUi9tLGO%2FPezPOb2W63C8cwEcUqx1aXflxVQtmqgLFQWVnAZVVMUcNnVKiFLTXHuLhjOEMz03JlZancydelNLDS5UKLAiQDFqisAQHrgKmFcqDKgy52oDDdgEFrpVpQ7FRa4wClAlOtUK9KI5kFqCSwS1RQGQ6lf6FAYSrtmECjqXImNMzERAyVtJQo0KzEDEOfcNcCesoVIY3kTJemnNskNJ58LNU1ajt4KuZCWS2VkbMno0bCLgctF%2FH%2BE2VsvpRS2XZnl0O9c1cjIVW7AydAqKE0fnK0UIhbOIV%2BrzfYH4%2FQGLHA9osj1wzrIHbtBl2naG%2BQKuo5we4IZHsCR51QCj8kbrhRPMglzVnWXpIwN71bYLJP02grrfap3iLeIRdqzLFtyur4YreezntlUhWF0Jv9yQ7CPFN8SnhpciDYM%2FxeyYeCEu0ZWtSFVOiaDIzl3PkEFGKGZLCS%2BlyRl9hSjt9VYsinCNLGcEGsVH2g%2FBSuhZZimmNCo0AdLHYK795HpoSbpZwt4UbmeQxYEzPuDiBrSlTfR1FZ5vOIp%2FDt%2B6DhHkdL5%2FRnIn%2FiRXQ5%2FMxp8GR2S2%2B5kX6S%2BPhaMH5oLD38K8JvCk%2FR7U5EuW2QGz9qpzB2UznUWmyGhm6kzXAHZhx7Yx1YcQ5cGmkl7TJIrXFdoQmSxNaMHU1ZYkFDVSOUlTUy87f%2BuKdDuW%2FdyL171uJwHzL%2Bz%2BskwdOmdtqSg714TRMP87yccZ%2FCL8u6l8ogrEnyKz5uH2g8Qe8ql0M3Tj5vbleaAz90%2FWi%2FnotMUDcZxFB11el5WtOft9cHnEOm5K8CL76IPV2nsKQ1%2Fab3EmZLoWg18xQis4aNnoXpiuHue39%2B%2FaZJdM1Tof%2BQDvf9fU7DLSMfFxV1LfIKo9bTUVp3%2BjpaAgdVxZ8gj7HbC%2F6DFC2GxoT4t3Rdowe60bVTgC%2BL7sn10Xu8D3fXhlaBoUHdU8dK0ceVj6wsXOkbEJoHKaeyMauxHOcjlbjx8LPn6iWlDz68D4bJRzaBLkvL9ML6ip0FyBmismVBCs9Enm%2F2VYHwbkHSPPBNyfQ%2FkjBt29ZfD7iEnLgIAAA%3D) for a test extension.

npm/ux/qsharp-ux.css
```diff
@@ -27,7 +50,16 @@ html {
   padding: 0;
 }
 
-body {
+html {
```
billti, line 53: Note: The contents of this file haven't actually changed much, mostly just been moved around to group together related styles.

npm/ux/histogram.tsx
```diff
@@ -242,218 +244,195 @@ export function Histogram(props: {
...
-              </rect>
-              {
-                <text
-                  class="bar-label"
-                  x={labelX}
-                  y="85"
-                  visibility={showLabels ? "visible" : "hidden"}
-                  transform={`rotate(90, ${labelX}, 85)`}
+    <>
+      {props.shotsHeader ? (
```
billti, line 248: This and the next 2 lines are largely the only change in the output here. The rest is mostly whitespace updates.

package.json
```diff
@@ -18,23 +18,28 @@
...
     "@types/mocha": "^10.0.2",
     "@types/node": "^18.17",
     "@types/vscode": "^1.83.0",
+    "@types/vscode-webview": "^1.57.3",
     "@typescript-eslint/eslint-plugin": "^6.8.0",
     "@typescript-eslint/parser": "^6.8.0",
     "@vscode/debugadapter": "^1.63.0",
     "@vscode/debugprotocol": "^1.63.0",
     "@vscode/extension-telemetry": "^0.8.5",
+    "@vscode/markdown-it-katex": "^1.0.0",
```
billti, line 31: The webview uses markdown-it and katex to align more closely with how VS Code renders Notebooks and LaTeX. These packages are also faster and smaller. We can consider if the playground should move also at a later time.

vscode/src/webview/help.tsx
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function HelpPage() {
+  return (
+    <div style="line-height: 1.5">
+      <h1>Azure Quantum Development Kit</h1>
+      <h2>Overview</h2>
+      <p>
+        Welcome to the Azure Quantum Development Kit! Or the "QDK", as we like
```
billti, line 10: Obviously this help content needs to be fleshed out, but want to get eyes on the rest of the PR ASAP.

vscode/src/webviewPanel.ts
```diff
@@ -0,0 +1,240 @@
...
+  <!DOCTYPE html>
+  <html lang="en">
+    <head>
+      <meta charset="UTF-8">
+      <meta name="viewport" content="width=device-width, initial-scale=1.0">
+      <title>Q#</title>
+      <link rel="stylesheet" href="${githubCss}" />
+      <link rel="stylesheet" href="${katexCss}" />
+      <link rel="stylesheet" href="${webviewCss}" />
+      <script src="${webviewJs}"></script>
```
billti, line 202: To simply the build and keep the number/size of scripts minimal, the WebView is actually identical html, js, and css for all scenarios, and renders different UX depending on the messages sent to it.

npm/ux/qsharp-ux.css
```diff
@@ -27,7 +50,16 @@ html {
...
+  box-sizing: border-box;
+  font-size: 16px;
+}
+
+/* TODO: Make the below playground specific classes */
```
idavis, line 58: @billti Are you planning to address this here or in a future PR?

idavis (approved):

npm/ux/qsharp-ux.css
billti: In a later PR. I'll probably move the styles specific to the playground out of the common package and back to a playground-specific CSS file.

vscode/src/webview/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "lib": ["DOM", "ES2020"],
+    "rootDir": ".",
+    "strict": true /* enable all strict type-checking options */,
+    "jsx": "react-jsx",
+    "jsxImportSource": "preact",
+    "isolatedModules": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true
+  },
+  "exclude": ["node_modules"]
```
minestarks, line 15: Does this do anything? Surely node_modules is in the default excludes?

vscode/src/webviewPanel.ts
```diff
@@ -0,0 +1,240 @@
...
+    </head>
+    <body>
+    </body>
+  </html>
+`;
+  }
+
+  sendMessage(message: any) {
+    if (this._ready) {
+      console.log("Sending message to webview", message);
```
minestarks, line 212: Can these use `qscLog` too? (At debug/trace level maybe?)

vscode/src/webviewPanel.ts
```diff
@@ -0,0 +1,240 @@
...
+    commands.registerCommand("qsharp-vscode.showHelp", async () => {
+      const message = {
+        command: "help",
+      };
+      sendMessageToPanel("help", true, message);
+    }),
+  );
+
+  context.subscriptions.push(
+    commands.registerCommand("qsharp-vscode.showHistogram", async () => {
```
minestarks, line 55: This spurred some UX confusion in the Monday demo, and I tend to agree that it's a bit counterintuitive that "show histogram" does real work (i.e. runs the program) besides just showing a window. I think a more cohesive UX here would be to use the existing "Run Q# File" command to actually run the program, and then display the Histogram based on the simulation that was just run. It just feels strange that we have two ways of running a Q# program now, and one is called "show" and shows a histogram, and the other one doesn't. How would a user know which one to pick?

Perhaps for now, rename to something like "Run Q# File (Histogram)"? I'm not sure if I like my own suggestion.

Not for this PR, but I think taking a holistic look at the UX and maybe aligning some of the disjoint experiences would be nice before we ship.

vscode/package.json
```diff
@@ -156,6 +156,18 @@
           "command": "qsharp-vscode.getQir",
           "when": "resourceLangId == qsharp"
         },
+        {
+          "command": "qsharp-vscode.showHistogram",
+          "when": "resourceLangId == qsharp"
```
minestarks, line 161: Is it OK to show this command during a debug session? I'm guessing it won't interfere with the currently running program but the user also shouldn't expect that the histogram is relevant to the currently running program.

minestarks (approved):

vscode/src/webviewPanel.ts
billti: Yeah. The existing 'Run' (Ctrl+F5) experience does what you'd expect, and runs the code once, outputting messages to the debug console. For some algorithms (or when developing) that's exactly what you want, and histograms aren't interesting (e.g. all the samples on the playground that default to 1 shot are that way because a histogram is of no value).

When you run a histogram you typically are interested in running hundreds (or thousands) of shots and seeing the statistical results, and the thousands of lines that likely end up in your debug console are not interesting.

I do see them as two very different use cases, but we should agree on how to make the experience intuitive.

vscode/package.json
billti: Good catch. I could add that check and prohibit it. But as you outline, it runs in a separate worker, and there's actually no reason why you can't bring up a histogram for the code your debugging. Maybe that is interesting or useful when debugging something?

billti: Actually, I'll leave it. I see no reason to prohibit it.

vscode/src/webviewPanel.ts
billti: I changed the command text to 'Run file and show histogram' (it's already prefixed with 'Q#')

vscode/src/webviewPanel.ts
billti: Good catch. Fixed.

PR #869, opened 2023-11-22
Check whether number is NaN

msoeken:


library/std/math.qs
```diff
@@ -45,6 +45,24 @@ namespace Microsoft.Quantum.Math {
...
+    /// Returns whether a given floating-point value is not a number (i.e. is
+    /// NaN).
+    ///
+    /// # Input
+    /// ## d
+    /// A floating-point value to be checked.
+    ///
+    /// # Output
+    /// `true` if and only if `d` is not a number.
+    function IsNaN(d : Double) : Bool {
```
swernli, line 62: Could you add a test case to https://github.com/microsoft/qsharp/blob/d183b05a863d8f2365f61392839d3fd703eedb46/library/tests/src/test_math.rs to cover this new function?

msoeken: Added test case.

swernli (approved):

PR #871, opened 2023-11-23
LS: Dynamic Indentation for Auto Open

ScottCarda-MS:
Changes the hardcoded indentation used for auto-open to use an indentation based on the text found before the insertion location. This will allow the indentation to be accurate for both the namespace experience, and the notebook experience.


language_service/src/completion.rs
```diff
@@ -19,11 +19,49 @@ const PRELUDE: [&str; 3] = [
     "Microsoft.Quantum.Intrinsic",
 ];
 
+struct IndentMaker<'a> {
+    contents: &'a str,
+    offset: u32,
+}
+
+impl<'a> IndentMaker<'a> {
+    fn new(compilation: &'a Compilation, source_name: &'_ str) -> Self {
```
minestarks, line 28: Surprised you need a lifetime for `source_name: &'_ str` . What happens if you remove it?

language_service/src/completion.rs
```diff
@@ -19,11 +19,49 @@ const PRELUDE: [&str; 3] = [
...
+        let source = compilation
+            .user_unit()
+            .sources
+            .find_by_name(source_name)
```
minestarks, line 32: Since you already have the package offset for the cursor, it's going to be more efficient if you look up the source contents with `find_by_offset`, instead of looking up using the name.

Or maybe you can have `source_offset_to_package_offset` return the `SourceContents` as well? Since it's getting called in the next line anyway.

language_service/src/completion.rs
```diff
@@ -19,11 +19,49 @@ const PRELUDE: [&str; 3] = [
     "Microsoft.Quantum.Intrinsic",
 ];
 
+struct IndentMaker<'a> {
+    contents: &'a str,
```
minestarks, line 23: I think you can just use a `SourceContents` (reference counted string, the same type that is stored in the `SourceMap`) here instead of having to futz with lifetimes.

language_service/src/completion.rs
```diff
@@ -343,8 +387,9 @@ impl CompletionListBuilder {
                                                     additional_edits.push((
                                                         protocol::Span { start, end: start },
                                                         format!(
-                                                            "open {};\n    ",
-                                                            namespace.name.clone()
+                                                            "open {};{}",
+                                                            namespace.name.clone(),
+                                                            indent_maker.get_ident(start),
```
minestarks, line 392: This gets called for every single completion we add to the list - almost every single global in the program. Should you cache the indent string?

minestarks (changes requested):

language_service/src/completion.rs
ScottCarda-MS: done

language_service/src/completion.rs
ScottCarda-MS: Avoided this by using contents without a need to annotate types, so it implicitly uses `SourceContents`.

language_service/src/completion.rs
ScottCarda-MS: I removed the `IndentMaker` in favor of using a single function.

minestarks (approved):

PR #876, opened 2023-11-29
removed unnecessary hashes around raw string literals

filipw:
The solution does not compile with Rust/Cargo 1.74.0 and clippy 0.1.74 due to to a flood of [unnecessary hashes around raw string literal](https://rust-lang.github.io/rust-clippy/master/index.html#/needless_raw_string_hashes) clippy complaints

### DIFF
--- a/compiler/qsc/src/interpret/debug/tests.rs
+++ b/compiler/qsc/src/interpret/debug/tests.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(clippy::needless_raw_string_hashes)]
+
 use indoc::indoc;
 use miette::Result;
 use qsc_eval::{output::CursorReceiver, val::Value};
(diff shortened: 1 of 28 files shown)


ScottCarda-MS (approved):
Thanks!

ScottCarda-MS (changes requested):
After discussing this PR with the team, we've decided the way forward with this is to disable the specific lint instead of removing the hashes for the test files. The reason for this is that the expect-test macro will still generate strings with the hashes, even when hashes aren't strictly necessary. The team's preference is to avoid having to remove those hashes every time the expect-macro generates a string by allowing the hashes.

Would you be willing to add `#![allow(clippy::needless_raw_string_hashes)]` to the top of the test files instead of removing the hashes?

filipw:
> Would you be willing to add `#![allow(clippy::needless_raw_string_hashes)]` to the top of the test files instead of removing the hashes?

sure no problem, changed it accordingly (the end result is the same, the build is green for me 🙂  )

ScottCarda-MS (approved):
Thanks for making the changes!

swernli (approved):

PR #878, opened 2023-11-29
Reworked adders and tests

DmitryVasilevsky:
Added and reworked all adders and helper operations. Adders supported:

* General in-place adder `IncByLE`. Implemented via TTK adder. Specific implementations:
   * `RippleCarryTTKIncByLE`
   * `RippleCarryCGIncByLE`
   * `FourierTDIncByLE`
* General `IncByI`. Implemented via `IncByIUsingIncByLE(RippleCarryTTKIncByLE)`.
* General `IncByL` implemented via `IncByLUsingIncByLE(RippleCarryTTKIncByLE)`.
* General out-of-place adder `AddLE`. Implementated via CG adder. Specific implementations:
   * `RippleCarryCGAddLE`
   * `LookAheadDKRSAddLE`
* Helper function `IncByLEUsingAddLE` to use two out-of-place adders to add in-place.
* Helper function `IncByIUsingIncByLE` to increase by Int using an in-place adder.
* Helper function `IncByLUsingIncByLE` to increase by BigInt using an in-place adder.

TTK adder updated to reduce simulation time by ~7%, CG adder updated to reduce simulation time by ~30%.

Added and unified tests.

All adders are now in the Microsoft.Qunatum.**Unstable**.Arithmetic namespace to signify that they will change in future and will be moved to a separate library when we have support for it.
Note that we will entirely remove Microsoft.Qunatum.Arithmetic namespace from the standard library by moving a few remaining operations to a different namespace where they fit better. This will be in a separate PR.


library/std/arithmetic_internal.qs
```diff
@@ -7,30 +7,123 @@ namespace Microsoft.Quantum.Arithmetic {
...
     open Microsoft.Quantum.Convert;
 
+    /// # Summary
+    /// Implements the outer operation for RippleCarryAdderTTK to conjugate
+    /// the inner operation to construct the full adder. Only Length(xs)
+    /// qubits are processed.
+    ///
+    /// # Input
+    /// ## xs
+    /// LittleEndian qubit register encoding the first integer summand
```
msoeken, line 17: LittleEndian is not type anymore

msoeken (approved):

library/std/arithmetic_internal.qs
DmitryVasilevsky: Updated comments.

samples/algorithms/Shor.qs
```diff
@@ -11,6 +11,7 @@ namespace Sample {
     open Microsoft.Quantum.Random;
     open Microsoft.Quantum.Math;
     open Microsoft.Quantum.Arithmetic;
+    open Microsoft.Quantum.Unstable.Arithmetic;
```
cesarzc, line 14: I curious about the reason for using unstable functions/operations on our samples? Wouldn't it be preferable to use stable ones on our own samples code?

library/std/arithmetic_internal.qs
```diff
@@ -1,36 +1,127 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.Arithmetic {
+namespace Microsoft.Quantum.Unstable.Arithmetic {
```
cesarzc, line 4: Based on the file name, was this the intended namespace?

samples/algorithms/Shor.qs
DmitryVasilevsky: This sample uses IncByLE function from the unstable namespace because we don't have anything stable for it yet. We don't want to include this function in the standard library, but we don't have any other way of distributing library code yet. Once we have a way to distribute libraries, we will remove this from here. Hence it is unstable.

library/std/arithmetic_internal.qs
DmitryVasilevsky: Yes. All these functions are unstable. And internal.

cesarzc (approved):

compiler/qsc_frontend/src/compile.rs
```diff
@@ -401,6 +401,10 @@ pub fn std(store: &PackageStore, target: TargetProfile) -> CompileUnit {
                 "arithmetic_internal.qs".into(),
                 include_str!("../../../library/std/arithmetic_internal.qs").into(),
             ),
+            (
+                "arithmetic_unstable.qs".into(),
```
swernli, line 405: Nitpick: it would be nice to have the name be "unstable_arithmetic.qs" so that it matched the order of the namespace name and so that unstable files would all be together in the alphabetical order (assuming we have more in the future).

swernli (approved):

compiler/qsc_frontend/src/compile.rs
DmitryVasilevsky: Renamed files containing unstable code to start with "unstable_"

PR #880, opened 2023-12-01
[WIP] Include locals in the completion list

minestarks:
Lots more cleanup to be done.


compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -366,10 +445,15 @@ fn check_unit(expect: &Expect, actual: &Increment) {
                 output
             })
     );
+    let locals = format!("\nlocals:\n{:#?}", actual.ast.locals);
```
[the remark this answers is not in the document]
minestarks: Ha. I totally agree with you - I hate how this looks, but could not make up my mind on how to clean it up. These tests don't have access to the fields  `Locals` so I can't do the "pretty" formatting in the test code. I _could_ implement `Display` for `Locals` in the product crate, but that felt a little presumptuous - should I implement `Display` just to use it in these couple test cases?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -131,7 +135,91 @@ impl Scope {
...
+
+        all_locals
+    }
+
+    fn for_each_scope_at_offset<F>(&self, offset: u32, mut f: F)
+    where
+        F: FnMut(&Scope),
+    {
+        // reverse to go from innermost -> outermost
+        self.scopes.iter().rev().for_each(move |scope| {
```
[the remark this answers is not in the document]
minestarks: I don't think so? What do you mean invalidate?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -131,7 +135,91 @@ impl Scope {
...
+pub(super) type ScopeId = usize;
+
+#[derive(Debug, Clone, Default)]
+pub struct Locals {
+    // order is ascending by span (outermost -> innermost)
+    scopes: Vec<Scope>,
+}
+
+impl Locals {
+    pub(super) fn get_scopes<'a>(
```
minestarks, line 147: these can just be private

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -131,7 +135,91 @@ impl Scope {
     }
 }
 
-struct GlobalScope {
+pub(super) type ScopeId = usize;
```
minestarks, line 138: Can this be private?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -890,6 +1060,39 @@ fn resolve_scope_locals(
...
     None
 }
+
+fn get_scope_locals(scope: &Scope, offset: u32, vars: bool) -> Vec<Local> {
+    let mut names = Vec::new();
+
+    // variables
+    if vars {
+        names.extend(scope.vars.iter().filter_map(|(name, (valid_at, id))| {
+            if offset >= *valid_at {
```
minestarks, line 1070: There's a problem here

```qsharp
let var = 3;
// if you invoke completions here you won't see `var`
let var = "foo";
```

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -131,7 +135,91 @@ impl Scope {
...
+
+            if scope.kind == ScopeKind::Callable {
+                // Since local callables are not closures, hide local variables in parent scopes.
+                vars = false;
+            }
+        });
+
+        // deduping by name will effectively make locals in a child scope
+        // shadow the locals in its parent scopes
+        all_locals.dedup_by(|a, b| a.name == b.name);
```
swernli, line 187: I think you just posted a comment about this, but how can we be sure the right duplicate will get dropped here based on the cursor position?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -131,7 +135,91 @@ impl Scope {
     }
 }
 
-struct GlobalScope {
+pub(super) type ScopeId = usize;
+
+#[derive(Debug, Clone, Default)]
+pub struct Locals {
```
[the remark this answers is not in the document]
ScottCarda-MS: Resolved offline. It can just stay as "Locals".

compiler/qsc_frontend/src/resolve.rs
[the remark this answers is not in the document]
ScottCarda-MS: Maybe I'm just misunderstanding a rust thing. I thought `move` would take each scope in `scopes` and move the data out of `scopes` such that it isn't in `scopes` anymore, or at least that it would be an error to reference `scopes` after this point.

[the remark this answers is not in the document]
minestarks: Ah I see. `move` only applies to the variables captured by the closure. In this case here, that's `f`. If we capture `f` by (immutable) reference, that makes rust unhappy as we're calling `f`, which mutates it. By using `move`, we make the closure _own_ `f` instead of borrowing it.
https://doc.rust-lang.org/std/keyword.move.html

As for `self.scopes`, there's nothing special there, we're just iterating over it using `iter()`. `scope` is a reference.

compiler/qsc_frontend/src/resolve.rs
minestarks: Ah, yeah it's related to the other bug I commented on. When you have a variable shadowed in the _same_ scope:

```qsharp
let var = 3;
let var = "foo";
```

By the time we get to this code that you commented on, the first declaration doesn't even exist in the `Scope` anymore. This is because `vars` is kept as a hashmap, and the second declaration overwrites the first one, with its own `valid_at` offset.

So, deduping within the same scope isn't a concern here, because there is no dupe. Which is itself a bug. ([this one](https://github.com/microsoft/qsharp/pull/880#discussion_r1419537650)) . I'm completely stumped on how to fix it in this PR and would like to open a separate GH issue for it.

compiler/qsc_frontend/src/resolve.rs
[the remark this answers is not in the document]
minestarks: Huh. I lied. While my general explanation still is (hopefully) correct, removing the `move` here _doesn't_ cause rust to complain, because `f` is already captured by value. I have no idea why I put this there then, must be from an earlier revision of the code!

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -101,18 +101,22 @@ pub(super) enum Error {
...
+#[derive(Debug, Clone)]
+pub struct Scope {
+    span: Span,
     kind: ScopeKind,
     opens: FxHashMap<Rc<str>, Vec<Open>>,
     tys: FxHashMap<Rc<str>, ItemId>,
     terms: FxHashMap<Rc<str>, ItemId>,
-    vars: FxHashMap<Rc<str>, NodeId>,
+    // u32 is the `valid_at` - the lowest offset at which the variable name can be used
+    vars: FxHashMap<Rc<str>, (u32, NodeId)>,
```
swernli, line 112: would reversing this data structure help? Using `FxHashMap<NodeId, (u32, Rc<str>)>` would make sure that shadowed variables aren't overwritten but rather are all present.

swernli: Oh, never mind. I keep forgetting which stage of compilation we are at, and here we need to match on string because it's during resolution, not afterward.

swernli (approved):

idavis (approved):

compiler/qsc_frontend/src/resolve.rs
minestarks: added bug comment.

ScottCarda-MS (approved):

PR #881, opened 2023-12-02
[API Proposal] Array CircularlyShifted array function

filipw:
I recently needed that and was surprised that in the reach pool of built-in array functions there was no rotation function (rotate left or right).

This PR adds a `Rotated` function which performs just that. It follows the pattern of `Padded` function, in that a positive step indicates rotation right, and negative indicates a rotation left.

Hope this makes sense.


library/std/arrays.qs
```diff
@@ -985,6 +985,54 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// ## 'T
+    /// The type of the array elements.
+    ///
+    /// # Input
+    /// ## stepCount
+    /// The amount of positions by which the array elements will be shifted.
+    /// If this is positive, `array` is rotated to the right.
+    /// If this is negative, `array` is rotated to the left.
+    /// ## array
+    /// Array to be padded.
```
DmitryVasilevsky, line 1001: Array to be rotated?

DmitryVasilevsky:
I'm feel a little uncomfortable about 'Rotated' name for this function. Rotation typically means something else in quantum. Could be it confusing to the users?

filipw:
I see what you mean. Matlab refers to it as [circular shift](https://nl.mathworks.com/help/matlab/ref/circshift.html) and numpy calls it [roll](https://numpy.org/doc/stable/reference/generated/numpy.roll.html) so maybe one of these?

DmitryVasilevsky:
> I see what you mean. Matlab refers to it as [circular shift](https://www.mathworks.com/help/matlab/ref/circshift.html) and numpy calls it [roll](https://numpy.org/doc/stable/reference/generated/numpy.roll.html) so maybe one of these?

I was thinking about some sort of a shift so circular shift sounds reasonable.

filipw:
Thanks a lot for a quick feedback, I renamed it accordingly, sorted the code and force pushed.

swernli:
@filipw would you mind merging main into your branch? We're circling back to getting this merged, but we also made changes to the CI pipeline that must be in the PR for the branch rules to be satisfied.

filipw:
sure, done!

library/std/arrays.qs
```diff
@@ -985,6 +1033,54 @@ namespace Microsoft.Quantum.Arrays {
...
+    ///
+    /// # Example
+    /// ```qsharp
+    /// let array = [10, 11, 12];
+    /// // The following line returns [11, 12, 10].
+    /// let output = Rotated(2, array);
+    /// // The following line returns [12, 10, 11].
+    /// let output = Rotated(-2, array);
+    /// ```
+    function Rotated<'T> (stepCount : Int, array : 'T[]) : 'T[] {
```
swernli, line 1063: Based on the PR comments, I would've expected this to be replaced by `CircularlyShifted` but currently the change includes both even though they have the same functionality and implementation. Don't we just want to keep one?

filipw: I am sorry something went wrong between the force push and the merge 😄 should be fine now, please squash at merge

swernli (approved):

PR #884, opened 2023-12-05
Fix span offsets in `additional_text_edits` in completions

minestarks:
This bug was exposed while testing multi-file projects.

Plus, some test utilities for testing language service features with multiple sources.


language_service/src/test_utils.rs
```diff
@@ -100,44 +70,47 @@ pub(crate) fn compile_with_fake_stdlib(source_name: &str, source_contents: &str)
     }
 }
 
-pub(crate) fn compile_notebook_with_fake_stdlib_and_markers(
-    cells_with_markers: &[(&str, &str)],
+pub(crate) fn compile_project_with_fake_stdlib_and_markers(
```
ScottCarda-MS, line 73: Isn't the usual pattern to get marker offsets with `get_source_and_marker_offsets` first, then call into `compile_with_fake_stdlib`? Why do we need this function?

minestarks: Because there are multiple files, and it's nontrivial to iterate over them, extract markers, validate that there is only one cursor offset, etc. I thought it was worth encapsulating all of that in a utility.

You're right that now there's an asymmetry between this and _original_ way of compiling a single file ... I was thinking I would extract the single-file compilation into a helper method too, but that would involve touching all the test files. I'm happy to do it in this PR if you don't mind reviewing ;)

language_service/src/completion/tests.rs
```diff
@@ -241,6 +262,52 @@ fn in_block_from_other_namespace() {
...
+    check_project(
+        &[
+            (
+                "foo.qs",
+                indoc! {r#"namespace Foo { operation FooOperation() : Unit {} }
+                "#},
+            ),
+            (
+                "bar.qs",
+                indoc! {r#"namespace Bar { operation BarOperation() : Unit { ↘ } }
```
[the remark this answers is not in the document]
minestarks: You betcha

language_service/src/test_utils.rs
ScottCarda-MS: Yah I think that might be the right way to go. It will make for a bigger PR, but I'd rather have consistency in how the tests work.

language_service/src/completion/tests.rs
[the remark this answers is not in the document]
ScottCarda-MS: The arrow character indicates where the user's cursor is in the test case.

ScottCarda-MS (approved):

PR #887, opened 2023-12-05
Add get_qir_str() method to qsharp.compile() result

minestarks:
This is to support the scenario where the user wants to generate an input file to be submitted with the `az quantum` cli.


shenyingjun (approved):

pip/qsharp/_qsharp.py
```diff
@@ -115,4 +118,17 @@ def __init__(self, name: str, ll_str: str):
     # The name of this method is defined
     # by the protocol and must remain unchanged.
     def _repr_qir_(self, **kwargs) -> bytes:
-        return self._ll_str.encode("utf-8")
+        return self.get_qir_str().encode("utf-8")
+
+    def get_qir_str(self) -> str:
+        """
+        Returns the QIR LLVM string representation of this program as a bytes object.
```
minestarks, line 125: This doc comment is wrong..

```suggestion
        Returns the QIR LLVM string representation of this program.
```

pip/qsharp/_qsharp.py
```diff
@@ -115,4 +118,17 @@ def __init__(self, name: str, ll_str: str):
     # The name of this method is defined
     # by the protocol and must remain unchanged.
     def _repr_qir_(self, **kwargs) -> bytes:
-        return self._ll_str.encode("utf-8")
+        return self.get_qir_str().encode("utf-8")
+
+    def get_qir_str(self) -> str:
```
swernli, line 123: Does this need to be a method? Since it's not doing any expensive processing, it could be a nicely named property or perhaps just the default representation or display/to_string for the class? I'm not sure how Python-ic of a pattern that would be, to be fair.

minestarks: Re: method vs attribute - I wanted to futureproof this in case we ever wanted to defer generating the string, but that may be overkill.

I could implement `__str__()` which is the default string representation. I thought, what if we have other string representations in the future - do I want to squat on the default string representation now? Again, a futureproofing concern which may be overkill.

I have zero intuition about QIR generation or its future so I could use your judgment on this. I'm happy to change it to `__str__()` if we think the string representation will be stable.

minestarks: Side question: do we want to call out whether this is qir.v1, qir.v2, etc?

DmitryVasilevsky: Probably makes sense to include version if we are planning to support several at the same time...

swernli: The QIR generated already has attributes in it that describe it (the module flags) per the QIR base profile spec. Anything else we add might be confusing to consumers and would be extra description not part of the spec. I think we should leave the string as just the string and trust the QIR to self-identify.

swernli: > I could implement __str__() which is the default string representation. I thought, what if we have other string representations in the future - do I want to squat on the default string representation now?

I think this makes sense, personally, and clarifies what this actually is. Regarding your other concerns about future proofing, I do think it's good to consider just to make sure and think you are right that it would likely be overkill for now. All else fails, we can have a new class that represents a more complex object in the future, should that need arise.

minestarks: Okay, sounds like we're landing on `__str__()` , updating the PR

swernli (approved):

PR #888, opened 2023-12-05
Update UI to use profiles, compiler to use capabilities.

idavis:
The capabilities are pulled from Cesar's spike PR trying to lay the integration point so that we have the compiler only ever works with capabilities. The profiles are currently defined in `qsc` so that the compiler can't take a dependency on profiles.

I'm not sure how many capabilities we'll need to map, so the flags enum may need to change to `u64` or a bit array.

Still not sure what to call the values in the `@Config()` attribute. Currently `Base`/`Full` are used. `None`/`Any` has been proposed.

Effects #863 
Replaces #842

### DIFF
--- a/compiler/qsc/src/lib.rs
+++ b/compiler/qsc/src/lib.rs
@@ -8,9 +8,10 @@ pub mod compile;
 pub mod error;
 pub mod incremental;
 pub mod interpret;
+pub mod target;
 
 pub use qsc_frontend::compile::{
-    CompileUnit, PackageStore, SourceContents, SourceMap, SourceName, TargetProfile,
+    CompileUnit, PackageStore, RuntimeCapabilityFlags, SourceContents, SourceMap, SourceName,
 };
 
 pub mod resolve {
--- a/compiler/qsc_frontend/src/compile.rs
+++ b/compiler/qsc_frontend/src/compile.rs
@@ -12,6 +12,7 @@ use crate::{
     resolve::{self, Names, Resolver},
     typeck::{self, Checker, Table},
 };
+use bitflags::bitflags;
 use miette::{Diagnostic, Report};
 use preprocess::TrackedName;
 use qsc_ast::{
@@ -35,13 +36,24 @@ use qsc_hir::{
 use std::{fmt::Debug, str::FromStr, sync::Arc};
 use thiserror::Error;
 
+bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct RuntimeCapabilityFlags: u32 {
+        const ConditionalForwardBranching = 0b0000_0001;
+        const IntegerComputations = 0b0000_0010;
+        const FloatingPointComputation = 0b0000_0100;
+        const BackwardsBranching = 0b0000_1000;
+        const HigherLevelConstructs = 0b0001_0000;
+    }
+}
+
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub enum TargetProfile {
+pub enum ConfigAttr {
     Full,
     Base,
 }
 
-impl TargetProfile {
+impl ConfigAttr {
     #[must_use]
     pub fn to_str(&self) -> &'static str {
         match self {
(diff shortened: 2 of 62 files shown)


compiler/qsc_frontend/src/compile.rs
```diff
@@ -35,13 +36,24 @@ use qsc_hir::{
 use std::{fmt::Debug, str::FromStr, sync::Arc};
 use thiserror::Error;
 
+bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct RuntimeCapabilityFlags: u32 {
```
minestarks, line 41: What benefit does it bring us to make this a `bitflags` rather than a struct with `bool` fields?

compiler/qsc/src/lib.rs
```diff
@@ -4,13 +4,57 @@
...
 
+use std::str::FromStr;
+
 pub mod compile;
 pub mod error;
 pub mod incremental;
 pub mod interpret;
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum TargetProfile {
```
minestarks, line 15: Can we pull this code into a module? Seems out of place sitting in the top level `lib` when everything else here is a reexport.

compiler/qsc/src/lib.rs
```diff
@@ -4,13 +4,57 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Unrestricted" => Ok(TargetProfile::Unrestricted),
+            "Base" => Ok(TargetProfile::Base),
+            _ => Err(()),
+        }
+    }
+}
+
```
minestarks, line 46: These string conversions don't seem to be used anywhere.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -118,23 +118,27 @@ impl MutVisitor for Conditional {
...
-                    _ => true,
+                    // If there is no config attribute, then we assume that the item matches
+                    // the target. We can't do membership tests on the capabilities because
+                    // Base is not a subset of any capabilities, it is a lack of capabilities.
+                    ExprKind::Path(path) => match ConfigAttr::from_str(path.name.name.as_ref()) {
+                        Ok(ConfigAttr::Full) => capabilities.is_all(),
+                        Ok(ConfigAttr::Base) => capabilities.is_empty(),
+                        _ => true,
+                    },
+                    _ => true, // Unknown config attribute, so we assume it matches
```
cesarzc, line 134: `matches_config` is currently used to check whether to include an item in the compilation. However, it can silently behave in an unexpected way, most likely including something that was not meant to be included (e.g. `@Config(Ful)` when compiling for `Base`).

It might be worth to not silently include things for which the `@Config` attribute is not "well-defined", but we could wait until we decide how the attribute will be used moving forward.

compiler/qsc/src/lib.rs
idavis: Agreed, do you have any suggestions?

compiler/qsc_frontend/src/compile/preprocess.rs
idavis: This is the same behavior as the original code, just commented to explain what is going on. @swernli - do you want to weigh in here?

compiler/qsc_frontend/src/compile.rs
idavis: A struct with bool fields with have a size of `n` bytes where `n` is the number of bools. The bitflags is much denser than that.

minestarks: Is there a particular perf reason that this needs to be small? There's basically only one instance of it that we're passing around at initialization. 

No objections to leaving it this way, just probing out of curiosity.

compiler/qsc_frontend/src/compile/preprocess.rs
swernli: This behavior is intentional. The code in lowering checks whether or not the `@Config` attribute is correct, but if we drop stuff we don't recognize then it won't get checked by lowering. So we keep things we don't recognize and let it trigger the appropriate failure in a later check.

swernli (approved):

PR #889, opened 2023-12-05
Multifile Support for the VS Code Debugger

sezna:
We can't access any of the vscode api in the `npm` module, and that module gets called from within a worker, limiting our ability to pass functions back and forth. Instead, we can directly access the wasm in the `vscode` module).

Things that don't work, that I tried:
1. Loading the project in the `InlineDebugAdapterFactory` constructor. Can't have an async constructor.
2. Loading the files in the `npm` package at all. Can't access `vscode.fs` related APIs, and can't pass them in as callbacks.
3. Loading the files in the `wasm` layer -- the callbacks have to pass through a worker boundary via `npm`, so that doesn't fly either.

# Summary

- Refactor the transformers for JS async callbacks, for better readability [1](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/wasm/src/language_service.rs#L32)
- Create a shared project loader, to be used by both the debugger and the language service (eventually)  [1](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/compiler/qsc_project/src/js.rs#L31) [2](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/wasm/src/project_system.rs)
- Updates the debugger to load all project sources in the `vscode` layer. [1](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/vscode/src/debugger/session.ts#L113) [2](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/vscode/src/debugger/session.ts#L134).

# Notes

We discussed a few ways of approaching this:

One was to pass callbacks in through the JS, all the way into the WASM -- the way the language service does. This, of course, doesn't work in workers, so it won't work for the debugger. 

The next thought was to directly construct the `vscode.fs` callbacks in the `npm` directory, instead of passing them in. That way, the callbacks never cross the worker boundary. Unfortunately, we don't have access to `vscode.fs` in `npm`, so that would not work either.

The final approach, and the one that I ultimately ended up implementing, is to provide a more minimal project loading API from the WASM, allowing the `/vscode` layer to directly access our project system and load the sources ahead of time, before the worker ever needs them. This allows us to pass them in to the worker as strings, which are structurally cloneable types. This works!

The difficult (impossibility) to work with here is that the `/vscode` crate didn't have direct access to the project logic, which is contained within the wasm. Because of that, I had to directly access the wasm types from `vscode`. I'm not sure we want this, but I'm also not sure of any other way to get access to `qsc_project` while in `/vscode` , which is the *only* place we can actually use `vscode.fs`.

TODO: 
- [x] docstrings on transformer funcs
- [x] build consensus and validate that importing wasm into `vscode` is okay
  - [x] does this manner of importing wasm even work in a packaged vsix? [1](https://github.com/microsoft/qsharp/blob/983b875855cb0edfdf4ad0dacf3a0189c84b3df4/vscode/src/debugger/session.ts#L8)
  - [x] Actually, I should wrap it in an API in `npm`
- [x] figure out what the correct thing to do is [here](https://github.com/microsoft/qsharp/pull/889/files#diff-c2b7e9f81359638edbfc422c7f3df0deda6bd218866e6ebff36e57a85dc780c6R141)
- [x] test in windows

### DIFF
--- a/vscode/src/debugger/activate.ts
+++ b/vscode/src/debugger/activate.ts
@@ -194,28 +194,25 @@ export const workspaceFileAccessor: FileAccessor = {
 class InlineDebugAdapterFactory
   implements vscode.DebugAdapterDescriptorFactory
 {
-  createDebugAdapterDescriptor(
+  async createDebugAdapterDescriptor(
     session: vscode.DebugSession,
     _executable: vscode.DebugAdapterExecutable | undefined,
-  ): vscode.ProviderResult<vscode.DebugAdapterDescriptor> {
+  ): Promise<vscode.DebugAdapterDescriptor> {
     const worker = debugServiceWorkerFactory();
-    return loadProject(
-      workspaceFileAccessor.resolvePathToUri(session.configuration.program),
-    )
-      .then(
-        (sources) =>
-          new QscDebugSession(
-            workspaceFileAccessor,
-            worker,
-            session.configuration,
-            sources,
-          ),
-      )
-      .then((qscSession) =>
-        qscSession.init(getRandomGuid()).then(() => {
-          return new vscode.DebugAdapterInlineImplementation(qscSession);
-        }),
-      );
+    const uri = workspaceFileAccessor.resolvePathToUri(
+      session.configuration.program,
+    );
+    const sources = await loadProject(uri);
+    const qscSession = new QscDebugSession(
+      workspaceFileAccessor,
+      worker,
+      session.configuration,
+      sources,
+    );
+
+    await qscSession.init(getRandomGuid());
+
+    return new vscode.DebugAdapterInlineImplementation(qscSession);
   }
 }
 async function loadProject(configUri: vscode.Uri): Promise<[string, string][]> {


wasm/src/debug_service.rs
```diff
@@ -31,17 +106,32 @@ impl DebugService {
...
+        read_file: ReadFileCallback,
+        list_directory: ListDirectoryCallback,
+        get_manifest: GetManifestCallback,
     ) -> String {
-        let source_map = SourceMap::new(
-            [(path.into(), source.into())],
-            entry.as_deref().map(|value| value.into()),
-        );
+        let loader = DebugServiceProjectLoader::new(read_file, list_directory, get_manifest);
+        let manifest = loader.get_manifest(&path).await;
```
idavis, line 120: Does it make sense to avoid taking in the source contents and instead have a `try_load_project` that takes an `impl AsRef<Path>`.  Then if the project has a manifest, you try to return the project. Otherwise we create a default manifest and load the supplied path as the input file (if it is a file) returning it as a project. Then all code always assumes a project and all project logic stays in `qsc_project`. Something like this (not real code):
```rust
    fn try_load_project<T: AsRef<Path>>(&self, input: T) -> miette::Result<Project> {
        if let Some(manifest) = Manifest::load_from_path(input.as_ref())? {
            self.load_project(manifest)
        } else {
            let manifest = ManifestDescriptor {
                manifest: Manifest {
                    ..Default::default()
                },
                manifest_dir: input.as_ref().to_path_buf(),
            };
            self.load_project(manifest)
        }
    }
```

compiler/qsc_project/src/js.rs
```diff
@@ -0,0 +1,39 @@
+use std::{convert::Infallible, path::PathBuf, sync::Arc};
```
billti, line 1: License header

npm/test/basics.js
```diff
@@ -781,8 +791,10 @@ test("debug service getting breakpoints after loaded source succeeds when file n
   const debugService = getDebugServiceWorker();
   try {
     const result = await debugService.loadSource(
-      "test.qs",
-      `namespace Sample {
+      [
```
billti, line 794: Did you add any tests for multiple files, or just update the single file tests for the new structure?

vscode/src/debugger/activate.ts
```diff
@@ -3,11 +3,15 @@
...
 
-import * as vscode from "vscode";
 import { IDebugServiceWorker, getDebugServiceWorker } from "qsharp-lang";
 import { FileAccessor, qsharpExtensionId, isQsharpDocument } from "../common";
 import { QscDebugSession } from "./session";
 import { getRandomGuid } from "../utils";
+import { getManifest, readFile, listDir } from "../projectSystem.js";
+
+import * as vscode from "vscode";
+import * as wasm from "../../../npm/lib/web/qsc_wasm.js";
```
billti, line 13: I don't see 'wasm' used anywhere in this file. Why is this needed?

vscode/src/debugger/session.ts
```diff
@@ -5,6 +5,8 @@
 
 import * as vscode from "vscode";
 
+import * as wasm from "../../../npm/lib/web/qsc_wasm.js";
```
billti, line 8: The npm package already exposes methods to load the wasm and call methods on it, which is what is used everywhere else (VS Code and browser). We shouldn't be loading files directly from the npm folder structure, but by importing from 'qsharp-lang' (and adding new methods and types there if needed)

minestarks (approved):
:shipit:

PR #893, opened 2023-12-07
Added comparisons to Arithmetic namespace

DmitryVasilevsky:
Added comparisons and tests.

### DIFF
--- a/library/std/unstable_arithmetic.qs
+++ b/library/std/unstable_arithmetic.qs
@@ -540,11 +540,15 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
     /// Computes `target ^= (c == x)`, that is, inverts `target`
     /// if a BigInt value `c` is equal to the little-endian qubit register `x`
     operation InvertIfEqualL (c : BigInt, xs : Qubit[], target : Qubit) : Unit is Adj + Ctl {
-        let bits = BigIntAsBoolArray(c, Length(xs));
-        within {
-            ApplyPauliFromBitString(PauliX, false, bits, xs);
-        } apply {
-            Controlled X(xs, target);
+        let cBitSize = BitSizeL(c);
+        let xLen = Length(xs);
+        if (cBitSize <= xLen) {
+            let bits = BigIntAsBoolArray(c, Length(xs));
+            within {
+                ApplyPauliFromBitString(PauliX, false, bits, xs);
+            } apply {
+                Controlled X(xs, target);
+            }
         }
     }
 
--- a/library/tests/src/resources/compare.qs
+++ b/library/tests/src/resources/compare.qs
@@ -15,7 +15,7 @@ namespace Test {
             use qs = Qubit[n];
             use t = Qubit();
 
-            for a in 0 .. 2^n-1 {
+            for a in 0 .. 2^n+1 { // We want b to have more bits sometimes...
                 for b in 0 .. 2^n-1 {
                     ApplyXorInPlace(b, qs);
                     quantumComparator(IntAsBigInt(a), qs, t);


library/std/convert.qs
```diff
@@ -97,6 +97,36 @@ namespace Microsoft.Quantum.Convert {
         result
     }
 
+    /// # Summary
+    /// Produces a binary representation of a non-negative BigInt, using the
+    /// little-endian representation for the returned array.
+    ///
+    /// # Input
+    /// ## number
+    /// A non-negative BigInt to be converted to an array of boolean values.
```
msoeken, line 106: Just noticing an inconsistency of boolean vs. Boolean, but I see both versions throughout the docs strings in the main branch.

msoeken (approved):

library/std/convert.qs
DmitryVasilevsky: Updated to use capital letter - Boolean.

library/std/unstable_arithmetic_internal.qs
```diff
@@ -404,4 +414,150 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
         }
     }
 
+    //
+    // Internal operations for comparisons
+    //
+
+    internal operation CompareGreaterThanOrEqualConstant<'T>(
```
cesarzc, line 421: Even though these are internal operations, it would be nice to have a summary of what it does since it is fairly complex. Additionally, it would be helpful to name the input parameters to reflect what they represent.

library/std/unstable_arithmetic.qs
```diff
@@ -511,4 +512,223 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
...
+            Controlled X(xs, target);
+        }
+    }
+
+    /// # Summary
+    /// Computes `target ^= (c >= x)`, that is, inverts `target`
+    /// if a BigInt value `c` is greater or equal to the little-endian qubit register `x`
+    operation InvertIfGreaterOrEqualL (c : BigInt, x : Qubit[], target : Qubit) : Unit is Adj {
+        CompareGreaterThanOrEqualConstant(false, X, c + 1L, x, target);
+        X(target);
```
cesarzc, line 556: Isn't the inversion already happening when calling `CompareGreaterThanOrEqualConstant`?

library/std/unstable_arithmetic_internal.qs
cesarzc: Maybe this could be named `ApplyActionIfGreaterThanOrEqualConstant`?

library/std/unstable_arithmetic_internal.qs
```diff
@@ -404,4 +414,150 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
...
+                for i in 0..Length(cs1)-1 {
+                    let op =
+                        cNormalized &&& (1L <<< (i+1)) != 0L ?
+                        ApplyAndAssuming0Target | ApplyOrAssuming0Target;
+                    op(cs1[i], xNormalized[i+1], qs[i]);
+                }
+            } apply {
+                let control = IsEmpty(qs) ? Tail(x) | Tail(qs);
+                within {
+                    if invertControl {
```
cesarzc, line 461: Why is `invertControl` needed?

library/std/unstable_arithmetic_internal.qs
DmitryVasilevsky: I added summary and description of the parameters. Also renamed function as you suggested.

library/std/unstable_arithmetic.qs
DmitryVasilevsky: I updated the code and hopefully it's easier to understand now. I also set invertControl to true to avoid additional X on target.
The function is named InvertIfGreaterOrEqualL(c, x), or it should invert target if c >= x. The helper function it calls checks if x is greater or equal than constant (as the name suggests), so it checks x >= c. We use this helper function as follows: c>=x  ⇔  c+1>x  ⇔  x<c+1  ⇔ not(x>=c+1). So we add 1 to c and invert control.

cesarzc (approved):
It looks much better and easy to understand now, thank you @DmitryVasilevsky!

PR #894, opened 2023-12-07
Throw vscode dialog if qsharp.json isn't valid

sezna:
@idavis pointed out that if a `qsharp.json` file is invalid (empty, or invalid JSON), we log to the
extension logs and fail. This is not a great user experience, as the error is almost completely
undiscoverable.


minestarks:
Showing up a dialog every time we try to happen to parse the manifest is not good UX...  Can't we just show this in the problems view? (i.e. surface it as an error from the language service?) If this is supposed to be a quick-and-dirty fix - and you don't want to mess with the language service right now - you can also manipulate the Problems list directly through the vscode api (and clear the error when once we've successfully parsed the qsharp.json)

PR #899, opened 2023-12-08
Removed Arithmetic namespace from standard library

DmitryVasilevsky:
Moved ApplyXorInPlace and ApplyXorInPlaceL to Canon, and MeasureInteger to Measurement namespace.
Removed Microsoft.Quantum.Arithmetic namespace that was left empty.
The rest is a fallout.
(Also removed a few unnecessary explicit references to Microsoft.Quantum.Canon namespace, but mostly by accident.)


swernli (approved):
The changes look good to me, might be worth directly confirming with @cesarzc about kata content change and whether there is anyone else we should notify. Note that unless the katas get re-deployed they will still use the old code, which when opened in VSCode might show errors.

katas/content/multi_qubit_measurements/measuring_one.qs
```diff
@@ -1,6 +1,6 @@
 namespace Kata {
     open Microsoft.Quantum.Diagnostics;
-    open Microsoft.Quantum.Arithmetic;
+    open Microsoft.Quantum.Measurement;
```
swernli, line 3: This change would need to be deployed to the katas site, which we should coordinate to make sure there isn't a difference between katas namespaces and what VSCode.dev supports.

samples/algorithms/DeutschJozsa.qs
```diff
@@ -8,7 +8,6 @@
 ///
 /// This Q# program implements the Deutsch–Jozsa algorithm.
 namespace Sample {
-    open Microsoft.Quantum.Canon;
```
cesarzc, line 11: Was this intentional?

cesarzc:
> The changes look good to me, might be worth directly confirming with @cesarzc about kata content change and whether there is anyone else we should notify. Note that unless the katas get re-deployed they will still use the old code, which when opened in VSCode might show errors.

We need to coordinate a deployment of this package to the Katas site, but since the package will have a lot of large changes, we should be diligent testing the new package before deploying it.

samples/algorithms/DeutschJozsa.qs
DmitryVasilevsky: I believe Canon is open by default.

katas/content/multi_qubit_measurements/measuring_one.qs
DmitryVasilevsky: Cesar says we'll coordinate once this is in.

PR #908, opened 2023-12-11
Pin hatchling to 1.18

swernli:
We eventually need to update build settings to adjust to https://github.com/pypa/hatch/issues/1113.

### DIFF
--- a/jupyterlab/pyproject.toml
+++ b/jupyterlab/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.4.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
+requires = ["hatchling<=1.18.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
 build-backend = "hatchling.build"
 
 [project]


jupyterlab/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.4.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
+requires = ["hatchling<=1.18.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
```
minestarks, line 2: Could you add a comment with a link to the issue you're working around?

(I'm trying to avoid a future like [this](https://github.com/microsoft/iqsharp/blob/main/src/Python/requirements.txt) where you can never be sure exactly what you're breaking by unpinning a dependency)

minestarks (approved):

ofek:
You need to provide something like this https://hatch.pypa.io/latest/config/build/#packages

Previously the wheel target had a fallback to include every Python file anywhere it could find that didn't look like a test file.

swernli:
@ofek Looks like I needed both that and a force includes section to make sure that `./setup.py` gets added as well.

swernli:
I'll replace this PR with one that adds the newly required sections to the pyproject.toml.

PR #909, opened 2023-12-11
Fix hatchling build by adding required sections

swernli:
This adds the newly required sections to the hatchling build and making sure the right files get included in the wheel.

See https://github.com/pypa/hatch/issues/1113 for context.

### DIFF
--- a/jupyterlab/pyproject.toml
+++ b/jupyterlab/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.4.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
+requires = ["hatchling>=1.19.0", "jupyterlab==4.0.0", "hatch-nodejs-version"]
 build-backend = "hatchling.build"
 
 [project]
@@ -59,6 +59,12 @@ build_cmd = "build:dev"
 npm = ["jlpm"]
 source_dir = "src"
 build_dir = "qsharp-jupyterlab/labextension"
+
+[tool.hatch.build.targets.wheel]
+packages = ["qsharp-jupyterlab"]
+
+[tool.hatch.build.targets.wheel.force-include]
+"./setup.py" = "setup.py"
 
 [tool.check-wheel-contents]
 ignore = ["W002"]


idavis (approved):

jupyterlab/pyproject.toml
```diff
@@ -60,5 +60,11 @@ npm = ["jlpm"]
 source_dir = "src"
 build_dir = "qsharp-jupyterlab/labextension"
 
+[tool.hatch.build.targets.wheel]
+packages = ["qsharp-jupyterlab"]
+
+[tool.hatch.build.targets.wheel.force-include]
```
ofek, line 66: Do you want this in the wheel or source distribution?

jupyterlab/pyproject.toml
```diff
@@ -60,5 +60,11 @@ npm = ["jlpm"]
 source_dir = "src"
 build_dir = "qsharp-jupyterlab/labextension"
 
+[tool.hatch.build.targets.wheel]
+packages = ["qsharp-jupyterlab"]
```
ofek, line 64: This directory does not exist, are you sure this is correct?

swernli: that folder is a child folder under jupyterlab, and from testing locally it looks like it pulls in the right contents.

jupyterlab/pyproject.toml
swernli: It was in the wheel previously, we don't do an sdist for this but I wanted to make sure the contents were the same as before.

jupyterlab/pyproject.toml
ofek: Sorry to bother here, I'm just trying to help! Which of the following mappings do you want the wheel to use as a top-level directory?

1. `jupyterlab/qsharp-jupyterlab` ->`qsharp-jupyterlab` (`packages = ["jupyterlab/qsharp-jupyterlab"]`)
2. `jupyterlab` ->`jupyterlab` (`packages = ["jupyterlab"]`)

swernli: when I unzip the wheel, the contents look like:
image
this matches the contents of the wheel that we have before the upgrade to hatch 1.19, so I think this is intentional.

ofek: Okay thank you! I would strongly recommend the following config if it produces the contents that you expect:

```toml
[tool.hatch.build.targets.wheel]
packages = ["jupyterlab/qsharp-jupyterlab"]
```

ofek: Oh I see what's happening, there are multiple project files in this repository and I was not looking at where it was located. My apologies! Everything looks great

ofek (approved):

PR #910, opened 2023-12-11
Added some unimplemented functions and operations

DmitryVasilevsky:
This adds a few functions we removed with `@Unimplemented` attribute and a doc comment to direct users to the suggested replacement. This will help users to fix their code. The functions do not include implementations.
Known problems:
* We have to re-introduce namespaces we removed
* Unimplemented functions show in completion lists

NOTE: Not for merging. For experiments only.


DmitryVasilevsky:
This was deemed wrong. Removing.

PR #915, opened 2023-12-12
Async updates for the language service

minestarks:
This PR allows us to use async methods (e.g. for loading the project) when updating the compilation state in the language service, without interfering with read operations such as `getCompletions`. 

We can divide language service methods into two kinds of operations: reads and updates. The `CompilationState` is shared.
* Read operations, which are always synchronous, immutably borrow the `CompilationState`.
* Update operations, instead of performing the update directly, post a message to a channel which is then read by an update worker that processes the updates one by one. The update worker can do asynchronous work without interfering with any other language service operations. The update worker only mutably borrows the `CompilationState` briefly to update it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,17 +87,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2e1373abdaa212b704512ec2bd8b26bd0b7d5c3f70117411a5d9a451383c859"
 
 [[package]]
-name = "async-recursion"
-version = "1.0.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fd55a5ba1179988837d24ab4c7cc8ed6efdeff578ede0416b4225a5fca35bd0"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "async-trait"
 version = "0.1.74"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1096,7 +1085,6 @@ dependencies = [
 name = "qsls"
 version = "0.0.0"
 dependencies = [
- "async-recursion",
  "async-trait",
  "enum-iterator",
  "expect-test",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,6 +357,95 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4f4cdac9e4065d7c48e30770f8665b8cef9a3a73a63a4056a33a5f395bc7cf75"
 
 [[package]]
+name = "futures"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da0290714b38af9b4a7b094b8a37086d1b4e61f2df9122c3cad2577669145335"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-channel"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ff4dd66668b557604244583e3e1e1eada8c5c2e96a6d0d6653ede395b78bbacb"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+]
+
+[[package]]
+name = "futures-core"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eb1d22c66e66d9d72e1758f0bd7d4fd0bee04cad842ee34587d68c07e45d088c"
+
+[[package]]
+name = "futures-executor"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0f4fb8693db0cf099eadcca0efe2a5a22e4550f98ed16aba6c48700da29597bc"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-io"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8bf34a163b5c4c52d0478a4d757da8fb65cabef42ba90515efee0f6f9fa45aaa"
+
+[[package]]
+name = "futures-macro"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "53b153fd91e4b0147f4aced87be237c98248656bb01050b96bf3ee89220a8ddb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "futures-sink"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e36d3378ee38c2a36ad710c5d30c2911d752cb941c00c72dbabfb786a7970817"
+
+[[package]]
+name = "futures-task"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "efd193069b0ddadc69c46389b740bbccdd97203899b48d09c5f7969591d6bae2"
+
+[[package]]
+name = "futures-util"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a19526d624e703a3179b3d322efec918b6246ea0fa51d41124525f00f1cc8104"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-io",
+ "futures-macro",
+ "futures-sink",
+ "futures-task",
+ "memchr",
+ "pin-project-lite",
+ "pin-utils",
+ "slab",
+]
+
+[[package]]
 name = "fuzz"
 version = "0.0.0"
 dependencies = [
@@ -697,6 +786,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8afb450f006bf6385ca15ef45d71d2288452bc3683ce2e2cacc0d18e4be60b58"
 
 [[package]]
+name = "pin-utils"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
+
+[[package]]
 name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -964,6 +1059,7 @@ name = "qsc_wasm"
 version = "0.0.0"
 dependencies = [
  "expect-test",
+ "futures-util",
  "getrandom",
  "indoc",
  "js-sys",
@@ -1004,6 +1100,8 @@ dependencies = [
  "async-trait",
  "enum-iterator",
  "expect-test",
+ "futures",
+ "futures-util",
  "indoc",
  "log",
  "miette",
@@ -1208,6 +1306,15 @@ dependencies = [
  "itoa",
  "ryu",
  "serde",
+]
+
+[[package]]
+name = "slab"
+version = "0.4.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f92a496fb766b417c996b9c5e57daf2f7ad3b0bebe1ccfca4856390e3d3bb67"
+dependencies = [
+ "autocfg",
 ]
 
 [[package]]
--- a/language_service/src/lib.rs
+++ b/language_service/src/lib.rs
@@ -135,7 +135,7 @@ impl LanguageService {
     pub fn update_notebook_document<'b, I>(
         &mut self,
         notebook_uri: &str,
-        notebook_metadata: &NotebookMetadata,
+        notebook_metadata: NotebookMetadata,
         cells: I,
     ) where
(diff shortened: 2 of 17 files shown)


Cargo.lock
minestarks, line 1: NOTE TO REVIEWER: I still need to add lots of documentation to the changes in this PR. The code changes are done, feel free to review if you don't mind the lack of comments.

language_service/src/lib.rs
```diff
@@ -445,146 +217,166 @@ impl<'a> LanguageService<'a> {
...
+                    }
+                }
+            }
+        }
+        Update::NotebookDocument { notebook_uri, .. } => {
+            if let Some(last) = pending_updates.last_mut() {
+                if let Update::NotebookDocument {
+                    notebook_uri: last_uri,
+                    ..
+                } = last
```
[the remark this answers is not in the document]
minestarks: Ah. I thought this was a great suggestion but when I applied it I realized I need the `last` binding here: `*last = update` . So unfortunately I can't combine these patterns into one.

language_service/src/lib.rs
```diff
@@ -445,146 +217,166 @@ impl<'a> LanguageService<'a> {
...
+}
 
-        // Clear errors from any documents that previously had errors
-        for uri in last_docs_with_errors.difference(&self.documents_with_errors) {
-            self.publish_diagnostics_for_doc(uri, vec![]);
+impl UpdateWorker<'_> {
+    #[allow(clippy::await_holding_refcell_ref)]
+    pub async fn run(&mut self) {
+        while let Some(update) = self.recv.next().await {
+            trace!("start applying updates");
```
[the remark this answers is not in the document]
minestarks: Hmm, I took a look at how it looks in practice. We actually had a _different_, enormous, event, that was generating a lot of noise (publish diagnostics), I fixed that. Now when typing in a document generates the following events in `trace` mode:

```
2023-12-12 10:17:13.140 [trace] [%s] update_document: file:///c%3A/src/qsharp/samples/language/DataTypes.qs 21 qsls
2023-12-12 10:17:13.141 [trace] [%s] start applying updates qsls
2023-12-12 10:17:13.239 [trace] [%s] asynchronous callback from wasm returned JsValue(null) qsc_wasm::language_service
2023-12-12 10:17:13.240 [trace] [%s] Running in single file mode qsls::state
2023-12-12 10:17:13.240 [trace] [%s] compiling single-file document file:///c%3A/src/qsharp/samples/language/DataTypes.qs qsls::compilation
2023-12-12 10:17:13.240 [trace] [%s] Loaded project uri file:///c%3A/src/qsharp/samples/language/DataTypes.qs with 1 sources qsls::state
2023-12-12 10:17:13.240 [trace] [%s] publishing diagnostics for file:///c%3A/src/qsharp/samples/language/DataTypes.qs qsls::state
2023-12-12 10:17:13.240 [trace] [%s] publishing diagnostics for file:///c%3A/src/qsharp/samples/language/DataTypes.qs Some(21)): 2 errors qsls::state
2023-12-12 10:17:13.241 [trace] [%s] end applying updates updates qsls
```

Which I think is an appropriate amount for this level - considering you'd only want `trace` level if you're debugging a race condition or something. 

I did notice the typo in the trace message though, fixing that. Also logging the # of updates being applied to make it a bit more useful

minestarks:
> I notice that we don't check if the queue is getting backed up. Using a bounded queue (even with large bounds -- like 100-1000) would eliminate that as a possible failure mode. Do you think that's worth addressing?

This is a great question. I think we need to step back and ask ourselves "wtf are we doing" if we get even _close_ to 100 messages backed up in the queue. Hence the `warn!` message about perf. Which I know doesn't _fix_ anything but should at least prompt us to look for optimizations if we ever see it in real usage.

It would be nice to go further and use a bounded queue to force us "do something" if the queue is full, but what can the sender even do? The sender is calling from synchronous code and needs to post a message and get out fast. Blocking/awaiting here would defeat the purpose of the whole system. Dropping messages will lead to inconsistent compiler state. Having some kind of retry loop is equivalent to having _another_ message queue, and what's the point. The receiver is already draining messages as fast as it can, backpressure or not. So I don't know what benefit a bounded queue would bring in this setup.

language_service/src/lib.rs
```diff
@@ -16,145 +16,70 @@ mod qsc_utils;
...
+                list_directory,
+                get_manifest,
+            ),
+            recv,
+        };
+        self.state_updater = Some(send);
+        worker
+    }
+
+    ///
```
idavis, line 78: Looks like a comment was intended but forgotten

language_service/src/lib.rs
```diff
@@ -164,18 +89,9 @@ impl<'a> LanguageService<'a> {
     /// LSP: workspace/didChangeConfiguration
     pub fn update_configuration(&mut self, configuration: &WorkspaceConfigurationUpdate) {
         trace!("update_configuration: {configuration:?}");
```
idavis, line 91: Pass owned value of config and avoid the clone below.

language_service/src/state.rs
```diff
@@ -0,0 +1,465 @@
...
+    {
+        let state = self.state.borrow();
+        f(&state)
+    }
+
+    fn with_state_mut<F, T>(&self, f: F) -> T
+    where
+        F: FnOnce(&mut CompilationState) -> T,
+    {
+        let mut state = self.state.borrow_mut();
```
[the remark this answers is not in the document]
minestarks: Yeah good concern, I'm absolutely certain this won't panic. I can explain in a comment too, but here's the breakdown:

So, the `CompilationStateUpdater` isn't `Send`/`Sync` (nothing is - we're always single threaded). So we have this blanket guarantee that this state will never be shared across threads. That's one reassurance.

So the only remaining thing to be concerned about would be `await` points. The current code doesn't keep the mutable borrow across an await point. And it never ever should in the future. There is a comment warning against this:
https://github.com/microsoft/qsharp/blob/19f5fb5cd6ec59cf8eaf7f4ac418f75866dd509c/language_service/src/state.rs#L88-L91

In case the comment doesn't have enough teeth there's also a Clippy warning (which saved me many times in development)
https://rust-lang.github.io/rust-clippy/master/#/await_holding_refcell_ref

I'll add another comment right by this `borrow_mut` so no one is ever tempted to change it to `async` in the future.

language_service/src/protocol.rs
```diff
@@ -100,7 +100,7 @@ pub struct ParameterInformation {
     pub documentation: Option<String>,
 }
 
-#[derive(Default)]
+#[derive(Default, Clone)]
```
idavis, line 103: Should this be `Copy` instead of `Clone`?

language_service/src/state.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// It increases strictly with each text change, though this knowledge should
+    /// not be important. The version is only ever used when publishing
+    /// diagnostics to help the client associate the list of diagnostics
+    /// with a snapshot of the document.
+    pub version: u32,
+    pub compilation: CompilationUri,
+}
+
+#[derive(Debug, Clone)]
+struct Configuration {
```
idavis, line 67: Impl `Copy` instead of `Clone`?

language_service/src/state.rs
```diff
@@ -0,0 +1,465 @@
...
+    fn default() -> Self {
+        Self {
+            target_profile: Profile::Unrestricted,
+            package_type: PackageType::Exe,
+        }
+    }
+}
+
+#[derive(Default, Clone)]
+struct PartialConfiguration {
```
idavis, line 82: Impl Copy instead of Clone?

language_service/src/state.rs
```diff
@@ -0,0 +1,465 @@
...
+
+        let need_recompile = self.apply_configuration(configuration);
+
+        // Some configuration options require a recompilation as they impact error checking
+        if need_recompile {
+            self.recompile_all();
+        }
+    }
+
+    pub(super) async fn update_document(&mut self, uri: &str, version: u32, text: &str) {
```
idavis, line 148: This method is an exmple of the project special casing we discussed. Loading a project from a path should always return a project/manifest regardless of whether there is a project. Then all logic is merged.

language_service/src/tests.rs
```diff
@@ -267,427 +79,268 @@ async fn target_profile_update_fixes_error() {
...
         "namespace Foo { open Microsoft.Quantum.Diagnostics; @EntryPoint() operation Main() : Unit { DumpMachine() } }",
     );
 
+    // we intentionally don't await work to test how LSP features function when
+    // a document hasn't fully loaded
+
     // this should be empty, because the doc hasn't loaded
     assert!(ls.get_completions("foo.qs", 76).items.is_empty());
+
+    worker.apply_pending().await;
```
idavis, line 224: Why are we awaiting after all asserts?

npm/src/language-service/language-service.ts
```diff
@@ -142,7 +147,7 @@ export class QSharpLanguageService implements ILanguageService {
   }
 
   async updateConfiguration(config: IWorkspaceConfiguration): Promise<void> {
```
idavis, line 149: Why is this function async when we don't await?

npm/src/language-service/language-service.ts
```diff
@@ -142,7 +147,7 @@ export class QSharpLanguageService implements ILanguageService {
...
-    await this.languageService.update_configuration(config);
+    this.languageService.update_configuration(config);
   }
 
   async updateDocument(
```
idavis, line 153: Why is this function async when we don't await?

npm/src/language-service/language-service.ts
```diff
@@ -151,7 +156,7 @@ export class QSharpLanguageService implements ILanguageService {
     code: string,
   ): Promise<void> {
     this.code[documentUri] = code;
-    await this.languageService.update_document(documentUri, version, code);
+    this.languageService.update_document(documentUri, version, code);
   }
 
   async updateNotebookDocument(
```
idavis, line 162: Why is this function async when we don't await?

npm/src/language-service/language-service.ts
```diff
@@ -166,24 +171,20 @@ export class QSharpLanguageService implements ILanguageService {
...
     }
-    await this.languageService.update_notebook_document(
-      notebookUri,
-      metadata,
-      cells,
-    );
+    this.languageService.update_notebook_document(notebookUri, metadata, cells);
   }
 
   async closeDocument(documentUri: string): Promise<void> {
```
idavis, line 177: Why is this function async when we don't await?

npm/src/language-service/language-service.ts
```diff
@@ -166,24 +171,20 @@ export class QSharpLanguageService implements ILanguageService {
...
     delete this.code[documentUri];
-    await this.languageService.close_document(documentUri);
+    this.languageService.close_document(documentUri);
   }
 
   async closeNotebookDocument(
```
idavis, line 182: Why is this function async when we don't await?

wasm/Cargo.toml
```diff
@@ -15,6 +15,7 @@ crate-type = ["rlib", "cdylib"]
 doctest = false
 
 [dependencies]
+futures-util = { workspace = true }
```
idavis, line 18: It may be the web interface, but I can't see where this dep is used.

npm/src/language-service/language-service.ts
minestarks: Same answer for all your comments: `ILanguageService`, which this class implements, has to define all of its methods as `async`. This is because `ILanguageService` can be proxied to a web worker (same as the debugger and compiler). So even though the implementation you're looking at is synchronous, the _interface_ needs to be async, to account for the fact that we may be running the language service in a worker.

wasm/Cargo.toml
minestarks: It's here:
https://github.com/microsoft/qsharp/blob/19f5fb5cd6ec59cf8eaf7f4ac418f75866dd509c/language_service/src/lib.rs#L27

language_service/src/state.rs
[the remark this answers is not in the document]
minestarks: I don't really know how to protect against your examples except to say, "don't do that"? I don't think the patterns you showed exist in the code today, and I agree they're bad. Your examples have nested borrows in the same function, which is fairly easy to catch in development. I tried to address the less obvious cases (await points) in my other comment. 

Anything else I can do to cover all bases here?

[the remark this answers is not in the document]
minestarks: I'll leave a big scary comment. Not disagreeing that care should be taken. 

In general I'd suggest keeping `with_borrow_mut` scopes really small so you don't accidentally introduce reentrancy. But I feel like you'd have to be trying pretty hard to shoot yourself in the foot.

language_service/src/state.rs
minestarks: Why? `Copy` requires `Clone` so I'd have to implement both. Nobody requires `Copy` on this struct at the moment.

language_service/src/tests.rs
minestarks: Yeah, I guess I had it here for "completeness" since in real usage, updates would always be applied. And clippy didn't like that `worker` was unused. Basically this and the test below call `get_completions()` and `apply_pending()` in flipped order and I wanted that to be clear. But perhaps it looks confusing instead - I'll remove it.

language_service/src/state.rs
minestarks: Finally got it. The phrase "instead of" threw me off - you need to derive `Copy` _and_ `Clone` . Anyway it's a good change thank you!

language_service/src/state.rs
minestarks: Forwarding this comment to @sezna  to take into account as he reworks the project load / open buffer merge code path. I didn't want to touch any of the logic here since he's working off my branch.

minestarks:
Will close and reopen this to trigger checks. I may have messed something up.

PR #918, opened 2023-12-12
Tracking branch: Multi-File Support

sezna:
This PR contains the re-application of #901, plus:
1. QIR generation for projects
1. RE for projects
1. Histogram for projects
1. Prioritize open text buffers over disk contents, if present
1. Various bugfixes and memory usage improvements 
1. Python project support


wasm/src/project_system.rs
```diff
@@ -0,0 +1,272 @@
...
+
+pub(crate) fn to_js_function(val: JsValue, help_text_panic: &'static str) -> js_sys::Function {
+    let js_ty = val.js_typeof();
+    TryInto::<js_sys::Function>::try_into(val).unwrap_or_else(|_| {
+        panic!(
+            "expected a valid JS function ({help_text_panic}), received {:?}",
+            js_ty
+        )
+    })
+}
```
swernli, line 95: Newer clippy gets mad about this line because the conversion is actually infallible. If you want to check the type, you should probably use `.is_function()` and panic if the result is false, then just return the `Into` conversion.


```suggestion
    assert!(
        val.is_function(),
        "expected a valid JS function ({help_text_panic}), received {:?}",
        js_ty
    );
    Into::<js_sys::Function>::into(val)
```

wasm/src/lib.rs
```diff
@@ -209,7 +236,7 @@ where
                 // TODO: handle multiple errors
                 // https://github.com/microsoft/qsharp/issues/149
```
swernli, line 237: I know this isn't part of your changes, but I'm pretty sure this todo can be removed. Evaluation can only ever return one error that caused evaluation to end, so the vector is known to be of size one. I missed this in my last refactor.

vscode/src/diagnostics.ts
```diff
@@ -4,13 +4,23 @@
...
 } from "qsharp-lang";
 import * as vscode from "vscode";
 import { loadDocument, qsharpLanguageId } from "./common.js";
 
 export function startCheckingQSharp(
   languageService: ILanguageService,
+): vscode.Disposable[] {
+  return [
+    ...startLanguageServiceDiagnostics(languageService),
+    ...startQsharpJsonDiagnostics(),
```
swernli, line 18: This isn't really an issue for now, since json doens't produce any diagnositics, but if it does in the future I think it would be helpful to have those errors first in the list as they are likely to affect the way the rest of the files are processed. I know I sometimes work my way through the problems list from top to bottom trying to fix things, and would hate to get all the way to the bottom and see that the real problem was a misconfiguration of the project itself.

npm/src/compiler/compiler.ts
```diff
@@ -17,13 +17,13 @@ export interface ICompiler {
...
   run(
-    code: string,
+    sources: [string, string][],
     expr: string,
     shots: number,
     eventHandler: IQscEventTarget,
   ): Promise<void>;
-  getQir(code: string): Promise<string>;
-  getEstimates(code: string, params: string): Promise<string>;
+  getQir(sources: [string, string][]): Promise<string>;
```
swernli, line 25: We'll need to coordinate this with consumers, as this is a breaking API change. Is it worth exploring alternatives? As much as it would bloat the interface, we could leave the existing APIs unchanged and have them wrap new ones that take arrays of sources. That would make it easy for consumers to switch to the project-aware APIs when they are ready and leave the behavior unchanged in the meantime.

npm/src/main.ts
```diff
@@ -96,9 +115,17 @@ export function getDebugServiceWorker(): IDebugServiceWorker {
...
-export function getLanguageService(): ILanguageService {
+export function getLanguageService(
+  readFile?: (uri: string) => Promise<string | null>,
+  listDir?: (uri: string) => Promise<[string, number][]>,
+  getManifest?: (uri: string) => Promise<{
+    excludeFiles: string[];
+    excludeRegexes: string[];
+    manifestDirectory: string;
+  } | null>,
+): ILanguageService {
```
swernli, line 126: Just to make sure I understand, the expected behavior when `getLanguageService()` is called without passing any of these parameters is that the callbacks are null and it's the responsibility of the caller to not use any of the file system APIs? From what I can tell, there are no checks after this point so the language service is just holding those null pointers and may try to call them if the file APIs are used when the langauge service hasn't been initialized with callbacks, resulting in a null dereference crash. Not sure if it's worth the extra effort to add checks to the Rust internals that panic with a message before trying to invoke callbacks versus letting it crash and forcing a dev to connect a debugger... it's a development scenario either way, so it's really a question of how easy we want it to be for developers to diagnose the crash when they are using the npm module improperly.

wasm/src/project_system.rs
swernli: whoops, looks like something in the formatting here got messed up, so github isn't going to put the suggestion in the right spot. Might have to copy paste manually.

swernli (approved):

wasm/src/project_system.rs
swernli: will be addressed in follow up PRs on main

vscode/src/diagnostics.ts
swernli: will be addressed in follow up PRs on main

npm/src/compiler/compiler.ts
swernli: @minestarks, @sezna, let's circle back to this thought in a later PR.

PR #922, opened 2023-12-13
Add histogram widget

billti:
This adds a histogram IPyWidget similar to that used in the playground and VS Code, complete with animating incoming results, menu, zoom, scroll, etc. as shown below.

Open to discussing the API. By having 'run' on the widget launch the execution, it allows it to easily wire up events and animate as they come in.

image


widgets/src/qsharp_widgets/__init__.py
```diff
@@ -35,3 +36,46 @@ class EstimateDetails(anywidget.AnyWidget):
...
+        if time.time() - self._last_message >= 0.05:
+            self._update_ui()
+
+    def __init__(self, entry_expr=None, shot_count=None):
+        super().__init__()
+        if entry_expr and shot_count:
+            self.run(entry_expr, shot_count)
+
+    def run(self, expr, shots):
+        import qsharp
```
swernli, line 76: Now that this imports qsharp, should qsharp be listed as a dependency of the package? Alternatively, this could do a try-catch and wrap the exception with some additional help text to tell the user that histograms depend on qsharp and they need to ensure it is installed properly.

swernli (commented):
I'm still wondering if we need an additional API that lets someone construct a histogram from a results list they've already got, rather than forcing them to rerun the simulation in that scenario.

pip/qsharp/_qsharp.py
```diff
@@ -84,6 +84,22 @@ def callback(output):
     return get_interpreter().run(entry_expr, shots, callback)
 
 
+def run_histogram(entry_expr, shot_count, on_result=None):
```
minestarks, line 87: I'd be happy to just fold this into the `run()` method. `on_result` can be an optional named argument, and if it's passed in, then we run in "callback mode".

```python
def run(entry_expr, *shots, on_result=None):
  if on_result is None:
      def callback(output):
          print(output)
      return get_interpreter().run(entry_expr, shots, callback)
  else:
      results = []
      
      def OnEvent(output):
          results[len(results) - 1]["events"].append(output)
      
      for i in range(shot_count):
          results.append({"events": [], "result": None})
          results[i]["result"] = (get_interpreter().run(entry_expr, 1, OnEvent))[0]
          if on_result:
              on_result(results[i])
      
      return results
```

Later on, maybe the native interpreter take result callbacks directly, then we can clean this up without changing the api signature.

minestarks: Now that I think about it, I'm not sure if we even need that first `if` case.... we can just use the full implementation from `run_histogram`. I think I just didn't want the `shots` parameter on the `run()` method to go to waste :)

swernli: I'm inclined to agree. Then native code refactoring can be a follow up.

widgets/src/qsharp_widgets/__init__.py
```diff
@@ -35,3 +36,46 @@ class EstimateDetails(anywidget.AnyWidget):
     def __init__(self, estimates):
         super().__init__()
         self.estimates = estimates
+
+
+class Histogram(anywidget.AnyWidget):
```
minestarks, line 41: Does this do anything with DumpMachine/Message output? If not, I think it should just `print()` the output, just like we do in the original `run()` implementation.

widgets/src/qsharp_widgets/__init__.py
```diff
@@ -35,3 +36,46 @@ class EstimateDetails(anywidget.AnyWidget):
...
+    buckets = traitlets.Dict().tag(sync=True)
+    shot_count = traitlets.Integer().tag(sync=True)
+
+    _new_buckets = {}
+    _new_count = 0
+    _last_message = time.time()
+
+    results = []
+
+    def _update_ui(self):
```
minestarks, line 55: Is there a logic behind which names start with `_` here?

widgets/src/qsharp_widgets/__init__.py
billti: Adding as a dependency might make sense. I really don't see how you'd use qsharp_widgets without qsharp. I expect most folks will get both at once anyway via the optional extra on the qsharp package with `pip install "qsharp-lang[widgets]"`

swernli: I agree, though now that you mention the optional dependency I'm curious how pip handles what looks like a cycle there... `pip install 'qsharp-lang[widgets]'` will install qsharp-lang and then qsharp-widgets, but qsharp-widgets depends on qsharp-lang...

widgets/src/qsharp_widgets/__init__.py
billti: The intent is that eventually you'd be able to filter and navigate shots details by clicking on the histogram, same as playground does today. I really don't like the 'print by default'. If I'm running 10,000 shots and each has a handful of message and state events, I don't want 50,000 prints in the output by default.

pip/qsharp/_qsharp.py
billti: Yeah, that was my original inclination. I just didn't want to mess with run when experimenting with getting something working. I can do the consolidation.

widgets/src/qsharp_widgets/__init__.py
minestarks: Related: I think we should do the rename from `qsharp-lang` to `qsharp` this week, so I think it's okay to go ahead and put the new name, `qsharp`, as a dependency now.

widgets/src/qsharp_widgets/__init__.py
billti: It is internal, and not something intended for users  to call (see https://stackoverflow.com/a/1301369/1674945)

widgets/src/qsharp_widgets/__init__.py
minestarks: If we eventually intend to fold these into the UI then I think that's alright. (As long as we still print() in non-callback mode of course)

widgets/src/qsharp_widgets/__init__.py
minestarks: I was familiar with the "internal" connotation, but I couldn't make sense of which ones are deemed internal here and why. What about `add_result`? And why are some parameters internal and some not? Maybe I'm just not getting it, but the internal-ness doesn't seem to line up with which ones actually get used outside this class and which ones don't.

widgets/src/qsharp_widgets/__init__.py
billti: If we get multi-file in, then I wouldn't mind pushing out another dev build. Might we weird if this preview QDK starts publishing builds to the old 'qsharp' pip package now.

widgets/src/qsharp_widgets/__init__.py
billti: comp, buckets, and shot_count are part of the contract (those constructed from the 'traitlets' module are what get synced with the front-end for display). add_result should probably be internal, and result itself should probably only be access external via an API (if at all) so the user doesn't mutate it without some action. Will update for consistency.

billti:
OK. With this change 'run' has been updated to be both backwards compatible, and flexible enough for the histogram to use (both with or without capturing shot output for later details addition).

With this, you can either create a Histogram object then ask it to execute some shots:

image


Or just call 'run' as you did previously, and pass the output of that to a Histogram constructor.

image

pip/qsharp/_qsharp.py
```diff
@@ -93,23 +93,60 @@ def eval_file(path: str) -> Any:
...
+    events: List[Output]
+    result: Any
+
+
+def run(
+    entry_expr: str,
+    shots: int,
+    on_result: Optional[Callable[[ShotResult], None]] = None,
+    save_output: bool = False,
+) -> Any:
```
minestarks, line 110: Shall we require the new arguments to be named? I _think_ that gives us more flexibility to change this signature down the line. 

```suggestion
def run(
    entry_expr: str,
    shots: int,
    *,
    on_result: Optional[Callable[[ShotResult], None]] = None,
    save_output: bool = False,
) -> Any:
```

pip/qsharp/_qsharp.py
```diff
@@ -93,23 +93,60 @@ def eval_file(path: str) -> Any:
...
 
     :param entry_expr: The entry expression.
     :param shots: The number of shots to run.
+    :param on_result: A callback function that will be called with each result.
+    :param save_output: If true, the output of each shot will be saved. If false, they will be printed.
 
-    :returns values: A list of results or runtime errors.
+    :returns values: A list of results or runtime errors. If `save_output` is true,
+    a Dict is returned with a "results" property containing the list of results,
+    and an "events" property containing a list of lists of events for each result.
```
minestarks, line 123: Should `"events"` really be called `"outputs"` (my preference)? Or should `save_outputs` be called `save_events`? Events are just output events, so from a user perspective it reads like we're calling the same thing two different names.

pip/qsharp/_qsharp.py
```diff
@@ -93,23 +93,60 @@ def eval_file(path: str) -> Any:
...
+        outputs.append([])
+        run_results = get_interpreter().run(
+            entry_expr, 1, on_save_output if save_output else print_output
+        )
+        results.append(run_results[0])
+        if on_result:
+            on_result({"events": outputs[i], "result": results[i]})
+
+    if save_output:
+        return {"results": results, "events": outputs}
```
minestarks, line 147: Wouldn't it be more ergonomic to return a list where results are associated with their outputs instead of two lists, like  :

```python
    [
      { "result" : Zero, "events": ["hello world"] },
      { "result" : One, "events": ["hi", "hi there"] }
    ]
```

billti: It would, but that would also be a breaking change to the existing API. I'm not sure how many existing samples make use of the return value (I assume no customers do)

pip/qsharp/_qsharp.py
minestarks, line 1: It'd be great to add unit tests in `test_qsharp.py` for this. I wrote some that should pass with your current changes




```python

def test_run_with_result(capsys) -> None:
    qsharp.init()
    qsharp.eval('operation Foo() : Result { Message("Hello, world!"); Zero }')
    results = qsharp.run("Foo()", 3)
    assert results == [qsharp.Result.Zero, qsharp.Result.Zero, qsharp.Result.Zero]
    stdout = capsys.readouterr().out
    assert stdout == "Hello, world!\nHello, world!\nHello, world!\n"


def test_run_with_result_callback(capsys) -> None:
    def on_result(result):
        nonlocal called
        called = True
        assert result["result"] == qsharp.Result.Zero
        assert str(result["events"]) == "[Hello, world!]"

    called = False
    qsharp.init()
    qsharp.eval('operation Foo() : Result { Message("Hello, world!"); Zero }')
    results = qsharp.run("Foo()", 3, on_result=on_result, save_output=True)
    assert (
        str(results)
        == "{'results': [Zero, Zero, Zero], 'events': [[Hello, world!], [Hello, world!], [Hello, world!]]}"
    )
    stdout = capsys.readouterr().out
    assert stdout == ""
    assert called

```

```
cd pip
pip install -e .
cd tests
pytest
```

pip/qsharp/_qsharp.py
billti: Oh, I see what you mean. Leave the result without 'save_output' as it, but change this structure. I'll look into it.

minestarks (approved):

pip/qsharp/_qsharp.py
billti: I didn't like this either. In the compiler today we call them events (https://github.com/microsoft/qsharp/blob/main/npm/src/compiler/common.ts#L82), and that's what is surfaced in the wasm interface too. But I see they got called an 'Output' when the Python class was implemented.

Peronally I'd rather keep them consistent across APIs with 'events', so can make that change.

billti: > Events are just output events

There's also no guarantee this will always be true. We may at some point have tracing or profiling events we want to capture, for example.

minestarks: "events" sounds good to me, thanks

swernli: I disagree slightly, but not enough to block. The reason why the line you point to Bill uses the term "events" is that it is for messages and dump machine, not for the final result of execution. These are two different things that we are combining for the first time here.

pip/qsharp/_qsharp.py
billti: Thanks! Done.

pip/qsharp/_qsharp.py
billti: This is just for Message and DumpMachine. We save the results of execution regardless. We only save the events (Message and DumpMachine currently) in this field (and only if save_events is True).

PR #925, opened 2023-12-13
Prioritize open file buffers over files on the fs in the extension

sezna:
I have one primary question about this: is there a way to get the original source contents for an `OpenDocument` back from a `Compilation`? I could not find one, and that's why I added `latest_str_content`, to keep track of the latest `str` reported by `update_document`. It isn't the most elegant, but it might be the only option, since `Compilation` doesn't expose the original input at all.

### DIFF
--- a/language_service/src/state.rs
+++ b/language_service/src/state.rs
@@ -296,8 +296,6 @@ impl<'a> CompilationStateUpdater<'a> {
         self.publish_diagnostics();
     }
 
-    // TODO
-
     pub(super) fn close_notebook_document<'b>(
         &mut self,
         notebook_uri: &str,
--- a/language_service/src/state/tests.rs
+++ b/language_service/src/state/tests.rs
@@ -646,7 +646,7 @@ async fn update_doc_updates_project() {
         .update_document(
             "this_file.qs",
             1,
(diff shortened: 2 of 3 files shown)


language_service/src/state.rs
```diff
@@ -276,6 +296,8 @@ impl<'a> CompilationStateUpdater<'a> {
         self.publish_diagnostics();
     }
 
+    // TODO
```
minestarks, line 299: I'm intrigued...

language_service/src/state/tests.rs
```diff
@@ -627,6 +629,189 @@ fn close_notebook_clears_errors() {
...
+            SourceMap {
+                sources: [
+                    Source {
+                        name: "other_file.qs",
+                        contents: "namespace Foo { @EntryPoint() operation Main() : Unit {} }",
+                        offset: 0,
+                    },
+                    Source {
+                        name: "this_file.qs",
+                        contents: "namespace Foo {/* we should see this comment in the source*/ }",
```
minestarks, line 680: I don't know if it's intentional in the test, but looks like you're getting a parse error about the `/* */` comment syntax which Q# doesn't like

minestarks (approved):
Looking good, thanks!

PR #926, opened 2023-12-13
Projects can generate QIR

idavis:


vscode/package.json
```diff
@@ -342,6 +342,9 @@
         "extensions": [
           ".qs"
         ],
+        "filenames": [
+          "qsharp.json"
+        ],
```
minestarks, line 347: This causes qsharp.json files to be treated as Q# ... we don't want that. I'm getting Q# compiler syntax errors from the json file :)

vscode/src/projectSystem.ts
```diff
@@ -1,10 +1,17 @@
...
 // Licensed under the MIT License.
 
-import { log } from "qsharp-lang";
 import { Utils, URI } from "vscode-uri";
 import * as vscode from "vscode";
 
+import { getProjectLoader, log } from "qsharp-lang";
+
+export async function isQsharpProject(uri: vscode.Uri): Promise<boolean> {
+  // we look for a manifest file in the parent directory of the current file
```
minestarks, line 10: This comment is a bit misleading since it implies we only look in the _parent_ directory

```suggestion
  // we look for the manifest in the directory of the current file and all of its parent directories
```

vscode/src/qirGeneration.ts
```diff
@@ -22,9 +23,13 @@ export class QirGenerationError extends Error {
 export async function getQirForActiveWindow(): Promise<string> {
   let result = "";
   const editor = vscode.window.activeTextEditor;
-  if (!editor || !isQsharpDocument(editor.document)) {
+
+  if (
+    !editor ||
+    !(isQsharpDocument(editor.document) || isQsharpProject(editor.document.uri))
```
minestarks, line 29: The `isQsharpProject` check seems unnecessary now since we can't ever get to this function if we're not in a Q# file. 
(Since you had to remove the `qsharp.json` file type from the language registration)

vscode/src/qirGeneration.ts
```diff
@@ -46,19 +51,23 @@ export async function getQirForActiveWindow(): Promise<string> {
...
+        vscode.Uri.parse(source[0]),
+      );
+      if (diagnostics?.length > 0) {
+        throw new QirGenerationError(
+          "The current program contains errors that must be fixed before submitting to Azure",
+        );
+      }
+    }
+  } catch (e: any) {
+    throw new QirGenerationError(e.message);
```
minestarks, line 68: This `try/catch` should probably be directly around `loadProject` since that's the actual thing it seems to be intended to catch. Otherwise we might be catching and rethrowing the exception we just threw in line 63... which feels a bit unnecessary.

vscode/src/statusbar.ts
```diff
@@ -20,8 +21,12 @@ export function activateTargetProfileStatusBarItem(): vscode.Disposable[] {
   statusBarItem.command = "qsharp-vscode.setTargetProfile";
 
   disposables.push(
-    vscode.window.onDidChangeActiveTextEditor((editor) => {
-      if (editor && isQsharpDocument(editor.document)) {
+    vscode.window.onDidChangeActiveTextEditor(async (editor) => {
+      if (
+        editor &&
+        (isQsharpDocument(editor.document) ||
+          (await isQsharpProject(editor.document.uri)))
```
minestarks, line 28: Same comment - `isQsharpProject` should be unnecessary since we can only get here in Q# files

minestarks (approved):

PR #927, opened 2023-12-13
Error from Qubit Released in Non-Zero State Causes Leak

### LINKED ISSUE #923: Error from Qubit Released in Non-Zero State Causes Leak
**Describe the bug**

When releasing a qubit in a non-zero state, a runtime error is thrown. This error may be causing a memory leak.

**To Reproduce**

Run the following code in the playground with 100 shots specified:
```qsharp
namespace input {
    @EntryPoint()
    operation Main() : Unit {
        use q = Qubit();
        H(q);
    }
}
```
You will see the shots become slower and slower.

**Expected behavior**

Shots should be run very fast.

**Additional context**

Similar code with the same runtime error does not have the perf issue:
```qsharp
namespace input {
    @EntryPoint()
    operation Main() : Unit {
        use q = Qubit();
        X(q);
    }
}
```

```qsharp
namespace input {
    @EntryPoint()
    operation Main() : Unit {
        use q = Qubit();
        H(q);
        M(q);
    }
}
```

swernli:
Fixes #923

### DIFF
--- a/compiler/qsc/src/interpret/stateful.rs
+++ b/compiler/qsc/src/interpret/stateful.rs
@@ -264,7 +264,7 @@ impl Interpreter {
         };
 
         eval_expr(
-            &mut self.state,
+            &mut State::new(self.source_package),
             expr,
             &globals,
             &mut Env::with_empty_scope(),
--- a/vscode/src/webviewPanel.ts
+++ b/vscode/src/webviewPanel.ts
@@ -315,7 +315,7 @@ export function registerWebViewCommands(context: ExtensionContext) {
         await worker.run(code, "", parseInt(numberOfShots), evtTarget);
         clearTimeout(compilerTimeout);
       } catch (e: any) {
-        log.error("Codegen error. ", e.toString());
+        log.error("Histogram error. ", e.toString());
         throw new Error("Run failed");
       } finally {
         worker.terminate();
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -13,7 +13,7 @@ use qsc::{
         stateful::{self, re::estimate_entry},
     },
     target::Profile,
-    PackageStore, PackageType, SourceContents, SourceMap, SourceName,
+    PackageStore, PackageType, SourceContents, SourceMap, SourceName, SparseSim,
 };
 use qsc_codegen::qir_base::generate_qir;
 use serde_json::json;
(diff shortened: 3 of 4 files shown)


ScottCarda-MS (approved):

compiler/qsc/src/interpret/stateful.rs
```diff
@@ -264,7 +264,7 @@ impl Interpreter {
         };
 
         eval_expr(
-            &mut self.state,
+            &mut State::new(self.source_package),
```
cesarzc, line 267: I'm curious about why a new state is created here rather than `self.state`.

vscode/src/webviewPanel.ts
```diff
@@ -315,7 +315,7 @@ export function registerWebViewCommands(context: ExtensionContext) {
         await worker.run(code, "", parseInt(numberOfShots), evtTarget);
         clearTimeout(compilerTimeout);
       } catch (e: any) {
-        log.error("Codegen error. ", e.toString());
+        log.error("Histogram error. ", e.toString());
```
cesarzc, line 318: Was this just something you noticed but not exactly related to the fix?

wasm/src/tests.rs
```diff
@@ -248,6 +248,36 @@ fn test_run_error_program_multiple_shots() {
     .assert_eq(&output.join("\n"));
 }
 
+#[test]
+fn test_run_error_program_multiple_shots_qubit_leak() {
+    // If qubits are leaked from execution, the runtime will fail with an out of memory
+    // error pretty quickly.
```
DmitryVasilevsky, line 254: I'm a bit hesitant about this "pretty quickly" statement. What if it hangs for a long time instead trying to come up with a memory? Maybe we can do something more deterministic...

compiler/qsc/src/interpret/stateful.rs
swernli: If there is an exception, we abort execution but do not clear the "tape" used to track execution. This means it can leave leftover expressions in the state that will cause failures on subsequent runs. We fixed a variation of this bug for Python at one point, but I missed it when implementing this variation and incorrectly reused the state from `self`. This works fine as long as execution completes, since that means the tape in the state gets fully evaluated.

wasm/src/lib.rs
```diff
@@ -181,25 +181,27 @@ where
...
+            let msg = json!(
+                {"type": "Result", "success": false, "result": diag});
+            (out.event_cb)(&msg.to_string());
+            return Err(Box::new(e));
+        }
+    };
+
     for _ in 0..shots {
-        let result = interpreter.eval_entry(&mut out);
+        let result = interpreter.eval_entry_with_sim(&mut SparseSim::new(), &mut out);
```
cesarzc, line 204: Was the sparse simulator being used by default already when calling `eval_entry`?

compiler/qsc/src/interpret/stateful.rs
swernli: "Exception" is the wrong word... I mean a Q# runtime failure, like a divide by zero or an explicit `fail`, or in the sample case, a qubit released in a non-zero state.

vscode/src/webviewPanel.ts
swernli: Yes. The dialog box that pops up on histogram execution failures had the wrong strong, and it made me thing I had run the wrong command!

wasm/src/tests.rs
swernli: out of the 100 shots, I was seeing it fail around shot 24. Leaks are hard to be deterministic about... I suppose we could try peeking into the simulator state between shots to confirm it is the same?

wasm/src/lib.rs
swernli: Yes, though it was reusing the one from the stateful interpreter rather than throwing it away between shots.

compiler/qsc/src/interpret/stateful.rs
cesarzc: I see, thank you for the explanation!

cesarzc (approved):

wasm/src/tests.rs
swernli: I don't think I want to change how the test is invoked... the critical part of the change is that the undelrying API does the right thing with state between shot loops, and doing something like checking the state would require interrupting the loop to do other things in between. I'm going to leave this as-is, since it reliably reproduced the issue for us so far.

minestarks (approved):

compiler/qsc/src/interpret/stateful.rs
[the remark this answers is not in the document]
swernli: This is a good idea... I was joking with Ian that we are due for the semi-monthly evaluator/interpreter refactor ;). I think this change is something we should consider as part of a clean up after we've gotten some other things merged. For example, now that we have only the stateful interpreter, we can move it up a folder and get rid of the `stateful` module, which frees up the `stateful` name for use in some of our function names.

PR #936, opened 2023-12-13
Reset before releasing `Measure` auxiliary qubit

cesarzc:


cesarzc:
Closed in favor of #938

PR #937, opened 2023-12-13
Updating UI and tooltip for target profile selection.

### LINKED ISSUE #935: Add tooltip to status bar for base/unrestricted
https://code.visualstudio.com/api/references/vscode-api#StatusBarItem

idavis:
Closes #935

### DIFF
--- a/vscode/src/statusbar.ts
+++ b/vscode/src/statusbar.ts
@@ -66,6 +66,7 @@ export function activateTargetProfileStatusBarItem(): vscode.Disposable[] {
     const targetProfile = getTarget();
 
     statusBarItem.text = getTargetProfileUiText(targetProfile);
+    statusBarItem.tooltip = getTargetProfileUiToolTip(targetProfile);
     statusBarItem.show();
   }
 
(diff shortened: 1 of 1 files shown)


minestarks (approved):

billti:
Do we need a target specific tooltip? I assume that folks who hover on it for more info may not know what "Q# Base Profile QIR" means. I had put together just a generic/friendly tooltip regardless of selection:

```ts
    statusBarItem.tooltip = new vscode.MarkdownString(`## Q# target profile

  The target profile determines the set of operations that are available to Q#
  programs, in order to generate valid QIR for the target platform. For more
  details see <https://aka.ms/qdk.qir>.`);
```

Either is better I supposed and includes the same link, so go with either.

vscode/src/statusbar.ts
```diff
@@ -89,27 +90,40 @@ function registerTargetProfileCommand() {
...
-  { configName: "unrestricted", uiText: "Unrestricted" },
+  { configName: "base", uiText: "Q#: QIR base" },
+  { configName: "unrestricted", uiText: "Q#: unrestricted" },
 ];
 
 function getTargetProfileUiText(targetProfile?: string) {
   switch (targetProfile) {
     case "base":
-      return "QIR:Base";
+      return "Q#: QIR base";
```
billti, line 100: Out of interest, why do we repeat the string literal here, rather than use the targetProfiles const map that includes it in lines 93/94 above?

billti: (I guess that would be a little awkward its the value not the key to look up on, so never mind)

billti (approved):

PR #939, opened 2023-12-14
Fix bug in qubit reset on measure logic

swernli:
For base profile, we need to "reset" the qubit identifier for qubits that are measured such that later use of the same ID gets a fresh hardware ID. Because the underlying bug was a subtle misuse of the two types (where we have a map from `usize` to `usize`), this change adds a little bit of extra type safety by ensuring that mapped qubits are a new wrapper type `HardwareId`. Added a test case that confirms qubit reuse behaves as expected.

Thanks @cesarzc for figuring this one out!

### DIFF
--- a/compiler/qsc_codegen/src/qir_base.rs
+++ b/compiler/qsc_codegen/src/qir_base.rs
@@ -118,11 +118,14 @@ pub(super) fn get_global(
         })
 }
 
+#[derive(Copy, Clone, Default)]
+struct HardwareId(usize);
+
 pub struct BaseProfSim {
     next_meas_id: usize,
     next_qubit_id: usize,
-    next_qubit_hardware_id: usize,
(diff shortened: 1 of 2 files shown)


minestarks (approved):

minestarks:
Actually, I think I might be misunderstanding the change. I saw this as a mere refactoring and signed off. Is there a behavior change somewhere I'm missing?

swernli:
> Actually, I think I might be misunderstanding the change. I saw this as a mere refactoring and signed off. Is there a behavior change somewhere I'm missing?

Yes, this a bug fix that resulted from the refactoring because we were removing the wrong id from the map (called remove on the value instead of on the key). Using more specific types helps avoid this in the future.

This also adds a new test case that shows the effect of the fix, but it's hard to see. The added test case alone would fail without the refactoring/fix because the wrong qubit ids would be used.

PR #940, opened 2023-12-14
Support Copy-and-Update for Array with Range

### LINKED ISSUE #933: Copy-and-Update for Array with Range causes WASM Panic
**Describe the bug**

On the playground, attempting to use the Copy-and-Update syntax on an array with a range specified causes as WASM panic at runtime.

image

**To Reproduce**

Run the following code on the playground:

```qsharp
namespace input {

    @EntryPoint()
    operation Main() : Unit {
        let x = [10, 11, 12, 13];
        let y = x w/ 0..1 <- [20, 30];
        Message($"{y}");
    }
}
```

swernli:
This change adds full support for copy-update and assign-update using ranges. With this fix, patterns like the following are now correctly supported:
```qsharp
[0, 1, 2] w/ 1..2 <- [10, 11]
// yields [0, 10, 11]

[0, 1, 2, 3] w/ ...2... <- [10, 11]
// yields [10, 1, 11, 3]

mutable x = [1, 2, 3, 4];
set x w/= 2..-1..0 <- [10, 11];
// x updated in-place to [1, 11, 10, 4]
```

Fixes #933

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -1196,13 +1196,29 @@ impl State {
     fn eval_update_index(&mut self, span: Span) -> Result<(), Error> {
         let values = self.pop_val().unwrap_array();
         let update = self.pop_val();
-        let index = self.pop_val().unwrap_int();
+        let index = self.pop_val();
         let span = self.to_global_span(span);
+        match index {
+            Value::Int(index) => self.eval_update_index_single(&values, index, update, span),
+            Value::Range(start, step, end) => {
+                self.eval_update_index_range(&values, start, step, end, update, span)
+            }
+            _ => unreachable!("array should only be indexed by Int or Range"),
+        }
+    }
+
+    fn eval_update_index_single(
+        &mut self,
+        values: &[Value],
+        index: i64,
+        update: Value,
+        span: PackageSpan,
+    ) -> Result<(), Error> {
         if index < 0 {
             return Err(Error::InvalidNegativeInt(index, span));
         }
         let i = index.as_index(span)?;
-        let mut values = values.iter().cloned().collect::<Vec<_>>();
+        let mut values = values.to_vec();
         match values.get_mut(i) {
             Some(value) => {
                 *value = update;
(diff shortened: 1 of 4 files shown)


compiler/qsc_eval/src/tests.rs
```diff
@@ -2148,6 +2148,121 @@ fn update_range_to_end() {
     check_expr("", "...2..3 w/ End <- 10", &expect!["...2..10"]);
 }
 
+#[test]
```
ScottCarda-MS, line 2151: These tests are great! I wasn't expecting support for open-ended ranges, but it makes sense. Can you add a test for using the fully open range: `...`?

swernli: oh, good idea! I forgot about the fully open ended case.

compiler/qsc_eval/src/tests.rs
```diff
@@ -2148,6 +2148,121 @@ fn update_range_to_end() {
...
+        "[0, 1, 2, 3] w/ ...2... <- [10, 11]",
+        &expect!["[10, 1, 11, 3]"],
+    );
+}
+
+#[test]
+fn update_array_with_range_end() {
+    check_expr(
+        "",
+        "[0, 1, 2, 3] w/ ...2 <- [10, 11, 12, 13]",
```
ScottCarda-MS, line 2178: Oh interesting, the size of the new value array can be too big for the range. Can you add some tests for the following scenarios:
- new values array is too small for open-ended range
- new values array is too big for close-ended range
- new values array is too small for close-ended range
- new values array is empty

ScottCarda-MS (approved):

compiler/qsc_eval/src/tests.rs
swernli: Sure thing! I think some of these cases are covered here and there in these tests, but not exhaustively. I'll use your list here to double check and make sure I have individual tests for all of them.

compiler/qsc_eval/src/lib.rs
```diff
@@ -1369,19 +1418,67 @@ impl State {
...
+        match &lhs.kind {
+            &ExprKind::Var(Res::Local(node), _) => match env.get_mut(node) {
                 Some(var) if var.is_mutable() => {
-                    var.value.update_array(index, rhs);
+                    var.value.update_array(index, rhs).map_err(|idx| {
+                        Error::IndexOutOfRange(idx.try_into().expect("index should be valid"), span)
+                    })?;
+                }
+                Some(_) => {
+                    unreachable!("update of mutable variable should be disallowed by compiler")
```
minestarks, line 1439:
```suggestion
                    unreachable!("update of immutable variable should be disallowed by compiler")
```

Unless I'm misreading something?

minestarks (approved):

compiler/qsc_eval/src/lib.rs
swernli: good catch, typo from me!

PR #943, opened 2023-12-14
Run with hist now accepts projects

idavis:


wasm/src/tests.rs
```diff
@@ -310,34 +301,35 @@ fn test_runtime_error_related_spans() {
...
-        "",
+    };
+    run_internal(
+        SourceMap::new([("test.qs".into(), code.into())], None),
         |s| output.push(s.to_string()),
         1,
     )
     .expect("code should compile and run");
-    expect![[r#"{"result":{"code":"Qsc.Eval.UserFail","end_pos":1,"message":"runtime error: program failed: Cannot allocate qubit array with a negative length","related":[{"end_pos":429,"message":"explicit fail","source":"core/qir.qs","start_pos":372}],"severity":"error","start_pos":0},"success":false,"type":"Result"}"#]]
+    expect![[r#"{"result":{"code":"Qsc.Eval.UserFail","end_pos":429,"message":"runtime error: program failed: Cannot allocate qubit array with a negative length","related":[{"end_pos":429,"message":"explicit fail","source":"core/qir.qs","start_pos":372}],"severity":"error","start_pos":372},"success":false,"type":"Result"}"#]]
```
minestarks, line 333: These positions seem off to me. I think I know what to do to map source map offsets, but I need to look for a bit

wasm/src/tests.rs
```diff
@@ -226,57 +218,56 @@ fn test_run_simple_program_multiple_shots() {
...
+        SourceMap::new([("test.qs".into(), code.into())], None),
         |s| output.push(s.to_string()),
         3,
     )
     .expect("code should compile and run");
     expect![[r#"
-        {"result":{"code":"Qsc.Eval.QubitUniqueness","end_pos":1,"message":"runtime error: qubits in gate invocation are not unique","severity":"error","start_pos":0},"success":false,"type":"Result"}
-        {"result":{"code":"Qsc.Eval.QubitUniqueness","end_pos":1,"message":"runtime error: qubits in gate invocation are not unique","severity":"error","start_pos":0},"success":false,"type":"Result"}
-        {"result":{"code":"Qsc.Eval.QubitUniqueness","end_pos":1,"message":"runtime error: qubits in gate invocation are not unique","severity":"error","start_pos":0},"success":false,"type":"Result"}"#]]
+        {"result":{"code":"Qsc.Eval.QubitUniqueness","end_pos":1937,"message":"runtime error: qubits in gate invocation are not unique","severity":"error","start_pos":1920},"success":false,"type":"Result"}
```
minestarks, line 237: Offsets seem wrong here as well

wasm/src/diagnostic.rs
```diff
@@ -91,17 +88,14 @@ impl VSDiagnostic {
     /// Any labels with associated messages are included as "related information"
     /// objects in the [VSDiagnostic], whether they fall in the current document or not.
     /// Editors can display these as links to other source locations.
-    fn new<T>(labels: Vec<Label>, source_name: &str, err: &T) -> VSDiagnostic
```
minestarks, line 94: Why remove `source_name`? It was being used to get the correct offsets for the source

minestarks: If the `label` is coming from a file that is _not_ equal to `source_name` (as is the case with the broken test below), then the offsets should apply to _that_ file, not the one indicated by `source_name`. That's what this line was ensuring

wasm/src/tests.rs
minestarks, line 1: could you add a test case where we run a project with multiple files and get a runtime error? Clearly the changes have implications on error reporting here (as evidenced by the broken test) so let's make sure to cover it

wasm/src/lib.rs
```diff
@@ -197,18 +197,16 @@ where
...
+fn run_internal<F>(sources: SourceMap, event_cb: F, shots: u32) -> Result<(), Box<stateful::Error>>
 where
     F: FnMut(&str),
 {
-    let source_name = "code";
+    let source_name = sources
+        .find_by_offset(0)
+        .expect("There must be a source to process")
+        .name
+        .to_string();
```
minestarks, line 208: `find_by_offset(0)` returns the entry expression if there is one, which is probably not want you want. You'll want to get the name of the first _file_, from the original array

minestarks (approved):
If you could please add a GH issue link, we can ignore the error span issue for now (since playground isn't broken and the rest of the bug is invisible). Thank you!

PR #945, opened 2023-12-14
Compile/Design-Time Checks for Variable Scopes Don't Work Properly with Qubit Allocation Blocks

### LINKED ISSUE #932: Compile/Design-Time Checks for Variable Scopes Don't Work Properly with Qubit Allocation Blocks
**Describe the bug**

When using the qubit block allocation syntax, the bound variable is not correctly scoped by the compile/design-time checks for valid variables.

**To Reproduce**

The following code incorrectly produces no errors until runtime:
```qsharp
namespace input {
    open Microsoft.Quantum.Measurement;

    @EntryPoint()
    operation Main() : Unit {
        use q = Qubit() {
            H(q);
            MResetZ(q);
        };
        H(q);
        MResetZ(q);
    }
}
```

**Expected behavior**

The last two statements using `q` should give errors about invalid variable name `q` because `q` is out of scope at that point. The code should be treated equivalently to:

```qsharp
namespace input {
    open Microsoft.Quantum.Measurement;

    @EntryPoint()
    operation Main() : Unit {
        {
            use q = Qubit();
            H(q);
            MResetZ(q);
        };
        H(q);
        MResetZ(q);
    }
}
```

swernli:
Fixes #932

### DIFF
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -667,10 +667,22 @@ impl AstVisitor<'_> for With<'_> {
             }
             ast::StmtKind::Qubit(_, pat, init, block) => {
                 ast_visit::walk_qubit_init(self, init);
-                // The binding is valid after end of the statement.
-                self.resolver.bind_pat(pat, stmt.span.hi);
                 if let Some(block) = block {
-                    self.visit_block(block);
+                    self.with_scope(block.span, ScopeKind::Block, |visitor| {
+                        // The binding is valid after end of the statement, but scoped to the current block.
+                        visitor.resolver.bind_pat(pat, stmt.span.hi);
+
+                        for stmt in &*block.stmts {
+                            if let ast::StmtKind::Item(item) = &*stmt.kind {
+                                visitor.resolver.bind_local_item(visitor.assigner, item);
+                            }
+                        }
+
+                        ast_visit::walk_block(visitor, block);
+                    });
+                } else {
+                    // The binding is valid after end of the statement.
+                    self.resolver.bind_pat(pat, stmt.span.hi);
                 }
             }
             ast::StmtKind::Empty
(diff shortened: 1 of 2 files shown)


ScottCarda-MS (approved):

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -667,10 +667,22 @@ impl AstVisitor<'_> for With<'_> {
             }
             ast::StmtKind::Qubit(_, pat, init, block) => {
                 ast_visit::walk_qubit_init(self, init);
-                // The binding is valid after end of the statement.
-                self.resolver.bind_pat(pat, stmt.span.hi);
                 if let Some(block) = block {
-                    self.visit_block(block);
+                    self.with_scope(block.span, ScopeKind::Block, |visitor| {
```
minestarks, line 671: I think you may want to use the `with_pat` helper here. This mimics what we do for, say, for loops - this is kind of the same idea where the binding happens outside the block, but only applies to the block.

https://github.com/microsoft/qsharp/blob/b8a94e8d417dba5a8aef256103c4a14abd5937f5/compiler/qsc_frontend/src/resolve.rs#L701-L703

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -667,10 +667,22 @@ impl AstVisitor<'_> for With<'_> {
...
+                        // The binding is valid after end of the statement, but scoped to the current block.
+                        visitor.resolver.bind_pat(pat, stmt.span.hi);
```
minestarks, line 673: `block.span.lo` would be the more correct offset to use here (but if you take my suggestion to use `with_pat`, you don't need to worry about it)

minestarks (approved):

compiler/qsc_frontend/src/resolve.rs
swernli: Oh, great suggestion! I missed that I could use an existing function for that, definitely a better way to go.

compiler/qsc_frontend/src/resolve.rs
swernli: I'll use your other suggestion.

PR #946, opened 2023-12-14
Design-Time Errors for Incorrect Functor Generators

### LINKED ISSUE #914: Design-Time Errors for Incorrect Functor Generators
The user should receive immediate feedback in the form of errors for trying to match functor specializations with inappropriate generators. For example:

```qsharp
operation Foo (q : Qubit) : Unit is Adj {
    body auto;
    adjoint distribute;
}
```

should generate errors for `body` being matched with `auto` and `adjoint` being matched with `distribute`. These issues generate runtime errors, but it seems like this should be catchable at design-time.

swernli:
Fixes #914

### DIFF
--- a/compiler/qsc_passes/src/spec_gen.rs
+++ b/compiler/qsc_passes/src/spec_gen.rs
@@ -35,6 +35,30 @@ pub enum Error {
     #[error(transparent)]
     AdjGen(adj_gen::Error),
 
+    #[error("invalid specialization generator")]
+    #[diagnostic(code("Qsc.SpecGen.InvalidAdjGen"))]
+    #[diagnostic(help(
+        "valid specialization generators for adjoint are `auto`, `invert`, and `self`"
+    ))]
+    InvalidAdjGen(#[label] Span),
+
+    #[error("invalid specialization generator")]
+    #[diagnostic(code("Qsc.SpecGen.InvalidBodyGen"))]
+    #[diagnostic(help("body specialization only supports `intrinsic`"))]
+    InvalidBodyGen(#[label] Span),
+
+    #[error("invalid specialization generator")]
+    #[diagnostic(code("Qsc.SpecGen.InvalidCtlGen"))]
+    #[diagnostic(help(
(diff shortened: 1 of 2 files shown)


compiler/qsc_passes/src/spec_gen.rs
```diff
@@ -35,6 +35,30 @@ pub enum Error {
...
+    #[error("invalid specialization generator")]
+    #[diagnostic(code("Qsc.SpecGen.InvalidAdjGen"))]
+    #[diagnostic(help(
+        "valid specialization generators for adjoint are `auto`, `invert`, and `self`"
+    ))]
+    InvalidAdjGen(#[label] Span),
+
+    #[error("invalid specialization generator")]
+    #[diagnostic(code("Qsc.SpecGen.InvalidBodyGen"))]
+    #[diagnostic(help("body specialization only supports `intrinsic`"))]
```
ScottCarda-MS, line 47: Is `intrinsic` only valid for `body` or can other specializations be `intrinsic`? And do we need a check to make sure that if `body` is `intrinsic` the other specializations are either not explicitly listed or are also `intrinsic`?

swernli: Intrinsic is only valid for body, and only if there are no other specializations. There are existing test cases that cover this, which is why they aren't in the diff.

ScottCarda-MS (approved):

compiler/qsc_passes/src/spec_gen/tests.rs
```diff
@@ -49,6 +49,218 @@ fn generate_specs_body_intrinsic_should_fail() {
...
+                        hi: 71,
+                    },
+                ),
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn genrate_specs_body_self_should_fail() {
```
minestarks, line 76:
```suggestion
fn generate_specs_body_self_should_fail() {
```

minestarks (approved):

PR #949, opened 2023-12-15
Updated comparisons to make them more useful

DmitryVasilevsky:
Updated comparison operations.

* Replaced operations that inverted target qubit based on comparison result of a BigInt and LE register with operations that apply action to a target. The new operations are more general operations, and they are needed for the upcoming samples.
* Removed operations that inverted target qubit based on comparison result of two LE registers. We already have corresponding set of more general operations that would apply actions.
* Previously all comparisons except for equality comparisons allowed singly controlled actions. Actions had to support multiqubit control to be passed into equality comparisons. This restriction has been lifted. Actions that only support single-qubit control can now be passed to equality comparisons.

NOTE that the former more specific functions didn't implement any optimizations based on the fact that the action was an X gate. Also, to use the old functionality one needs to pass an X gate as the first argument.

### DIFF
--- a/library/std/unstable_arithmetic.qs
+++ b/library/std/unstable_arithmetic.qs
@@ -569,7 +569,7 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
             within {
                 ApplyPauliFromBitString(PauliX, false, bits, xs);
             } apply {
-                Controlled action(xs, target);
+                Controlled ApplyAsSinglyControlled(xs, (a => action(a), target));
             }
         }
     }
@@ -647,7 +647,7 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
                 X(y[i]);
             }
         } apply {
-            Controlled action(y, target);
+            Controlled ApplyAsSinglyControlled(y, (a => action(a), target))
         }
     }
 
--- a/library/std/unstable_arithmetic.qs
+++ b/library/std/unstable_arithmetic.qs
@@ -512,7 +512,7 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
     }
 
     //
-    // Compariosons
+    // Comparisons
     //
     // Compare BigInt and qubit register in a little-endian format and apply action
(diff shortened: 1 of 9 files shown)


library/std/unstable_arithmetic.qs
```diff
@@ -512,140 +512,92 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
     }
 
     //
-    // Compare BigInt and qubit register in a little-endian format
+    // Compariosons
```
ScottCarda-MS, line 515:
```suggestion
    // Comparisons
```

library/std/unstable_arithmetic.qs
```diff
@@ -512,140 +512,92 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
...
+
         let cBitSize = BitSizeL(c);
         let xLen = Length(xs);
         if (cBitSize <= xLen) {
             let bits = BigIntAsBoolArray(c, Length(xs));
             within {
                 ApplyPauliFromBitString(PauliX, false, bits, xs);
             } apply {
-                Controlled X(xs, target);
+                Controlled action(xs, target);
```
msoeken, line 572: I suggest to use AndChain here and only singly-controlled the action.

msoeken (approved):

library/std/unstable_arithmetic.qs
DmitryVasilevsky: I would prefer not to do this, at least for now. I can see that all other (not equality) comparisons result in a singly controlled action. But if I add the And chain here, user won't be able to avoid it. On the other hand, if I do not add it here, user will be able to provide it if needed.
I'm even thinking we could provide a generic operation that would turn controllable operation into a singly-controlled operation.

DmitryVasilevsky: I gave it a try and I updated code to apply AND chain. It is faster in simulation and should be better for resource estimation.

swernli (approved):

PR #951, opened 2023-12-15
Report when qsharp.json can't be parsed

minestarks:
The error is surfaced when we try to parse the Q# manifest for any reason (running the program, typing in Q# files etc).

image


vscode/qsharp.schema.json
minestarks, line 1: This has nothing to do with reporting errors, but just provides completion help in the editor. I thought it would be nice to have.

PR #955, opened 2023-12-22
Add performance benchmarks to CI

sezna:
goals:
- [x] as non-flakey as possible
    -  Accomplished by not failing the stage based on any threshold
- [x] provide high-level insights to flag potential degradation, but don't block PRs

This PR adds a CI stage to benchmark a feature branch against main. It only works on criterion benches, so we are using it in `compiler/qsc`. I believe this is a good starting point, as that encapsulates all compiler logic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,4 +162,4 @@ jobs:
       - uses: boa-dev/criterion-compare-action@v3
         with:
           branchName: ${{ github.base_ref }}
-          cwd: "qsc"
+          cwd: "compiler/qsc"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,3 +162,4 @@ jobs:
       - uses: boa-dev/criterion-compare-action@v3
         with:
           branchName: ${{ github.base_ref }}
+          cwd: "qsc"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,22 +55,6 @@ jobs:
       - name: Clippy Lints
         run: cargo clippy --all-targets --all-features -- -D warnings
 
-  benches:
-    name: Benches
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          submodules: "true"
-      - name: Setup rust toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -153,3 +153,13 @@ jobs:
...
         run: python ./build.py --no-check --no-test --wasm --npm --vscode --integration-tests
+
+  runBenchmark:
+    name: run benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: boa-dev/criterion-compare-action@v3
+        with:
+          branchName: ${{ github.base_ref }}
```
minestarks, line 164: Do we want to limit this to run on pull requests only? (Or, what exactly does this do when it runs on a merge group check or manually invoking a workflow?)

Do we want to combine this with the "benches" job?

swernli: Looking at how the output is summarized and what details are available in the logs, I think this will supersede the existing benches run, so I would be in favor of this PR including a removal of the existing one.

swernli:
> ## Benchmark for [b0d8aa4](https://github.com/microsoft/qsharp/commit/b0d8aa496538c7b48e0958c5ea894726503052d2)
> Click to view benchmark

I think this benchmark run is the perfect example of the benefits and detriments of using the default benchmark settings with the benches we have checked in. Here we have what looks like a 3% regression in standard library compilation with no changes to the source code and run back-to-back on the same machine. That this infra makes it so easy to see this is exactly what I was hoping for from adding more perf logic to the pipeline, so I think getting this checked in is the right first step for improving our benchmark visibility on each PR. Then as follow up, we can try changing benchmark settings (like increase the number of samples) to see if that reduces noise.

Good find, @sezna!

swernli (approved):

.github/workflows/ci.yml
```diff
@@ -153,3 +137,13 @@ jobs:
...
+          cwd: "compiler/qsc"
```
swernli, line 149: I think you can add a condition here to make it play nice with merge queues:


```suggestion
          cwd: "compiler/qsc"
        if: ${{ github.base_ref != null }}
```

PR #956, opened 2023-12-27
Add artifact validation for CI builds

### LINKED ISSUE #950: Include Python down level validation in CI
We've hit a few times where Python code uses newer 3.10 features that then cause failures in supported down-level versions we only catch during the release pipeline. We should add some CI validation of version features so that we don't end up getting surprised.

idavis:
Python 3.8 is excluded as the build can't be run from 3.8 but we allow 3.8 as a target Python version.

Build is moved to ubuntu-20.04 for binary compat generation. This caused issues with the rust caching action as it wasn't separating by target os, so we were getting symbos issues with glibc versions.

This adds ~10s to the main build for packaging and uploading artifacts. The secondary (non-required) build for checks is ~20s-120s.

Fixes #950

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,52 @@ jobs:
         run: python ./prereqs.py --install
       - name: Build and Test
         run: python ./build.py --no-check
-
+      # Now we need to prepare the artifacts for upload.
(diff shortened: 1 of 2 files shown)


idavis:
Closing. We can add this if desired in the future.

PR #966, opened 2024-01-02
Diagnostics event for closed document is not really an error

minestarks:
Just changing the severity of the log since we can actually hit this in practice.

Repro:
1. Create a new untitled file.
2. Change the language to Q#.
3. Type some bad Q# into the file to make errors show up.
4. Save the file to disk as `foo.qs`.
5. Observe the Output window for an error log.

The reason for this error is that the rename gets handled as a pair of events: `update_document`, under the new name, followed by `close_document` under the old name. The first event causes all the diagnostics to get republished, including the ones associated with the old name (this is where we log the error). The second event clears the diagnostics associated with the old name. Then all is good.

### DIFF
--- a/npm/src/language-service/language-service.ts
+++ b/npm/src/language-service/language-service.ts
@@ -440,11 +440,18 @@ export class QSharpLanguageService implements ILanguageService {
       if (code === null && !empty) {
         // We need the contents of the document to convert error offsets to utf16.
         // But the contents aren't available after a document is closed.
-        // It is possible to get a diagnostics event after a document is closed,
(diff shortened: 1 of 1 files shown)


minestarks:
> Huh, good catch. This log has helped us track down other errors in the past, though, as it is often a symptom of messing up the event logic. I suppose we can still keep track of the info log, but perhaps this is an appropriate place to use a warning?

I got a little shy making it a warning because I thought it may happen a lot, and I don't want other devs thinking something wrong is happening. Maybe I'll try it out for a little bit, and I'll make it a warning if it doesn't happen in regular usage (like when just opening and closing files)

PR #969, opened 2024-01-03
Delete compilation from LS if last document in project is closed

sezna:
Also, some `Debug` implementations that were useful when debugging

[1] The question of whether or not to unit test private APIs is hotly contested. I personally find it useful, so I did it. Happy to discuss the tradeoffs there with anybody.

### DIFF
--- a/language_service/src/state.rs
+++ b/language_service/src/state.rs
@@ -24,7 +24,7 @@ type PinnedFuture<T> = Pin<Box<dyn Future<Output = T>>>;
 /// `'a` represents the lifetime of the contained `dyn Fn`.
 type AsyncFunction<'a, Arg, Return> = Box<dyn Fn(Arg) -> PinnedFuture<Return> + 'a>;
 
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub(super) struct CompilationState {
     /// A `CompilationUri` is an identifier for a unique compilation.
     /// It is NOT required to be a uri that represents an actual document.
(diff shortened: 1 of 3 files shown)


compiler/qsc_frontend/src/compile.rs
```diff
@@ -210,6 +210,12 @@ pub struct PackageStore {
     next_id: PackageId,
 }
 
+impl Debug for PackageStore {
```
minestarks, line 213: Can we impl `Display` instead? It feels unconventional to have the `Debug` emit a "lossy" representation - I would expect it to display every bit of data in a struct. But it's entirely possible I'm making this up since I can't really find any documentation to back up my opinion...

minestarks (approved):

minestarks:
Didn't realize this was still draft but overall approach LGTM

compiler/qsc_frontend/src/compile.rs
minestarks: I'm with you on `Display` but we already break that rule all over the place for HIR/AST types - the `Display` impls we have are more of a developer-facing representation, so [we ended up adding another `display()` method](https://github.com/microsoft/qsharp/blob/09ded186cf91547ac1c0701d287b5c6b7fe8102e/compiler/qsc_hir/src/ty.rs#L68) that _is_ for actual users . It's already a bit messy, I'm okay whichever way you choose to go on this.

minestarks (approved):

PR #971, opened 2024-01-03
Use empty string for signature help when no documentation present

### LINKED ISSUE #913: Signature Help Causes Language Server Crash on Playground: Cannont read properties of undefined (reading 'length')
**Describe the bug**

image

This error occurs when receiving signature help at either end of a callable's parameter list.

**To Reproduce**

```qsharp
namespace input {
    function Foo() : Unit {}

    @EntryPoint()
    operation Main() : Unit {
        Foo();
    }
```
Put your cursor in the parentheses of the `Foo();` call and activate signature help by using the keyboard shortcut `ctrl-shift-space`. This should trigger the error above.

swernli:
VSCode/Monaco expect the documentation member of `SignatureInformation` and `ParamterInformation` to always be defined in order to check the length (see https://github.com/microsoft/vscode/blob/eb55c74a90679ba9471ea942b7cb8e6df2b24fad/src/vs/editor/contrib/parameterHints/browser/parameterHintsWidget.ts#L285-L296). As a result, we should not be serializing `None` values as undefined, but rather convert those to the empty string before sending them as part of the serialized `ISignatureHelp` type.

Fixes #913

### DIFF
--- a/wasm/src/language_service.rs
+++ b/wasm/src/language_service.rs
@@ -190,7 +190,7 @@ impl LanguageService {
                     .into_iter()
                     .map(|sig| SignatureInformation {
                         label: sig.label,
-                        documentation: sig.documentation,
+                        documentation: sig.documentation.unwrap_or_default(),
                         parameters: sig
                             .parameters
                             .into_iter()
@@ -199,7 +199,7 @@ impl LanguageService {
                                     start: param.label.start,
                                     end: param.label.end,
                                 },
-                                documentation: param.documentation,
+                                documentation: param.documentation.unwrap_or_default(),
                             })
                             .collect(),
                     })
@@ -347,12 +347,12 @@ serializable_type! {
     SignatureInformation,
     {
         label: String,
-        documentation: Option<String>,
+        documentation: String,
         parameters: Vec<ParameterInformation>,
     },
     r#"export interface ISignatureInformation {
         label: string;
-        documentation?: string;
+        documentation: string;
         parameters: IParameterInformation[];
     }"#
 }
@@ -361,11 +361,11 @@ serializable_type! {
     ParameterInformation,
     {
         label: Span,
-        documentation: Option<String>,
+        documentation: String,
     },
     r#"export interface IParameterInformation {
         label: ISpan;
-        documentation?: string;
+        documentation: string;
     }"#
 }
 


minestarks:
I can't quite wrap my head around this fix, but I'm assuming you've verified that it fixes the issue, and it's straightforward enough.

The part that puzzles me is that this expression would short circuit if `signature.documentation` is `undefined`, so I don't think it's what's throwing the exception.

```js
if (signature.documentation && typeof signature.documentation === 'string' && assertIsDefined(signature.documentation).length > 0) {
```

So it's probably this line that's throwing, and for some reason we're returning `documentation` which is an `object` , and it's actually `signature.documentation.value` that's `undefined`?

```js
signature.documentation && typeof signature.documentation === 'object' && assertIsDefined(signature.documentation.value).length > 0
```

I don't really see how this is happening, something mysterious about serde serialization maybe? But your fix certainly sidesteps the issue by making it a string. So I guess I won't worry about it :)

minestarks (approved):

swernli:
> So it's probably this line that's throwing, and for some reason we're returning `documentation` which is an `object` , and it's actually `signature.documentation.value` that's `undefined`?
> 
> ```js
> signature.documentation && typeof signature.documentation === 'object' && assertIsDefined(signature.documentation.value).length > 0
> ```
> 
> I don't really see how this is happening, something mysterious about serde serialization maybe? But your fix certainly sidesteps the issue by making it a string. So I guess I won't worry about it :)

Yup, you are right here, it's actually `signature.documentation.value` that is undefined. I was able to see this in the developer tools for the web by breaking in on the exception or setting a breakpoint on the offending function:
image
I think this is indeed because serde serialization turns a Rust `None` into a JS Object with an undefined value, and it breaks the logic here. At first, I thought of filing an issue on VSCode and suggesting the fix you have above, but it that would still throw (see the definition of `assertIsDefined` [here](https://github.com/microsoft/vscode/blob/d8a75eacf7c70e7327a460a8224dec14a65e1e27/src/vs/base/common/types.ts#L97-L103)). Looking at the VSCode logic again with the assertion in mind, it's really saying that documentation is either a string or it's an object whose `.value` must be a string. So, while the VSCode logic could be more resilient, by using `Option<String>` in Rust and relying on serde we are allowing types through that VSCode does not support.

I did confirm via the browser debugger that the change makes the exception go away, so that's one less noisy exception for us to worry about!

PR #974, opened 2024-01-05
Format compilation errors from project import in Python

swernli:
Format compilation errors when failing to compile project files provided to `qsharp.init()`.

Before:
image

After:
image

### DIFF
--- a/pip/src/fs.rs
+++ b/pip/src/fs.rs
@@ -35,20 +35,20 @@ struct FsHooks {
 
 #[derive(Debug)]
 struct Entry {
-    entry_type: EntryType,
+    ty: EntryType,
     path: String,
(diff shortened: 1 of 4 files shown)


pip/src/fs.rs
```diff
@@ -35,20 +35,20 @@ struct FsHooks {
 
 #[derive(Debug)]
 struct Entry {
-    entry_type: EntryType,
+    ty: EntryType,
```
swernli, line 38: These name changes are just me responding to clippy 1.75. There are other new clippy warnings-as-errors in the repo but we'll get to those in a separate PR.

minestarks (approved):
Thanks for catching this and even adding a test!

PR #975, opened 2024-01-05
Use src dir for Q# projects

sezna:
This PR updates `qsc_project` logic so Q# projects:
- Default to `/src/**/*.qs` (instead of all sibling files)
- Ignore all dot entries 
- Don't support exclude files and regexes

Aside: This PR nicely proves that our project logic is indeed centralized. We only need to update `qsc_project`, and we get project

### DIFF
--- a/compiler/qsc_project/tests/harness.rs
+++ b/compiler/qsc_project/tests/harness.rs
@@ -15,7 +15,7 @@ pub fn check(project_path: PathBuf, expect: &Expect) {
         .unwrap()
         .unwrap();
     let fs = StdFs;
-    let mut project = fs.load_project(dbg!(&manifest)).unwrap();
+    let mut project = fs.load_project(&manifest).unwrap();
 
     // remove the prefix absolute path
     for (path, _contents) in project.sources.iter_mut() {
--- a/vscode/src/projectSystem.ts
+++ b/vscode/src/projectSystem.ts
@@ -20,6 +20,19 @@ export async function getManifest(uri: string): Promise<{
     return null;
   }
 
+  try {
+    JSON.parse(manifestDocument.content);
+  } catch (e) {
+    log.warn(
+      `failed to parse manifest at ${manifestDocument.uri.toString()}`,
+      e,
+    );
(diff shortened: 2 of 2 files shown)


compiler/qsc_project/src/project.rs
```diff
@@ -61,34 +60,36 @@ pub trait FileSystemAsync {
...
         let listing = self.list_directory(initial_path).await?;
         let mut files = vec![];
-        for item in listing {
-            let file_name = item.entry_name();
-            let name = item.path().to_string_lossy().to_string();
-            if regex_matches(exclude_patterns, &name) || exclude_files.contains(&file_name) {
-                continue;
+        for item in filter_hidden_files(listing.into_iter()) {
+            match item.entry_type() {
+                Ok(EntryType::File) if item.entry_extension() == "qs" => files.push(item),
```
swernli, line 72: Doesn't this line mean that .qs files that are peers to the src folder will also be included? I think this pattern should be updated to match non-async version below which only begins gathering files once it has found the src folder.

minestarks:
What's supposed to happen in this setup? This doesn't look quite right to me.

image

swernli (approved):

swernli (approved):

swernli:
Good catches with the behavior of searching up for the containing folder. I ran this through some tests manually on my end, and the behavior of the extension looks good to me.

compiler/qsc_project/tests/harness.rs
```diff
@@ -15,7 +15,7 @@ pub fn check(project_path: PathBuf, expect: &Expect) {
         .unwrap()
         .unwrap();
     let fs = StdFs;
-    let mut project = fs.load_project(&manifest).unwrap();
+    let mut project = fs.load_project(dbg!(&manifest)).unwrap();
```
minestarks, line 18: Did you mean to leave this in?

minestarks (changes requested):
I think the corrupt qsharp.json error reporting in vscode may have been unintentionally changed, the rest LGTM

minestarks (approved):

PR #977, opened 2024-01-05
Updates for latest bug fixes and feature tweaks

swernli:


minestarks (approved):

library/std/re.qs
```diff
@@ -127,4 +127,59 @@ namespace Microsoft.Quantum.ResourceEstimation {
...
+    /// Assumed number of repetitions, factor to multiply the cost with
+    operation BeginRepeatEstimates(count : Int) : Unit {
+        body ... {
+            BeginRepeatEstimatesInternal(count);
+        }
+        adjoint self;
+    }
+
+    internal operation BeginRepeatEstimatesInternal(count : Int) : Unit {
+        body intrinsic;
```
DmitryVasilevsky, line 150: Do we need internal operations to provide adjoint of the public ones?

DmitryVasilevsky (approved):
Approving the library part.

library/std/re.qs
swernli: Only if they are intrinsic. Callables that are body intrinsic can't also be adjoint.

PR #978, opened 2024-01-06
Fix: @EntryPoint completion inserts two @ characters

### LINKED ISSUE #812: `@EntryPoint` completion inserts two `@` characters
In the Q# Playground or in VS Code, typing `@` in a namespace correctly brings up the `@EntryPoint()` attribute, but committing the completion results in `@@EntryPoint()`.

The insertion range probably needs to consume the `@` before the cursor, which is something I missed initially.

Seen as of a153981617a6c8a2bb745a8d15176755d399af44

minestarks:
Fixes #812

Good news: This was a one line fix
Bad news: That one line was a regex

### DIFF
--- a/vscode/language-configuration.json
+++ b/vscode/language-configuration.json
@@ -22,7 +22,7 @@
     ["{", "}"],
     ["\"", "\""]
   ],
-  // This is the sample word pattern from https://code.visualstudio.com/api/language-extensions/language-configuration-guide
+  // This is the default word pattern from https://code.visualstudio.com/api/language-extensions/language-configuration-guide
   // with the following modifications:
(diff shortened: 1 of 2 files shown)


vscode/language-configuration.json
```diff
@@ -12,12 +12,29 @@
   "autoClosingPairs": [
     ["[", "]"],
     ["(", ")"],
-    ["{", "}"]
+    ["{", "}"],
+    { "open": "\"", "close": "\"", "notIn": ["string"] }
```
minestarks, line 16: Unrelated fix: Noticed we were missing quotes in the auto-close list.

ScottCarda-MS (approved):

PR #979, opened 2024-01-07
Remove division by zero check for floating point

### LINKED ISSUE #870: Behavior of division by 0.0 (for `Double`)
**Describe the bug**

This just describes a non-documented inconsistency with the previous QDK. It seems that division by 0.0 returned `NaN` in the previous QDK (see, e.g., [implementation of `NaN()`](https://github.com/microsoft/qsharp-runtime/blob/3af588c549aa7d3d6d512e00e61ab697597b2fed/src/Simulation/QSharpFoundation/Math/FloatingPoint.qs#L22)). The current implementation results in a simulation error and program interruption.

I suggest to have a discussion on whether to change simulation behavior or adding the change to the [simulation section](https://github.com/microsoft/qsharp/wiki/Differences-from-the-previous-QDK#simulation
) of the wiki.

**To Reproduce**

[Q# playground](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAEz1Oy2rDQAy871cMcQ%2F2JZtzS6GHhp4MCUk%2FQLWFI%2FA%2B2JVTQvC%2Fd%2B2EDkKgmdFI1lpUOJGLIxtbhi9WFT8gKyXlfuFWvsIn5y5JVAl%2BZc4XyShFcOLF0YhjhZjCkMhBL6ToyOOHMWXuoeERid8k64Fi7kLPW%2BPJcY7UMdrbcSKvkzs8U%2B7GoOBj7zXdDkG81s3KhMiJlk%2FQkvi6wSu%2BvWhZwBMjK654x267g13627%2FUcs40cP2yWQz367xpHuJsZvMHsuIAuxEBAAA%3D)

**Expected behavior**

I don't suggest an expected behavior, but note the inconsistency.

swernli:
This change removes division by zero checks for floating point values and returns to previous behavior for Q#. Specifically:
 - 0.0 / 0.0 produces NaN
 - 1.0 / 0.0 produces inf
 - -1.0 / 0.0 produces -inf

To work with infinite values, this also reintroduces `Microsoft.Quantum.Math.IsInfinite` as a counterpart to `IsNaN` in the same namespace.

Fixes #870

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -1731,11 +1731,7 @@ fn eval_binop_div(lhs_val: Value, rhs_val: Value, rhs_span: PackageSpan) -> Resu
         }
         Value::Double(val) => {
             let rhs = rhs_val.unwrap_double();
-            if rhs == 0.0 {
-                Err(Error::DivZero(rhs_span))
-            } else {
-                Ok(Value::Double(val / rhs))
-            }
+            Ok(Value::Double(val / rhs))
         }
         _ => panic!("value should support div"),
     }
--- a/compiler/qsc_eval/src/tests.rs
+++ b/compiler/qsc_eval/src/tests.rs
@@ -614,27 +614,18 @@ fn binop_div_double() {
 }
 
 #[test]
-fn binop_div_double_zero() {
-    check_expr(
-        "",
-        "1.2 / 0.0",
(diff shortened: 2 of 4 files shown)


minestarks (approved):
Huh, I had no idea this was standard for floating point.

DmitryVasilevsky (approved):

PR #981, opened 2024-01-08
Add CI stage for memory benchmark

sezna:
This PR adds the CI stage for #976, which compares `main`'s result to the feature branch's result. This CI will not pass until #976 goes in.

Eventually, we should probably write our own Github Action to combine the benchmarks from #955 and these memory benchmarks into one PR comment. That'll reduce the noisy github-actions bot comments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,7 +181,7 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: `# Memory Report for ${{ github.sha }}
+              body: `## Memory Report for ${{ github.sha }}
             | Test                        | This Branch | On Main  |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -148,3 +148,41 @@ jobs:
...
+          echo "${{env.BRANCH_MEASUREMENT}}"
+          echo $BRANCH_MEASUREMENT
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `# Memory Report for ${{ github.sha }}
```
swernli, line 184: nitpick: can this be `##` so that it's not as big in the comment?

swernli (approved):

Issue #941, opened 2023-12-14
Cannot document UDT entries


msoeken:
**Describe the bug**

As of recently, I was able to document UDT entries. This doesn't work anymore in current `main`.

**To Reproduce**

https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAE4WQT0vEMBDF7%2FkUD3ppQS3uURE8%2BAcPC7uoH2Bsx25gMwnJlKXIfneTVMSbw1zmN%2FNeHun7Hg1eyYUjmz4Pz6xqZUJSispjYZU3eOA0RBvUeqnk7WATchOcFevoiH2DEP0UyUEPpBhI8MGYE49Qv1riFG19IB8PfuQrI%2BQ4BRoY22U%2Fk%2Bjsdj8uX8Ygl%2FBJl8B48h53aCsrVVKMfrhMGovjp4%2BgOF3%2F7suAG7yIXvyj2fzVbFZNRd3tGuH%2BUTQuO29F264SHzhS%2BQxsyUrbZdG7WM2Zy%2FZszuYbAb3J3VwBAAA%3D

**Expected behavior**

No error.

sezna:
Closing as this was resolved in #982

PR #982, opened 2024-01-08
Anticipate invalid docstrings in some locations and don't generate a parse error.

sezna:
See #941 for more context -- we don't want the modern compiler to reject previously accepted Q# code.

### DIFF
--- a/compiler/qsc_parse/src/item.rs
+++ b/compiler/qsc_parse/src/item.rs
@@ -209,7 +209,6 @@ fn parse_open(s: &mut Scanner) -> Result<Box<ItemKind>> {
 }
 
 fn parse_newtype(s: &mut Scanner) -> Result<Box<ItemKind>> {
-    throw_away_doc(s);
     token(s, TokenKind::Keyword(Keyword::Newtype))?;
     let name = ident(s)?;
     token(s, TokenKind::Eq)?;
--- a/compiler/qsc_parse/src/item/tests.rs
+++ b/compiler/qsc_parse/src/item/tests.rs
@@ -214,20 +214,20 @@ fn callable_return_doc() {
     check(
         parse,
         "
-        operation Foo(input: Int): 
+        operation Foo(input: Int):
         /// the return type
         Unit {}
         ",
         &expect![[r#"
-            Item _id_ [9-80]:
-                Callable _id_ [9-80] (Operation):
+            Item _id_ [9-79]:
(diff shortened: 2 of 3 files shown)


compiler/qsc_ast/src/ast.rs
```diff
@@ -317,14 +317,27 @@ impl Display for Attr {
...
+        TyDef {
+            id: NodeId::default(),
+            span: Span::default(),
+            kind: Box::default(),
+            doc: Rc::from(""),
+        }
+    }
 }
 
 impl Display for TyDef {
```
swernli, line 343: The display for `TyDef` should be updated here, which will make the tests actually show the parsed doc-comments.

swernli:
Hmm... at the high level, I'm not sure we want this. As it stands, any `///` comment is a doc comment, and those can only appear on namespaces and items (callables and newtypes). A user can still put a standard `//` comment on individual items which are just for those reading the code. But a doc comment implies that this will work with our language service to display the values when the user hovers. Now we have doc strings that don't get used anywhere in our system, which feels odd.

That said, this would be the first step to fully supporting comments on individual UDT properties throughout the stack and a second PR could add logic for showing these in the language service. I just think we need to confirm that we are going to do that second part before we take on this first part, since I wouldn't want to add new data to the AST that we then never end up using and just accrues cost there.

minestarks:
> Hmm... at the high level, I'm not sure we want this. As it stands, any `///` comment is a doc comment, and those can only appear on namespaces and items (callables and newtypes). A user can still put a standard `//` comment on individual items which are just for those reading the code. But a doc comment implies that this will work with our language service to display the values when the user hovers. Now we have doc strings that don't get used anywhere in our system, which feels odd.
> 
> That said, this would be the first step to fully supporting comments on individual UDT properties throughout the stack and a second PR could add logic for showing these in the language service. I just think we need to confirm that we are going to do that second part before we take on this first part, since I wouldn't want to add new data to the AST that we then never end up using and just accrues cost there.

Our language documentation claims that there should be a "named items" section that documents each individual "field" (which is what a "named item" is I suppose?)  . We could follow that.

https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/1_ProgramStructure/7_Comments.md

Sure this is supportable in the language service, just a bit more work.

billti (approved):

swernli (approved):

PR #985, opened 2024-01-08
scatter chart for RE

ivanbasov:


msoeken (approved):

ivanbasov:
@billti and @swernli , could you please take a look?

I have recently added Jupyter notebook support and removed a reference to an external colormap package. A Jupyter notebook sample is also there.

Besides careful testing, the only item to add seems to be unit tests for ticks.tsx. Hope to proceed with them soon.

widgets/js/index.tsx
```diff
@@ -20,18 +34,25 @@ export function mdRenderer(input: string) {
...
 type RenderArgs = {
   model: AnyModel;
   el: HTMLElement;
 };
 
 export function render({ model, el }: RenderArgs) {
+  console.error("render");
   const componentType = model.get("comp");
+  console.log("componentType");
+  console.error("Some error");
```
billti, line 47: Remove these console.error logs.

npm/ux/colormap.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    // map index points from 0..1 to 0..n-1
+    const indicies = cmap.map(function (c) {
+      return Math.round(c.index * colorsCount);
+    });
+
+    const steps = cmap.map((c, i) => {
+      const rgba = cmap[i].rgb.slice();
+
+      // if user supplies their own map use it
```
billti, line 22: How does anyone supply their own map?

npm/ux/data.tsx
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// Data structures and functions for the scatter chart and the results table.
```
billti, line 3: Add a newline between the license header and any other comments.

npm/ux/generate_report_code.py
```diff
@@ -0,0 +1,227 @@
+import re
```
billti, line 1: I'm not clear what uses this file? Does any other code use it? If it's something that needs to be run manually, how? It could do with some notes/instructions.

billti: (And a license header)

widgets/js/index.tsx
```diff
@@ -66,6 +87,58 @@ function renderChart({ model, el }: RenderArgs) {
   model.on("change:estimates", onChange);
 }
 
+function renderScatterChart({ model, el }: RenderArgs) {
+  const onChange = () => {
+    const results = model.get("estimates");
+
+    const estimates = [];
```
billti, line 94: Ideally the render call in the widget should just render the component - similar to those above. Does this next 20 lines of logic belong here or in the component? Do we need a new component to wrap it up this logic and the two child components maybe? (Just thinking here - not proposing either of these yet).

npm/ux/index.ts
```diff
@@ -5,7 +5,25 @@
...
 import "./qsharp-ux.css";
 
+export { ColorMap } from "./colormap.js";
+export {
+  ColumnNames,
+  CreateReData,
+  InitialColumns,
+  ReDataToRow,
+  ReDataToRowScatter,
+  xAxis,
```
billti, line 15: Do all of these need to be exported at the top level of the package? Seems like some of these should be internal to the components. (e.g. does a consumer need to worry about xAxis and yAxis?)

npm/ux/colormap.tsx
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// essentially borrowed from https://www.npmjs.com/package/colormap
+// source are also available at https://github.com/bpostlethwaite/colormap
```
minestarks, line 5: Don't we need license notices if we're copying code?

widgets/js/index.tsx
```diff
@@ -66,6 +87,58 @@ function renderChart({ model, el }: RenderArgs) {
...
+      estimates.push(results);
+    } else {
+      for (const estimate of Object.values(results)) {
+        estimates.push(estimate);
+      }
+    }
+
+    (estimates as Array<any>).forEach(
+      (item, idx) =>
+        (item.jobParams.runName = `(${String.fromCharCode(0x61 + idx)})`),
```
minestarks, line 105: Wouldn't this start to look a bit weird after `idx` exceeds 26? Is it important to use letters or can we use a number suffix?

ivanbasov: Good question! Thank you, @minestarks ! On one hand, the chart and the table could show many dozens of series and rows correspondingly. We even prepare colormap for the purpose. However, it is pretty hard to analyze the chart and the summary table with the number of series/rows greater than 10-15, ok 20. So, technically we could replace (a), (b), (c) with (1), (2), (3). But do we want to support such scenarios? Or maybe we could show (a), (b), (c) if the number of series/rows is less than 27, and numbers otherwise.

We have the same code on VS Code extension side (this one is a copy for Jupyter widgets).
@billti and @minestarks , what do you think?

npm/ux/colormap.tsx
ivanbasov: Do we need to support custom maps?
1. Map definitions can be provided as arrays for colors with 'distances' between them to set interpolations. It seems to be a complication for the API.
2. Supporting custom maps also requires some kind of error handling for situations such as the number of map intervals is greater than the number of colors need to produce.
3. Most of maps provided in colormap original samples provide shades of few number of gradients. It seems to be not applicable for the scatter chart purposes.

I do not think that providing custom color maps would be beneficial for customers.
Maybe we could supply custom colors for individual series at the Jupyter notebook scenario? E.g. draw majorana with '#00FF00' and gate-based with '#FF0000'.

What do you think @billti?

ivanbasov: Added support of custom colors for the Jupyter notebook scenario. Hopefully, this provides enough flexibility for users.

widgets/js/index.tsx
ivanbasov: Here are my thoughts:
1. for VS Code scenario labels like "BitFlip (a)", "BitFlip (b)" are enough because we would consider a few items in the run (now just 8) and add "-2", "-3" for next runs.
2. For Jupyter notebook scenario, it is worth to have numbers (not letters for easiest correspondence with params sent for the estimation). I have added this starting with 1. Please let me know if you think we should start with 0.
3. For Jupyter scenario, it could be helpful to show custom labels for runNames. I have added this as well and provided in a sample.

npm/ux/report.tsx
```diff
@@ -0,0 +1,814 @@
...
+
+  return r;
+}
+
+// THIS CODE HAS BEEN AUTOMATICALLY GENERATED WITH generate_report_code.py from output_data.md
+export function CreateReport(result: ReData): ReportData {
+  const groups = [] as ReportGroup[];
+  let entries = [] as ReportEntry[];
+
+  const map = createPathMap(result);
```
ivanbasov, line 239: still debugging here

resource_estimator/src/estimates/data/job_params.rs
```diff
@@ -190,6 +190,6 @@ pub enum EstimateType {
 
 impl Default for EstimateType {
     fn default() -> Self {
-        Self::SinglePoint
+        Self::Frontier
```
ivanbasov, line 193: @msoeken , do you agree it is a good time to make it the default option?

pip/tests/test_re.py
```diff
@@ -100,19 +100,31 @@ def test_qsharp_estimation_with_multiple_params() -> None:
     params = EstimatorParams(3)
     params.items[0].qubit_params.name = QubitParams.GATE_US_E3
     params.items[0].error_budget = 0.333
+    params.items[0].estimate_type = "singlePoint"
```
swernli, line 103: Hmm... I think the fact that this test had to change is a strong indication that we don't want to change the default in Python. This pattern here is used in notebook samples and docs pages, which means this would be a **breaking change** for Python consumers.

ivanbasov: Sure. This is a breaking change for code we just released. It is not great to make the change the next day. But if we ever want to switch it could be the only option. So, let us wait for more news today to make the final decision: if we want the frontier by default anywhere either in Python or in VS Code.

swernli:
I think a merge from main might help with the memory task, as main now skips writing the comment if the PR doesn't change memory values and this PR does not change them.

swernli:
Oh! This PR is coming from a fork. I think that is affecting how the task is able to write back. Take a look at [the logs for the benchmark task](https://github.com/microsoft/qsharp/actions/runs/7547644236/job/20559372973#step:3:454) that show the same failure, but that task treats it as a warning instead of a failure:

> Warning: Failed to comment: HttpError: Resource not accessible by integration
Commenting is not possible from forks.

@sezna I think you'll need to adjust the behavior of the memory task as well, otherwise it will continue to fail when PRs come from forks.

@ivanbasov you should be able to get into a good state by instead creating your branch in this repo rather than in your fork and then opening the PR from there. Then you can close this PR.

resource_estimator/src/estimates/data/job_params.rs
ivanbasov: Decided to return the API back to SinglePoint by default but use the Frontier as the only option in the VS Code scenaition

pip/tests/test_re.py
ivanbasov: agreed on defaults. reverted to agreed ones

samples/estimation/estimation-frontier-widgets.ipynb
```diff
@@ -0,0 +1,86 @@
...
+   "outputs": [],
+   "source": [
+    "import qsharp\n",
+    "from qsharp_widgets import SpaceChart, EstimateDetails, Summary\n",
+    "with open(\"Dynamics.qs\", \"r\") as f:\n",
+    "    contents = f.read()\n",
+    "qsharp.eval(contents)\n",
+    "\n",
+    "# esitmate for the shortest runtime and the default qubit\n",
+    "result1 = qsharp.estimate(\"QuantumDynamics.Main()\")"
```
billti, line 16: Do you do anything with result1? I don't see it referenced anywhere else. It seems to be dead code/unused and this line should be removed.

npm/src/ux/utils.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function formatThousandSepF64(val: number): string {
```
billti, line 4: Not sure we need either of these methods. The number formatter with options such as 'maximumFractionDigits', and includes locale-specific separators by default, has been widely implemented for about 10 years across browsers.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

npm/test/basics.js
```diff
@@ -879,3 +880,251 @@ test("debug service compiling multiple sources - web worker", async () => {
...
+
+test("CreateIntegerTicks: below 100", () => {
+  runAndAssertIntegerTicks(1, 1, [1]);
+  runAndAssertIntegerTicks(3, 3, [3]);
+  runAndAssertIntegerTicks(4, 6, [4, 5, 6]);
+  runAndAssertIntegerTicks(1, 100, [1, 10, 100]);
+  runAndAssertIntegerTicks(1, 10, [1, 10]);
+  runAndAssertIntegerTicks(1, 9, [1]);
+  runAndAssertIntegerTicks(2, 10, [10]);
+  runAndAssertIntegerTicks(2, 9, [2, 3, 4, 5, 6, 7, 8, 9]);
```
billti, line 898: Am I reading the above tests right that 
 - if my data runs 1 to 9, I just get a tick at 1
 - if my data runs 2 to 10, I just get a tick at 10
 - if my data runs 2 to 9, I get 8 ticks from 2 through 9

Maybe that 'feels' right in practice, but seems a little odd.

npm/test/basics.js
```diff
@@ -879,3 +880,251 @@ test("debug service compiling multiple sources - web worker", async () => {
...
+
+test("CreateTimeTicks: nanoseconds below 100", () => {
+  runAndAssertTimeTicks(1, 1, ["nanosecond"]);
+  runAndAssertTimeTicks(3, 3, ["3 nanoseconds"]);
+  runAndAssertTimeTicks(4, 6, [
+    "4 nanoseconds",
+    "5 nanoseconds",
+    "6 nanoseconds",
+  ]);
+  runAndAssertTimeTicks(1, 100, ["nanosecond"]);
```
billti, line 928: Should we prefix the singular with the value as we do for the plural (e.g. "1 nanosecond")?

npm/test/basics.js
```diff
@@ -879,3 +880,251 @@ test("debug service compiling multiple sources - web worker", async () => {
...
+
+const second = 1e9;
+const minute = 60 * second;
+const hour = 60 * minute;
+const day = 24 * hour;
+const week = 7 * day;
+const month = 30 * day;
+const year = 365 * day;
+const decade = 10 * year;
+const century = 10 * decade;
```
billti, line 1104: For readability, can we have all these consts above all the code where they're used, rather than after.

npm/test/basics.js
ivanbasov: Thank you, @billti! The intention was to show 1, 10, 100, ... if possible. If at least one of them is shown, then do not show anything else. Otherwise, show intermediate results on the lower level, e.g. 2, 3, 4, 5; or 50, 60, 70, 80, 90; or 34, 35, 36 etc.
Yes, in the "worst" case it could be 2 through 9 or similar 8 numbers. Could be worse in case of 1, 10, ... 1'000'000'000, etc.

So, the number of ticks could be from 1 to 8. My original goal was to find at least a single tick. Everything else was secondary. Maybe we could set more elegant goals like to find from 2 to 5 ticks or so. Thoughts?

npm/ux/data.ts
```diff
@@ -0,0 +1,57 @@
...
+  logicalQubit: any;
+  tfactory: any;
+  errorBudget: any;
+  physicalCounts: any;
+  physicalCountsFormatted: any;
+};
+
+export function CreateReData(
+  input: ReData,
+  frontierEntryIndex: number | null = null,
```
billti, line 27: If this doesn't need to be provided, then just mark it optional, i.e. parameter would be declared `frontierEntryIndex?: number` (see https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters)

npm/ux/data.ts
```diff
@@ -0,0 +1,57 @@
...
+    return {
+      status: input.status,
+      jobParams: input.jobParams,
+      physicalCounts: entry.physicalCounts,
+      physicalCountsFormatted: entry.physicalCountsFormatted,
+      logicalQubit: entry.logicalQubit,
+      tfactory: entry.tfactory,
+      errorBudget: entry.errorBudget,
+      logicalCounts: input.logicalCounts,
+      frontierEntries: input.frontierEntries,
```
billti, line 53: So here it's returning one frontier entry, but you still set a property called 'frontierEntries' which is all the frontierEntries. Is that needed and used for anything?

npm/ux/data.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type FrontierEntry = {
+  logicalQubit: any;
+  tfactory: any;
+  errorBudget: any;
+  physicalCounts: any;
+  physicalCountsFormatted: any;
+};
+
+export function CreateReData(
```
billti, line 25: Should this just be called `GetFrontierEntry` and require the second param? There's little point in calling it unless you pass the second param to get the specific entry, as it just reflects the first param directly.

vscode/src/webview/rePage.tsx
```diff
@@ -96,32 +60,31 @@ export function RePage(props: {
...
-        <ResultsTable
-          columnNames={columnNames}
-          data={props.estimatesData}
-          initialColumns={initialColumns}
-          ensureSelected={true}
-          onRowSelected={onRowSelected}
-          onRowDeleted={onRowDeleted}
-        />
-      </details>
+      <Summary
```
billti, line 63: We might want a different name than `Summary` for the custom component, as `summary` is a standard HTML tag that is also used in this code (and is something very different). I was confused for a second until I realized that distinction.

npm/test/basics.js
ivanbasov: I thought to follow https://quantum.microsoft.com/en-us/experience/quantum-cryptography where we use units without values. But I do not have a strong preference.

npm/package.json
```diff
@@ -20,6 +20,8 @@
     "./compiler-worker": "./dist/compiler/worker-browser.js",
     "./language-service-worker": "./dist/language-service/worker-browser.js",
     "./debug-service-worker": "./dist/debug-service/worker-browser.js",
+    "./ux/ticks": "./src/ux/ticks.ts",
+    "./ux/utils": "./src/ux/utils.ts",
```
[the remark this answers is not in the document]
ivanbasov: Tried once again. Unfortunately, basics.js could not refer to a typescript file. So, likely this solution with moving ticks.ts to ./src/ux (discussed a week ago with @billti  and @minestarks ) is the only solution to provide unit tests for this ts file. Please let me know if there are more ideas to try.

npm/ux/data.ts
ivanbasov: Actually, ReData was used for two different purposes: 
- original binding strucutre for controls
- structure received from the RE, which was extended with FrontierEntries

Now, I have intorduced SingleEstimateResult for the former purpose. Hopefully, it would clarify data conversions.

npm/test/basics.js
ivanbasov: Let us consider this aside from the current PR. I have created https://github.com/microsoft/qsharp/issues/1036

npm/package.json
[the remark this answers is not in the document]
billti: OK. I see what you were trying to do now. If it's working, let's get this in now and clean it up after. It is a bit messy/confusing when you have compiled JS, source only TSX, and package exports mixed together.

npm/ux/scatterChart.tsx
```diff
@@ -0,0 +1,355 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import {
+  CreateIntegerTicks,
+  CreateTimeTicks,
+  Tick,
+} from "qsharp-lang/ux/ticks";
```
billti, line 8: I think this should just be a relative path to the source, then you can remove the package.json changes to exports entirely, i.e.

`} from "../src/ux/ticks.js"`

(Note: Still use a .js extension and it will resolve).

This way, the unit tests in basics.js are still testing the compiled source under 'dist', but any package importing the components as source are still only pulling source, and aren't pulling a mix of source tsx and built js for the ux components.

Not sure if that makes sense, but try it. (Or leave as-is, and I'll do a follow up)

ivanbasov: Let me try

billti (approved):

npm/ux/scatterChart.tsx
ivanbasov: Seems to work. Thank you, @billti !

DmitryVasilevsky (approved):

PR #991, opened 2024-01-09
Add version check to Python package

swernli:
This adds a best effort version check to the Python package. On import it will try to check pypi.org for the latest version and compare it to the metadata for the currently installed version. It uses a 3 second timeout for the check and will catch and ignore any failures so that it doesn't interfere with using the package. The message to users is displayed on stderr to avoid having it show up along with normal script output while still showing on first import in a jupyter environment.

### DIFF
--- a/pip/qsharp/__init__.py
+++ b/pip/qsharp/__init__.py
@@ -36,3 +36,27 @@ __all__ = [
     "TargetProfile",
     "StateDump",
 ]
+
+__version__ = "0.0.0"
+
+# Check if there is a newer published version of the package
+# and warn the user if there is. This is best effort only.
+
+try:
+    import urllib.request
+    import json
+    import sys
+    from packaging.version import parse as parse_version
+
+    url = 'https://pypi.org/pypi/qsharp-lang/json'
+    with urllib.request.urlopen(url, timeout=3) as f:
+        pypi_info = json.loads(f.read().decode('utf-8'))
+    latest_version = pypi_info['info']['version']
+
+    if parse_version(latest_version) > parse_version(__version__):
+        print(
+            f'You are using qsharp {__version__}, but version {latest_version} is available.\n'
+            'Consider upgrading via the "pip install --upgrade qsharp" command.',
+            file=sys.stderr)
+except:
+    pass
--- a/pip/qsharp/__init__.py
+++ b/pip/qsharp/__init__.py
@@ -37,7 +37,8 @@ __all__ = [
     "StateDump",
 ]
 
-__version__ = "0.0.0"
+import importlib.metadata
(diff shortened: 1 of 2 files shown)


minestarks:
I don't think we should include this, but if we have to, we should write some kind of file to disk to remember the last update check and only do the check once a day. Many users will hit the 3-second timeout every single time they run a script, I guarantee it.

minestarks:
> I think that's a fair concern. I think it's worth tabling this for now, but I would like to pull out just the `__version__` property part and have that merged. Does that seem reasonable? Having that property set would be nice for those who expect it.

Seems reasonable, `__version__` appears to be standard. I think you may need to export it explicitly through `__all__` for `qsharp.__version__` to work, not sure.

minestarks:
@swernli nevermind - the PEP I was looking at says "rejected" on top. https://peps.python.org/pep-0396/#pep-rejection . This stackoverflow answer has some commentary. https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module . So maybe it's _not_ a great idea?

swernli:
> @swernli nevermind - the PEP I was looking at says "rejected" on top. https://peps.python.org/pep-0396/#pep-rejection . This stackoverflow answer has some commentary. https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module . So maybe it's _not_ a great idea?

Yeah, I was reading some of the same links. From what I gather, it's an informal convention many libraries follow but is not part of any formal specification, so it's totally optional. Given that a user can just themselves call `importlib.metadata.version('qsharp')` and get the same info, it's definitely not critical. That said, the property is something a lot of other packages do, so I could be convinced either way.

swernli:
I'm just going to drop this for now. If this ends up being a customer pain point (out of date pip packages) we can circle back to exploring how to address it.

billti:
> I don't think we should include this, but if we have to, we should write some kind of file to disk to remember the last update check and only do the check once a day. Many users will hit the 3-second timeout every single time they run a script, I guarantee it.

Did we look at what 'pip' itself does? Being that it also gives a similar message, and is far more used than we will be, I'm guessing they've dealt with any major feedback on impact.

minestarks:
> Did we look at what 'pip' itself does? Being that it also gives a similar message, and is far more used than we will be, I'm guessing they've dealt with any major feedback on impact.

The difference in my mind is that `pip` is a command-line tool, but `qsharp` is a library (more comparable to something like `numpy` IMO) . `pip` interacts with pypi _by definition_. I don't think it's a great comparison.

I also mentioned to @swernli about this: We did talk about evolving our Python package into a more full-featured CLI tool. I think an update check would be perfectly appropriate for such a tool. But I don't think `qsharp` is it, in its current form. It's a small, fast, library.

PR #993, opened 2024-01-09
Rename pip package qsharp-lang -> qsharp

minestarks:


swernli (approved):

billti (changes requested):
See lines 13 & 22 in https://github.com/microsoft/qsharp/blob/main/vscode/src/notebookTemplate.ts also.

minestarks:
> See lines 13 & 22 in https://github.com/microsoft/qsharp/blob/main/vscode/src/notebookTemplate.ts also.

Oh! Missed that because it was in a sea of TypeScript files I was ignoring (since we're not renaming the npm package). I'll do another pass to be sure

billti (approved):

PR #1001, opened 2024-01-10
Update sample notebooks

minestarks:
Just rerun the sample notebooks with the latest bits to pick up any differences in output. Nothing major.


pip/samples/azure_submission.ipynb
minestarks, line 1: You may want to hide whitespace when reviewing these, though there's really not much interesting in the changes.

image

PR #1005, opened 2024-01-10
Remove syntax highlighting rule that breaks codemirror

minestarks:
This causes JupyterLab to throw an error when opening .qs documents. 

They were way too many backslashes in the regex... the CodeMirror parser didn't like that, failed parsing, which brought down the whole editor for some reason.


jupyterlab/src/index.ts
```diff
@@ -68,7 +68,7 @@ function registerQSharpLanguage(
     },
     {
       token: "operatorKeyword",
-      regex: String.raw`(=)|(!)|(<)|(>)|(\\+)|(-)|(\\*)|(\\/)|(\\^)|(%)|(\\|)|(\\&\\&\\&)|(\\~\\~\\~)|(\\.\\.\\.)|(\\.\\.)|(\\?)`,
```
billti, line 71: So.... was this never working/parsing right?

Shouldn't `(\\/)` still leave one slash? (ie.g. `(\/)` instead of `(/)`?

minestarks: No, it was never working.

`/` is not a special character in regex, they just need to be escaped when using the `/regex/` syntax. We're using a string here so it should be fine.

https://javascript.info/regexp-escaping#a-slash

So intuitive right?

minestarks: Just to be clear: this line was never working, but it was only relatively recently added. I didn't mean to imply syntax highlighting was broken from the start.

billti: OK. I've no idea how CodeMirror works, so as long as it's not getting eval'ed into a RegEx literal or something funky like that, the / should be fine (I think it's benign to still escape though). Thanks!

billti (approved):

Issue #767, opened 2023-10-03
Add advanced quantum algorithms samples


cesarzc:
This issue tracks the progress towards expanding advanced Q# samples.
- [ ] 1️⃣ Shor's quantum subroutine
- [ ] 1️⃣ GHZ
- [ ] 1️⃣ Cat state
- [ ] 2️⃣ CHSH game
- [ ] 2️⃣ Noisy amplitude estimation
- [ ] 2️⃣ Oracle synthesis
- [ ] 2️⃣ Order finding
- [ ] 2️⃣ Repeat until success
- [ ] 2️⃣ Reversible logic synthesis
- [ ] 2️⃣ Sudoku Grover
- [ ] 2️⃣ VQE
- [ ] 2️⃣ Quantum adder
- [ ] 2️⃣ Phase estimation
- [ ] 1️⃣ Bit-flip code
- [ ] 1️⃣ Phase-flip code
- [ ] 2️⃣ Shor code
- [ ] 2️⃣ Error syndrome
- [ ] 2️⃣ Modular addition operator
- [ ] 2️⃣ H2 simulation
- [ ] 2️⃣ Simple Ising model
- [ ] 2️⃣ QAOA

Code:
- 1️⃣ Samples that should be done first.
- 2️⃣ Samples that should be done second.

limitless-100:
Is this still an open and active issue? @cesarzc

PR #1010, opened 2024-01-11
Add more samples

swernli:
This adds some additional samples folks can use for estimation.


samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
+/// # Sample
+/// Resource Estimation for Shor's algorithm
+///
+/// # Description
+/// In this sample we concentrate on costing the `EstimateFrequency`
+/// operation, which is the core quantum operation in Shor's algorithm, and
+/// we omit the classical pre- and post-processing.
+namespace Shors {
```
DmitryVasilevsky, line 8: So, this is the quantum part of Shor? Would it make sense to try some new library arithmetic and see if estimates change?

DmitryVasilevsky (approved):

samples/estimation/ShorRE.qs
swernli: I think this one was tuned already to use `CCNOT` directly for AND in order to keep the counts consistent with our other samples. But definitely worth letting @msoeken take a look and refine in a later follow up too.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+    /// # Description
+    /// Given a non-empty quantum register |𝑦⟩ of length 𝑛+1 and a positive
+    /// constant 𝑐 < 2ⁿ, computes |𝑦 + c⟩ into |𝑦⟩.
+    ///
+    /// # Input
+    /// ## c
+    /// Constant number to add to |𝑦⟩.
+    /// ## y
+    /// Quantum register of second summand and target; must not be empty.
+    internal operation AddConstant(c : Int, y : Qubit[]) : Unit is Adj + Ctl {
```
msoeken, line 216: This can be replaced by the library version.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+    /// ## c
+    /// Constant value for comparison.
+    /// ## x
+    /// Quantum register to compare against.
+    /// ## target
+    /// Target qubit for comparison result.
+    ///
+    /// # Reference
+    /// This construction is described in [Lemma 3, arXiv:2201.10200]
+    internal operation CompareGreaterThanOrEqualConstant(c : Int, x : Qubit[], target : Qubit)
```
msoeken, line 258: This can be replaced by the library version.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+                }
+            } apply {
+                ApplyLowTCNOT(Tail(qs), target);
+            }
+        }
+    }
+
+    /// # Summary
+    /// Internal operation used in the implementation of GreaterThanOrEqualConstant.
+    internal operation ApplyOr(control1 : Qubit, control2 : Qubit, target : Qubit) : Unit is Adj {
```
msoeken, line 293: This is not needed after replacing CompareGreaterThanOrEqualConstant.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+        within {
+            ApplyToEachA(X, [control1, control2]);
+        } apply {
+            ApplyAnd(control1, control2, target);
+            X(target);
+        }
+    }
+
+    internal operation ApplyAnd(control1 : Qubit, control2 : Qubit, target : Qubit)
```
msoeken, line 302: This is not needed after replacing CompareGreaterThanOrEqualConstant.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// # Summary
+    /// Returns the number of trailing zeroes of a number
+    ///
+    /// ## Example
+    /// ```qsharp
+    /// let zeroes = NTrailingZeroes(21); // = NTrailingZeroes(0b1101) = 0
+    /// let zeroes = NTrailingZeroes(20); // = NTrailingZeroes(0b1100) = 2
+    /// ```
+    internal function NTrailingZeroes(number : Int) : Int {
```
msoeken, line 325: This is not needed after replacing CompareGreaterThanOrEqualConstant.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,364 @@
...
+            set nZeroes += 1;
+            set copy /= 2;
+        }
+        return nZeroes;
+    }
+
+    /// # Summary
+    /// An implementation for `CNOT` that when controlled using a single control uses
+    /// a helper qubit and uses `ApplyAnd` to reduce the T-count to 4 instead of 7.
+    internal operation ApplyLowTCNOT(a : Qubit, b : Qubit) : Unit is Adj+Ctl {
```
msoeken, line 338: This is not needed after replacing CompareGreaterThanOrEqualConstant.

msoeken (changes requested):
Made some comments on where to use library operations.

ivanbasov (commented):
Examples are great! They add a great variety from small snippets to mid-size academic problems.

Maybe it is worth to correct title comments. Now they state that these are 'Resource Estimator' samples. But actually those are programs which is interesting to consider in the resource estimator scope.

For example, if we would consider Jupyter notebooks for those samples, we could put qs files aside and have qsharp.estimate(...) in an actual sample.

samples/estimation/ShorRE.qs
swernli: Thanks for the feedback! I have incorporated that in both this and the simulator sample, after testing to confirm behavior is as expected.

swernli:
> Examples are great! They add a great variety from small snippets to mid-size academic problems.
> 
> Maybe it is worth to correct title comments. Now they state that these are 'Resource Estimator' samples. But actually those are programs which is interesting to consider in the resource estimator scope.
> 
> For example, if we would consider Jupyter notebooks for those samples, we could put qs files aside and have qsharp.estimate(...) in an actual sample.

Fair point, I'll adjust my phrasing. Right now the Jupyter samples are all in other folders, we might want to consider consolidating into the single samples folder here at the top level.

samples/estimation/ShorRE.qs
```diff
@@ -0,0 +1,203 @@
+/// # Sample
```
cesarzc, line 1: A quantum-only Shor sample has been in the plan for more advanced algorithmic sample (#767). Since this sample is exactly that, would it be make sense to move it to the algorithms folder as `ShorEstimateFrequency` or something similar? Or is the objective to have a specific set of samples dedicated to RE only and those are expected to be in the `estimation` folder to be able to refer people to it?

swernli: For now the intent is to have this only be for RE. As configured, it's too big to simulate reasonably, and the entry point doesn't return any values. I think keeping this as estimation specific might make more sense until we have a better plan for what we want out of the quantum only version.

billti (approved):

cesarzc (approved):

samples/algorithms/Shor.qs
```diff
@@ -297,62 +298,35 @@ namespace Sample {
...
-    /// `generator` and N is `modulus`.
+    /// Interprets `target` as encoding unsigned little-endian integer k
+    /// and performs transformation |k⟩ ↦ |gᵖ⋅k mod N ⟩ where
+    /// p is `power`, g is `generator` and N is `modulus`.
     ///
     /// # Input
     /// ## generator
-    /// The unsigned integer multiplicative order (period) of which is being
-    /// estimated. Must be co-prime to `modulus`.
+    /// The unsigned integer multiplicative order ( period )
```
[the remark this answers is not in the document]
swernli: Fixed!

samples/algorithms/Shor.qs
```diff
@@ -297,62 +298,35 @@ namespace Sample {
...
-        }
-        return nZeroes;
+    /// Register interpreted as  little-endian which is multiplied by
+    /// given power of the generator. The multiplication is performed modulo
+    /// `modulus`.
+    internal operation ApplyOrderFindingOracle(
+        generator : Int, modulus : Int, power : Int, target : Qubit[]
+    )
+    : Unit
+    is Adj + Ctl {
```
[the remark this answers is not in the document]
swernli: I'm not a fan, to be honest; I prefer them all on the same line, but it's fine.

PR #1011, opened 2024-01-11
Completion List Updates

ScottCarda-MS:
Ignore "Microsoft.Quantum.Unstable" and internal items from completion list. Internal items are only ignored if the user would not have access to them at the cursor.


minestarks (approved):
I think this will also exclude items that are declared `internal` in a notebook cell, but I'm ok with that (I don't see an easy way to "fix" it). Declaring internal items in notebook cells would be a really weird thing to do.

PR #1014, opened 2024-01-12
Formally introduce an FIR `PackageStore` struct and consolidate lookups

cesarzc:
This change formally introduces an FIR `PackageStore` such that it can be later used in runtime capabilities analysis (RCA).

It also refactors the way lookups are performed on an FIR package store in a consolidated way, reducing code duplication and making the lookup functionality available to the RCA implementation.

### DIFF
--- a/compiler/qsc_fir/src/fir.rs
+++ b/compiler/qsc_fir/src/fir.rs
@@ -327,6 +327,20 @@ pub struct ItemLookupId {
     pub item: LocalItemId,
 }
 
+impl Display for ItemLookupId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "<item {} in package {}>", self.item, self.package)
+    }
+}
+
+impl ItemLookupId {
+    /// Creates a `ItemLookupId`.
+    #[must_use]
+    pub fn from(package: PackageId, item: LocalItemId) -> Self {
+        Self { package, item }
+    }
+}
+
 /// A unique identifier for a block within a package store.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct BlockLookupId {
@@ -334,6 +348,20 @@ pub struct BlockLookupId {
     pub package: PackageId,
     /// The item ID.
     pub block: BlockId,
+}
+
+impl Display for BlockLookupId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "<block {} in package {}>", self.block, self.package)
+    }
+}
+
+impl BlockLookupId {
+    /// Creates a `BlockLookupId`.
+    #[must_use]
+    pub fn from(package: PackageId, block: BlockId) -> Self {
+        Self { package, block }
+    }
 }
 
 /// A unique identifier for an expression within a package store.
@@ -345,6 +373,20 @@ pub struct ExprLookupId {
     pub expr: ExprId,
 }
 
+impl Display for ExprLookupId {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "<expression {} in package {}>", self.expr, self.package)
+    }
+}
+
+impl ExprLookupId {
+    /// Creates an `ExprLookupId`.
+    #[must_use]
+    pub fn from(package: PackageId, expr: ExprId) -> Self {
+        Self { package, expr }
+    }
+}
+
 /// A unique identifier for a pattern within a package store.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct PatLookupId {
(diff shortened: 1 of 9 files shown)


compiler/qsc_fir/src/fir.rs
```diff
@@ -351,6 +587,36 @@ impl Display for Package {
...
+        match &self.items.get(id)?.kind {
+            ItemKind::Callable(callable) => Some(Global::Callable(callable)),
+            ItemKind::Namespace(..) => None,
+            ItemKind::Ty(..) => Some(Global::Udt),
+        }
+    }
+
+    fn get_item(&self, id: LocalItemId) -> &Item {
+        self.items.get(id).expect("Item not found")
+    }
```
minestarks, line 609: `get_item` is almost the same as `get_global`, but it returns namespaces as well. It's also unused (rather, it's called from the corresponding `PackageStore` method, but _that_ method is unused). Do we need it in the interface, or can we remove it to reduce ambiguity?

compiler/qsc_fir/src/fir.rs
```diff
@@ -310,6 +313,239 @@ impl Display for Res {
...
+pub enum Global<'a> {
+    /// A global callable.
+    Callable(&'a CallableDecl),
+    /// A global user-defined type.
+    Udt,
+}
+
+/// A unique identifier for an item within a package store.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct StoreItemId {
```
minestarks, line 326: Since we're going with names that parallel the corresponding HIR types (which I like), can we call this just `ItemId` to mirror `hir::ItemId`?

compiler/qsc_fir/src/fir.rs
```diff
@@ -310,6 +313,239 @@ impl Display for Res {
...
+    fn from(tuple: (PackageId, StmtId)) -> Self {
+        Self {
+            package: tuple.0,
+            stmt: tuple.1,
+        }
+    }
+}
+
+/// A trait to find elements in a package store.
+pub trait PackageStoreLookup {
```
minestarks, line 445: Does this really need to be a trait if it has only one impl?

minestarks (approved):
I think this is a great idea!

swernli (approved):
I like how this turned out, thanks for the refactor!

compiler/qsc_fir/src/fir.rs
```diff
@@ -310,6 +313,239 @@ impl Display for Res {
...
+    pub fn insert(&mut self, id: PackageId, package: Package) {
+        self.0.insert(id, package);
+    }
+
+    /// Gets a package store iterator.
+    #[allow(clippy::iter_not_returning_iterator)]
+    #[must_use]
+    pub fn iter(&self) -> Iter<PackageId, Package> {
+        self.0.iter()
+    }
```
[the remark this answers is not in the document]
cesarzc: This refactoring was in part intended to make runtime capabilities analysis (RCA) more convenient, and even though an iterator is not used in these changes, I'm using it in the RCA implementation.

I'll implement `Iterator` properly :)

compiler/qsc_fir/src/fir.rs
cesarzc: I think we can remove it. I'll do that.

compiler/qsc_fir/src/fir.rs
cesarzc: I think keeping this trait even if it has only one implementation was a few advantages:
- IMO it keeps code more organized because otherwise we would implement these functions in the `fir::PackageStore` implementation.
- When the trait is used in arguments to functions, it makes it clear that the argument is only used as a lookup. The functions that receive a `PackageStoreLookup` argument can't do other things like iterate through packages or get individual packages.

Having said that, I don't have enough experience with Rust to know if this is a common pattern so I'm open to change it if you think these advantages are not enough to warrant a trait with only one implementation.

compiler/qsc_fir/src/fir.rs
cesarzc: There's already an `fir::ItemId` type, which is analogous to `hir::ItemId`. However, there is a slight difference between `fir::StoreItemId` and `fir::ItemId` which is the `package` field (one is an `Option` and the other one is not).

I think in this case it makes sense to keep `StoreItemId` for the additional type safety of always having a `package` field that is not an `Option`.

compiler/qsc_fir/src/fir.rs
minestarks: I like it this way, actually. It's funny because I did the exact same thing in a PR where I named a trait `Lookup`, had only one implementation, and got the same question that I just asked you 😄  see: https://github.com/microsoft/qsharp/pull/759/files#r1380776877  . So I completely understand your reasoning, I was just curious what your take was!

compiler/qsc_fir/src/fir.rs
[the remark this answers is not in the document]
cesarzc: Looking at the code more closely, the underlying `IndexMap` data structure seems like it properly implements `Iterator` so I'm just surfacing the function here. Also, the clippy warning seems like a non-issue now.

compiler/qsc_fir/src/fir.rs
cesarzc: Removed.

compiler/qsc_fir/src/fir.rs
cesarzc: I liked the God object mention in your response!

Issue #963, opened 2024-01-02
Completions integration test is flaky


minestarks:
I've seen this test fail about once a week in merge queue checks. Likely some race condition in the test.

Logs:
https://github.com/microsoft/qsharp/actions/runs/7389502439/job/20102481251#step:9:493

On `macos-latest`:

```
%c INFO color: #33f %s%s    Q# Language Service Tests
    Q# Language Service Tests
  GET /static/mount/test.qs?stat 200 5.370 ms - 65
  GET /static/mount/test.qs 200 13.141 ms - 158
  GET /static/build/node_modules/vscode-textmate/release/main.js 200 16.697 ms - 57923
  GET /static/build/node_modules/vscode-oniguruma/release/main.js 200 18.833 ms - 11728
  GET /static/devextensions/language-configuration.json 200 22.019 ms - 403
  %c INFO color: #33f     ✅ %s (%dms) Q# language is registered 91
      ✅ Q# language is registered (91ms)
  GET /static/build/node_modules/vscode-oniguruma/release/onig.wasm 200 21.984 ms - 466610
  GET /static/build/extensions/emmet/dist/browser/emmetBrowserMain.js 200 3.848 ms - 250922
  GET /static/devextensions/out/extension.js 200 39.392 ms - 1313610
  GET /static/devextensions/syntaxes/qsharp.tmLanguage.json 200 1.863 ms - 2274
  GET /static/build/out/vs/base/worker/workerMain.js 200 14.800 ms - 348151
  GET /static/build/out/vs/editor/common/services/editorSimpleWorker.nls.js 200 6.659 ms - 761
  GET /static/build/out/vs/workbench/services/textMate/browser/backgroundTokenization/worker/textMateTokenizationWorker.worker.js 200 3.085 ms - 86489
  GET /static/build/node_modules/vscode-textmate/release/main.js 200 15.927 ms - 57923
  GET /static/build/node_modules/vscode-oniguruma/release/main.js 200 1.885 ms - 11728
  GET /static/build/node_modules/vscode-oniguruma/release/onig.wasm 200 5.261 ms - 466610
  GET /static/devextensions/wasm/qsc_wasm_bg.wasm 200 3.033 ms - 2295989
  %c INFO color: #33f     %d) %s 1 Completions
      1) Completions
  %c INFO color: #33f     ✅ %s (%dms) Definition 1916
      ✅ Definition (1916ms)
  %c INFO color: #33f     ✅ %s Diagnostics
[...]

PR #1015, opened 2024-01-12
More predictable init and better logging in VS Code tests

minestarks:
This is an attempt at fixing or mitigating #963, but since I can't repro that one reliably, we won't be sure for a while.

* Product change: fix VS Code log formatting.

Before:
Screenshot 2024-01-12 110423
After:
Screenshot 2024-01-12 110110

Test changes: 

* Explicitly activate the Q# extension before the tests start. This should eliminate issues where the activation itself is taking a long time for whatever reason and timing out the test cases, which have a reasonably conservative timeout of 2s.
* Write Q# logs to the console in the integration tests.
* Disable the ugly verbose vscode / webserver logs by default. They were helpful to debug infrastructure issues but we don't need them at the moment.

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -41,9 +41,24 @@ import { createReferenceProvider } from "./references.js";
 import { activateTargetProfileStatusBarItem } from "./statusbar.js";
 import { initFileSystem } from "./memfs.js";
 import { getManifest, readFile, listDir } from "./projectSystem.js";
+import {
+  Logging,
+  initLogForwarder,
+  initOutputWindowLogger,
+} from "./logging.js";
 
-export async function activate(context: vscode.ExtensionContext) {
-  initializeLogger();
+export async function activate(
+  context: vscode.ExtensionContext,
+): Promise<ExtensionApi> {
+  const api: ExtensionApi = {};
+
+  if (context.extensionMode !== vscode.ExtensionMode.Test) {
+    initOutputWindowLogger();
+  } else {
+    // Don't log to the output window in tests, forward to a listener instead
+    api.logging = initLogForwarder();
+  }
+
   log.info("Q# extension activating.");
   initTelemetry(context);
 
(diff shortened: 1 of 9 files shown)


swernli (approved):

swernli:
I think you might've meant to link this to #963. It doesn't seem related to the linked issue, which is assigned to @ScottCarda-MS.

vscode/src/extension.ts
```diff
@@ -41,9 +41,24 @@ import { createReferenceProvider } from "./references.js";
...
+  initLogForwarder,
+  initOutputWindowLogger,
+} from "./logging.js";
+
+export async function activate(
+  context: vscode.ExtensionContext,
+): Promise<ExtensionApi> {
+  const api: ExtensionApi = {};
+
+  if (context.extensionMode !== vscode.ExtensionMode.Test) {
```
idavis, line 55: nit: reverse the conditional blocks so that we don't have to negate the check.

idavis (approved):

vscode/test/suites/extensionUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import * as vscode from "vscode";
+import { type ExtensionApi } from "../../src/extension";
+
+// Q# extension log level. Increase this for debugging.
+const extensionLogLevel = "warn";
+
+export async function activateExtension() {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const ext = vscode.extensions.getExtension("quantum.qsharp-lang-vscode-dev")!;
```
billti, line 12: The extension id will be different between dev and release. Is that safe here? (And if so, please add a comment as to why)

vscode/test/suites/extensionUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (ext.isActive) {
+    return;
+  }
+
+  const start = performance.now();
+  const extensionApi: ExtensionApi = await ext.activate();
+  const logForwarder = extensionApi.logging;
+  if (!logForwarder) {
+    throw new Error(`qsharp-tests: extension did not return a log forwarder`);
+  } else {
```
cesarzc, line 22: Maybe a naive comment since I don't know ts that well. Is the `else` needed? Since an error is thrown inside the `if` block already, could the logic inside the `else` block just be part of the function block?

vscode/test/suites/extensionUtils.ts
billti: (I'm guessing it's because it's test code and thus we assume the 'dev' extension, but wouldn't hurt to be clear)

vscode/test/suites/extensionUtils.ts
cesarzc: IMO it is more legible having this logic outside the `else` block, but it is just a preference and there might be some reason for doing it this way that I don't know about so feel free to ignore this suggestion if it doesn't make sense.

cesarzc (approved):

billti (approved):

vscode/test/suites/extensionUtils.ts
billti: (I wouldn't worry about it if it means pushing more commits and waiting though, so resolving for now)

vscode/test/suites/extensionUtils.ts
minestarks: I totally agree, thanks for the suggestion.

vscode/src/extension.ts
minestarks: Ha, I had an internal battle between "the preferred code branch should come first" and "no inequality checks", and the first one won. I'll switch it though.

PR #1017, opened 2024-01-12
Add custom user agent header for service submissions 

sezna:


vscode/src/telemetry.ts
```diff
@@ -250,3 +250,19 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
...
+  if (navigator.userAgentData) {
+    const browser =
+      navigator.userAgentData.brands[navigator.userAgentData.brands.length - 1];
+    return `${browser.brand}/${browser.version}`;
+  } else {
+    return "";
+  }
+}
+
+export function getUserAgent(context: vscode.ExtensionContext): string {
```
billti, line 264: There's a LOT of plumbing through the context everywhere in this PR. Being that you already need to pass the context to `initTelemetry` (line 227 above) which already constructs a singleton (`reporter`) for use in any calls to send telemetry (line 237 above), could you not just construct the user agent string once in `initTelemetry` (it's not going to change for the lifetype of the app) and use it generally in `sendTelemetryEvent`?

Or do we absolutely specifically only want this propery on Azure network request events?

vscode/src/telemetry.ts
```diff
@@ -250,3 +250,19 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
     )}`,
   );
 }
+
+function getBrowserRelease(): string {
+  if (navigator.userAgentData) {
+    const browser =
+      navigator.userAgentData.brands[navigator.userAgentData.brands.length - 1];
```
billti, line 257: We discussed already in a separate thread that this isn't cross-browser available, and that we run in the browser. If I try and evaluate this expression in Safari I get the below. Please include the appropriate error handling so this doesn't throw exceptions.

image

billti (changes requested):
See comments.

vscode/src/telemetry.ts
billti: Doh! How'd I miss that. Yeah, should do it.

Just in case some browser (ahem... Safari) ever adds 'userAgentData' but still doesn't add the child properties like 'brands', might want to change the condition on 255 to be `navigator.userAgentData?.brands` for extra safety).

vscode/src/telemetry.ts
```diff
@@ -250,3 +254,19 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
     )}`,
   );
 }
+
+function getBrowserRelease(): string {
+  if (navigator.userAgentData) {
+    const browser =
+      navigator.userAgentData?.brands[
```
billti, line 261: There's no point adding this here, you've already checked `navigator.userAgentData' is not 'falsy' in the if check. The if check is where `navigator.userAgentData?.brands` should be used, which is a safe way to check userAgentData.brands exists on navigator before you get to this line.

vscode/src/telemetry.ts
```diff
@@ -250,3 +254,19 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
...
+        navigator.userAgentData.brands.length - 1
+      ];
+    return `${browser.brand}/${browser.version}`;
+  } else {
+    return "";
+  }
+}
+
+export function getUserAgent(): string {
+  return userAgentString || "";
```
billti, line 271: Shouldn't it fall back to 'navigator.userAgent' rather than an empty string?

billti: Not sure I follow. You set the User-Agent header to the value returned by this function, correct? (line 84 in networkRequests.ts). The "User-Agent" header on HTTP requests is by default the same value as navigator.userAgent, is my understanding:

Network trace:

image

'navigator.userAgent' value

image

So why would this be an invalid default if userAgentString doesn't have a value?

Or am I missing something?

vscode/src/telemetry.ts
```diff
@@ -2,6 +2,8 @@ import * as vscode from "vscode";
 import TelemetryReporter from "@vscode/extension-telemetry";
 import { log } from "qsharp-lang";
 
+/// <reference types="user-agent-data-types" />
```
billti, line 5: This should go just after the license header, but just noticed this file forget to add one. Please add. (First  4 lines should look similar to https://github.com/microsoft/qsharp/blob/main/vscode/src/webview/webview.tsx - but with the different package name).

vscode/src/telemetry.ts
```diff
@@ -250,3 +256,19 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
     )}`,
   );
 }
+
+function getBrowserRelease(): string {
+  if (navigator.userAgentData) {
+    const browser =
+      navigator.userAgentData?.brands[
```
billti, line 263:
```suggestion
  if (navigator.userAgentData?.brands) {
    const browser =
      navigator.userAgentData.brands[
```

billti (approved):

PR #1019, opened 2024-01-13
Samples for resource estimation of simple circuits

tcNickolas:
This change adds two Python Jupyter notebooks that define a random or a hardcoded Q# circuit and run resource estimation for it.


swernli (approved):
Samples look good, notebooks run nice and fast!

swernli (approved):
Samples look good, notebooks run nice and fast!

PR #1020, opened 2024-01-15
Only comment memory benchmark results if there is a difference

sezna:
With this change, our pipeline only comments about memory usage if there is a diff. If not, we omit the comment but log in the pipeline to confirm that there was no diff. I hope this reduces per-PR noise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,13 +178,18 @@ jobs:
       - uses: actions/github-script@v6
         with:
           script: |
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: `## Memory Report for ${{ github.sha }}
-            | Test                        | This Branch | On Main  |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -178,13 +178,18 @@ jobs:
...
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: `## Memory Report for ${{ github.sha }}
-            | Test                        | This Branch | On Main  |
-            |-----------------------------|-------------|----------|
-            | compile core + standard lib | ${{ env.BRANCH_MEASUREMENT }} bytes | ${{ env.MAIN_MEASUREMENT }} bytes |`
-            })
+            if (${{ env.BRANCH_MEASUREMENT }} !== ${{ env.MAIN_MEASUREMENT }}) {
```
swernli, line 181: Would it be possible to combine this condition into the if below? Then there would be a single place to check whether the step will run rather than two.

swernli (approved):

PR #1022, opened 2024-01-16
Scatter charts for RE

ivanbasov:


ivanbasov:
run memory profile is still failing. let us return to the original PR

PR #1027, opened 2024-01-17
Drop `stateful` distinction in favor of single interpreter

swernli:
This is a move only change and does not introduce any implementation differences. The code that was previously in `interpret::stateful` has been moved to `interpret` so that the stateful distinction can be dropped in favor of what was already the single interpreter implementation.


compiler/qsc/src/interpret.rs
minestarks, line 1: Is this just a copy of stateful.rs? Can I skip reading it?

swernli: Yup, that's exactly what it is.

minestarks (approved):

cesarzc (approved):

ivanbasov (approved):

PR #1028, opened 2024-01-17
Compute distance and logical error rate from lookup table

msoeken:
As alternative to formula coefficients, a lookup table can be provided to compute code distance for logical error rate (and vice versa). The lookup table contains triples with values for physical error rate, logical error rate, and code distance.

Left to do:

- [ ] Add test case
- [ ] Integrate with `ProtocolSpecification`


msoeken:
Closing this one, because its design will look different in light of #1059

PR #1029, opened 2024-01-17
Turn core/std lib into a crate

idavis:
This change makes core/std lib sources available via a crate instead of loading through a macro in the compiler. General effect is that the sources should only be compiled when changes are made to the library vs the compiler speeding up workflows in the compiler.

This can also help in future use of the language services if needed to have direct access to the library sources.

The tests have been updated to follow more established patterns for tests used in the rest of the repository.


swernli (approved):

library/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+// The core prefix on the name is needed to disambiguate from the std
+// files of the same name. This comes in during debugging when we need
+// to load a core/std file from the library.
+pub const CORE_LIB: &[(&str, &str)] = &[
+    ("core/core.qs", include_str!("../core/core.qs")),
+    ("core/qir.qs", include_str!("../core/qir.qs")),
+];
+
+pub const STD_LIB: &[(&str, &str)] = &[
+    ("arrays.qs", include_str!("../std/arrays.qs")),
```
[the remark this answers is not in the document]
idavis: @DmitryVasilevsky To my knowledge, there is no macro that can enumerate the file system and do this at compile time.

library/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+        &Value::Array(vec![Value::Int(1), Value::Int(1)].into()),
+    );
+    test_expression(
+        "Repeated(true, 3)",
+        &Value::Array(vec![Value::Bool(true), Value::Bool(true), Value::Bool(true)].into()),
+    );
+}
+
+#[test]
+fn check_exp_with_cnot() {
```
cesarzc, line 109: Isn't exp part of the `Microsoft.Quantum.Intrinsic` namespace? Maybe this tests belong in the same file that other intrinsic callables are tested.

idavis: I didn't change the effective location of those tests. If they make sense in another file, we can move them in a follow up PR.

PR #1034, opened 2024-01-18
Rename a couple of runtime capability flags for consistency

cesarzc:
Rename a couple of runtime capability flags for consistency before they are extensively used in runtime capabilities analysis.

### DIFF
--- a/compiler/qsc_frontend/src/compile.rs
+++ b/compiler/qsc_frontend/src/compile.rs
@@ -39,9 +39,9 @@ use thiserror::Error;
 bitflags! {
     #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct RuntimeCapabilityFlags: u32 {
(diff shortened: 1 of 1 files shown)


swernli (approved):

compiler/qsc_frontend/src/compile.rs
```diff
@@ -39,9 +39,9 @@ use thiserror::Error;
 bitflags! {
     #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct RuntimeCapabilityFlags: u32 {
-        const ConditionalForwardBranching = 0b0000_0001;
```
DmitryVasilevsky, line 42: So, they aren't used at all at this time?

cesarzc: Not yet in main but I'm using them in this [branch](https://github.com/microsoft/qsharp/tree/cesarzc/rca) from which I'll open a PR sometime soon.

PR #1035, opened 2024-01-18
Refactor common interpreter patterns

swernli:
This change refactors the interpreter API to unify around some common patterns and reduce places that look like exceptions. Several of the APIs used what appeared to be different patterns for how they invoked the evaluator when it turns out they were all minor, surface differences except for a few differences in input values. Using a wrapper function clarifies where the common patterns are and makes it easier for future updates to reuse them.

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -170,45 +170,45 @@ impl Display for Spec {
     }
 }
 
-/// Evaluates the given expr with the given context.
-/// # Errors
-/// Returns the first error encountered during execution.
-/// # Panics
-/// On internal error where no result is returned.
-pub fn eval_expr(
-    state: &mut State,
-    expr: ExprId,
-    globals: &impl PackageStoreLookup,
-    env: &mut Env,
-    sim: &mut impl Backend<ResultType = impl Into<val::Result>>,
-    out: &mut impl Receiver,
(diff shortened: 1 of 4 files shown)


compiler/qsc_eval/src/lib.rs
```diff
@@ -170,45 +170,46 @@ impl Display for Spec {
...
     receiver: &mut impl Receiver,
 ) -> Result<Value, (Error, Vec<Frame>)> {
     let mut state = State::new(package);
-    state.push_stmt(stmt);
+    match id {
+        EvalId::Expr(expr) => state.push_expr(expr),
+        EvalId::Stmt(stmt) => state.push_stmt(stmt),
+    }
     let res = state.eval(globals, env, sim, receiver, &[], StepAction::Continue)?;
     let StepResult::Return(value) = res else {
```
idavis, line 211: This function is named `eval`

compiler/qsc_eval/src/lib.rs
```diff
@@ -170,45 +170,46 @@ impl Display for Spec {
...
-        panic!("eval_stmt should always return a value");
+        panic!("eval_expr should always return a value");
```
swernli, line 212:
```suggestion
        panic!("eval should always return a value");
```

idavis (approved):

PR #1038, opened 2024-01-18
Return utf-16 line and column offsets directly from the language service and debugger

### LINKED ISSUE #851: Converting utf-8 offsets to utf-16 line/column positions should not depend on file i/o
This isn't an issue today but will become an issue when we add multi-file support.

For features such as go-to-definition, the language service returns result spans in utf-8 offsets. The VS Code / Monaco APIs expect utf16 line/column values. In order to map from one to the other, we need the file contents.

This mapping is done today in the npm package:
https://github.com/microsoft/qsharp/blob/main/npm/src/language-service/language-service.ts#L197

However, when the result points to a file that isn't currently open in the editor, we don't have the contents for that file available at that layer. For std/core libs, we work around this problem by retrieving the file contents via a dedicated function in the wasm crate. But for other files this wouldn't be an option.

This will mean that the offsets won't get mapped properly, and the cursor may be shifted if the target file contains complex Unicode characters .

I can put together a real repro once we have the multi-file feature working.

This likely impacts the debugger to, as it also does utf8-utf16 mapping, but I'm less familiar with it to think of a good repro.

There are two ways we could fix this:
- Add a function to read source contents for "remote" files through the WASM create (same as we do for std/core libs today, but for user files)
- Do the utf8-utf16 mapping in the Rust layer and have the language service return utf16 offsets (better yet, line/column pairs) directly.

minestarks:
Fixes #851

- Introduce the `Position` and `Range` types which contain line/column offsets. These are the equivalent of `offset` and `Span` in the compiler, but with line/columns instead of offsets. 

- Given a source string and offset, we can generate line/column positions in terms of both UTF-8 and UTF-16 code units. UTF-16 is useful for JavaScript, UTF-8 everywhere else we might use them.

- The `BreakpointSpan` and `StackFrame`s returned by the debugger now use these line/column `Range`s instead of offsets.

- All the language service methods now take `Position`s instead of `offset`, and return `Range`s instead of `Span`s in any response structs.

- DELETE all the code that has to deal with line/column/utf-16 mapping in the JavaScript layer, getting rid of a lot of document loading and reading code!


idavis (approved):

language_service/src/protocol.rs
```diff
@@ -83,25 +77,23 @@ impl Eq for CompletionItem {}
 use std::hash::{Hash, Hasher};
 
 impl Hash for CompletionItem {
-    // exclude sort text for hashing
+    // use only label and kind for hashing
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.label.hash(state);
         self.kind.hash(state);
-        self.detail.hash(state);
```
ScottCarda-MS, line 90: I think we still need to hash on everything except sort text to be able to distinguish between two items that have the same label, but are actually different in the changes they will make.

compiler/qsc_data_structures/src/line_column.rs
```diff
@@ -0,0 +1,192 @@
...
+    // with more utf-8 code units (bytes) than utf-16 code units, to demonstrate
+    // how code unit offset will differ depending on the chosen encoding.
+
+    // chars                    | 𝑓                 (        𝑥                 ⃗        )     <eof>
+    // unicode code point       | 1d453             28       1d465             20d7     29
+    // utf-8 units (bytes)      | f09d9193          28       f09d91a5          e28397   29
+    // utf-16 units             | d835     dc53     28       d835     dc65     20d7     29
+    // char offset              | 0                 1        2                 3        4
+    // utf-8 byte offset        | 0                 4        5                 9        12
+    // utf-16 code unit offset  | 0                 2        3                 5        6
```
billti, line 137: UTF-16, as the name says, always has 16 bit code units, so there is never an odd offset. Looks like 3 & 5 are wrong here, and in line 134 above the code units should be 0x0028 and 0x0029, not 28 & 29.

compiler/qsc_data_structures/src/line_column.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// This function takes a (const sized) array to avoid allocations.
+fn positions_from_utf8_byte_offsets<const N: usize>(
+    encoding: Encoding,
+    contents: &str,
+    sorted_utf8_byte_offsets: [u32; N],
+) -> [Position; N] {
+    // The below example contains characters that are encoded
+    // with more utf-8 code units (bytes) than utf-16 code units, to demonstrate
+    // how code unit offset will differ depending on the chosen encoding.
```
billti, line 129: It's probably also worth calling and out including examples of where there are more utf-16 code units than characters, such as those above the BMP from U+010000 to U+10FFFF (where one code point is encoded as a UTF-16 'surrogate pair' of code units between 0xD800 and 0xDFFF - see https://en.wikipedia.org/wiki/UTF-16 )

billti: (I guess the first char below shows that, but worth a note in the comments maybe)

compiler/qsc_data_structures/src/line_column.rs
billti: Ah! I guess those are 'unit' offsets, not 'byte' offsets, so my mistake. Still, might be clearer to show utf-16 code units as 0x0028, rather than 28, to highlight it is 2 bytes long.

compiler/qsc_data_structures/src/line_column/tests.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+#[test]
+fn grapheme_clusters() {
+    // grapheme clusters, both encodings have multi-unit characters
+    // string       | 𝑓(𝑥⃗) ≔ Σᵢ 𝑥ᵢ 𝑟ᵢ
+    // chars        | 𝑓                 (        𝑥                 ⃗        )                 ≔                 Σ        ᵢ                 𝑥                 ᵢ                 𝑟                 ᵢ
+    // code points  | 1d453             28       1d465             20d7     29       20       2254     20       3a3      1d62     20       1d465             1d62     20       1d45f             1d62
+    // utf-8 units  | f09d9193          28       f09d91a5          e28397   29       20       e28994   20       cea3     e1b5a2   20       f09d91a5          e1b5a2   20       f09d919f          e1b5a2
+    // utf-16 units | d835     dc53     28       d835     dc65     20d7     29       20       2254     20       3a3      1d62     20       d835     dc65     1d62     20       d835     dc5f     1d62
```
billti, line 231: Again, if these are units, should probably make each be 2 bytes long (0028 instead of 28) so that it reads the same as the byte offset/memory layout.

language_service/src/signature_help.rs
```diff
@@ -36,6 +41,29 @@ pub(crate) fn get_signature_help(
...
+                for param in &mut sig.parameters {
+                    let range = Range::from_span(
+                        position_encoding,
+                        &sig.label,
+                        &Span {
+                            lo: param.label.0,
+                            hi: param.label.1,
+                        },
+                    );
+                    assert!(
```
ScottCarda-MS, line 57: Were you running into problems with signature help having line-breaks?

vscode/src/completion.ts
```diff
@@ -73,8 +73,11 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
       item.additionalTextEdits = c.additionalTextEdits?.map((edit) => {
         return new vscode.TextEdit(
           new vscode.Range(
```
ScottCarda-MS, line 75: Can this use the toVsCodeRange function instead?

vscode/src/hover.ts
```diff
@@ -11,23 +11,21 @@ export function createHoverProvider(languageService: ILanguageService) {
...
     const hover = await this.languageService.getHover(
       document.uri.toString(),
-      document.offsetAt(position),
+      position,
     );
     return (
       hover &&
       new vscode.Hover(
         new vscode.MarkdownString(hover.contents),
         new vscode.Range(
```
ScottCarda-MS, line 23: Same here, can this be the toVscodeRange function?

ScottCarda-MS (approved):

language_service/src/signature_help.rs
minestarks: Not at all! I'm just being paranoid here - if we add line breaks to signature help down the line, this assert should remind us the offsets will be broken.

vscode/src/hover.ts
minestarks: Yep, thanks for catching.

language_service/src/protocol.rs
ScottCarda-MS: Discussed offline: there is a risk that if two signatures defined in different namespaces are identical, this hash will merge them together despite the fact that they would have resulted in separate auto-open behavior covered by the `addition_text_edits`. We decided that this should be handled in a separate PR.

compiler/qsc_data_structures/src/line_column.rs
```diff
@@ -0,0 +1,192 @@
...
+use crate::span::Span;
+
+/// A line and column pair that describes a position in a string.
+#[derive(Clone, Copy, PartialEq, Debug, Eq, Hash)]
+pub struct Position {
+    /// Line offset.
+    pub line: u32,
+    /// Column offset.
+    /// When created using [`Encoding::Utf8`], this is the byte offset.
+    /// When created using [`Encoding::Utf16`], this is the code unit offset.
```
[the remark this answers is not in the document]
minestarks: The truth is a bit more subtle - I'm adding a big scare comment to explain.

PR #1039, opened 2024-01-18
Separate Debugger into a distinct struct from Interpreter

minestarks:
The immediate, selfish motivation for this change is that I want to add a parameter to the `Interpreter` constructor (in #1038) that is really only relevant to debugger scenarios, and I didn't want to update every single call to `Interpreter::new`.

But I think this is a good change regardless. Some methods and fields in `Interpreter` are only applicable to debugging, and the rest are only applicable to not-debugging. It's really two separate use cases and two separate sets of data. I wanted to make this distinction clearer.

Very open to feedback and opinions, and suggestions for further cleanup.


npm/test/basics.js
```diff
@@ -697,16 +697,6 @@ async function testCompilerError(useWorker) {
 test("compiler error on run", () => testCompilerError(false));
 test("compiler error on run - worker", () => testCompilerError(true));
 
-test("debug service get breakpoints without service loaded returns empty - web worker", async () => {
```
minestarks, line 700: I removed this because my changes in the `wasm` crate was causing this to fail now. In practice, I didn't think it made sense to support calling `getBreakpoints` before we ever called `loadSource`. I verified that we never actually do this in the product code.

wasm/src/debug_service.rs
```diff
@@ -4,33 +4,26 @@
...
-        Self {
-            interpreter: Interpreter::new(
-                false,
-                SourceMap::default(),
-                PackageType::Lib,
-                Profile::Unrestricted.into(),
-            )
-            .expect("Couldn't create interpreter"),
-        }
+        Self::default()
```
minestarks, line 26: The reason I removed the initialization here is that the `Interpreter` object created here isn't actually good for anything. `get_breakpoints` will return an empty list and any attempt to execute code will throw a "no entrypoint" error. We also never use it in practice. So I thought it's clearer to make it an optional field.

compiler/qsc/src/interpret/stateful/tests.rs
```diff
@@ -1236,16 +1238,11 @@ mod given_interpreter {
             ];
 
             let sources = SourceMap::new(sources, None);
-            let interpreter = Interpreter::new(
-                true,
-                sources,
-                PackageType::Lib,
```
minestarks, line 1242: You'll notice that the test used to pass `PackageType::Lib` here, but the new `Debugger` always passes in `PackageType::Exe` at initialization. I think this change made sense because you can't actually debug a `Lib` in practice (it will always fail with no entrypoint error) , but I'm open to feedback if I missed anything

compiler/qsc/src/interpret/stateful/stepping_tests.rs
```diff
@@ -139,96 +138,76 @@ mod given_interpreter_with_sources {
         #[test]
```
swernli, line 138: I can't comment on the line directly as it is outside the diff, but the module here is called `given_interpreter_with_sources` and that can probably be updated to something simpler like `given_sources` or `with_sources`

minestarks:
@cesarzc would you mind taking a look? I'm touching a couple of folders where you're the owner.

swernli (approved):

billti (approved):

cesarzc (approved):

PR #1041, opened 2024-01-18
add resource estimation samples to playground

sezna:
This PR:
1. Adds the RE samples to the playground
2. Adds `omitFromTests` to `samples.mjs` to denote omission of a sample from the test suite.


samples/samples.mjs
```diff
@@ -23,4 +23,7 @@ export default [
     { title: "Hidden Shift", file: "./algorithms/HiddenShiftNISQ.qs", shots: 1 },
     { title: "Hidden Shift (Advanced)", file: "./algorithms/HiddenShift.qs", shots: 1 },
     { title: "Shor", file: "./algorithms/Shor.qs", shots: 1 },
+    { title: "Dynamics", file: "./estimation/Dynamics.qs", shots: 1 },
+    { title: "Precalculated", file: "./estimation/Precalculated.qs", shots: 1 },
+    { title: "Shor's Resource Estimation", file: "./estimation/ShorRE.qs", shots: 1 },
```
swernli, line 28: since these won't show up in a different subfolder, I think giving them a name that differentiates them would be helpful (like the "(Advanced)" used above):

```suggestion
    { title: "Dynamics (Resource Estimation)", file: "./estimation/Dynamics.qs", shots: 1 },
    { title: "Precalculated (Resource Estimation)", file: "./estimation/Precalculated.qs", shots: 1 },
    { title: "Shor (Resource Estimation)", file: "./estimation/ShorRE.qs", shots: 1 },
```

swernli (approved):

swernli:
I'm not sure how, but it appears this is actually causing a failure in our tests: https://github.com/microsoft/qsharp/actions/runs/7578885466/job/20662715903?pr=1041#step:9:3060. I think the addition of these samples may be causing them to get picked up by one of the npm tests, and then that test tries to simulate the programs that are meant for resource estimation, causing it to get hung.

billti (approved):

PR #1044, opened 2024-01-19
Skip macOS in PR builds, include integration tests

swernli:
This change updates the logic for the build matrix to skip macOS when building for PRs and enables integration tests. This keeps the contentions and slower macOS runners reserved for merge queue to main.

To make this work nicely with branch protection rules, a "Status Check" task is introduced that depends on previous jobs, only executes on failure of dependencies, and always fails. Branch protection treats this job as required, and then PR build will require only those matrix combinations enabled in that context while merge queues will require the full matrix to run and pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,11 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
+        is_pr:
+          - ${{ github.event_name == 'pull_request' }}
+        exclude:
+          - os: macos-latest
+            is_pr: true
 
     runs-on: ${{matrix.os}}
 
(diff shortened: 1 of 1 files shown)


minestarks:
@swernli while you're changing this, think you could enable integration tests for PR builds on ubuntu as well? They don't have to be a _required_ check yet, but it would be nice if they ran automatically now.

billti (approved):

.github/workflows/ci.yml
```diff
@@ -193,3 +202,13 @@ jobs:
               console.log("no change in memory usage detected by benchmark");
             }
         if: ${{ github.base_ref != null }}
+
+  status-check:
+    name: Status Check
+    needs: [format, clippy, web-check, build, integration-tests, runBenchmark]
+    runs-on: ubuntu-latest
```
minestarks, line 209: If we're running the build on Ubuntu & Windows, then it makes sense to run the status check on Ubuntu & Windows as well, right?

minestarks: Or maybe I just didn't read this right - does this job take into account the outcomes from the previous Windows jobs?

swernli: Nope, it only needs to run in one place, because the `needs` section identifies the list of jobs it requires (so waits on them to finish) and the `if: failed()` causes it to only trigger if any of those other jobs failed. If they all succeed, this job doesn't meet it's precondition and is skipped, and from branch protection's perspective a skipped job is treated the same as a succeeded job.

minestarks: Got it, thanks! I'd probably name this this `Fail Workflow` or something else that implies failure, instead of the innocuous sounding "status check" then. When I see "Status Check = skipped" I panic 😜

PR #1045, opened 2024-01-19
Update FIR data structures related to specializations to encode assumptions after running default passes

cesarzc:
This change refactors FIR data structures related to callables' specializations to encode assumptions about what is expected of FIR after running default passes (specifically, the replace qubit allocations pass).

The assumptions are the following:
- Intrinsic callables do not have specializations.
- Specializations for callables that have them are implementations (no generators).

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -24,8 +24,8 @@ use num_bigint::BigInt;
 use output::Receiver;
 use qsc_data_structures::span::Span;
 use qsc_fir::fir::{
-    self, BinOp, BlockId, ExprId, ExprKind, Field, Functor, Global, Lit, LocalItemId, Mutability,
-    NodeId, PackageId, PackageStoreLookup, PatId, PatKind, PrimField, Res, SpecBody, SpecGen,
+    self, BinOp, BlockId, CallableImplementation, ExprId, ExprKind, Field, Functor, Global, Lit,
+    LocalItemId, Mutability, NodeId, PackageId, PackageStoreLookup, PatId, PatKind, PrimField, Res,
     StmtId, StmtKind, StoreItemId, StringComponent, UnOp,
 };
 use qsc_fir::ty::Ty;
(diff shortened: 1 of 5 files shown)


compiler/qsc_fir/src/fir.rs
```diff
@@ -765,6 +759,56 @@ impl Display for CallableDecl {
...
         write!(indent, "\noutput: {}", self.output)?;
         write!(indent, "\nfunctors: {}", self.functors)?;
+        write!(indent, "\nimplementation: {}", self.implementation)?;
+        Ok(())
+    }
+}
+
+/// A callable implementations.
+#[derive(Clone, Debug, PartialEq)]
+pub enum CallableImplementation {
```
swernli, line 769: nitpick: the name is a bit long, so following the patterns of abbreviations we used elsewhere this could be `CallableImpl` instead.

compiler/qsc_fir/src/fir.rs
```diff
@@ -765,6 +759,56 @@ impl Display for CallableDecl {
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// A specialized implementation.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SpecializedImplementation {
```
swernli, line 796: nitpick: same here, we usually abbreviate to "Spec" so something like `SpecImpl` would be shorter and match other patterns.

compiler/qsc_eval/src/lower.rs
```diff
@@ -175,30 +196,23 @@ impl Lowerer {
...
     }
 
     fn lower_spec_decl(&mut self, decl: &hir::SpecDecl) -> fir::SpecDecl {
+        let (controls, block) = if let SpecBody::Impl(pat, block) = &decl.body {
+            let pat_id = pat.as_ref().map(|p| self.lower_spec_decl_pat(p));
+            let block_id = self.lower_block(block);
+            (pat_id, block_id)
+        } else {
+            panic!("if a SpecDecl is some, then it must be an implementation");
+        };
```
swernli, line 210: You can simplify this one into a let-else pattern:

```suggestion
        let SpecBody::Impl(pat, block) = &decl.body else {
            panic!("if a SpecDecl is some, then it must be an implementation");
        };
        let controls = pat.as_ref().map(|p| self.lower_spec_decl_pat(p));
        let block = self.lower_block(block);
```

compiler/qsc_eval/src/lower.rs
```diff
@@ -153,18 +153,39 @@ impl Lowerer {
     }
 
     fn lower_callable_decl(&mut self, decl: &hir::CallableDecl) -> fir::CallableDecl {
+        fn is_instrinsic(spec_body: &SpecBody) -> bool {
+            match spec_body {
+                SpecBody::Gen(spec_gen) => matches!(spec_gen, SpecGen::Intrinsic),
+                SpecBody::Impl(_, _) => false,
+            }
+        }
+
```
swernli, line 162: clippy is going to complain about this unused function now, so it can just be removed.
```suggestion
```

compiler/qsc_fir/src/fir.rs
cesarzc: Renamed.

compiler/qsc_fir/src/fir.rs
cesarzc: Renamed.

compiler/qsc_eval/src/lower.rs
cesarzc: This is very neat! Updated.

compiler/qsc_eval/src/lib.rs
```diff
@@ -1087,35 +1087,33 @@ impl State {
...
-            Spec::Adj => callee.adj.as_ref(),
-            Spec::Ctl => callee.ctl.as_ref(),
-            Spec::CtlAdj => callee.ctl_adj.as_ref(),
-        }
-        .ok_or(Error::MissingSpec(spec.to_string(), callee_span))?
-        .body;
-        match block_body {
-            SpecBody::Impl(input, body_block) => {
+        match &callee.implementation {
+            CallableImplementation::Intrinsic(_, _) => {
```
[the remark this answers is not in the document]
cesarzc: Updated.

swernli (approved):

PR #1047, opened 2024-01-19
[Missing API] Added BigIntAsBoolArray 

filipw:
There currently exist the following std lib functions in the `Microsoft.Quantum.Convert` namespace:

 - `BoolArrayAsInt`
 - `IntAsBoolArray`
 - `BigIntAsBoolArray`

However, `BoolArrayAsBigInt` is missing (and it used to exist in the old QDK too). 
I think it makes sense, for completeness sake, to have it here as well, so this PR introduces it.

I also added a test for `BigIntAsBoolArray` as it did not have an associated test.


filipw:
@sezna I didn't work, I think it's because the promise escapes the try-catch block - https://github.com/microsoft/qsharp/pull/1056

.github/workflows/ci.yml
```diff
@@ -189,7 +189,7 @@ jobs:
           script: |
             if (${{ env.BRANCH_MEASUREMENT }} !== ${{ env.MAIN_MEASUREMENT }}) {
               try {
-                github.rest.issues.createComment({
+                await github.rest.issues.createComment({
```
swernli, line 192: Thanks for fixing this! The warning is showing up as expected now, so this change is exactly what was missing.

PR #1050, opened 2024-01-19
Ignore failed PR comment when PRs are from forks

orpuente-MS:
`github.rest.issues.createComment({ ... })` is throwing an error when running the memory profile on PRs from forks. We want to treat this failure as warning instead of an error.


orpuente-MS:
@microsoft-github-policy-service agree company="Microsoft"

swernli (approved):

PR #1052, opened 2024-01-19
ux improvements for the scatter chart and the summary table

ivanbasov:


npm/ux/resultsTable.tsx
```diff
@@ -194,9 +204,9 @@ export function ResultsTable(props: {
   }
 
   function rowClicked(rowId: string) {
-    if (selectedRow === rowId && props.ensureSelected) return;
+    if (props.selectedRow === rowId && props.ensureSelected) return;
 
-    const newSelectedRow = selectedRow === rowId ? "" : rowId;
+    const newSelectedRow = props.selectedRow === rowId ? "" : rowId;
     setSelectedRow(newSelectedRow);
     props.onRowSelected(newSelectedRow);
```
billti, line 211: Being that setSelectedRow now just calls `props.setSelectedRow()`, do we still need to call `props.onRowSelected()` to notify the parent if we set it by calling the parent?

npm/ux/scatterChart.tsx
```diff
@@ -305,21 +327,18 @@ export function ScatterChart(props: {
...
-                  }
-                  onClick={() =>
-                    drawTooltip(item.label, x, y, seriesIndex, pointIndex, true)
-                  }
+                  onMouseOver={(e) => {
+                    drawTooltip(e.currentTarget, false);
+                    deselectPoint();
+                  }}
+                  onClick={(e) => {
+                    drawTooltip(e.currentTarget, true);
```
billti, line 338: Maybe this call should be inside the SelectPoint method?

billti (approved):

npm/ux/resultsTable.tsx
ivanbasov: Yep! We could keep just one callback out of 2.

PR #1053, opened 2024-01-19
Support setting classical and quantum random seeds

### LINKED ISSUE #1013: Support setting classical and quantum random seeds in Python
Both classical and quantum randomness during simulation could benefit from the ability to set specific seeds before execution. At least initially, this would be exposed through Python, but we can follow up if/how it makes sense to expose this for non-Python scenarios. This becomes valuable when writing tests around Q# code to verify behavior, and predictability and reproducibility are critical.

As a starting point for consideration, here are two options:
 - Expose functions that can be called at any point, like `qsharp.set_classical_seed(x)` and `qsharp.set_quantum_seed(x)` that take effect after they have been called.
 - Add seeds as optional parameters to `qsharp.init()` so that the compilation has to start from a given seed value.

swernli:
This adds new Python APIs `qsharp.set_quantum_seed` and `qsharp.set_classical_seed` that allow users to fix the seed and sequence of random quantum measurements and/or random classical numbers generated from simulating Q#.

Fixes #1013

### DIFF
--- a/pip/qsharp/__init__.py
+++ b/pip/qsharp/__init__.py
@@ -7,6 +7,8 @@ from ._qsharp import (
     run,
     compile,
     estimate,
+    set_quantum_seed,
+    set_classical_seed,
     dump_machine,
 )
 
@@ -27,6 +29,8 @@ __all__ = [
     "init",
     "eval",
     "run",
+    "set_quantum_seed",
+    "set_classical_seed",
     "dump_machine",
     "compile",
     "estimate",
--- a/pip/qsharp/_qsharp.py
+++ b/pip/qsharp/_qsharp.py
@@ -247,7 +247,8 @@ def estimate(
 
 def set_quantum_seed(seed: Optional[int]) -> None:
     """
-    Sets the seed for the quantum random number generator used for measurements.
+    Sets the seed for the random number generator used for quantum measurements.
+    This applies to all Q# code executed, compiled, or estimated.
 
     :param seed: The seed to use for the quantum random number generator.
         If None, the seed will be generated from entropy.
@@ -256,8 +257,9 @@ def set_quantum_seed(seed: Optional[int]) -> None:
 
 def set_classical_seed(seed: Optional[int]) -> None:
     """
-    Sets the seed for the classical random number generator used for standard
-    library random number operations.
+    Sets the seed for the random number generator used for standard
+    library classical random number operations.
+    This applies to all Q# code executed, compiled, or estimated.
 
     :param seed: The seed to use for the classical random number generator.
         If None, the seed will be generated from entropy.
--- a/pip/qsharp/_qsharp.py
+++ b/pip/qsharp/_qsharp.py
@@ -245,6 +245,24 @@ def estimate(
         json.loads(get_interpreter().estimate(entry_expr, json.dumps(params)))
     )
 
+def set_quantum_seed(seed: Optional[int]) -> None:
+    """
+    Sets the seed for the quantum random number generator used for measurements.
+
+    :param seed: The seed to use for the quantum random number generator.
+        If None, the seed will be generated from entropy.
+    """
+    get_interpreter().set_quantum_seed(seed)
+
+def set_classical_seed(seed: Optional[int]) -> None:
+    """
+    Sets the seed for the classical random number generator used for standard
+    library random number operations.
+
+    :param seed: The seed to use for the classical random number generator.
+        If None, the seed will be generated from entropy.
+    """
+    get_interpreter().set_classical_seed(seed)
 
 def dump_machine() -> StateDump:
     """
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -98,6 +98,12 @@ pub struct Interpreter {
     source_package: PackageId,
     /// The default simulator backend.
     sim: SparseSim,
+    /// The quantum seed, if any. This is cached here so that it can be used in calls to
+    /// `run_internal` which use a passed instance of the simulator instead of the one above.
+    quantum_seed: Option<u64>,
+    /// The classical seed, if any. This needs to be passed to the evaluator for use in intrinsic
+    /// calls that produce classical random numbers.
+    classical_seed: Option<u64>,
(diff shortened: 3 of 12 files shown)


tcNickolas (commented):
Looking at the way it would be used in test_qsharp.py, looks perfect for my scenario (fixing seed for randomly generated circuits so that they produce consistent results). An example of a scenario where I would like to use this is 
https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-random-circuit.ipynb - with fixed seed it would always produce exactly the same resource estimates

swernli:
> Looking at the way it would be used in test_qsharp.py, looks perfect for my scenario (fixing seed for randomly generated circuits so that they produce consistent results). An example of a scenario where I would like to use this is https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-random-circuit.ipynb - with fixed seed it would always produce exactly the same resource estimates

@tcNickolas, would you want me to update that sample as part of this PR? I can add the right call and a comment to note why that is being done and how to undo it.

pip/qsharp/_qsharp.py
```diff
@@ -245,6 +245,24 @@ def estimate(
...
+
+def set_classical_seed(seed: Optional[int]) -> None:
+    """
+    Sets the seed for the classical random number generator used for standard
+    library random number operations.
+
+    :param seed: The seed to use for the classical random number generator.
+        If None, the seed will be generated from entropy.
+    """
+    get_interpreter().set_classical_seed(seed)
```
minestarks, line 265: Will users want to control these separately? To keep the api simple I wonder if we should just expose one `set_random_seed`. 

1. Is the distinction really clear to users which library operations will use the classical rng and which end up using the quantum rng? (Honest question, perhaps this is a concept that's clear to Quantum devs, but I feel like it's pretty opaque)
2. If you're touching this, it's probably because you want your program to be fully deterministic, so I feel like users would just get into the habit of calling _both_ every time to be safe.

minestarks (approved):

pip/qsharp/__init__.py
```diff
@@ -27,6 +29,8 @@
     "init",
     "eval",
     "run",
+    "set_quantum_seed",
+    "set_classical_seed",
```
minestarks, line 33: I have no suggestions on how to improve this, but I feel like it's really unclear whether these apply to any/all of `qsharp.eval`, `qsharp.run` or `qsharp.compile` (QIR generation). How can we make it clearer to the user exactly what they're changing?

tcNickolas:
If we modify the sample right away, will it stop working with the currently released QDK version? If that's the case, let's keep it as is for now, and modify it once this change is shipped to the released version. If not, let's change it now

pip/qsharp/__init__.py
tcNickolas: (Also, qsharp.estimate - quantum seed would not be relevant but classical seed would impact the results)

pip/qsharp/_qsharp.py
swernli: I really went back and forth debating on that one too. Every time I convinced myself of one side, I'd eventually flip to the other. As you say, it seems very likely that users will always set both because their interest is running the program fully deterministically. But then again, a user who is running many shots of some combined quantum-classical algorithm might want to see the behavior of the quantum with fixed classical randomness, so that makes me want to keep them separate. But in that case, couldn't the user just replace calls to `DrawRandomInt` and/or `DrawRandomDouble` with fixed values? But that is hard to do if they come from a library, which we know we eventually want to support external dependencies... I think I just need someone else to weigh in with a strong preference.

@minestarks and @tcNickolas, do either of you feel strongly one way or the other?

pip/qsharp/__init__.py
swernli: All great points, and the answer is that the setting is respected all the scenarios except debugging. But as you point out, neither `compile` for QIR generation nor `estimate` for resource estimation do any quantum simulation, so even though the quantum setting is passed through in those cases only the classical setting changes behavior.

pip/qsharp/_qsharp.py
tcNickolas: In my use cases, classical random numbers affect the structure of the circuit (which gates are applied, with which parameters, etc.), while quantum seed affects the simulation results. 

I'm more interested in the first one, since we don't usually need to reproduce the exact measurement outcomes, while we might want to reproduce the exact circuit we estimate on subsequent runs. (The user doesn't replace calls to DrawRandomInt with fixed values because they're generating 1000 random gates in a for loop, and replacing calls with fixed numbers means effectively generating a hardcoded Q# circuit using non-Q# means, like we did for https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-hardcoded-circuit.ipynb - it's possible but not really readable, or even easy to scroll!)

But I can definitely imagine wanting to get different simulation results on each run for that fixed circuit, for example, if I want to gather some statistics about the outcomes. If somebody wants to generate a fixed circuit but then gets deterministic measurement results each time, they might start questioning the fundamental randomness of quantum computing, we don't want that to happen :-)

So, I'd say it's best to keep them separate - it doesn't look like that's too much work (and it's already done anyways :-)), and it's better to give the user more flexibility than less.

pip/qsharp/__init__.py
swernli: To put it another way, since debugging can't be invoked through Python and these are only exposed through Python, once you make the call `qsharp.set_classical_seed(val)` it be respected for all calls to `qsharp` after that until you either reinitialize or set the seed again. The same is true for `set_quantum_seed` with the caveat that it doesn't affect those calls that don't perform quantum simulation, namely `compile` and `estimate`.

swernli:
> If we modify the sample right away, will it stop working with the currently released QDK version? If that's the case, let's keep it as is for now, and modify it once this change is shipped to the released version. If not, let's change it now

Good point, I don't want to break existing users browsing the samples. I'll leave the sample as is.

pip/qsharp/__init__.py
minestarks: I can wrap my head around it now, thanks - but maybe it's worth explaining in the doc comments.

PR #1055, opened 2024-01-20
Add dump_operation to Python API

### LINKED ISSUE #1009: Add dump_operation to Python API
I'd love to be able to write unit tests for Q# code that implements unitary operations by getting the matrix of the transformation from Python.

Something similar to `dump_machine`, where we call a method from Python and it returns a 2D array of coefficients of the matrix.

I could implement this myself based on `dump_machine` functionality, but that requires writing extra wrapper code in Q# and parsing code in Python, and this functionality is frequent enough to be implemented in QDK itself.

**Additional context**
Classic QDK used to have `DumpOperation` that would print the matrix of an operation from Q# (not necessarily a unitary, it could use measurements as well, then the matrix would not be unitary). The tests were written using `AssertOperationsEqualInPlace` and `AssertOperationsEqualReferenced`. This can be done using Choi–Jamiołkowski isomorphism (see https://devblogs.microsoft.com/qsharp/inside-the-quantum-katas-part-2-testing-quantum-programs/ for a small example). Since we're moving to using Python for tests in Modern QDK, it makes sense to expose this functionality in Python.

swernli:
This adds a new `dump_operation` to Python that when given an operation expression to test and a number of qubits will execute that operation in isolation and produce an output matrix (array of arrays) of complex numbers representing the operation.

Fixes #1009

### DIFF
--- a/pip/qsharp/_native.pyi
+++ b/pip/qsharp/_native.pyi
@@ -114,7 +114,7 @@ class Interpreter:
         ...
     def dump_operation(self, operation: str, num_qubits: int) -> StateDump:
         """
-        Returns the sparse state vector of the simulator describing the affect of
+        Returns the sparse state vector of the simulator describing the effect of
         the given operation on the specified number of qubits. Internally, twice
         this number of qubits are needed to perform the analysis.
 
--- a/pip/qsharp/_native.pyi
+++ b/pip/qsharp/_native.pyi
@@ -112,18 +112,6 @@ class Interpreter:
         :returns: The state of the simulator.
         """
         ...
-    def dump_operation(self, operation: str, num_qubits: int) -> StateDump:
-        """
-        Returns the sparse state vector of the simulator describing the effect of
-        the given operation on the specified number of qubits. Internally, twice
-        this number of qubits are needed to perform the analysis.
-
-        :param operation: The operation to run.
-        :param num_qubits: The number of qubits needed to execute the operation.
-
-        :returns: The state of the simulator after the given operation.
-        """
-        ...
 
 class Result(Enum):
     """
@@ -152,6 +140,7 @@ class Output:
     def __repr__(self) -> str: ...
     def __str__(self) -> str: ...
     def _repr_html_(self) -> str: ...
+    def state_dump(self) -> Optional[StateDump]: ...
 
 class StateDump:
     """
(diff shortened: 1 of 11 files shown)


pip/qsharp/_native.pyi
```diff
@@ -112,6 +112,18 @@ class Interpreter:
         :returns: The state of the simulator.
         """
         ...
+    def dump_operation(self, operation: str, num_qubits: int) -> StateDump:
+        """
+        Returns the sparse state vector of the simulator describing the affect of
```
minestarks, line 117:
```suggestion
        Returns the sparse state vector of the simulator describing the effect of
```

I think?

pip/src/interpreter.rs
```diff
@@ -174,6 +174,40 @@ impl Interpreter {
         ))
     }
 
+    /// Dumps the quantum state corresponding to the application of a given operation.
```
minestarks, line 177: I don't think this function belongs here. This file is supposed to be just interop "glue" and the code gets confusing when we start putting actual implementations in these sorts of files. Instead it should either go up or down a layer. Can we implement this in Python (in `_qsharp.py`) or in the native `Interpreter` and plumb it up?

minestarks (changes requested):

pip/src/interpreter.rs
swernli: Fair point, I can move this down. That is what we would need to do anyway if we want to make this functionality available somewhere other than Python.

swernli: I moved this down as you pointed out, and added unit tests at that layer.

minestarks (dismissed):
Looks very nice!

minestarks (approved):
Even nicer! Let's discuss separately whether we want to take it into the qsharp module and/or a submodule.

pip/qsharp/_qsharp.py
```diff
@@ -4,6 +4,7 @@
 from ._native import Interpreter, TargetProfile, StateDump, QSharpError, Output
 from typing import Any, Callable, Dict, Optional, TypedDict, Union, List
 from .estimator._estimator import EstimatorResult, EstimatorParams
+import math
```
swernli, line 7: whoops, left this behind. Not needed here anymore.

PR #1056, opened 2024-01-22
await createComment() call in CI

filipw:
The catch block is never reached, because the function  is asynchronous


swernli:
Could you go ahead and merge this change to your other branch where the check is failing? The check only attempts to write a comment if there is a change in memory allocations, which your other PR does have, so should be able to show the fix working in that context.

PR #1058, opened 2024-01-22
Factoring algorithm for resource estimation

msoeken:
This adds a factoring algorithm based on work from Ekara-Hastad with optimizations to modular arithmetic subroutines based on work from Gidney (see https://arxiv.org/abs/1905.09749)


tcNickolas:
I would add this sample to the readme at https://github.com/microsoft/qsharp/tree/main/samples/estimation, so that we document all samples uniformly.

Since this is a two-file sample (Jupyter notebook requires the .qs file to work), do we want to move both files to a separate subfolder to emphasize this? Not sure if the intent is to use them together only, or to allow using Q# file on its own as well - if it's the latter, we might want to keep them in this folder and document them separately in the README

msoeken:
> I would add this sample to the readme at https://github.com/microsoft/qsharp/tree/main/samples/estimation, so that we document all samples uniformly.
> 
> Since this is a two-file sample (Jupyter notebook requires the .qs file to work), do we want to move both files to a separate subfolder to emphasize this? Not sure if the intent is to use them together only, or to allow using Q# file on its own as well - if it's the latter, we might want to keep them in this folder and document them separately in the README

Thanks, I addresses this comment in a recent commit.

swernli (approved):

PR #1059, opened 2024-01-22
Traits for EC and physical qubits

msoeken:
This creates traits for error correction and physical qubits.


resource_estimator/src/estimates/modeling/physical_qubit.rs
```diff
@@ -13,6 +13,64 @@ use super::super::{
 };
 use serde::{de::Error, Deserialize, Serialize};
 
+pub trait TPhysicalQubit {
```
ivanbasov, line 16: T is for trait. Correct?
Do we already follow this pattern?

msoeken: No, this is not final. A good name would be `PhysicalQubit`, but we are already using it. I still want to change the names.

resource_estimator/src/estimates/modeling/physical_qubit.rs
```diff
@@ -13,6 +13,64 @@ use super::super::{
...
+    fn instruction_set(&self) -> PhysicalInstructionSet;
+    fn t_gate_error_rate(&self) -> f64;
```
ivanbasov, line 18: Just to confirm that we expect all new qubits to support those properties, and we do not need more formulas. 

For example, currently we use those error rates for t-factories construction, e.g. in custom formulas. Should we consider a wider (flexible?) set of custom formulas where we would consume not 3 but more parameters in polynomials?

msoeken: Possibly yes, for now I was just extracting the interface based on the current types and made small adjustments on the ErrorCorrection trait.

resource_estimator/src/estimates/modeling/physical_qubit.rs
msoeken: In fact, we don't need a trait on the physical qubits, so I removed this one. We require no properties on the physical qubit for estimation, as it is all derived from the ErrorCorrection trait.

resource_estimator/src/estimates/modeling/physical_qubit.rs
msoeken: All these properties are part of the implementation of the error correction, there is no more trait for physical qubits.

resource_estimator/src/estimates.rs
```diff
@@ -79,21 +82,28 @@ fn estimate_single(
...
+        TFactoryBuilder::default(),
+        logical_resources,
+        partitioning,
+    );
     if let Some(logical_depth_factor) = job_params.constraints().logical_depth_factor {
         estimation.set_logical_depth_factor(logical_depth_factor);
     }
     if let Some(max_t_factories) = job_params.constraints().max_t_factories {
-        estimation.set_max_t_factories(max_t_factories);
+        estimation.set_max_factories(max_t_factories);
```
ivanbasov, line 96: to confirm that this is itended

resource_estimator/src/estimates/data/result.rs
```diff
@@ -184,12 +184,12 @@ fn create_physical_resource_counts_breakdown<L: Overhead + Clone>(
         num_tstates: result
             .layout_overhead()
             .num_tstates(num_ts_per_rotation.unwrap_or_default()),
-        num_tfactories: result.num_tfactories(),
-        num_tfactory_runs: result.num_tfactory_runs(),
-        physical_qubits_for_tfactories: result.physical_qubits_for_tfactories(),
+        num_tfactories: result.num_factories(),
+        num_tfactory_runs: result.num_factory_runs(),
+        physical_qubits_for_tfactories: result.physical_qubits_for_factories(),
```
ivanbasov, line 189: same here. do we want to remove "t" from tfactories?

resource_estimator/src/estimates/modeling/logical_qubit.rs
```diff
@@ -30,16 +27,12 @@ pub struct LogicalQubit<P: TPhysicalQubit> {
     logical_error_rate: f64,
 }
 
-impl<P: TPhysicalQubit> LogicalQubit<P> {
+impl<P> LogicalQubit<P> {
     pub fn new(
-        ftp: &impl ErrorCorrection<PhysicalQubit = P>,
+        ftp: &impl ErrorCorrection<Qubit = P>,
```
ivanbasov, line 32: to confirm that we want to use Qubit for PhysicalQubit and omit Physical

resource_estimator/src/estimates/stages/physical_estimation/tests.rs
```diff
@@ -264,13 +296,12 @@ pub fn test_hubbard_e2e_measurement_based() -> super::super::super::Result<()> {
     validate_result_invariants(&result);
 
     let output_t_error_rate = result
-        .required_logical_tstate_error_rate()
+        .required_logical_magic_state_error_rate()
```
ivanbasov, line 299: should it be magic state, factory or tfactory?

resource_estimator/src/estimates/data/result.rs
msoeken: Same comment as above.

resource_estimator/src/estimates.rs
msoeken: Yes, the estimator is generic over `FactoryBuilder` which just considers that there are some magic states. The system is based on T states.

resource_estimator/src/estimates/modeling/logical_qubit.rs
msoeken: Yes, it's more generic. I can envision situations in which error correction is on top of logical qubits.

resource_estimator/src/estimates/stages/physical_estimation/tests.rs
msoeken: magic state, because the output rate of a factory can be different, if it produces multiple states.

resource_estimator/src/estimates.rs
ivanbasov: I am seeing changes in naming. I am not against them but would like:
- them to be consistent
- understand what would be the standard naming

There are two t_factories on lines 97-98 and one factory without 't'.

There are magic states in code below instead of t-states. If we want to start a renaming, let us agree and apply it everywhere.

msoeken: You're right. I missed quite some instances for renaming and will fix this soon.

I suggest to use `t_states` in the specific system-related code, and `magic_states` in the generic estimator-related code.

ivanbasov (approved):

PR #1061, opened 2024-01-23
Better visual alignment in Debug Console Dump Machine output

minestarks:
Compute variable column width based on the values so the columns align better.

Before: 

Screenshot 2024-01-23 091251

After:

image


vscode/src/debugger/output.ts
```diff
@@ -20,25 +20,65 @@ export function createDebugConsoleEventTarget(out: (message: string) => void) {
...
+
+    function formatPhase(real: number, imag: number) {
+      const phase = Math.atan2(imag, real);
+      return phase.toFixed(4);
+    }
+
+    function column(header: string, values: string[]) {
+      // The length of the longest value, or the header length, whichever is longer
+      const columnWidth = values.reduce((max, value) => {
+        if (value.length > max) {
```
DmitryVasilevsky, line 36: I'm wondering, is it necessary to compute max length? Could we set max length upfront and force fit everything into that max length? (Not suggesting this would be a better solution, just asking).

minestarks: It's a good suggestion, and it would probably make the code simpler.

The reason for the computed column width is that Basis State labels can exceed the maximum width we hardcode if we have a lot of qubits. Amplitude and Phase are also unbounded, though in practice, I'm not sure if they'd ever exceed 100... 

I don't know. I may have overcomplicated this for the sake of perfection, perhaps it's okay to have a few misaligned columns if your values get really large.

DmitryVasilevsky: If the values get large, you can use scientific notation (Although amplitude magnitude shoulnd't be more that 1 and phase should be in -pi..pi). Title length could be also checked upfront. Number of qubits is also known. If you set things upfront, you won't need to iterate through all the data. Not sure if it has any benefits here - we still have to display all the data... or not? Maybe if the data is too large, we will truncate it someday. Anyway, it's up to you.

idavis (approved):

vscode/src/debugger/output.ts
minestarks: I took your suggestion and simplified quite a bit. Thanks!

idavis (approved):

PR #1067, opened 2024-01-24
Samples for resource estimation of Lattice Model simulation circuits

aarthims:
This change adds three Python+Q# Jupyter notebooks that perform resource estimation for simulating (i) a 2D Ising Model; (ii) a 2D Heisenberg Model; and (iii) a 2D Hubbard Model. 

The change also removes the Dynamics.qs file which is outdated with respect to the new 2D Ising model notebook.


tcNickolas:
Dynamics.qs was used by https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-frontier-widgets.ipynb so we want to also figure out what to do with that

samples/estimation/estimation-ising-2D.ipynb
```diff
@@ -0,0 +1,509 @@
...
+        "version": 3
+      },
+      "file_extension": ".py",
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.11.7"
+    },
+    "microsoft": {
```
tcNickolas, line 495: This section is possibly not needed - I think it appears in notebooks that were downloaded from hosted notebooks, but now it's pointing to an archived repo. I deleted it from my notebooks

aarthims:
With newest commits, the change replaces the old Dynamics.qs file with updated code to execute the same sample.

samples/estimation/estimation-ising-2D.ipynb
aarthims: Done. Removed the reference to hosted notebooks.

billti:
Dynamics was a standalone Q# sample that was included in our playground and completion list (via https://github.com/microsoft/qsharp/blob/main/samples/samples.mjs#L26 - which would also have to be updated).

Generally we like having standalone Q# samples, as they work well in the browser and directly in VS Code. Once things are moved into a Notebook they don't run on web, don't support debugging, etc. so we'd need a strong reason to remove stand alone Q# samples in favor of notebooks

swernli (approved):

PR #1068, opened 2024-01-24
Pure state preparation

DmitryVasilevsky:
Pure state preparation: `PreparePureStateD` and `ApproximatelyPreparePureStateCP` and tests.
For simple scenarios use `PreparePureStateD`. It takes real coefficients (including negative ones). 
For more complex scenarios use `ApproximatelyPreparePureStateCP`, which takes complex polar coefficients and tolerance.
`PreparePureStateD` is almost the same as the legacy `PrepareArbitraryStateD`, but properly handles negative coefficients, hence the different name. Also, it is uses words "pure state" to distinguish it from potential future addition of the mixed state preparation.
Note that state prepared may not have the same coefficients as ones provided as global phase may be different.
This PR also adds `IsRangeEmpty` to core namespace and tests for range functions from core namespace.
This PR also includes a way to check quantum state directly by inspecting DumpMachine output (courtesy of @swernli).


library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,446 @@
...
+    /// ## coefficients
+    /// Array of up to 2ⁿ real coefficients. The j-th coefficient
+    /// indexes the number state |j⟩ encoded in little-endian format.
+    ///
+    /// ## qubits
+    /// Qubit register encoding number states in little-endian format. This is
+    /// expected to be initialized in the computational basis state |0...0⟩.
+    ///
+    /// # Remarks
+    /// `coefficients` will be normalized and padded with
```
cesarzc, line 36: I think the pattern of padding arrays if they don't have the expected number of elements can lead to inadvertently using the operation incorrectly. For this reason, I think it is more robust if the operation fails when the number of coefficients is not 2^n.

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,457 @@
...
+            // For tolerance 0.0 it is always applied due to >= comparison.
+            if AbsD(coefficient) >= tolerance {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    internal function AnyOutsideToleranceCP(tolerance : Double, coefficients : ComplexPolar[]) : Bool {
+        for coefficient in coefficients {
```
cesarzc, line 448: It seems like the note/comment in `AnyOutsideToleranceD` also applies to this function so it might be worth also including it here if that's the case.

library/std/unstable_state_preparation.qs
DmitryVasilevsky: I understand the concern and yes, this is a valid concern. However, it looks like we need to support this functionality. (1) It used to be like that in the old QDK (2) You may want to prepare a sate on only the first five "numbers". Then you won't need to provide coefficients for the other three. (3) It's kind-of consistent with other state preparation routines that aren't ported yet. For example, there's a non-trivial operation that prepares uniform superposition of the first N basis vectors. If N was always 2^n, we wouldn't need that at all, as a simple H and CNOTs would be sufficient.
Given that I'd say we keep the ability to provide only first few coefficients omitting trailing 0s.

library/std/unstable_state_preparation.qs
DmitryVasilevsky: This is not entirely correct. It is used as a recursion termination condition, but only in addition to the exhaustion of array elements. Another caveat is that this condition can technically become false even if tolerance is 0.0 (unlike the other one).
I added that comment to the other function because it was quite unexpected at first glance. (if that other function was used as the only recursion termination condition with tolerance 0.0, the recursion would never stop). This doesn't look like it's totally out of the ordinary so I'm not sure what comment would be appropriate here.

cesarzc (approved):

tcNickolas (changes requested):

library/src/tests/resources/state_preparation.qs
```diff
@@ -0,0 +1,137 @@
...
+            // Test missing coefficients
+            [1.0986553, 0.359005, 0.465689, -0.467395, 0.419893, 0.118445]
+        ];
+
+        for coefficients in testCases {
+            let L = Length(coefficients);
+            let N = Ceiling(Log(IntAsDouble(L))/LogOf2() - 0.001);
+            use q = Qubit[N];
+            PreparePureStateD(coefficients, q);
+            // NOTE: We cannot check that the state is actually correct.
```
swernli, line 100: We actually can check the state from Rust using DumpMachine and the registered receiver. We can make a new test utility to help with this. Let’s sync up on that one, I think it will be handy going forward, and might even be something we can adopt for the intrinsic tests.

library/src/tests/resources/state_preparation.qs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    operation CheckEndianness(attempts: Int, number: Int, qs: Qubit[]): Unit {
+        let n = Length(qs);
+        for attempt in 1..attempts {
+            mutable c = Repeated(0.0, 2^n);
+            set c w/= number <- 1.0;
+            PreparePureStateD(c, qs);
+            let measured = MeasureInteger(Reversed(qs));
+            if number == measured {
```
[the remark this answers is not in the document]
tcNickolas: (If we end up adopting Stefan's suggestion on tests, we can replace iteration and measurement with that approach to validate the state directly)

library/src/tests/resources/state_preparation.qs
```diff
@@ -0,0 +1,137 @@
...
+        use q = Qubit();
+        PreparePureStateD([Sqrt(0.5), -Sqrt(0.5)], [q]);
+        // Ucompute minus
+        H(q);
+        X(q);
+        if ( not CheckZero(q) ) {
+            fail "|-> preparation failed.";
+        }
+    }
+
```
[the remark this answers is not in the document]
tcNickolas: (If we end up adopting Stefan's suggestion on tests, this scenario will be covered by the test with random amplitudes, so a separate test won't be necessary)

library/src/tests/resources/state_preparation.qs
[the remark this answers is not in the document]
DmitryVasilevsky: Having stable tests is more favorable than having tests that detect all failures. That is if we have to choose. Like Stefan says it's about time to add a bit of infrastructure so that we don't have to choose.

library/src/tests/resources/state_preparation.qs
DmitryVasilevsky: Yes, it's about time!

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+                H(target);
+            } apply {
+                ApproximatelyMultiplexPauli(tolerance, coefficients, PauliZ, control, target);
+            }
+        } elif pauli == PauliY {
+            within {
+                Adjoint S(target);
+            } apply {
+                ApproximatelyMultiplexPauli(tolerance, coefficients, PauliX, control, target);
+            }
```
msoeken, line 220: This can be made more concise with `M.Q.Intrinsic.MapPauli`.

library/src/tests/resources/state_preparation.qs
swernli: I tried out a couple alternatives for how to verify dump output and finally found one I like best. You can check out the change here: https://github.com/microsoft/qsharp/commit/5a31cc7b5d46adddcd886a875a930efee1250b33. Feel free to cherry-pick that commit if you want to get it into your branch or we can discuss more if you want to go over it before adopting the changes.

msoeken: I think it's a great utility to be able to verify the quantum state in tests cases. I suggest to consider numerical robustness. Some internal changes to the operation or to other operations that it's referring to could add some negligible numerical imprecision that changes the output. Also it could add a global phase, which doesn't change the quantum state, but the way it's going to be displayed. Further, there may be test cases in which one wants to compare a subset of amplitudes.

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+    /// $$
+    /// \begin{align}
+    ///     U \ket{0\cdots 0} = \ket{\psi} = \frac{\sum_{j=0}^{2^n-1}\alpha_j \ket{j}}{\sqrt{\sum_{j=0}^{2^n-1}|\alpha_j|^2}}.
+    /// \end{align}
+    /// $$
+    ///
+    /// # Input
+    /// ## coefficients
+    /// Array of up to 2ⁿ real coefficients. The j-th coefficient
+    /// indexes the number state |j⟩ encoded in little-endian format.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Yes, with the change of endianness recently introduced this should be big-endian.

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+    /// `coefficients` will be normalized and padded with
+    /// elements 𝑎ⱼ = 0.0 if fewer than 2ⁿ are specified.
+    ///
+    /// # Example
+    /// The following snippet prepares the quantum state |𝜓⟩=√(1/8)|0⟩+√(7/8)|2⟩
+    /// in the qubit register `qubitsLE`.
+    /// ```qsharp
+    /// let amplitudes = [Sqrt(0.125), 0.0, Sqrt(0.875), 0.0];
+    /// use qubits = Qubit[2];
+    /// let qubitsLE = LittleEndian(qubits);
```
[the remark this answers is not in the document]
DmitryVasilevsky: Updated and added test for this specific case.

library/src/tests/resources/state_preparation.qs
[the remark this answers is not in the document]
DmitryVasilevsky: We used approach suggested by Stefan.

library/src/tests/resources/state_preparation.qs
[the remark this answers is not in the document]
DmitryVasilevsky: We are inspecting the actual state now.

library/src/tests/resources/state_preparation.qs
DmitryVasilevsky: We are inspecting the state now as displayed by DumpMachine. Its precision is only about 10^-4, but this should be enough for the test cases we currently have.

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    internal function AnyOutsideToleranceCP(tolerance : Double, coefficients : ComplexPolar[]) : Bool {
+        for coefficient in coefficients {
+            if AbsComplexPolar(coefficient) > tolerance {
+                return true;
+            }
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
DmitryVasilevsky: Updated.

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+    internal function AnyOutsideToleranceD(tolerance : Double, coefficients : Double[]) : Bool {
+        for coefficient in coefficients {
+            // NOTE: This function is not used in a recursion termination condition
+            // only to determine if the multiplex step needs to be applied.
+            // For tolerance 0.0 it is always applied due to >= comparison.
+            if AbsD(coefficient) >= tolerance {
+                return true;
+            }
+        }
+        return false;
```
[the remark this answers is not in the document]
DmitryVasilevsky: Updated.

library/std/unstable_state_preparation.qs
DmitryVasilevsky: Postponed for now. Filed https://github.com/microsoft/qsharp/issues/1104

library/std/unstable_state_preparation.qs
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        controlled (controlRegister, ...) {
+            // pad coefficients length to a power of 2.
+            let coefficientsPadded = Padded(2 ^ (Length(control) + 1), 0.0, Padded(-2 ^ Length(control), 0.0, coefficients));
+            let (coefficients0, coefficients1) = MultiplexZCoefficients(coefficientsPadded);
+            ApproximatelyMultiplexZ(tolerance,coefficients0, control, target);
+            if AnyOutsideToleranceD(tolerance,coefficients1) {
+                within {
+                    Controlled X(controlRegister, target);
```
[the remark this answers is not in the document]
DmitryVasilevsky: Postponed for now. Filed https://github.com/microsoft/qsharp/issues/1104

tcNickolas (approved):
I would like to understand for myself why, as soon as we have one of the real amplitudes negative, there appears a global phase in the DumpMachine output... But I'm counting on the fact that the state prep uses correct gates and that it's a quirk of the simulator, which should be harmless, if confusing

One small doc improvement suggestion, other than that, looks good!

swernli (approved):

minestarks (approved):

PR #1069, opened 2024-01-24
hover and selected tooltips for RE scatter chart; EstimatesPanel for the Jupyter notebook

ivanbasov:


npm/ux/utils.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function findParentByClassName(
```
billti, line 4: You don't need this helper. Use <https://developer.mozilla.org/en-US/docs/Web/API/Element/closest> .  See resultsTable.tsx for examples, which uses it extensively.

npm/ux/scatterChart.tsx
```diff
@@ -202,46 +196,93 @@ export function ScatterChart(props: {
...
+  useEffect(() => {
+    chart = scatterChartContainerRef.current as Element | null;
+    updateCoordinates();
+  });
 
   return (
     <div style="display: flex; flex-wrap: wrap; margin-top: 8px;">
       <div class="chart-container">
         <svg
+          class="qs-scatterChart-svg"
```
billti, line 257: I don't see this class defined in the CSS. Is it used for any styling or just a marker?

minestarks:
A suggestion: for UI changes, I find it really helpful to see a screenshot of the changes in the PR description, it gives context when doing the code review. Thanks!

ivanbasov:
> A suggestion: for UI changes, I find it really helpful to see a screenshot of the changes in the PR description, it gives context when doing the code review. Thanks!

image

Thanks, @minestarks ! Now we support two tooltips: one for the selected point which als corrseponds the selected row in the table and data shown on the space diagram and report details below. And the second tooltip follows the cursor.

ivanbasov:
Also added the extended experience with 4 controls together to Jupyter notebook

vscode/src/webview/webview.tsx
```diff
@@ -6,9 +6,8 @@
 const vscodeApi = acquireVsCodeApi();
 
 import { render } from "preact";
-import { Histogram, type ReData } from "qsharp-lang/ux";
+import { EstimatesPanel, Histogram, type ReData } from "qsharp-lang/ux";
 import { HelpPage } from "./help";
-import { RePage } from "./rePage";
```
billti, line 11: I'm confused here, as this PR doesn't show rePage being deleted, but when I check out your PR it's gone, but it's still in main (https://github.com/microsoft/qsharp/blob/main/vscode/src/webview/rePage.tsx). Weird.

billti: Ah! I see. It got renamed and moved into npm/ux/estimatesPanel.tsx. False alarm! Thanks.

npm/ux/estimatesOverview.tsx
```diff
@@ -179,60 +193,64 @@ export function EstimatesOverview(props: {
...
-        <ScatterChart
-          xAxis={xAxis}
-          yAxis={yAxis}
-          data={props.estimatesData.map((dataItem, index) =>
-            reDataToRowScatter(dataItem, colormap[index]),
-          )}
-          onPointSelected={onPointSelected}
-        />
-      </details>
+      <g id="qs-estimatesOverview">
```
billti, line 223: `g` is an SVG element, so you shouldn't be using it in an HTML context such as here. You should probably just make the parent `<></>` element be a `div` with the id you want. (Though if using it for selection, you should use a className rather than id to find it in case there are multiple on the page)

widgets/js/index.tsx
```diff
@@ -118,6 +128,64 @@ function renderEstimatesOverview({ model, el }: RenderArgs) {
...
+        calculating={false}
+        onRowDeleted={onRowDeleted}
+      ></EstimatesPanel>,
+      el,
+    );
+  };
+
+  onChange();
+  model.on("change:estimates", onChange);
+  model.on("change:colors", onChange);
```
billti, line 166: Should probably add runNames here for completeness.

billti (approved):

cesarzc (approved):
Approving the samples changes.

Issue #1023, opened 2024-01-16
Expand automated testing for samples


swernli:
As we add more samples, both things like multi-file with qsharp.json and notebooks, we need to ensure those are getting automatically tested. Notebook validation in particular needs some investigation.

PR #1070, opened 2024-01-24
Consolidate samples, run notebooks in build

swernli:
This change consolidates our samples under the top level samples folder by moving the three notebooks that were under pip/samples.

It also adds a new build step controlled by `--notebooks` that uses a virtual python environment to run the notebooks and verify the cells execute successfully. It does not verify the contents of the output, that can come in a later PR.


swernli:
This helps address part of #1023

build.py
```diff
@@ -32,6 +32,7 @@
 parser.add_argument("--npm", action="store_true", help="Build the npm package")
 parser.add_argument("--play", action="store_true", help="Build the web playground")
 parser.add_argument("--samples", action="store_true", help="Compile the Q# samples")
+parser.add_argument("--notebooks", action="store_true", help="Run Jupyter notebook samples")
```
minestarks, line 35: Any thoughts on putting this under the `--integration-tests` step? That step is excluded by default (i.e. if you run `build.py` with no arguments) but runs in CI, as you already know.

(If you do do this, you probably want to update ci.yml as well, to build the pip package in addition to vscode for the integration test run.)

build.py
```diff
@@ -341,6 +344,91 @@ def step_end():
...
+            venv.main([venv_dir])
+
+        # Check if .venv/bin/python exists, otherwise use .venv/Scripts/python.exe (Windows)
+        python_bin = os.path.join(venv_dir, "bin", "python")
+        if not os.path.exists(python_bin):
+            python_bin = os.path.join(venv_dir, "Scripts", "python.exe")
+        print(f"Using python from {python_bin}")
+    else:
+        # Already in a virtual environment, use current Python
+        python_bin = sys.executable
```
minestarks, line 374: This is now repeated three times, maybe we should pull it into a helper function?

build.py
```diff
@@ -341,6 +344,91 @@ def step_end():
...
+    # copy the process env vars
+    pip_env: dict[str, str] = os.environ.copy()
+
+    # Install the qsharp package
+    pip_install_args = [
+        python_bin,
+        "-m",
+        "pip",
+        "install",
+        "-e",
```
minestarks, line 385: This will rebuild the Rust module from scratch. Also the install doesn't need to be editable. Elsewhere in the file (when we run the pip package tests) we install the package from the already built wheel. I'd recommend doing that instead. (Same for the widgets package).

minestarks (changes requested):
Just some thoughts about structuring the build scripts.

build.py
```diff
@@ -341,6 +344,91 @@ def step_end():
         subprocess.run((cargo_args + ["--", "--qsharp-json", project]), check=True, text=True, cwd=root_dir)
     step_end()
 
+if build_notebooks:
+    step_start("Running notebook samples")
+    # Find all notebooks in the samples directory. Skip the basic sample and the azure submission sample, since those won't run
+    # nicely in automation.
+    notebook_files = [os.path.join(dp, f) for dp, _, filenames in os.walk(samples_src) for f in filenames if f.endswith(".ipynb")
+                      and not (f.startswith("sample.") or f.startswith("azure_submission."))]
```
minestarks, line 352: Do we trust ourselves not to check in sample notebooks with arbitrary web requests, pip installs, and other stuff that might make automation flaky? Should notebooks be explicitly included instead?

build.py
swernli: good call! I can do that as part of these changes.

build.py
swernli: That's a interesting question... my thinking with going with an exclude list instead is that it makes adding a new sample easier, and any misbehaving samples will (best case) get caught when the PR pipeline runs before they merge or (worst case) show as flakiness later that forces us to stop and take a look at why that notebook is flaky. Then we will be incentivised to keep our notebook samples well-contained and cleanly running, instead of defaulting to not testing them and having to remember to add good ones to the list here.

minestarks: Fair enough - we can always change later.

build.py
swernli: It actually doesn't rebuild it from scratch, but rather benefits from the incremental builds. So in practice the Rust compilation is nearly zero if the binary was built before and then it just links to the Python files in this folder rather than needing to zip everything into a wheel and then unzip again into the venv's module folder. In the ubuntu build pipeline this took about 2 seconds since it didn't need to rebuild anything. And if run in isolation (without the `--pip` flag) this line will build the needed Rust components if they haven't been recently built. I prefer that to requiring that folks remember to specify both `--pip` and the command runs the samples.

build.py
swernli: I waffled back and forth on this... I do like the idea of them being integration tests since they depend on more components, I just also wanted to make it easy to run them directly without also running the VSCode tests for local testing. Maybe we can make the existing `--integration-tests` a wrapper argument and have two separate arguments for the scenarios, something like `--integration-vscode` and `--integration-notebooks`, where specifying the wrapper one implies enabling all individual integration tests?

build.py
```diff
@@ -341,6 +344,91 @@ def step_end():
...
+        result = subprocess.run([python_bin,
+                        "-m",
+                        "nbconvert",
+                        "--to",
+                        "notebook",
+                        "--stdout",
+                        "--ExecutePreprocessor.timeout=60",
+                        "--execute",
+                        notebook],
+                        check=False, text=True, cwd=root_dir, env=pip_env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
swernli, line 425: Platform dependencies rear their ugly head again... it seems like one of the notebooks produces output that the Python infrastructure on Windows can't convert into a character string. I need to figure out how to adjust for that without needing to go back to spamming the logs with the very, very long output of the converted notebook.

swernli: Ok, explicitly specifying encoding seems to have resolved the issue. Apparently Windows is the only platform that doesn't set that by default.

build.py
minestarks: There's some demonic behavior that occurs with `pip install` where it will corrupt the `.whl` file that we built in the previous step. I had to fix this in #539 and I strongly suggest you avoid it as well :)

swernli: oh, good tip! I'll avoid this pattern then.

build.py
swernli: we talked a bit more about this, capturing where we landed here. Like the `--test` command today, I'll keep it so that `--integration-tests` runs the tests for whatever components have been specified. So for these new ones, you'll have to have included `--pip` and `--widgets` to have built the wheels. Then I'll update the ci.yml to include `--pip` and `--widgets`.

build.py
swernli: From local testing, it looks like a non-editable `pip install` with the folder path to widgets src does what we want (I checked the targets/wheel folder and saw the wheel built earlier is not touched when using this approach).

minestarks (approved):
Love it, thank you!

DmitryVasilevsky (approved):

cesarzc (approved):

PR #1079, opened 2024-01-27
Reverse Endianness

swernli:
Based on feedback, we are reversing the endianness of ket displays from the QDK along with associated code. This includes reversing the display for state dumps from within Q#, in the debugger, and in Jupyter output. Included in this change is the revert of updates to kata content that tried to accomodate the different ordering between exercise text and the state display. The change also includes update to the new `dump_operation` in Python so that it is consistent with the endianness and outputs matrices that match the same conventions.


minestarks:
@swernli Shouldn't we reverse the order of the formatted tables as well so they still go from 00001 => 11111?

ScottCarda-MS:
> @swernli Shouldn't we reverse the order of the formatted tables as well so they still go from 00001 => 11111?

image
Yes I think that is what I'm seeing here as well.

swernli:
> @swernli Shouldn't we reverse the order of the formatted tables as well so they still go from 00001 => 11111?

No, because that is the correct order for what those entries are. `|10⟩` is the first entry in the state vector, `|01⟩` is the second entry, etc.

If there were 4 qubits in equal superposition, it would go:
```|0000⟩
|1000⟩
|0100⟩
|1100⟩
|0010⟩
|1010⟩
|0110⟩
|1110⟩
|0001⟩
|1001⟩
|0101⟩
|1101⟩
|0011⟩
|1011⟩
|0111⟩
|1111⟩
```

tcNickolas:
The order of the printed entries is the second part of our discussion today, do we interpret bit strings as integers using little endian (least significant bit first) or big endian (most significant bit first)? For the CNOT matrix to have the right ("textbook") shape, we need to use big endian, so it makes sense to have the same for ordering the entries. (|0000⟩ then |0001⟩ then |0010⟩ etc.)

ScottCarda-MS:
> > @swernli Shouldn't we reverse the order of the formatted tables as well so they still go from 00001 => 11111?
> 
> No, because that is the correct order for what those entries are. `|10⟩` is the first entry in the state vector, `|01⟩` is the second entry, etc.
> 
> If there were 4 qubits in equal superposition, it would go:
> 
> ```
> |0000⟩
> |1000⟩
> |0100⟩
> |1100⟩
> |0010⟩
> |1010⟩
> |0110⟩
> |1110⟩
> |0001⟩
> |1001⟩
> |0101⟩
> |1101⟩
> |0011⟩
> |1011⟩
> |0111⟩
> |1111⟩
> ```

Is there something imposing a particular order other than what we want to display? Displaying the kets in that order seems confusing.

katas/content/multi_qubit_systems/prepare_with_complex/solution.md
```diff
@@ -4,17 +4,17 @@ $$ \begin{bmatrix} 1 \\\ 0 \\\ 0 \\\ 0 \end{bmatrix} = \begin{bmatrix} 1 \\\ 0 \
...
 
 For the first qubit, we'll start by applying the **H** gate, getting the state $\frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ 1 \end{bmatrix}$, as we've seen in the previous task. Afterwards we'll apply the **S** gate with the following result:
 
 $$ \begin{bmatrix} 1 & 0 \\\ 0 & i \end{bmatrix} \cdot \frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ 1 \end{bmatrix} = \frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ i \end{bmatrix}$$
 
-If we recall that $i = e^{i\pi/2}$, we can write the final state of the first qubit as:
+If we recall that $i = e^{i\pi/2}$, we can write the final state of the second qubit as:
 $$ \frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ e^{i\pi/2} \end{bmatrix} $$
 
 For the second qubit. we'll apply the **H** gate, followed by the **T** gate, with the following result:
```
ScottCarda-MS, line 20:
```suggestion
For the first qubit. we'll apply the **H** gate, followed by the **T** gate, with the following result:
```

katas/content/multi_qubit_systems/prepare_with_complex/solution.md
```diff
@@ -4,17 +4,17 @@ $$ \begin{bmatrix} 1 \\\ 0 \\\ 0 \\\ 0 \end{bmatrix} = \begin{bmatrix} 1 \\\ 0 \
...
+\frac{1}{\sqrt2}\big(|0\rangle + e^{i\pi/2}|1\rangle\big) \otimes \frac{1}{\sqrt2}\big(|0\rangle + e^{i\pi/4}|1\rangle\big) $$
 
 We will again need to adjust the states of both qubits independently.
 
 For the first qubit, we'll start by applying the **H** gate, getting the state $\frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ 1 \end{bmatrix}$, as we've seen in the previous task. Afterwards we'll apply the **S** gate with the following result:
 
 $$ \begin{bmatrix} 1 & 0 \\\ 0 & i \end{bmatrix} \cdot \frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ 1 \end{bmatrix} = \frac{1}{\sqrt2} \begin{bmatrix} 1 \\\ i \end{bmatrix}$$
 
-If we recall that $i = e^{i\pi/2}$, we can write the final state of the first qubit as:
+If we recall that $i = e^{i\pi/2}$, we can write the final state of the second qubit as:
```
ScottCarda-MS, line 17:
```suggestion
If we recall that $i = e^{i\pi/2}$, we can write the final state of the first qubit as:
```

ScottCarda-MS: This was a small typo in the original that I fixed in my changes, so if you were reversing my changes, we should keep this edit in to fix the typo.

swernli: Good catch, thanks!

swernli:
Given the feedback and sticking with the theme of trying to make things less confusing, I'll make sure the state display is sorted according to the display format rather than the original state vector ordering.

billti:
> No, because that is the correct order for what those entries are. |10⟩ is the first entry in the state vector, |01⟩ is the second entry, etc.

Is this correct? If |01> is the tensor product of |0> and |1>, then it is [1, 0]T x [0, 1]T which is [0, 1, 0, 0]T. In other words, the kets should be ordered from |00>, |01>, |10>, |11> to map to the way state vectors are typically laid out horizontally, i.e. mapping to top-to-bottm entries of [1, 0, 0, 0]T, [0, 1, 0, 0]T, [0, 0, 1, 0]T, and [0, 0, 0, 1]T.

swernli:
Now all the endianness matches and the ordering of the state entries are also sequential:
image
The state display in the debugger also matches this too.

minestarks (approved):
Phew, looks pretty thorough. Thanks!

cesarzc (approved):

PR #1080, opened 2024-01-28
RE and widget updates

billti:
In trying to add some desired features and improvements, I hit a few existing bugs and glitches. In the process of trying to address those I ended up doing quite a bit of refactoring. (This PR actually removes more code than it adds, despite adding several new features).

- Fixed issue where result name was blank if only one row was present (i.e. the run selected just one qubit type)
    - Note: There is still a bug with sometimes losing the current selection when new results/rows are added, which can cause the row 'name' to change due to the recent change to dynamically calculate the series/row name. This is because the name is used as the unique key for some saved state. I look more into this later.
- Fixed issue where multiple selections could occur on the scatter chart.
- Fixed issue where an error was reporting in the editor for UX components importing from the ticks module.
- Fixed some color rendering issues with JupyterLab in the browser in dark theme.
- Made some of the DOM manipulation code more declarative and canonical, reducing code size and complexity.
- Updated hover, selection, and tooltip display behavior to be more consistent.
- Added keyboard up/down handler to change the selected row when the results table is focused.
- The column selection menu now closes when you click outside of it.
- Minor adjustments to chart layout (e.g. reducing the left padding)

I think we want much of this for 1.1, so hopefully can get it in tomorrow (Monday). I've tested in VS Code and Jupyter Notebooks. CC @ivanbasov


.ado/publish.yml
```diff
@@ -124,7 +124,7 @@ jobs:
     displayName: Install Prereqs and set version
 
   - script: |
-      python ./build.py --jupyterlab --widgets --no-check
+      python ./build.py --npm --jupyterlab --widgets --no-check
```
billti, line 127: Ugh, just realized npm depends on wasm, so we'll end up building the whole thing here. Will think about it, but might go back to ignoring the import warning from TypeScript when importing 'out of tree' modules :-/

billti: Fixed it via a tsconfig.json change, so no build changes needed - thankfully.

npm/ux/data.ts
```diff
@@ -10,7 +10,6 @@ export type ReData = {
   tfactory: any;
   errorBudget: any;
   logicalCounts: any;
-  new: boolean;
```
ivanbasov, line 13: Maybe isNew could be considered instead

billti: Not sure I follow. The 'new' property was only used when 'ensureSelected' was set, but one of your changes made that it never is, so I removed all the code related to 'ensureSelected', including the 'new' property.

npm/src/browser.ts
```diff
@@ -249,3 +249,5 @@ export type {
 } from "../lib/web/qsc_wasm.js";
 export { type IStructStepResult, StepResultId } from "../lib/web/qsc_wasm.js";
 export { type LanguageServiceEvent } from "./language-service/language-service.js";
+
+export * as utils from "./utils.js";
```
ivanbasov, line 253: just to confirm that this export is required

npm/src/main.ts
```diff
@@ -145,3 +145,5 @@ export function getLanguageServiceWorker(): ILanguageServiceWorker {
 
   return proxy;
 }
+
+export * as utils from "./utils.js";
```
ivanbasov, line 149: just to confirm that this export is required

npm/ux/scatterChart.tsx
```diff
@@ -201,199 +85,216 @@ export function ScatterChart(props: {
...
+        break;
+      default:
+        console.error("Unknown event type: ", eventType);
     }
   }
 
-  const scatterChartContainerRef = useRef<SVGSVGElement | null>(null);
-  let chart: Element | null = null;
-
-  function updateCoordinates() {
```
ivanbasov, line 232: This one was recently introduced to consolidate all coordinate computations and separate them from rendering (which was originally provided).

A potential for this was a chart resizing scenario. Let us discuss if we would like to consider the resizing scenario and if so, if the proposed update would support it.

npm/src/browser.ts
billti: If we're going to include it in the main package code (e.g. so we can test it), it should be exposed somehow. For example, if you look at the unit tests now, it imports this export.

npm/src/main.ts
billti: See above. unit tests use this export now.

npm/ux/scatterChart.tsx
billti: Resizing should update some state (either stored in the passed in props or locally with useState), which would cause a re-render, so the new approach would work fine with resizing.

minestarks (approved):

ivanbasov (approved):

PR #1089, opened 2024-01-29
Add opt-in language features to the Qsharp.json manifest

### LINKED ISSUE #543: Unstable features
A concern that keeps coming up when designing new language features: a fear of breaking the feature shortly after it's implemented if we discover new issues after trying to use it in practice. Being able to put new features behind an opt-in unstable flag, like Rust, would give us more freedom to experiment with feature designs.

sezna:
Closes #543 

## Overview

This PR adds two things:
1. The ability to opt-in to language features via the project manifest, `qsharp.init()`, or command-line flags
2. A simple feature, `v2-preview-syntax`, ostensibly created to allow users to opt-in to breaking syntax that's coming in the future, but really created so I could wire up and show how a real language feature works. Only one syntax change was made in this feature: removal of scoped qubit blocks. This is because a scoped qubit block is identical to a scoped block with a qubit in it. `use qubit = Qubit () {};` is equivalent to `{ use qubit = Qubit(); }`. We therefore prefer the latter.

## Implementation Notes

This PR includes the addition of `ParserContext`. It became clear to me that language features need to be globally available to _at least_ the parser, and probably other areas of the compiler later. Instead of plumbing this all around manually as an argument, I promoted `Scanner` into a `ParserContext`. 

This PR also refactors manifest loading in `projectSystem.ts` -- we had some duplicate code in `getManifestThrowsOnParseFailure` and `getManifest`. I DRY'd that up.

### Why opt-in features?

In our discussions of evolving the language, it became clear that we prioritize both improving the language, including in breaking ways, and maintaining compatibility. As a result, introducing the notion of being able to opt-in to a specific language feature became a blocker for most further language design work.

### How will packages with different feature sets work together?

In general, we should aim to write features that don't break compatibility with other packages. The language server and all compiler scenarios assess language features on a per-package basis, so as long as the individual compilation units are sane and well-formed, their feature sets should be interoperable.

## Tests

These tests are considered _passing_ if turning on the language feature `v2-preview-syntax` causes the compiler to reject scoped qubit block syntax.

✅ Histogram:

https://github.com/microsoft/qsharp/assets/12157751/631dded2-d016-4933-83fe-bdff67852f50

✅ Language Service

https://github.com/microsoft/qsharp/assets/12157751/964574f6-c6bd-4225-acd4-6eb2afaec41c

✅  Jupyter Notebook

https://github.com/microsoft/qsharp/assets/12157751/fc5cc8b8-1c47-4dd9-b478-a233f16fc9e3

✅ `qsc`

https://github.com/microsoft/qsharp/assets/12157751/48e746fc-f7fe-4e9a-a658-d428b640d062

✅ `qsi`

https://github.com/microsoft/qsharp/assets/12157751/8154de70-80c5-444c-ac4e-f33d775a2fa9

✅  Debugger

https://github.com/microsoft/qsharp/assets/12157751/5aa7fb5a-3ec4-43fe-b256-6b75c808fe7e

✅ Language service inside of notebook cells

image

✅  QIR Generation

https://github.com/microsoft/qsharp/assets/12157751/eed50ca5-85c4-4434-9443-a6c67ef10337


resource_estimator/src/counts/tests.rs
```diff
@@ -12,6 +12,7 @@ use qsc::{
     target::Profile,
     PackageType, SourceMap,
 };
+use qsc_data_structures::language_features::LanguageFeatures;
```
ivanbasov, line 15: I do not think that we should introduce references to compiler into the resource_estimator. In an ideal case, RE should be independent from the compiler and specific language. So, I would prefer to avoid this. However, if there is a cascade referencing requiring this, then OK - let us proceed.

ivanbasov: Sounds good. We may consider separating RE from the qsharp packaging since it could support other languages and should not depend on qsharp. So, if there is anything we could do to prevent potential issues with this, it would be better to take it into account.

swernli: It's also worth noting that this part of the resource estimator is the part that connects to the evaluator and handles gathering logical counts during simulation. It requires a tight connection to the Q# infrastructure to accomplish that. We could separate that into its own crate that depends on the physical estimator and qsc, such that the physical estimator would not depend on another code in the repo. Some of the original designs had such a division, we just haven't had the chance to do that refactoring yet.

compiler/qsc_data_structures/src/language_features.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct LanguageFeatures(BTreeSet<LanguageFeature>);
+pub type LanguageFeatureIncompatibility = miette::ErrReport;
+pub type UnrecognizedLanguageFeature = miette::ErrReport;
+
+impl LanguageFeatures {
+    /// Checks that the current set of language features is compatible and well-formed.
+    /// Returns a descriptive error message if not.
+    pub fn check_compatibility(&self) -> Result<(), LanguageFeatureIncompatibility> {
```
cesarzc, line 34: Since this method is currently not really doing anything, I think it would be better to not introduce it preemptively. If we add features in the future that are not compatible with each other, we can add a compatibility check then.

compiler/qsc_data_structures/src/language_features.rs
```diff
@@ -0,0 +1,76 @@
...
+            "v2-preview-syntax" => Ok(LanguageFeature::V2PreviewSyntax),
+            _ => Err(UnrecognizedLanguageFeature::msg(format!(
+                "Unrecognized language feature: {s}"
+            ))),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct LanguageFeatures(BTreeSet<LanguageFeature>);
```
cesarzc, line 27: Do we expect to have more information associated with each language feature that just present/not-present?
- If it's just present/not-present, we could use bitflags instead of a set.
    - Using bitflags would have the benefit of using much less space and avoiding `clone()` in a few places.
- If not, do we expect to be able to populate the additional information associated with each language feature from the CLI, manifest and other compiler APIs?
    - IMO, this option seems like a lot of complexity where a simple present/not-present would be enough.

cesarzc: Using bitflags would also have the benefit of consistency since now runtime capabilities and language features would work the same way.

compiler/qsc_data_structures/src/language_features.rs
```diff
@@ -0,0 +1,76 @@
...
+impl LanguageFeatures {
+    /// Checks that the current set of language features is compatible and well-formed.
+    /// Returns a descriptive error message if not.
+    pub fn check_compatibility(&self) -> Result<(), LanguageFeatureIncompatibility> {
+        // we currently only have one language feature, so no checking is required.
+        Ok(())
+    }
+
+    #[must_use]
+    pub fn none() -> Self {
```
cesarzc, line 40: I think it would be better to have a `default()` method instead of a `none()` method. Since we can use language features to experiment with things that might not be ready for release but are not breaking changes (e.g. performing RCA), we could release these features by including them in `default()`.

npm/src/compiler/compiler.ts
```diff
@@ -15,15 +15,23 @@ type Wasm = typeof import("../../lib/node/qsc_wasm.cjs");
 // for running the compiler in the same thread the result will be synchronous (a resolved promise).
 export interface ICompiler {
   checkCode(code: string): Promise<VSDiagnostic[]>;
-  getHir(code: string): Promise<string>;
+  getHir(code: string, language_features: string[]): Promise<string>;
   run(
     sources: [string, string][],
     expr: string,
     shots: number,
+    language_features: string[],
```
cesarzc, line 23: This would be a breaking API change. Could we make this an optional parameter?

katas/src/lib.rs
```diff
@@ -13,6 +13,8 @@ use qsc::{
     PackageType, SourceContents, SourceMap, SourceName,
 };
 
+use qsc_data_structures::language_features::LanguageFeatures;
```
swernli, line 16: Same here and through out the other uses. That should also remove the need to add `qsc_data_structures` to the Cargo.toml files.

language_service/src/state.rs
```diff
@@ -186,12 +200,16 @@ impl<'a> CompilationStateUpdater<'a> {
     async fn load_manifest(
         &self,
         doc_uri: &Arc<str>,
-    ) -> Option<(Arc<str>, Vec<(Arc<str>, Arc<str>)>)> {
+    ) -> Option<(Arc<str>, Vec<(Arc<str>, Arc<str>)>, LanguageFeatures)> {
```
swernli, line 203: nitpick: this type is starting to get long (which I'm often guilty of myself), so it seems like it might be worth putting it in a struct to make the intentions clear and use named access (rather than the `.0`, `.1`, `.2` below). Mostly a personal preference, but when I see `.0` or `.1` I fine with it because its just a pair, but once it gets to `.2` it starts looking messy to me.

compiler/qsc_data_structures/src/language_features.rs
```diff
@@ -0,0 +1,76 @@
...
+        val.0
+    }
+}
+impl From<BTreeSet<LanguageFeature>> for LanguageFeatures {
+    fn from(val: BTreeSet<LanguageFeature>) -> Self {
+        LanguageFeatures(val)
+    }
+}
+
+impl ValueEnum for LanguageFeature {
```
swernli, line 64: When converting to bitflags as suggested above, I think it would be worth dropping the dependency on `ValueEnum` and clap and just implementing a try-conversion from `String` similar to what we did for target profiles (see compiler/qsc/src/target.rs line 33) along with a matching `to_str` implementation. It's a little more manual to keep in sync, but that shouldn't be a concern until we have very many values at which point we can revisit.

compiler/qsc_frontend/src/compile/tests.rs
```diff
@@ -1103,3 +1169,95 @@ fn duplicate_intrinsic_from_dependency() {
     "#]]
     .assert_debug_eq(&unit.errors);
 }
+#[test]
```
swernli, line 1172: missing newline

compiler/qsc_frontend/src/compile/tests.rs
```diff
@@ -1103,3 +1169,95 @@ fn duplicate_intrinsic_from_dependency() {
...
+                            },
+                        ),
+                    ),
+                ),
+            ),
+        ]
+    "#]]
+    .assert_debug_eq(&unit.errors);
+}
+#[test]
```
swernli, line 1230: missing newline

compiler/qsc_parse/src/stmt.rs
```diff
@@ -111,15 +111,20 @@ fn parse_qubit(s: &mut Scanner) -> Result<Box<StmtKind>> {
     let lhs = pat(s)?;
     token(s, TokenKind::Eq)?;
     let rhs = parse_qubit_init(s)?;
-    let block = opt(s, parse_block)?;
+    let block = if !s.contains_language_feature(LanguageFeature::V2PreviewSyntax) {
```
swernli, line 114: nitpick: swapping the then-else blocks will let you drop the `!` and improves readability, in my opinion.

npm/src/compiler/compiler.ts
minestarks: There is a fairly idiomatic way to make this change backward compatible. You define method overloads in the interface and check the argument type at runtime. See my [example on the TypeScript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyg9gWwgeQEYCsIGNgBVzQC8UA3lKgDYCGAFgFxQDOwATgJYB2A5lAL4DcAKEGdgEFgDMqWaAEkAYnDilBUNVAD0AKi1QAAgBMIYFtipiDUMBQhVG0AK72owGtA4QA7lDgA3cRRwVJZaGqrqEkoAFFQsLFQgAIIsXAzM7NwA2gC6AJQMvnBsBkLhapFwUXAY2MDJqbCIKDU4+JAANFCx8UkpaaycXDn5UIXFQrzCWNSMjFCKymwI1hBIHMBzCkoq6lAVMXEJyCxomDj1-RlD2VAAPo1Ip7VtEJ3dCfUA-JeDw6RQZV2UDYEigUWSPQAdGxGBCEgcescnucUrlciRAUD1Fg4BxGHAbJDAlwogAiCRsFjMHz+FiBYIAQlJuSEWKxGg0UBeMCw7DAwCgHG6cE8c1c0FAkBcynSv2ynXsErcUBofnEwI4kR8HC6hxASJadRSmLZOLxBIgRLgJNJjFVDgollQ0FlWWyDFJbz1BrORq4LJNWLN+MJxLJdrgDqdjg4RgpHgMHq9PXqAbZQODFqtJLTbN4UAgFGcGPTu0zoet4ewuMsarpQQMTNzpc0nO5vLY-MFwtFLmVkolyngj0NLwVECV0FVtI1WtxusRJ0N9UDGdxIctYdt9sd5Ggw+avpedE9C6OS99qf4AJbanLm8r28ju+dTAGbqTZ96-tZt-v2aiZsgUmXZJkmQRgwFCoGC2ZRiA8bwFkAoQKkhfZMlJKhT1JVBsKwUk8hQpQ0OiMhKFoD1yJoUk+E6DCsM6HC8IIlkgA) . TypeScript is smart enough to narrow the type of the argument at compile time, and JavaScript provides runtime types so that you can check at runtime which version of the function is being called.

Here, we'd check the type of the fourth argument at runtime, if it's a string array, we know the caller is using the new overload. If not, we know they're calling the old version.

Now, that being said. This will get us through in the short term, but maybe this is a good point to pause and consider our strategy for handling breaking changes in this group of functions going forward.

1. I propose changing the signatures of these methods to something like:

```ts
run(config: ProgramConfig, eventHandler: IQscEventTarget);
getEstimates(config: ProgramConfig, params: string);
..... etc ....

type ProgramConfig  = {
    sources: {
        name: string;
        contents: string;
    }[];
    languageFeatures?: string[];
    // future-proof - we can add other optional properties as time goes on
}
```

IMO, it's more maintainable to make backward-compatible changes to object properties than to function arguments. Optional properties look nicer at the call site (`foo({ bar: "bar"})` looks better than `foo(undefined, undefined, undefined, undefined, "bar")`). Also, since most of these functions take basically the same set of arguments; it makes sense to me to group them together into an object.

2. Requiring the `IQscEventHandler` argument to be the last argument wasn't the greatest decision, and maybe I can change that too while we're at it. I can make it the first argument - gives us a bit more flexibility with future optional arguments.

cesarzc: I liked @minestarks's suggestion and the new `run`. I'm in favor of doing this for the rest of the API.

minestarks: @sezna yep that's exactly what I had in mind! I can also take this opportunity to move `IQscEventHandler` to be the first argument? But that may be a bit too much to add to this PR... perhaps it could be a separate change after this PR is merged, but before we ship the next build (since there's no need to have _two_ deprecated signatures if we can avoid it).

minestarks: Sounds fine by me. We can iron out the remaining wrinkles in the follow up PR

npm/src/compiler/compiler.ts
```diff
@@ -15,22 +15,43 @@ type Wasm = typeof import("../../lib/node/qsc_wasm.cjs");
 // for running the compiler in the same thread the result will be synchronous (a resolved promise).
 export interface ICompiler {
   checkCode(code: string): Promise<VSDiagnostic[]>;
-  getHir(code: string): Promise<string>;
+  getHir(code: string, language_features?: string[]): Promise<string>;
+  /** @deprecated -- switch to using `ProgramConfig`-based overload **/
```
minestarks, line 19: Can you make this string _extremely_ verbose and helpful? The first person to read the deprecation error is going to freak out even though we're not actually breaking anything ;)

pip/src/interpreter.rs
```diff
@@ -535,3 +549,21 @@ fn get_dict_opt_string(dict: &PyDict, key: &str) -> PyResult<Option<String>> {
...
+    match list
+        .iter()
+        .map(|item| {
+            item.downcast::<PyString>()
+                .map(|s| s.to_string_lossy().into())
+        })
+        .collect::<std::result::Result<Vec<String>, _>>()
+    {
+        Ok(list) => Ok(list),
+        Err(e) => Err(QSharpError::new_err(e.to_string())),
```
minestarks, line 567: This error type doesn't look quite right to me, `QSharpError` is more for like compiler errors, but I'm getting a bit lost in the types... what's `e` here, and is there a more direct conversion? Do you need to convert at all? I remember doing a bunch of work in this file's error handling but none of it makes sense anymore...

wasm/src/project_system.rs
```diff
@@ -165,14 +166,34 @@ pub(crate) fn get_manifest_transformer(js_val: JsValue, _: String) -> Option<Man
                 v
             )
         }),
-                    Err(_) => unreachable!("our typescript bindings should guarantee that an object with a manifestDirectory property is returned here"),
+        Err(_) => unreachable!("our typescript bindings should guarantee that an object with a manifestDirectory property is returned here"),
```
minestarks, line 169: I feel like most of this function could be replaced with `serde_wasm_bindgen::from_value` as long as you define a `ManifestDescriptor` struct with the proper fields and marked as `Deserializable`.

minestarks (changes requested):
Looks good overall, just a few cleanups

language_service/src/state.rs
```diff
@@ -77,12 +77,21 @@ impl Default for Configuration {
     }
 }
 
-#[derive(Default, Clone, Copy, Debug)]
+#[derive(Default, Clone, Debug)]
 struct PartialConfiguration {
     pub target_profile: Option<Profile>,
     pub package_type: Option<PackageType>,
+    pub language_features: Option<LanguageFeatures>,
```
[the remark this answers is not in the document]
minestarks: No, this is the intended and we follow the same pattern for every other setting that apply to the compilation, like target_profile and package_type.  Think of it like a workspace default. Yes, you can override it with a qsharp.json, but you have to have _something_ to fall back to. Right now you're still implicitly harcoding a default, I'm just suggesting moving the hardcoded default to be at the workspace-level to be consistent with the other settings.

wasm/src/project_system.rs
minestarks: Oops, I didn't realize this was just a whitespace change. I still think `serde_wasm_bindgen::from_value` should be directly callable, I'll try to sketch up an example, but don't worry about holding up the PR for that.

minestarks (approved):

language_service/src/state.rs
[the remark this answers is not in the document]
minestarks: Thanks! While I'm with you on not wanting to expose it to the user, I don't think it's nonsensical at all... workspace level settings apply to open documents without a qsharp.json. It's just a default. Projects/python environments could override it, but we have to have a default somewhere.... I don't think it's a stretch to think that we may want to let the user configure that default at some point in the future.

swernli (approved):

PR #1098, opened 2024-01-31
Update spans used for some type mismatch errors

### LINKED ISSUE #970: Need better error placement for return type mismatch in case when actual type is Unit
**Describe the bug**

When the declared return type of a function doesn't match the actual type returned, an error message is produced. The error message is bound to the last statement, which doesn't always produce intuitive results. For example, if a function is declared as returning Int, but the return statement is actually missing, the last statement (producing Unit) is underlined. It may be a large statement, which makes it not appealing visually and makes it a bit misleading.

Moreover, when you are just writing code and haven't gotten to the last return statement yet, this error gets in the way and is very annoying. It also masks other errors in the code you are typing.

**To Reproduce**

https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACnWLywqCQBhG9%2FMUHxKiRKLbQmjTUrC20WKKXxlwZmQuoMi8exgtmqKz%2By5HcUl25A9CM589V87L1ujecImFAcDxpJyZWy2Uy%2FJXo0cy3Amt0HChshx7XMj6wV1vb2el0wYCQqEsiqr8GFakd%2Fw%2BECbUKA%2FRJDoIpGmKCnWNb3HFksOEbY0qNgNosPRX2P0KUWrIWt5TtkmWKSQ5i1%2BBBfYEtHZmjiwBAAA%3D

Also in VS Code:

image

**Expected behavior**

Maybe we should treat the mismatch case when the actual return type is Unit as a missing return. In this case, it would probably make more sense to bind the error to the return type rather than the last satatement.

swernli:
This change adds some additonal logic to better select error spans and expected/actual types for both if-expressions and callable blocks.

Fixes #970


swernli (commented):
See screenshots and discussion in linked issue for more context

minestarks (approved):

PR #1100, opened 2024-02-01
Added support for boolean exclusive disjunction in standard library

### LINKED ISSUE #1096: Missing XOR support for booleans
The old QDK had a handy [Xor](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.logical.xor) function which could be used to perform XOR between two booleans. 

It was a super simple function but very useful, so it could just be added back

```qsharp
    function Xor(a : Bool, b : Bool) : Bool {
        (a or b) and ((not a) or (not b))
    }
```

Alternatively the `^^^` (or `^`) could be extended to work with booleans - at the moment it works with ints only. For example in C# I could write `true ^ true` or `false ^ true`

1. do nothing and force everyone to write this by hand
2. add the function
3. extend the operator

**Additional context**
None

filipw:
Fixes #1096 

The motivation is discussed in that issue as well. I put in the `Microsoft.Quantum.Logical` namespace, but we can move it into a different one if needed.

### DIFF
--- a/library/std/logical.qs
+++ b/library/std/logical.qs
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Logical {
+
+    /// # Summary
+    /// Returns the boolean exclusive disjunction (XOR) of two input boolean values.
+    ///
+    /// # Input
+    /// ## first
+    /// The first boolean value to be considered.
+    ///
+    /// ## second
+    /// The second boolean value to be considered.
+    ///
+    /// # Output
+    /// A `Bool` which is `true` if and only if exactly one of `first` and `second` is `true`.
+    ///
+    /// # Example
+    /// ```qsharp
+    ///
+    /// let result = Xor(true, false);
+    /// // result is true
+    /// ```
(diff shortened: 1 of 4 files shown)


library/std/logical.qs
```diff
@@ -0,0 +1,28 @@
...
+    /// A `Bool` which is `true` if and only if exactly one of `first` and `second` is `true`.
+    ///
+    /// # Example
+    /// ```qsharp
+    ///
+    /// let result = Xor(true, false);
+    /// // result is true
+    /// ```
+    function Xor(first : Bool, second : Bool) : Bool {
+        (first or second) and ((not first) or (not second))
```
msoeken, line 26:
```suggestion
        first != second
```

msoeken (changes requested):
I suggest to simplify the expression.

library/std/logical.qs
filipw: yes makes sense, thanks!

msoeken (approved):
Looks good to me. Thanks @filipw

DmitryVasilevsky (approved):
Thank you! We decided to go with the compatibility argument.

PR #1103, opened 2024-02-01
Session now exits when there is a runtime failure when running without debugging

### LINKED ISSUE #1082: Debugging panel disables the Run button if Q# program throws exception
**Describe the bug**

If I get an exception in the output to debug console (for example, qubits released not in zero state), the debugging panel pops up, and the run button doesn't work any longer. At the same time, the debugging panel is not really visible (it's placed over the tabs with different files, so it blends in really well). This can make rerunning the code after fixing the bug very confusing.

**Expected behavior**

Keep the run button enabled in addition to the debugging panel. I don't think there's any state of the program the debugging tool captures at this point that needs to be preserved (debugging the program needs to start from scratch). And I can often fix the exception without going through debugging, so making the panel mandatory is not necessary.

**System information**

- QDK 1.0.33
- Ubuntu

idavis:
Fixes #1082

### DIFF
--- a/vscode/src/debugger/session.ts
+++ b/vscode/src/debugger/session.ts
@@ -390,25 +390,34 @@ export class QscDebugSession extends LoggingDebugSession {
     // This will be replaced when the interpreter
     // supports shots.
     for (let i = 0; i < args.shots; i++) {
-      const result = await this.debugService.evalContinue(
-        bps,
-        this.eventTarget,
-      );
-      if (result.id != StepResultId.Return) {
-        await this.endSession(`execution didn't run to completion`, -1);
-        return;
-      }
-      this.writeToDebugConsole(`Finished shot ${i + 1} of ${args.shots}`);
-      // Reset the interpreter for the next shot.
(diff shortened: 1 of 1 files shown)


vscode/src/debugger/session.ts
```diff
@@ -390,25 +390,34 @@ export class QscDebugSession extends LoggingDebugSession {
...
-      // Reset the interpreter for the next shot.
-      // The interactive interpreter doesn't do this automatically,
-      // and doesn't know how to deal with shots like the stateless version.
-      await this.init(associationId);
-      if (this.failureMessage != "") {
-        log.info(
-          "compilation failed. sending error response and stopping execution.",
+      const success = await this.debugService
+        .evalContinue(bps, this.eventTarget)
+        .then(
```
minestarks, line 395: Mixing `await` and `then` patterns is pretty confusing - could you rewrite this as just straight line code with `await`s? Errors can simply be `try`/`catch`ed.

minestarks (approved):

PR #1107, opened 2024-02-02
Add integration tests for VS Code debugging, eliminate `FileAccessor`

minestarks:
Depends on #1106

VSCode's debugging related APIs sometimes use filesystem paths instead of URIs to represent sources. Previously, our solution was to have a `FileAccessor` which tried to treat the path both ways (first as a filesystem path, then a URI). But by being strategic about exactly where we expect filesystem paths and sanitizing them, we can just consistently use URIs in the rest of the code.

Additionally this PR introduces integration tests for the debugger.

We also fix two small-ish bugs here:
- launch.json variables like `{file}` and `{workspaceFolder}` weren't treated properly on all platforms/filesystems
- when launching the debugger on an untitled file, the user may be prompted to save the file. If they choose to save the file, the debugger will fail to launch. (This is probably not as big an issue anymore, since we disabled the auto-save behavior by default since I've opened this PR.)


billti:
I wish I could remember the issues we hit (maybe @idavis does) but we hit repeated problems with this code when implementing the debugger and when to use strings vs Uris. (I was hoping we'd not touch it again, else we should have commented it with the details).

The biggest challenge I recall was specifically with the GitHub VFS/scheme, so would only occur when you'd open the editor on github.dev, for which I needed to deploy a test extension to be able to test and debug.

If I vaguely recall the root cause, it was that at some point in the infra it marshals one of the values via JSON, which for a string works fine, but for a Uri converting it to JSON and back loses the Uri methods (as the Uri base class is not on the prototype chain anymore). However the GitHub file system would detect that one of the Uri properties is present (e.g. 'scheme'), assume IT IS a Uri, and then try and call Uri methods on it (like `.from()` or `.with()`) which would crash with the expected `TypeError: from is not a function`.

Either way, if touching this stuff again, be sure to REALLY test it, included when it's running the browser as an extension working with the GitHub virtual file system on github.dev

idavis:
The issue was around paths coming in from VS being windows paths instead of URIs. Normally we'd be able to tell the underlying debug session that we want paths always as URIs, but it is broken in web extensions. See this issue: https://github.com/microsoft/vscode-debugadapter-node/issues/298 .

I haven't dug into this PR, but when the call to `createDebugAdapterDescriptor` is made, we sanitize the input to get it as a URI before feeding it into the project resolver. This way we never read the value `config.program` as the project resolver was fed the sanitized URI and loaded all related source.

We also use `resolvePathToUri` to sanitize paths coming in for `breakpointLocationsRequest`, `setBreakPointsRequest`, and `stackTraceRequest` callbacks.

I've left out some historical detail as the project loader changed how the initialization of the debug session is performed.

minestarks:
@billti @idavis yeah I'm in no rush to check this in without extensive testing, since so far every environment I tried had a unique behavior which made me revise my assumptions... like even Windows _browser_ behaves differently from Linux _browser_, even when you're using a virtual filesystem - what the heck. 

I have this test matrix which I'm manually going through but I can't test github.dev / vscode.dev without shipping the dev build, I think. But please let me know if there's another way!

Test:

Launch configs:

    - launch.json ${file}
    - launch.json ${workspaceFolder}
    - launch.json no program
    - debug command
    - debug command on untitled buffer, saving
    - debug command on untitled buffer, without saving

Actions:

    - Set and run until breakpoints in main file
    - step in/out of stdlib
    - Set and run until breakpoints in stdlib
    - Inspect call stacks
    - (Multi-file only)
        - step in/out of child files, if any
        - Set and run until breakpoints in child files

VSCode environments:

    - desktop, local folder
    - desktop, wsl
    - desktop, playground (qsharp-vfs://)
    - locally hosted vscode web (@vscode/test-web), test fs (vscode-test-web://)
    - vscode.dev, playground (qsharp-vfs://)
    - vscode.dev, local folder
    - github.dev, local folder

Platforms:

    - Windows
    - Linux
    - Mac

vscode/src/debugger/session.ts
```diff
@@ -874,4 +813,57 @@ export class QscDebugSession extends LoggingDebugSession {
...
+      return;
+    }
+
+    const uri = this.asUri(source.path);
+    if (!uri) {
+      return;
+    }
+
+    try {
+      const doc = await vscode.workspace.openTextDocument(uri);
```
billti, line 836: If I'm reading the doc right, this actually opens the document in the editor, correct? Is this the desired behavior for 'tryLoadSource'? It looks like this is called when resolving breakpoints, which happens at the start of the session I think, so does this mean when I start debugging, every doc that has a breakpoint in it will be opened in the editor?

vscode/src/debugger/session.ts
```diff
@@ -874,4 +813,57 @@ export class QscDebugSession extends LoggingDebugSession {
...
+  /**
+   * Attemps to resolve a DebugProtocol.Source.path to a URI.
+   *
+   * This method will attempt to undo the lossy conversion
+   * from URI to path that the VS Code debug adapter client performs.
+   * See: https://github.com/microsoft/vscode/blob/3246d63177e1e5ae211029e7ab0021c33342a3c7/src/vs/workbench/contrib/debug/common/debugSource.ts#L90
+   *
+   * If the path was originally constructed from a file:// URI, it's a
+   * filesystem path. We use our knownPaths map to resolve it back to a URI.
+   * Filesystem paths we don't know about *won't* be resolved, and that's ok.
```
billti, line 855: I've been looking at the comments here and on the `knownPaths` member for a bit and still can't fully grok why we need this or what problem it solves. Can you try to clarify a bit? Thanks.

vscode/src/debugger/session.ts
```diff
@@ -666,79 +656,28 @@ export class QscDebugSession extends LoggingDebugSession {
...
+            this.convertDebuggerLineToClient(f.location.span.start.line),
+            this.convertDebuggerColumnToClient(f.location.span.start.character),
+          );
+          sf.endLine = this.convertDebuggerLineToClient(
+            f.location.span.end.line,
+          );
+          sf.endColumn = this.convertDebuggerColumnToClient(
+            f.location.span.end.character,
+          );
+          return sf;
```
billti, line 680: I see a bunch of code was deleted related to debugging into the built-in core/std library. Does that "just work" now and any special handling is no longer necessary?

vscode/test/suites/debugger/debugger.test.ts
```diff
@@ -0,0 +1,567 @@
...
+    ]);
+
+    // step into call (will be a call into intrinsic.qs)
+    await vscode.commands.executeCommand("workbench.action.debug.stepInto");
+
+    await waitUntilPaused([
+      {
+        id: 1,
+        source: {
+          name: "qsharp-library-source:intrinsic.qs",
```
billti, line 373: Nice! I see we have a test for stepping into the stdlib, so maybe ignore my last comment :-)

vscode/test/suites/extensionUtils.ts
```diff
@@ -31,3 +31,66 @@ export async function activateExtension() {
...
+ *
+ * Adding an unconditional delay to tests is guaranteed to slow tests down
+ * unnecessarily. It also reduces reliability when we're not clear about
+ * exactly what we're waiting on.
+ */
+export async function delay(timeoutMs: number) {
+  try {
+    await waitForCondition(
+      () => false,
+      () => ({ dispose() {} }),
```
billti, line 89: What 'dispose' is this calling?

vscode/src/debugger/session.ts
minestarks: No, `openTextDocument` doesn't open a visible text editor for the document (that's `showTextDocument`). Think of it like VS Code opening the document in the background - it's handy when you need information from the document model but don't want to show UI. We used to use `openTextDocument` all over the place to map offsets to line+columns, without opening the doc (but we don't anymore since lines/columns are mapped in Rust now).

vscode/src/debugger/session.ts
minestarks: Yes, you can't quite tell in this PR, but the changes in the `interpret.rs` mean that the special stdlib URLs are now returned directly from the native debugger, so we don't need to "guess" like we were doing up here.

This behavior now matches what the Language Service does (e.g. when returning URLs to stdlib in go-to-def).

billti (approved):

vscode/test/suites/extensionUtils.ts
minestarks: This is really hard to follow, but I'm actually defining an object here with a `dispose()` method that does nothing. (The horribly ambiguous `() => { foo }`/`() => ({ foo })`  syntax!)

This is to satisfy the type of this parameter, `vscode.Event<any>`. `vscode.Event<any>` has to return a `Disposable` when invoked.

vscode/test/suites/extensionUtils.ts
```diff
@@ -31,3 +31,66 @@ export async function activateExtension() {
...
+  await new Promise<void>((resolve, reject) => {
+    let done = false;
+    setTimeout(() => {
+      if (!done) {
+        reject(new Error(timeoutErrorMsg));
+      }
+    }, timeoutMs);
+
+    disposable = wakeUpOn(() => {
+      if (!done && condition()) {
```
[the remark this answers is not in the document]
minestarks: Yeah good catch, I went back on forth on this. Eventually I settled on multiple wakeups, to support a use case like this:

```ts
await waitForCondition(
      () =>
        !!notebook
          .getCells()
          .find((cell) => cell.document.languageId === "qsharp"),
      vscode.workspace.onDidChangeNotebookDocument,
      50,
      "timed out waiting for a Q# code cell",
    );
```

`onDidChangeNotebookDocument` will fire multiple times for unrelated reasons, but unfortunately we don't have a more precise event to wait for. This is why we can't give up after the first wakeup - we have to keep checking.

vscode/test/suites/extensionUtils.ts
billti: Ah! Got it.

I think part of what confused me is that in waitForCondition, disposable is `vscode.Disposable | undefined;` and its checked for nullness before invoking `disposable?.dispose();`, so seems like you shouldn't have to provide it if it's optional.

billti: Maybe the 'disposable' above shouldn't be nullable if invoking an Event<> is required to return one?

minestarks: Took another look, and I don't think I can change it. Even though in practice `disposable` should always be set, the null-check is necessary to make TypeScript happy. `disposable` needs to be declared outside the promise (so we can dispose after the promise resolves) but it's initialized _inside_ the promise, so we have to assume it's nullable in the outer scope.

PR #1108, opened 2024-02-02
Save dirty files before launching histogram and QIR generation

sezna:
This PR aims to solve the issue of executing histogram or QIR generation with dirty files.

### DIFF
--- a/vscode/src/qirGeneration.ts
+++ b/vscode/src/qirGeneration.ts
@@ -49,6 +49,7 @@ export async function getQirForActiveWindow(): Promise<string> {
     }
   }
   let sources: [string, string][] = [];
+  await vscode.workspace.saveAll(false);
   try {
     sources = await loadProject(editor.document.uri);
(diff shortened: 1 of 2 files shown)


billti (changes requested):
I think we should be going in the other direction and trying to save files less often, not more. Let's resolve ongoing discussions before making any changes.

minestarks:
@billti since we're aiming for consistency, let's also consider that VS Code automatically saves files when the "Run Q# File" and "Debug Q# File" editor commands are used. See: https://github.com/microsoft/vscode/blob/9794c5e919eb3a44c8159a72d6db90bf1bee9c38/src/vscode-dts/vscode.d.ts#L16287

I do like the solution in this PR a lot, especially since we pass `includeUntitled=false` . If the file isn't already backed by the filesystem, we won't try to save it. Which means the unsaved buffer / filesystem-free scenario will still work, while projects also behave consistently.

PR #1109, opened 2024-02-02
Web workers: refactor away some boilerplate code

minestarks:
There was a lot of repeated code among the compiler, debugger and language service workers. Just extracted all of that to some new helper functions (with separate Node.js and browser implementations where needed).


npm/src/browser.ts
minestarks, line 1: I ran "Organize Imports" on most of the files, which is why you'll see a lot of imports moved around. Sorry!

cesarzc (approved):
I don't have a lot of experience with TS, but I really like the new organization. Extracting the worker logic in a single place really helped readability too! Now it's easier to see how the components work because the boilerplate code is gone :)

billti (approved):

PR #1110, opened 2024-02-02
Prefer open file contents to disk contents

billti:
This seems like what we should be doing anyway, but has the benefit that when using the language service or the debugger in a multi-file project, then everything just works even if the files haven't been saved to disk yet.

Putting up early for review, but in initial testing looked to behave as expected.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -397,7 +397,12 @@
           }
         }
       }
-    ]
+    ],
+    "configurationDefaults": {
+      "[qsharp]": {
+        "debug.saveBeforeStart": "none"
+      }
+    }
   },
   "scripts": {
     "build": "npm run tsc:check && node build.mjs",
--- a/vscode/src/projectSystem.ts
+++ b/vscode/src/projectSystem.ts
@@ -53,6 +53,9 @@ async function findManifestDocument(
   // vscode-vfs://github%2B7b2276223a312c22726566223a7b2274797065223a332c226964223a22383439227d7d/microsoft/qsharp/samples/shor.qs
   const currentDocumentUri = URI.parse(currentDocumentUriString);
 
+  // Untitled documents don't have a file location, thus can't have a manifest
+  if (currentDocumentUri.scheme === "untitled") return null;
+
   // just the parent
   // e.g.
   // file://home/foo/bar/src
@@ -64,17 +67,12 @@ async function findManifestDocument(
 
   while (attempts > 0) {
     attempts--;
-    // we abort this check if we are going above the current VS Code
-    // workspace. If the user is working in a multi-root workspace [1],
-    // then we do not perform this check. This is because a multi-
-    // root workspace could contain different roots at different
-    // levels in each others' path ancestry.
-    // [1]: https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces
+
+    // Make sure that the path doesn't go above one of the open workspaces.
     if (
-      // TODO: It's perfectly valid to have > 1 workspace folders open in VS Code
-      vscode.workspace.workspaceFolders?.length === 1 &&
-      Utils.resolvePath(vscode.workspace.workspaceFolders[0].uri, "..") ===
-        uriToQuery
+      !vscode.workspace.workspaceFolders?.some((root) =>
+        uriToQuery.toString().startsWith(root.uri.toString()),
+      )
     ) {
       log.debug("Aborting search for manifest file outside of workspace");
       return null;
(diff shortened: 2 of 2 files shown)


vscode/src/projectSystem.ts
```diff
@@ -145,7 +162,12 @@ async function readFileUri(
       };
     });
   } catch (_err) {
-    // `readFile` returns `err` if the file didn't exist.
+    // `readFile` should throw the below if the file is not found
+    if (
+      !(_err instanceof vscode.FileSystemError && _err.code === "FileNotFound")
```
billti, line 167: Drive by improvement to log an error if unexpected, rather than swallow all.

vscode/src/projectSystem.ts
```diff
@@ -184,6 +206,12 @@ let projectLoader: any | undefined = undefined;
 export async function loadProject(
   documentUri: vscode.Uri,
 ): Promise<[string, string][]> {
+  // If the file is not even named yet, it won't be on disk and can't be in a project
+  if (documentUri.scheme === "untitled") {
```
billti, line 210: Not necessary for this change. Just a minor optimization to avoid some work if the file isn't backed by disk yet.

vscode/src/projectSystem.ts
```diff
@@ -137,6 +138,22 @@ async function readFileUri(
   const uri: vscode.Uri = (maybeUri as any).path
     ? (maybeUri as vscode.Uri)
     : vscode.Uri.parse(maybeUri as string);
+
+  // If any open documents match this uri, return their contents instead of from disk
+  const opendoc = vscode.workspace.textDocuments.find(
+    (opendoc) =>
+      opendoc.uri.scheme === uri.scheme &&
+      opendoc.uri.authority === uri.authority &&
+      opendoc.uri.path == uri.path,
```
minestarks, line 147: Is there a subtle reason this is a double equals while the other ones are triple?

vscode/src/projectSystem.ts
```diff
@@ -71,6 +71,7 @@
     // levels in each others' path ancestry.
     // [1]: https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces
     if (
+      // TODO: It's perfectly valid to have > 1 workspace folders open in VS Code
```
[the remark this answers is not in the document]
minestarks: What's the effect of the bug implied by this TODO? I think it would be better to track in an actual GitHub issue.

vscode/package.json
```diff
@@ -397,7 +397,12 @@
           }
         }
       }
-    ]
+    ],
+    "configurationDefaults": {
+      "[qsharp]": {
+        "debug.saveBeforeStart": "none"
```
minestarks, line 403: Instead of making this configurable, I think we can pass in `suppressSaveBeforeStart` with the `DebugSessionOptions` object when we call `startDebugging`. I tend to think that would be simpler, what do you think?

vscode/src/projectSystem.ts
billti: Yes, because I'm an idiot. (I guess that's not that subtle). Thanks for the catch! (Surprised ESLint didn't flag it).

In reality, I could probably just do `opendoc.toString() === uri.toString()`, but I'm not sure if query and fragment might have values at some point that should be ignored, so these seems like the right fields of a URI to worry about for comparison.

vscode/package.json
billti: Either way. I think by doing it here it allows the user to set it if they'd prefer to save on launch (similar to have we have `"editor.formatOnSave": true` in our .vscode settings for some languages. If we set it in the session, the user can't override.

billti: RE the TODO: Plan was to fix that before I'd merge, just wanted to get the PR up for feedback last night. Will fix in this PR before merging.

vscode/src/projectSystem.ts
minestarks: I've been doing the `uri1.toString() === uri2.toString()` comparison elsewhere in the code, and it hasn't caused me trouble so far. But I think this is good too. Surprised they didn't put in an `equals` method.

Yeah, ESLint has a rule for triple equals (`eqeqeq`) but it's not included in the recommended set.

I personally have conditioned myself to *always* use `===` since I don't trust myself to remember when it's okay to use `==`, it always surprises me. I wouldn't mind enabling the lint rule to enforce this, but that discussion is for another time...

billti: Switched to comparing the toString() results.

minestarks (approved):

PR #1132, opened 2024-02-07
Check for duplicate intrinsics globally

### LINKED ISSUE #1130: Two `body intrinsic` callables with the same name but different signature can cause panic in evaluator
Because `body intrinsic` callables must have globally unique names (not including namespace) but nothing today enforces this, it is possible to define a new intrinsic with the same name as an existing one but a different signature, resulting in a panic during evaluation. For example:

[Playground repro:](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAE02OPW%2FCUAxF9%2FyKK2WBhex0aSuqigEB%2FVAHxOAkBiwlfpGfn6oI8d%2BrhILweHx9fYqiQI5ParuGs6Io8M7uokdEJ3OuBzbyHAuOlUnnEnQkXyeJkAhCKyotNdjm6CwcjVr4iRwVKUpGilzDw7USvybjg22OKtQ8y5Rajh1VjFW%2FTaSe2s1%2FyznLAOD5Td36TRD1yXQkoWOjwQQrEp1MMccHx9T4bo%2FzmBhmkFwv1nP8mDijD8lub3Fi49k9uVR%2FiYuQyoYnB2oiT5%2FuO2NPptjtr%2BhydTokrUaBx9MSc7yG0Aw%2B3yr%2B4FKGuoeom2iU6lZ1yf4Axkxy8oEBAAA%3D)
image

The fix here is likely a new pass that enforces unique naming among all intrinsic callables in the current compilation.

swernli:
This adds new logic to name resolution to verify that any callable declared as `body intrinsic` has a globally unique name.

This fixes #1130.

### DIFF
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -6,7 +6,7 @@ mod tests;
 
 use miette::Diagnostic;
 use qsc_ast::{
-    ast::{self, CallableDecl, Ident, NodeId, TopLevelNode},
+    ast::{self, CallableBody, CallableDecl, Ident, NodeId, SpecBody, SpecGen, TopLevelNode},
     visit::{self as ast_visit, walk_attr, Visitor as AstVisitor},
 };
 use qsc_data_structures::{index_map::IndexMap, span::Span};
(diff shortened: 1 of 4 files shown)


compiler/qsc_frontend/src/resolve.rs
```diff
@@ -881,29 +897,42 @@ fn bind_global_item(
...
             names.insert(decl.name.id, res);
+            let mut errors = Vec::new();
             match scope
                 .terms
                 .entry(Rc::clone(namespace))
                 .or_default()
                 .entry(Rc::clone(&decl.name.name))
             {
-                Entry::Occupied(_) => Err(Error::Duplicate(
+                Entry::Occupied(_) => errors.push(Error::Duplicate(
```
minestarks, line 914: Why accumulate these two errors instead of just returning the first one? If we already saw the name is duplicated, is it worth _also_ checking for duplicate intrinsic?

minestarks (approved):

compiler/qsc_frontend/src/resolve.rs
swernli: I had to think about this for a while, as I forgot what my motivation was when I did the update this way. It turns out that a callable can be both a name collision (if the fully qualified namespace.name is a direct match) and an intrinsic collision (if declared as `body intrinsic` and another intrinsic in ANY namespace has the same name). This can be true even in the case where the duplicate intrinsic is an internal from a distinct namespace. I didn't want either error to mask the other, so I wanted to make sure I could report both.

swernli: I added a test case to show that both errors can be returned if the callable name violates each rule separately.

PR #1139, opened 2024-02-09
Superposition kata: add first 4 tasks

tcNickolas:
Includes minor refactoring of common testing operations


npm/test/basics.js
```diff
@@ -308,6 +308,11 @@ test("multi_qubit_gates kata is valid", async () => {
   await validateKata(kata, true, true, true);
 });
```
cesarzc, line 310:
```suggestion

// This test is commented because the superposition kata is still a work in progress.
```

cesarzc: Small suggestion to provide context on why the test is commented.

cesarzc (approved):

PR #1140, opened 2024-02-09
Add infrastructure for linting and creating new Lints

orpuente-MS:
This PR adds infrastructure for linting and creating new Lints. It also adds integration of the linter to the language service.

# End user experience

The users can set the lint level from the `qsharp.json` file as follows,

```json
{
  "lints": [
    {
      "lint": "needlessParens",
      "level": "allow",
    },
    {
      "lint": "deadCode",
      "level": "warn",
    },
    {
      "lint": "divisionByZero",
      "level": "error",
    }
  ]
}
```

# Contributor experience

The entry points to the linter is the `run_lints` function, which takes a `qsc_frontend::CompileUnit` as input and outputs a `Vec<Lint>`.

## Example

```rust
use linter::run_lints;;
use qsc::compile::compile;

let unit: CompileUnit = compile(...);

// The second argument is an optional user configuration.
let lints: Vec<Lint> = run_ast_lints(&package, None);
```

## How to add a new Lint

We can add a new lint in two steps:
 1. Declaring the lint: here we set the lint name, the default `LintLevel`, the message, and the help text the user will see.
 2. Implementing the lint: here we write the pattern matching logic of the new lint.

Below is a full example of how to add a new AST lint.

## Example

First, we add our lint to `src/lints/ast.rs`.
```rust
declare_ast_lints! {
  ...
  (DoubleParens, LintLevel::Warn, "unnecesary double parentheses", "remove extra parentheses for clarity"),
}
```

Then we implement the right `LintPass` for our new lint, in this case `linter::ast::AstLintPass`

```rust
impl linter::ast::AstLintPass for DoubleParens {
    // we only need to impl the relevant check_* method, all the other ones
    // will default to an empty method that will get optmized by rust
    fn check_expr(&self, expr: &qsc_ast::ast::Expr, buffer: &mut Vec<Lint>) {
        // we match the relevant pattern
        if let ExprKind::Paren(ref inner_expr) = *expr.kind {
            if matches!(*inner_expr.kind, ExprKind::Paren(_)) {
                // we push the lint to the buffer
                push_lint!(self, expr.span, buffer);
            }
        }
    }
}
```


compiler/qsc_linter/src/linter/hir.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.check_qubit_init(init);
+        visit::walk_qubit_init(self, init);
+    }
+
+    fn visit_ident(&mut self, ident: &'a Ident) {
+        self.check_ident(ident);
+    }
+}
+
+#[allow(unused_variables)]
```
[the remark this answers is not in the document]
orpuente-MS: I think this is a legitimate use case for `#[allow(unused_variables)]` since it is part of the design of this trait. I want to provide default empty implementation for these methods, but I don't want rust-analyzer to show the parameter names as `_block` and `_buffer` when future Lint writers are adding new Lints.

swernli:
I think the pattern used to integrate the lints into compilation and gather the results is good, but I think we want to integrate this a bit higher in the flow. Instead of being part of the frontend, the code there can be moved to compiler/qsc/src/compile.rs, that way it doesn't prevent the passes from running. We get a lot of error information from the passes and wouldn't want lint warnings suppressing those (if you look on [this line](https://github.com/microsoft/qsharp/blob/main/compiler/qsc/src/compile.rs#L45), you'll see we only run passes if the compile unit has no errors as determined by the length of the errors list). With that change, compiler/qsc_frontend/src/compile.rs would effectively be reverted and these same changes would be made to the `qsc::compile::compile` function and `qsc::compile::ErrorKind`. From there, I think we also want to skip adding the lint errors to the same list returned from compile for similar reasons: we don't want callers to treat them as errors. 

I'm not sure exactly how to return the lints, but here are a couple options you could consider and try out:

1. Updating the return type from the function to produce 3-tuple `(CompileUnit, Vec<Error>, Vec<Error>)` and then update all the callers to know the third entry is the output from the linter.
2. Introduce a new type to use in the 3-tuple rather than add lints to the existing error kind, so that `(CompileUnit, Vec<Error>, Vec<Lint>)` type more clearly separates lints.
3. Leave the signature unchanged but add the lints to a new field in the `CompileUnit` such that consumers can choose when to read the lints or not.

The downside of any of these is that some extra work will need to be done in the language service to display lints in addition to the errors rather than being able to piggyback off of the existing handling. But I think that's worth it compared to the price of updating all callers (and tests) that check if the errors list is empty to instead filter the list between lints and other errors.

orpuente-MS:
> I think the pattern used to integrate the lints into compilation and gather the results is good, but I think we want to integrate this a bit higher in the flow. Instead of being part of the frontend, the code there can be moved to compiler/qsc/src/compile.rs, that way it doesn't prevent the passes from running. We get a lot of error information from the passes and wouldn't want lint warnings suppressing those (if you look on [this line](https://github.com/microsoft/qsharp/blob/main/compiler/qsc/src/compile.rs#L45), you'll see we only run passes if the compile unit has no errors as determined by the length of the errors list). With that change, compiler/qsc_frontend/src/compile.rs would effectively be reverted and these same changes would be made to the `qsc::compile::compile` function and `qsc::compile::ErrorKind`. From there, I think we also want to skip adding the lint errors to the same list returned from compile for similar reasons: we don't want callers to treat them as errors.
> 
> I'm not sure exactly how to return the lints, but here are a couple options you could consider and try out:
> 
> 1. Updating the return type from the function to produce 3-tuple `(CompileUnit, Vec<Error>, Vec<Error>)` and then update all the callers to know the third entry is the output from the linter.
> 2. Introduce a new type to use in the 3-tuple rather than add lints to the existing error kind, so that `(CompileUnit, Vec<Error>, Vec<Lint>)` type more clearly separates lints.
> 3. Leave the signature unchanged but add the lints to a new field in the `CompileUnit` such that consumers can choose when to read the lints or not.
> 
> The downside of any of these is that some extra work will need to be done in the language service to display lints in addition to the errors rather than being able to piggyback off of the existing handling. But I think that's worth it compared to the price of updating all callers (and tests) that check if the errors list is empty to instead filter the list between lints and other errors.

Nice catch. I think I like options 2 & 3. But before I move forward, I noticed `qsc::compile::compile(...)` does something very weird. It drains the `CompileUnit.errors` vector into an external vector and it returns a (CompileUnit, Vec<Error>) tuple where the `CompileUnit.errors` vector is effectively empty. Does this mean `CompileUnit` isn't meant to have a `CompileUnit.errors` field in the first place?

swernli:
> Nice catch. I think I like options 2 & 3. But before I move forward, I noticed `qsc::compile::compile(...)` does something very weird. It drains the `CompileUnit.errors` vector into an external vector and it returns a `(CompileUnit, Vec<Error>)` tuple where the `CompileUnit.errors` vector is effectively empty. Does this mean `CompileUnit` isn't meant to have a `CompileUnit.errors` field in the first place?

Hmm... good question. I think that this is an example of two different design philosophies merging. Throughout the internals of the compiler, the field is used as a way to collect errors associated with that unit (all the tests read from that field, for example). But at the boundary of where we expose it out of the compiler the errors are treated as separate data, and to avoid copying the vector and having two instances it's drained. There's definitely an opportunity for refactor here, where `qsc::compile::compile` is updated to just return the `CompileUnit` and callers read the errors from inside of there. Not sure if you want to pull that into the scope of this PR or not, but it might make sense as a pattern update where callers look at `CompileUnit.errors` and `CompileUnit.lints` directly.

compiler/qsc_linter/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+//!         // ... some other lints
+//!
+//!         DoubleParens::check_expr(expr); // add your new lint here
+//!     }
+//! }
+//! ```
+
+#![deny(missing_docs)]
+#![warn(clippy::mod_module_files, clippy::pedantic, clippy::unwrap_used)]
+#![allow(clippy::missing_errors_doc, clippy::missing_panics_doc)]
```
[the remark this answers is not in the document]
swernli: we haven't started using the global clippy settings yet (I think that was introduced in 1.76, and our repo currently only required 1.75). So this snippet:
```rust
#![warn(clippy::mod_module_files, clippy::pedantic, clippy::unwrap_used)]
#![allow(clippy::missing_errors_doc, clippy::missing_panics_doc)]
```
Is repeated in every top-level lib.rs file. I'm hoping we can switch to the global settings eventually, but that alone didn't seem like reason to force 1.76. In the meantime, these two lines need to appear at the top of the file to conform to our current pattern of clippy settings.

For the `deny(missing_docs)`, I don't mind trying it out here and then tackling a follow up action to make that a common part of the codebase. One more reason to prefer global settings eventually: applying that more broadly right now will require adding it to each module manually rather than one central location.

compiler/qsc_linter/src/tests.rs
```diff
@@ -0,0 +1,54 @@
...
+    let lints = run_ast_lints(&package);
+
+    for lint in &lints {
+        let range = line_column::Range::from_span(line_column::Encoding::Utf8, source, &lint.span);
+        let chunk = get_lines(source, range);
+        for line in chunk {
+            println!("{line}");
+        }
+        println!("{lint:?}");
+    }
```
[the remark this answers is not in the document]
orpuente-MS: Wow, this is amazing! I'm doing it now.

compiler/qsc_linter/src/linter/hir.rs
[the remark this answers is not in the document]
orpuente-MS: There is one drawback to this approach. When you use trait objects and call one of those methods, rust-analyzer will show you the following as a hint `my_trait_object.custom_intrinsic(_name: "W-gate", _arg: 7)`, "suggesting" to the user that `_name` and `_arg` aren't being used.

orpuente-MS:
I realized that computing the lints at `qsc::compile::compile(...)` might be inefficient, because that function is called in a lot of places, however it only makes sense to compute the lints if it is being called from the language server. So, I have come up with the following solution:
 1. `qsc::compile::compile(...)` will remain unchanged, it returns a `(Unit, Vec<Error>)`.
 2. `Unit` will have a method `Unit::lints(&self) -> Vec<Lint>` that will compute & return the lints. It makes sense for it to be a method of Unit because "computing the lints" is an operation that is only applied to a compiled unit.

swernli:
> I realized that computing the lints at `qsc::compile::compile(...)` might be inefficient, because that function is called in a lot of places, however it only makes sense to compute the lints if it is being called from the language server. So, I have come up with the following solution:
> 
> 1. `qsc::compile::compile(...)` will remain unchanged, it returns a `(CompileUnit, Vec<Error>)`.
> 2. `CompileUnit` will have a method `CompileUnit::lints(&self) -> Vec<Lint>` that will compute & return the lints. It makes sense for it to be a method of `CompileUnit` because "computing the lints" is an operation that is only applied to a compiled unit.

Oh, good point. I like this approach, and makes sense given that lints are optional and only those who want them will pay the cost for generating them.

compiler/qsc_linter/src/linter/ast.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    fn check_functor_expr(&self, _expr: &FunctorExpr) {}
+
+    fn check_ty(&self, _ty: &Ty) {}
+
+    fn check_block(&self, _block: &Block) {}
+
+    fn check_stmt(&self, _stmt: &Stmt) {}
+
+    fn check_expr(&mut self, expr: &Expr) {
```
ScottCarda-MS, line 68: I could put any function or logic in these that I want, right? In general each lint listed here is a function that checks the given ast node, and puts a Lint in the self.buffer if needed, right?

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -0,0 +1,38 @@
...
+use qsc_ast::ast::{BinOp, ExprKind, Lit};
+
+declare_lint!(
+    DoubleParens,
+    LintLevel::Warning,
+    "unnecesary double parentheses"
+);
+declare_lint!(
+    DivisionByZero,
+    LintLevel::Allow,
```
ScottCarda-MS, line 12: Should division by zero be an error?

compiler/qsc_linter/src/linter.rs
```diff
@@ -0,0 +1,67 @@
...
+        let source_span = miette::SourceSpan::from(self.span);
+        let labeled_span = LabeledSpan::new_with_span(Some(self.to_string()), source_span);
+        Some(Box::new(vec![labeled_span].into_iter()))
+    }
+}
+
+/// A lint level. This defines if a lint will be treated as a warning or an error,
+/// and if the lint level can be overriden by the user.
+#[derive(Debug, Clone)]
+pub enum LintLevel {
```
ScottCarda-MS, line 44: Is ForceWarning and ForceError common among other linters? If there is not a precedent for this, it doesn't seem like something that should be used, as users should be able to control how to treat the lints.

compiler/qsc_linter/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+miette = { workspace = true }
+qsc_ast = { path = "../qsc_ast" }
+qsc_fir = { path = "../qsc_fir" }
```
ScottCarda-MS, line 14: Can the fir dependency be removed? I didn't see the fir being used, but maybe I missed it.

compiler/qsc_linter/src/tests.rs
```diff
@@ -0,0 +1,115 @@
...
+                    message: "unnecesary double parentheses",
+                    level: Warning,
+                },
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn multiple_lints() {
```
ScottCarda-MS, line 27: For completeness, could you add a test that _only_ test the division by zero lint?

compiler/qsc_linter/src/linter/hir.rs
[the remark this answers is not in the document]
orpuente-MS: Never mind. I just realized rust-analyzer autocompletes it correctly even if it has _ as a prefix

compiler/qsc_linter/src/linter.rs
orpuente-MS: Yes, for example, it is used in the rust linter, for example, to force future deprecation lints, some stylistic conventions, and things like dead-code.

compiler/qsc_linter/Cargo.toml
orpuente-MS: Nice catch

compiler/qsc_linter/src/lints/ast.rs
orpuente-MS: I'm not sure. When I showed it on the demo I think Dmitry said something about it being possible to do in Q#. I just left it there in case the users wanted to enable that lint manually.

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -0,0 +1,36 @@
+use super::{declare_lint, push_lint};
+use crate::linter::{ast::AstLintPass, Lint, LintLevel};
+use qsc_ast::ast::{BinOp, ExprKind, Lit};
+
+declare_lint!(
+    DoubleParens,
+    LintLevel::Warning,
+    "unnecesary double parentheses"
```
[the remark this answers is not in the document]
orpuente-MS: Good point. Changed the message to `"unnecessary parentheses"`, but left the Lint name as `DoubleParens` since that's what it matches against in the AST, is that reasonable?

compiler/qsc_linter/src/tests.rs
```diff
@@ -0,0 +1,131 @@
...
+use qsc_data_structures::language_features::LanguageFeatures;
+
+#[test]
+fn multiple_lints() {
+    check_ast(
+        "let x = ((1 + 2)) / 0;",
+        &expect![[r#"
+        [
+            SrcLint {
+                source: "((1 + 2)) / 0",
```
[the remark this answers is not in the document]
orpuente-MS: That can be done with another Lint that matches against unnecessary parentheses instead of extra set of parentheses. For example, Clippy has two lints, one called `DoubleParens` and another called `NeedlessParens`. We could follow the same pattern here.

compiler/qsc_linter/src/linter/ast.rs
```diff
@@ -0,0 +1,185 @@
+use qsc_ast::{
```
billti, line 1: You'll need to include the standard MIT license header in these new files (see other files for reference)

orpuente-MS: Nice catch! Fixed. I wonder if there is way to make the compilation fail if those are missing @swernli

compiler/qsc_linter/src/tests.rs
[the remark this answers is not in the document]
orpuente-MS: Added `NeedlessParens` lint.

orpuente-MS:
## Reduced boilerplate when adding a new lint.

First, we add our lint to `src/lints/ast.rs`.
```rust
declare_ast_lints! {
  ...
  (DoubleParens, LintLevel::Warn, "unnecesary double parentheses"),
}
```

Then we implement the right `LintPass` for our new lint, in this case `linter::ast::AstLintPass`

```rust
impl linter::ast::AstLintPass for DoubleParens {
    // we only need to impl the relevant check_* method, all the other ones
    // will default to an empty method that will get optmized by rust
    fn check_expr(expr: &qsc_ast::ast::Expr, buffer: &mut Vec<Lint>) {
        // we match the relevant pattern
        if let ExprKind::Paren(ref inner_expr) = *expr.kind {
            if matches!(*inner_expr.kind, ExprKind::Paren(_)) {
                // we push the lint to the buffer
                push_lint!(Self, expr.span, buffer);
            }
        }
    }
}
```

Previously we also needed to add our new lint to `impl CombinedAstLints { ... }` in `src/linter/ast.rs`.

```
impl CombinedAstLints {
    fn check_expr(&mut self, expr: &Expr) {
        // ... some other lints

        DoubleParens::check_expr(expr, &mut self.buffer); // add your new lint here
    }
}
```

Now, this logic is included in the `declare_ast_lints!` macro.

compiler/qsc_linter/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+serde = { workspace = true }
+thiserror = { workspace = true }
+
+[dev-dependencies]
+expect-test = { workspace = true }
+qsc_parse = { path = "../qsc_parse" }
+serde_json = { workspace = true }
+
+[lib]
+doctest = false
```
swernli, line 25: I know I had just told you to enable lints for the project with explicit clippy module attributes in the lib.rs, but as of https://github.com/microsoft/qsharp/pull/1224 you can now just inherit the lint settings from the workspace by adding a few lines here:

```suggestion
[lints]
workspace = true

[lib]
doctest = false
```

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use super::push_lint;
+use crate::linter::ast::declare_ast_lints;
+use qsc_ast::ast::{BinOp, ExprKind, Lit, StmtKind};
+use qsc_data_structures::span::Span;
+
+declare_ast_lints! {
+    (DivisionByZero, LintLevel::Allow, "attempt to divide by zero"),
+    (NeedlessParens, LintLevel::Warning, "unnecessary parentheses"),
+    (RedundantSemicolons, LintLevel::Warning, "redundant semicolons"),
```
swernli, line 12: I think it would be helpful to include miette-style diagnostic help like we do for other errors. Lints in particular benefit from some additional explanatory text for the user. For example, "redundant semicolons" could come with the additional help text, "remove the redundant semicolons for clarity."

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -0,0 +1,15 @@
+use crate::linter::hir::declare_hir_lints;
+
+use super::push_lint;
+
+declare_hir_lints! {
+    (Stump, LintLevel::Allow, "remove this stump after addding the first HIR lint"),
```
swernli, line 6: I know this is just a placeholder, but to confirm HIR lints show up it would be good to include in one of your unit tests.

orpuente-MS: Just added a unit test for the placeholder, and there was indeed a bug. Nice catch!

compiler/qsc/src/compile.rs
```diff
@@ -27,6 +27,7 @@ pub enum ErrorKind {
     /// These errors are typically related to optimization, transformation, code generation, passes,
     /// and static analysis passes.
     Pass(#[from] qsc_passes::Error),
+    Lint(#[from] qsc_linter::Lint),
```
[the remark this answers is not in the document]
orpuente-MS: Fixed

compiler/qsc_linter/src/lints/ast.rs
orpuente-MS: Yes, you are right. I just added help messages to the lints

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -0,0 +1,117 @@
...
+        ExprKind::BinOp(op, _, _) => match op {
+            BinOp::Exp => 3,
+            BinOp::Div | BinOp::Mod | BinOp::Mul => 4,
+            BinOp::Add | BinOp::Sub => 5,
+            BinOp::Shl | BinOp::Shr => 6,
+            BinOp::Gt | BinOp::Gte | BinOp::Lt | BinOp::Lte => 7,
+            BinOp::Eq | BinOp::Neq => 8,
+            BinOp::OrB | BinOp::XorB | BinOp::AndB => 9,
+            BinOp::OrL | BinOp::AndL => 10,
+        },
```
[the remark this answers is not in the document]
billti: Also, do we warn on any parens that are not needed due to precedence? e.g. would `(2 ^ i) * 5` cause a warning? If so, I think we should have these off by default (not warn), as a lot of "coding guidelines" encourage parens for clarity to help those mere mortals who haven't memorized the precedence table :-)

Also, logical OR and AND are different precedence (line 112)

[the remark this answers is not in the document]
orpuente-MS: I fixed the precedence following your suggestion Stefan, there were indeed some errors.

Bill, you are right, I myself add extra parentheses sometimes due to precedence paranoia. I went ahead and set the level to Allow by default. We can have a pedantic lint group later where we enable it.

samples/language/MultiFileProject/src/Main.qs
```diff
@@ -13,7 +13,7 @@ namespace MyQuantumApp {
   operation Main() : Unit {
     let particleA = Particle(0, 0, 0);
     let particleB = Particle(1, 1, 1);
-
+
```
minestarks, line 16: Did this sneak in by accident?

```suggestion

```

wasm/src/project_system.rs
```diff
@@ -187,14 +188,41 @@ pub(crate) fn get_manifest_transformer(js_val: JsValue, _: String) -> Option<Man
 
     };
 
+    let lints: Vec<LintConfig> = match js_sys::Reflect::get(&js_val, &JsValue::from_str("lints")) {
```
minestarks, line 191: I'm not 100% sure but we should be able to simplify all this manual conversion code by declaring `Serializable` structs and just letting wasm-bindgen take care of the conversions. I'll try to come up with some sample code - the docs are pretty nonobvious.

language_service/src/compilation.rs
```diff
@@ -69,6 +72,13 @@ impl Compilation {
             language_features,
         );
 
+        let lints = qsc::linter::run_lints(&unit, Some(lints_config));
```
minestarks, line 75: We should do this for notebooks as well (though, feel free to open a new issue/PR for this so as not to bloat this one any further). It should be super easy - just insert a corresponding call into the `new_notebook` function after we build up the `CompileUnit` for the notebook.

compiler/qsc_project/tests/tests.rs
```diff
@@ -36,6 +36,7 @@ fn basic_manifest() {
                     ),
                     license: None,
                     language_features: [],
+                    lints: [],
```
billti, line 39: Does this need to be specified if you're not specifying anything? Ideally not, I assume.

Or if this is NOT testing actually settings in qsharp.json take effect, we should add tests for that.

wasm/src/project_system.rs
```diff
@@ -187,14 +188,41 @@ pub(crate) fn get_manifest_transformer(js_val: JsValue, _: String) -> Option<Man
...
+                    let lint = lint.as_string().expect("manifest callback returned {lint:?}, but we expected a string representing a lint name");
+                    let level = level.as_string().expect("manifest callback returned {level:?}, but we expected a string representing a lint level");
+
+                    let kind = format!(r#""{lint}""#);
+                    let level = format!(r#""{level}""#);
+                    let kind: LintKind = serde_json::from_str(&kind).ok()?;
+                    let level: LintLevel = serde_json::from_str(&level).ok()?;
+
+                    Some(LintConfig { kind, level })
+                })
```
minestarks, line 209: OK, since you already declared `LintConfig` as `Serializable`, this works quite nicely. I tested this:

```suggestion
                .filter_map(|x| serde_wasm_bindgen::from_value::<LintConfig>(x).ok())
```

I'm pretty sure the same could be applied to this entire function (since all of the `Manifest` struct is serializable) but maybe it's best not to risk it right now.

compiler/qsc_linter/src/lints.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub(super) mod ast;
+pub(super) mod hir;
+
+macro_rules! push_lint {
+    ($lint:expr, $span:expr, $buffer:expr) => {
+        $buffer.push(Lint {
```
minestarks, line 9: Just personal preference but I think this macro obfuscates a little too much. I'd find it easier to follow `buffer.push(lint!(self, expr.span))` at the call site, and it's not any longer than writing `push_lint!(self, expr.span, buffer)`.

minestarks: I should have clarified that the reason I'm averse to macros is that the editor experience (code navigation, hover info, parameter hints etc) for macros tends to be pretty terrible, so in general I think they should only be used if they really help save _a lot_ of code.

minestarks (approved):

compiler/qsc_project/tests/tests.rs
orpuente-MS: Just added a unit-test to `language_service/src/state/tests.rs` to make sure the lints_config are getting loaded from the manifest.

compiler/qsc_linter/src/lints.rs
orpuente-MS: Yes, I agree with your preference haha.  Just changed it as you suggested.

wasm/src/project_system.rs
orpuente-MS: Yes, this indeed works!

language_service/src/state/tests.rs
```diff
@@ -1252,6 +1253,66 @@ async fn doc_switches_project_on_close() {
...
+
+    let received_errors = RefCell::new(Vec::new());
+    let updater = new_updater_with_file_system(&received_errors, &fs);
+
+    // Load the manifest.
+    let manifest = updater
+        .load_manifest(&"project/src/this_file.qs".into())
+        .await
+        .expect("manifest should exist");
+
```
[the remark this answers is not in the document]
orpuente-MS: Changed the test to check the lints. I would need to extend the in-memory file system to modify the manifest. I'll take a look at it tomorrow.

swernli (approved):

language_service/src/state/tests.rs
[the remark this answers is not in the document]
orpuente-MS: Done!

language_service/src/compilation.rs
orpuente-MS: Filed an issue as a remainder.

PR #1149, opened 2024-02-12
Fix panics on In-place update optimization

### LINKED ISSUE #1146: In-place update optimization can cause panic on array of arrays
The optimization introduced in https://github.com/microsoft/qsharp/pull/874 used an assumption that in-place operations will only happen on arrays that are uniquely referenced. However, if trying to use `w/=` syntax on a mutable array that is itself a member of another mutable array, the check can fail and cause a panic.

See for example this code ([Playground link](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAE32PwU7CQBCG732KP%2BkFEm3rEZTEA8YTAdTEQ8NhKCMMaXeb2W0IEt7dbAW1kDi3%2FWe%2Bb2bTNEWMV6rqkqM0TfHM3otZw3lSz6uQtXmMMbtCpfZiTZu8bcRBHAiVGKmoxDxGrXatVMFvyKMggyWjcbyCt99K7FTaBfMYhV1xEhmq2NVUMCb7eUPGN9XsZDlEEQA8Phmv%2B5kV43v9NrE1K4VLMCExvT6GeGHXlD5f4NBOhApHTsfTId5VPGNvGz2vxYaVk5%2FJXyZU1XhalgxSxQh5nt3AySdjhLssW3Qe9x3wwyoEYpAlyWBwYf1rFmM4uEk1lwvJWbT9TxTKsT%2BJdukIWzzcYnutOl4lgQtfC5QEqpV0yS5FqtF1R9k3apCf7j9Gx%2BgL2dWABk8CAAA%3D)):
```qsharp
mutable arr = [[0, size = 100], size = 100];
for i in 0..99 {
  mutable inner = arr[i];
    for j in 0..99 {
        set inner w/= j <- j;
    }
    set arr w/= i <- inner;
}
arr
```

Instead of panicking, the code should check the applicability of the optimization before choosing which path to go down.

swernli:
This checks ahead of time if the array we want to update in-place is uniquely referenced and correctly falls back to the previous, unoptimized path that makes a copy.

Fixes #1146

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -24,9 +24,9 @@ use num_bigint::BigInt;
 use output::Receiver;
 use qsc_data_structures::span::Span;
 use qsc_fir::fir::{
-    self, BinOp, BlockId, CallableImpl, ExprId, ExprKind, Field, Functor, Global, Lit, LocalItemId,
-    Mutability, NodeId, PackageId, PackageStoreLookup, PatId, PatKind, PrimField, Res, StmtId,
-    StmtKind, StoreItemId, StringComponent, UnOp,
+    self, BinOp, BlockId, CallableImpl, Expr, ExprId, ExprKind, Field, Functor, Global, Lit,
+    LocalItemId, Mutability, NodeId, PackageId, PackageStoreLookup, PatId, PatKind, PrimField, Res,
+    StmtId, StmtKind, StoreItemId, StringComponent, UnOp,
 };
 use qsc_fir::ty::Ty;
 use rand::{rngs::StdRng, SeedableRng};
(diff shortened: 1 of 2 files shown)


idavis (approved):

swernli:
> Do we have a formal/documented notion of the pass-by-value or pass-by-reference semantics of data structures in Q#?

All values are read-only in Q#, even on mutable variables. The expectation for mutable variables is that they are allowed to be updated to "point" to a new value, rather than the actual value being mutated. This is why array index updates are only possible with "copy-update" expressions which are expected to copy. But when looking for ways to improve performance, we made it so aggregate data types, Arrays and Tuples, are actually `Rc` of value, which means we can avoid the overhead of copying the items every time. This made it so things like:
```qsharp
let x = [1, 2, 3];
let y = x;
```
would still only have one copy of `[1, 2, 3]` in memory and each variable has a reference. The further optimization we tried to do was in the case where "assign-update" and "assign-add" are used to update a mutable array and use that as an opportunity to work in-place:
```qsharp
mutable arr = [1, 2, 3];
set arr += [4, 5, 6];
```
Such that the append (or index update) can happen without creating an entire copy of the source array. The bug was that we always did it, running into a panic if there was another reference:
```qsharp
mutable arr = [1, 2, 3];
let copy = arr;
set arr += [4, 5, 6]; // in-place update fails because the source is not uniquely owned
```
With this change, we've effectively made it so that uniquely owned arrays can be updated in-place, otherwise we fall back to the old behavior of copy-on-write.

PR #1160, opened 2024-02-14
Modify 'getAllKatas works' test to reduce friction when working on new katas

cesarzc:
This is meant to make it easier for people authoring on new katas.

### DIFF
--- a/npm/test/basics.js
+++ b/npm/test/basics.js
@@ -272,15 +272,9 @@ async function validateKata(
   }
 }
 
-test("all katas work", async () => {
+test("getAllKatas works", async () => {
   const katas = await getAllKatas();
-  // N.B. If you update the expected katas count, make sure to add a validation test for your newly added kata.
-  const expectedKatasCount = 9;
-  assert.equal(
-    katas.length,
-    expectedKatasCount,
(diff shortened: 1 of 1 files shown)


tcNickolas (approved):

npm/test/basics.js
```diff
@@ -272,15 +272,9 @@ async function validateKata(
...
+test("getAllKatas works", async () => {
   const katas = await getAllKatas();
-  // N.B. If you update the expected katas count, make sure to add a validation test for your newly added kata.
-  const expectedKatasCount = 9;
-  assert.equal(
-    katas.length,
-    expectedKatasCount,
-    `Expected ${expectedKatasCount} katas, but found ${katas.length} katas`,
-  );
+  assert.ok(katas.length > 0, "katas should not be empty");
```
DmitryVasilevsky, line 277: Since we already have 9 katas and we want to improve _addition_ of katas, consider comparing to 9 rather than 0. Removing katas would still be possible, but the test would have to be updated.

DmitryVasilevsky (approved):

billti (approved):

npm/test/basics.js
cesarzc: Seems a bit silly to have a seemingly random number so I'll leave 0 for now.

PR #1164, opened 2024-02-15
Implement Runtime Capabilities Analysis (RCA)

cesarzc:
Runtime Capabilities Analysis (RCA) is the process of determining the capabilities a quantum kernel needs to be able to run a particular program. This implementation also identifies program elements that can be pre-computed before execution on a quantum kernel and does not consider these elements when determining the capabilities. Additionally, this implementation provides details on why the program requires each capability.

One way to look at it is that RCA answers the following questions:
- What capabilities are needed to run a program?
- Which program elements require these capabilities and why?
- Which program elements can be computed before execution in a quantum kernel?


compiler/qsc_rca/src/rca.rs
```diff
@@ -0,0 +1,2340 @@
...
+            if let ValueKind::Dynamic = last_stmt_quantum_properties.value_kind {
+                // If the block's last statement's compute kind is quantum, the block's compute kind must be quantum too.
+                let ComputeKind::Quantum(block_quantum_properties) = &mut block_compute_kind else {
+                    panic!("block's compute kind should be quantum");
+                };
+
+                // The block's value kind must be static since this is the first time a value kind is set.
+                let ValueKind::Static = block_quantum_properties.value_kind else {
+                    panic!("block's value kind should be static");
+                };
```
swernli, line 2040: Since there is no assignment being used from the let-binding, this can just be an assert.
```suggestion
                assert!(block_quantum_properties.value_kind == ValueKind::Static, "block's value kind should be static");
```

compiler/qsc_rca/src/rca.rs
```diff
@@ -0,0 +1,2340 @@
...
+                    }
+                    input_param_exprs
+                }
+                _ => panic!("expected tuple expression"),
+            }
+        }
+    }
+}
+
+fn propagate_applications_table_through_expr(
```
swernli, line 1860: This recursive propagation through exprs, stmts, and blocks feels like it could have been implemented with the visitor pattern... have you tried that?

compiler/qsc_rca/src/common.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum SpecKind {
+    Body,
+    Adj,
+    Ctl,
+    CtlAdj,
+}
```
idavis, line 180: Should FunctorSetValue be used here. We are operating on HIR and these values are defined already for HIR.

compiler/qsc_rca/src/common.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub ty: Ty,
+    pub node: Option<NodeId>,
+}
+
+pub fn derive_callable_input_params(
+    callable: &CallableDecl,
+    pats: &IndexMap<PatId, Pat>,
+) -> Vec<InputParam> {
+    let input_elements = derive_callable_input_pattern_elements(callable, pats);
+    let mut input_params = Vec::new();
```
idavis, line 77: The length of `input_params` is known based on the number of `input_elements`. Initialize it with capacity: `let mut input_params = Vec::with_capacity(input_elements.len());`

compiler/qsc_rca/src/common.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+        ExprKind::Var(res, _) => match res {
+            Res::Item(item_id) => resolve_item_callee(package_id, *item_id),
+            Res::Local(node_id) => {
+                try_resolve_local_callee(*node_id, package_id, package, locals_map)
+            }
+            Res::Err => panic!("callee resolution should not be an error"),
+        },
+        // N.B. More complex callee expressions might require evaluation so we don't try to resolve them at compile
+        // time.
```
idavis, line 286: All comments are “Nota Bene” otherwise they wouldn’t be there. Prefixing comments with "N.B." is just noise and reduces readiblity.

compiler/qsc_rca/src/common.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+impl LocalsLookup for FxHashMap<NodeId, Local> {
+    fn find(&self, node_id: NodeId) -> Option<&Local> {
+        self.get(&node_id)
+    }
+}
+
+pub fn initalize_locals_map(input_params: &Vec<InputParam>) -> FxHashMap<NodeId, Local> {
```
idavis, line 141: `initialize_locals_map`

compiler/qsc_rca/src/scaffolding.rs
```diff
@@ -0,0 +1,296 @@
...
+};
+use qsc_data_structures::index_map::IndexMap;
+use qsc_fir::fir::{
+    BlockId, ExprId, LocalItemId, PackageId, PackageStore, StmtId, StoreBlockId, StoreExprId,
+    StoreItemId, StoreStmtId,
+};
+
+/// Scaffolding used to build the package store compute properties.
+#[derive(Debug, Default)]
+pub struct PackageStoreScaffolding(IndexMap<PackageId, PackageScaffolding>);
```
idavis, line 17: Given that all of these types are in the scaffolding module, it is redundant to suffix the types with the moddule name which already prefixes them.

compiler/qsc_rca/src/scaffolding.rs
```diff
@@ -0,0 +1,296 @@
...
+            .expect("block compute properties should exist")
+    }
+
+    fn get_expr(&self, id: StoreExprId) -> &ApplicationsTable {
+        self.find_expr(id)
+            .expect("expression compute properties should exist")
+    }
+
+    fn get_item(&self, _: StoreItemId) -> &ItemComputeProperties {
+        panic!("not implemented")
```
idavis, line 50: `notimplemented!()`

compiler/qsc_rca/src/scaffolding.rs
```diff
@@ -0,0 +1,296 @@
...
+            .get_mut(id.callable.package)
+            .expect("package should exist")
+            .items;
+        if let Some(item_scaffolding) = items.get_mut(id.callable.item) {
+            if let ItemScaffolding::Specializations(specializations) = item_scaffolding {
+                // The item already exists but not the specialization.
+                specializations.insert(SpecializationIndex::from(id.spec_kind), value);
+            } else {
+                panic!("item should be a callable");
+            }
```
idavis, line 134: Should be able to use let..else here.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+};
+use qsc_data_structures::index_map::IndexMap;
+use qsc_fir::fir::{BlockId, ExprId, NodeId, Pat, PatId, PatKind, SpecDecl, StmtId};
+use rustc_hash::FxHashMap;
+
+#[derive(Debug)]
+pub struct ApplicationInstancesTable {
+    pub inherent: ApplicationInstance,
+    pub dynamic_params: Vec<ApplicationInstance>,
+    is_settled: bool,
```
idavis, line 20: What does `is_settled` mean? From the code it looks like a flag indicating that the object has been fully populated.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+}
+
+/// An instance of a callable application.
+#[derive(Debug, Default)]
+pub struct ApplicationInstance {
+    /// A map of locals with their associated compute kind.
+    pub locals_map: LocalsComputeKindMap,
+    /// The currently active dynamic scopes in the application instance.
+    pub active_dynamic_scopes: Vec<ExprId>,
+    /// The return expressions througout the application instance.
```
idavis, line 230: throughout

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+            return_expressions: Vec::new(),
+            blocks: FxHashMap::default(),
+            stmts: FxHashMap::default(),
+            exprs: FxHashMap::default(),
+            is_settled: false,
+            was_flushed: false,
+        }
+    }
+
+    pub fn settle(&mut self) {
```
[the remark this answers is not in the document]
idavis: +1 It looks like we have two levels of settled, and I'm not sure how they interact/relate.

compiler/qsc_rca/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+            if *is_dynamic {
+                compute_kind = aggregate_compute_kind(compute_kind, dynamic_param_application);
+            }
+        }
+        compute_kind
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum ComputeKind {
```
idavis, line 320: `ComputeKind` (and its components) should implement `Copy`

compiler/qsc_rca/src/rca.rs
```diff
@@ -0,0 +1,2340 @@
...
+    for _ in 0..spec_functor.controlled {
+        let expr = package.get_expr(remainder_expr_id);
+        match &expr.kind {
+            ExprKind::Tuple(pats) => {
+                assert!(pats.len() == 2);
+                controls.push(pats[0]);
+                remainder_expr_id = pats[1];
+            }
+            _ => panic!("expected tuple expression"),
+        }
```
idavis, line 2298: let-else
```rust
        let ExprKind::Tuple(pats) = &expr.kind else {
            panic!("expected tuple expression")
        };
        assert!(pats.len() == 2);
        controls.push(pats[0]);
        remainder_expr_id = pats[1];
```

compiler/qsc_rca/src/applications.rs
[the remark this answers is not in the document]
cesarzc: In this context, settled was a way to convey that the data inside the structure should not change anymore.

Based on an offline conversation with @sezna, I changed the data structures in this file to use a builder pattern and the flags I was using to check things were being done in the right order and that no invalid operations were being performed are no longer needed.

I think the code is clearer now. Thank you @sezna for the feedback!

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+        }
+    }
+
+    fn clear_locals(&mut self) {
+        // Cannot clear locals until the application instance has been settled.
+        assert!(self.is_settled);
+        self.locals_map.clear();
+    }
+
+    fn mark_flushed(&mut self) {
```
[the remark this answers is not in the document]
cesarzc: This is no longer needed after the refactoring of this file based on our conversation a few days ago.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+
+    pub fn close(
+        &mut self,
+        package_scaffolding: &mut PackageScaffolding,
+        main_block: Option<BlockId>,
+    ) -> Option<ApplicationsTable> {
+        // We can close only if this structure is not yet settled and if all the internal application instances are
+        // already settled.
+        assert!(!self.is_settled);
+        assert!(self.inherent.is_settled);
```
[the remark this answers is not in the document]
cesarzc: This is no longer applicable after the refactoring.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+        let mut dynamic_param_applications =
+            Vec::<Option<ValueKind>>::with_capacity(self.dynamic_params.len());
+        for application_instance in self.dynamic_params.iter_mut() {
+            dynamic_param_applications.push(application_instance.collect_value_kind());
+        }
+
+        (inherent, dynamic_param_applications)
+    }
+
+    fn flush_compute_properties(&mut self, package_scaffolding: &mut PackageScaffolding) {
```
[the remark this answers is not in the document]
cesarzc: Flush meant saving to a package. For the refactoring, I used "save" instead of "flush" to avoid confusion.

compiler/qsc_rca/src/rca.rs
cesarzc: When I started implementing RCA, I tried using the visitor pattern. However, the way I was originally implementing it caused some conflicts with the borrow checker when using the visitor pattern.

I think now that I've implemented RCA, I can refactor it to use the visitor pattern. I'll work on it.

compiler/qsc_rca/src/common.rs
cesarzc: Yes. We should move it into a shared location and use it as you suggested for `FunctorApp`.

compiler/qsc_rca/src/common.rs
```diff
@@ -0,0 +1,354 @@
...
+pub struct Callee {
+    pub item: StoreItemId,
+    pub spec_functor: SpecFunctor,
+}
+
+#[derive(Debug, Default)]
+pub struct SpecFunctor {
+    pub adjoint: bool,
+    pub controlled: u8,
+}
```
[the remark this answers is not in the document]
cesarzc: Yes, I think moving it and re-using it is the way to go.

compiler/qsc_rca/src/common.rs
cesarzc: Well, `FunctorSetValue` probably does not need to be moved.

compiler/qsc_rca/src/rca.rs
```diff
@@ -0,0 +1,2221 @@
...
+}
+
+fn analyze_non_intrinsic_callable(
+    id: StoreItemId,
+    callable: &CallableDecl,
+    input_params: &Vec<InputParam>,
+    package_store: &PackageStore,
+    package_store_scaffolding: &mut PackageStoreScaffolding,
+) {
+    // This function is not meant for instrinsics.
```
idavis, line 242: intrinsics

compiler/qsc_rca/src/analyzer.rs
```diff
@@ -0,0 +1,90 @@
...
+/// A runtime capabilities analyzer.
+#[derive(Default)]
+pub struct Analyzer {
+    compute_properties: PackageStoreComputeProperties,
+}
+
+impl Analyzer {
+    /// Creates a new runtime capabilities analyzer for a package store. The provided package store is analyzed when a
+    /// new analyzer is created, which makes this a computationally intensive operation.
+    pub fn new(package_store: &PackageStore) -> Self {
```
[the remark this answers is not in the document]
cesarzc: Renamed.

compiler/qsc_rca/src/analyzer.rs
```diff
@@ -0,0 +1,90 @@
...
+        let package_compute_properties = self
+            .compute_properties
+            .0
+            .get_mut(package_id)
+            .expect("package should exist");
+        package_compute_properties.clear();
+
+        // Re-analyze the package.
+        let mut package_store_scaffolding = PackageStoreScaffolding::default();
+        let package = package_store.get(package_id).expect("package should exist");
```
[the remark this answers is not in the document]
cesarzc: PR #1183  incorporates this feedback.

compiler/qsc_rca/src/common.rs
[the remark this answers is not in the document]
cesarzc: PR #1182 moves `FunctorApp` to a shared location.

compiler/qsc_rca/src/analyzer.rs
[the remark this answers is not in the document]
cesarzc: PR merged and changes incorporated.

compiler/qsc_rca/src/common.rs
cesarzc: Updated to use `FunctorSetValue`.

compiler/qsc_rca/src/common.rs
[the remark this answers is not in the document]
cesarzc: PR merged and code updated to use `FunctorApp`.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,405 @@
...
+    },
+    scaffolding::PackageScaffolding,
+    ApplicationsTable, ComputeKind, QuantumProperties, RuntimeFeatureFlags, ValueKind,
+};
+use qsc_data_structures::index_map::IndexMap;
+use qsc_fir::fir::{BlockId, ExprId, NodeId, Pat, PatId, PatKind, SpecDecl, StmtId};
+use rustc_hash::FxHashMap;
+
+#[derive(Debug)]
+pub struct ApplicationInstancesTable {
```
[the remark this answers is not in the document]
cesarzc: Renamed to `ApplicationsGeneratorSetBuilder` so the name better matches the intent of the data structure.

compiler/qsc_rca/src/scaffolding.rs
cesarzc: Did you mean `unimplemented!()`?

compiler/qsc_rca/src/applications.rs
cesarzc: That's exactly what it meant. After refactoring, the flag is no longer needed.

compiler/qsc_rca/src/common.rs
cesarzc: Removed all "N.B." occurrences.

compiler/qsc_rca/src/common.rs
cesarzc: Fixed.

compiler/qsc_rca/src/scaffolding.rs
cesarzc: Updated the code to use `unimplemented!()`.

compiler/qsc_rca/src/applications.rs
cesarzc: Fixed.

compiler/qsc_rca/tests/common/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+) {
+    let callable_id = fir_package_store
+        .find_callable_id_by_name(callable_name)
+        .expect("callable should exist");
+
+    let callable_compute_properties = package_store_compute_properties.get_item(callable_id);
+    expect.assert_eq(&callable_compute_properties.to_string());
+}
+
+pub fn check_last_statement_compute_propeties(
```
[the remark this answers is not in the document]
cesarzc: Fixed.

compiler/qsc_rca/src/lib.rs
cesarzc: Fixed.

compiler/qsc_rca/src/rca.rs
cesarzc: Updated the code to use let-else.

compiler/qsc_rca/src/rca.rs
cesarzc: Fixed.

compiler/qsc_rca/src/scaffolding.rs
cesarzc: Renamed the data structures in this module that had the redundant suffix.

cesarzc:
Converting to draft while I work on the refactoring to use the visitor pattern.

compiler/qsc/benches/rca.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    fn analyze_open_package(&mut self) {
+        let Some(compute_properties) = &mut self.compute_properties else {
+            panic!("cannot analyze open package if the other packages have not been analyzed");
+        };
+
+        // Clear the compute properties of the open package.
+        let open_package_id = map_hir_package_to_fir(self.compiler.package_id());
+        let package_compute_properties = compute_properties.get_mut(open_package_id);
+        package_compute_properties.clear();
```
swernli, line 106: random question: if we haven't yet computed the properties for the open package, why do they need to be cleared?

swernli: Oh, never mind, it's because of the repeated benchmarking.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,666 @@
...
+            package_compute_properties
+                .exprs
+                .insert(expr_id, application_generator_set);
+        }
+    }
+}
+
+/// An instance of a callable application.
+#[derive(Debug, Default)]
+pub struct ApplicationInstance {
```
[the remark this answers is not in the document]
cesarzc: A term that I have seen more frequently is "call application", so what do you think about `CallApplication`?

compiler/qsc_rca/tests/intrinsics.rs
```diff
@@ -0,0 +1,1289 @@
...
+                    dynamic_param_applications:
+                        [0]: [Parameter Type Element] Quantum: QuantumProperties:
+                            runtime_features: RuntimeFeatureFlags(UseOfDynamicString)
+                            value_kind: Element(Static)
+                adj: <none>
+                ctl: <none>
+                ctl-adj: <none>"#
+        ],
+    );
+}
```
swernli, line 248: Future: Hmm, this brings to mind something that was a bit of a debate in the old stack: should dynamic strings be rejected or ignored? I tend to be of the opinion that asking a user to comment out their strings used for local printf-debugging is not a huge burden, but others have brought up the desire to let folks write code with debug `Message` calls and string interpolation of dynamic values and insisted that the compiler should know to ignore these during analysis for code generation. I don't think it should affect this PR, just wanted to note for the future that a similar complaint might come up if we treat anything that uses dynamic strings as disallowed...

compiler/qsc_rca/tests/literals.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
swernli, line 3: This file doesn't have any tests for string literals or string interpolation. Literals might fit here, but I wonder if there needs to be a string.rs file that handles both, since strings can start to get complex in their own right with nested interpolation, concatenation, equality comparisons, etc.

compiler/qsc_rca/tests/qubits.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use common::{check_last_statement_compute_properties, CompilationContext};
+use expect_test::expect;
+
+#[test]
+fn check_rca_for_static_single_qubit_allcation() {
+    let mut compilation_context = CompilationContext::new();
+    compilation_context.update(r#"use q = Qubit();"#);
+    let package_store_compute_properties = compilation_context.get_compute_properties();
+    check_last_statement_compute_properties(
```
swernli, line 14: I'm worried these tests aren't validating what you think they are validating... because the qubit statement gets transformed into an initial call to allocate and a final call to release, the last statemetn in the block will be just part of what was generated from the pass. For example, using [this program in the playground](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAEy2OwWrDMBBE7%2FqKAV%2FsS3VPKOTQ0lPAJukHbOwlWYhWirQihJB%2FL1YzxzfwZrz36HCgkK7svPf4YTPRM4pRNl5W1niHLy5zlmQStZHjRQqkgBBEJdAVU4eU4zlTgF3IMJPixKiFF1j8V%2BKepQ1MHea48IdTClwSzYz9Y6qkVsP4tjydA4Ddt1p%2BjFHU%2BqGRmDjT%2BgR7Eu0HbPCrYni2dk0tjBs%2BMdWTWD9sW%2FFyL%2FcHHVFUxvIAAAA%3D) and peeking at the HIR, the last statement of the block will be the call qubit release. So this compute properties is only for the call to `__quantum__rt__qubit_release`. If that's intentional, then it might be worth a comment here to clarify.

compiler/qsc_rca/tests/qubits.rs
```diff
@@ -0,0 +1,43 @@
...
+                inherent: Quantum: QuantumProperties:
+                    runtime_features: RuntimeFeatureFlags(0x0)
+                    value_kind: Element(Static)
+                dynamic_param_applications: <empty>"#
+        ],
+    );
+}
+
+#[test]
+fn check_rca_for_static_multi_qubit_allcation() {
```
swernli, line 28: Might be worth a test for dynamic qubit array, something like `use qs = Qubit[s];` where `s` is dynamic. Though given the transformations done by the passes, it might be hard to find the right statement to check the properties on, since it may not be the last one.

compiler/qsc_rca/tests/binops.rs
```diff
@@ -0,0 +1,133 @@
...
+    );
+}
+
+#[test]
+fn check_rca_for_nested_bin_ops_with_a_dynamic_operand() {
+    let mut compilation_context = CompilationContext::new();
+    compilation_context.update(
+        r#"
+        use q = Qubit();
+        let i = M(q) == Zero ? 0 | 1;
```
swernli, line 118: food for thought: if you want to reduce the complexity of some tests that require a dynamic integer you could use `DrawRandomInt` instead of relying on branching from qubit measurement. That said, the qubit measurement is more indicative of the kind of dynamism programs will use, so it's a trade off.

compiler/qsc_rca/tests/udts.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[test]
+fn check_rca_for_udt_constructor_with_a_dynamic_value() {
+    let mut compilation_context = CompilationContext::new();
+    compilation_context.update(
+        r#"
+        open Microsoft.Quantum.Math;
+        use q = Qubit();
+        let r = M(q) == Zero ? 0.0 | 1.0;
```
swernli, line 36: similar note here, you could use `DrawDynamicDouble` if you wanted to skip qubits.

compiler/qsc_rca/tests/vars.rs
```diff
@@ -0,0 +1,108 @@
...
+                inherent: Quantum: QuantumProperties:
+                    runtime_features: RuntimeFeatureFlags(0x0)
+                    value_kind: Element(Static)
+                dynamic_param_applications: <empty>"#
+        ],
+    );
+}
+
+#[test]
+fn check_rca_for_dynamic_result_var() {
```
swernli, line 88: this might be getting too complex, but I'm curious what the analysis would say for things like locally defined lambdas that capture a value. That opens the door to the lambda itself being dynamic, the lambda being static but the captured variables being dynamic, etc. Might be out of scope for now though.

compiler/qsc_rca/src/lib.rs
```diff
@@ -0,0 +1,841 @@
...
+            None => write!(indent, "\nctl-adj: <none>")?,
+        }
+        Ok(())
+    }
+}
+
+/// A set of compute properties associated to a callable or one of its elements, from which the properties of any
+/// particular call application can be derived.
+#[derive(Clone, Debug)]
+pub struct ApplicationGeneratorSet {
```
swernli, line 295: This is one of the names we wanted to explore alternatives for.

compiler/qsc_rca/src/lib.rs
```diff
@@ -0,0 +1,841 @@
...
+                write!(indent, "\n[{param_index}]: {param_application}")?;
+            }
+        }
+        Ok(())
+    }
+}
+
+impl ApplicationGeneratorSet {
+    #[must_use]
+    pub fn derive_application_compute_kind(&self, args_value_kinds: &[ValueKind]) -> ComputeKind {
```
swernli, line 327: on the topic of naming, I think it would be helpful to have this reflect whatever verb gets used in the struct name or go with a simple `get`. So right now, I would lean toward either `generate_application_compute_kind` or `get_application_compute_kind`.

compiler/qsc_rca/src/lib.rs
```diff
@@ -0,0 +1,841 @@
...
+        };
+
+        // Return the aggregated compute kind.
+        ComputeKind::Quantum(QuantumProperties {
+            runtime_features,
+            value_kind,
+        })
+    }
+
+    pub(crate) fn aggregate_runtime_features(
```
swernli, line 492: I would've expected this to take or produce `RuntimeFeatureFlags` since that's what other functions mean when they talk about "runtime_features" but these are instead calculated internally. Not sure how to convey this in the name, but really it calculates and then aggregates the runtime features.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ num-complex = "0.4"
 num-traits = "0.2"
 probability = "0.20"
 indenter = "0.3"
+itertools = "0.10.5"
```
swernli, line 55: Looking closer, I don't think this dependency is actually needed... I'll clarify in other comments.

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,666 @@
...
+                    .for_each(|value_kind| {
+                        element_param_application.aggregate_value_kind(*value_kind);
+                    });
+            }
+        };
+    }
+
+    fn close_inherent(&mut self) -> ApplicationInstanceComputeProperties {
+        // The inherent param application is always the first one.
+        let mut variants = self.application_instances[0].drain(..).collect_vec();
```
swernli, line 252: This can use standard library for this rather than `collect_vec` like this:

```suggestion
        let mut variants = self.application_instances[0].drain(..).collect::<Vec<_>>();
```

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,666 @@
...
+    fn close_param(&mut self, param_index: InputParamIndex) -> ParamApplicationComputeProperties {
+        const DYNAMIC_ELEMENTS_PARAM_VARIANTS: usize = 1;
+        const DYNAMIC_ARRAY_PARAM_VARIANTS: usize = 3;
+
+        // We need to offset the index since the first top-level vector in application instances is reserved for the
+        // inherent variant.
+        let variants_index = usize::from(param_index) + 1usize;
+        let variants = &mut self.application_instances[variants_index]
+            .drain(..)
+            .collect_vec();
```
swernli, line 268: Same thing here:

```suggestion
            .collect::<Vec<_>>();
```

compiler/qsc_rca/src/overrider.rs
```diff
@@ -0,0 +1,469 @@
...
+            let callables = namespace_items
+                .iter()
+                .filter_map(|item_id| {
+                    let item = self.get_item(*item_id);
+                    match &item.kind {
+                        ItemKind::Callable(decl) => Some((item.id, decl.name.name.to_string())),
+                        _ => None,
+                    }
+                })
+                .collect_vec();
```
swernli, line 373: Again here, this is not needed:

```suggestion
                .collect::<Vec<_>>();
```

compiler/qsc_rca/tests/common/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+pub fn check_last_statement_compute_properties(
+    package_store_compute_properties: &PackageStoreComputeProperties,
+    expect: &Expect,
+) {
+    let last_package_id = package_store_compute_properties
+        .iter()
+        .map(|(package_id, _)| package_id)
+        .sorted()
+        .last()
+        .expect("at least one package should exist");
```
swernli, line 132: This one is a tricky one... `sorted` seems like it it sorts on the fly, but internally it actually creates a copy of the vector, sorts it, and then returns a new owning iterator for the sorted items. In general, I would rather have the code do explicit copies and sort the copy so it's clear that copy is happening. That said, since all this code needs is the highest package id, which we can get from an stdlib function `max`:

```suggestion
    let last_package_id = package_store_compute_properties
        .iter()
        .map(|(package_id, _)| package_id)
        .max()
        .expect("at least one package should exist");
```

compiler/qsc_rca/tests/common/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    let package_compute_properties = package_store_compute_properties.get(last_package_id);
+    let last_statement_id = package_compute_properties
+        .stmts
+        .iter()
+        .map(|(stmt_id, _)| stmt_id)
+        .sorted()
+        .last()
+        .expect("at least one statement should exist");
```
swernli, line 140: The same thing can be done here:

```suggestion
    let last_statement_id = package_compute_properties
        .stmts
        .iter()
        .map(|(stmt_id, _)| stmt_id)
        .max()
        .expect("at least one statement should exist");
```

Cargo.toml
swernli: The other comments explain stdlib alternatives to `collect_vec` and `sorted` so that this dependency can be dropped.

compiler/qsc_rca/tests/intrinsics.rs
cesarzc: Even if strings are mostly used with `Message`, I think the safest thing to do is to reject dynamic strings because otherwise we could have something like this:

```qsharp
use q = Qubit();
let s = if M(q) == Zero { "Foo" } else { "Bar" };
if s == "Foo" {
    X(q);
}
```

I don't think people are doing this, but that doesn't mean they can't.

compiler/qsc_rca/tests/literals.rs
cesarzc: I'll create a string.rs test file and add tests with interpolation, concatenation and equality comparisons.

Cargo.toml
cesarzc: After merging your suggestions that removed uses of itertools, I removed this dependency, thanks Stefan!

swernli: Whoops, looks like the usage in the module was removed but this line is still pulling it into the dependencies even though nothing uses it. This can be removed now too:

```suggestion
```

compiler/qsc_rca/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+            runtume_capabilities |= RuntimeCapabilityFlags::HigherLevelConstructs;
+        }
+        if self.contains(RuntimeFeatureFlags::UseOfDynamicString) {
+            runtume_capabilities |= RuntimeCapabilityFlags::HigherLevelConstructs;
+        }
+        if self.contains(RuntimeFeatureFlags::UseOfDynamicallySizedArray) {
+            runtume_capabilities |= RuntimeCapabilityFlags::HigherLevelConstructs;
+        }
+        if self.contains(RuntimeFeatureFlags::UseOfDynamicUdt) {
+            runtume_capabilities |= RuntimeCapabilityFlags::all();
```
swernli, line 806: I wonder if using `all()` here might be confusing... it's not that we know it's using every capability but rather we can no longer verify what capabilities it uses because it's too complex. My reading of "Higher level constructs" was that this was meant to be the catch all for "things we can't handle yet." I think it will work either way, just wanted to confirm the intention.

compiler/qsc_rca/tests/literals.rs
cesarzc: Created strings.rs specifically to tests RCA on strings and consolidated the tests for literals in other data types into the types.rs file.

compiler/qsc_rca/src/cyclic_callables.rs
```diff
@@ -0,0 +1,300 @@
...
+            // Functions are inherently classically pure.
+            inherent: ComputeKind::Classical,
+            dynamic_param_applications,
+        }
+    }
+
+    fn create_operation_specialization_application_generator_set(
+        input_params: &Vec<InputParam>,
+        output_type: &Ty,
+    ) -> ApplicationGeneratorSet {
```
swernli, line 170: nitpick: since this function and `create_operation_specialization_application_generator_set` don't need `self` or othewise interact with the containing type, then can be moved outside of the `impl` block and just live in this file as utility functions. Then the call sites won't need `Self::` to scope the call.

compiler/qsc_rca/tests/qubits.rs
cesarzc: This is a great catch! I updated the tests to verify the behavior using the variable binding to look at the compute properties of the expression binded to the variable.

compiler/qsc_rca/tests/qubits.rs
cesarzc: Updated the tests to include a dynamic qubit array, as well as a dynamic single qubit.

Using the trick of checking an expression's properties through the variable it is binded too worked here as well.

For the dynamic single qubit allocation, I verified the behavior by encapsulating the dynamic qubit allocation in a callable.

compiler/qsc_rca/tests/binops.rs
cesarzc: For tests I would prefer to use explicit qubit measurements as sources of dynamism to make it clear for people that might not be as familiar with the implementation details of RCA.

compiler/qsc_rca/tests/udts.rs
cesarzc: For tests, I would prefer to use explicit qubit measurements as sources of dynamism to make it clear for people that might not be as familiar with the implementation details of RCA.

compiler/qsc_rca/src/scaffolding.rs
```diff
@@ -0,0 +1,308 @@
...
+    fir::{
+        self, BlockId, ExprId, LocalItemId, PackageId, StmtId, StoreBlockId, StoreExprId,
+        StoreItemId, StoreStmtId,
+    },
+    ty::FunctorSetValue,
+};
+
+/// Scaffolding used to build the package store compute properties.
+#[derive(Debug)]
+pub struct PackageStoreComputeProperties(IndexMap<PackageId, PackageComputeProperties>);
```
swernli, line 19: On the topic of naming, I think using the same name for the scaffolding class that is internal and the public facing class can be a bit confusing, even though they are qualified by the module within the crate. Maybe something like `InternalStoreComputeProperies`?

compiler/qsc_rca/src/applications.rs
```diff
@@ -0,0 +1,665 @@
...
+            stmts: self.stmts,
+            exprs: self.exprs,
+            value_kind,
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct LocalsComputeKindMap(FxHashMap<LocalVarId, LocalComputeKind>);
+
```
swernli, line 550: Since this uses `LocalVarId` and those are low numbered and (mostly) continguous, this could potentially be an `IndexMap` instead of `FxHashMap` for some perf gains.

compiler/qsc_rca/tests/vars.rs
cesarzc: The analysis for lambdas is currently not thorough at all. We basically just mark it with the `UseOfClosure` runtime feature and consider any lambda as dynamic. This has the consequence that the output of calling a lambda is also dynamic.

I added some tests to document the current expectations from lambdas both in (new) closures.rs and calls.rs test files.

compiler/qsc_rca/src/core.rs
```diff
@@ -0,0 +1,2121 @@
...
+
+        // Add the value expression ID to the return expressions tracked by the application instance.
+        let application_instance = self.get_current_application_instance_mut();
+        let value_expression_compute_kind =
+            *application_instance.get_expr_compute_kind(value_expr_id);
+        application_instance.return_expressions.push(value_expr_id);
+
+        // The compute kind of the return expression itself consists of only the runtime features of the value
+        // expression.
+        let default_value_kind = ValueKind::Element(RuntimeKind::Static);
```
swernli, line 710: For the default kind here, we might need to inherit something from the context, since the expression inside the return may be static, the actual condition of the return being evaluated may be dynamic. For example:
```qsharp
operation f() : Int {
    use q = Qubit();
    if M(q) == Zero {
        return 0;
    }
    return 1;
}
let x = f();
x
```
The value of `x` I would expect to be dynamic, but today it looks like it comes out as static.

compiler/qsc_rca/tests/arrays.rs
```diff
@@ -0,0 +1,179 @@
...
+        &expect![
+            r#"
+            ApplicationsGeneratorSet:
+                inherent: Quantum: QuantumProperties:
+                    runtime_features: RuntimeFeatureFlags(UseOfDynamicBool | UseOfDynamicInt | UseOfDynamicDouble | UseOfDynamicallySizedArray)
+                    value_kind: Array(Content: Dynamic, Size: Dynamic)
+                dynamic_param_applications: <empty>"#
+        ],
+    );
+}
```
swernli, line 179: I think a couple of test cases here that show mutable arrays being appended into inherit the right properties would be useful. Here's a couple I tried that show great results:

```suggestion
}

#[test]
fn check_rca_for_mutable_array_statically_appended() {
    let mut compilation_context = CompilationContext::new();
    compilation_context.update(
        r#"
        mutable arr = [];
        use q = Qubit();
        for i in 0..10 {
            set arr += [M(q)];
        }
        arr"#,
    );
    let package_store_compute_properties = compilation_context.get_compute_properties();
    check_last_statement_compute_properties(
        package_store_compute_properties,
        &expect![[r#"
            ApplicationsGeneratorSet:
                inherent: Quantum: QuantumProperties:
                    runtime_features: RuntimeFeatureFlags(0x0)
                    value_kind: Array(Content: Dynamic, Size: Static)
                dynamic_param_applications: <empty>"#]],
    );
}

#[test]
fn check_rca_for_mutable_array_dynamically_appended() {
    let mut compilation_context = CompilationContext::new();
    compilation_context.update(
        r#"
        mutable arr = [0, 1];
        use q = Qubit();
        if M(q) == Zero {
            set arr += [2];
        }
        arr"#,
    );
    let package_store_compute_properties = compilation_context.get_compute_properties();
    check_last_statement_compute_properties(
        package_store_compute_properties,
        &expect![[r#"
            ApplicationsGeneratorSet:
                inherent: Quantum: QuantumProperties:
                    runtime_features: RuntimeFeatureFlags(UseOfDynamicInt | UseOfDynamicallySizedArray)
                    value_kind: Array(Content: Dynamic, Size: Dynamic)
                dynamic_param_applications: <empty>"#]],
    );
}

```

swernli (commented):
Thanks for already addressing a bunch of my ongoing feedback! There are a few more things to circle back to, most notably the issue we discussed with ealry returns in dynamic contexts. Otherwise, looking good and almost ready to go in!

compiler/qsc_rca/src/applications.rs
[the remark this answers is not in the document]
cesarzc: I created #1285 to track some of the names we want to reconsider and change them in a follow up PR.

compiler/qsc_rca/src/lib.rs
cesarzc: Added it to #1285 to follow up on name alternatives.

compiler/qsc_rca/src/lib.rs
cesarzc: Yes, this one is tricky. I also added it to #1285 to reconsider what name should we give to this function.

compiler/qsc_rca/src/lib.rs
cesarzc: I agree, using `all()` here might be confusing and I think it was me being overly cautious before more pieces were in place. I'll update this function to replace `all()` with `HigherLevelConstructs`.

compiler/qsc_rca/src/cyclic_callables.rs
cesarzc: Moved outside the `impl` block.

compiler/qsc_rca/src/lib.rs
cesarzc: Updated.

compiler/qsc_rca/src/lib.rs
cesarzc: Renamed.

compiler/qsc_rca/src/scaffolding.rs
cesarzc: Renamed!

compiler/qsc_rca/src/applications.rs
cesarzc: Good catch! Using `IndexMap` instead of `FxHashMap` now.

compiler/qsc_rca/tests/arrays.rs
cesarzc: Thanks for catching this test gap! These test cases have been added now.

PR #1166, opened 2024-02-15
Add new nested iteration benchmark

swernli:
This adds a new benchmark to evaluate performance of deeply nested iteration and mutation of arrays.


swernli (commented):
I'm posting this as a separate PR so that later perf improvements can be evaluated compared to main.

idavis (approved):

PR #1172, opened 2024-02-16
Basic Q# Formatter

ScottCarda-MS:
Adds a Q# formatter to the compiler with some basic abilities.

The formatter is added to the VScode extension and is triggered manually with the VScode format command, which may be set to run on save, depending on user settings.


compiler/qsc_parse/src/lex/formatter.rs
minestarks, line 1: Someone else might have a better suggestion, but to me this is calling for a new crate again :) qsc_format?

compiler/qsc_parse/src/lex/formatter.rs
```diff
@@ -0,0 +1,270 @@
...
+    match (&left.kind, &right.kind) {
+        (
+            ConcreteTokenKind::Cooked(TokenKind::Open(l)),
+            ConcreteTokenKind::Cooked(TokenKind::Close(r)),
+        ) if l == r => {
+            rule_no_space(left, whitespace, right, &mut edits);
+        }
+        (ConcreteTokenKind::Comment | ConcreteTokenKind::Cooked(TokenKind::DocComment), _) => {
+            rule_trim_comments(left, &mut edits, code);
+            rule_indentation(left, whitespace, right, &mut edits, indent_level);
```
minestarks, line 150: I think, in theory, you don't need to have the `rule_indentation` calls sprinkled throughout the `match` expression like this. If the `rule_` functions returned the _new whitespace_ instead of the edits, you could rewrite this to look like:

1. Big match statement - apply `rule_xyz` functions to get `new_whitespace`
2. _After_ the match statement, call `fix_whitespace` to update `new_whitespace`
3. If new whitespace is not equal 
4. If `new_whitespace != old_whitespace` , generate the edits 

This would be more readable (and easier to follow logically) I think.

compiler/qsc_parse/src/lex/concrete.rs
```diff
@@ -0,0 +1,111 @@
...
+use super::{cooked, raw};
+
+pub(super) struct ConcreteToken {
+    pub kind: ConcreteTokenKind,
+    pub span: Span,
+}
+
+#[derive(Debug, PartialEq)]
+pub(super) enum ConcreteTokenKind {
+    Cooked(cooked::TokenKind),
```
minestarks, line 17: The way this enum is organized is throwing me off a bit at the usage site (in the bit match statement in formatter.rs). 

Instead of "leaking" the low-level concept of cooked tokens, I'd hope to see this enum organized a bit more abstractly, like:

```rs
pub(super) enum ConcreteTokenKind {
    Syntax(cooked::TokenKind),  // nontrivial tokens. Not sure if this is the best name to use. Maybe `Compilation`?
    Error(cooked::Error),
    Whitespace,
    Comment // this would INCLUDE doc comments, since you don't actually need the differentiate them from regular comments for formatting
}

File: compiler/qsc_parse/src/lex/formatter.rs [hunk 3]
```diff
@@ -0,0 +1,270 @@
...
+        if indent_level > 0 {
+            indent_level - 1
+        } else {
+            indent_level
+        }
+    } else {
+        indent_level
+    };
+
+    match (&left.kind, &right.kind) {
```
minestarks, line 144: As discussed, if you alias names to be as short as possible I'm hoping this match statement will line up a little better without the line breaks.

compiler/qsc_parse/src/lex/formatter.rs
```diff
@@ -0,0 +1,270 @@
...
+                rule_indentation(left, whitespace, right, &mut edits, indent_level);
+            }
+        },
+        (_, ConcreteTokenKind::Cooked(TokenKind::Close(Delim::Brace))) => {
+            rule_indentation(left, whitespace, right, &mut edits, indent_level);
+        }
+        (
+            _,
+            ConcreteTokenKind::Cooked(TokenKind::Keyword(
+                Keyword::Operation | Keyword::Function | Keyword::Newtype | Keyword::Namespace,
```
minestarks, line 168: I know your version is technically shorter, but I wonder if this would look better if you expand this out to be 4 separate lines like 

```rs
 ConcreteTokenKind::Cooked(TokenKind::Keyword(Keyword::Operation)) |
 ConcreteTokenKind::Cooked(TokenKind::Keyword(Keyword::Function)) |
 ConcreteTokenKind::Cooked(TokenKind::Keyword(Keyword::Newtype)) |
 ConcreteTokenKind::Cooked(TokenKind::Keyword(Keyword::Namespace)) |
```

I'm hoping that if we can have everything lined up, looking like a table, things will be much easier to follow. Sometimes repetition is actually better IMO!

compiler/qsc_parse/src/lex/formatter.rs
```diff
@@ -0,0 +1,270 @@
...
+            ConcreteTokenKind::Cooked(TokenKind::Close(Delim::Brace)),
+            ConcreteTokenKind::Cooked(TokenKind::Semi),
+        ) => {
+            rule_no_space(left, whitespace, right, &mut edits);
+        }
+        (ConcreteTokenKind::Cooked(TokenKind::Close(Delim::Brace)), _) => {
+            rule_indentation(left, whitespace, right, &mut edits, indent_level);
+        }
+        (ConcreteTokenKind::Cooked(cooked_left), ConcreteTokenKind::Cooked(cooked_right)) => {
+            match (cooked_left, cooked_right) {
```
minestarks, line 183: Similar to my last comment. Instead of having nested `match`es I think it would be better to write this out line by line, like:

```rs
// abc abc
(CTK::Cooked(TokenKind::Ident), CTK::Cooked(TokenKind::Ident)) => 
(CTK::Cooked(TokenKind::Keyword(_), CTK::Cooked(TokenKind::Ident)) => 
// abc :
(CTK::Cooked(_), CTK::Cooked(TokenKind::Colon)) => 
// : abc
(CTK::Cooked(TokenKind::Colon), CTK::Cooked(_)) => 
// , abc
(CTK::Cooked(TokenKind::Comma), CTK::Cooked(_)) => 
``
 etc.

(You can take it or leave it on the comment suggestion, but I thought it would be helpful to have a visual example of what the rule is)

File: compiler/qsc_parse/src/lex/formatter.rs [hunk 6]
```diff
@@ -0,0 +1,270 @@
...
+                | (TokenKind::Ident, TokenKind::Open(Delim::Bracket))
+                | (TokenKind::Ident, TokenKind::Comma)
+                | (TokenKind::Open(_), _)
+                | (_, TokenKind::Close(_)) => {
+                    rule_no_space(left, whitespace, right, &mut edits);
+                }
+                _ => {}
+            }
+        }
+        _ => {}
```
minestarks, line 201: You'd still want to fix the indentation for the catch-all cases right?

ScottCarda-MS: In all other cases, we don't want to be authoritative about whitespace. The `rule_indentation` removes any whitespace other than new-lines and indentation, so it wouldn't work well to have that be called in situations where we don't want that behavior, like in `let x = 3;`. Those spaces are desired in that example.

compiler/qsc_parse/src/lex/formatter.rs
ScottCarda-MS: Not all of the rules rewrite the whitespace. Look at the `rule_trim_comments` rule, which edits the contents of the comment.

compiler/qsc_parse/src/lex/formatter.rs
ScottCarda-MS: I tried it out, and I think the verbosity of having `CTK::Cooked(...)` everywhere really hurts my ability to quickly get at the relevant information of what two token kinds are being compared.
image
versus
image

compiler/qsc_parse/src/lex/formatter.rs
ScottCarda-MS: done

compiler/qsc_parse/src/lex/concrete.rs
ScottCarda-MS: Discussed offline: The doc comments are logically different from regular comments and are compiled tokens, so it makes sense to keep them separate. Renaming the "Cooked" case makes sense and will be done.

compiler/qsc_parse/src/lex/formatter.rs
```diff
@@ -0,0 +1,270 @@
...
+    let mut edits = vec![];
+
+    let mut indent_level = 0;
+
+    #[allow(unused_assignments)] // there's probably a better way of doing this, but this works
+    let mut one = None;
+    let mut two = None;
+    let mut three = None;
+
+    for token in tokens {
```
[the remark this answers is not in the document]
ScottCarda-MS: It might be possible to do a larger refactor of the code in the future, including removing whitespace tokens in favor of a kind of whitespace trivia approach. Since this is under time pressure, let's table that possible refactoring for a future PR.

compiler/qsc_parse/src/lex/concrete.rs
ScottCarda-MS: Renamed to `Syntax`

compiler/qsc_parse/src/lex/formatter.rs
ScottCarda-MS: I like that suggestion a lot, and didn't know Rust supported that. I _did_ have to rename the `TokenKind::String` case though to avoid it conflicting with the built-in `String` type.

compiler/qsc_parse/src/lex/formatter.rs
```diff
@@ -0,0 +1,258 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::keyword::Keyword;
```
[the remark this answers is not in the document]
ScottCarda-MS: added

compiler/qsc_parse/src/lex/formatter.rs
ScottCarda-MS: I've made the qsc_formatter crate, which required exposing most of the date structures in the lexer.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,263 @@
...
+/// token stream of the given source code string. It then uses a sliding window
+/// over this token stream to apply formatting rules when the selected tokens
+/// match certain patterns. Formatting rules will generate text edit objects
+/// when the format of the input string does not match the expected format, and
+/// these edits are returned on using the formatter.
+use qsc_data_structures::span::Span;
+use qsc_frontend::keyword::Keyword;
+use qsc_frontend::lex::{
+    concrete::{self, ConcreteToken, ConcreteTokenKind::*},
+    cooked::TokenKind::*,
```
orpuente-MS, line 8: These `use::MyEnum::*` should be as local as possible (even if you use them in multiple functions), so that the individual variants are not mistaken with individual types or between enums. For example, from reading the code I don't know if Whitespace belongs to `TokenKind` or to `ConcreteTokenKind`. So, instead of importing them globally it is better to import them right before the match statement where they are used, like this:
```rust
use qsc_frontend::lex{concrete::ConcreteTokenKind, cooked::TokenKind};

pub fn format(code: &str) -> Vec<Edit> {
  ...
  use TokenKind::*;
  let mut edits_for_triple = match (&one, &two, &three) { ... }
}

fn apply_rules(...) -> Vec<Edit> {
  ...
  use ConcreteTokenKind::*;
  match (&left.kind, &right.kind) { ... }
}
```

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,263 @@
...
+                if let Syntax(Open(Delim::Brace)) = one.kind {
+                    indent_level += 1;
+                }
+
+                // if the token is a }, decrease the indent level
+                if let Syntax(Close(Delim::Brace)) = one.kind {
+                    #[allow(clippy::implicit_saturating_sub)]
+                    if indent_level > 0 {
+                        indent_level -= 1;
+                    }
```
orpuente-MS, line 65: This can be written as follows so you don't need the if-branch or the `#[allow(clippy::implicit_saturation_sub)`:
```rust
ident_level = ident_level.saturating_sub(0);
```

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I've moved the Enum::* uses to the match statement, though I'm not sure that will help with the ambiguity you pointed out as both `ConcreteTokenKind` and `TokenKind` are used in the same match statement.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I had tried to use the `saturation_sub(0)` before, but it didn't seem to produce the right behavior. Maybe I was just not using it correctly. I'll try again...

ScottCarda-MS: Ah it is because the `indent_level` is a usize, which is ambiguous and can't use the saturating_sub.

orpuente-MS: That is weird, `usize` is actually well defined and has `usize::saturating_sub(rhs)` as a method. It is true that its size depends on the target architecture, but it is still well defined.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,257 @@
...
+                        one,
+                        get_token_contents(code, two),
+                        three,
+                        code,
+                        indent_level,
+                    )
+                } else {
+                    // one, two are adjacent tokens with no whitespace in the middle
+                    apply_rules(one, "", two, code, indent_level)
+                }
```
orpuente-MS, line 83: I think this whole logic can be rewritten as:
```rust
match (one.kind, two.kind) {
  (Syntax(Open(Delim::Brace)), _) => ident_level += 1,
  (Syntax(Close(Delim::Brace)), _) => ident_level = ident_level.saturating_sub(1),
  (WhiteSpace, _) => (),
  (_, Whitespace) => apply_rules(one, get_token_contents(code, two), three, code, indent_level),
  _ => apply_rules(one, "", two, code, indent_level),
}
```
and in that way you can better express the intent that it depends on `one.kind` and `two.kind`.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: fixed

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: This doesn't allow the indentation to be adjusted in the same iteration as the rules applied, so it is a bit different.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,254 @@
...
+impl Edit {
+    fn new(lo: u32, hi: u32, new_text: &str) -> Self {
+        Self {
+            span: Span { lo, hi },
+            new_text: new_text.to_string(),
+        }
+    }
+}
+
+fn make_indent_string(level: usize) -> String {
```
minestarks, line 29: If the original document uses tabs or a different indent width should we respect that? I think VS Code provides these options via `FormattingOptions` https://code.visualstudio.com/api/references/vscode-api#FormattingOptions

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,254 @@
...
+        } else {
+            indent_level
+        }
+    } else {
+        indent_level
+    };
+
+    use qsc_frontend::keyword::Keyword;
+    use ConcreteTokenKind::*;
+    use TokenKind::*;
```
minestarks, line 142: Oh I didn't know you could do this! Looks much better than using an alias.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,254 @@
...
+    let mut edits = vec![];
+    // when we get here, neither left nor right should be whitespace
+
+    // if the right is a close brace, the indent level should be one less
+    let indent_level = if let Syntax(Close(Delim::Brace)) = right.kind {
+        if indent_level > 0 {
+            indent_level - 1
+        } else {
+            indent_level
+        }
```
orpuente-MS, line 135: This can also be just a call to `ident_level.saturating_sub(1)`

ScottCarda-MS: good catch, thanks!

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    edits
+}
+
+fn fix_whitespace(whitespace: &str, indent_level: usize) -> String {
+    //
+    // when you see newline, insert the indent string
+    // and trim until the next newline or the end of the string
+    //
```
orpuente-MS, line 106: This docstring should be before the function

```rust
/// When seeing a new line, insert the indent string and
/// trim until the next newline or the end of the string.
fn fix_whitespace(whitespace: &str, indent_level: usize) -> String { ... }
```

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+    right: &ConcreteToken,
+    edits: &mut Vec<Edit>,
+) {
+    if whitespace != " " {
+        edits.push(Edit::new(left.span.hi, right.span.lo, " "));
+    }
+}
+
+fn rule_trim_comments(left: &ConcreteToken, edits: &mut Vec<Edit>, code: &str) {
+    // fix trailing spaces on the comment
```
orpuente-MS, line 220: Same here, you can make it a docstring. Also, if the rule processes single comment, you can change name of the function to `rule_trim_comment`
```rust
/// Fixes trailing spaces on the comment
fn rule_trim_comment(left: &ConcreteToken, edits: &mut Vec<Edit>, code: &str) { ... }
```

compiler/qsc_parse/src/lex/cooked.rs
```diff
@@ -139,7 +148,7 @@ pub(crate) enum TokenKind {
     /// `;`
     Semi,
     /// A string literal.
-    String(StringToken),
+    StringToken(StringToken),
```
idavis, line 151: Why is this renamed?

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -0,0 +1,307 @@
...
+use expect_test::{expect, Expect};
+use indoc::indoc;
+
+fn check(input: &str, expect: &Expect) {
+    let actual = super::format(input);
+    expect.assert_debug_eq(&actual);
+}
+
+#[test]
+fn test_formatting() {
```
idavis, line 13: What is this testing?

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -0,0 +1,307 @@
...
+            },
+            new_text: "",
+        },
+    ]
+"#]],
+    );
+}
+
+#[test]
+fn remove_trailing_spaces() {
```
idavis, line 38: When are trailing spaces removed? Do we need more source states to show that it is getting removed in all cases? Should there be a test per kind of trailing space that gets cleaned up?

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -0,0 +1,307 @@
...
+                    },
+                    new_text: "\n",
+                },
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn correct_indentation() {
```
idavis, line 88: What kind of indentation is being corrected? What kinds of indents can be corrected? What test cases are we missing?

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -0,0 +1,307 @@
...
+                    },
+                    new_text: "",
+                },
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn test_sample() {
```
idavis, line 258: What is being asserted? What is being tested? What is the expected output?

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -0,0 +1,307 @@
...
+                    },
+                    new_text: "\n",
+                },
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn correct_empty_delimiters() {
```
idavis, line 199: What does "correct" mean in this context?

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+}
+
+fn make_indent_string(level: usize) -> String {
+    "    ".repeat(level)
+}
+
+/// Applies formatting rules to the given code str, generating edits where
+/// the source code needs to be changed to comply with the format rules.
+pub fn format(code: &str) -> Vec<Edit> {
```
idavis, line 35: This API seems very limited to supporting the language service. For testing, and other uses, it would seem handy to have `pub fn format(code: &str) -> String` with the formatted output and have this function be `pub fn calculate_edits(code: &str) -> Vec<Edit>`.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+                    ConcreteTokenKind::WhiteSpace => continue,
+                    _ => {}
+                }
+
+                if let ConcreteTokenKind::WhiteSpace = one.kind {
+                    // first token is whitespace, continue scanning
+                    continue;
+                } else if matches!(two.kind, ConcreteTokenKind::WhiteSpace) {
+                    // whitespace in the middle
+                    apply_rules(
```
idavis, line 69: What rules do we have? Do we have tests showing each rule being applied?

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+            }
+        };
+
+        edits.append(&mut edits_for_triple);
+    }
+
+    edits
+}
+
+fn fix_whitespace(whitespace: &str, indent_level: usize) -> String {
```
idavis, line 101: What does `fix` mean here?

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    println!(
+        "edits for `{}` : {edits:?}",
+        &code[left.span.lo as usize..right.span.hi as usize]
+    );
+    edits
+}
+
+fn rule_no_space(
```
idavis, line 197: What does this mean?

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,249 @@
...
+    whitespace: &str,
+    right: &ConcreteToken,
+    edits: &mut Vec<Edit>,
+) {
+    if !whitespace.is_empty() {
+        edits.push(Edit::new(left.span.hi, right.span.lo, ""));
+    }
+}
+
+fn rule_single_space(
```
idavis, line 208: What does `rule_single_space` mean?

language_service/src/lib.rs
```diff
@@ -221,6 +222,17 @@ impl LanguageService {
         )
     }
 
+    /// LSP: textDocument/format
+    #[must_use]
+    pub fn get_format_changes(&self, uri: &str) -> Vec<TextEdit> {
+        self.document_op(
```
idavis, line 228: Does this have the correct encoding offsets applied?

compiler/qsc_parse/src/lex/formatter.rs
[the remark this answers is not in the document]
ScottCarda-MS: I looked into doing this refactor and the logic to embed the whitespace spans into the tokens was much more complicated than I anticipated. The current structure is preferred for simplicity.

compiler/qsc_parse/src/lex/cooked.rs
idavis: Agreed. I'd like to keep any refactorings like this to an isolated PR.

idavis: Though I don't like * imports unless there is a real prelude being imported.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I had thought of implementing that as future work for the formatter, along with several other features related to further integration with the VS Code API.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: applied your suggestions to use the `match!()` syntax

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: function was removed

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: removed comment

compiler/qsc_parse/src/lex/cooked.rs
ScottCarda-MS: I've reversed the rename.

language_service/src/protocol.rs
```diff
@@ -92,6 +92,12 @@ pub struct Hover {
     pub span: Range,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct TextEdit {
```
billti, line 96: If VS Code is the main consumer, could we just match the struct there and call these 'newText' and 'range' rather than 'contents' and 'span'?

See https://code.visualstudio.com/api/references/vscode-api#TextEdit

compiler/qsc_formatter/src/formatter.rs
billti: I'm OK if we start with a highly 'opinionated' formatter (a-la Prettier) and only allow configuration away from the "one truth path" if users really scream about it. You keep the code (and the tests) way simpler the less config permutations you have.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -0,0 +1,239 @@
...
+    concrete::{self, ConcreteToken, ConcreteTokenKind},
+    cooked::TokenKind,
+    Delim,
+};
+
+#[cfg(test)]
+mod tests;
+
+#[derive(Debug)]
+pub struct Edit {
```
billti, line 15: See comment in protocol.rs where TextEdit was added with the fields 'span' and 'contents'. Should we have one struct here, consistent with the VS Code API struct if possible?

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: removed

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: That is a nice suggestion and made the tests much more readable. Thanks!

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: removed

language_service/src/lib.rs
ScottCarda-MS: The offsets encoding is fine as this function is of the same form as the other LS feature functions in this file. I believe @minestarks recently (within the last 2 months) made changes to this infrastructure regarding encoding, so she might know more.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: We can have a struct here that looks just like the one in the protocol.rs, but the language service package is further down the dependency stream, so the formatter mod can't take a dependency on it to use that TextEdit struct.

language_service/src/protocol.rs
ScottCarda-MS: done, although "newText" had to be "new_text" because of Rust naming conventions

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: All trailing spaces are removed. I'm not really sure what you are getting at here. The only exception I can think of is with a multi-line string, as the contents of strings will not be edited. I could write a test to check that that is the case.

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: I've renamed this to `remove_spaces_between_empty_delimiters` for better clarity.

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: I've renamed this to `sample_has_no_formatting_changes` for clarity.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: There a many rules. We have tests for each in the test.rs file.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I've renamed the function, but it is a function that is meant to ensure that the given whitespace is empty, and if it isn't it will generate an edit to make it so.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: Similar to the above comment, I've renamed the function. It is a function that is meant to ensure that the given whitespace has only one space in it, and if that isn't the case, it will generate an edit to make it so.

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: This function is meant to ensure the given whitespace has the correct indentation and at least one newline. If the whitespace doesn't, it will generate an edit with the correct indentation and at least one newline.

wasm/src/language_service.rs
```diff
@@ -196,6 +196,30 @@ impl LanguageService {
             .collect()
     }
 
+    pub fn get_format_changes(&self, uri: &str) -> Vec<ITextEdit> {
+        let edits = self.0.get_format_changes(uri);
+        edits
+            .into_iter()
+            .map(|edit| {
+                // ToDo: is there an easier way to convert data types?
```
minestarks, line 204: Yes, you should be able to say `range: edit.range.into()` , see the other usage of `TextEdit` in this file

language_service/src/protocol.rs
```diff
@@ -89,6 +89,12 @@ pub struct Hover {
...
```
minestarks, line 93: Hmm, ideall this would be the same data type as the one used by `additional_text_edits` in `CompletionItem`, because they're defined the same type in the vscode/lsp api. Could you refactor `CompletionItem` to match?

vscode/src/format.ts
```diff
@@ -0,0 +1,46 @@
...
+  return new QSharpFormatProvider(languageService);
+}
+
+class QSharpFormatProvider implements vscode.DocumentFormattingEditProvider {
+  constructor(public languageService: ILanguageService) {}
+
+  async provideDocumentFormattingEdits(document: vscode.TextDocument) {
+    // telemetry start format
+    const associationId = getRandomGuid();
+    sendTelemetryEvent(EventType.FormatStart, { associationId }, {});
```
minestarks, line 20: I love that you thought of telemetry, but what exactly is the goal of formatting telemetry? I can't imagine formatting possibly being a performance burden worthy of tracking --like, I almost think sending the telemetry event itself might take longer than traversing a document and collecting format edits.

vscode/src/format.ts
```diff
@@ -0,0 +1,46 @@
...
+    const lsEdits = await this.languageService.getFormatChanges(
+      document.uri.toString(),
+    );
+
+    if (!lsEdits) return [];
+    const edits = [];
+    for (const edit of lsEdits) {
+      const referenceRange = toVscodeRange(edit.range);
+      edits.push(new vscode.TextEdit(referenceRange, edit.newText));
+    }
```
minestarks, line 32: More concise (and a little more idiomatic) to write this using `.map`

```typescript
const edits = lsEdits.map(edit => new vscode.TextEdit(toVsCodeRange(edit.range), edit.newText))
```

language_service/src/lib.rs
minestarks: Yeah the offset/position mappings should all fall into place.

language_service/src/lib.rs
```diff
@@ -218,6 +219,17 @@ impl LanguageService {
...
+            format::get_format_changes,
+            "get_format_changes",
+            uri,
+            Position { line: 0, column: 0 },
```
minestarks, line 229: Since position will never be a relevant parameter, I suggest taking the parameter out of this `get_format_changes` altogether. Here you can just pass in a lambda instead of the name of the function. Then to reduce confusion further you can pass `()` instead of `Position { line: 0, column: 0 }`. 

An example to copy is `get_code_lenses` in this file.

language_service/src/format.rs
```diff
@@ -0,0 +1,30 @@
...
+    _position: Position,
+    encoding: Encoding,
+) -> Vec<TextEdit> {
+    let contents = compilation
+        .user_unit()
+        .sources
+        .find_by_name(source_name)
+        .expect("can't find source by name")
+        .contents
+        .clone();
```
minestarks, line 21: Is the `.clone()` necessary?

language_service/src/format.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    calculate_format_edits(&contents)
+        .iter()
+        .map(|edit| TextEdit {
+            new_text: edit.new_text.clone(),
```
minestarks, line 26: I think you can avoid the `.clone()` here if you use `into_iter()` instead of `iter()` to move the values out of the source vector.

(Though maybe then the reference to `&edit.span` may be a problem... I wonder if Rust would allow that if you flipped the order of `new_text` and `range` so that you use `&edit.span` before you consume `edit.new_text`.)

compiler/qsc_parse/src/lex/raw.rs
```diff
@@ -162,22 +162,34 @@ enum StringKind {
...
+    pub fn new(input: &'a str) -> Self {
         Self {
             chars: input.char_indices().peekable(),
             interpolation: 0,
+            starting_offset: 0,
+        }
+    }
+
+    #[must_use]
+    pub fn new_with_starting_offset(input: &'a str, starting_offset: u32) -> Self {
```
minestarks, line 188: Hmm, do you really need this additional method for this use case? You could just adjust the starting offset when you pass in the string slice:

```rs
`Lexer::new(&input[offset..])
```

Feels much cleaner to me.

compiler/qsc_parse/src/lex/concrete.rs
minestarks, line 1: I think this needs a few unit test cases to show how some of the weirder edge cases are handled, like the error tokens, apostrophe etc.

(Unless you consider this temporary until the "v2" implementation of the Formatter - then I'm not too concerned about it).

compiler/qsc_parse/src/lex/cooked.rs
```diff
@@ -66,11 +66,20 @@ impl Error {
             Self::Unknown(c, span) => Self::Unknown(c, span + offset),
         }
     }
+
+    pub(crate) fn get_span(self) -> Span {
```
minestarks, line 70: Just a naming convention thing - we don't tend to use `get_` for getters. https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

(To me `get` implies collection access, like `Vec::get` or `HashMap::get`)

compiler/qsc_parse/src/lex/concrete.rs
```diff
@@ -0,0 +1,132 @@
...
+                }
+                Err(err) => {
+                    let next_lo = self.get_next_lo();
+                    let span = err.get_span();
+                    self.get_tokens_from_span(span.hi, next_lo);
+                    let error = ConcreteToken {
+                        kind: ConcreteTokenKind::Error(err),
+                        span,
+                    };
+                    Some(error)
```
minestarks, line 127: What exactly happens if we swallow errors and keep going? Is there ever a next token after the error? Should we just stop instead of returning an Error token? An test case would be helpful, e.g. using an unterminated string literal.

compiler/qsc_formatter/src/formatter/tests.rs
minestarks, line 1: These are great!

minestarks (commented):
No major concerns, just a few more suggested changes

language_service/src/format.rs
ScottCarda-MS: Normally yes, but since `contents` is only used by reference later on, we can omit the clone here in favor of making a reference here.

compiler/qsc_parse/src/lex/raw.rs
ScottCarda-MS: The issue is that the raw tokens won't be created with the proper offsets for their spans, because to this instantiation of the Lexer, the beginning of the slice is the beginning of the whole user code. This is why it is necessary to have some way of telling the Lexer to start creating token offsets at a value other than zero.

wasm/src/language_service.rs
ScottCarda-MS: done

language_service/src/protocol.rs
ScottCarda-MS: It is now using the TextEdit defined in the LS protocol.rs file. Thanks!

vscode/src/format.ts
ScottCarda-MS: done

language_service/src/lib.rs
ScottCarda-MS: It would be relevant when we add the functionality to format highlighted sections of code, but since that feature is not implemented yet, it doesn't make sense to leave the parameter in. Removed.

compiler/qsc_parse/src/lex/cooked.rs
ScottCarda-MS: done

vscode/src/format.ts
ScottCarda-MS: We want to be able to track how often the feature is used to learn how effective it is.

compiler/qsc_parse/src/lex/raw.rs
minestarks: Sorry, yeah I got this backwards. In that case, my suggestion would be to add the offset the spans at the very end, when you're creating the Concrete tokens.

After all, this is a problem that we already have in the compiler. The lexer is created with offset zero, but then each source file needs to be offsetted per their SourceMap offset. The current solution in the compiler is to do it _after_ parsing (using the `Offsetter` visitor). Similarly, I think in your implementation, you can also do it at the very end. 

IMO, if the compiler didn't need this extra method on the offset neither should your implementation.

minestarks (approved):

compiler/qsc_parse/src/lex/concrete.rs
ScottCarda-MS: It should just skip past error tokens. Tests have been added.

compiler/qsc_parse/src/lex/concrete.rs
ScottCarda-MS: I've added a couple more test cases and addressed handling of error tokens. Thanks!

compiler/qsc_parse/src/lex/raw.rs
minestarks: Resolved offline - I finally grok the purpose of this method and think it's fine as is

library/std/arrays.qs
```diff
@@ -647,8 +647,11 @@ namespace Microsoft.Quantum.Arrays {
         for index in 0..interleavedLength - 1 {
             let originalIndex = index / 2;
             let value =
-                if index % 2 == 0 {first[originalIndex]}
-                else {second[originalIndex]};
+                if index % 2 == 0 {
+                first[originalIndex]
```
billti, line 651: Just browsing these files, and this doesn't look right. I assume it should be indented from the if statement above it?

library/std/unstable_arithmetic_internal.qs
```diff
@@ -323,12 +322,12 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
...
     internal operation ComputeCarries(ps : Qubit[], gs : Qubit[]) : Unit is Adj {
         let n = Length(gs);
-        Fact(Length(ps)+1 == n, "Register gs must be one qubit longer than register gs.");
+        Fact(Length(ps) + 1 == n, "Register gs must be one qubit longer than register gs.");
 
         let T = Floor(Lg(IntAsDouble(n)));
         use qs = Qubit[n - HammingWeightI(n) - T];
 
-        let registerPartition = MappedOverRange(t -> Floor(IntAsDouble(n) / IntAsDouble(2^t)) - 1, 1..T - 1);
+        let registerPartition = MappedOverRange(t -> Floor(IntAsDouble(n) / IntAsDouble(2 ^ t)) - 1, 1..T - 1);
```
billti, line 330: Hmmmm, I wonder if we should put spaces around exponentiation operators. That one does look a little weird.

samples/algorithms/PhaseFlipCode.qs
```diff
@@ -140,14 +140,14 @@ namespace Sample {
         // parity measurements.
         let indexOfError =
             if (parity01, parity12) == (One, Zero) {
-                0
-            } elif (parity01, parity12) == (One, One) {
-                1
-            } elif (parity01, parity12) == (Zero, One) {
-                2
-            } else {
+            0
```
billti, line 143: Looks like another case where this line and the following aren't indented correctly. Looks from this diff like the same indent level as the 'if' statement that it's within a block of.

samples/estimation/df-chemistry/src/df_chemistry.qs
```diff
@@ -20,28 +20,28 @@ namespace Microsoft.Quantum.Applications.Chemistry {
...
-        OneBodyEigenValues: Double[],
-        // eigenvectors in the EVD of the one-electron Hamiltonian (Rₖ, p. 54, eq. 67)
-        OneBodyEigenVectors: Double[][],
-        // norms inside Λ_SH (p. 56, eq. 77)
-        Lambdas: Double[],
-        // eigenvalues in the EVDs of the two-electron Hamiltonian for all r (λₖ⁽ʳ⁾, p. 56, eq. 77)
-        TwoBodyEigenValues: Double[][],
-        // eigenvectors in the EVDs of the two-electron Hamiltonian for all r (R⁽ʳ⁾ₖ, p. 56, eq. 77)
-        TwoBodyEigenVectors: Double[][][],
+    // Number of orbitals (N, p. 8)
```
billti, line 23: Again, should these be indented a level relative to line 22?

samples/estimation/df-chemistry/src/df_chemistry.qs
```diff
@@ -86,9 +86,9 @@ namespace Microsoft.Quantum.Applications.Chemistry {
     /// electron operators, which are computed from the double factorized
     /// problem and parameters.
     internal newtype DoubleFactorizedChemistryConstants = (
-        RotationAngleBitPrecision: Int,
-        StatePreparationBitPrecision: Int,
-        TargetError: Double
+        RotationAngleBitPrecision : Int,
```
billti, line 89: Note how this is indented correctly compared to a couple of the examples above. I wonder if it's the comments causing the issue. I notice this doesn't have comments on the line following the open parens, but a couple examples of it not indenting correctly above do have comments on the next line following the open paren.

library/src/tests/resources/state_preparation.qs
```diff
@@ -49,52 +49,52 @@ namespace Test {
...
         Adjoint PrepareComplex(q);
     }
 
-    operation TestPreparationCompletion(): Unit {
+    operation TestPreparationCompletion() : Unit {
         let testCases = [
-            // Test positive coefficients
-            [0.773761, 0.633478],
+        // Test positive coefficients
+        [0.773761, 0.633478],
```
billti, line 66: Another case where looks like comments on the line following an opening brace have cause a lack of indentation. Lines 65 & 66 should be in one level compared to 64 (interesting that 67 - 69 are indented correctly)

library/src/tests/resources/state_preparation.qs
```diff
@@ -49,52 +49,52 @@ namespace Test {
...
             [0.183017, 0.406973, 0.604925, 0.659502],
             [0.0986553, 0.359005, 0.465689, 0.467395, 0.419893, 0.118445, 0.461883, 0.149609],
             [0.271471, 0.0583654, 0.11639, 0.36112, 0.307383, 0.193371, 0.274151, 0.332542, 0.130172, 0.222546, 0.314879, 0.210704, 0.212429, 0.245518, 0.30666, 0.22773],
 
-            // Test negative coefficients; should give same probabilities as positive coefficients
-            [-0.773761, 0.633478],
+        // Test negative coefficients; should give same probabilities as positive coefficients
+        [-0.773761, 0.633478],
```
billti, line 72: Yup... definitely the comments causing it. Looks like most of these changes in this area are wrong, and always when there's a comment.

ScottCarda-MS: This, as well as the other places you noticed the indentation is off, is a known limitation of the way the formatter is structured currently. There is no way to fix this without completely rewriting the formatter, unfortunately.

library/std/unstable_arithmetic_internal.qs
ScottCarda-MS: I asked other people what they prefer. I can special-case the logic for the carrot character to have no spaces, if we really want.

library/src/tests/resources/state_preparation.qs
ScottCarda-MS: Alternatively, I can remove any indentation logic from the formatter, if that is preferrable.

samples/algorithms/BitFlipCode.qs
```diff
@@ -119,14 +119,14 @@ namespace Sample {
         // parity measurements.
         let indexOfError =
             if (parity01, parity12) == (One, Zero) {
```
minestarks, line 121: I think if we just change our sample so that `let indexOfError = if (parity01, parity12) .....` is on one line instead of being broken up with a newline, it works better with the current behavior.

vscode/package.json
```diff
@@ -108,6 +108,11 @@
             "The minimal set of capabilities required to run a quantum program. This option maps to the Base Profile as defined by the QIR specification."
           ],
           "description": "Setting the target profile allows the Q# extension to generate programs that are compatible with a specific target. The target is the hardware or simulator which will be used to run the Q# program. The target profile is a description of a target's capabilities."
+        },
+        "Q#.enableFormatting": {
+          "type": "boolean",
+          "default": "true",
+          "description": "Enables the Q# formatter to be used to format code."
```
billti, line 115: I think could just truncate this at `Enable the Q# formatter`

vscode/src/config.ts
```diff
@@ -26,3 +26,18 @@ export async function setTarget(target: TargetProfile) {
...
   );
 }
+
+export function getEnableFormating(): boolean {
+  return vscode.workspace
+    .getConfiguration("Q#")
+    .get<boolean>("enableFormatting", true);
+}
+
+export async function setEnableFormating(isEnabled: boolean) {
```
billti, line 36: Does this need to be implemented for any reason? Does anything call this?

vscode/package.json
ScottCarda-MS: done

vscode/src/config.ts
ScottCarda-MS: Hmm, I thought I had this being set somewhere programmatically, but looking now, I guess not. Removed.

samples/algorithms/BitFlipCode.qs
ScottCarda-MS: Formatting of the library and samples has been removed from this PR. They may be added in a near future PR.

library/std/arrays.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

library/std/unstable_arithmetic_internal.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

ScottCarda-MS: Let's keep this spacing for now. If people really complain about this, it is an easy one to update later.

samples/algorithms/PhaseFlipCode.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

samples/estimation/df-chemistry/src/df_chemistry.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

samples/estimation/df-chemistry/src/df_chemistry.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

library/src/tests/resources/state_preparation.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

library/src/tests/resources/state_preparation.qs
ScottCarda-MS: Indentation logic will be addressed by a future PR.

vscode/src/extension.ts
```diff
@@ -181,8 +182,24 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     ...registerQSharpNotebookCellUpdateHandlers(languageService),
   );
 
+  // format document
+  const isFormattingEnabled = getEnableFormating();
+  const formatterEvent = {
+    event: undefined as vscode.Disposable | undefined,
```
billti, line 188: Calling this 'formatterEvent' (and the member 'event') confused me for a bit. There's not any event handling here, right? Or even any event-related objects. It's just a container to hold the Disposable returned from registering the edit provider (if I'm reading the code correctly). I clearer name would be less confusing.

vscode/src/format.ts
```diff
@@ -0,0 +1,44 @@
...
+    // telemetry start format
+    const associationId = getRandomGuid();
+    sendTelemetryEvent(EventType.FormatStart, { associationId }, {});
+    const start = performance.now();
+
+    const lsEdits = await this.languageService.getFormatChanges(
+      document.uri.toString(),
+    );
+
+    if (!lsEdits) return [];
```
billti, line 27: Shouldn't you still send a FormatEnd event if it returns early? Else you have the FormatStart with no FormatEnd whenever this code path is taken.

billti (commented):
A couple minor things to fix up, and I didn't look at the Rust code too much as I assume Mine did, else looks good. Thanks!

billti (approved):

vscode/src/extension.ts
ScottCarda-MS: renamed to FormatterHandle

vscode/src/format.ts
ScottCarda-MS: Sending FormatEnd in no lsEdit case.

PR #1173, opened 2024-02-17
Add DumpRegister

### LINKED ISSUE #1138: Add DumpRegister
When I work on larger code that might have several different qubit registers serving different purposes, I want to be able to dump just the state of some of them, not of all systems. Typically, in such scenarios I know that these qubits are not entangled with the rest of the system, for example:

* In the katas, `ShowQuantumStateComparison` reuses the input register to show multiple states that can be prepared from it, uncomputing the preparation operation between them. This is inconvenient, since the state preparation operation might not have an adjoint defined. Instead, we could allocate an extra register and print the state of just that register.
* Similarly, for oracle-based problems, I need to check if the state of the input register didn't change, and if it did, let the user know - ideally printing its state separately from the state of the target qubit that is also allocated at this time.

`DumpRegister` functionality similar to that of `DumpMachine` but allowing to the state of just one register. (This was mentioned as part of https://github.com/microsoft/qsharp/issues/579, but I think it's worth a separate callout.)

If the qubits are entangled with others, I'm fine with a message telling me this, like the classic QDK did, I'm usually interested in cases when they're not.

swernli:
This change adds an implementation of `DumpRegister` as an intrinsic in the standard library with matching implementation in the evaluator. It works by getting a full state dump and calculating if the state corresponding to the given qubits is separable from the rest, returning an error if it is not.

Fixes #1138


library/std/diagnostics.qs
```diff
@@ -4,10 +4,71 @@
...
+    /// The actual information generated and the semantics are specific to each target machine.
+    ///
+    /// For the local sparse-state simulator distributed as part of the
+    /// Quantum Development Kit, this method will write the wave function as a
+    /// one-dimensional array of pairs of state indices and complex numbers, in which each element represents
+    /// the amplitudes of the probability of measuring the corresponding state.
+    ///
+    /// # Example
+    /// When run on the sparse-state simulator, the following snippet dumps
+    /// the Bell state $(\ket{00} + \ket{11}) / \sqrt{2}$ to the console:
```
billti, line 21: Didn't we decide to limit LaTeX to the 'Remarks' section? (CC @ScottCarda-MS)

ScottCarda-MS: Yes we should avoid LaTex except in the Remarks section. In this case, we can use the ket unicode character: `〉` and the sqrt unicode character: `√` although we've also just referred to that as `sqrt(2)` in other places.

compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+    )
+}
+
+trait NearlyZero {
+    fn is_nearly_zero(&self) -> bool;
+}
+
+impl NearlyZero for f64 {
+    fn is_nearly_zero(&self) -> bool {
+        self.max(0.0) - 0.0_f64.min(*self) <= 1e-10
```
billti, line 119: Minor suggestion, but was dealing with some math libs over the weekend and most commonly saw this done as something like `self.abs() <= 1e-10`. That might read a little easier and be less operations to perform.

compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // If the product of the collected states is not equal to the total number of input states, then that
+    // implies some states are zero amplitude that would have to be non-zero for the state to be separable.
+    if state.len() != dump_state.len() * other_state.len() {
+        return None;
+    }
+
+    let dump_norm = 1.0 / dump_norm.sqrt();
+    Some(
```
DmitryVasilevsky, line 86: When reading this code it's hard to tell which numbers are real and which are complex. Maybe explicit typing would help understanding.

compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+        if !(dump_val * other_val - base_val * curr_val)
+            .norm()
+            .is_nearly_zero()
+        {
+            // Coefficients are not equal, so the state is not separable.
+            return None;
+        }
+
+        if let Entry::Vacant(entry) = dump_state.entry(dump_label) {
+            let amplitude = curr_val / other_val;
```
cesarzc, line 63: I'm not sure I understand why the amplitude is the division of the current label value over the other value.

compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+use num_bigint::BigUint;
+use num_complex::{Complex, Complex64};
+use num_traits::{One, Zero};
+use rustc_hash::FxHashMap;
+
+pub fn split_state(
+    qubits: &[usize],
+    state: Vec<(BigUint, Complex64)>,
+    qubit_count: usize,
+) -> Option<Vec<(BigUint, Complex64)>> {
```
cesarzc, line 15: I think returning `Result<Vec<(BigUint, Complex64)>, ()>` would be more appropriate since it conveys the fact that an error happens when trying to split a non-separable state.

cesarzc: And it would also make it easier to quickly see the cases where this happens.

minestarks:
Should we add some sample code for this? In `samples/language/Diagnostics.qs` maybe?

compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Try to split out the state for the given qubits from the whole state, detecting any entanglement
+    // and returning `None` if the qubits are not separable.
+    let mut state_iter = state.iter();
+    let (base_label, base_val) = state_iter.next().expect("state should never be empty");
+    let dump_base_label = base_label & &dump_mask;
+    let other_base_label = base_label & &other_mask;
+    let mut dump_norm = 1.0_f64;
+    dump_state.insert(dump_base_label.clone(), Complex64::one());
+    other_state.insert(other_base_label.clone(), Complex64::one());
+    for (curr_label, curr_val) in state_iter {
```
DmitryVasilevsky, line 41: This function is rather long. Maybe extract part that tries to separate qubits and extract another part that computes new coefficients for separated qubits into their own functions.

compiler/qsc_eval/src/intrinsic/utils.rs
swernli: I added a comment to try and clarify. The amplitude we are capturing here is intermingled in the original state vector, so we calculated the amplitude of the "other" state and divide that out to get the amplitude of just the desired qubits.

compiler/qsc_eval/src/intrinsic/utils.rs
swernli: I split it out into several helper functions, hopefully that helps.

swernli:
> Should we add some sample code for this? In `samples/language/Diagnostics.qs` maybe?

Probably, though likely not until we have actually released a version with `DumpRegister`.

cesarzc (approved):

idavis (approved):

Issue #1081, opened 2024-01-29
Remove "Message:" prefix from the output from Message function (running in VS Code)


tcNickolas:
**Describe the bug**

Prefacing the output from the Message function with "Message:" (when running in VS Code) looks very jarring to me, and not what I expect. It makes sense to call out "Result" explicitly, since this happens once per program, and DumpMachine output is large enough that a "DumpMachine" prefix doesn't stand out, but the "Message" prefix looks noisy.

**Screenshots**

For example, in the following example I use newlines to add a break between different parts of the example, and having "Message:" strings makes it look odd:

![image](https://github.com/microsoft/qsharp/assets/10113024/fb463f15-dc56-44dc-b170-302e3a966dcb)

**System information**

- QDK 1.0.33

colommar:
Hi, I'd like to clarify the requirements. Is it ideal to remove the `message:` and directly output the content in the message?

tcNickolas:
Yes

PR #1175, opened 2024-02-20
message will directly output the content in the message without typing the `Message:`

colommar:
## What have been changed
delete the debugger output `Message:`

## Issue mentioned
finish the open issue #1081.

## What will happen
message will directly output the content in the message without typing the `Message:`


vscode/src/debugger/output.ts
```diff
@@ -7,7 +7,7 @@ export function createDebugConsoleEventTarget(out: (message: string) => void) {
   const eventTarget = new QscEventTarget(false);
 
   eventTarget.addEventListener("Message", (evt) => {
-    out(`Message: ${evt.detail}`);
+    out(`${evt.detail}`);
```
billti, line 10: There's no point in having an interpolated string any more. May as well just `out(evt.detail);`.

minestarks (approved):

Issue #1120, opened 2024-02-05
Rust library panic strings not surfaced in WASM


swernli:
Currently any panics from the underlying Rust code that come from a web worker all show up as a generic `Error: unexpected` in the VSCode extension. We need to find the right behavior for our WASM panic handling to make sure at the very least the original strings (and possibly later some additional explanatory text) can get surfaced to improve diagnosability without having to attach a debugger.

One current theory: the existing panic hook that gets configured as part of setting up logging is either not configured or not able to function as expected when run from the web worker. Three scenarios today use the web worker: QIR code generation, histogram/shots execution, and resource estimation. We can test with explicit panics in these code paths and find a mechanism (perhaps message passing?) that lets the worker report the panic back to the main thread.

minestarks:
Fixed by #1176

PR #1176, opened 2024-02-20
Forward log messages from worker thread to main thread

minestarks:
Proposed fix for #1120

Note that this fix has some downsides as well. So I'm not 100% sure we should take it yet. Looking for opinions.

### Problem
- When workers (i.e. the compiler and debugger workers in VS Code) write logs, these are not surfaced in the VS Code output window along with other logs. This applies to logging from JavaScript (e.g `log.info()` calls), logs from Rust (e.g. `info!()` macros), and Rust panics. 

### Fix
- Workers, instead of logging directly to the console, now send a message to the main thread. The main thread logs the message as normal (to the Developer Tools console for playground/website, to the Output window for VS Code). All logs are now visible in the VS Code output window.

### Concerns about this fix
- Logging becomes a nontrivial code path, which may defeat the purpose of using logging to debug tricky bugs.
- For log messages, the original callstack in the worker is lost. Instead the callstack from the message handler in the main thread is shown (see screenshots). This will *not* be a problem for panic messages, which capture a stack trace explicitly. Personally, I wouldn't expect to see callstacks in log messages anyway, it's just something that Chrome does.
- Log messages now become part of the message traffic between the main thread and the worker thread, which _could_ be a perf concern if noisy logs start slowing down the "real" communication between the threads. This is hypothetical; I haven't been able to find an actual scenario where this is a problem. We can mitigate this by being careful we're not writing super-noisy logs at the default log level ("info" and up).

### Alternate solution
- Do nothing! There's actually a way to access the console logs for the worker thread in VS Code. You can bring up the Chrome developer tools using the "Toggle Developer Tools" VS Code command. It's not as convenient as having them in the output window, but they're there.

### Notes
- Once a worker is started, there's no way to set the log level for that worker. This is a preexisting issue, and probably not a big deal in practice, since most of the workers (compiler, debugger) are short-lived and can just be run again with a different log level if we're trying to repro a specific issue.
- A bonus fix here is increasing the stack trace frames to 20 (from 10) when logging panics, since most of the stack trace tends to get taken up by the logging code itself.

### Screenshots

These show an `error!` log and a `panic!` surfaced from the WASM module.

VS Code:

**Before, release build, VS Code**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 150303

</details>

**After, release build, VS Code** 

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 145822

</details>

**Before, debug build, VS Code**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 150600

</details>

**After, debug build, VS Code**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 145544

</details>

**Before, release build, Developer Tools window in VS Code**

<details>
<summary>Screenshot</summary>

image

</details>

Playground - no real improvement here, since the worker logs were already shown in the Developer Tools console:

**Before, release build, playground**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 150200

</details>

**After, release build, playground**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 142645

</details>

**Before, debug build, playground**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 150503

Screenshot 2024-02-18 150519

</details>

**After, debug build, playground**

<details>
<summary>Screenshot</summary>

Screenshot 2024-02-18 145153

</details>


npm/src/workers/common.ts
```diff
@@ -425,28 +448,57 @@ export function initService<
...
+      detail: { level, target, data },
+    });
+  }
+
+  // Override the global logger
+  log.error = (...args) => postLogMessage(1, "worker", ...args);
+  log.warn = (...args) => postLogMessage(2, "worker", ...args);
+  log.info = (...args) => postLogMessage(3, "worker", ...args);
+  log.debug = (...args) => postLogMessage(4, "worker", ...args);
+  log.trace = (...args) => postLogMessage(5, "worker", ...args);
```
billti, line 492: Does this send all the messages from the worker without filtering on level first? That can be quite chatty.

minestarks: Ouch, yes, good catch. I'll fix this

minestarks:
@billti fixed the perf issue and tested it, could you sign off if all looks good?

billti (approved):

PR #1190, opened 2024-02-22
Clean up QRNG kata

tcNickolas:
Also adds task 1.5 to Superposition kata and updates docs on `@[exercise]` macro

### DIFF
--- a/katas/README.md
+++ b/katas/README.md
@@ -20,10 +20,14 @@ Macros are meant to insert interactive elements into the content defined through
 The following macros are available for katas composition:
 - @[exercise]: Used to create Q# code exercises that we can be automatically verified.
     - id: Unique identifier for the exercise.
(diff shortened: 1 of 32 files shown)


cesarzc (approved):
The contents of this PR look good, but I think I found an error that I missed during the review of the first part of the superposition kata.

In exercise 4 "Even superposition of all basis vectors on two qubits with phase flip", I think the goal should be the following instead of what it currently has:

> **Goal:**  Change the state of the qubits to \frac{1}{2} \big(|00\rangle + |01\rangle + |10\rangle - |11\rangle\big)$.

Removing "$|+\rangle \otimes |+\rangle =" from the current goal description.

This PR might be a good time to do the change so we don't forget.

tcNickolas:
Well spotted! I fixed that

PR #1191, opened 2024-02-22
Generic code parameter in error correction trait

msoeken:
The `ErrorCorrection` trait is generalized such that it uses a code parameter instead of a code distance to compute the logical error rate, number of physical qubits, and cycle time. The code parameter's type is an associated type `Parameter` to the `ErrorCorrection` trait. It could be a numeric type, e.g., to represent a code distance, but it could also be a tuple type, to represent multiple parameters that influence the metrics above. A new trait function is added to traverse the possible space of values for the code parameter.

Further, this PR
- adds a test case to the system that checks consistency of the result output, in order to detect changes that break this format,
- removes serialization from the `estimate` folder and moves it to the `system` folder.


resource_estimator/src/system/data/report.rs
```diff
@@ -67,7 +67,7 @@ impl Report {
...
-        entries.push(ReportEntry::new("physicalCountsFormatted/logicalErrorRate", "Logical qubit error rate", r#"Logical qubit error rate"#, &format!(r#"The logical qubit error rate is computed as ${} \cdot \left(\dfrac{{{}}}{{{}}}\right)^\frac{{{} + 1}}{{2}}$"#, job_params.qec_scheme().crossing_prefactor.expect("crossing prefactor should be set"), result.logical_qubit().physical_qubit().clifford_error_rate(), job_params.qec_scheme().error_correction_threshold.expect("error correction threshold should be set"), result.logical_qubit().code_distance())));
+        entries.push(ReportEntry::new("physicalCountsFormatted/logicalErrorRate", "Logical qubit error rate", r#"Logical qubit error rate"#, &format!(r#"The logical qubit error rate is computed as ${} \cdot \left(\dfrac{{{}}}{{{}}}\right)^\frac{{{} + 1}}{{2}}$"#, job_params.qec_scheme().crossing_prefactor.expect("crossing prefactor should be set"), result.logical_qubit().physical_qubit().clifford_error_rate(), job_params.qec_scheme().error_correction_threshold.expect("error correction threshold should be set"), result.logical_qubit().code_parameter())));
```
msoeken, line 70: Double check if code_distance needs to be renamed in markdown generation file.

resource_estimator/src/system/modeling/fault_tolerance.rs
```diff
@@ -487,16 +489,16 @@ impl ErrorCorrection for Protocol {
     /// extraction time, is based on physical operation times specified in the
     /// qubit and usually some factor based on the choice of stabilizer
     /// extraction circuit.
-    fn logical_cycle_time(&self, qubit: &Self::Qubit, code_distance: u64) -> Result<u64, String> {
-        let mut context = Self::create_evaluation_context(Some(qubit), code_distance);
+    fn logical_cycle_time(&self, qubit: &Self::Qubit, code_distance: &u64) -> Result<u64, String> {
```
msoeken, line 492: Use associated type or assigned type (choose for consistency)

ivanbasov (approved):
reviewed on a sync

resource_estimator/src/system/data/report.rs
msoeken: Double-checked. Generated output is based on JSON report structure, which is not changed by this PR.

resource_estimator/src/system/modeling/fault_tolerance.rs
msoeken: Changed to use assigned type for `Self::Qubit`.

PR #1194, opened 2024-02-22
Doc Gen Improvements

ScottCarda-MS:
Some improvements to the docs generation:
 - Removes newlines from Summary in metadata section.
 - Prefixes uid's with "Qdk." to disambiguate the new generated docs from the legacy docs.
 - Alphabetizes the Table of Contents.
 - Keeps the metadata separate from the user-facing content of the files until the npm layer.


compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -260,8 +261,12 @@ fn get_metadata(ns: Rc<str>, item: &Item, display: &CodeDisplay) -> Option<Metad
         ItemKind::Namespace(_, _) => None,
     }?;
 
+    let summary = parse_doc_for_summary(&item.doc)
+        .replace("\r\n", " ")
+        .replace('\n', " ");
+
     Some(Metadata {
-        uid: format!("{ns}.{name}"),
+        uid: format!("Qdk.{ns}.{name}"),
```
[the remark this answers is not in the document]
ScottCarda-MS: This is just a temporary workaround while the legacy docs are in deprecation. There are currently no plans to generate docs that aren't Microsoft docs.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -260,8 +261,12 @@ fn get_metadata(ns: Rc<str>, item: &Item, display: &CodeDisplay) -> Option<Metad
         ItemKind::Namespace(_, _) => None,
     }?;
 
+    let summary = parse_doc_for_summary(&item.doc)
+        .replace("\r\n", " ")
+        .replace('\n', " ");
```
billti, line 266: Why is replacing newlines with spaces needed here?

ScottCarda-MS: They had said that newlines in the summary were not compatible with their tooling and so needed to be removed.

PR #1202, opened 2024-02-26
Use optimized AND for decomposition 

### LINKED ISSUE #1137: Modern ApplyControlledOnInt implementation less efficient for multiple controls compared to classic ControlledOnInt
**Describe the bug**

When I run resource estimation for a small program consisting of a series of multi-controlled gates, I get the same number of qubits in the resource estimate, but about 80% longer runtime (and larger circuit depth). 

**To Reproduce**

In the Modern QDK resource estimator, a sequence of 6 gates, each with 5 control qubits and 1 target, takes 144 T gates and 6 CCZ -> 162 logical depth. In the Classic QDK, the same sequence took 18 CCiX, 18 measurement, and 6 CCZ -> 90 logical depth.

This appears to be (as @msoeken explained) due to the difference in how `ApplyControlledOnInt` is implemented: classic QDK decomposed multi-controlled Toffoli gates into pairs of AND + Adjoint AND (which corresponds to one CCiX + one measurement each, in total 18), and modern QDK decomposes them into less efficient relative-phase Toffoli gate (and its adjoint), which require 4 Ts each (i.e., 18 * (2 * 4) = 144 T gates)

**Expected behavior**

It would be good to use a more efficient decomposition that would not increase resource estimates compared to the classic QDK. (@msoeken suggested an implementation using `ApplyAsSinglyControlled`)

swernli:
This allows the "unrestricted" configuration to use an optimized `AND` for multi-controlled decomposition that relies on measurement-based adjoint. The existing `PhaseCCX` will still be used for "base" configuration since branching measurement is not allowed there.

Fixes #1137

### DIFF
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -12,6 +12,7 @@ mod debugger_tests;
 pub use qsc_eval::{
     debug::Frame,
     output::{self, GenericReceiver},
+    val::Result,
     val::Value,
     StepAction, StepResult,
 };
@@ -112,7 +113,7 @@ pub struct Interpreter {
 }
 
 #[allow(clippy::module_name_repetitions)]
-pub type InterpretResult = Result<Value, Vec<Error>>;
+pub type InterpretResult = std::result::Result<Value, Vec<Error>>;
 
 impl Interpreter {
     /// Creates a new incremental compiler, compiling the passed in sources.
@@ -123,7 +124,7 @@ impl Interpreter {
         sources: SourceMap,
         package_type: PackageType,
         capabilities: RuntimeCapabilityFlags,
-    ) -> Result<Self, Vec<Error>> {
+    ) -> std::result::Result<Self, Vec<Error>> {
         let mut lowerer = qsc_eval::lower::Lowerer::new();
         let mut fir_store = fir::PackageStore::new();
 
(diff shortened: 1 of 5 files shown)


msoeken (approved):
Looks good to me, thank you.

We could replace `M.Q.Unstable.Arithmetic.ApplyAndAssuming0Target` by `AND` in this PR.

Further, I suggest to consider making `AND` a public operation.

swernli:
> Looks good to me, thank you.
> 
> We could replace `M.Q.Unstable.Arithmetic.ApplyAndAssuming0Target` by `AND` in this PR.
> 
> Further, I suggest to consider making `AND` a public operation.

We'll definitely be discussing when/how to expose this as a public API. For now though, we can't have them both use the same utility function because they are in different namespaces and cannot see the callables marked as `internal` in the other.

msoeken:
> > Looks good to me, thank you.
> > We could replace `M.Q.Unstable.Arithmetic.ApplyAndAssuming0Target` by `AND` in this PR.
> > Further, I suggest to consider making `AND` a public operation.
> 
> We'll definitely be discussing when/how to expose this as a public API. For now though, we can't have them both use the same utility function because they are in different namespaces and cannot see the callables marked as `internal` in the other.

We could make it public in the unstable namespace to avoid duplication before discussing the public API in a non-unstable namespace.

swernli:
> We could make it public in the unstable namespace to avoid duplication before discussing the public API in a non-unstable namespace.

That's certainly possible, yes. I just would feel a bit awkward about having most callables in the Microsoft.Quantum.Intrinsic namespace depend on something from an unstable namespace. Feels a bit like an inversion...

The other concern I have is with the adjoint and how it is used, but that might be a longer conversation than just for this PR. What I'd like is to get this merged and then open an issue about making AND a supported intrinsic in QIR and public in Q#. Then we can get into the details of the design in that issue.

DmitryVasilevsky (approved):
Approving library change.

minestarks (approved):

PR #1205, opened 2024-02-26
RE add code distance to scatter chart label and connecting scatter circles

### LINKED ISSUE #1102: RE space-time diagram: add code distance to tooltips
The code distance explains gaps between points on the diagram. So, we could proactively respond to potential questions regarding the nature of specific features of pareto frontiers.

### LINKED ISSUE #1189: Resource estimator: circles on the scatter chart makes an illusion that plots are not monotonous but they actually are
Circles have radius > 1
If there are two solutions with very close numbers of qubits (10000 vs 10001) which should be placed with the same y-value and there is a large time gap between them (for the scale of the chart).

Then the bottom of the circle for the shorter solution would be displayed below the top of the circle for the logner solution. Their centers are on the same line. But visually they appear not monotonically decreasing.

We may consider adding a thin line below/above circles to connect their centers.

ivanbasov:
fixes #1189 
fixes #1102

### DIFF
--- a/npm/ux/estimatesOverview.tsx
+++ b/npm/ux/estimatesOverview.tsx
@@ -82,7 +82,9 @@ function frontierEntryToPlotEntry(entry: FrontierEntry): PlotItem {
     label:
       entry.physicalCountsFormatted.runtime +
       ", physical qubits: " +
-      entry.physicalCountsFormatted.physicalQubits,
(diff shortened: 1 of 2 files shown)


ivanbasov:
image

ivanbasov:
image

PR #1217, opened 2024-02-28
Fix ordering of output from `DumpRegister`

swernli:
`DumpRegister` output was incorrectly non-deterministic and not ordered because it was coming from a hash map. This fixes that by sorting the reordered labels before returning the vector of results.

### DIFF
--- a/compiler/qsc_eval/src/intrinsic/utils.rs
+++ b/compiler/qsc_eval/src/intrinsic/utils.rs
@@ -39,12 +39,14 @@ pub fn split_state(
     }
 
     let dump_norm = 1.0 / dump_norm.sqrt();
-    Ok(dump_state
+    let mut dump_state = dump_state
         .into_iter()
         .filter_map(|(label, val)| {
             normalize_and_reorder(val, dump_norm, qubits, &label, qubit_count)
         })
-        .collect())
+        .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


compiler/qsc_eval/src/intrinsic/utils.rs
```diff
@@ -39,12 +39,14 @@ pub fn split_state(
...
     let dump_norm = 1.0 / dump_norm.sqrt();
-    Ok(dump_state
+    let mut dump_state = dump_state
         .into_iter()
         .filter_map(|(label, val)| {
             normalize_and_reorder(val, dump_norm, qubits, &label, qubit_count)
         })
-        .collect())
+        .collect::<Vec<_>>();
+    dump_state.sort_by(|(a, _), (b, _)| a.cmp(b));
```
billti, line 48: So the 'reorder' part of 'normalize_and_reorder' doesn't handle the sorting then?

billti (approved):

minestarks (approved):

compiler/qsc_eval/src/intrinsic/utils.rs
swernli: It reorders the bits in the label, but doesn't reorder the entries themselves. As in, `010` might become `100`, but something separately needs to look at all the reordered labels and sort the tuples by them.

idavis (approved):

PR #1219, opened 2024-02-29
refactor compiler API in npm package

### LINKED ISSUE #1201: Refactor `compiler.ts` API
See [this comment](https://github.com/microsoft/qsharp/pull/1089#discussion_r1501112035) and [this comment thread](https://github.com/microsoft/qsharp/pull/1089#discussion_r1491828824) for context. We'd like to introduce config objects, swap argument ordering, and remove `shots` from `ProgramConfig`, going forward.

sezna:
This PR refactors the compiler API in the npm package to be extensible, but backwards-compatible.
In doing so, this PR closes #1201.

This PR introduces `ProgramConfig`, an abstraction meant to encapsulate whatever inputs go into
a compile. For now, this is just the source code and what language features you want turned on.

We believe this will be more extensible going forward, as modifying a config object type is easier to
make non-breaking than modifying a parameters directly.

### DIFF
--- a/npm/src/compiler/compiler.ts
+++ b/npm/src/compiler/compiler.ts
@@ -24,51 +24,44 @@ type Wasm = typeof import("../../lib/node/qsc_wasm.cjs");
 // for running the compiler in the same thread the result will be synchronous (a resolved promise).
 export interface ICompiler {
   checkCode(code: string): Promise<VSDiagnostic[]>;
+
   getHir(code: string, languageFeatures?: string[]): Promise<string>;
+
+  /** @deprecated -- switch to using `ProgramConfig`-based overload. Instead of passing
+   * sources and language features separately, pass an object with named properties. This change was made
+   * for the sake of extensibility and future-compatibility. Note that only the new API
+   * supports passing language features. If you need to pass language features, you must use
+   * the new API.
+   **/
+  run(
+    sources: [string, string][],
+    expr: string,
+    shots: number,
+    eventHandler: IQscEventTarget,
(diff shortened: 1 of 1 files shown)


npm/src/compiler/compiler.ts
```diff
@@ -37,31 +53,35 @@ export interface ICompiler {
     shots: number,
     eventHandler: IQscEventTarget,
   ): Promise<void>;
-  run(config: ProgramConfig, eventHandler: IQscEventTarget): Promise<void>;
+
+  /** @deprecated -- switch to using `ProgramConfig`-based overload. Instead of passing
+   * sources and language features separately, pass an object with named properties. This change was made
+   * for the sake of extensibility and future-compatibility. Note that only the new API
+   * supports passing guage features. If you need to pass language features, you must use
```
minestarks, line 60: Typo (here and in the copy-pasted comments): passing guage features

npm/src/compiler/compiler.ts
```diff
@@ -25,8 +25,24 @@ type Wasm = typeof import("../../lib/node/qsc_wasm.cjs");
 export interface ICompiler {
   checkCode(code: string): Promise<VSDiagnostic[]>;
   getHir(code: string, languageFeatures?: string[]): Promise<string>;
+  run(
```
minestarks, line 28: Extreme nit: I prefer to organize these methods such that the overloads are next to each other. Would make the diff easier to read too I think

minestarks (approved):
Thank you!

PR #1220, opened 2024-02-29
Correct parameter in rotation synthesis formula

### LINKED ISSUE #1218: Resource Estimator - T states per rotation formula
**Describe the bug**

The formula used by the Resource Estimator for number of T states to implement an arbitrary rotation does not match the cited paper. 

The formula used by the Resource Estimator is:
T ( \epsilon ) = 0.53 log2 ( 1 / \epsilon ) + 5.3

In practice, the number of T states per rotation is:
```
                (NUM_TS_PER_ROTATION_A_COEFFICIENT
                    * ((self.rotation_count as f64) / eps_synthesis).log2()
                    + NUM_TS_PER_ROTATION_B_COEFFICIENT)
                    .ceil() as _,
``` 
[(see logical_counts.rs)](https://github.com/microsoft/qsharp/blob/main/resource_estimator/src/system/data/logical_counts.rs)
where:
```
/// A coefficient in Ts per rotation
pub const NUM_TS_PER_ROTATION_A_COEFFICIENT: f64 = 0.53;

/// A coefficient in Ts per rotation
pub const NUM_TS_PER_ROTATION_B_COEFFICIENT: f64 = 5.3;
```
[(see constants.rs)](https://github.com/microsoft/qsharp/blob/main/resource_estimator/src/system/constants.rs)

This information is seen by the user in the "Resource estimates breakdown" section of the estimate report. For example, the double-factorized chemistry sample shows:

The number of T states to implement a rotation with an arbitrary angle is ⌈0.53log2(11,988,044/0.0033333333333333335)+5.3⌉ [arXiv:2203.10064]. For simplicity, we use this formula for all single-qubit arbitrary angle rotations, and do not distinguish between best, worst, and average cases.

However, the linked paper [arXiv:2203.10064](https://arxiv.org/abs/2203.10064) does not include this formula. 

The table on page 8 of the paper lists a "Mixed fallback" approximation protocol for Clifford+T (T-count) as `0.53 log2 (1/ \epsilon ) + 4.86`  and the "Mix fallback" approximation protocol for Clifford+sqrt(T) (T count) as `0.56 log2 (1 /\epsilon) + 5.30`. 

image

**To Reproduce**
N/A

**Expected behavior**

Either the formula should match the paper or an explanation of the formula should be provided.

**Screenshots**

image

**System information**
N/A

**Additional context**
N/A

msoeken:
Fixes #1218


billti (approved):

tcNickolas:
One other place where this constant shows up is https://github.com/microsoft/qsharp/blob/9754ef65578814e6788128623bace43bd2eee00e/npm/ux/output_data.md?plain=1#L137 - not sure if that's part of a unit test or the text from which field explanations are taken

msoeken:
@sezna, @tcNickolas I updated other occurrences of the constants and fixed most of the failing tests. One test still fails, and I am looking into what is causing this.

PR #1226, opened 2024-03-03
Allow QEC to encode multiple logical qubits

msoeken:
This generalizes the `ErrorCorrection` trait to allow encoding multiple logical qubits, thereby enabling the modeling of codes such as LDPC. The change also lead to rename the `LogicalQubit` struct into `LogicalPatch` to account the possibility of multiple logical qubits in the name of the struct.

The system does not change, since all the planar codes modeled in it encode one logical qubit.


msoeken:
> Just to confirm: the RE result and UI (tables, charts) will still contain LogicalQubit so far. Correct?

Correct.

PR #1227, opened 2024-03-04
Update `dump_machine` and `dump_operation` based on feedback

### LINKED ISSUE #1177: Python API dump_machine should represent amplitudes as complex numbers instead of tuples
**Describe the bug**

Python API `dump_machine()` represents amplitudes of the state as tuples. This is very inconvenient, since the first thing I need to do with its return is get rid of the global phase, which requires converting the amplitudes to complex type. (The complex global phase shows up even in simple scenarios that use only Ry gates, so I couldn't find a case where having only tuples would suffice.)

**To Reproduce**

Run `dump_machine()`, for example, using https://github.com/microsoft/qsharp/blob/36a091c23144014a3f9e64169210a30262f9095f/pip/tests/test_qsharp.py#L63. 

**Expected behavior**

Return a dictionary of complex numbers, similar to Qiskit statevector behavior.

**Additional context**

I can unpack the tuple representing the amplitude and convert its fields into `complex` type manually, but I'll need to do it every time I use `dump_machine()`, so it really makes sense to have the API do it instead.

### LINKED ISSUE #1178: Python API dump_operation does not work if operation has Adj/Ctl specializations
**Describe the bug**

Python API `dump_operation()` fails with `Qsc.TypeCk.FunctorMismatch: expected empty set, found Adj + Ctl` error if operation has Adj/Ctl specializations. This doesn't really make sense, since Q# can ignore additional specializations in any other context.

**To Reproduce**

```qsharp
%%qsharp

operation ControlledRxWrapper(
    qs : Qubit[],
    theta : Double
) : Unit is Adj + Ctl {            // Remove "is Adj + Ctl here to make the code work"
    Controlled Rx(Most(qs), (theta, Tail(qs)));
}
```

```python
import qsharp.utils
qsharp.utils.dump_operation("ControlledRxWrapper(_, 1.0)", 2)
```

**Expected behavior**

I expect `dump_operation()` to produce a valid result regardless of specializations the operation has.

**Additional context**

Here is the more detailed error message; looks like the highlighted assignment is too strict in specializations handling.

```QSharpError: Qsc.TypeCk.FunctorMismatch

  × type error
  ╰─▶ expected empty set, found Adj + Ctl
   ╭─[<entry>:1:1]
 1 │ {
 2 │         let op : (Qubit[] => Unit) = ControlledRxWrapper(_, 1.0);
   ·                                      ───────────────────────────
 3 │         use (targets, extra) = (Qubit[2], Qubit[2]);
   ╰────
```

### LINKED ISSUE #1223: Python dump_machine output is not iterable?
**Describe the bug**

The map that is the output of `qsharp.dump_machine()` call does not support `in` to check whether a certain state is present in it. This behavior seems to point to it not being iterable, though there might be a different reason.

**To Reproduce**

Here is an example of the code that tests Q# state preparation logic by using `dump_machine` and doing the checks in Python. Adding `if ind in state:` part to the test makes the code stuck here forever, not responsive to Ctrl+C; without this part, the test completes successfully.

```python
def test_prep_two_qubit(a):
  qsharp.init(project_root='.')
  qsharp.eval(f"use qs = Qubit[2]; StatePreparation.PrepTwoQubits(qs, {a});")
  state = qsharp.dump_machine()
  first_ind = -1
  first_amp_cp = 0
  for ind in range(4):
    if abs(a[ind]) > 1E-9:
      (real, imag) = state[ind]
      # Do the check
    else:
      if ind in state:             # This is stuck forever
        raise "Non-zero amp that should be zero"
```

**Expected behavior**

I'd expect this code to run successfully to check whether the map has an element with this key.

**System information**

- qsharp 1.1.3

swernli:
This updates the Python functions for dumping state vectors and operation matrices to both use complex numbers, provide
better compatibility with input, and allow for common iteration patterns.

Fixes #1178
Fixes #1177
Fixes #1223


pip/tests/test_qsharp.py
```diff
@@ -72,12 +72,12 @@ def test_dump_machine() -> None:
...
     assert state_dump.qubit_count == 2
     assert len(state_dump) == 1
-    assert state_dump[2] == (1.0, 0.0)
+    assert state_dump[2] == complex(1.0, 0.0)
     qsharp.eval("X(q2);")
     state_dump = qsharp.dump_machine()
     assert state_dump.qubit_count == 2
     assert len(state_dump) == 1
-    assert state_dump[3] == (1.0, 0.0)
+    assert state_dump[3] == complex(1.0, 0.0)
```
tcNickolas, line 81: I would add a test for iteration through state_dump somewhere here, I don't think there's a test for it right now?

pip/tests/test_qsharp.py
```diff
@@ -72,12 +72,12 @@ def test_dump_machine() -> None:
...
 
 def test_dump_operation() -> None:
```
tcNickolas, line 82: Is there a test for dump_operation of operation that has Adj/Ctl? It looks like lambdas should cover that, but the existing tests passed before the fix, so they probably don't

tcNickolas (commented):
Looking at the tests, the changes look good, but might need more tests to cover the scenarios that are being fixed

pip/tests/test_qsharp.py
swernli: Good call, I'll add an iteration case that also includes a check for membership to confirm both of those syntaxes work as expected.

pip/tests/test_qsharp.py
swernli: I can update the `ApplySWAP` test below to explicitly use `Ctl + Adj` so that we know that compatibility is being tested.

tcNickolas (approved):

pip/qsharp/_native.pyi
```diff
@@ -141,9 +141,9 @@ class Output:
     def __repr__(self) -> str: ...
     def __str__(self) -> str: ...
     def _repr_html_(self) -> str: ...
-    def state_dump(self) -> Optional[StateDump]: ...
+    def state_dump(self) -> Optional[StateDumpData]: ...
 
-class StateDump:
```
DmitryVasilevsky, line 146: For my own curiosity, why was the class renamed?

DmitryVasilevsky (approved):

pip/qsharp/_native.pyi
swernli: This file, as explained to me by @minestarks, is `.pyi` instead of `.py` making it an interface file. So there is no implementation allowed here, it's basically just the manually declared interfaces for what comes out of Rust via PyO3. Because I wanted to add implementation, I had to create a wrapper type to expose the functionality I wanted, and renamed the native type that should stay internal.

That said, your question made me realize the editor rename incorrectly updated the name of the exposed type, which was not intentional. I'll get that fixed!

pip/qsharp/_qsharp.py
```diff
@@ -273,10 +273,46 @@ def set_classical_seed(seed: Optional[int]) -> None:
     """
     get_interpreter().set_classical_seed(seed)
 
+class StateDump:
```
minestarks, line 276: I wonder if you'd get to avoid duplicating methods if you made this a subclass of `StateDumpData` instead of having it as a member?

pip/qsharp/__init__.py
```diff
@@ -10,9 +10,11 @@
     set_quantum_seed,
     set_classical_seed,
     dump_machine,
+    StateDump,
+    ShotResult,
 )
 
-from ._native import Result, Pauli, QSharpError, TargetProfile, StateDump
+from ._native import Result, Pauli, QSharpError, TargetProfile, StateDumpData
```
minestarks, line 17: Is this import used?

minestarks (approved):

pip/qsharp/__init__.py
swernli: Good catch, the added `StateDumpData` is not! That's a holdover from a renaming.

pip/qsharp/_qsharp.py
swernli: Hmm... I tried this in the hopes that it would work, but got some arcane Python errors. Given that it won't save us too much in the implementation, I think we should leave this as-is for now.

PR #1228, opened 2024-03-05
Create Deutsch algorithm kata

tcNickolas:


katas/content/deutsch_algo/implement_algo/solution.md
```diff
@@ -0,0 +1,13 @@
+Follow the algorithm as outlined in the previous section:
+
+1. Allocate a qubit - it starts in the $|0\rangle$ state.
+2. Apply the $H$ gate to the qubit.
+3. Apply the oracle. The syntax for applying the oracle is the same as for applying any other gate or operation.
+4. Apply the $H$ gate to the qubit again.
+5. Measure the qubit: if the measurement result is `Zero`, the function is constant, otherwise it is balanced.
+Since we need to return `true` if the function is constant, we can just return the result of comparing the measurement result with `Zero`.
+
+@[solution]({
```
cesarzc, line 10: Since we have the ability of including multiple code solution, it might be worth adding a solution that uses `within apply` to introduce that Q# construct.

tcNickolas: I'm reluctant to do that here, since we're not really "uncomputing" the computation, and the code would become much longer. We're introducing within-apply in Oracles kata and use it quite a lot there, when we're dealing with proper uncomputation

cesarzc:
If we are not including the section with the images, and given that it's likely the SVGs will change, we should probably remove them before merging this PR.

cesarzc (approved):

PR #1233, opened 2024-03-06
Add `Microsoft.Quantum.Measurement` to compiler prelude

### LINKED ISSUE #1193: Move MResetZ to Intrinsic namespace?
Now that Q# requires resetting qubits to |0> before releasing them, not just measuring them, I find myself using `MResetZ` a lot more often than with the classic QDK, and every time I need to open `Microsoft.Quantum.Measurement` namespace. (There's also no Quick Fix to automate this.)
Moving it to Intrinsic seems logical, since it's so frequently used in the code.

eureka-cpu:
Closes #1193 

As suggested by @swernli, this PR adds `Microsoft.Quantum.Measurement` to the prelude to avoid constant importing.


swernli:
Hi @eureka-cpu , thanks for the contribution! As a way of updating the tests to depend on this change, could you also remove the instances of `open Microsoft.Quantum.Measurement;` from the following files:

- compiler/qsc_codegen/src/qir_base/tests.rs
- compiler/qsc_eval/src/tests.rs
- library/src/tests/arithmetic.rs
- library/src/tests/canon.rs
- library/src/tests/measurement.rs
- library/src/tests/resources/add_le.qs
- library/src/tests/resources/compare.qs
- library/src/tests/resources/inc_by_le.qs
- library/src/tests/resources/select.qs
- library/src/tests/resources/state_preparation.qs
- library/std/internal.qs
- library/std/unstable_table_lookup.qs
- resource_estimator/src/counts/tests.rs

The reasoning for just that subset is that we don't want to remove the `open` directive from files that are part of public samples until this change makes it into a release.

swernli (approved):

PR #1234, opened 2024-03-06
Generate kata tests from index.json

tcNickolas:
Reading the list of katas from index.json and generating a test for each kata automatically seems more scalable and easier to manage than hardcoding the list of katas in two files separately. This approach generates individual tests for each kata that fail separately from each other, so there's no regression in the experience testing them.


cesarzc (approved):
This is awesome, thank you!

PR #1236, opened 2024-03-07
Add logic to detect missing comma token in array parsing

### LINKED ISSUE #1114: Improve error message for arrays with missing commas
An array with missing commas between elements `[1. 0. 0.]` yields a confusing error message: "expected `]`, found float". This looks like a case of unbalanced brackets rather than the actual issue of missing commas. I ran into this when generating a very long Q# code string for df-chemistry sample and spent quite some time looking for a missing closing bracket before I realized the real issue.

An error message saying "expected `]` or `,`" to better point at the possible source of the issue.

**Additional context**
image

eureka-cpu:
Closes #1114 

Adds a conditional arg to add better error handling when array elements are missing a comma. This may fall under the "parse don't validate" category, but it's a start. Welcoming any feedback on how to approach this otherwise, or happy to close if the current behavior is expected.

### DIFF
--- a/compiler/qsc_parse/src/expr.rs
+++ b/compiler/qsc_parse/src/expr.rs
@@ -280,7 +280,18 @@ fn expr_array_core(s: &mut ParserContext) -> Result<Box<ExprKind>> {
     };
 
     if token(s, TokenKind::Comma).is_err() {
-        return Ok(Box::new(ExprKind::Array(vec![first].into_boxed_slice())));
+        match s.peek().kind {
+            TokenKind::Eq | TokenKind::Close(Delim::Bracket) => {
+                return Ok(Box::new(ExprKind::Array(vec![first].into_boxed_slice())));
+            }
+            _missing_comma => {
+                return Err(Error(ErrorKind::Token(
+                    TokenKind::Comma,
+                    s.peek().kind,
+                    s.peek().span,
+                )));
+            }
+        }
     }
 
     let second = expr(s)?;
--- a/compiler/qsc_parse/src/expr/tests.rs
+++ b/compiler/qsc_parse/src/expr/tests.rs
@@ -978,6 +978,26 @@ fn array_pair() {
 }
 
 #[test]
+fn array_of_floats_missing_comma() {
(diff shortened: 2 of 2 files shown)


eureka-cpu:
> Maybe something like this?
> 
> ```rust
> ...
>     #[error("expected {0}, found {1}")]
>     #[diagnostic(code("Qsc.Parse.MultiToken"))]
>     MultiToken(MultiToken<2>, TokenKind, #[label] Span)
> ...
> 
> 
> #[derive(Clone, Copy, Debug, Diagnostic, Eq, Error, PartialEq)]
> /// supports expecting up to N tokens
> pub struct MultiToken<const N: usize> {
>     tokens: [TokenKind; N],
> }
> 
> impl<const N: usize> std::fmt::Display for MultiToken<N> {
>     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
>         use std::fmt::Write;
>         let mut buf = String::new();
>         let len = N;
>         for ix in 0..len {
>             let token = self.tokens[ix];
>             if ix == len - 1 {
>                 write!(buf, "or {}", token)?;
>             } else if ix == 0 {
>                 write!(buf, "{}", token)?;
>             } else {
>                 write!(buf, ", {}", token)?;
>             }
>         }
>         write!(f, "{}", buf)
>     }
> }
> ```

Was trying to find a way to keep it dynamic, but was running into the `Copy` trait constraint. Is there a way to do this that is dynamic? The problem I can see with this impl would be the use cases where there are a different number of expected `TokenKind`s for any one use case, how would we handle those cases? Maybe by making the expected type `&'static str`, then creating a `to_err_msg` method that matches on `self` could work, and that could return a `const`. It's not elegant but it would be somewhat more dynamic than restricting the list of `TokenKind`s to a static number. Just thinking out loud here, feel free to correct me if this is just better off as a static list that gets formatted by the error impl.

eureka-cpu:
> Feel free to reopen if you pick this up again.

Thanks, it's been a rough couple of weeks.

PR #1239, opened 2024-03-07
Update type inference to constrain indexed container as an array

swernli:
This change updates the type inference logic to have an additional constraint on indexed containers that they must be an array of a new inference type. This is enough to allow further inference to determine the type of the array if items from the array are used in places with inferred types. If any other types support the `HasIndex` class in the future this approach can be revisited.

This fixes an issue where certain code with inferrable types would produce type ambiguous errors, such as [this code](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAEz3OMWvDMBAF4F2%2F4oEXe6k6t6R0aMkUiNsOBePhal8TgXVSpBMlhPz3YqX0xu8e785aiwbv5OPCxlqLLas6OSArJeV5teoNXjhPyUV1Qap8HF2GyyB4J87Tgr5BTOGQyEOPpJhI8MUomWdouFXiJ7l6oG8whZnvjJDnHGli7M59IdHi938tF2MA4PlVNJ33wYm2XZUQOdH6CXbkpO3wgDfOZdFhxKUm1llY8Y0NThmbJ3y2pzzcj93j%2Fz6xliQYxhtdzdX8Al5OMIEQAQAA):

```qsharp
let f = qs => X(qs[0);
```

### DIFF
--- a/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/compiler/qsc_frontend/src/typeck/rules.rs
@@ -312,9 +312,18 @@ impl<'a> Context<'a> {
                 )
             }
             ExprKind::Index(container, index) => {
+                let container_span = container.span;
                 let container = self.infer_expr(container);
                 let index = self.infer_expr(index);
                 let item_ty = self.inferrer.fresh_ty(TySource::not_divergent(expr.span));
+                let container_item_ty = self
+                    .inferrer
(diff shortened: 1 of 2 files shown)


minestarks (approved):
LGTM but a real compilerhead should look at it :)

PR #1243, opened 2024-03-07
Add some docstrings and tests for resolution and parsing

sezna:
I accumulated some docs and tests from my hierarchical namespaces branch. I figured I'd pull them out into a separate PR
for easier review. These tests help me modify namespace parsing and semantics, as well as resolver semantics with greater
confidence.

### DIFF
--- a/compiler/qsc_parse/src/item/tests.rs
+++ b/compiler/qsc_parse/src/item/tests.rs
@@ -1654,21 +1654,21 @@ fn namespace_with_attributes_and_docs() {
         "/// Documentation for LevelOne
         namespace LevelOne {
             @ExampleAttr()
-            /// Documentation for InnerItem
+            /// Documentation that shouldn't show up, since docstrings go above attrs
             function InnerItem() : Unit {}
         }",
         &expect![[r#"
-            Namespace _id_ [0-183] (Ident _id_ [49-57] "LevelOne"):
+            Namespace _id_ [0-225] (Ident _id_ [49-57] "LevelOne"):
                 doc:
                     Documentation for LevelOne
-                Item _id_ [72-173]:
+                Item _id_ [72-215]:
                     Attr _id_ [72-86] (Ident _id_ [73-84] "ExampleAttr"):
                         Expr _id_ [84-86]: Unit
-                    Callable _id_ [99-173] (Function):
-                        name: Ident _id_ [152-161] "InnerItem"
-                        input: Pat _id_ [161-163]: Unit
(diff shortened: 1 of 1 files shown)


compiler/qsc_parse/src/item/tests.rs
```diff
@@ -1557,3 +1557,139 @@ fn disallow_qubit_scoped_block() {
...
+}
+
+#[test]
+fn namespace_with_attributes_and_docs() {
+    check_vec(
+        parse_namespaces,
+        "/// Documentation for LevelOne
+        namespace LevelOne {
+            @ExampleAttr()
+            /// Documentation for InnerItem
```
swernli, line 1657: This documentation is actually missing from the HIR below because it is getting treated as a normal comment (the parser recovery for `///` you worked on). It should appear before any attributes to correctly show up.

compiler/qsc_parse/src/item/tests.rs
```diff
@@ -1557,3 +1557,139 @@ fn disallow_qubit_scoped_block() {
...
+}
+
+#[test]
+fn namespace_with_conflicting_names() {
+    check_vec(
+        parse_namespaces,
+        "namespace Conflicts {
+            function Item() : Unit {}
+            newtype Item = Int;
+        }",
```
swernli, line 1682: I would've expected this to produce a compilation error as well... should the test verify that in addition to ensuring recovery knows how to produce HIR?

compiler/qsc_parse/src/prim.rs
```diff
@@ -77,6 +77,8 @@ pub(super) fn ident(s: &mut ParserContext) -> Result<Box<Ident>> {
     }
 }
 
+/// This function takes a [Path] and converts it into a single ident, which
```
swernli, line 80: nitpick: I don't see this as "taking" a path since the path is not an argument, but rather parsing the path from the given stream context.

compiler/qsc_parse/src/item/tests.rs
swernli: Ah, that makes sense then. Maybe to help clarify intent, chance the string to something like "Documentation treated as comment" so it's clear that it's not like the one above.

compiler/qsc_parse/src/item/tests.rs
swernli: Good clarification. Never mind then!

swernli (approved):

PR #1244, opened 2024-03-08
Fix generated type in replace qubit allocation pass

cesarzc:
During the replace qubit allocations pass, calls to qubit allocation operations are generated, but the type of the call expression was always set to `Qubit` regardless of whether the allocation was for a single qubit or multiple qubits.

This change sets the appropriate types for the qubit allocation call expression. `Qubit` for single qubit allocation and `Qubit[]` for multi-qubit allocation.

### DIFF
--- a/compiler/qsc_passes/src/replace_qubit_allocation.rs
+++ b/compiler/qsc_passes/src/replace_qubit_allocation.rs
@@ -480,10 +480,15 @@ fn create_qubit_alloc_call_expr(
     call_expr: Expr,
     array_size: Option<Expr>,
 ) -> Expr {
+    let ty = if array_size.is_none() {
+        Ty::Prim(Prim::Qubit)
+    } else {
+        Ty::Array(Box::new(Ty::Prim(Prim::Qubit)))
+    };
     Expr {
         id: assigner.next_node(),
         span,
-        ty: Ty::Prim(Prim::Qubit),
+        ty,
         kind: ExprKind::Call(
             Box::new(call_expr),
             Box::new(array_size.unwrap_or(Expr {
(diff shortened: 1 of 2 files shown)


swernli (approved):

minestarks:
@cesarzc @swernli @sezna  This kind of bug makes me nervous because it sounds so hard to catch - would it be technically possible to do a typecheck run after we do all the passes? Or do I worry too much?

swernli:
> @cesarzc @swernli @sezna This kind of bug makes me nervous because it sounds so hard to catch - would it be technically possible to do a typecheck run after we do all the passes? Or do I worry too much?

All the typecheck infrastructure runs on AST, not HIR, so it would be tricky to do a rerun that included these modified/generated constructs as those are not in in the AST. We could write something that doesn't have the full inference logic and just verifies types match, but it would still be new code. Worth considering for sure, but not as trivial as just calling what we already have again.

cesarzc:
> > @cesarzc @swernli @sezna This kind of bug makes me nervous because it sounds so hard to catch - would it be technically possible to do a typecheck run after we do all the passes? Or do I worry too much?
> 
> All the typecheck infrastructure runs on AST, not HIR, so it would be tricky to do a rerun that included these modified/generated constructs as those are not in in the AST. We could write something that doesn't have the full inference logic and just verifies types match, but it would still be new code. Worth considering for sure, but not as trivial as just calling what we already have again.

I think it would be worth it even if it's new code and not re-using the typecheck infra that runs on AST. The two times I've encountered this kind of bug, it was hard to catch because it manifested as a side effect in other parts of the code.

Issue #1085, opened 2024-01-29
Circuit visualization


minestarks:
We'd like to add the ability to generate circuit diagrams from a Q# program.

These can appear in:
- VS Code
- Python notebooks
- Python command line (ascii art)

Small programs without classical constructs (e.g. the Bell State sample) are fairly easy to generate diagrams for, but we'll have to figure how best to visualize more complex cases.

Design considerations:
- How to represent classical constructs, such as loops, visually
- Whether the circuit should map to the hardware behavior (i.e. closer to the generated QIR, showing auxiliary qubits, unrolled loops etc), or to the high level input Q# program
- UI to visually map parts of the circuit to source code

For reference, Q# notebooks in the classic QDK had the ability to generate diagrams using the `%trace` magic and the `quantum_viz` Python package.

minestarks:
More 

- [ ] State annotations on the circuit
- [ ] Q# variable names for qubit wire labels instead of q0, q1, etc.
- [ ] When running with simulator, annotate measurement results on circuit
- [ ] Run multiple shots for dynamic circuits and display multiple outcomes in some way

PR #1247, opened 2024-03-08
Circuit generation (internals)

minestarks:
Circuit generation for Q# programs and operations.

This PR covers the native code bits - the implementation of the circuit generation itself. 

I decided to make the UX parts (VS Code and Python) into a separate PR, since the feature is large enough as is.

Implements part of #1085


compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,676 @@
...
+#[derive(PartialEq, Clone, Serialize, Debug)]
+pub struct Qubit {
+    pub id: usize,
+    #[serde(rename = "numChildren")]
+    pub num_children: usize,
+}
+
+#[derive(Clone, Debug, Copy, Default)]
+pub struct Config {
+    // TODO: I don't think it ever makes sense to
```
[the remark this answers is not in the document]
minestarks: Oops, next commit takes care of this.

compiler/qsc_circuit/src/operations.rs
```diff
@@ -0,0 +1,307 @@
...
+    hir::{Item, ItemKind},
+    ty::{Prim, Ty},
+};
+use std::cell::RefCell;
+
+#[derive(Debug)]
+pub struct OperationInfo {
+    pub operation: String,
+    pub total_num_qubits: u32,
+}
```
minestarks, line 14: Need to move this into the language service since it's really only used there.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,673 @@
...
+#[derive(PartialEq, Clone, Serialize, Debug)]
+pub struct Qubit {
+    pub id: usize,
+    #[serde(rename = "numChildren")]
+    pub num_children: usize,
+}
+
+#[derive(Clone, Debug, Copy, Default)]
+pub struct Config {
+    pub no_qubit_reuse: bool,
```
minestarks, line 58: I'm really struggling with the name of this parameter. What it really means is "match base profile code generation". See `remapper.rs` for what that entails -- another struct whose naming I really struggled with.

compiler/qsc/src/interpret/circuit_tests.rs
minestarks, line 1: Note to reviewer: this is one where you'll probably want to read the test cases to understand some gotchas in the behavior.

DmitryVasilevsky: Should there be comments in the source code in this case?

minestarks: Yeah. I tried my best to do so, but some of it is just emergent behavior. Each individual part is already documented in the relevant source code, but when you put them all together, you sometimes get surprising (but expected) results. 

I figured sometimes it's just easier to see the end-to-end input->output demonstrated in a test case to really "grok" what all the implications are.

That being said, this kind of stuff should probably go in user-facing documentation... I'll work on that.

compiler/qsc/src/interpret.rs
```diff
@@ -361,6 +457,18 @@ impl Interpreter {
     }
 }
 
+/// Describes the entry point for circuit generation.
+pub enum CircuitEntryPoint {
+    /// An operation. This must be an callable name or a lambda
```
orpuente-MS, line 462: Typo in docstring
```suggestion
    /// An operation. This must be a callable name or a lambda
```

compiler/qsc/src/interpret.rs
```diff
@@ -292,6 +320,74 @@ impl Interpreter {
...
+                    }
+                    _ => {
+                        return Err(vec![Error::NoCircuitForOperation]);
+                    }
+                };
+
+                if maybe_invoke_expr.is_none() {
+                    return Err(vec![Error::NoCircuitForOperation]);
+                }
+                maybe_invoke_expr
```
orpuente-MS, line 376: `Option<T>` has a method called `ok_or(...)` for this kind of situation. This becomes:

```suggestion
                maybe_invoke_expr.ok_or(vec![Error::NoCircuitForOperation])
```

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,674 @@
...
+    pub targets: Vec<Register>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub children: Vec<Operation>,
+}
+
+#[derive(Serialize, Debug, Eq, Hash, PartialEq, Clone)]
+pub struct Register {
+    #[serde(rename = "qId")]
+    pub q_id: usize,
+    pub r#type: usize, // 0: quantum, 1: classical
```
cesarzc, line 43: It would be nice to use constants to know which value represents a quantum register type and which one represents a classical register (e.g. `const QUANTUM: usize = 0;`, `const CLASSICAL: usize = 0;`). There are a couple of places in the code where the value of `r#type` is checked and I think having these constants will make the intent clearer.

minestarks: Good suggestion, I found this schema somewhat annoying to work with, but I couldn't think of this solution!

compiler/qsc/src/interpret.rs
```diff
@@ -79,6 +84,12 @@ pub enum Error {
     #[error("unsupported runtime capabilities for code generation")]
     #[diagnostic(code("Qsc.Interpret.UnsupportedRuntimeCapabilities"))]
     UnsupportedRuntimeCapabilities,
+    #[error("expression does not evaluate to an operation that takes qubits")]
+    #[diagnostic(code("Qsc.Interpret.NoCircuitForOperation"))]
+    #[diagnostic(help(
+        "provide the name of a callable or a lambda expression that only takes qubits"
```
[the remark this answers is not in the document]
minestarks: Reworded.

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -0,0 +1,804 @@
...
+                    Controlled CNOT(cs, (q2, q1));
+                    CNOT(q1, q2);
+                }
+            }
+        }",
+        Profile::Unrestricted,
+    );
+
+    let circ = interpreter
+        .circuit(CircuitEntryPoint::Operation("Adjoint Test.SWAP".into()))
```
swernli, line 451: I think it would be helpful to test the adjoint of an operation that is not effectively "adjoint self" such that the circuit diagram will look different for the two. You could have a custom operation that has `X` as its body and `Y` as it's adjoint, since technically nothing verifies that explicit adjoints are actually the inverse of their body implementation. That would make it very obvious that the circuit below is for the adjoint!

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -0,0 +1,804 @@
...
+                    CNOT(q1, q2);
+                    Controlled CNOT(cs, (q2, q1));
+                    CNOT(q1, q2);
+                }
+            }
+        }",
+        Profile::Unrestricted,
+    );
+
+    let circ = interpreter
```
cesarzc, line 492: Was this test meat to check circuit generation for `SWAP` (instead of `H`) using a lambda?

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -0,0 +1,804 @@
...
+        .circuit(CircuitEntryPoint::EntryPoint)
+        .expect("circuit generation should succeed");
+
+    // This is one gate that spans ten target wires, even though the
+    // text visualization doesn't convey that clearly.
+    expect![[r"
```
cesarzc, line 282: For this test, I'm not sure how to read the circuit. By looking at the definition of `AccountForEstimatesInternal`, it seems like that operation takes an array of tuples of ints, an integer and a qubit register. However, by looking at the generated circuit, I'm not sure how the arguments match.

compiler/qsc_circuit/src/operations.rs
```diff
@@ -0,0 +1,296 @@
...
+            ({operation_expr})({call_args});
+            let r: Result[] = [];
+            r
+        }}"#
+    )
+}
+
+/// The number of qubits to allocate for each qubit array
+/// in the operation arguments.
+static NUM_QUBITS: u32 = 2;
```
cesarzc, line 107: Would using `const` be more appropriate than using `static` in this case?

language_service/src/code_lens.rs
```diff
@@ -27,43 +28,75 @@ pub(crate) fn get_code_lenses(
...
+                    .parent
+                    .and_then(|parent_id| user_unit.package.items.get(parent_id))
+                    .map(|parent| &parent.kind)
+                {
+                    let namespace = ns.name.to_string();
+                    let range = into_range(position_encoding, decl.span, &user_unit.sources);
+                    let name = decl.name.name.clone();
+
+                    if item.attrs.iter().any(|a| a == &Attr::EntryPoint) {
+                        // If there is more than one entrypoint, not our problem, we'll go ahead
```
cesarzc, line 50: I chuckled with this one 😆

compiler/qsc_circuit/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![warn(clippy::mod_module_files, clippy::pedantic, clippy::unwrap_used)]
+#![allow(clippy::missing_errors_doc, clippy::missing_panics_doc)]
```
swernli, line 5: The pattern for the clippy lints changed since you started working on this, where now instead of adding these to the top of the lib.rs you can just add a workspace lint entry to the Cargo.toml. See https://github.com/microsoft/qsharp/blob/main/compiler/qsc_eval/Cargo.toml#L30-L31 as an example.

compiler/qsc_circuit/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+qsc_frontend = { path = "../qsc_frontend" }
+qsc_fir = { path = "../qsc_fir" }
+qsc_hir = { path = "../qsc_hir" }
+rustc-hash = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+
+[dev-dependencies]
+expect-test = { workspace = true }
+indoc = { workspace = true }
```
swernli, line 28: huh, looks like indoc shows up twice here. If it's used by production code then only the one above should be needed and this can be deleted., if only test then this one can kept and the one above deleted.

compiler/qsc_circuit/src/operations.rs
minestarks: I forgot `const` was a thing! Updating.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub is_measurement: bool,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub controls: Vec<Register>,
+    pub targets: Vec<Register>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub children: Vec<Operation>,
+}
+
+static QUANTUM_REGISTER: usize = 0;
+static CLASSICAL_REGISTER: usize = 1;
```
orpuente-MS, line 40: These two can be const.
```suggestion
const QUANTUM_REGISTER: usize = 0;
const CLASSICAL_REGISTER: usize = 1;
```

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,612 @@
...
+static QUBIT_WIRE: &str = "───────";
+static CLASSICAL_WIRE: &str = "═══════";
+static QUBIT_WIRE_CROSS: &str = "───┼───";
+static CLASSICAL_WIRE_CROSS: &str = "═══╪═══";
+static CLASSICAL_WIRE_START: &str = "   ╘═══";
+static QUBIT_WIRE_DASHED_CROSS: &str = "───┆───";
+static CLASSICAL_WIRE_DASHED_CROSS: &str = "═══┆═══";
+static VERTICAL_DASHED: &str = "   ┆   ";
+static VERTICAL: &str = "   │   ";
+static BLANK: &str = "       ";
```
orpuente-MS, line 205: All these can also be `const`. I think the difference is that if you mark it at `const` they get evaluated at compile time. So maybe you get some extra fancy optimizations from `rustc`.
```suggestion
const COLUMN_WIDTH: usize = 7;
const QUBIT_WIRE: &str = "───────";
const CLASSICAL_WIRE: &str = "═══════";
const QUBIT_WIRE_CROSS: &str = "───┼───";
const CLASSICAL_WIRE_CROSS: &str = "═══╪═══";
const CLASSICAL_WIRE_START: &str = "   ╘═══";
const QUBIT_WIRE_DASHED_CROSS: &str = "───┆───";
const CLASSICAL_WIRE_DASHED_CROSS: &str = "═══┆═══";
const VERTICAL_DASHED: &str = "   ┆   ";
const VERTICAL: &str = "   │   ";
const BLANK: &str = "       ";
```

orpuente-MS (commented):
Looks good!

compiler/qsc_circuit/src/circuit.rs
minestarks: Makes sense, thanks!

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -0,0 +1,804 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
minestarks: Thanks for the suggestions. For the second one, I think maybe I'll do a classical for loop just to make it a bit more interesting and to demonstrate how classical control flow impacts circuits.

compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Good point. I wanted a realistic example but didn't know enough to come up with one. I think I'll just have a very simple and obvious one instead.

compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Yeah, I didn't put a lot effort into custom intrinsics, so the formatting is very unclear here. 

What this supposed to show is all the runtime values of all the classical arguments. Qubit arguments are just shown as wires that go into the gate.

I think this one is extra challenging to read, because the gate spans across 10 qubits and qubit wires aren't even visible (the gate label is too wide for the "column" width I chose). I didn't want to sink too much time into rendering ASCII art nicely, when I expect the SVG image to be the main scenario.

I think I can spend a little more time making the classical arguments look better though. At the moment I'm not grouping tuples or arrays, and a flat list is really hard to read.

cesarzc: Thanks for the explanation!

FWIW, I don't think making the classical arguments look better when rendering ASCII is something that needs to be done in this PR. I was just trying to understand the test case, so if you want to make the classical arguments look better in another change, I don't have any problem with that.

compiler/qsc_circuit/src/operations.rs
```diff
@@ -0,0 +1,296 @@
...
+fn get_array_dimension(input: &Ty) -> Option<u32> {
+    match input {
+        Ty::Prim(Prim::Qubit) => Some(0),
+        Ty::Array(ty) => get_array_dimension(ty).map(|d| d + 1),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
swernli, line 150: nitpick: we've been following the pattern in the repo where tests are not at the bottom of the file but instead in a separate file, so the `mod tests;` would be at the top of this file and these would be moved to operations/tests.rs.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -0,0 +1,612 @@
...
+        for row in rows {
+            row.fmt(f, end_column)?;
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
swernli, line 345: nitpick: same here, the tests should be moved to circuit/tests.rs to match the repo patterns.

cesarzc (approved):

compiler/qsc/src/interpret/circuit_tests.rs
minestarks: I ended up changing the formatting a little bit anyway. The classical args (the `[(0,1), (1,2)...` part) get displayed in the SVG circuit as well... so this should help there.

I won't worry about the ASCII art wires for now.

compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Uh oh, copy paste error. It is actually supposed to test the lambda, but the SWAP code got left in there from a different test I copied. Thanks!

billti (approved):

PR #1249, opened 2024-03-10
Native libraries use mimalloc as global allocator

idavis:
This PR moves the global allocator to MS's [mimalloc](https://github.com/microsoft/mimalloc). The performance gains are significant, and it adds `98KB` to the Python wheel and `215KB` to the native dylib used in the python wheel. This dependency adds `2.5s` to the build time.

We can investigate wasm usage in the future, but this would likely require having a wasi-libc installation available for the build process to give mimalloc a wasi compatible libc to link against.

Benchmarks on Apple M1 Max 64GB
```
Teleport evaluation     time:   [44.024 µs 44.043 µs 44.063 µs]
                        change: [-21.789% -21.595% -21.420%] (p = 0.00 < 0.05)
                        Performance has improved.

Deutsch-Jozsa evaluation
                        time:   [3.0302 ms 3.0336 ms 3.0375 ms]
                        change: [-23.878% -23.671% -23.493%] (p = 0.00 < 0.05)
                        Performance has improved.

Large file parity evaluation
                        time:   [28.956 ms 28.972 ms 28.989 ms]
                        change: [-1.9178% -1.8288% -1.7431%] (p = 0.00 < 0.05)
                        Performance has improved.

Array append evaluation time:   [306.58 µs 306.78 µs 307.02 µs]
                        change: [-14.312% -14.068% -13.881%] (p = 0.00 < 0.05)
                        Performance has improved.

Array update evaluation time:   [304.71 µs 304.84 µs 305.00 µs]
                        change: [-10.777% -10.651% -10.537%] (p = 0.00 < 0.05)
                        Performance has improved.

Array literal evaluation
                        time:   [144.03 µs 144.32 µs 144.68 µs]
                        change: [-39.800% -39.455% -39.171%] (p = 0.00 < 0.05)
                        Performance has improved.

Large nested iteration  time:   [30.418 ms 30.446 ms 30.476 ms]
                        change: [-11.368% -11.252% -11.131%] (p = 0.00 < 0.05)
                        Performance has improved.

Large input file        time:   [15.869 ms 15.883 ms 15.900 ms]
                        change: [-30.391% -30.267% -30.147%] (p = 0.00 < 0.05)
                        Performance has improved.

Standard library        time:   [9.5701 ms 9.5801 ms 9.5919 ms]
                        change: [-28.527% -28.371% -28.224%] (p = 0.00 < 0.05)
                        Performance has improved.
```


swernli (approved):
I love it!

billti:
Does this mean we have a dependency on a C compiler in our build now also if users want to set up a Q# development environment? (Which I guess we kind of implicitly do anyway as Rust on Windows requires MSVC or C++ Build Tools are installed, Linux generally has a C compiler already... but I'm not sure if that's already a requirement on macOS).

idavis:
> Does this mean we have a dependency on a C compiler in our build now also if users want to set up a Q# development environment? (Which I guess we kind of implicitly do anyway as Rust on Windows requires MSVC or C++ Build Tools are installed, Linux generally has a C compiler already... but I'm not sure if that's already a requirement on macOS).

Yes, you need to have a C compiler handy. I removed the dep on Ninja so that you don't need the generator.

idavis:
Build warnings on macos when xcode isn't installed will be fixed by https://github.com/rust-lang/cc-rs/pull/1007

build.py
```diff
@@ -403,6 +403,7 @@ def use_python_env(folder):
         "-m",
         "pip",
         "wheel",
+        "--verbose",
```
billti, line 406: Why do we want the JupyterLab pip package to be verbose as part of this?

Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [workspace]
 members = [
+    "allocator",
+    "allocator/mimalloc-sys",
```
billti, line 4: Why do we need two different crates rather than just the one?

allocator/mimalloc-sys/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+project(allocator_external)
+include(ExternalProject)
+
+ExternalProject_Add(mimalloc
+  GIT_REPOSITORY    https://github.com/microsoft/mimalloc.git
+  GIT_TAG           $ENV{ALLOCATOR_MIMALLOC_TAG}
+  GIT_SHALLOW       TRUE
+  GIT_PROGRESS      TRUE
+  CONFIGURE_COMMAND ""
```
billti, line 15: My CMake is a little lacking, but why do all these commands just have an empty string? If they're empty, why do you need the 'USES_TERMINAL*' settings on them? (maybe some explanatory comments would be useful).

billti (approved):

allocator/mimalloc-sys/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version.workspace = true
+
+[dependencies]
+
+[lints]
+workspace = true
+
+[build-dependencies]
+cmake = "0.1"
+cc = "1.0"
```
minestarks, line 19: Is it possible to put all of this behind a feature flag? The new dependency is a bit "ouch". I love the idea of omitting it for local dev builds but using it in CI & release builds.

billti: If it adds 2.5 seconds to build time, I'd rather we dogfood it and it be in dev builds by default, rather than all our local testing NOT match what we build/ship and then being surprised by something when we do.

README.md
```diff
@@ -23,12 +23,13 @@ Code from this repository powers the Q# development experience on <https://quant
...
 ## Building
 
-To build this repository there are 4 dependencies that need to be installed. These are:
+To build this repository there are dependencies that need to be installed. These are:
 
 - Python (<https://python.org>)
 - Rust (<https://www.rust-lang.org/tools/install>)
 - Node.js (<https://nodejs.org/>)
 - wasm-pack (<https://rustwasm.github.io/wasm-pack/installer/>)
+- cmake (<https://cmake.org/>) and a C compiler
```
minestarks, line 32: @idavis if we do keep the C dependency for dev builds, could you add it to the devcontainer.json (if the container doesn't already have it)? I frequently use Codespaces for development.

minestarks, line 32: Could you also update `prereqs.py` to account for the new dependency?

minestarks: And perhaps a more precise link to on how to install the "C compiler" for various platforms (mac, ubuntu, windows). Will any compiler do or do we need a specific one in the toolchain?

(I know I sound like I'm nitpicking here, but it's been about 15 years since I've manually installed a C compiler 😆 .  This would definitely stop me in my tracks if I was new to contributing to this repo, or setting up a fresh environment after having been away for a while.)

allocator/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
minestarks, line 1: I searched for `mimalloc` on crates.io... would it simplify things at all if we just took a dependency on https://crates.io/crates/mimalloc  (or one of the many alternatives)?

README.md
billti: Ian and I discussed this. We'll update the docs. Windows is actually the easy one (which I think you use), as to get Rust working you must have already installed the C++ compiler (either with VS or the Build Tools - see https://learn.microsoft.com/en-us/windows/dev-environment/rust/setup ).

allocator/mimalloc-sys/Cargo.toml
minestarks: Either way, a feature flag approach would at least clean up the checks that look like `[target.'cfg(not(any(target_family = "wasm")))'.dependencies]` . (It could be on by default but wasm can opt out).

allocator/mimalloc-sys/build.rs
```diff
@@ -0,0 +1,84 @@
...
+    build.compile("mimalloc");
+
+    println!(
+        "cargo:rustc-link-search=native={}",
+        dst.join("lib").display()
+    );
+    println!("cargo:rustc-link-lib=static=mimalloc");
+}
+
+// Use cmake to download mimallloc source code and stage
```
minestarks, line 63: Looks like you allocated an extra l here :P

minestarks (commented):
I'm frankly on the fence about taking this. The perf gains are undeniable, but I have several reservations about the change overall.

1. WASM is one our two main use cases and this change wouldn't improve it.
2. I'd rather reassess impact after we look into optimizations that reduce allocations across the board, such as #1261 . I still think we have a lot of low hanging fruit here.

If we do decide it's worth it, these would make the change more palatable for me:

1. An easy way to opt out of the CMake/C dependency for local builds. We have external contributors and we love them! One more prereq before you can build _at all_ can be a barrier to engagement.
2. Can we use an external implementation via crates.io? This code is a perfect candidate for just using OSS, it's a very general implementation, and exactly the kind of code I wouldn't want to be stuck maintaining.

allocator/Cargo.toml
idavis: Great call out. We need to customize our compilation of mimalloc, otherwise I would have used it as you suggested.

allocator/mimalloc-sys/Cargo.toml
idavis: @billti @minestarks There is the question of whether we have this set in all libs. Right now, it is only set on distribution units, so any tests in those projects will use mimalloc. Any low level libs, for their individual crate tests, would still be using the default allocator. We can change this, but we have the same issue with WASM. Don't run all of the tests compiled to wasm, just the integration tests.

minestarks (approved):
Thanks for addressing the concerns, excited about the perf improvement!

PR #1256, opened 2024-03-12
Make Base Profile result type check specific to entry expressions

### LINKED ISSUE #1246: "non-Result return type in entry expression" when executing Q# expressions via Python in Base Profile
When target it set to Base profile, we run the return type check on _all_ cells:

In Python:

```python
import qsharp
qsharp.init(target_profile=qsharp.TargetProfile.Base)
qsharp.eval("123")
```

```
QSharpError: Qsc.BaseProfCk.ReturnNonResult

  × non-Result return type in entry expression
   ╭─[line_0:1:1]
 1 │ 123
   · ───
   ╰────
  help: returning types other than Result from the entry expression is not
        supported when performing base profile QIR generation
```

The base profile compliance check is overly restrictive - I'm not generating code or even evaluating an entry expression.

minestarks:
Fixes #1246

The current base profile result type check was overly broad, because it ran the check on all fragments (not just entry expressions). 

This was because during incremental compilation, entry expressions were added to the AST package as a top-level statement. After that point, the passes could not differentiate between the entry expression and a top-level statement.

We don't _actually_ have to convert the entry expression to a statement - we can just keep the entry expression in the `entry` field in the AST and HIR. So, now, if passes want to apply any special treatment to the entry expression, they have enough information to do so.

### DIFF
--- a/compiler/qsc_eval/src/lower.rs
+++ b/compiler/qsc_eval/src/lower.rs
@@ -81,6 +81,8 @@ impl Lowerer {
         fir_package: &mut fir::Package,
         hir_package: &hir::Package,
     ) -> Vec<StmtId> {
+        let new_entry = hir_package.entry.as_ref().map(|e| self.lower_expr(e));
+
         let items: IndexMap<LocalItemId, fir::Item> = hir_package
             .items
             .values()
@@ -99,6 +101,8 @@ impl Lowerer {
         for (k, v) in items {
             fir_package.items.insert(k, v);
         }
+
+        fir_package.entry = new_entry;
 
         qsc_fir::validate::validate(fir_package);
 
--- a/compiler/qsc_eval/src/lower.rs
+++ b/compiler/qsc_eval/src/lower.rs
@@ -81,8 +81,6 @@ impl Lowerer {
         fir_package: &mut fir::Package,
         hir_package: &hir::Package,
     ) -> Vec<StmtId> {
-        let new_entry = hir_package.entry.as_ref().map(|e| self.lower_expr(e));
-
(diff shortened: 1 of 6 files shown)


compiler/qsc_eval/src/lower.rs
```diff
@@ -81,6 +81,8 @@ impl Lowerer {
         fir_package: &mut fir::Package,
         hir_package: &hir::Package,
     ) -> Vec<StmtId> {
+        let new_entry = hir_package.entry.as_ref().map(|e| self.lower_expr(e));
```
swernli, line 84: Since @cesarzc is taking a dependency on FIR for RCA and has refactored lowering some, it might be worth verifying with him that this change to how these expressions are tracked won't invalidate any assumptions or logic there.

swernli (approved):
Looks good to me, just one note to sync up with Cesar on the change.

compiler/qsc_eval/src/lower.rs
cesarzc: This change does not invalidate assumptions in RCA. The only adjustment I have to make is to include an entry expression in the analysis data structure to save the properties of the entry expression.

cesarzc: On a related note, @minestarks does it matter the order in which things are lowered? I would expect to first lower items, then top-level statements and then the entry expression since it can depend on items and top-level statements.

minestarks: Good question. I don't think it matters in practice, because I can see some existing unit tests that would have been broken if it mattered :) But to keep the order more logical, I'll follow your suggestion and do the entry expression last.

cesarzc (approved):

PR #1257, opened 2024-03-13
Fix required logical qubit error rate for multi-qubit patches

msoeken:
The required logical error rate is not correctly computed for logical patches which have more than one qubit. This PR fixes this. The problem had no impact on our system implementation for resource estimation, since it considers quantum error correction that encodes one logical qubit.


ivanbasov (approved):

swernli:
Oops, sorry, clicked on the wrong PR!

PR #1261, opened 2024-03-13
Optimize evaluator perf via Control Flow Graph

swernli:
This change adds new content to FIR that captures the control flow graph for expressions. With this pre-calculated during lowering, the evaluator is refactored to follow that sequencing rather than calculating (and recalculating) the execution sequence on the fly.

In addition to the performance numbers from the CI benchmarks (which show larger gains in the Linux container), here is some data gathered from runs on my machine. This change tested in WASM shows overall performance benefit but <s>with a trade off of reduced performance on large array literals</s> with the latest change also improving performance on array literals:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
</head>

<body link="#467886" vlink="#96607D">

WASI (time in µs) |   |   |  
-- | -- | -- | --
Benchmark | Main | PR | Perf Impact
Teleport eval | 162.55 | 129.86 | -20.11%
DJ eval | 9835.53 | 7910.8 | -19.57%
Large File eval | 37323 | 35022 | -6.17%
Array Append | 1835 | 1265.6 | -31.03%
Array Update | 2089.2 | 1556 | -25.52%
Array Literal | 694.17 | 377.02 | -45.69%
Large Nested | 195790 | 135720 | -30.68%

</body>

</html>

The performance gains in native code are more significant for those benchmarks that do large numbers of iterations through code. Both the "Main" and "PR" below include the mimalloc optimization that has already merged:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
</head>

<body link="#467886" vlink="#96607D">

Native (time in µs) |   |   |  
-- | -- | -- | --
Benchmark | Main | PR | Perf Impact
Teleport eval | 39.034 | 32.269 | -17.33%
DJ eval | 2650.8 | 2220.6 | -16.23%
Large File eval | 25327 | 23411 | -7.57%
Array Append | 255.74 | 152.52 | -40.36%
Array Update | 254.13 | 170.67 | -32.84%
Array Literal | 117.93 | 79.975 | -32.18%
Large Nested | 25544 | 14966 | -41.41%

</body>

</html>

Finally, the performance gains for these combined changes on resource estimation workloads show promise in the larger cases, specifically:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/swernli/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
</head>

<body link="#467886" vlink="#96607D">

Resource Estimation   (time in sec) |   |   |  
-- | -- | -- | --
Workload | v1.2 | PR | Perf Impact
dfchem XVIII-cas4-fb-64e-56o | 45 | 30 | -33.33%
dfchem nitrogenase-54e-54o | 123 | 80 | -34.96%
Windowed multiplication* | 548 | 333 | -39.23%

</body>

</html>

*from https://arxiv.org/abs/2402.01891


swernli:
> ## Benchmark for [482ca5f](https://github.com/microsoft/qsharp/commit/482ca5fcadcf886a4cd5eb5b1694ca76979640ed)
> Click to view benchmark
> Test	Base	PR	%
> Array append evaluation	756.0±7.50µs	**359.4±4.36µs**	**-52.46%**
> Array literal evaluation	**507.2±20.92µs**	538.2±21.06µs	**+6.11%**
> Array update evaluation	759.2±36.53µs	**438.1±3.06µs**	**-42.29%**
> Deutsch-Jozsa evaluation	7.0±0.06ms	**5.8±0.06ms**	**-17.14%**
> Large file parity evaluation	35.0±0.15ms	**33.8±0.08ms**	**-3.43%**
> Large input file	41.4±1.25ms	41.3±1.16ms	-0.24%
> Large nested iteration	74.5±0.41ms	**35.6±1.63ms**	**-52.21%**
> Standard library	23.3±0.80ms	23.5±0.78ms	+0.86%
> Teleport evaluation	95.4±1.78µs	**76.7±1.33µs**	**-19.60%**

It's interesting to see that the CI benchmark benefits much more from this change (even without mimalloc yet) than my local runs, which gives me hope that the combined benefit of both changes will be even higher for older or slower machines than my local one.

allocator/Cargo.toml
minestarks, line 1: Are these cargo.toml changes meant to be in this PR?

swernli: Yes, because without them running local benchmarks using `cargo wasi bench` was broken. I believe this better matches the intent of the macro, which is to make it available everywhere but become a no-op when compiling for WASM. Without this change, compiling `qsc` for WASM would fail due to not finding the allocator dependency.

compiler/qsc_eval/src/val.rs
```diff
@@ -16,18 +16,32 @@ pub enum Value {
...
     Bool(bool),
-    Closure(Rc<[Value]>, StoreItemId, FunctorApp),
+    Closure(Box<Closure>),
     Double(f64),
     Global(StoreItemId, FunctorApp),
     Int(i64),
     Pauli(Pauli),
     Qubit(Qubit),
-    Range(Option<i64>, i64, Option<i64>),
+    Range(Box<Range>),
```
minestarks, line 25: Why did these have to become boxed?

compiler/qsc_eval/src/lib.rs
```diff
@@ -334,9 +323,13 @@ impl Env {
     }
 
     fn leave_scope(&mut self) {
-        self.0
-            .pop()
-            .expect("scope should be entered first before leaving");
+        // Only pop the scope if there is more than one scope in the stack,
+        // because the global/top-level scope cannot be exited.
```
minestarks, line 327: How could this happen? Are the pushes and pops not balanced?

compiler/qsc_eval/src/val.rs
swernli: That was another performance enhancement that was in the branch I was working on, and I didn't want to tease it out. These two tuples were each larger than a pointer, so boxing them makes the whole structure smaller and improves performance. This is one of the recommendations in the Rust Performance Book as a possible optimization: https://nnethercote.github.io/perf-book/heap-allocations.html#box

compiler/qsc_eval/src/lib.rs
swernli: The "global" scope used by the interpreter when running top level statements is not really a frame and is never actually pushed. So top-level statements are a kind of weird exception that this logic handles.

compiler/qsc_eval/src/lib.rs
minestarks, line 1: It'd be nice to have a header comment explaining how the evaluator works, at least the basic approach taken.

minestarks (approved):
Of course I couldn't follow every single line but changes look reasonable and it seems pretty thoroughly covered by tests. Thanks for the perf gains!

compiler/qsc_eval/src/lib.rs
minestarks: What if we just push a scope in the very beginning?

compiler/qsc_eval/src/lib.rs
swernli: That's a great idea, I'll work on adding one!

compiler/qsc_eval/src/lib.rs
swernli: We sort of do... `Env::default` implemented down on line 376-381 (side note: why is it all the way down there? I'm going to move it up to make it easier to find...) always creates a new environment with an intial global scope. That seemed easier than forcing callers to always have to use `let mut env = Env::default;` followed by `env.push_scope(0);`. But more importantly, because a user can write Q# `return` as a top-level statement, we need to make sure that doesn't accidentally trigger a scope pop that causes all the global variables to dissapper. So scope 0 really is a special, permanent scope that can never be left.

compiler/qsc_eval/src/lib.rs
```diff
@@ -381,51 +394,12 @@ struct Scope {
...
-}
-
 pub struct State {
-    cont_stack: Vec<Cont>,
-    action_stack: Vec<Action>,
-    vals: Vec<Value>,
+    exec_graph_stack: Vec<Rc<[ExecGraphNode]>>,
+    idx: u32,
+    idx_stack: Vec<u32>,
+    curr_val: Option<Value>,
```
cesarzc, line 401: From our conversation about this PR yesterday, this seems like what we referred to as the "result register", is this correct?

If that's the case, could we consider renaming this to `val_register`? It's a small change, but I think it would help map the way we talk about the evaluator with the code itself.

compiler/qsc_eval/src/lib.rs
```diff
@@ -594,6 +561,52 @@ impl State {
...
+                    self.leave_frame();
+                    env.leave_scope();
+                    continue;
+                }
+                None => {
+                    // We have reached the end of the current graph without reaching an explicit return node,
+                    // usually indicating the partial execution of a single sub-expression.
+                    // This means we should pop the execution graph but not the current environment scope,
+                    // so bound variables are still accessible after completion.
+                    self.exec_graph_stack.pop();
```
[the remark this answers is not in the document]
cesarzc: Should we add something like this to verify that we have effectively reached the end of the current graph?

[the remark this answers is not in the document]
swernli: I don't think this would work, because the graph is read-only, so traversing it doesn't shorten it. When we pop it, it will still have all the contents it had before.

compiler/qsc_eval/src/lib.rs
swernli: Yeah, I think that's reasonable. I can change the variable name and the accessor methods too, just make sure they are consistent.

compiler/qsc_eval/src/lib.rs
```diff
@@ -607,346 +620,157 @@ impl State {
...
+            ExprKind::ArrayLit(arr) => self.eval_arr_lit(arr, globals),
+            ExprKind::ArrayRepeat(..) => self.eval_arr_repeat(expr.span)?,
+            ExprKind::Assign(lhs, _) => self.eval_assign(env, globals, *lhs)?,
+            ExprKind::AssignOp(op, lhs, rhs) => {
+                let rhs_span = globals.get_expr((self.package, *rhs).into()).span;
+                let (is_array, is_unique) =
+                    is_updatable_in_place(env, globals.get_expr((self.package, *lhs).into()));
+                if is_array {
+                    if is_unique {
+                        self.eval_array_append_in_place(env, globals, *lhs)?;
```
cesarzc, line 652: As a way to make the intent even clearer, would adding something like `assert!(mathches!(op, BinOp::Add))` make sense?

compiler/qsc_eval/src/lib.rs
```diff
@@ -607,346 +620,157 @@ impl State {
...
+            ExprKind::AssignField(record, field, _) => {
+                self.eval_update_field(field.clone());
+                self.eval_assign(env, globals, *record)?;
             }
-            ExprKind::BinOp(op, lhs, rhs) => self.cont_binop(globals, *op, *rhs, *lhs),
-            ExprKind::Block(block) => self.push_block(env, globals, *block),
+            ExprKind::AssignIndex(lhs, mid, _) => {
+                let mid_span = globals.get_expr((self.package, *mid).into()).span;
+                let (_, is_unique) =
+                    is_updatable_in_place(env, globals.get_expr((self.package, *lhs).into()));
```
cesarzc, line 670: Along the same lines than my previous comment, would it be worth not discarding the `is_array` value and adding something like `assert!(is_array)` to make the assumptions more evident?

compiler/qsc_eval/src/lib.rs
swernli: We could, just as a double check. Because of type checking, there is no way to have the target of the operation be of type array without the operation being an Add. The assert would just validate this assumption from type restrictions.

compiler/qsc_eval/src/lib.rs
swernli: This one I don't think is necessary, because `AssignIndex` is an expression that can only ever make sense for arrays.

swernli: It's worth noting that `assert!` is not free. Unlike other languages where asserts are only compiled out by default, in Rust this check will still run in release unless you use the `debug_assert!` macro instead. We've avoid debug asserts so far, though I can't remember the motivation for why, perhaps just that we don't get much protection from them since we don't regularly run in debug mode even in tests.

compiler/qsc_fir/src/fir.rs
```diff
@@ -1040,6 +1076,7 @@ impl Display for ExprKind {
         let mut indent = set_indentation(indented(f), 0);
         match self {
             ExprKind::Array(exprs) => display_array(indent, exprs)?,
+            ExprKind::ArrayLit(exprs) => display_array(indent, exprs)?,
```
cesarzc, line 1079:
```suggestion
            ExprKind::Array(exprs) | ExprKind::ArrayLit(exprs) => display_array(indent, exprs)?,
```

cesarzc (approved):
Looks good!

Thanks for walking me through the changes yesterday, it was really helpful and made reviewing this PR much easier.

compiler/qsc_fir/src/fir.rs
swernli: Good catch!

PR #1262, opened 2024-03-13
Send OpenDocument telemetry event upon first opening of document

### LINKED ISSUE #1253: `OpenedDocument` telemetry event doesn't get sent for the first Q# document
Prep:

1. Open VSCode
2. Set log level to "debug"
3. Close VSCode

Repro steps:

4. Open VSCode
5. Open a Q# file for the first time
6. Check Q# logs in the Output window

Expected:

* `OpenedDocument` telemetry event should be in the logs

Actual:

* No `OpenedDocument` event
* Opening _another_ Q# file does cause the event to get sent

sezna:
Closes #1253

We register subscriptions on first doc open, but that means we skip the telemetry event on first doc open, since the subscriptions aren't active yet.

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -103,6 +103,21 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
     updateIfQsharpDocument(document);
   });
 
+  // we manually send an OpenDocument telemetry event if this is a Q# document, because the
+  // below subscriptions won't fire for documents that are already open when the extension is activated
+  vscode.workspace.textDocuments.forEach((document) => {
+    if (isQsharpDocument(document)) {
+      const documentType = isQsharpNotebookCell(document)
+        ? QsharpDocumentType.JupyterCell
+        : QsharpDocumentType.Qsharp;
+      sendTelemetryEvent(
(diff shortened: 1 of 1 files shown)


vscode/src/extension.ts
```diff
@@ -103,6 +103,21 @@ function registerDocumentUpdateHandlers(languageService: ILanguageService) {
     updateIfQsharpDocument(document);
   });
 
+  // we manually send an OpenDocument telemetry event if this is a Q# document, because the
+  // below subscriptions won't fire for documents that are already open when the extension is activated
+  vscode.workspace.textDocuments.forEach((document) => {
+    if (isQsharpDocument(document)) {
+      const documentType = isQsharpNotebookCell(document)
```
billti, line 110: So if I open a notebook with a couple dozen Q# cells, is this going to send a couple dozen telemetry events? Or do we consider opened a notebook one document?

billti (approved):

vscode/src/extension.ts
minestarks: This will get fired for each individual Q# cell. 

That being said I personally am not concerned about over-firing here.

PR #1263, opened 2024-03-13
Don't install all the browsers for integration tests

minestarks:
FYI @swernli


billti:
I know VS Code itself is based off Chromium, but do/can the tests at least give us some coverage for browser functionality on vscode.dev (seeing as it loads the same extension)? Might be nice to also run integration tests on Safari/WebKit if that is the case.

billti (approved):

minestarks:
@billti Yes, that could be valuable. I ultimately decided not to, thinking there's not much that our integration tests touch, that could be browser-specific. Maybe the web worker internals?

Maybe if we add integration tests for the webview stuff, then that would be a good time to add cross-browser coverage.

But for now I think it's ok to remove this unnecessary step.

PR #1264, opened 2024-03-14
Local variable identifiers are incorrectly overridden in global scope used for fragments

### LINKED ISSUE #1250: Local variable identifiers are incorrectly overridden in global scope used for fragments
When assigning sequential local variable identifiers to optimize variable access (see #1167), the assigner is reset after each callable declaration. Unfortunately, that will reset the ids for the global scope used by Python `eval` or Jupyter cells, causing variables to become unbound:
image
The interpreter should keep a separate global assigner to ensure this doesn't happen.

swernli:
This fixes the logic on resetting local variable IDs to not lose global scope context, which ensures that variables defined at the global scope before an item defined in the global scope is still valid later in interpreter usage.

Fixes #1250


cesarzc (approved):

minestarks (approved):
I verified this works though I can't explain exactly why.

PR #1267, opened 2024-03-15
Circuit API for Python

minestarks:
- Add `qsharp.dump_circuit()` and `qsharp.circuit()` to Python library
- Add circuit widget with SVG circuit generated by `quantum-viz.js`
- Add sample notebook

JupyterLab light theme:
image

JupyterLab dark theme:
image

VS Code default light theme:
image

VS Code default dark theme:
image


widgets/js/index.tsx
```diff
@@ -210,3 +214,8 @@ function renderHistogram({ model, el }: RenderArgs) {
   model.on("change:buckets", onChange);
   model.on("change:shot_count", onChange);
 }
+
+function renderCircuit({ model, el }: RenderArgs) {
+  const circuitJson = model.get("circuit_json") as string;
+  prender(<Circuit circuit={JSON.parse(circuitJson)}></Circuit>, el);
```
billti, line 220: If you follow the above pattern with rendering in `onChange` and an event handler, then you get updates to the rendering for free (i.e. if the user holds a reference to the widget and sets a new circuit trace on it, it will update the circuit drawing in place)

npm/ux/qsharp-circuit.css
```diff
@@ -0,0 +1,180 @@
...
+
+/*
+  Styles for Q# Circuits.
+  Copied and modified from the default CSS generated by
+  https://github.com/microsoft/quantum-viz.js/blob/288e0cb506ee866ecca1b414c522e2d634f7b36d/src/styles.ts
+*/
+
+/* Background color for the circuit diagram */
+.qs-circuit {
+  background: var(--vscode-editor-background, #ffffff);
```
billti, line 12: You'll see in qsharp-ux.css variables that it accounts not only for the VS Code theme, but also for the Jupyter Labs theme. Have you tried this in Jupyter with different themes? Any reason you're not leveraging the variables in qsharp-ux.css that already account for these?

minestarks: No, no good reason, I just really struggled with this part. I see the "common" variables in qsharp-ux.css now, but they don't cover everything. e.g. some objects in the histogram/estimates use `--vscode-button-background` , but their fallbacks aren't consistent. 

The fallbacks here are just the default colors that are declared in `quantum-viz.js`, so they look okay but they of course don't respect JupyterLab's dark theme.

I just checked Histogram and Estimates, and I think Histogram also doesn't respect the theme - so I'll try to follow Estimates as an example.

widgets/js/index.tsx
minestarks: Makes sense, I'll add the handler. It wasn't clear to me why we'd _want_ that kind of dynamic behavior, but I think I just don't have much experience with widgets.

npm/ux/qsharp-circuit.css
minestarks: I used whatever common variables I could, but a few things (gates) just didn't have clear counterparts in the other widgets. I verified with JupyterLab light / dark themes and a bunch of custom VS Code themes.

billti (approved):

cesarzc (approved):

swernli (approved):

minestarks:
Related #1085

PR #1269, opened 2024-03-18
Circuit panel in VS Code

minestarks:
Circuit code lens and webview showing rendered circuit in VS Code.

Debugging not included - that will be in the next PR.


vscode/src/webview/webview.tsx
```diff
@@ -155,6 +178,15 @@ function App({ state }: { state: State }) {
...
         />
       );
+    case "circuit":
+      return (
+        <CircuitPanel
+          title={state.title}
+          subtitle={state.subtitle}
+          circuit={state.circuit}
+          errorHtml={state.errorHtml}
+        ></CircuitPanel>
```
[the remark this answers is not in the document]
minestarks: Yes, I was caught off guard by this too, but it's been like this for a while. Apparently a pedant has declared

> self-closing tags do not exist in HTML.

https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags

I don't know the reason for this.

vscode/src/codeLens.ts
```diff
@@ -51,14 +60,25 @@ function mapCodeLens(cl: ICodeLens): vscode.CodeLens {
       command = "qsharp-vscode.runEditorContents";
       tooltip = "Run program";
       break;
-    default:
-      throw new Error(`Unknown code lens command: ${cl.command}`);
+    case "circuit":
+      if (
+        !vscode.workspace.getConfiguration("Q#").get<boolean>("featureCircuits")
```
billti, line 65: Is there any UX for this settings (by declaring it in the package.json)? Or do we want/expect users to manually put this in the JSON file to enable it?

vscode/src/webview/webview.tsx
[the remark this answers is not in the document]
billti: They are valid in (P)react. For example, see the HelpPage immediately following.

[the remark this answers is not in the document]
billti: (Or EstimatesPanel above, which also has attributes)

npm/ux/circuit.tsx
```diff
@@ -18,3 +18,27 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
...
+  circuit?: qviz.Circuit;
+  errorHtml?: string;
+}) {
+  return (
+    <div>
+      <div>
+        <h1>{props.title}</h1>
+        <h2>{props.subtitle}</h2>
+      </div>
+      {props.circuit ? <Circuit circuit={props.circuit}></Circuit> : <></>}
```
billti, line 34: Typically you just return `null` (rather than `<></>`) if you don't want content in the alternate case.

vscode/src/circuit.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!editor || !isQsharpDocument(editor.document)) {
+    throw new Error("The currently active window is not a Q# file");
+  }
+
+  sendMessageToPanel("circuit", true, undefined);
+
+  // Start the worker, run the code, and send the results to the webview
+  const worker = getCompilerWorker(compilerWorkerScriptPath);
+  const compilerTimeout = setTimeout(() => {
+    log.info("terminating circuit worker due to timeout");
```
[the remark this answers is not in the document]
minestarks: Added.

vscode/src/codeLens.ts
minestarks: Yes, I thought we may want to "hide" the feature for now, though on second thought I don't know why I thought that :) . I'll add a proper description for it in package.json.

npm/ux/circuit.tsx
minestarks: Thanks, I didn't know this was possible.

vscode/src/circuit.ts
```diff
@@ -27,9 +32,16 @@ export async function showCircuitCommand(
...
 
+  let timeout = false;
+
   // Start the worker, run the code, and send the results to the webview
   const worker = getCompilerWorker(compilerWorkerScriptPath);
   const compilerTimeout = setTimeout(() => {
+    timeout = true;
+    sendTelemetryEvent(EventType.CircuitEnd, {
+      associationId,
+      flowStatus: UserFlowStatus.Aborted,
```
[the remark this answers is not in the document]
minestarks: I'm happy to add timeout to the enum! Wasn't sure if it was okay for me to mess with the general schema.

Isn't generating a circuit a user flow? I'm not fully getting the distinction.

npm/ux/circuit.tsx
```diff
@@ -18,3 +18,31 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
...
+        <h1>{props.title}</h1>
+        <h2>{props.subtitle}</h2>
+      </div>
+      {props.circuit ? <Circuit circuit={props.circuit}></Circuit> : null}
+      <div>
+        Tip: you can generate a circuit diagram for any operation that takes
+        qubits or arrays of qubits as input.
+      </div>
+      <div>
+        {props.errorHtml ? (
```
billti, line 40: Does an error have any styling on it (e.g. a red box, bold text, etc.)? Or is just a regular div.

npm/ux/circuit.tsx
```diff
@@ -18,3 +18,31 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
...
+          <div dangerouslySetInnerHTML={{ __html: props.errorHtml }}></div>
+        ) : null}
+      </div>
+    </div>
+  );
+}
+
+export type CircuitData = qviz.Circuit;
```
billti, line 48: Why not also reexport this at the top level? (see following comment)

vscode/src/webview/webview.tsx
```diff
@@ -6,13 +6,19 @@
...
+  Histogram,
+  type ReData,
+} from "qsharp-lang/ux";
 import { HelpPage } from "./help";
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore - there are no types for this
 import mk from "@vscode/markdown-it-katex";
 import markdownIt from "markdown-it";
+import { CircuitData } from "qsharp-lang/ux/circuit";
```
billti, line 21: I think today we import everything from either `qsharp-lang` or `qsharp-lang/ux`. If you reexport this type at the top level as you do for Circuit and CircuitPanel already, then just add this import with those on line 9.

vscode/src/config.ts
```diff
@@ -27,8 +27,26 @@ export async function setTarget(target: TargetProfile) {
...
 export function getEnableFormating(): boolean {
   return vscode.workspace
     .getConfiguration("Q#")
     .get<boolean>("enableFormatting", true);
 }
+
+export function getShowCircuitCodeLens(): boolean {
+  return vscode.workspace
+    .getConfiguration("Q#")
+    .get<boolean>("showCircuitCodeLens", false);
```
billti, line 51: Should we remove this now that we'll show it by default in dev builds? (Or at least default to 'true')

vscode/src/circuit.ts
```diff
@@ -0,0 +1,134 @@
...
+  } finally {
+    log.info("terminating circuit worker");
+    worker.terminate();
+  }
+}
+
+function errorsToHtml(errors: [string, VSDiagnostic][]): string {
+  let errorHtml = "";
+  for (const error of errors) {
+    const uri = Uri.parse(error[0]);
```
billti, line 113: What is at element 0? Can you add some comments as to what these commands are you're embedding into the html.

vscode/src/circuit.ts
```diff
@@ -0,0 +1,134 @@
...
+    const fsPath = escapeHtml(uri.fsPath);
+    const errorText = escapeHtml(
+      `:${error[1].range.start.line}:${error[1].range.start.character}: ${error[1].message} (${error[1].code})`,
+    );
+    errorHtml += `<a href="${openCommandUri}">${fsPath}</a>${errorText}<br/>`;
+  }
+  return errorHtml;
+}
+
+function escapeHtml(unsafe: string): string {
```
billti, line 127: Is encodeUriComponent not suitable here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent (Or `encodeURI` - I can never remember which to use when)

billti: (Actually, I guess that' for URLs, not for escaping HTML characters. Ignore me)

minestarks: This gets me every time. I'm outraged there's no built-in HTML escape function in JavaScript...

vscode/src/circuit.ts
```diff
@@ -0,0 +1,134 @@
...
+import { loadProject } from "./projectSystem";
+import type { IOperationInfo, VSDiagnostic } from "../../npm/lib/web/qsc_wasm";
+import { getTarget, getTargetFriendlyName } from "./config";
+import { sendMessageToPanel } from "./webviewPanel";
+import { getRandomGuid } from "./utils";
+import { EventType, UserFlowStatus, sendTelemetryEvent } from "./telemetry";
+
+const compilerRunTimeoutMs = 1000 * 60 * 5; // 5 minutes
+
+export async function showCircuitCommand(
```
billti, line 16: Looks good for now, but much of the logic here does seem to be boilerplate that's repeated for the Histogram and Estimates panel as well. We may want to refactor out the commonality at some point.

vscode/package.json
```diff
@@ -113,6 +113,11 @@
           "type": "boolean",
           "default": "true",
           "description": "Enables the Q# formatter."
+        },
+        "Q#.showCircuitCodeLens": {
+          "type": "boolean",
+          "default": false,
```
billti, line 119: Do you need to provide a default in the package.json and then a default value when reading the property in code also? I noticed this when Scott added the formatting setting also.

Not important. Just seems odd to have a 'default' in both places.

billti: (And as per the comment in the code that reads it, we should default to 'true' while this is going out only in dev builds)

vscode/package.json
```diff
@@ -270,6 +275,11 @@
         "title": "Help",
         "category": "Q#"
       },
+      {
+        "command": "qsharp-vscode.showCircuit",
+        "title": "Show circuit",
+        "category": "Q#"
```
billti, line 281: I think this shows up by default in the Command Palette unless you add a condition (in this case, should be conditions on the config value to enable it). See <https://code.visualstudio.com/api/extension-guides/command#controlling-when-a-command-shows-up-in-the-command-palette>

wasm/src/diagnostic.rs
```diff
@@ -187,3 +187,27 @@ where
         message: labeled_span.label().map(ToString::to_string),
     }
 }
+
+pub fn interpret_errors_into_vs_diagnostics(
+    errs: &[interpret::Error],
+) -> Vec<(String, VSDiagnostic)> {
+    let default_uri: Arc<str> = Arc::from("<project>");
```
billti, line 194: Does this need to be `Arc<str>` or could it be `Rc<str>`? (I recall we had some Miette requirements that may come into play here)

wasm/src/lib.rs
```diff
@@ -133,6 +135,44 @@ pub fn get_estimates(
...
+        })
+        .map_err(interpret_errors_into_vs_diagnostics_json)?;
+
+    serde_wasm_bindgen::to_value(&circuit).map_err(|e| e.to_string())
+}
+
+#[allow(clippy::needless_pass_by_value)]
+fn interpret_errors_into_vs_diagnostics_json(errs: Vec<qsc::interpret::Error>) -> String {
+    serde_json::to_string(&interpret_errors_into_vs_diagnostics(&errs))
+        .expect("serializing errors to json should succeed (or should it?)")
```
billti, line 173: Should it? :-p

minestarks: Uh oh, forgot I left this in! Yes, it should succeed...

wasm/src/diagnostic.rs
minestarks: Yeah, `Arc<str>` did come from Miette, but actually I can avoid some unnecessary conversions by just declaring this as a string. Updating...

vscode/src/config.ts
minestarks: I wanted to add a package name check for the "enable in dev builds" thing and doing that in a different PR. But if that sounds overly complicated I can just set it to true now.

vscode/src/circuit.ts
minestarks: I'd love to do that in a future PR. The error handling should probably be consolidated too - I print runtime errors into the webview now, but couldn't find a clean way to do it for the other panel (where would we even put that in the estimates?). Requires a bit of thought.

vscode/package.json
minestarks: Yeah, I was copying his example. The annoying thing is that `vscode.workspace.getConfiguration("Q#").get<boolean>("showCircuitCodeLens")` technically returns `boolean | undefined`, so I went with the lesser of two evils between adding an `!` and specifying a default value. At least I'll add a comment to clarify.

npm/ux/circuit.tsx
minestarks: I just added a CSS class so I could tweak the padding. I didn't want to do any styling because it seemed like a big time sink, though I might like to do that in the future. This is how it looks when a circuit fails now. Not pretty but gets the job done?

image

vscode/src/circuit.ts
minestarks: Added comments, and in the process discovered a bug handling project-level errors. Thanks!

npm/src/browser.ts
```diff
@@ -158,6 +158,17 @@ export function getLanguageServiceWorker(
...
+export type {
+  IBreakpointSpan,
+  ICodeLens,
+  ILocation,
+  IOperationInfo,
+  IPosition,
+  IRange,
+  IStackFrame,
+  VSDiagnostic,
+} from "../lib/web/qsc_wasm.js";
```
billti, line 171: This isn't a new problem (I just spotted it in other code checked in over the past few months), but we used to always import from "../lib/node/qsc_wasm.js". The Node types are a (slightly) simpler subset of the web types which should work in all environments (but not vice-versa)

Don't worry about it here if a pain to change. I may go clean this up across the code-base later (and try to think how we could catch it)

vscode/src/circuit.ts
```diff
@@ -0,0 +1,172 @@
...
+    let location;
+    const document = error[0];
+    try {
+      // If the error location is a document URI, create a link to that document.
+      // We use the `vscode.open` command (https://code.visualstudio.com/api/references/commands#commands)
+      // to open the document in the editor.
+      // The line and column information is displayed, but are not part of the link.
+      //
+      // At the time of writing this is the only way we know to create a direct
+      // link to a Q# document from a Web View.
```
billti, line 135: A WebView can also send a message back to the main thread, which can then take whatever action the extension can take. We don't handle any custom messages currently, but I put the infrastructure in (see webviewPanel.ts@504)

npm/src/browser.ts
minestarks: I don't mind taking care of it now, but, just to confirm:

We already have a _runtime_ import of `../lib/web/qsc_wasm.js` in this file. I believe that has to stay as `web`. Are you suggesting importing the types from `node`, but the runtime exports from `web`, even if they both occur in the same file?

If so, something like this 
```ts
import {
  IOperationInfo,
  TargetProfile,
  type VSDiagnostic,
} from "../../lib/web/qsc_wasm.js";
```

Has to turn into two separate import statements which seems a bit overkill.

vscode/src/circuit.ts
minestarks: Oh, that's potentially even better... is it ok if I just update this comment and worry about it later? The current link approach seemed like the easiest way to get the job done.

billti (approved):

swernli (approved):

Issue #1185, opened 2024-02-21
Migrate more katas to quantum.microsoft.com


tcNickolas:
There are multiple katas in the Quantum Katas project that cannot be used with the modern QDK, and are straightforward to migrate to the [new katas experience](https://quantum.microsoft.com/en-us/experience/quantum-katas). I don't have detailed notes for each of them, so I think we can track them all in this one issue.

* [x] [BasicGates](https://github.com/microsoft/QuantumKatas/blob/main/BasicGates): makes sense to split it, add tasks 1.1-1.7 to the end of the Single-Qubit Gates kata, 1.8-1.10 - to the end of the Multi-Qubit Systems kata, 2.1-2.6 - to Multi-Qubit Gates kata 
* [x] [Superposition](https://github.com/microsoft/QuantumKatas/blob/main/Superposition): I started migration already, need to finish the rest of the tasks (I've updated the infrastructure to support solutions without adjoint, so we can migrate all content) 
* [x] [Measurements](https://github.com/microsoft/QuantumKatas/blob/main/Measurements) (renamed to Distinguishing Quantum States): add as a standalone kata after Multi-Qubit Measurements, testing infra should be similar to that one
* [x] [DistinguishUnitaries](https://github.com/microsoft/QuantumKatas/blob/main/DistinguishUnitaries): add as a standalone kata after Measurements, needs its own infra but should be similar to existing Measurements 
* [x] [Teleportation](https://github.com/microsoft/QuantumKatas/blob/main/Teleportation): add after DistinguishUnitaries
* [x] [SuperdenseCoding](https://github.com/microsoft/QuantumKatas/blob/main/SuperdenseCoding): add after Teleportation 
* [x] [MarkingOracles](https://github.com/microsoft/QuantumKatas/blob/main/MarkingOracles): add after existing Oracles kata, include tasks 1.5, 1.6, and 1.8 from https://github.com/microsoft/qsharp/issues/1156 (task 1.8 is a building block for majority task from MarkingOracles)

jkingdon-ms:
I will be working on migrating Superposition Kata - Task 1.6 (Bell state)

jkingdon-ms:
I will be working on migrating Superposition Kata - Task 1.7 (All Bell states)

jkingdon-ms:
I will be working on migrating Superposition Kata - Tasks 1.8 & 1.9

SakthiVijayS:
I will be working on migrating Superdense Coding Kata

[...]

PR #1271, opened 2024-03-18
Add All Bell States task to Superposition kata

jkingdon-ms:
# Adding All Bell States Kata

### DIFF
--- a/katas/content/superposition_kata/all_bell_states/Verification.qs
+++ b/katas/content/superposition_kata/all_bell_states/Verification.qs
@@ -22,11 +22,9 @@ namespace Kata.Verification {
     operation CheckSolution() : Bool {
         for index in 0 .. 3 {
             if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
-                Kata.AllBellStates,
-                AllBellStates_Reference,
-                2,
-                index) {
-                Message("Incorrect");
+                Kata.AllBellStates(_, index),
+                AllBellStates(_, index),
+                2) {
                 Message($"The test case for index = {index} did not pass.");
                 return false;
             }
--- a/katas/content/superposition_kata/all_bell_states/Verification.qs
+++ b/katas/content/superposition_kata/all_bell_states/Verification.qs
@@ -21,7 +21,7 @@ namespace Kata.Verification {
     @EntryPoint()
     operation CheckSolution() : Bool {
         for index in 0 .. 3 {
-            if not CheckOperationsEquivalenceOnZeroStateAndIndexWithFeedback(
+            if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
                 Kata.AllBellStates,
                 AllBellStates_Reference,
                 2,
--- a/katas/content/superposition_kata/all_bell_states/Verification.qs
+++ b/katas/content/superposition_kata/all_bell_states/Verification.qs
@@ -30,7 +30,7 @@ namespace Kata.Verification {
             }
         }
 
-        Message("Correct");
+        Message("Correct!");
         return true;
     }
 }
--- a/katas/content/superposition_kata/all_bell_states/Verification.qs
+++ b/katas/content/superposition_kata/all_bell_states/Verification.qs
@@ -21,16 +21,15 @@ namespace Kata.Verification {
     @EntryPoint()
     operation CheckSolution() : Bool {
         for index in 0 .. 3 {
+            Message($"Testing index = {index}...");            
             if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
                 Kata.AllBellStates(_, index),
-                AllBellStates(_, index),
+                AllBellStates_Reference(_, index),
                 2) {
-                Message($"The test case for index = {index} did not pass.");
                 return false;
             }
         }
 
-        Message("Correct!");
         return true;
     }
 }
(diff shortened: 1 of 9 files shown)


katas/content/superposition_kata/all_bell_states/Verification.qs
```diff
@@ -0,0 +1,38 @@
...
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        for index in 0 .. 3 {
+            if not CheckOperationsEquivalenceOnZeroStateAndIndexWithFeedback(
+                Kata.AllBellStates,
+                AllBellStates,
+                2,
+                index) {
+                Message("Incorrect");
```
tcNickolas, line 29: We can simplify this code using [partial application](https://learn.microsoft.com/en-us/azure/quantum/user-guide/language/expressions/closures#partial-application), and then we won't need the changes to the libraries at all!
```suggestion
            if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
                Kata.AllBellStates(_, index),
                AllBellStates(_, index),
                2) {
```

We'll be using different signatures for the states we want to compare - sometimes integers, other times Booleans or arrays, so it's not really practical to introduce a new library for each signature of the operation we want to test. Instead, we bind the integer parameter of these operations to the index we want to test and leave qubit parameters unbound - the result is an operation with a single parameter that we can pass to a regular library. Partial application is very useful and especially in testing - let me know if you want more details, I'll try to dig up more docs.

(And with this, you can revert the change to the KatasLibrary.qs)

katas/content/superposition_kata/all_bell_states/Verification.qs
```diff
@@ -0,0 +1,38 @@
...
+                Message($"The test case for index = {index} did not pass.");
+                return false;
+            }
+        }
+
+        Message("Correct");
```
tcNickolas, line 35: Let's cheer for the learner :-) (I like to keep this kind of messages uniform across all katas)
```suggestion
        Message("Correct!");
```

tcNickolas (changes requested):
Looks good overall, I left a few comments. The most significant one regarding using partial application to fix index when testing instead of introducing new libraries - you'll be able to use this trick a couple of tasks later when dealing with preparing states with other additional classical variables.

Thank you!

tcNickolas:
I've played with this a bit more, and my rewrite suggestion can be improved: right now if I just prep the first Bell state (index = 0), it prints Correct! before continuing to Incorrect on the next test. We can do it slightly different, print "Testing for index = 0..." before doing that code, so that the verdict Correct! is just for that test case, and then we don't need a separate failure message - Incorrect will say it for us. (You can see an example of that done in katas\content\random_numbers\random_n_bits\Verification.qs, verdict is printed by IsSufficientlyRandom)

katas/content/superposition_kata/all_bell_states/Verification.qs
jkingdon-ms: good to know! thanks for the tip.

jkingdon-ms:
> I've played with this a bit more, and my rewrite suggestion can be improved: right now if I just prep the first Bell state (index = 0), it prints Correct! before continuing to Incorrect on the next test. We can do it slightly different, print "Testing for index = 0..." before doing that code, so that the verdict Correct! is just for that test case, and then we don't need a separate failure message - Incorrect will say it for us. (You can see an example of that done in katas\content\random_numbers\random_n_bits\Verification.qs, verdict is printed by IsSufficientlyRandom)

Good point, I can update it like that for clarity.

katas/content/superposition_kata/all_bell_states/solution.md
```diff
@@ -0,0 +1,65 @@
...
+$$\frac{1}{\sqrt2} \begin{bmatrix} 1 & 0 & 1 & 0 \\\ 0 & 1 & 0 & 1 \\\ 0 & 1 & 0 & -1 \\\ \underset{|\Phi^{+}\rangle}{\underbrace{1}} & \underset{|\Psi^{+}\rangle}{\underbrace{0}} & \underset{|\Phi^{-}\rangle}{\underbrace{-1}} & \underset{|\Psi^{-}\rangle}{\underbrace{0}} \end{bmatrix}$$
+
+
+Notice that each of the columns in the unitary matrix corresponds to one of the Bell states.
+This unitary transformation transforms the computational basis into the Bell basis, which is exactly what the task asks us to do.
+
+We see that this transformation converts $|00\rangle$ into the first Bell state, $|01\rangle$ into the second Bell state, etc. 
+We just need to make sure we set the qubits to the correct state before applying this transformation, using $X$ gates to change the initial $|0\rangle$ states to $|1\rangle$ if needed. 
+
+In Q#, we can use the <a href="https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.convert.intasboolarray">IntAsBoolArray</a> function to convert the input `index` to the right bit pattern.
```
tcNickolas, line 60: The API docs format changed a bit with the Modern QDK, sorry I forgot to check that on the first round of review!
```suggestion
In Q#, we can use the <a href="https://learn.microsoft.com/qsharp/api/qsharp-lang/microsoft.quantum.convert/intasboolarray">IntAsBoolArray</a> function to convert the input `index` to the right bit pattern.
```

tcNickolas (approved):
Looks great! Just one link discrepancy I missed on the first review - I think I can resolve it myself by applying a suggestion.

Thank you!

PR #1272, opened 2024-03-18
Fix "initializeruntime" telemetry error

sezna:
We are currently oversending initialize runtime end events. This was because 
we incorrectly updated the telemetry when the project system was added. So,
there are currently far more _End_ events than _Start_ events in our metrics.

This PR fixes that by only sending one _End_ event per initialization.

### DIFF
--- a/vscode/src/debugger/session.ts
+++ b/vscode/src/debugger/session.ts
@@ -142,7 +142,13 @@ export class QscDebugSession extends LoggingDebugSession {
     }
     sendTelemetryEvent(
       EventType.InitializeRuntimeEnd,
-      { associationId, flowStatus: UserFlowStatus.Succeeded },
+      {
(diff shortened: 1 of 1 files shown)


vscode/src/debugger/session.ts
```diff
@@ -137,15 +137,6 @@ export class QscDebugSession extends LoggingDebugSession {
         this.breakpointLocations.set(path, mapped);
       } else {
         log.warn(`compilation failed. ${failureMessage}`);
-        sendTelemetryEvent(
```
minestarks, line 140: Doesn't that mean the `UserFlowStatus.Succeeded` event below is now incorrect? Should we be sending the failure information in that event?

minestarks (approved):

billti (approved):

PR #1273, opened 2024-03-18
Update qsharp.schema.json for language features

sezna:
This was never updated when we merged language features in.


minestarks (approved):

billti:
Should we also add lints to this if that's about to go in? (Or have Oscar also update in his PR)

orpuente-MS:
I just added it to my PR

PR #1275, opened 2024-03-19
Format on Document Range

ScottCarda-MS:
Enables the feature to format highlighted Q# code, rather than formatting the whole file.

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -39,7 +39,7 @@ import { initCodegen } from "./qirGeneration.js";
 import { createReferenceProvider } from "./references.js";
 import { createRenameProvider } from "./rename.js";
 import { createSignatureHelpProvider } from "./signature.js";
-import { createFormatProvider } from "./format.js";
+import { createFormattingProvider } from "./format.js";
 import { activateTargetProfileStatusBarItem } from "./statusbar.js";
 import {
   EventType,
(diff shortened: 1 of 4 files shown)


vscode/src/format.ts
```diff
@@ -37,10 +49,17 @@ class QSharpFormatProvider implements vscode.DocumentFormattingEditProvider {
...
 
-    const edits = lsEdits.map(
+    let edits = lsEdits.map(
       (edit) => new vscode.TextEdit(toVscodeRange(edit.range), edit.newText),
     );
 
+    if (range) {
+      edits = edits.filter(
+        (e) =>
+          range.start.isBefore(e.range.end) && range.end.isAfter(e.range.start),
```
minestarks, line 59: Does `range.intersection(e.range) !== undefined` do the trick here?

Also confirming: you want all the edits that overlap the requested range, not just the ones wholly contained in the range?

vscode/src/format.ts
minestarks, line 1: Clever - it wouldn't have occurred to me to implement it at the vscode layer.

Would you please add an integration test for it though? It's the only way to get coverage at code implemented at this layer.

minestarks (approved):

vscode/src/extension.ts
```diff
@@ -290,13 +296,19 @@ async function updateLanguageServiceEnableFormatting(
...
+      );
+    formatterHandle.DocumentRangeFormattingHandle =
+      vscode.languages.registerDocumentRangeFormattingEditProvider(
+        qsharpLanguageId,
+        createFormattingProvider(languageService),
       );
   } else {
-    formatterHandle.handle?.dispose();
+    formatterHandle.DocumentFormattingHandle?.dispose();
+    formatterHandle.DocumentRangeFormattingHandle?.dispose();
```
billti, line 311: Minor nit: But best practice would be to set both those values to null after you dispose them.

vscode/src/format.ts
ScottCarda-MS: Yes we want the edits that overlap.

The `range.intersection(e.range) !== undefined` has slightly different behavior that is not desirable. That one is equal to `range.start.isBeforeOrEqual(e.range.end) && range.end.isAfterOrEqual(e.range.start)`, causing edits to happen if they are adjacent to the highlighted text, not inside the highlighted text.

minestarks: Okay, fair enough!

billti (approved):

vscode/src/extension.ts
ScottCarda-MS: I set them to `undefined` after disposal.

vscode/src/format.ts
ScottCarda-MS: added

PR #1276, opened 2024-03-19
Require cmake 3.21 version

DmitryVasilevsky:
cmake 3.21 is required to support "-G" "Visual Studio 17 2022", which is needed for mimalloc-sys.

### DIFF
--- a/prereqs.py
+++ b/prereqs.py
@@ -200,7 +200,7 @@ def check_prereqs(install=False):
             cmake_major == cmake_ver[0] and cmake_minor < cmake_ver[1]
         ):
             print(
-                f"CMake v{cmake_ver[0]}.{cmake_ver[1]} or later is required. Please update to Latest Release from https://cmake.org/download/"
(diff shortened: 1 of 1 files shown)


prereqs.py
```diff
@@ -200,7 +200,7 @@ def check_prereqs(install=False):
             cmake_major == cmake_ver[0] and cmake_minor < cmake_ver[1]
         ):
             print(
-                f"CMake v{cmake_ver[0]}.{cmake_ver[1]} or later is required. Please update."
+                f"CMake v{cmake_ver[0]}.{cmake_ver[1]} or later is required. Please update to Latest Release from https://cmake.org/download/"
```
billti, line 203: Why is 'Latest Release' capitalized?

billti (approved):

prereqs.py
DmitryVasilevsky: If you follow the link, the first section would be "Release Candidate" and the second section would be "Latest Release". I wanted to draw attention to the fact that one needs to scroll down to that second section. Maybe there's a better way.

DmitryVasilevsky: Updated text.

PR #1278, opened 2024-03-20
Revert "Native libraries use mimalloc as global allocator (#1249)"

billti:
This reverts commit 93f01d0de2eb70945b40d219dc8daec9b6b5a6c3.

Reverting until we can figure out why Windows ARM64 build is failing.


swernli (changes requested):
This should not be needed and we can use https://github.com/microsoft/qsharp/pull/1279 instead.

billti:
Closing in favor of #1279 (for now)

PR #1286, opened 2024-03-21
Add createProject command

billti:


billti:
Context menu:

image

Resulting file structure

image

If run from the command palette instead of the context menu, it creates the project in the root workspace folder.

vscode/src/createProject.ts
```diff
@@ -0,0 +1,66 @@
...
+              { placeHolder: "Pick the workspace folder for the project" },
+            );
+            if (!workspaceChoice) return;
+            folderUri = workspaceChoice.uri;
+          }
+        }
+
+        const edit = new vscode.WorkspaceEdit();
+        const projUri = vscode.Uri.joinPath(folderUri, "qsharp.json");
+        const mainUri = vscode.Uri.joinPath(folderUri, "src", "main.qs");
```
minestarks: Should this be `Main.qs` to stick to our naming convention? (seem to recall @sezna bringing this up in the namespaces spec?)

minestarks: nvm, outdated comment.

PR #1287, opened 2024-03-21
Multiple magic states and factory types

msoeken:
This PR
- simplifies the `Overhead` trait by requiring information about the number of magic states per rotation (moved to implementation)
- allows to specify an index for the type of magic state (both in `Overhead` and in `FactoryBuilder` to determine how many different magic states are provided)
- changes the return type of `FactoryBuilder::find_factories` to return `Vec<Cow<Factory>>` instead of `Vec<Factory>` such that implementors do not need to copy factories in case they are already pre-computed
- adds helper structs to pass no factories or to merge two separate factories to produce a factory builder that produces two different magic states


resource_estimator/src/estimates/factory/dispatch.rs
```diff
@@ -0,0 +1,101 @@
+use crate::estimates::{ErrorCorrection, Factory, FactoryBuilder};
+use std::borrow::Cow;
+
+#[derive(Clone)]
+pub enum FactoryDispatch2<F1, F2> {
```
ivanbasov, line 5: Not sure if enum name with index and enums themselves are descriptive. Can we have maybe much longer name but self-descriptive?

resource_estimator/src/estimates/factory/dispatch.rs
```diff
@@ -0,0 +1,101 @@
...
+                )
+                .into_iter()
+                .map(|f| Cow::Owned(FactoryDispatch2::F2(f.into_owned())))
+                .collect(),
+            _ => unreachable!("factory builder only has two magic state types"),
+        }
+    }
+
+    fn num_magic_state_types(&self) -> usize {
+        2
```
ivanbasov, line 99: a magic number for a magic state :-)

resource_estimator/src/system/data/report.rs
```diff
@@ -36,12 +36,14 @@ impl Report {
         formatted_counts: &FormattedPhysicalResourceCounts,
     ) -> Self {
         let logical_counts = result.layout_overhead();
+        let part = result.factory_parts()[0].as_ref();
```
ivanbasov, line 39: Dispalying just the first one? Why?

ivanbasov (changes requested):

resource_estimator/src/estimates/factory/dispatch.rs
msoeken: I cannot come up with a better name, so I added a doc string to make clearer what this is used for. Eventually there might be `FactoryDispatch3` and `FactoryDispatch4`, but since there is no use case for more than 2 magic state types for now, I have not added them yet.

resource_estimator/src/estimates/factory/dispatch.rs
msoeken: It's the number of different magic state types this factory builder can support. It's documented now.

resource_estimator/src/system/data/report.rs
msoeken: The system instantiation only supports T states as magic states, but the generic implementation allows to model systems with more than one magic state type.

msoeken: I added a comment.

resource_estimator/src/estimates/factory/dispatch.rs
ivanbasov: I see. Thank you!
 A couple ideas:
1.  Would it be easier to read if generic parameters would be Factory1 and Factory2 instead of F1 and F2?
2. Same for enum

Or FactoryType1 and FactoryType2

msoeken: Thanks, I updated the code and use the longer names for the generic parameters, both for the factory and the builder.

ivanbasov (approved):

Issue #1025, opened 2024-01-17
Q# Formatter


ScottCarda-MS:
Let's use this issue to collect feedback and gather ideas about implementing a formatter for Q# to help standardize a consistent coding style across Q# code.
```[tasklist]
### Formatter Rules to Implement
- [x] Correct Indentation #1289
- [x] Correct Brace Style #1289
- [x] Correct spacing around operators #1172
- [x] Correct spacing around type-annotation colon #1172
- [x] Correct spacing around parentheses, brackets, and commas (sequences) #1289
- [ ] Remove unnecessary newlines
- [ ] ~Remove trailing commas in sequences with more than one element~ will be handled by the Linter
- [ ] ~Remove duplicate semicolons~ will be handled by the Linter
```
```[tasklist]
### Tasks
- [x] Decide on data structures for formatter
- [x] Format document on save #1172
- [x] Format code snippets #1275
- [ ] Format on typing
- [x] Format in Playground #1307
- [x] Format Samples/Libraries #1303
- [x] Add format of Samples/Libraries to build pipeline #1329
- [ ] Use indentation style specified by VsCode (or other endpoint)
- [ ] ~Support formatting around unary `-` and `+`~ not doable with this version of the formatter
- [x] Investigate removal of unauthoritative newlines #1303
```

PR #1289, opened 2024-03-21
Formatter Indentation and Delimiter Logic

ScottCarda-MS:
Added some state to the formatter for the purposes of better handling of delimiters, i.e. `[]`, `()`, `{}`, and type parameter list delimiters `<>`. The formatting of delimited blocks will be determined by whether there is a newline in the whitespace following the open delimiter.
 - If there is a newline, then the indentation is increased for the length of the block, and separators, i.e. ',' and ';', require newlines after them, with the closing delimiter having a newline in front of it.
 - If there are no newlines, then delimiters have no spacing after the opening and before the closing delimiters, except the brace delimiters which have a singe space instead, and separators require a single space after them.

### DIFF
--- a/compiler/qsc_formatter/src/formatter.rs
+++ b/compiler/qsc_formatter/src/formatter.rs
@@ -36,6 +36,21 @@ enum NewlineContext {
     Spaces,
 }
 
+#[derive(Clone, Copy)]
+enum TypeParameterListState {
+    NoState,
+    SeenCallableKeyword,
+    SeenCallableName,
+    InTypeParamList,
+}
+
+struct FormatterState<'a> {
+    code: &'a str,
+    indent_level: usize,
+    delim_newlines_stack: Vec<NewlineContext>,
+    type_param_state: TypeParameterListState,
+}
+
 fn make_indent_string(level: usize) -> String {
     "    ".repeat(level)
 }
@@ -62,8 +77,12 @@ pub fn calculate_format_edits(code: &str) -> Vec<TextEdit> {
     let tokens = concrete::ConcreteTokenIterator::new(code);
     let mut edits = vec![];
 
-    let mut indent_level: usize = 0;
-    let mut delim_context_stack: Vec<NewlineContext> = vec![];
+    let mut state = FormatterState {
+        code,
+        indent_level: 0,
+        delim_newlines_stack: vec![],
+        type_param_state: TypeParameterListState::NoState,
+    };
 
     // The sliding window used is over three adjacent tokens
     #[allow(unused_assignments)]
@@ -84,24 +103,10 @@ pub fn calculate_format_edits(code: &str) -> Vec<TextEdit> {
                     continue;
                 } else if matches!(two.kind, ConcreteTokenKind::WhiteSpace) {
                     // whitespace in the middle
-                    apply_rules(
-                        one,
-                        get_token_contents(code, two),
-                        three,
-                        code,
-                        &mut indent_level,
-                        &mut delim_context_stack,
-                    )
+                    apply_rules(one, get_token_contents(code, two), three, &mut state)
                 } else {
                     // one, two are adjacent tokens with no whitespace in the middle
-                    apply_rules(
-                        one,
-                        "",
-                        two,
-                        code,
-                        &mut indent_level,
-                        &mut delim_context_stack,
-                    )
+                    apply_rules(one, "", two, &mut state)
                 }
             }
             (None, None, Some(three)) => {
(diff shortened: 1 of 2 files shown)


ScottCarda-MS:
I've run this version of the formatter against the samples and library files to test that nothing unreasonable or unexpected occurs. I don't want to add the updated samples/libraries to this PR though, as I think that would be better for a future PR where I could include running the formatter automatically in the build pipeline.

compiler/qsc_formatter/src/formatter.rs
minestarks, line 1: In the spirit of the comment I left on Stefan's evaluator PR ([which led to this doc comment](https://github.com/microsoft/qsharp/blob/072d85de6ccc32cc78e4537fdc14b18b5b71affc/compiler/qsc_eval/src/lib.rs#L4-L15) - I think it would be nice if there was a doc comment explaining the general approach taken with the formatter. Not necessarily the implementation details, but a general "what are we doing here". Perhaps mentioning the convolution approach, what kind of context/state we keep track of etc.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -29,6 +29,28 @@ impl TextEdit {
...
+    Spaces,
+}
+
+#[derive(Clone, Copy)]
+enum TypeParameterListState {
+    NoState,
+    SeenCallableKeyword,
+    SeenCallableName,
+    InTypeParamList,
+}
```
minestarks, line 45: Can these have doc comments explaining what each context/state means?

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -121,36 +133,124 @@ fn apply_rules(
     left: &ConcreteToken,
     whitespace: &str,
     right: &ConcreteToken,
-    code: &str,
-    indent_level: usize,
+    state: &mut FormatterState,
```
minestarks, line 136: Minor style nit: I think once you have a parameter named `state`, you're basically writing the `impl` for that struct. It would seem more intuitive to me, if instead of being a free function, `apply_rules` was a method of `FormatterState` (or just `Formatter`) and took `&mut self` instead of `state` as a named parameter.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I put a doc comment like that on the lib.rs file as module-level documentation.

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -121,36 +133,124 @@ fn apply_rules(
...
+    let mut newline_context = state
+        .delim_newlines_stack
+        .last()
+        .map_or(NewlineContext::NoContext, |b| *b);
 
     use qsc_frontend::keyword::Keyword;
     use qsc_frontend::lex::cooked::ClosedBinOp;
     use ConcreteTokenKind::*;
     use TokenKind::*;
+
```
minestarks, line 151: With the two new `match` expressions, this function is quite long - would you extracting out a couple of helper functions? `update_type_param_state` and `update_delim_state` maybe

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -686,6 +685,213 @@ fn formatting_corrects_indentation() {
...
+
+            a,  b, c
+        ];
+"#,
+        &expect![[r#"
+            let x = [a, b, c];
+            let y = [a, b, c];
+            let z = [
+
+
```
minestarks, line 886: Not suggesting it as a change on this PR, but just curious: is it theoretically possible for us to have authoritative logic around collapsing multiple new lines into one? 

I have seen some formatters do it, and some don't. I can't even make up my mind about whether it's desirable.

compiler/qsc_formatter/src/formatter.rs
minestarks: There it is. Sorry I missed it! 

Maybe consider adding the reasoning behind all the newline/indentation handling we do, but either way I'm satisfied with it as is.

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: I think it should be possible, at least to stop more the 2 newlines from showing up anywhere. I should really put together a little todo task list for improvements like this to the formatter.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: Done

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: Done

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: `apply_rules` as well as the new state updater functions are put into the new `impl` for the renamed `Formatter` struct. Thanks for the suggestion!

minestarks (approved):

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: I added a sentence to the description about the use of state to determine indentation. I don't want to add too much detail to that very high-level doc comment, though, so I think the one sentence should be good enough for that.

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: Turns out I already had a todo-list style issue for the formatter: #1025. I've updated it a bit since I last looked at it.

PR #1290, opened 2024-03-21
update to Rust 1.77.0

filipw:
Rust [1.77.0](https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html) was released earlier today. Unfortunately the repo no longer builds once it is installed, due to the usual new clippy checks.


billti (changes requested):
We use a Microsoft build of Rust internally to prevent supply-chain attacks. There's a slight lag between a new public Rust version and updating internally. We can't merge this until the internal version is updated else it will break our publishing pipeline.

filipw:
I was wondering what the cryptic `ms-$(RUST_TOOLCHAIN_VERSION)` stood for 😅

filipw:
> We can't merge this until the internal version is updated else it will break our publishing pipeline.

I reverted the changes to CI/CD pipeline and left the code changes only. This way it still builds with 1.76, but also work fine with 1.77 so no one has a broken repo if they (like me) upgrade Rust locally.

billti (approved):
Thanks!

minestarks (approved):

Issue #1285, opened 2024-03-21
Reconsider the names of some data structures and functions in the RCA implementation


cesarzc:
The Runtime Capabilities Analysis (RCA) implementation uses some data structures and functions that rely on concepts that are difficult to convey through names. However, we can improve some of them to make the algorithm easier to understand. Here's a list of the names to reconsider:
- [ ] `ApplicationGeneratorSet`
    - **Description**: A set of compute properties associated to a callable or one of its elements, from which the properties of any
 particular call application can be computed.
    - **Location**: _compiler/qsc_rca/src/lib.rs_
- [ ] `aggregate_runtime_features`
    - **Description**: Using the `self` `ComputeKind` as the basis, it calculates a new `ComputeKind` but only taking into account the runtime features of the `ComputeKind` to aggregate.
- [ ] `ApplicationInstance`
    - **Description**: A representation of the compute properties of a particular callable application. In other words, a representation of a function/operation call in which its arguments are binded to static or dynamic values.
    - **Location**: _compiler/qsc_rca/src/application.rs_

PR #1291, opened 2024-03-21
Runtime capabilities check pass

cesarzc:
This change introduces an FIR pass that checks whether a program can run on a target with specific capabilities.

Part of the change also improves RCA so we can provide less noisy error messages.


compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+mod base;
+
+#[cfg(test)]
+mod adaptive;
+
+#[cfg(test)]
+mod adaptive_plus_integers;
+
+#[cfg(test)]
+pub mod common;
```
swernli, line 14: nitpick: could you prepend these with `tests_` so it's easier to see at a glance that they are not modules for production code?

compiler/qsc_rca/tests/callables.rs
```diff
@@ -142,7 +142,7 @@ fn check_rca_for_operation_with_one_classical_return_and_one_dynamic_return() {
             Callable: CallableComputeProperties:
                 body: ApplicationsGeneratorSet:
                     inherent: Quantum: QuantumProperties:
-                        runtime_features: RuntimeFeatureFlags(UseOfDynamicBool | ForwardBranchingOnDynamicValue | ReturnWithinDynamicScope)
+                        runtime_features: RuntimeFeatureFlags(UseOfDynamicBool | ReturnWithinDynamicScope)
```
swernli, line 145: Why did forward branching capability get removed here? Shouldn't it still require that since there is a branch on measurement result?

swernli: Oh, whoops... that's the intent of the change! I should read the other files too ;)

compiler/qsc_passes/src/lib.rs
```diff
@@ -154,3 +159,16 @@ pub fn run_core_passes(core: &mut CompileUnit) -> Vec<Error> {
         .chain(base_prof_errors.into_iter().map(Error::BaseProfCk))
         .collect()
 }
+
+pub fn run_fir_passes(
+    package: &fir::Package,
```
swernli, line 164: just to confirm, the change to refactor the FIR lowering out of qsc_eval will be part of a later PR, right? At that point, we could consider pulling this up into the `PassContext` as the path for integration into the existing scenarios, but that's out of scope for now, making this effectively a test only API at the moment.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+};
+use qsc_frontend::compile::RuntimeCapabilityFlags;
+use qsc_rca::{ComputeKind, ItemComputeProperties, PackageComputeProperties, RuntimeFeatureFlags};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
+pub enum Error {
+    #[error("cannot use a dynamic boolean value")]
+    #[diagnostic(help(
+        "using a dynamic boolean value, a boolean value that depends on a measurement result, is not supported by the target"
```
swernli, line 35: I'm sure there is lots of room for tweaking how we communicate these things to users, but as a first feedback I think repeating the term "dynamic boolean" is not needed, and "target" could be more clearly stated as "current target" to imply that this is configurable. So something like this (and similar for the others below):

```suggestion
        "using a bool value that depends on a measurement result is not supported by the current target"
```

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+    ty::FunctorSetValue,
+    visit::Visitor,
+};
+use qsc_frontend::compile::RuntimeCapabilityFlags;
+use qsc_rca::{ComputeKind, ItemComputeProperties, PackageComputeProperties, RuntimeFeatureFlags};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
+pub enum Error {
+    #[error("cannot use a dynamic boolean value")]
```
swernli, line 33: nitpick: I think it's either "bool" or "Boolean" since the latter is a person's last name and a proper noun while the former is the common shorthand. That said, Q# types split the difference and are all capitalized, leading to "Bool." Which ever we pick, we should be consistent between these errors and the ones used in other places, which might mean we should unify all errors to use "Bool" and "Int" etc. since that is why type inference will output.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+
+    #[error("cannot call a cyclic function with a dynamic value as argument")]
+    #[diagnostic(help(
+        "calling a cyclic function with a dynamic value as argument, an argument value that depends on a measurement result, is not supported by the target"
+    ))]
+    #[diagnostic(code("Qsc.CapabilitiesCk.CallToCyclicFunctionWithDynamicArg"))]
+    CallToCyclicFunctionWithDynamicArg(#[label] Span),
+
+    #[error("cannot define a cyclic operation specialization")]
+    #[diagnostic(help("cyclic operation specializations, specializations that contain cycles, are not supported by the target"))]
```
swernli, line 125: maybe something like this would add more context for the user:

```suggestion
    #[diagnostic(help("operation specializations that contain call cycles are not supported by the current target"))]
```

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+    #[diagnostic(code("Qsc.CapabilitiesCk.CyclicOperationSpec"))]
+    CyclicOperationSpec(#[label] Span),
+
+    #[error("cannot call a cyclic operation")]
+    #[diagnostic(help("calling a cyclic operation is not supported by the target"))]
+    #[diagnostic(code("Qsc.CapabilitiesCk.CallToCyclicOperation"))]
+    CallToCyclicOperation(#[label] Span),
+
+    #[error("cannot call a function or operation whose resolution is dynamic")]
+    #[diagnostic(help("calling a function or operation whose resolution is dynamic, a resolution that depends on a measurement result, is not supported by the target"))]
```
swernli, line 135: nitpick: I might be worth thinking a bit about whether we can find a different term other than "resolution" because that can lead to some confusion. This came up when working on the compiler internals that use the term, and it is the right noun to use, but might not be the clearest thing for a help message. That said, I don't have any alternative suggestions, so feel free to table this one for a later discussion with a broader group.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,538 @@
...
+            }
+        };
+    }
+
+    fn visit_expr(&mut self, expr_id: ExprId) {
+        let expr = self.get_expr(expr_id);
+
+        // We do not want to generate errors for auto-generated expressions.
+        if self.is_expr_auto_generated(expr) {
+            println!("{expr_id}");
```
swernli, line 235: whoops, was this left in from testing?

swernli (approved):
Some minor feedback and wordsmithing for error messages, but that's likely to be a long tail anyway as we refine user feedback (and something others can weigh in on further into integration). The core structure looks good!

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,555 @@
...
+                    .get_global(current_callable_id)
+                    .expect("callable not present in package");
+                let Global::Callable(callable_decl) = current_callable else {
+                    panic!("");
+                };
+
+                self.missing_features_map
+                    .entry(callable_decl.name.span)
+                    .and_modify(|f| *f |= missing_spec_level_runtime_features)
+                    .or_insert(missing_spec_level_runtime_features);
```
swernli, line 399: You don't actually need to use itertools for this, and can instead use built-in stdlib stuff as long as `RuntimeFeatureFlags` gets support for `Default`:

```suggestion
                *(self
                    .missing_features_map
                    .entry(callable_decl.name.span)
                    .or_default()) |= missing_spec_level_runtime_features;
```

Cargo.toml
```diff
@@ -50,6 +50,7 @@ indoc = "2.0"
 js-sys = "0.3"
 libfuzzer-sys = "0.4"
 log = "0.4"
+itertools = "0.10.5"
```
swernli, line 53: The dependency on itertools can be removed if you use the `entry().or_default()` pattern in the capabilities check.

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: This part of the code does not need itertools. The place where I am using it is [here](https://github.com/microsoft/qsharp/blob/644bacd3436fc78cbd32ed37f6d1904eae91dc1f/compiler/qsc_passes/src/capabilitiesck.rs#L437), to being able to always produce the errors in an expected order (otherwise tests are flaky because of the ordering).

Cargo.toml
cesarzc: `and_modify` and `or_insert` can be used without itertools. I included it to be able to sort the generated errors to get an expected order because tests that yield multiple errors are flaky if the errors are not in an expected order.

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: oops, sorry, removed :)

compiler/qsc_passes/src/capabilitiesck.rs
swernli: Ah, I commented on that line then. I think it would be better to avoid itertools, as it makes code look nicer but introduces unexpected costs.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -0,0 +1,555 @@
...
+                } else {
+                    None
+                }
+            })
+    }
+
+    fn generate_errors(&mut self) -> Vec<Error> {
+        let mut errors = Vec::new();
+
+        for (span, missing_features) in self.missing_features_map.drain().sorted() {
```
swernli, line 437: We should avoid `sorted` as it sneakily introduces a copy (not always, but often). I'd prefer to be explicit about it so ownership is easy to track:

```suggestion
        let mut missing_features_map = self.missing_features_map.drain().collect::<Vec<_>>();
        missing_features_map.sort_unstable();
        for (span, missing_features) in missing_features_map {
```

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: Removed itertools dependency using your suggestion :)

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: Updated to not use `sorted`.

Cargo.toml
cesarzc: itertools dependency removed.

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: Modules renamed.

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: Updated.

compiler/qsc_passes/src/capabilitiesck.rs
cesarzc: I included this in #1285, which tracks the names/words we want to re-consider.

PR #1292, opened 2024-03-21
Residualized Intermediate Representation (RIR)

cesarzc:


compiler/qsc_rir/src/rir.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// The input type of the callable.
+    pub input_type: Vec<Ty>,
+    /// The output type of the callable.
+    pub output_type: Option<Ty>,
+    /// The callable body.
+    /// N.B. `None` bodys represent an intrinsic.
+    pub body: Option<BlockId>,
+}
+
+pub enum Instruction {
```
swernli, line 39: Thinking about it some more, I think removing the concept of statement means some instructions need to have a Variable field for their result. Take LogicalNot as an example. Where does the result of that instruction get written to? I think Call will need an Option<Variable> and others that always produce a value will need Variable as a field.

swernli (commented):
Just needs a couple tweaks!

compiler/qsc_rir/src/rir.rs
cesarzc: Good catch! I'll update the instructions that need a variable field.

compiler/qsc_rir/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![warn(clippy::mod_module_files, clippy::pedantic, clippy::unwrap_used)]
```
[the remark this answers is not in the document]
cesarzc: Updated.

compiler/qsc_rir/src/rir.rs
cesarzc: Updated.

swernli (approved):

PR #1293, opened 2024-03-21
Avoid cyclic override for intrinsics

swernli:
This change updates the implementation of the internal `AND` utility to call QIR intrinsics directly, which avoids a cycle in the call graph and removes the need for overrides in the Runtime Capabilities Analysis.


compiler/qsc_rca/tests/overrides.rs
```diff
@@ -104,14 +96,12 @@ fn check_rca_for_controlled_h() {
     let package_store_compute_properties = compilation_context.get_compute_properties();
     check_last_statement_compute_properties(
```
swernli, line 97: These tests are technically no longer of overrides, but I do think it would be useful to keep them (and add a few for the properties when compiling with `@Config(Base)`), but could use some help deciding where to put them.

cesarzc: I created some tests in callables.rs that replace these ones. The ones that I added are a bit more extensive since they check the compute properties of the whole callable and expand verification to callables for which the controlled specializations differ between `base` and `unrestricted`.

cesarzc (approved):

minestarks (approved):

PR #1303, opened 2024-03-24
Formatter: Remove Unauthoritative Newlines

ScottCarda-MS:
Removes unauthoritative newlines from the formatter, ensuring that all newlines are strictly enforced between either a single style or one of several well-defined styles.
This PR also contains all the formatted Samples and Libraries Q# files.


library/src/tests/resources/add_le.qs
```diff
@@ -25,12 +26,9 @@ namespace Test {
...
 
-                Fact(xActual == xValue,
-                   $"{name}: Incorrect x={xActual}, expected={xValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
-                Fact(yActual == yValue,
-                   $"{name}: Incorrect y={yActual}, expected={yValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
-                Fact(zActual == zExpected,
-                   $"{name}: Incorrect z={zActual}, expected={zExpected}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
+                Fact(xActual == xValue, $"{name}: Incorrect x={xActual}, expected={xValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
+                Fact(yActual == yValue, $"{name}: Incorrect y={yActual}, expected={yValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
+                Fact(zActual == zExpected, $"{name}: Incorrect z={zActual}, expected={zExpected}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}.");
```
DmitryVasilevsky, line 31: This line is fairly long (~160 characters). Is there a limit after which we try to split long lines?

library/core/qir.qs
```diff
@@ -15,7 +15,7 @@ namespace QIR.Runtime {
             fail "Cannot allocate qubit array with a negative length";
         }
         mutable qs = [];
-        for _ in 0..size-1 {
+        for _ in 0..size - 1 {
```
DmitryVasilevsky, line 18: So, looks like you decided to not put spaces around the ".."
The expression now looks like you are subtracting 1 from the range (for whatever meaning one can put into such an interpretation) rather than subtracting 1 from the upper bound.
Anyway, maybe it's just my taste. If majority of stakeholders decided that this is the right way - I do not object, just close this comment.

library/std/canon.qs
```diff
@@ -287,9 +284,7 @@ namespace Microsoft.Quantum.Canon {
     /// X(q);
     /// ```
     operation ApplyP(pauli : Pauli, target : Qubit) : Unit is Adj + Ctl {
-        if   pauli == PauliX { X(target); }
```
DmitryVasilevsky, line 290: Is there a way to override formatting for a block if one feels strongly about a particular formatting in a particular place? Something like #pragma(DoNotAutoformat) <- not an actual directive, just made it up.

library/std/math.qs
```diff
@@ -803,10 +797,7 @@ namespace Microsoft.Quantum.Math {
         Fact(n >= 0, "The factorial is not defined for negative inputs.");
         Fact(n <= 20, "The largest factorial that can be stored as an Int is 20!. Use FactorialL or ApproximateFactorial.");
 
-        [1, 1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800,
-        39916800, 479001600, 6227020800, 87178291200, 1307674368000,
-        20922789888000, 355687428096000, 6402373705728000,
-        121645100408832000, 2432902008176640000][n]
+        [1, 1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800, 39916800, 479001600, 6227020800, 87178291200, 1307674368000, 20922789888000, 355687428096000, 6402373705728000, 121645100408832000, 2432902008176640000][n]
```
DmitryVasilevsky, line 800: This string is very long with plenty of opportunities to split it. Are you sure it's the best way to format it?

library/src/tests/resources/add_le.qs
ScottCarda-MS: Line length limits are one of the things this formatter can't do. Since we are removing user's ability to put newlines where they want with this PR (that's the point of this PR), some lines may become longer than desired. In these cases, users can either write their Q# in a way that avoids the long lines, or do not format that section of their code.

library/std/math.qs
ScottCarda-MS: The formatter as it is cannot reason about the length of lines, so there is no way for it to know that this line needs to be split or where to split it.

library/std/canon.qs
ScottCarda-MS: You can turn the "save on format" setting off specifically for Q#. Then you would only manually either format the whole file, in which case you could adjust the formatting after words, or you could format only the sections of the code you want via highlighting. Finally, if you are not wanting to use the formatter at all, there is a setting to disable the Q# formatter.

library/src/tests/resources/compare.qs
```diff
@@ -4,49 +4,50 @@ namespace Test {
...
+        classicalComparator : (Int, Int) -> Bool
+    ) : Unit {
+
         for n in 1..bitwidth {
             use qs = Qubit[n];
             use t = Qubit();
 
-            for a in 0 .. 2^n+1 { // We want b to have more bits sometimes...
-                for b in 0 .. 2^n-1 {
+            for a in 0..2 ^ n + 1 {
```
DmitryVasilevsky, line 17: Have you considered not putting spaces around ^ (as it seems to be the highest priority operator)?

DmitryVasilevsky (approved):
Approving library part with comments.

library/src/tests/resources/add_le.qs
DmitryVasilevsky: But we are formatting entire library now as a policy. Long lines don't look good.

library/std/math.qs
DmitryVasilevsky: What about splitting of a similar array later in this file?

library/std/canon.qs
ScottCarda-MS: There's currently no built-in way to format multiple files at once like that. You'd have to go to each individual file anyway to format each of them.

library/src/tests/resources/add_le.qs
minestarks: But hang on... this particular line _can_ be improved by formatting the argument list vertically, right? I think we should just do that.

```qs
Fact(
    xActual == xValue,
    $"{name}: Incorrect x={xActual}, expected={xValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}."
);
Fact(
    yActual == yValue,
    $"{name}: Incorrect y={yActual}, expected={yValue}. |x|={xLen}, |y|={yLen}, |z|={zLen}, x={xValue}, y={yValue}."
);
```

The only "problem" here is that the formatter won't do this automatically for you based on line length - you have to add the first newline manually. But the formatter respects your choice after that.

library/std/math.qs
```diff
@@ -102,42 +101,36 @@ namespace Microsoft.Quantum.Math {
...
 
     /// # Summary
     /// Returns -1, 0 or +1 that indicates the sign of a number.
-    function SignL (a : BigInt) : Int {
-        if   (a < 0L) { -1 }
-        elif (a > 0L) { +1 }
-        else          {  0 }
+    function SignL(a : BigInt) : Int {
+        if (a < 0L) { -1 } elif (a > 0L) { + 1 } else { 0 }
     }
```
minestarks, line 118: For these, I feel like "regular" block formatting with newlines would look better.

Though either option is an improvement over the original formatting.

format_all.ps1
minestarks, line 1: We've sort of committed to using either Python or JavaScript for scripts in the repo (to minimize prerequisites on all supported OSes). Could you convert to one of those two?

samples/algorithms/BernsteinVazirani.qs
```diff
@@ -135,20 +135,19 @@ namespace Sample {
...
+    ) : Unit {
         // `xRegister` muts have enough qubits to represent the integer.
         let requiredBits = BitSizeI(bitStringAsInt);
         let availableQubits = Length(xRegister);
         Fact(
             availableQubits >= requiredBits,
-            $"Integer value {bitStringAsInt} requires {requiredBits} bits to " +
-            $"be represented but the quantum register only has " +
-            $"{availableQubits} qubits");
+            $"Integer value {bitStringAsInt} requires {requiredBits} bits to " + $"be represented but the quantum register only has " + $"{availableQubits} qubits"
```
minestarks, line 145: Probably best to take the `+` operators out of the string here if it's going to be all on one line anyway....

minestarks (commented):
Very satisfying to see all the ad-hoc formatting standardized in the libraries! Definite improvement IMO.

I don't have an opinion either way on the subjective calls (e.g. spaces around certain binary operators). 

Just a couple of requests, will sign off when those are addressed.

library/src/tests/resources/compare.qs
ScottCarda-MS: I think the consensus around this has shifted to not wanting the spaces (I also think it looks better without spaces), so I will add a rule to the formatter to special-case the `^` to not have spaces around it.

library/std/canon.qs
ScottCarda-MS: This seems to me like an ask for an additional formatting feature, unless you disagree and believe such a change is integral to the other changes in this PR. We could consider adding this in the future if the demand for it is strong enough. Would you add a GitHub issue for this please?

format_all.ps1
ScottCarda-MS: I should probably save such a script for the upcoming integration with the pipeline anyway. I will remove this script for this PR, and reimplement it with the a more appropriate language for that future PR 😊

library/std/math.qs
ScottCarda-MS: Whether or not all the elements of a sequence like this are separated by a space or by newlines is determined by whether or not there is a newline present in the whitespace between the opening delimiter and the fist item. We could change this array to be formatted vertically by adding a newline between the `[` and the first `1`.

library/src/tests/resources/add_le.qs
ScottCarda-MS: That does look better like that. Thanks Mine, I've applied your suggestion.

library/std/math.qs
ScottCarda-MS: This does look much better formatted vertically (in my opinion). I've applied that change here.

library/std/math.qs
ScottCarda-MS: done

samples/algorithms/BernsteinVazirani.qs
ScottCarda-MS: done

minestarks (approved):

format_all.ps1
ScottCarda-MS: removed

library/src/tests/resources/compare.qs
ScottCarda-MS: removed spacing around the `^` operator

PR #1305, opened 2024-03-25
Support Base Profile RIR to QIR

swernli:
This adds support for converting the subset of RIR that corresponds to base profile to QIR. Initial unit tests are included, along with test infrastructure for creating simple RIR programs.


compiler/qsc_codegen/src/qir/rir_utils.rs
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) Microsoft Corporation.
```
cesarzc, line 1: nit: this module could be renamed to `rir_building.rs` or `rir_builder.rs`. I think it describes a bit better what it is used for.

cesarzc (approved):
Just commented a small nit.

minestarks (approved):

PR #1310, opened 2024-03-25
Fixes Magic Getting Erased by Format in Notebooks

### LINKED ISSUE #1308: Formatting Q# cell erases `%%qsharp` magic
Repro steps:

1. In VS Code, run "Create: new Jupyter notebook" command to create a notebook.
2. Type this in a code cell:
```
%%qsharp

operation Foo() : Unit {
    
}
```
3. Hit Alt-Shift-F to format the cell.

Expected:

* Code is formatted and the `%%qsharp` cell remains at the top

Actual:

* `%%qsharp` cell is gone.

I'm fairly sure the code below is the issue. We erase the `%%qsharp` magic and replace it with whitespace when communicating with the language service, and the formatter of course gets rid of that "whitespace"... Maybe we can replace it with a comment or something instead.

https://github.com/microsoft/qsharp/blob/7c2296061bf46a3ef6203f1fdfa0fc09857033c6/vscode/src/notebook.ts#L200-L203

ScottCarda-MS:
This was caused by two different bugs that are addressed in this PR:
 - When the input Q# uses `\r\n` line endings, the raw token lexer would only consider the `\n` line ending, resulting in comment tokens with trailing `\r` characters in them.
 - In the notebooks, when there were newlines prior to the magic command, the offsets were not mapped correctly.

In order to preserve the magic command in the notebooks from getting removed as leading whitespace in a file (a formatter rule), the magic command is replaced by a comment instead of whitespace.

Fixes #1308

### DIFF
--- a/compiler/qsc_formatter/src/formatter/tests.rs
+++ b/compiler/qsc_formatter/src/formatter/tests.rs
@@ -965,6 +965,43 @@ fn format_with_crlf() {
 }
 
 #[test]
+fn format_does_not_edit_magic_comment() {
+    let content = indoc! {"\r\n\r\n    //qsharp    \r\n\r\noperation Foo() : Unit {\r\n\r\n}\r\n"};
+    check_edits(
+        content,
+        &expect![[r#"
+            [
+                TextEdit {
+                    new_text: "",
+                    span: Span {
+                        lo: 0,
+                        hi: 8,
+                    },
+                },
+                TextEdit {
+                    new_text: "//qsharp",
+                    span: Span {
+                        lo: 8,
+                        hi: 20,
+                    },
+                },
+                TextEdit {
+                    new_text: "",
+                    span: Span {
+                        lo: 48,
+                        hi: 52,
+                    },
+                },
+            ]
+        "#]],
+    );
+    check(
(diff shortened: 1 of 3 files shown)


ScottCarda-MS:
> Maybe a dumb question, but looking at the test, does this end up replacing `%%qsharp` with `//qsharp` in notebooks?

No, the text found in the notebook is only changed by the edits the formatter makes. Since there is nothing wrong format-wise with `//qsharp`, no format edits are issues to change that text, and the original `%%qsharp` found in the notebook goes unchanged.

compiler/qsc_formatter/src/formatter/tests.rs
minestarks, line 1: Perhaps another test case with some whitespace around the `//qsharp` so we don't regress this case? It's cricital we don't return an edit to the comment itself, and this case would be pretty easy to miss if we made changes to how we return edits...

minestarks (approved):

compiler/qsc_formatter/src/formatter/tests.rs
ScottCarda-MS: added

PR #1311, opened 2024-03-26
Codegen for basic branching on measurement

swernli:
This change upgrades RIR -> QIR to support basic branching on measurement results via `Jump` and `Branch` instructions. This does not include any support for mutable variables.

### DIFF
--- a/compiler/qsc_codegen/src/qir.rs
+++ b/compiler/qsc_codegen/src/qir.rs
@@ -166,8 +166,8 @@ impl ToQir<String> for rir::Callable {
             return format!(
                 "declare {output_type} @{}({input_type}){}",
                 self.name,
-                if self.name == "__quantum__qis__mz__body" {
-                    // The mz callable is a special case that needs the irreversable attribute.
+                if self.is_measurement {
+                    // Measurement callables are a special case that needs the irreversable attribute.
                     " #1"
                 } else {
                     ""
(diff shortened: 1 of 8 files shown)


compiler/qsc_codegen/src/qir.rs
```diff
@@ -127,10 +162,10 @@ impl ToQir<String> for rir::Callable {
             .collect::<Vec<_>>()
             .join(", ");
         let output_type = ToQir::<String>::to_qir(&self.output_type, program);
-        let signature = format!("{} @{}({})", output_type, self.name, input_type);
         let Some(entry_id) = self.body else {
             return format!(
-                "declare {signature}{}",
+                "declare {output_type} @{}({input_type}){}",
+                self.name,
                 if self.name == "__quantum__qis__mz__body" {
```
cesarzc, line 169: I missed this in the first PR, but would it make sense to include known attributes in RIR so that RIR -> QIR does not need to handle this through name?

cesarzc (approved):
Looks great!
I just left a comment regarding attributes but it doesn't have to be handled as part of this PR.

compiler/qsc_codegen/src/qir.rs
swernli: I think that would be preferable, yes. Already this check is incomplete... I forgot to add `__quantum__qis__mresetz__body`. Right now we only need two attributes, `#0` for the entry point function, and `#1` for any measurement functions. Entry point is already determined by the callable id saved in the program, so if we just had a property on `Callable` that identified measurements, maybe something like `is_measurement`, that would be enough. Does that seem reasonable? If so, I could make that change as part of this PR pretty easily.

swernli: We went with this approach.

idavis (approved):

PR #1318, opened 2024-03-26
Better UX and error display for circuits

minestarks:
This improves the circuit UX and the handling of various known error cases.

## Circuit

image

## Operation circuit

Screenshot 2024-03-26 151611

## Unsupported due to result comparison

Screenshot 2024-03-26 151551

## Runtime failure

Screenshot 2024-03-26 152053

## Compiler error

Screenshot 2024-03-26 152033

## Too much circuit

Screenshot 2024-03-26 152020

## Not enough circuit

Screenshot 2024-03-26 151736


npm/ux/circuit.tsx
```diff
@@ -4,7 +4,48 @@
...
+// render a circuit this big on a mid-grade laptop so we allow it.
+const MAX_OPERATIONS = 10000;
+const MAX_QUBITS = 1000;
+
+/* This component is shared by the Python widget and the VS Code panel */
 export function Circuit(props: { circuit: qviz.Circuit }) {
+  const errorDiv =
+    props.circuit.qubits.length === 0 ? (
+      <div>
+        <p>No circuit to display. The program has not allocated any qubits.</p>
```
cesarzc, line 19:
```suggestion
        <p>No circuit to display. The program does not allocate any qubits.</p>
```

cesarzc: This is super nit but using past tense in conjunction with "any" short-circuited my brain for some reason.

billti (approved):

cesarzc (approved):

npm/ux/circuit.tsx
minestarks: Ahh. I see the phrasing is awkward if you're just generating a circuit for a program. :(

I was thinking of the debugging scenario when I wrote this. If you're stepping through a program during debugging, this is the first thing you see. We don't have any qubits _yet_, but they may appear as the program continues. So present tense doesn't quite work here either.

I'll try to think of a wording that works in both scenarios...

swernli (approved):

minestarks:
Related #1085

PR #1319, opened 2024-03-27
RIR defer measurements and unreachable code passes

swernli:
This change introduces two RIR -> RIR passes: the first that defers measurements within a block and a second that is a wrapper on a set of unreachable code checks. 

Deferring of measurements updates the RIR and preserves relative ordering of all instructions, leaving the block in the format of:

 1. Operations
 2. Measurements and readout
 3. Output recording
 4. Block terminator

For Base Profile programs where the whole program is a single block, this performs whole program measurement deferral. For programs with more complex control flow, this ensures that measurements appear at the end of the block in a way that could eventually support quantum logic separation.

The unreachable code pass does not modify the RIR, instead acting as a way to verify that RIR is well-formed and panicking on any unreachable callables, blocks, or instructions.

This change also refactors some test utilities to support code reuse across tests for QIR codegen and passes.

### DIFF
--- a/compiler/qsc_codegen/src/rir_passes/defer_meas.rs
+++ b/compiler/qsc_codegen/src/rir_passes/defer_meas.rs
@@ -29,7 +29,7 @@ pub fn defer_measurements(program: &mut Program) {
             CallableType::OutputRecording => {
                 output_recording_ids.insert(id);
             }
-            CallableType::Other => {}
+            CallableType::Regular => {}
         }
     }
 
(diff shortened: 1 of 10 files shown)


compiler/qsc_codegen/src/rir_passes/defer_meas.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Defers measurements in each block of a program to the end of that block.
+/// Specifically, this function reorders instructions within a block to follow the given order:
+///
+/// 1. All instructions except for measurements and output recordings.
+/// 2. All measurements.
+/// 3. All output recordings.
+/// 4. Return, branch, and jump instructions (terminators).
+///
+/// Releative ordering within each section is maintained. Note that the scope is limited to each
```
cesarzc, line 18:
```suggestion
/// Relative ordering within each section is maintained. Note that the scope is limited to each
```

cesarzc (approved):
Looks good!

PR #1320, opened 2024-03-27
Mimalloc is vendored instead of fetched

### LINKED ISSUE #1281: Update mimalloc build infra
We should use a submodule which doesn't require cmake, ninja, etc. and only uses a C compiler with vendored sources. We should also remove the WoA64 exclusion and build all native binaries with the same allocator.

idavis:
These changes copy a snapshot of the Mimalloc source from `v2.1.2` into the mimalloc-sys folder so that we no longer need to use cmake to fetch and prepare the source. This also fixes `aarch64` Windows cross-compilation issues caused by cmake generators.

I have addressed all DevSkim feedback and nothing was applicable or actionable. 

I've run the release build and everything compiles. It shows a red X in the checks as I cancelled the second half of the build to prevent the release from going out.

Closes #1281


allocator/mimalloc-sys/mimalloc/include/mimalloc-new-delete.h
```diff
@@ -0,0 +1,66 @@
+/* ----------------------------------------------------------------------------
+Copyright (c) 2018-2020 Microsoft Research, Daan Leijen
+This is free software; you can redistribute it and/or modify it under the
+terms of the MIT license. A copy of the license can be found in the file
+"LICENSE" at the root of this distribution.
```
minestarks, line 5: I know this is Microsoft... but just for the sake of consistency, shall we copy in `LICENSE` into the parent folder as well?

allocator/mimalloc-sys/build.rs
```diff
@@ -52,28 +49,6 @@ fn compile_mimalloc(dst: &Path) {
     //build.opt_level(3);
 
     build.compile("mimalloc");
```
swernli, line 51: Since we get some spurious warning spew in the build when this part compiles, I wonder if it's worth including some kind of "no warn" flag here. I think that can be done with `.warnings(false)` (docs at https://docs.rs/cc/latest/cc/struct.Build.html#method.warnings).

swernli (approved):

minestarks (approved):

allocator/mimalloc-sys/build.rs
idavis: I'd rather wait on https://github.com/rust-lang/cc-rs/pull/1009

swernli: It doesn't work the way I thought it worked anyway, so it wouldn't have helped!

PR #1323, opened 2024-03-28
Remove debug with circuit menu item; more helpful handling for result comparison error

minestarks:
More UX improvements and an attempt to be more helpful with known error conditions.

## "Show circuit" command (or Code Lens)

image

## "Run file and show circuit diagram" command (i.e. running in simulator)

image

## Program that compares measurement results, Target Profile = Unrestricted

image

## Same program, but with "Run file and show circuit diagram" 

image

## Same program, but with Target Profile = Base

image

### DIFF
--- a/npm/ux/circuit.tsx
+++ b/npm/ux/circuit.tsx
@@ -13,6 +13,8 @@ const MAX_QUBITS = 1000;
 
 /* This component is shared by the Python widget and the VS Code panel */
 export function Circuit(props: { circuit: qviz.Circuit }) {
+  const circuitDiv = useRef<HTMLDivElement>(null);
+
   const errorDiv =
     props.circuit.qubits.length === 0 ? (
       <div>
@@ -42,13 +44,12 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
       </div>
     ) : undefined;
 
-  if (errorDiv) {
-    return <div class=".qs-circuit-error">{errorDiv}</div>;
-  }
-
-  const circuitDiv = useRef<HTMLDivElement>(null);
-
   useEffect(() => {
+    if (errorDiv !== undefined) {
+      circuitDiv.current!.innerHTML = "";
+      return;
+    }
+
     qviz.draw(props.circuit, circuitDiv.current!);
 
     // quantum-viz hardcodes the styles in the SVG.
@@ -57,7 +58,12 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
     styleElements?.forEach((tag) => tag.remove());
   }, [props.circuit]);
 
-  return <div class="qs-circuit" ref={circuitDiv}></div>;
+  return (
+    <div>
+      <div class="qs-circuit-error">{errorDiv}</div>
+      <div class="qs-circuit" ref={circuitDiv}></div>
+    </div>
+  );
 }
 
 /* This component is exclusive to the VS Code panel */
--- a/npm/ux/circuit.tsx
+++ b/npm/ux/circuit.tsx
@@ -3,6 +3,7 @@
 
 import * as qviz from "@microsoft/quantum-viz.js/lib";
 import { useEffect, useRef } from "preact/hooks";
+import { CircuitProps } from "./data";
 
 // For perf reasons we set a limit on how many gates/qubits
 // we attempt to render. This is still a lot higher than a human would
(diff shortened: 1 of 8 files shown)


npm/ux/circuit.tsx
```diff
@@ -3,6 +3,7 @@
 
 import * as qviz from "@microsoft/quantum-viz.js/lib";
 import { useEffect, useRef } from "preact/hooks";
+import { CircuitProps } from "./data";
```
billti, line 6:
```suggestion
import { CircuitProps } from "./data.js";
```

npm/ux/circuit.tsx
```diff
@@ -67,31 +68,39 @@ export function Circuit(props: { circuit: qviz.Circuit }) {
...
-              error(s) below.
-              <br />
-            </p>
-            <div dangerouslySetInnerHTML={{ __html: props.errorHtml }}></div>
-          </div>
-        ) : null}
-      </div>
-      <p>{props.subtitle /* target profile */}</p>
+      <div class="qs-circuit-error">{error}</div>
+      <p>{props.targetProfile /* target profile */}</p>
```
billti, line 91: I'm not sure the comment here is adding a lot of clarity :-)

vscode/package.json
```diff
@@ -285,11 +280,11 @@
         "icon": "$(play)"
       },
       {
-        "command": "qsharp-vscode.debugEditorContentsWithCircuit",
-        "title": "Debug Q# File with Circuit Diagram",
+        "command": "qsharp-vscode.runEditorContentsWithCircuit",
+        "title": "Run file and show circuit diagram",
```
billti, line 284: I think it's still good to have Q# in the command text. I usually include that when searching for commands.

vscode/src/debugger/activate.ts
```diff
@@ -90,7 +90,11 @@ function registerCommands(context: vscode.ExtensionContext) {
           shots: 1,
           ...config,
         },
-        options,
+        {
+          // no need to save the file, in fact better not to, since it may cause the document uri to change
+          suppressSaveBeforeStart: true,
```
billti, line 95: Do we need to specify this here? We already set this in the package.json (search for `"debug.saveBeforeStart": "none"`).

vscode/src/circuit.ts
```diff
@@ -219,7 +243,7 @@ function documentHtml(maybeUri: string, range?: IRange) {
   return location;
 }
 
-function escapeHtml(unsafe: string): string {
+export function escapeHtml(unsafe: string): string {
```
billti, line 246: I see in session.ts you now use `import { escapeHtml } from "markdown-it/lib/common/utils";`. Is that usable here also?

vscode/package.json
```diff
@@ -285,11 +280,11 @@
...
         "category": "Debug",
```
billti, line 285: This should probably just be in the `Q#` category now.

vscode/src/circuit.ts
```diff
@@ -108,28 +125,35 @@ export async function showCircuitCommand(
...
 ) {
   let title;
-  let subtitle;
-  if (operation) {
-    title = `${operation.operation} with ${operation.totalNumQubits} input qubits`;
-    subtitle = `Target profile: ${getTargetFriendlyName(targetProfile)} `;
+  let target;
+  if (params?.operation) {
+    title = `${params.operation.operation} with ${params.operation.totalNumQubits} input qubits`;
+    target = `Target profile: ${getTargetFriendlyName(targetProfile)} `;
```
billti, line 140: Looks like target is the same in both branches. Move it out of the if block and make it const.

vscode/src/debugger/session.ts
```diff
@@ -316,7 +317,12 @@ export class QscDebugSession extends LoggingDebugSession {
     }
 
     if (this.config.showCircuit) {
-      await this.showCircuit();
+      const e = error as any;
```
billti, line 320: It's `error` already an any going by line 312? Do you need to declare e and cast here? Or can just use `error` directly.

vscode/src/debugger/session.ts
```diff
@@ -388,12 +394,23 @@ export class QscDebugSession extends LoggingDebugSession {
...
+        }
         if (result.id != StepResultId.Return) {
           await this.endSession(`execution didn't run to completion`, -1);
           return;
         }
-      } catch (e) {
-        await this.endSession(`ending session due to error: ${e}`, 1);
+      } catch (error) {
+        if (this.config.showCircuit) {
+          const e = error as any;
```
billti, line 406: Just use `catch (error: any)` on line 404 and use directly.

vscode/src/debugger/activate.ts
```diff
@@ -55,13 +55,17 @@ function registerCommands(context: vscode.ExtensionContext) {
...
       (resource: vscode.Uri) =>
-        startDebugging(resource, {
-          name: "Debug Q# File",
-          stopOnEntry: true,
-          showCircuit: true,
-        }),
+        startDebugging(
+          resource,
+          {
+            name: "Run file and show circuit diagram",
```
billti, line 63: Does this `name` get used for anything? Does it show up in any UI?

npm/ux/circuit.tsx
minestarks: Let me clarify that for you: This parameter is the Target Profile. :P

vscode/package.json
minestarks: Good point... I was copying from the command below it, but those commands end up with a "Q#" prefix in the Command Palette while this one isn't. I'll update.

image

vscode/src/debugger/activate.ts
minestarks: Yeah, I should have explained this a little better. I'm changing how we pass `programUri` here, to make the command invocation work from the webview. But by doing this, I'm breaking the "save before start" scenario (when the document uri changes after debugging starts). We do turn off "save before start" in the settings, but in theory the user could override that setting. Here, I'm disabling it completely, even if the user tries to opt in, since it's not going to work. Unlikely scenario, but thought I'd cover all bases.

vscode/src/circuit.ts
minestarks: Wow, I had no idea - auto-import must have added that. I might as well use it here too then :)

vscode/package.json
minestarks: I was on the fence about this. It behaves almost exactly the same as "Run Q# File", also in the Debug category. The internals are the same and the output goes to the Debug Console.

On the other hand, it's also a lot like "Run File and Show Histogram", in that it just runs a whole simulation and shows the output... so I can see putting it into the same category as that one.

Even before this change, the similarity of "Debug: Run Q# File" and "Q#: Run File and Show Histogram" always bugged me a little bit, and this one just kind of overlaps with both.

vscode/src/debugger/session.ts
minestarks: No it's `unknown` 😠  . This code is kinda gross in other ways too, I'll clean it up.

vscode/src/debugger/activate.ts
minestarks: I've never seen it show up. Which probably explains why I didn't even have it correct in the first place.

billti (approved):

PR #1324, opened 2024-03-28
Fast inner-loop build watcher

billti:
See the comments in dev.mjs

Basically after initial build with './build.py', simply run 'node dev.mjs' and most changes to Rust, npm, VS Code, or playground can be reloaded in seconds.

I expect this may have issues on Windows (e.g. maybe need to us 'wasm-pack.exe' there), so looking for feedback on if it works and time it takes across platforms, as well as if this is something we should check in.

Note that as part of this I cleaned up the npm imports to use the web version EXCEPT where it's Node specific code. That's because this watcher only rebuilds the web version for speed.

Also, I exported some functions from the build scripts so they could be reused in this script.


dev.mjs
```diff
@@ -0,0 +1,92 @@
...
+  copyWasmToVsCode();
+  copyWasmToPlayground();
+}
+
+subscribe(coreDir, onRustChange);
+subscribe(libsDir, onRustChange);
+subscribe(vslsDir, onRustChange);
+subscribe(wasmDir, onRustChange);
+
+// TODO: Kick off: cd npm; npm run tsc:watch
```
billti, line 86: Oops. Need to push this bit. Will do later.

vscode/build.mjs
```diff
@@ -80,16 +90,20 @@ function buildBundle() {
   );
 }
 
-async function buildWatch() {
+export async function watchVsCode() {
```
minestarks, line 93: When I introduce errors in the VS Code files and save, I don't get the error output in the console. Would it be possible to surface them?

image

dev.mjs
```diff
@@ -0,0 +1,95 @@
...
+  // Build only web and not node targets to half time.
+  const buildDir = join(thisDir, "target", "wasm32", "debug", "web");
+  const result = spawnSync(
+    "wasm-pack",
+    ["build", "--dev", "--no-pack", "--target", "web", "--out-dir", buildDir],
+    { cwd: wasmDir },
+  );
+  console.log("wasm-pack done! ", result.stderr.toString());
+
+  console.log("Copying the wasm-pack ouput files to the npm package");
```
minestarks, line 62: Is it worth skipping this step if the Rust build failed?

image

npm/src/main.ts
```diff
@@ -27,32 +26,38 @@ import {
...
+import type { ProjectLoader } from "../lib/web/qsc_wasm.js";
 
 export { qsharpLibraryUriScheme };
 
 // Only load the Wasm module when first needed, as it may only be used in a Worker,
 // and not in the main thread.
-type Wasm = typeof import("../lib/node/qsc_wasm.cjs");
+
+// Use the types from the web version for... reasons.
+type Wasm = typeof import("../lib/web/qsc_wasm.js");
```
minestarks, line 37: Since we're consistently using the `web/` version now, can this turn into a regular `import` statement and live amongst the other imports above? `import type * as Wasm from "../lib/web/qsc_wasm.js";`  Just for consistency.

dev.mjs
billti: If the files didn't change its harmless, if they did change then you want the new broken ones, else refreshing will still give you the old working ones and the wrong impression that you've changes may have saved and built correctly (if not looking at the output each time)

npm/src/main.ts
billti: That doesn't work. It gives you the type of a namespace, not a value. Feel free to try a cleaner approach and see if something compiles. But I see errors when I try other ideas.

dev.mjs
minestarks, line 1: Silly nitpick but I think `watch.mjs` would be a more descriptive name.

minestarks (approved):
Tried it out, works super fast! Thank you!

vscode/build.mjs
billti: That's because esbuild doesn't do TypeScript type checking. If you look at the npm build script, it does a 'npm run tsc:check' (which doesn't build anything, just checks types) followed by an esbuild (which doesn't type check anything, just bundles).

I can look into running both (not sure if 'tsc:check' can run in a separate 'watch' mode also). But your editor should have TypeScript errors as you make changes that introduce type errors, and this script is intended to be as you're editing/saving code, not run in a pipeline.

dev.mjs
billti: Yeah. I agree.

vscode/build.mjs
billti: I added watchers that do the TypeScript checking on all projects now. See if that works for you.

PR #1325, opened 2024-03-28
Autogenerated docs in the playground

DmitryVasilevsky:
Show documentation autogenerated from doc comments in the playground.
Documentation is generated in the playground by calling generate_docs() exposed by compiler in WASM. Then playground code renders it from md to html and groups various documents by their namespace.

Example:
image


playground/src/docs.tsx
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation.
```
ScottCarda-MS, line 1: image
Parts of the returned string are in code-context blocks with the \`\`\` ```This is code.``` \`\`\`. Is there any way to render that to make it stand out more as code blocks, like perhaps giving them a slightly different background or put boards around the code blocks? I assume this is a css thing, but maybe it is more difficult than I realize. If it is too tricky to do, that's okay also.

playground/src/docs.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const regex = new RegExp("^qsharp.namespace: Microsoft.Quantum.(.+)$", "m");
+
+  for (const doc of docFiles) {
+    const match = regex.exec(doc.metadata); // Parse namespace out of metadata
+    if (match == null) {
+      continue; // Skip items with non-parsable metadata
+    }
+    // The next line contains "Zero-width space" unicode character
+    // to allow line breaks before the period.
+    const newNamespace = "… " + match[1].replace(".", "​.");
```
ScottCarda-MS, line 32: nit: I don't think the "... " is necessary and looks a bit strange to me given that the other sections of the nav don't use this syntax.

ScottCarda-MS (approved):
Looks good! Just added a few optional suggestions.

playground/src/docs.tsx
DmitryVasilevsky: I'll see what I can do, but if it requires customization of the markdown renderer, it's probably not worth it. It gets rendered in <pre> section with the <code> block. But so are other fragments, such as function names.

playground/src/docs.tsx
DmitryVasilevsky: Any suggestions? Keeping 'Microsoft.Quantum." in every line seems like a waste of space.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -25,6 +25,13 @@ import {
 // so use those as the set used by the shared compiler
 type Wasm = typeof import("../../lib/web/qsc_wasm.js");
 
+// This type mirrors DocFile in WASM, which is returned from generate_docs()
+export type DocFile = {
+  filename: string;
+  metadata: string;
+  contents: string;
+};
+
```
minestarks, line 34: This can easily be generated from the WASM crate to get strong typing and to avoid duplicating types across files like this.

You can apply this commit 
https://github.com/microsoft/qsharp/commit/31a3b31d3ebe4ee785dec90073799cb5473a5406

The generated signature of `this.wasm.generate_docs` will now be `IDocFile[]` instead of `any`. That type can be exported and used in the rest of the code.

minestarks (approved):

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: This is interesting! Thanks for the tip, serializable_type! macro is nice!

playground/src/docs.tsx
ScottCarda-MS: I was thinking to just drop the "... ".

playground/src/docs.tsx
DmitryVasilevsky: Please take a look. I updated the style.
image

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: I took your commit. After our discussion and perf testing this looks good!

playground/src/docs.tsx
DmitryVasilevsky: That was the first thing I tried. It looked strange - seeing just Convert and similar names without anything in front didn't associate these names to namespaces for me...

PR #1327, opened 2024-03-28
Distinguish between samples and katas in navigation in the playground

cesarzc:
This change distinguishes between samples and katas in the navigation bar so we can use the same name for a sample and a kata.

It also renames the superposition kata, which is the motivation for the navigation change in the playground.


swernli (approved):

tcNickolas (approved):
I won't be able to build this to test in the playground until tonight, but looks good. (I'm assuming you'll undo the change to index.json, superposition kata is not yet ready for its time in the spotlight :-))

tcNickolas:
I verified just now that it works in the playground!

billti (approved):

PR #1329, opened 2024-03-28
Check Samples and Libraries for Correct Format in Pipeline

ScottCarda-MS:
Expands on using the formatter from the command-line.
 - Allows folders to be given.
 - Adds the `--recursive` flag for specifying if you want the folders to be processed recursively. Without this, if given a folder it will only search the direct children of the folder.
 - Adds the `--write` flag for specifying that the formatter should overwrite unformatted Q# files with the format corrections. Without this flag, the formatter only checks if the Q# files are correctly formatted.
To use the command-line formatter run `cargo run --binqsc_formatter -- <args to formatter>`.

Adds a check to the CI pipeline to run the formatter against the libraries and samples, and stops a build if it finds files that are not formatted correctly.


ScottCarda-MS:
> I think it would be advantageous to have a tighter integration with the project system here. What if, instead of `--recursive`, the formatter queried the project system for the project files and formatted them? And for single file mode, it would just take the name of a file to format? Then, we could orchestrate the CI with a basic script that calls the formatter on each individual file.
> 
> `prettier` doesn't operate this way, because the ecosystem lacks a canonical "project" definition, but `cargo fmt` does. The advantageous side effect here is that we have a clean CI story as well as a project-aware formatter.

I'm open to the idea of having an option for formatting a Q# project, but that doesn't cover the same use-case as the folder option, so I still wouldn't want to remove that option. It is very useful to be able to format a specific folder regardless of it is is associated with a project or even a subfolder of a project. This option allows the user a degree of flexibility that wouldn't be covered by a project option.

compiler/qsc_formatter/src/main.rs
```diff
@@ -1,17 +1,195 @@
...
+    #[arg()]
+    path: PathBuf,
+
+    /// Search for Q# files recursively when a folder is given as PATH.
+    #[arg(short, long, default_value("false"))]
+    recursive: bool,
+
+    /// Overwrite the file contents with the formatted contents.
+    #[arg(short, long, default_value("false"))]
+    write: bool,
```
DmitryVasilevsky, line 27: Is "write" a common option in this case? If not, would you consider something a little more menacing, such as "overwrite"?

ScottCarda-MS: `--write` or `-w` seems to be common for cli options where you want to specify that you are writing to a file. Prettier, for example, uses `--write`. I'm not sure I like `--overwrite` because the shorthand of that would be `-o` which I don't associate with writing. That's just a personal preference.

.github/workflows/ci.yml
```diff
@@ -103,6 +103,23 @@ jobs:
...
+          submodules: "true"
+      - name: Setup rust toolchain
+        uses: ./.github/actions/toolchains/rust
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN_VERSION }}
+          components: ${{ env.RUST_TOOLCHAIN_COMPONENTS }}
+      - name: Check Formatting for Libraries
+        run: cargo run --release --bin qsc_formatter -- ./library/ -r
+      - name: Check Formatting for Samples
+        run: cargo run --release --bin qsc_formatter -- ./samples/ -r
```
minestarks, line 121: IMO this should be a part of `build.py --check` as well, so that it runs by default locally and so that it's consistent with how/when we run `prettier` and `cargo fmt` today.

.github/workflows/ci.yml
```diff
@@ -207,7 +224,7 @@ jobs:
 
   status-check:
     name: Status Check
-    needs: [format, clippy, web-check, build, integration-tests, runBenchmark, runMemoryProfile]
+    needs: [format, clippy, web-check, build, format_qsc, integration-tests, runBenchmark, runMemoryProfile]
```
minestarks, line 227: Nittiest of nits: can you use a dash instead of underscore in the name to match the style of the other job names?

(And as I say this, I realize some of them use camelCase so we're already inconsistent :/ )

.github/workflows/ci.yml
```diff
@@ -103,6 +103,23 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: "true"
+      - name: Setup rust toolchain
+        uses: ./.github/actions/toolchains/rust
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN_VERSION }}
+          components: ${{ env.RUST_TOOLCHAIN_COMPONENTS }}
+      - name: Check Formatting for Libraries
+        run: cargo run --release --bin qsc_formatter -- ./library/ -r
```
minestarks, line 119: What about katas?

.github/workflows/ci.yml
minestarks: Oh and if it fails, can it say "please fix the formatting by running `<exact command I can just copy and paste>`"?

compiler/qsc_formatter/src/main.rs
```diff
@@ -1,17 +1,195 @@
...
+            "{Error}{} files are in need of formatting:",
+            file_walker.changed_files.len()
+        );
+        for f in file_walker.changed_files.iter() {
+            println!("\t{f}");
+        }
+        println!("Run the formatter to correct formatting for the above files.{Reset}");
+        Err("Files are not formatted.".to_string())
+    } else if !file_walker.skipped_files.is_empty() {
+        println!(
```
minestarks, line 174: Is there a good reason we'd fail to read files? Shouldn't the command fail if we had to skip files?

minestarks (changes requested):

compiler/qsc_formatter/src/main.rs
ScottCarda-MS: It didn't seem very ergonomic to me to stop the whole formatting if one of the files encountered couldn't be read from due to permissions.

ScottCarda-MS: Or do you mean the formatter should still run all the way through but return an error code?

.github/workflows/ci.yml
ScottCarda-MS: done

.github/workflows/ci.yml
ScottCarda-MS: I added it to the build.py --check step, but I'm not sure about having it print out a custom message if that step fails. I think it always just tells the user that the subprocess failed.

.github/workflows/ci.yml
ScottCarda-MS: Discussed offline: katas will need a bit more work to integrated them into the formatting checks. This can be done in a future PR.

build.py
```diff
@@ -193,6 +193,20 @@ def use_python_env(folder):
             text=True,
             cwd=root_dir,
         )
+
+    step_start("Runing Q# format check")
```
minestarks, line 197: - I think this code should be contained inside the same `if` statement above, so that it only runs if we're building any Rust code.
- Instead of `step_start` could you simply do a `print()`? Unfortunately `step_start`/`step_end` logging doesn't support nested pairs (the pert timer gets thrown off) - and this is missing a `step_end` anyways

compiler/qsc_formatter/src/main.rs
```diff
@@ -1,17 +1,195 @@
...
+use clap::{crate_version, Parser};
+use formatter::{calculate_format_edits, format_str};
+
+#[derive(Debug, Parser)]
+#[command(name = "fmt", version = crate_version!())]
+#[command(author, about, next_line_help = true)]
+struct Cli {
+    /// Path to the file or folder contain files to be formatted.
+    #[arg()]
+    path: PathBuf,
```
minestarks, line 19: I think `clap` allows `Vec<>` arguments. If it's trivial to pass in multiple paths in the same command, maybe that would be nicer since we can avoid invoking the command twice in the build/ci scripts.

.github/workflows/ci.yml
minestarks: Either way. If it makes the script really awkward then I can live without a custom message :)

build.py
ScottCarda-MS: yes - `print()` instead of `step_start()`, but why would we only want to check the format of the Q# code if we compile Rust?

compiler/qsc_formatter/src/main.rs
ScottCarda-MS: done

.github/workflows/ci.yml
ScottCarda-MS: I tried to add an instructive statement to the situations where we want the user to run the formatter with the `--write` option.

build.py
minestarks: > why would we only want to check the format of the Q# code if we compile Rust?

Warning, subjective answer ahead :)

Because if I'm running `./build.py` with specific flags that skip compiling the Rust bits (which I often do, when working on just the JS/Python bits), it would be an annoying extra step to build & run the Q# formatter.

Almost everything in this file is contained within these `if build_xyz` checks and I think it's good to follow that example, instead of running anything unconditionally. The idea is, if you're working on a specific component, you want to be able to build & check only those components. 

So in which cases _should_ we run the formatter? There's no ideal answer but these would be my expectations:

- if I'm building the library (`build_cli`)
- if I'm building the samples (`build_samples`)
- if I'm building the formatter itself (`build_cli` again)

Some common use cases and my expectations on whether the Q# formatter should run:

`./build.py` - yes (since no arguments means ALL)
`./build.py --npm --vscode` - no
`./build.py --samples` - yes
`./build.py --cli` - yes
`./build.py --wasm --npm --vscode`  - no


So maybe I'm changing my answer slightly: `if build_cli or build_samples` seems like the ideal condition to check.

compiler/qsc_formatter/src/main.rs
```diff
@@ -169,17 +180,72 @@ fn main() -> Result<(), String> {
...
+                "{Skip}Updated {} files. {} files skipped.{Reset}",
+                file_walker.changed_files.len(),
+                file_walker.skipped_files.len()
+            );
+            Err(format!(
+                "Could not read/write from {} files",
+                file_walker.skipped_files.len()
+            ))
+        }
+        (true, true, false) => Ok(()),
```
minestarks, line 209: Love the detailed error handling, but I'd probably move all the success cases to the top to make it more immediately obvious what the success conditions are.

minestarks (approved):
Responded about `./build.py` flags - looks great otherwise, thanks! Please remember to check how this looks in CI one last time before merging

.github/workflows/ci.yml
```diff
@@ -103,6 +103,21 @@ jobs:
       - name: Build and Test
         run: python ./build.py --no-check
 
+  format-qsc:
```
minestarks, line 106: Sorry, noticed one more thing! This is purely about aesthetics again, but this should live just below the other `Format` job so that the CI output looks more organized.

image

ScottCarda-MS: I'm not entirely sure how this all works, but the reason I put it there was because the formatter needs to be built, so I put it after the build steps. Does that not have any impact?

build.py
ScottCarda-MS: The number of possible option combinations is a bit large, so I'll defer to you on this.

ScottCarda-MS: done, `if build_cli or build_samples`

compiler/qsc_formatter/src/main.rs
ScottCarda-MS: done

.github/workflows/ci.yml
minestarks: Oh no these jobs all run in parallel on different agents. I think it's just `cargo run` building your code.

ScottCarda-MS: done

PR #1330, opened 2024-03-28
Parse manifest in Rust instead of Python

orpuente-MS:
This PR regains some rust real-state. It delegates the parsing of the manifest in `pip/qsharp/_qsharp.py/init()` to rust, avoiding duplicate manifest-parsing logic in the project.

This change is needed to unblock #1313


minestarks (changes requested):
I'm afraid this change will interfere with how the notebook language service works (for which I'm realizing there's no integration tests, so nothing in the CI broke).

```python
    # Return the configuration information to provide a hint to the
    # language service through the cell output.
    return Config(target_profile, language_features)
```

This line is critical for the language service - it causes the `qsharp.json` configuration to be written out to the notebook cell output, which the language service will read to configure compiler options. (Can give you a demo to show you what I mean, but you can try by changing `languageFeatures` and re-running `qsharp.init` to see how the squiggles change in the editor).

I think this requirement forces us to parse the qsharp.json at the Python level. Even if we parse it in Rust, we'd need to return the contents of the manifest back to Python... which will be too convoluted to be worth it IMO.

pip/src/interpreter.rs
```diff
@@ -113,7 +110,15 @@ impl Interpreter {
             TargetProfile::Unrestricted => Profile::Unrestricted,
             TargetProfile::Base => Profile::Base,
         };
-        let language_features = language_features.unwrap_or_default();
+        // If no features were passed in as an argument, use the features from the manifest.
+        // this way we prefer the features from the argument over those from the manifest.
+        let language_features: Vec<String> = match (language_features, &manifest_descriptor) {
+            (Some(language_features), _) if !language_features.is_empty() => language_features,
```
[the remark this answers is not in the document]
orpuente-MS: I was just replicating the already existing behavior in the Python file, which is loading the settings from the manifest if `language_features` is an empty list.

image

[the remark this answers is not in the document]
orpuente-MS: However, what you suggest makes more sense than the previous behavior. I'm guessing we want to change the default argument of the init function to `None` instead of the current `[]`.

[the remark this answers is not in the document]
orpuente-MS: Fixed this.

orpuente-MS:
> I'm afraid this change will interfere with how the notebook language service works (for which I'm realizing there's no integration tests, so nothing in the CI broke).
> 
> ```python
>     # Return the configuration information to provide a hint to the
>     # language service through the cell output.
>     return Config(target_profile, language_features)
> ```
> 
> This line is critical for the language service - it causes the `qsharp.json` configuration to be written out to the notebook cell output, which the language service will read to configure compiler options. (Can give you a demo to show you what I mean, but you can try by changing `languageFeatures` and re-running `qsharp.init` to see how the squiggles change in the editor).
> 
> I think this requirement forces us to parse the qsharp.json at the Python level. Even if we parse it in Rust, we'd need to return the contents of the manifest back to Python... which will be too convoluted to be worth it IMO.

I implemented the solution we talked about. Passing the raw manifest around and parsing it when it gets to rust land.

wasm/src/language_service.rs
```diff
@@ -486,11 +498,13 @@ serializable_type! {
...
-        pub languageFeatures: Option<Vec<String>>
+        pub languageFeatures: Option<Vec<String>>,
+        pub manifest: Option<Manifest>,
     },
     r#"export interface INotebookMetadata {
         targetProfile?: "unrestricted" | "base";
         languageFeatures?: "v2-preview-syntax"[];
+        manifest?: string;
     }"#,
     INotebookMetadata
```
minestarks, line 509: The types of `manifest` don't match between the Rust and the TypeScript declaration here, which makes me think there might be some type error happening when this actually gets used.

pip/src/interpreter.rs
```diff
@@ -74,22 +74,19 @@ impl FromPyObject<'_> for PyManifestDescriptor {
...
-            .downcast::<PyDict>()?;
 
-        let language_features = get_dict_opt_list_string(manifest, "features")?;
+        let manifest = get_dict_opt_string(dict, "manifest")?.ok_or(PyException::new_err(
+            "missing key `manifest` in manifest descriptor",
+        ))?;
+
+        let manifest = serde_json::from_str::<Manifest>(&manifest).map_err(|_| {
+            PyErr::new::<PyException, _>(format!(
+                "Error parsing {manifest_dir}/qsharp.json. Manifest should be a valid JSON file."
```
minestarks, line 84: Nit: that dot after the file path becomes part of the URL in the UI, making the link invalid.

```suggestion
                "Error parsing {manifest_dir}/qsharp.json . Manifest should be a valid JSON file."
```

wasm/src/language_service.rs
minestarks: See https://github.com/microsoft/qsharp/blob/e3fa6744610bc253365f0c8372e27721d0e0f8ac/vscode/src/notebook.ts#L166
 
This function is throwing at the wasm-bindgen boundary. I'm guessing it receives a string when it expects a `Manifest` object.

Tested with a notebook and qsharp.json like this:

image

minestarks (changes requested):
Looking good but there seems to be a type error when sending the manifest through the language service.

I think this is the kind of scenario that would benefit from an integration test. LMK if you need help running/authoring those.

wasm/src/language_service.rs
orpuente-MS: Thanks! I fixed the type mismatch. I will reach out for help with the integration test.

orpuente-MS:
> Looking good but there seems to be a type error when sending the manifest through the language service.
> 
> I think this is the kind of scenario that would benefit from an integration test. LMK if you need help running/authoring those.

I fixed the type mismatch. I will add the integration test to #1313 instead of this PR. This is one is ready for review.

minestarks (approved):
LGTM. Thanks!

PR #1332, opened 2024-03-28
Implement defaults for all functions in the `Backend` trait

cesarzc:
This change implements defaults that panic for most of the functions in the backend trait. 

**Note on the motivation for this change**: a default backend trait will be used in partial evaluation to validate that no quantum operations are simulated.

### DIFF
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -14,82 +14,82 @@ pub trait Backend {
     type ResultType;
 
     fn ccx(&mut self, _ctl0: usize, _ctl1: usize, _q: usize) {
-        unimplemented!();
+        unimplemented!("ccx gate");
     }
     fn cx(&mut self, _ctl: usize, _q: usize) {
-        unimplemented!();
+        unimplemented!("cx gate");
     }
     fn cy(&mut self, _ctl: usize, _q: usize) {
-        unimplemented!();
+        unimplemented!("cy gate");
     }
     fn cz(&mut self, _ctl: usize, _q: usize) {
-        unimplemented!();
+        unimplemented!("cz gate");
     }
     fn h(&mut self, _q: usize) {
-        unimplemented!();
+        unimplemented!("h gate");
     }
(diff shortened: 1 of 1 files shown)


compiler/qsc_eval/src/backend.rs
```diff
@@ -13,32 +13,84 @@ use crate::val::Value;
...
-    fn t(&mut self, q: usize);
-    fn x(&mut self, q: usize);
-    fn y(&mut self, q: usize);
-    fn z(&mut self, q: usize);
-    fn qubit_allocate(&mut self) -> usize;
-    fn qubit_release(&mut self, q: usize);
-    fn capture_quantum_state(&mut self) -> (Vec<(BigUint, Complex<f64>)>, usize);
-    fn qubit_is_zero(&mut self, q: usize) -> bool;
+    fn ccx(&mut self, _ctl0: usize, _ctl1: usize, _q: usize) {
+        panic!("ccx is not supported");
```
swernli, line 17: nitpick: these can be `unimplemented!` instead, since that matches the intent here.

cesarzc: Updated.

compiler/qsc_eval/src/backend.rs
```diff
@@ -13,32 +13,84 @@ use crate::val::Value;
...
+        unimplemented!();
+    }
+    fn qubit_release(&mut self, _q: usize) {
+        unimplemented!();
+    }
+    fn capture_quantum_state(&mut self) -> (Vec<(BigUint, Complex<f64>)>, usize) {
+        unimplemented!();
+    }
+    fn qubit_is_zero(&mut self, _q: usize) -> bool {
+        unimplemented!();
```
swernli, line 92: this may need to be overriden by you later to return `true` otherwise qubit release in the evaluator will fail, but you will also probably want qubit allocate and release too, so they all go together.

swernli (approved):
One heads up for the future, otherwise looks good!

orpuente-MS (approved):
Makes sense to me. I think this is a good default behavior for the use cases I have seen.

PR #1334, opened 2024-03-28
Partial evaluation of intrinsic unitary operations with classical arguments

cesarzc:
This change implements the initial infrastructure for partial evaluation. More specifically, it implements support for intrinsic unitary operations with classical arguments.


compiler/qsc_partial_eval/src/tests.rs
```diff
@@ -0,0 +1,377 @@
...
+fn calls_to_rotation_operation_using_variables() {
+    check_rir(
+        indoc! {r#"
+        namespace Test {
+            open Microsoft.Quantum.Math;
+            open QIR.Intrinsic;
+            @EntryPoint()
+            operation Main() : Unit {
+                use q = Qubit();
+                let pi_over_two = PI() / 2.0;
```
cesarzc, line 285: Pick more predictable numbers.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+            .expect("the evaluation context does not have a current scope")
+    }
+}
+
+struct Scope {
+    package_id: PackageId,
+    callable: Option<(LocalItemId, FunctorApp)>,
+    args_runtime_properties: Vec<ValueKind>,
+    env: Env,
+    expression_value_map: FxHashMap<ExprId, Value>,
```
cesarzc, line 646: Rename rir::Value to rir::Operand and Icmp op to condition.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+        }
+
+        // If the expression can be classically evaluated, do it. Otherwise, generate instructions for it.
+        let current_package_id = self.get_current_package_id();
+        let store_expr_id = StoreExprId::from((current_package_id, expr_id));
+        let expr_generator_set = self.compute_properties.get_expr(store_expr_id);
+        let callable_scope = self.eval_context.get_current_scope();
+        let compute_kind = expr_generator_set
+            .generate_application_compute_kind(&callable_scope.args_runtime_properties);
+        if matches!(compute_kind, ComputeKind::Classical) {
```
cesarzc, line 512:
```suggestion
        if compute_kind == ComputeKind::Classical {
```

cesarzc: Cannot use `==` for type.

swernli: Oh, I see. This type and few others would have to impl `PartialEq` and `Eq`, which is doable but not important right now.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/tests.rs
cesarzc: Updated.

compiler/qsc_fir/src/extensions.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub ty: Ty,
+    pub var: Option<LocalVarId>,
+}
+
+pub trait PackageExt {
+    fn derive_callable_input_params(&self, callable: &CallableDecl) -> Vec<InputParam>;
+}
+
+impl PackageExt for Package {
+    fn derive_callable_input_params(&self, callable: &CallableDecl) -> Vec<InputParam> {
```
swernli, line 40: nitpick: this sort of comes down to personal preference, but since this trait is not used to be generic across implementers anywhere and is only expected to be implemented on `Package` you can actually just implement it as a function for that type right here:

```suggestion
impl Package {
    pub fn derive_callable_input_params(&self, callable: &CallableDecl) -> Vec<InputParam> {
```

Rust doesn't require that all `impl` for a type are in the same block, just that they are in the same crate that defines the type. Since you are in the same crate, this change could be done and the trait removed.

cesarzc: Making that change.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+    ) {
+        // Check if the callable is already in the program, and if not add it.
+        let callable_name = callable_decl.name.name.to_string();
+        #[allow(clippy::map_entry)]
+        if !self.callables_map.contains_key(&callable_name) {
+            let callable = self.create_intrinsic_callable(store_item_id, callable_decl);
+            let callable_id = self.assigner.next_callable();
+            self.program.callables.insert(callable_id, callable);
+            self.callables_map.insert(callable_name, callable_id);
+        }
```
swernli, line 261: you can actually avoid the clippy allow if you use `Entry::Vacant` and some intentional ordering of the lines inside an if-let (note you'll need to add a `use` of `std::collections::hash_map::Entry`):

```suggestion
        if let Entry::Vacant(entry) = self.callables_map.entry(callable_name) {
            let callable_id = self.assigner.next_callable();
            entry.insert(callable_id);
            let callable = self.create_intrinsic_callable(store_item_id, callable_decl);
            self.program.callables.insert(callable_id, callable);
        }
```

compiler/qsc_fir/src/extensions.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,710 @@
...
+pub enum Error {
+    EvaluationFailed(qsc_eval::Error),
+}
+
+struct PartialEvaluator<'a> {
+    package_store: &'a PackageStore,
+    compute_properties: &'a PackageStoreComputeProperties,
+    assigner: Assigner,
+    backend: QubitsAndResultsAllocator,
+    callables_map: FxHashMap<String, CallableId>,
```
swernli, line 44: I think if you use `Rc<str>` here instead of `String` you can void some memory allocations and just `.clone()` the value from the FIR, which is faster than using `.to_string()`:

```suggestion
    callables_map: FxHashMap<Rc<str>, CallableId>,
```

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,710 @@
...
+            &mut receiver,
+            &[],
+            StepAction::Continue,
+        );
+    }
+
+    fn generate_expr_call(&mut self, callee_expr_id: ExprId, args_expr_id: ExprId) {
+        let (store_item_id, functor_app, callable_decl) = self.eval_callee_expr(callee_expr_id);
+
+        // We generate instructions differently depending on whether
```
swernli, line 235: whoops, looks like the comment got cut off.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: I see. I opted to allow this clippy lint because I thought the "logical" order of operations is to insert the callable ID to the map until after the callable has been added to the program, but since that conflicted with the borrow checker, I decided to not use `entry`. However, it is not a big deal the order in which these operations happen as long as they happen "at the same time" before we need to check again.

Long way of saying, I'll update it to use your suggestion 😄

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -0,0 +1,710 @@
...
+    ) -> (StoreItemId, FunctorApp, CallableDecl) {
+        let current_package_id = self.get_current_package_id();
+        let store_callee_expr_id = StoreExprId::from((current_package_id, callee_expr_id));
+
+        // Verify that the callee expression is classical.
+        let callable_scope = self.eval_context.get_current_scope();
+        let callee_expr_generator_set = self.compute_properties.get_expr(store_callee_expr_id);
+        let callee_expr_compute_kind = callee_expr_generator_set
+            .generate_application_compute_kind(&callable_scope.args_runtime_properties);
+        assert!(matches!(callee_expr_compute_kind, ComputeKind::Classical));
```
swernli, line 156: nitpick: for this and a couple other asserts in the file, it's worth including a string for the panic message just to help guide investigation; something like `assert!(condition, "compute kind must be classical")`. It's a little redundant for checks as clear as this one, but I think still worth it for future devs (including us!) who might somehow break it.

swernli (approved):

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Added messages to asserts for the benefit of our future selves!

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: It seems like my brain `panicked` in this one 😆 
Updated with a complete comment.

PR #1342, opened 2024-03-30
RIR qubit reindex pass

swernli:
This pass reindexes qubits after measurement or reset, ensuring there is no qubit reuse in the program.

### DIFF
--- a/compiler/qsc_rir/src/passes/reindex_qubits.rs
+++ b/compiler/qsc_rir/src/passes/reindex_qubits.rs
@@ -1,0 +1,177 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#[cfg(test)]
+mod tests;
+
+use rustc_hash::FxHashMap;
+
+use crate::rir::{
+    Block, Callable, CallableId, CallableType, Instruction, Literal, Program, Ty, Value,
+};
+
+/// Reindexes qubits after they have been measured or reset. This ensures there is no qubit reuse in
+/// the program. As part of the pass, reset callables are removed and mresetz calls are replaced with
+/// mz calls.
+/// Note that this pass has several assumptions:
+/// 1. Only one callable has a body, which is the entry point callable.
+/// 2. The entry point callable has a single block.
+/// 3. No dynamic qubits are used.
+/// The pass will panic if the input program violates any of these assumptions.
+pub fn reindex_qubits(program: &mut Program) {
+    validate_assumptions(program);
+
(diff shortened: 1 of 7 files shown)


compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -0,0 +1,177 @@
...
+    // All reset function calls should be removed, so remove them from the callables.
+    let mut callables_to_remove = Vec::new();
+    for (callable_id, callable) in program.callables.iter() {
+        if callable.call_type == CallableType::Reset || Some(callable_id) == mresetz_id {
+            callables_to_remove.push(callable_id);
+        }
+    }
+    for callable_id in callables_to_remove {
+        program.callables.remove(callable_id);
+    }
```
idavis, line 111:
```rust
program.callables.retain(|(callable_id, callable)| {
                callable.call_type != CallableType::Reset && Some(callable_id) != mresetz_id
            });
```

swernli: oh, nice, thanks!

swernli: Darn, I'll have to add an impl for that to `IndexMap` since it's not defined yet. Probably worth it though.

compiler/qsc_rir/src/passes/reindex_qubits/tests.rs
```diff
@@ -0,0 +1,296 @@
...
+};
+
+use super::reindex_qubits;
+
+#[test]
+fn qubit_reindexed_after_reset_removes_reset() {
+    let mut program = Program::new();
+    program.num_qubits = 1;
+    program.callables.insert(CallableId(0), x_decl());
+    program.callables.insert(CallableId(1), reset_decl());
```
cesarzc, line 20: One thing that would be helpful when reading these tests would be to use constants for callable IDs:
 ```rust
const X: usize = 0;
const RESET: usize = 1;
program.callables.insert(X.into(), x_decl());
program.callables.insert(RESET.into(), reset_decl());
 ```
 
 This way a sequence of call instructions would be easier to read:
 ```rust
Instruction::Call(X.into(), vec![Value::Literal(Literal::Qubit(0))], None),
Instruction::Call(RESET.into(), vec![Value::Literal(Literal::Qubit(0))], None),
 ```

swernli: That's a good idea, thanks!

cesarzc (approved):

idavis (approved):

PR #1344, opened 2024-04-01
Move npm package

billti:
As discussed, moving the npm package to a directory named after it.

Note that calling the directory name 'qsharp' for now in anticipation of renaming the npm package from 'qsharp-lang' to 'qsharp' at some point in the near future. That will be a bigger change that needs coordinating with QCOM though, so will do separately.


swernli:
You need to update the CODEOWNERS file, which still lists an npm folder.

billti:
@swernli Everything is still under an `/npm` folder (just a level deeper), so codeowners still works as before, no?

swernli:
Oh, I see that now. I misunderstood and thought the npm folder was changing to be named qsharp. You're right, this should work as-is!

swernli (approved):

cesarzc (approved):

PR #1346, opened 2024-04-01
Add tasks 1.8, 1.9, 1.10 to Superposition Kata

jkingdon-ms:


katas/content/superposition/greenberger_horne_zeilinger/index.md
```diff
@@ -0,0 +1,5 @@
+**Input:** $N$ ($N \ge 1$) qubits in the $|0 \dots 0\rangle$ state.
+
+**Goal:**  Change the state of the qubits to an equal superposition of all basis vectors $\frac{1}{\sqrt{2^N}} \big (|0 \dots 0\rangle + \dots + |1 \dots 1\rangle\big)$. 
+
+> For example, for $N = 2$ the final state should be  $\frac{1}{2} \big (|00\rangle + |01\rangle + |10\rangle + |11\rangle\big)$.
```
tcNickolas, line 5: This doesn't look like GHZ state problem statement - leftover copy-paste issue?

katas/content/superposition/greenberger_horne_zeilinger/Verification.qs
```diff
@@ -0,0 +1,20 @@
...
+        H(qs[0]);
+        
+        for q in Rest(qs) {
+            CNOT(qs[0], q);
+        }
+    }
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        CheckOperationsEquivalenceOnZeroStateWithFeedback(
```
tcNickolas, line 15: You'll want to loop over N here, from 1 to 5 or something like that, and check that the state prepared is correct for all lengths

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,26 @@
+### Solution
+
+The single-qubit GHZ state is the plus state $\frac{1}{\sqrt{2}} \big (|0\rangle + |1\rangle\big)$ that we've discussed in task 1.1. As a reminder, that state is prepared by applying a Hadamard gate.
```
tcNickolas, line 3: The tasks are not numbered in the new interface (you can see examples at https://quantum.microsoft.com/en-us/experience/quantum-katas - they just have titles), so we need to refer to them in some other way. Here, we can probably just say "in the first task", since that doesn't force the learner to count much

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,26 @@
...
+
+The 2-qubit GHZ state is the Bell state $\frac{1}{\sqrt{2}} \big (|00\rangle + |11\rangle\big)$ that we've discussed in task 1.6.
```
tcNickolas, line 5: "in the eponymous task" is probably too fancy? :-) Maybe "in the two previous tasks"?

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,26 @@
...
+Let's use the 2-qubit circuit as a building block to construct the circuit for 3 qubits. First, let's add a third qubit to the above circuit.
+
+Comparing the state prepared by this circuit with the desired end state, we see that they differ only in the third (rightmost) qubit:
+
+$$|\Phi^+\rangle |0\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\color{red}0\rangle\big)$$
+$$|GHZ\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\color{red}1\rangle\big)$$
+
+Applying a controlled NOT operation using the first (leftmost) qubit as the control bit and the third (rightmost) qubit as the target qubit allows us to fix this difference.
+
+Similarly, the following circuit will prepare the GHZ state on four qubits $\frac{1}{\sqrt2} (|0000\rangle + |1111\rangle$
```
tcNickolas, line 19: Let's drop this sentence, since it doesn't make sense without the circuit

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,26 @@
...
+The single-qubit GHZ state is the plus state $\frac{1}{\sqrt{2}} \big (|0\rangle + |1\rangle\big)$ that we've discussed in task 1.1. As a reminder, that state is prepared by applying a Hadamard gate.
+
+The 2-qubit GHZ state is the Bell state $\frac{1}{\sqrt{2}} \big (|00\rangle + |11\rangle\big)$ that we've discussed in task 1.6.
+
+The next one is the 3-qubit GHZ state:
+$$|GHZ\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |111\rangle\big)$$
+
+Let's use the 2-qubit circuit as a building block to construct the circuit for 3 qubits. First, let's add a third qubit to the above circuit.
+
+Comparing the state prepared by this circuit with the desired end state, we see that they differ only in the third (rightmost) qubit:
```
tcNickolas, line 12: We need to rephrase this to account for not having circuits and talking about states instead. How about
```suggestion
Let's use the 2-qubit state as a building block to construct the state for 3 qubits. First, let's add a third qubit to the above state (on the right from the first two qubits).
Comparing this state with the desired end state, we see that they differ only in the third (rightmost) qubit:
```

katas/content/superposition/all_basis_vectors/index.md
```diff
@@ -0,0 +1,5 @@
+**Input:** $N$ ($N \ge 1$) qubits in the $|0 \dots 0\rangle$ state (stored in an array of length $N$).
```
tcNickolas, line 1: I think this task description was swapped with GHZ one, since they're otherwise exactly what they should be

katas/content/superposition/all_basis_vectors/Solution.qs
```diff
@@ -0,0 +1,8 @@
+namespace Kata {
+
```
tcNickolas, line 2: I like to not have empty line between namespace and first operation (unless there are open statements which deserve a line after to separate them from operation - but no empty line after namespace) for the sake of keeping code compact (not a lot of space if you view the modern katas on a laptop screen)
```suggestion
```

katas/content/superposition/all_basis_vectors/Placeholder.qs
```diff
@@ -0,0 +1,8 @@
+namespace Kata {
+
```
tcNickolas, line 2:
```suggestion
```

katas/content/superposition/all_basis_vectors/Verification.qs
```diff
@@ -0,0 +1,17 @@
...
+
+    operation AllBasisVectorsSuperposition_Reference (qs : Qubit[]) : Unit is Adj + Ctl {
+        for q in qs {
+            H(q);
+        }
+    }
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        CheckOperationsEquivalenceOnZeroStateWithFeedback(
```
tcNickolas, line 12: Same for the test here, need to loop over some Ns

katas/content/superposition/all_basis_vectors/solution.md
```diff
@@ -0,0 +1,24 @@
+### Solution
+
+As we've seen in task 1.3, to prepare a superposition of all basis vectors on 2 qubits we need to apply a Hadamard gate to each of the qubits.
```
tcNickolas, line 3: Let's refer to 1.3 by name instead

katas/content/superposition/even_odd/Placeholder.qs
```diff
@@ -0,0 +1,8 @@
+namespace Kata {
+
```
tcNickolas, line 2:
```suggestion
```

katas/content/superposition/even_odd/Solution.qs
```diff
@@ -0,0 +1,15 @@
+namespace Kata {
+    
+    operation EvenOddNumbersSuperposition(qs : Qubit[], isEven : Bool) : Unit is Adj + Ctl {
+        let N = Length(qs);
+
+        for i in 0..N-2 {
```
tcNickolas, line 6:
```suggestion
        for i in 0 .. N-2 {
```

katas/content/superposition/even_odd/solution.md
```diff
@@ -0,0 +1,21 @@
+### Solution
+
+Let’s look at some examples of basis states to illustrate the binary numbering system. 
+
+The 4 basis states on $N = 2$ qubits can be split in two columns, where the left column represents the basis states that form the required superposition state for `isEven = false` and the right column - the basis states that form the required superposition state for `isEven = true`.
```
tcNickolas, line 5: This explanation doesn't make much sense without the images... Maybe we can draw a table, with two columns: even (00 and 10) and odd (01 and 11), and instead of red frames use bold for the first N-1 binary digits (and another table for N=3), and rephrase accordingly

tcNickolas (changes requested):
I took a look and left some suggestions; now that we're getting into more interesting tasks, there is a bit of extra work in the tests and in rewording the solutions to account for not reusing images.

Thank you!

katas/content/superposition/greenberger_horne_zeilinger/solution.md
jkingdon-ms: hah yes let's go with 'in the two previous tasks'

katas/content/superposition/even_odd/Verification.qs
```diff
@@ -0,0 +1,30 @@
...
+        if not isEven {
+            X(qs[N-1]);
+        }
+    }
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        for boolVal in [false, true] {
+            Message($"Testing isEven = {boolVal}...");
+            if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
```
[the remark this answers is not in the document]
jkingdon-ms: decided on testing 1 through 5 qubits, not sure if that's overkill

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,23 @@
+### Solution
+
```
tcNickolas, line 2:
```suggestion
```

katas/content/superposition/greenberger_horne_zeilinger/solution.md
```diff
@@ -0,0 +1,23 @@
...
+The 2-qubit GHZ state is the Bell state $\frac{1}{\sqrt{2}} \big (|00\rangle + |11\rangle\big)$ that we've discussed in the two previous tasks.
+
+The next one is the 3-qubit GHZ state:
+$$|GHZ\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |111\rangle\big)$$
+
+Let's use the 2-qubit state as a building block to construct the state for 3 qubits. First, let's add a third qubit to the above state (on the right from the first two qubits).
+Comparing this state with the desired end state, we see that they differ only in the third (rightmost) qubit:
+
+$$|\Phi^+\rangle |0\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\color{red}0\rangle\big)$$
+$$|GHZ\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\color{red}1\rangle\big)$$
```
tcNickolas, line 14: I didn't leave a comment on this in the first round of review, sorry! I don't think `\color` tag works well in the new interface, it colors the whole end of the formula instead of just the piece that is needed, so we probably want to drop it. Making it bold instead works well
```suggestion
$$|\Phi^+\rangle |0\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\textbf{0}\rangle\big)$$
$$|GHZ\rangle = \frac{1}{\sqrt{2}} \big (|000\rangle + |11\textbf{1}\rangle\big)$$
```

katas/content/superposition/even_odd/solution.md
```diff
@@ -0,0 +1,44 @@
+### Solution
+
```
tcNickolas, line 2: We don't need Solution header here, it's provided by the interface
```suggestion
```

katas/content/superposition/all_basis_vectors/solution.md
```diff
@@ -0,0 +1,24 @@
+### Solution
+
```
tcNickolas, line 2:
```suggestion
```

katas/content/superposition/even_odd/solution.md
```diff
@@ -0,0 +1,44 @@
...
+Let’s look at some examples of basis states to illustrate the binary numbering system. 
+
+The 4 basis states on $N = 2$ qubits can be split in two columns, where the left column represents the basis states that form the required superposition state for `isEven = true` and the right column - the basis states that form the required superposition state for `isEven = false`.
+
+| even | odd |
+| ---- | --- |
+| 00   | 01  |
```
tcNickolas, line 9: Nice tables! I would make first digits bold in the first table, so that the format matches the rest of the tables

tcNickolas (approved):
Looks great, thank you!

Issue #1294, opened 2024-03-21
Adaptive Profile QIR generation


cesarzc:
The work to generate QIR that is compliant with the Adaptive Profile specification can be divided into a few major categories.

# Runtime Capabilities Analysis (RCA)
RCA is the process of determining the capabilities a quantum kernel needs to be able to run a particular program. RCA also identifies program elements that can be pre-computed before execution on a quantum kernel and does not consider these elements when determining the capabilities.
- [x] Detailed analysis for most of the Q# language (#1164).
- [x] Set dynamic qubits value kind as dynamic (#1304).

# Runtime Capabilities Check Pass
Once RCA has identified the capabilities needed to run a particular program, it can be used to check whether the program can run on a particular target, and if not, provide compile-time feedback to the user about why.
- [x] Implement runtime capabilities check pass (#1291).
- [x] Type of the entry expression should be part of the check (#1417).
- [x] Full support for partial eval of closures with captures (#1498)
- [x] Disallow dynamic exponential operations (if possible only when the exponential is dynamic) (#1539)

# Partial Evaluation
Partial evaluation is the process of generating a residual program by evaluating all purely classical computations. It starts with an entry expression for a program in the Flattened Intermediate Representation (FIR), uses RCA to identify purely classical computations and produces a program in the Residualized Intermediate Representation (RIR).
- [x] Design the RIR data structure (#1292).
- [x] Implement partial evaluation.
    - [x] Support custom backend to verify no quantum operations are simulated (#1332).
    - [x] Support for intrinsic unitary operations with classical arguments (#1334).
    - [x] Support for proper qubit and measurement results management, including optimal qubit allocation (#1367).
    - [x] Set qubit and result counts in output program (#1383).
    - [x] Output recording (#1387).
    - [x] Runtime failure for Result literals in output (#1409).
    - [x] Full support for non-intrinsic function/operation calls (#1386).
    - [x] Support for controlled functors (#1452).
    - [x] Support loop unrolling (#1378).
    - [x] Support basic branching based on measurement results. i.e. calls to unitary operations inside branches (#1376).
    - [x] Support for hybrid return expressions (#1421).
    - [x] Support for assignment, array repeat, array in-place concatenation, array in-place update of value at index, and array access to value at index (#1480).
    - [x] Support for dynamic binary operations and dynamic assign binary operations on Booleans and integers (except LHS dynamic Boolean and exponentiation) (#1499).
    - [x] Support LHS dynamic Booleans (#1540).
[...]

Issue #996, opened 2024-01-09
Conditional compilation for individual capabilities


sezna:
We currently support `@Config(Base)` and `@Config(Full)`. But, eventually, we will want to conditionally compile code for individual capabilities. See #882 for additional context.

This syntax could also be used to opt a Q# project in to new language features, like new UDT syntax. Similar to Rust's `#!` or Haskell's language extensions. See #543 for more context on that.

cesarzc:
@sezna if you start working on this, could we sync to talk through some options to make sure we can use it in the multiple scenarios conditional compilation would be useful?

sezna:
@swernli says we've done this, he will verify.

swernli:
Fixed by https://github.com/microsoft/qsharp/pull/1393

PR #1347, opened 2024-04-01
Initial integration of RCA, preview QIR gen, and adaptive profile

idavis:
In order to support adaptive profile selection in VS we have to add `Adaptive` as a compilation attribute in the std lib. This is a workaround until we design a full solution to address #996. This PR adds adaptive as a profile.

Functionality to enable adaptive profile VSCode is hidden behind a config value, as is using the preview QIR code generation from VSCode.

QIR generation for Adaptive/Unrestricted from Python will continue to raise exceptions.

Adaptive is a profile allowed in the Python APIs, but warnings will be issued to users indicating that this functionality is in preview.

The language service will surface RCA errors in VSCode and Python, but not the Notebooks (will be handled in a follow up PR).
Other integration features are being tracked in #1294 and this PR is not expected to have all functionality enabled.

### DIFF
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -476,17 +476,18 @@ impl Interpreter {
         let fir_package = fir_store.get_mut(self.package);
         let stmts = lowerer.lower_and_update_package(fir_package, &unit.hir);
 
-        let caps_errors =
+        let cap_results =
             PassContext::run_fir_passes_on_fir(&fir_store, self.package, self.capabilities);
         // if there are no errors, update the interpreter state
         // with the updated FIR store and lowerer so that we
         // don't have to do this twice.
-        if caps_errors.is_empty() {
+        let Err(caps_errors) = cap_results else {
             self.fir_store = fir_store;
             let graph = lowerer.take_exec_graph();
             self.lowerer = lowerer;
             return Ok((stmts, graph));
-        }
+        };
+
         // if there are errors, convert them to interpreter errors
         // and don't update the lowerer or FIR store.
         let mut errors = Vec::with_capacity(caps_errors.len());
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -23,6 +23,7 @@ pub use qsc_eval::{
     val::Value,
     StepAction, StepResult,
 };
+use qsc_lowerer::{map_fir_package_to_hir, map_hir_package_to_fir};
 
 use crate::{
     error::{self, WithStack},
@@ -46,7 +47,6 @@ use qsc_data_structures::{
 };
 use qsc_eval::{
     backend::{Backend, Chain as BackendChain, SparseSim},
-    debug::{map_fir_package_to_hir, map_hir_package_to_fir},
     output::Receiver,
     val, Env, State, VariableInfo,
 };
@@ -59,7 +59,7 @@ use qsc_frontend::{
     compile::{CompileUnit, PackageStore, RuntimeCapabilityFlags, Source, SourceMap},
     error::WithSource,
 };
-use qsc_passes::PackageType;
+use qsc_passes::{PackageType, PassContext};
 use rustc_hash::FxHashSet;
 use thiserror::Error;
 
@@ -112,7 +112,7 @@ pub struct Interpreter {
     // The FIR store
     fir_store: fir::PackageStore,
     /// FIR lowerer
-    lowerer: qsc_eval::lower::Lowerer,
+    lowerer: qsc_lowerer::Lowerer,
     /// The ID of the current package.
     /// This ID is valid both for the FIR store and the `PackageStore`.
     package: PackageId,
@@ -190,7 +190,7 @@ impl Interpreter {
         for (id, unit) in compiler.package_store() {
             fir_store.insert(
                 map_hir_package_to_fir(id),
-                qsc_eval::lower::Lowerer::new()
+                qsc_lowerer::Lowerer::new()
                     .with_debug(dbg)
                     .lower_package(&unit.package),
             );
@@ -203,7 +203,7 @@ impl Interpreter {
             lines: 0,
             capabilities,
             fir_store,
-            lowerer: qsc_eval::lower::Lowerer::new().with_debug(dbg),
+            lowerer: qsc_lowerer::Lowerer::new().with_debug(dbg),
             env: Env::default(),
             sim: BackendChain::new(
                 SparseSim::new(),
@@ -300,7 +300,7 @@ impl Interpreter {
             .compile_fragments_fail_fast(&label, fragments)
             .map_err(into_errors)?;
 
-        let (_, graph) = self.lower(&increment);
+        let (_, graph) = self.lower(&increment)?;
 
         // Updating the compiler state with the new AST/HIR nodes
         // is not necessary for the interpreter to function, as all
@@ -432,7 +432,7 @@ impl Interpreter {
 
         // `lower` will update the entry expression in the FIR store,
         // and it will always return an empty list of statements.
-        let (_, graph) = self.lower(&increment);
+        let (_, graph) = self.lower(&increment)?;
 
         // The AST and HIR packages in `increment` only contain an entry
         // expression and no statements. The HIR *can* contain items if the entry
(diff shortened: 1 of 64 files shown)


DmitryVasilevsky (approved):
Approving library changes given that this is a temporary solution.

vscode/package.json
```diff
@@ -101,7 +101,8 @@
           "default": "unrestricted",
           "enum": [
             "unrestricted",
-            "base"
+            "base",
+            "adaptive"
           ],
           "enumDescriptions": [
```
swernli, line 107: do you need to add an enum description for adaptive here, even just a placeholder?

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -168,6 +172,33 @@ pub enum Error {
...
+        fir_store.insert(map_hir_package_to_fir(id), package);
+    }
+    fir_store
+}
+
+pub fn run_rca_pass(
+    fir_store: &qsc_fir::fir::PackageStore,
+    package_id: qsc_fir::fir::PackageId,
+    capabilities: RuntimeCapabilityFlags,
+) -> Vec<crate::Error> {
```
swernli, line 191: If this functoion returned a tuple of `(Option<PackageStoreComputeProperties>, Vec<crate::Error>)` or some wrapper of that type, then the callers could return that as well and choose to reuse it when calling into `hir_to_qir` or `fir_to_qir`, which would save having to generate it twice. Right now the code runs `analyze_all` once to check errors and if there are none it runs again to prepare for partial evaluation.

swernli (approved):

compiler/qsc_codegen/src/qir.rs
```diff
@@ -7,11 +7,54 @@ mod instruction_tests;
...
+    fir_to_qir(&fir_store, fir_package_id, capabilities)
+}
+
+pub fn fir_to_qir(
+    fir_store: &qsc_fir::fir::PackageStore,
+    fir_package_id: qsc_fir::fir::PackageId,
+    capabilities: RuntimeCapabilityFlags,
+) -> Result<String, qsc_partial_eval::Error> {
+    let program = get_rir_from_compilation(fir_store, fir_package_id, capabilities)?;
+    Ok(ToQir::<String>::to_qir(&program, &program))
```
swernli, line 45: Right here is where we should call into the RIR passes:

```suggestion
    let mut program = get_rir_from_compilation(fir_store, fir_package_id, capabilities)?;
    check_and_transform(&mut program);
    if capabilities.is_empty() {
        defer_quantum_measurements(&mut program);
    }
    Ok(ToQir::<String>::to_qir(&program, &program))

```

compiler/qsc_codegen/src/qir.rs
```diff
@@ -7,11 +7,59 @@ mod instruction_tests;
...
+use qsc_hir::hir;
+use qsc_lowerer::map_hir_package_to_fir;
+use qsc_partial_eval::partially_evaluate;
 use qsc_rir::{
+    passes::{check_and_transform, defer_quantum_measurements},
     rir::{self, ConditionCode},
     utils::get_all_block_successors,
 };
 
+fn lower_store(package_store: &qsc_frontend::compile::PackageStore) -> qsc_fir::fir::PackageStore {
```
cesarzc, line 20: There are a couple of `lower_store` methods in multiple crates that duplicate this functionality. It would be worth it to implement it in the new `qsc_lowerer` crate and use it from there.

wasm/src/lib.rs
```diff
@@ -109,6 +123,46 @@ fn _get_qir(sources: SourceMap, language_features: LanguageFeatures) -> Result<S
...
+    let caps_results =
+        PassContext::run_fir_passes_on_hir(&package_store, package_id, profile.into());
+    // Ensure it compiles before trying to add it to the store.
+    if caps_results.is_err() {
+        // This should never happen, as the program should be checked for errors before trying to
+        // generate code for it. But just in case, simply report the failure.
+        return Err("Failed to generate QIR".to_string());
+    }
+
+    hir_to_qir(&package_store, package_id, profile.into()).map_err(|e| e.to_string())
```
cesarzc, line 163: I think there might be an opportunity here to re-use the `PackageStoreComputeProperties` returned by RCA here. It seems like `run_fir_passes_on_hir` lowers to FIR, runs RCA and then the capababilities check pass. However, neither FIR or the results of RCA persist outside of this function so the `hir_to_qir` function has to lower to FIR and run RCA again.

I don't think this is a blocking issue, but it is worth considering opportunities where we don't have to repeat transformations and analysis if we don't have to.

compiler/qsc/src/target.rs
```diff
@@ -26,6 +28,7 @@ impl From<Profile> for RuntimeCapabilityFlags {
         match value {
             Profile::Unrestricted => Self::all(),
             Profile::Base => Self::empty(),
+            Profile::Adaptive => Self::ForwardBranching,
```
cesarzc, line 31: For adaptive + reset + integer computations will we have another profile?

idavis: @cesarzc Yes, this is just a placeholder.

cesarzc (approved):

compiler/qsc_codegen/src/qir.rs
idavis: It can't be implemented in the lowerer crate due to extra deps. This is bigger refactoring that we can look at later.

wasm/src/lib.rs
idavis: @cesarzc Good catch, updating.

wasm/src/lib.rs
```diff
@@ -78,7 +79,20 @@ pub fn get_qir(
 ) -> Result<String, String> {
     let language_features = LanguageFeatures::from_iter(language_features);
     let sources = get_source_map(sources, &None);
-    _get_qir(sources, language_features)
+    if language_features.contains(LanguageFeatures::PreviewQirGen) {
+        // the use of AdaptiveProfileQirGen is currently being used to
+        // determine what profile to target as the get_qir interface doesn't
+        // have a way to specify the profile to target and the old codegen
+        // is hardcoded to use the Base profile.
```
minestarks, line 86: I'm not a fan of having logic in the WASM layer. Can we just plumb `TargetProfile` through from TypeScript (`compiler.ts`)? 

`targetProfile` can be added to `ProgramConfig` as an optional property, I think that's perfectly appropriate and backward compatible (I can take care of updating `getCircuit` to be consistent later).

https://github.com/microsoft/qsharp/blob/e3fa6744610bc253365f0c8372e27721d0e0f8ac/npm/qsharp/src/compiler/compiler.ts#L92-L97

wasm/src/lib.rs
```diff
@@ -109,6 +123,53 @@ fn _get_qir(sources: SourceMap, language_features: LanguageFeatures) -> Result<S
...
+            Some(compute_properties),
+        )
+        .map_err(|e| e.to_string()),
+        Err(_) => {
+            // This should never happen, as the program should be checked for errors before trying to
+            // generate code for it. But just in case, simply report the failure.
+            Err("Failed to generate QIR".to_string())
+        }
+    }
+}
```
minestarks, line 171: Again, too much behavior for what's supposed to be a thin interop layer (the wasm crate). Shouldn't all this code live in the `qsc_codegen` crate? Surely we'll want to share this code from Python too.

compiler/qsc/src/interpret.rs
```diff
@@ -454,13 +454,57 @@ impl Interpreter {
     fn lower(
         &mut self,
         unit_addition: &qsc_frontend::incremental::Increment,
-    ) -> (Vec<StmtId>, Vec<ExecGraphNode>) {
+    ) -> core::result::Result<(Vec<StmtId>, Vec<ExecGraphNode>), Vec<Error>> {
```
minestarks, line 457: Nit: I don't speak enough Rust to understand why, but the other functions in this file seem to use `std::result::Result` for return types. Are these the same type?

compiler/qsc/src/interpret.rs
```diff
@@ -454,13 +454,57 @@ impl Interpreter {
...
-        )
+        ))
+    }
+
+    fn run_fir_passes(
+        &mut self,
+        unit: &qsc_frontend::incremental::Increment,
+    ) -> core::result::Result<(Vec<StmtId>, Vec<ExecGraphNode>), Vec<Error>> {
+        let mut fir_store = self.fir_store.clone();
+        let mut lowerer = self.lowerer.clone();
```
minestarks, line 474: So we skip updating `self.fir_store` if there are pass errors? I have a feeling this will get us into an inconsistent state but I need to think about it to come up with a repro.

compiler/qsc/src/interpret.rs
```diff
@@ -454,13 +454,57 @@ impl Interpreter {
     fn lower(
```
minestarks, line 454: IMO we really need tests at this level that exercise the FIR pass errors.

compiler/qsc/src/interpret.rs
minestarks: This may be related:

```python
import qsharp
qsharp.init(target_profile=qsharp.TargetProfile.Adaptive)
qsharp.eval("operation Foo() : Int { use q = Qubit(); mutable x = 1; if MResetZ(q) == One { set x = 2; } x }")
qsharp.eval("Foo()")
```

minestarks: The incremental compiler `qsc/src/incremental.rs` has this "transactional" nature to it, we check the AST and HIR and only update the store if there are no errors. This prevents us from getting into these half-updated states.

Now that lowering is fallible, I have a feeling that these FIR checks need to occur _before_ the AST and HIR is updated, to avoid inconsistencies between the HIR and FIR. But lowering _depends_ on the HIR being updated. Which creates a bit of a chicken-and-egg problem. I'm not sure how we'll address it, but pretty sure it's important.

minestarks (commented):
Not finished, but publishing the first batch of comments

compiler/qsc_data_structures/src/language_features.rs
```diff
@@ -33,6 +35,8 @@ where
         iter.into_iter().fold(LanguageFeatures::empty(), |acc, x| {
             acc | match x.as_ref() {
                 "v2-preview-syntax" => LanguageFeatures::V2PreviewSyntax,
+                "preview-qir-gen" => LanguageFeatures::PreviewQirGen,
+                "adaptive-profile" => LanguageFeatures::AdaptiveProfile,
```
minestarks, line 39: Nothing seems to actually check this value.

Which actually kind of makes sense to me.... because doesn't the ability to pass `TargetProfile::Adaptive` imply that adaptive profile is supported? Is there ever a case where you'd want to pass in `TargetProfile::Adaptive` but NOT have the compiler respect it?

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -32,7 +34,10 @@ pub fn partially_evaluate(
     partial_evaluator.eval()
 }
 
+#[derive(Clone, Debug, Diagnostic, Error)]
 pub enum Error {
+    #[error("partial evaluation error: {0}")]
+    #[diagnostic(code("Qsc.Partial.Eval"))]
```
minestarks, line 40: The error code should probably follow the naming convention used for all the other errors: `Qsc.ErrorCategory.ErrorName`
```suggestion
    #[diagnostic(code("Qsc.PartialEval.EvaluationFailed"))]
```

language_service/src/compilation.rs
```diff
@@ -72,14 +72,34 @@ impl Compilation {
...
-            .map(|lint| WithSource::from_map(&unit.sources, qsc::compile::ErrorKind::Lint(lint)))
-            .collect();
-        errors.append(&mut lints);
-
         let package_id = package_store.insert(unit);
+        let unit = package_store
+            .get(package_id)
+            .expect("expected to find user package");
+
+        if errors.is_empty() && target_profile != Profile::Unrestricted {
```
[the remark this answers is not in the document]
minestarks: I ran your code and opted into the VS Code both features in user settings.

Set the target profile to **QIR: BASE**.

Grover.qs:

image


In the current VS Code playground, Grover.qs is Base Profile compatible.

https://insiders.vscode.dev/+quantum.qsharp-lang-vscode-dev/playground/

image


It's already known that some code will be Base Profile compliant but not Adaptive Profile compliant, so we should probably avoid running Adaptive Profile check if Base is selected.

compiler/qsc/src/interpret.rs
idavis: Added python test for it, will add Rust test shortly.

wasm/src/lib.rs
idavis: This can't be in the codegen crate as it would create a circular dependency

minestarks (approved):

PR #1351, opened 2024-04-02
Debounce Rust change watcher

minestarks:
When you save a Rust file multiple times while compilation is going on, the file watch notifications tend to accumulate, causing the wasm-pack build to run a few times in a row.

On my laptop, each "real" wasm-pack build takes ~10s and each "no-op" build still takes ~2s so they tend to add up :)


billti (approved):
It broke my brain a little trying to think through the event loop, what runs sequentially, and the order of events processed and builds called, but I _think_ this looks right. 😕  Thanks!

minestarks:
I agree it's finicky, though I'm pretty sure this is right -- I'll use this on my machine for a few days to put it through its paces before I merge it.

watch.mjs
```diff
@@ -99,6 +99,25 @@ function runWatcher(dir, name, watchTask = "tsc:watch") {
...
+    dirty = true;
+    setTimeout(() => {
+      if (dirty) {
+        dirty = false;
+        fn();
+      }
+    });
+  };
+}
+
```
billti, line 120: Is something like the below too naive? Seems like it would be clearer to those not steeped in the JS 'task' internals.

```js
// Minor delay to ensure all changes flush to disk before starting a build if 
// saving multiple files close together (e.g. a formatter running over a directory)
const buildDelayMs = 50;

let buildPending = 0; // timeout handle to track if a build is already queued
function onChangeNotification() {
    if (buildPending) return; // Already queued
    buildPending = setTimeout(() => {
        // The build task runs sychronously, so we can clear the timeout handle and
        // run the build knowing that nothing will interleave with those operations
        buildPending = 0;
        build();
    }, buildDelayMs);
}
```

minestarks: I blended our two changes. Your variable names and comments are better and the timeout is a good idea. And I guess my "function that takes a function" approach wasn't popular so I got rid of it ;) 

I stuck with using a `bool` instead of the "handle" in your code - seemed more straightforward to me, and also I get a type error with your code since `setTimeout` doesn't return a number.

I also increased the delay to 100ms. In VS Code, every save action results in two file changes due to the "format on save" feature, and 50ms wasn't long enough to batch those together for me. 100ms did the trick.

minestarks:
Tweaked the code a bit based on @billti 's suggestion - will give it another day of use before I merge

watch.mjs
billti: The `if (buildPending)` check is probably redundant, but doesn't hurt to be cautious. Thanks!

PR #1356, opened 2024-04-03
Use log-depth AND chain to reduce logical depth

msoeken:
Replaces the linear-depth AND chain in `ApplyAsSinglyControlled` with a log-depth version that uses the same number of operations and qubits. This can improve resource estimation and improves visualization.

### DIFF
--- a/library/std/unstable_arithmetic_internal.qs
+++ b/library/std/unstable_arithmetic_internal.qs
@@ -545,15 +545,44 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
             } else {
                 use aux = Qubit[n - 1];
                 within {
-                    ApplyAndAssuming0Target(ctls[0], ctls[1], aux[0]);
-                    for i in 1..n - 2 {
-                        ApplyAndAssuming0Target(aux[i - 1], ctls[i + 1], aux[i]);
-                    }
+                    LogDepthAndChain(ctls, aux);
                 } apply {
-                    Controlled op(aux[n - 2..n - 2], input);
+                    Controlled op([Tail(aux)], input);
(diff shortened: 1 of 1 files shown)


library/std/unstable_arithmetic_internal.qs
```diff
@@ -545,15 +545,44 @@ namespace Microsoft.Quantum.Unstable.Arithmetic {
...
         }
     }
 
+    /// # Summary
+    /// This helper function computes the AND of all control bits in `ctls` into
+    /// the last qubit of `tgts`, using the other qubits in `tgts` as helper
+    /// qubits for the AND of subsets of control bits.  The operation has a
+    /// logarithmic depth of AND gates by aligning them using a balanced binary
+    /// tree.
+    internal operation LogDepthAndChain(ctls : Qubit[], tgts : Qubit[]) : Unit is Adj {
```
DmitryVasilevsky, line 562: This function uses array slicing to create portions of the arrays to pass around. With our current approach to simulation this is fairly expensive. It could pass around original arrays and indexes/lengths instead. In this case the depth would still be log, but the simulation time won't be affected as much when comparing to original version. However, the code will become less pretty. Are we worried about simulation time here? If not, you can go ahead and check this in. I'm approving this PR to unblock.

DmitryVasilevsky (approved):

library/std/unstable_arithmetic_internal.qs
msoeken: Thanks for your comment. I don't expect this operation to be called too often in large scale programs, so we don't need to be too worried about simulation time in this case.

PR #1358, opened 2024-04-04
RIR SSA check pass

swernli:
This pass checks that the RIR is in Single Static Assignment (SSA) form. The check verifies that each variable is assigned exactly once before it is used, and that phi nodes are used correctly.

The change also updates how successors are calculated for branch instructions that have constant conditions, and updates the tests that used constant conditions to use dynamic bool values instead.

### DIFF
--- a/compiler/qsc_rir/src/passes/build_dominator_graph/tests.rs
+++ b/compiler/qsc_rir/src/passes/build_dominator_graph/tests.rs
@@ -10,8 +10,8 @@ use crate::{
     builder::new_program,
     passes::remap_block_ids,
     rir::{
-        Block, BlockId, Callable, CallableId, CallableType, Instruction, Operand, Ty, Variable,
-        VariableId,
+        Block, BlockId, Callable, CallableId, CallableType, Instruction, Operand, Program, Ty,
+        Variable, VariableId,
     },
     utils::build_predecessors_map,
 };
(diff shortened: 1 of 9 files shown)


compiler/qsc_rir/src/passes/build_dominator_graph/tests.rs
```diff
@@ -51,7 +37,8 @@ fn test_dominator_graph_single_block_dominates_itself() {
         .insert(BlockId(0), Block(vec![Instruction::Return]));
 
     remap_block_ids(&mut program);
```
idavis, line 39: Does it make sense to extract theses three calls into a common fn? All of the tests use the same setup.

compiler/qsc_rir/src/passes/build_dominator_graph/tests.rs
```diff
@@ -86,13 +74,38 @@ fn test_dominator_graph_sequential_blocks_dominated_by_predecessor() {
 #[test]
 fn test_dominator_graph_branching_blocks_dominated_by_common_predecessor() {
```
idavis, line 75: Test names don't really need to have test in the name, they are already attributed with `#[test]`.

compiler/qsc_rir/src/passes/ssa_check.rs
```diff
@@ -0,0 +1,230 @@
...
+    doms: &IndexMap<BlockId, BlockId>,
+) {
+    check_phi_nodes(program, preds);
+    let variable_assignments = get_variable_assignments(program);
+    let variable_uses = get_variable_uses(program);
+
+    for (var_id, uses) in variable_uses.iter() {
+        let (def_block_id, def_idx) = variable_assignments
+            .get(var_id)
+            .unwrap_or_else(|| panic!("{var_id:?} is used but not assigned"));
```
idavis, line 29: Would a let...else block be better here?

compiler/qsc_rir/src/passes/ssa_check.rs
```diff
@@ -0,0 +1,230 @@
...
+
+                // Multiple variables
+                Instruction::Call(_, vals, _) => {
+                    for val in vals {
+                        if let Operand::Variable(var) = val {
+                            add_use(var.variable_id, block_id, idx);
+                        }
+                    }
+                }
+                Instruction::Phi(vals, _) => {
```
idavis, line 191: In other parts of the code base we refer to the `Vec<_>` component of phi nodes as args. We should be consistent one way or another.

compiler/qsc_rir/src/passes/ssa_check.rs
swernli: Ah, good call. This went through a couple versions, so I didn't notice I had approached but not reached that pattern.

compiler/qsc_rir/src/passes/build_dominator_graph/tests.rs
swernli: I can do that here and in the other test file as well.

idavis (approved):

PR #1360, opened 2024-04-04
Disallow branching on literal in RIR

swernli:
Since branching on literal values is disallowed by dead code checks and should have been simplified by RIR, this change updates `Instruction::Branch` to only have a `Variable` for its condition rather than an operand.


cesarzc (approved):
I really like these kinds of changes that codify our assumptions into the data structures!

idavis (approved):

PR #1361, opened 2024-04-04
Circuit UX improvements, use simulator by default

minestarks:
- When using the Circuit command, fall back to "trace" mode (i.e. run with simulator) when circuit contains a result comparison, instead of showing an error.
- Zoom to fit larger circuits.
- Other minor UX tweaks to make the overall experience smoother. See GIFs below

<details><summary>Simple circuit</summary>

circuit-entanglement-unrestricted</details>

<details><summary>Zooming to fit larger circuit</summary>

circuit-bell-state-unrestricted-zoom-in-out</details>

<details><summary>Falling back to Trace mode for dynamic circuits</summary>

circuit-qrng-unrestricted-trace-zoom-to-fit</details>

<details><summary>Debugging</summary>

circuit-bell-state-debugging</details>

<details><summary>Debugging with a large, slow to render circuit</summary>

circuit-shor-debugging-rendering</details>

<details><summary>Too large to render</summary>

circuit-shor-unrestricted-trace-too-many-gates</details>

<details><summary>Empty circuit</summary>

circuit-array-no-gates</details>

<details><summary>Switching between target profiles Base and Unrestricted</summary>

circuit-bell-state-unrestricted-base</details>

<details><summary>Base profile compiler error</summary>

circuit-bernstein-base-compiler-error</details>

<details><summary>Runtime failure</summary>

circuit-runtime-fail</details>


samples/notebooks/circuits.ipynb
```diff
@@ -421,30 +421,45 @@
...
-      "\u001b[1;31mQSharpError\u001b[0m: Error: Result comparison is unsupported for this backend\nCall stack:\n    at ResetIfOne in line_0\n\u001b[31mQsc.Eval.ResultComparisonUnsupported\u001b[0m\n\n  \u001b[31m×\u001b[0m runtime error\n\u001b[31m  ╰─▶ \u001b[0mResult comparison is unsupported for this backend\n   ╭─[\u001b[36;1;4mline_0\u001b[0m:5:1]\n \u001b[2m5\u001b[0m │     let r = M(q);\n \u001b[2m6\u001b[0m │     if (r == One) {\n   · \u001b[35;1m             ─┬─\u001b[0m\n   ·               \u001b[35;1m╰── \u001b[35;1mcannot compare to result\u001b[0m\u001b[0m\n \u001b[2m7\u001b[0m │         Message(\"result was One, applying X gate\");\n   ╰────\n"
+      "result was Zero\n",
```
DmitryVasilevsky, line 426: Seems like a drastic change from 'result was One' :)

vscode/src/debugger/activate.ts
```diff
@@ -74,6 +74,14 @@ function registerCommands(context: vscode.ExtensionContext) {
     config: { name: string; [key: string]: any },
     options?: vscode.DebugSessionOptions,
   ) {
+    if (
+      vscode.debug.activeDebugSession !== undefined &&
+      vscode.debug.activeDebugSession.type === "qsharp"
+    ) {
+      // Multiple debug sessions disallowed, to reduce confusion
+      return;
+    }
```
idavis, line 83:
```suggestion
    if (vscode.debug.activeDebugSession?.type === "qsharp") {
      // Multiple debug sessions disallowed, to reduce confusion
      return;
    }
```

vscode/src/circuit.ts
```diff
@@ -46,6 +60,7 @@ export async function showCircuitCommand(
   const compilerTimeout = setTimeout(() => {
     timeout = true;
     sendTelemetryEvent(EventType.CircuitEnd, {
+      simulated: false.toString(),
```
idavis, line 63: Would it make sense to add a comment about how/why simulated starts false and may change in the circuit start/end calls for telemetry?

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -39,7 +39,7 @@ fn empty() {
     );
 
     let circ = interpreter
-        .circuit(CircuitEntryPoint::EntryPoint)
+        .circuit(CircuitEntryPoint::EntryPoint, false)
```
idavis, line 42: I could not find any tests which pass `true` for simulation. Are there test cases we should add?

vscode/src/circuit.ts
```diff
@@ -88,6 +109,7 @@ export async function showCircuitCommand(
 
     if (!timeout) {
       sendTelemetryEvent(EventType.CircuitEnd, {
+        simulated: false.toString(),
```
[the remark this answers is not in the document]
minestarks: Hmmm, this isn't actually correct 😬   - this is in a `catch`. There needs to be some twisty logic to handle the exception and also get the accurate value for `simulated`, and I think I just got lazy. I'll fix it (or if I can't, at least comment)

compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Good point - will add.

vscode/src/debugger/activate.ts
minestarks: I'm so happy to be taking fancy TypeScript syntax suggestions from you 😉

wasm/src/lib.rs
```diff
@@ -155,13 +156,16 @@ pub fn get_circuit(
...
 
     let circuit = interpreter
-        .circuit(match operation {
-            Some(p) => {
-                let o: language_service::OperationInfo = p.into();
-                CircuitEntryPoint::Operation(o.operation)
-            }
-            None => CircuitEntryPoint::EntryPoint,
-        })
+        .circuit(
```
idavis, line 159: nit: move into local var.

vscode/src/debugger/session.ts
```diff
@@ -389,17 +388,15 @@ export class QscDebugSession extends LoggingDebugSession {
           bps,
           this.eventTarget,
         );
-        if (this.config.showCircuit) {
-          this.showCircuit();
-        }
+
+        this.updateCircuit();
```
idavis, line 392: Should this be awaited?

vscode/src/debugger/session.ts
```diff
@@ -819,13 +816,14 @@ export class QscDebugSession extends LoggingDebugSession {
...
-          // the diagram won't get updated with each step even though the circuit
-          // panel is still being shown, which is misleading.
-          this.config.showCircuit = true;
+          if (!this.config.showCircuit) {
+            // Keep updating the circuit for the rest of this session, even if
+            // the Variables scope gets collapsed by the user. If we don't do this,
+            // the diagram won't get updated with each step even though the circuit
+            // panel is still being shown, which is misleading.
+            this.config.showCircuit = true;
+            this.updateCircuit();
```
idavis, line 825: Should this be awaited?

npm/qsharp/ux/data.ts
```diff
@@ -69,7 +69,10 @@ export type CircuitProps = {
   circuit?: CircuitData;
   errorHtml?: string;
   targetProfile: string;
-  simulating: boolean;
+  /** Circut was generated by running the simulator */
```
idavis, line 72:
```suggestion
  /** Circuit was generated by running the simulator */
```

compiler/qsc/src/interpret.rs
```diff
@@ -421,6 +419,36 @@ impl Interpreter {
...
+        sim: &mut impl Backend<ResultType = impl Into<val::Result>>,
+    ) -> InterpretResult {
+        let mut sink = std::io::sink();
+        let mut out = GenericReceiver::new(&mut sink);
+
+        let (package_id, graph) = if let Some(entry_expr) = entry_expr {
+            (self.package, self.compile_entry_expr(&entry_expr)?.into())
+        } else {
+            (self.source_package, self.get_entry_exec_graph()?)
+        };
```
idavis, line 434: Can you please add a comment here about each branch here?

vscode/src/circuit.ts
```diff
@@ -229,7 +403,7 @@ function documentHtml(maybeUri: string, range?: IRange) {
     );
     const fsPath = escapeHtml(uri.fsPath);
     const lineColumn = range
-      ? escapeHtml(`:${range.start.line}:${range.start.character}`)
+      ? escapeHtml(`:${range.start.line + 1}:${range.start.character + 1}`)
```
idavis, line 406: Do we need to pull this from the env? I can't remeber if it is just the debugger that has configurable offsets that we need to make calls for?

idavis (approved):

vscode/src/circuit.ts
minestarks: Good question I hadn't thought about it, but in this case, no. `line`+`character` are coming from our rust debugger which consistently uses 0-indexing. Here we're adding 1 for display purposes, which consistently need to be 1-indexed for the user. So we're good.

compiler/qsc_circuit/src/builder.rs
```diff
@@ -206,7 +206,13 @@ impl Backend for Builder {
                 Some(classical_args)
             },
         ));
-        Some(Ok(Value::unit()))
+
+        match name {
+            // Special case this known intrinsic to match the simulator
+            // behavior, so that our samples will work
+            "BeginEstimateCaching" => Some(Ok(Value::Bool(true))),
```
cesarzc, line 213: If the custom intrinsic is non-Unit and it is not `BeginEstimateCaching`, it seems like building the circuit might break some time later because the evaluator could get `Unit` when in reality it is expecting another kind of value.

If this is the case, I don't think this needs to be addressed in this PR since custom instrinsics outside the standard library are rare, but we should leave a comment or track it if we plan to address it.

cesarzc (approved):

compiler/qsc/src/interpret.rs
```diff
@@ -468,6 +466,38 @@ impl Interpreter {
...
+
+        if self.quantum_seed.is_some() {
+            sim.set_seed(self.quantum_seed);
+        }
+
+        eval(
+            package_id,
+            self.classical_seed,
```
billti, line 491: So this doesn't override the seed that might have been set by line 486 above?

billti: (I guess this is specifically the classical seed, so no)

npm/qsharp/ux/circuit.tsx
```diff
@@ -12,56 +13,157 @@ import { CircuitProps } from "./data.js";
...
+      ) : (
+        <ZoomableCircuit circuit={props.circuit} />
+      )}
+    </div>
+  );
+}
+
+function ZoomableCircuit(props: { circuit: qviz.Circuit }) {
   const circuitDiv = useRef<HTMLDivElement>(null);
+  const [zoomLevel, setZoomLevel] = useState<number | undefined>(undefined);
```
billti, line 40: Shouldn't this just be initialized to `100` and not have `undefined` in the type?

npm/qsharp/ux/circuit.tsx
```diff
@@ -12,56 +13,157 @@ import { CircuitProps } from "./data.js";
...
-    <div>
-      <div class="qs-circuit-error">{errorDiv}</div>
-      <div class="qs-circuit" ref={circuitDiv}></div>
-    </div>
-  );
+function ZoomControl(props: {
+  zoom?: number;
+  onChange: (zoom: number) => void;
+}) {
+  return props.zoom !== undefined ? (
```
billti, line 147: Don't need to check for 'undefined' or make zoom prop optional if initialized to a number as per prior comment.

npm/qsharp/ux/circuit.tsx
```diff
@@ -12,56 +13,157 @@ import { CircuitProps } from "./data.js";
...
+        {rendering
+          ? `Rendering diagram with ${props.circuit.operations.length} gates...`
+          : ""}
+      </div>
+      <div class="qs-circuit" ref={circuitDiv}></div>
+    </div>
+  );
+
+  function updateWidth() {
+    const svg = circuitDiv.current?.getElementsByClassName("qviz")[0];
```
billti, line 85: Can you replace `getElementsByClassName("qvis")[0]` with `querySelector(".qvis")`?

npm/qsharp/ux/circuit.tsx
```diff
@@ -12,56 +13,157 @@ import { CircuitProps } from "./data.js";
...
+    if (svg) {
+      const width = svg.getAttribute("width")!;
+      svg.setAttribute(
+        "style",
+        `max-width: ${width}; width: ${(parseInt(width) * (zoomLevel || 100)) / 100}; height: auto`,
```
billti, line 90: This is interesting/confusing. Initially I thought this might be recursively modifying the value, as you are getting the width on line 87, then setting it here, so on second run it would scale the prior set value, not the original value.

But on closer reading I guess the `svg` element itself has a `width` attribute, but you are then setting `width` via the CSS `style` attribute here - leaving the original `width` attribute untouched.

Is that correct (and deliberate)? Worth some comments if so, as it's subtle.

npm/qsharp/ux/circuit.tsx
```diff
@@ -78,41 +180,29 @@ export function CircuitPanel(props: CircuitProps) {
...
-        ) : null
-      }
       <p>
-        <a href="https://github.com/microsoft/qsharp/wiki/Circuit-Diagrams-from-Q%23-Code">
-          Learn more
-        </a>
+        {
+          props.simulated
+            ? "WARNING: This diagram shows the result of tracing a dynamic circuit, and may change from run to run."
+            : "\xa0" /* nbsp to keep line height consistent */
```
billti, line 193: Does this behave differently from using the common `&nbsp;` in HTML?

vscode/src/circuit.ts
```diff
@@ -17,137 +19,307 @@ import { loadProject } from "./projectSystem";
 import { EventType, UserFlowStatus, sendTelemetryEvent } from "./telemetry";
 import { getRandomGuid } from "./utils";
 import { sendMessageToPanel } from "./webviewPanel";
+import { ProgramConfig } from "../../npm/qsharp/dist/compiler/compiler";
```
billti, line 22: We shouldn't be reaching into the npm code directly via paths like this. We should be importing from qsharp-lang (as per line 6).

compiler/qsc_circuit/src/builder.rs
cesarzc: I think the behavior could be similar to the evaluator one:
```rust
"BeginEstimateCaching" => Some(Ok(Value::Bool(true))),
            "EndEstimateCaching"
            | "AccountForEstimatesInternal"
            | "BeginRepeatEstimatesInternal"
            | "EndRepeatEstimatesInternal" => Some(Ok(Value::unit())),
            _ => Some(Err("Custom intrinsic operations, such as '{name}' are not supported for circuit visualization")),
```

This way we and the user would know an unsupported scenario occurred instead of a seemingly random evaluation error.

npm/qsharp/ux/circuit.tsx
minestarks: `undefined` controls the visibility of the zoom control, so we don't show it before the circuit has finished rendering. But yeah, I think I can still simplify this by using the existing `rendering` flag and removing `undefined`.

npm/qsharp/ux/circuit.tsx
minestarks: OK, got rid of `undefined` , see response to prior comment

npm/qsharp/ux/circuit.tsx
minestarks: Yes, that's exactly it. I think I already have a comment about this somewhere but I'll add a better explanation here.

npm/qsharp/ux/circuit.tsx
minestarks: Well, I really wanted to put `&nbsp;` here, but evaluated JS strings get HTML-escaped in JSX elements 😢  (i.e. the user literally sees "&nbsp;" )

This does the same job, but looks clunky to me:

```jsx
{props.simulated ? (
  <p>
    WARNING: This diagram shows the result of tracing a dynamic circuit,
    and may change from run to run.{" "}
  </p>
) : (
  <p>&nbsp;</p> //  to keep line height consistent
)}
```

vscode/src/circuit.ts
minestarks: Auto-import strikes again! Fixed.

compiler/qsc/src/interpret.rs
minestarks: Correct. A bit clunky, but I'm copying the pattern we've already established in 2 other places in this file we call `set_seed`. This is getting to the point where it needs a refactor again...

compiler/qsc_circuit/src/builder.rs
minestarks: But then I think we're erring on the side of *too* restrictive. Custom intrinsics are okay to render in circuits, as long as they don't return a value that can manipulate classical control flow. 

One problem here is that we just don't know the intended return type of the custom intrinsic (because we don't have access to the compilation). If we did, we could narrow down the error case to "non-unit returning custom intrinsics are not supported").

Custom intrinsics that return anything other than `Unit` will fail Base Profile validation (`Qsc.BaseProfCk.UnsupportedIntrinsic`), so that helps a little bit. If you're using `Unrestricted`, then you're probably targeting the simulator and not implementing your own custom intrinsic.

Maybe we just need a circuits-specific pass to detect unsupported custom intrinsics _before_ we get to the evaluator. Once we're here, we just don't have enough information...

Agree that this should probably be a separate issue/PR.

cesarzc: I think the checks in base profile and the fact that they wouldn't be supported for simulation if someone is using `Unrestricted` seem enough reason to probably not worry about this.

Sorry for the churn!

minestarks: np, thanks for helping me think through the scenario.

minestarks:
Related #1085

Issue #1321, opened 2024-03-27
Use debug printing for expect_test in the parser and other places.


sezna:
We currently use `ToString::to_string` (i.e., `Display`) in our expect test tests for the parser ([see here](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_parse/src/tests.rs#L11)). This is problematic, because this trait is typically used for user-facing representations of data structures. We are forced to choose between a pretty user-facing representation, and a more technically accurate debug-style representation for testing for namespace names in #1312  as a result.

Pulkit1822:
@sezna , I would like to work on this issue.

Issue #493, opened 2023-07-20
Hierarchical Namespaces


bamarsha:
Namespaces don't make sense in Q#. Their only benefit is in organization and avoiding name conflicts, but they don't help you define abstractions or use encapsulation/information hiding. A module system does a better job at organization than namespaces and is also a good unit of abstraction. Namespaces should be deprecated and replaced with modules.

samarsha:
I don't think flat vs. hierarchical is the main problem with Q#'s namespace system. For me, the major issue is that namespaces are *open* instead of *closed*, meaning that a namespace can be defined multiple times, and the items in the namespace are the union of the items in all partial namespaces with the same name. This is related to *internal* being the lowest visibility in Q#, which is too coarse. A *private* visibility is needed but can't be meaningfully defined without a closed unit of abstraction.

I didn't see this mentioned explicitly in the new design, but it seems like because there is a one-to-one mapping from files to namespaces, namespaces are implicitly closed because there can only be one file at a given path. If so, it looks like this is circumvented by explicit namespace definitions which are still supported. I would argue that *private by default* is much better than *internal by default* so while this is an improvement, I don't think it goes far enough.

While making each file implicitly declare its own namespace is a good idea, it looks like this design also has each file be *at most* one namespace, because explicit namespace definitions are treated as a backwards compatibility feature. This seems like an unnecessary limitation. It's another factor that encourages too coarse-grained units of abstraction, because it's easier to put more definitions in the same namespace than to create a new file.

I'm also wondering what is truly *hierarchical* about this design? For example, does `Foo/Bar.qs` in namespace `Foo.Bar` automatically import items from `Foo.qs` in namespace `Foo`? Can `Foo.Bar` import from `Foo.Baz` using a relative path like `super.Baz`? Is there a visibility modifier for items that can be seen by parent or child namespaces but not by sibling namespaces? I didn't see any of these mentioned.

PR #1364, opened 2024-04-06
Use hierarchical namespaces in name resolution

### LINKED ISSUE #1336: Language Service no longer needs to auto-add `open Microsoft.Quantum.Measurement;`
As of https://github.com/microsoft/qsharp/pull/1233, Microsoft.Quantum.Measurement is part of the prelude and no longer needs to be open. However, the Language Service doesn't know this and still inserts open directives for that namespace when auto-completing things. We should include that namespace in the prelude logic there too.

sezna:
_Note: I see the performance degradation in the memory and performance benchmarks, and am looking into that_

This PR introduces a feature called _hierarchical namespaces_, a module-like system that maintains backward compatibility. It is the foundational groundwork that will pave the way for:
1. Naked namespaces (files without namespace declarations)
2. Package publishing and sharing
3. Explicit item imports
4. Exports and re-exports
5. and the rest of the stuff [here](https://github.com/microsoft/qsharp/issues/493#issuecomment-1979722979)

Key changes include:
1. `VecIdent` type introduction: This abstraction over `Vec<Ident>` represents the new namespaces in the AST. Alternate names could be `DotIdent` or `Path`, but those names were already loaded with meaning in our codebase. I'm open to feedback on nomenclature here.
7. `NamespaceId` usage: This replaces our previous flat map-style `Rc<str> -> Res` item tracking.
8. `NamespaceTree`: This tracks the namespace hierarchy separately in low-cost data structure, and allows us to keep our basic `tys` and `terms` maps relatively untouched.
9. Name resolution algorithm update: The algorithm in `resolve.rs` received a heavy facelift. Not only did I try to simplify the existing code, but I also added the required functionality to check candidate partially-opened or re-opened namespaces. 

This PR maintains 100% backward compatibility with the existing resolution engine, preserving features like re-opening namespaces, shadowing semantics, merging namespaces via aliases, and absolute-path opens. It also introduces the ability to reference a function in a different namespace if its parent namespace is opened.

Additional notes:

1.  Katas used a form of shadowing that wasn't tested in our compiler. This PR [adds a test](https://github.com/microsoft/qsharp/pull/1312/files#diff-7a3aa16f18f6d132b9e54a6b6ebd53e6c3b41cea805c8f8db9c456b533a3c96bR2787) for that.
2. Pursuing #1321 could help differentiate user-facing and compiler-engineer-facing string representations of structs, which are increasingly conflated.
3. There's potential for refactoring `ast`, `fir`, and `hir` to use shared generic types, reducing duplicate `Ident`/etc types.
4. Completions functionality remains the same, but it's not yet aware of the ability to partially open namespaces (e.g. `open Microsoft.Quantum`. This will be addressed in future updates.

Also closes #1336

### DIFF
--- a/compiler/qsc/src/interpret/tests.rs
+++ b/compiler/qsc/src/interpret/tests.rs
@@ -1769,11 +1769,14 @@ mod given_interpreter {
             let path = Path {
                 id: NodeId::default(),
                 span: Span::default(),
-                namespace: Some(Box::new(qsc_ast::ast::Ident {
-                    id: NodeId::default(),
-                    span: Span::default(),
-                    name: ns.into(),
-                })),
+                namespace: Some(
+                    std::iter::once(qsc_ast::ast::Ident {
+                        id: NodeId::default(),
+                        span: Span::default(),
+                        name: ns.into(),
+                    })
+                    .collect(),
+                ),
                 name: Box::new(qsc_ast::ast::Ident {
                     id: NodeId::default(),
                     span: Span::default(),
--- a/compiler/qsc_ast/src/ast.rs
+++ b/compiler/qsc_ast/src/ast.rs
@@ -1359,6 +1359,13 @@ impl<'a> IntoIterator for &'a VecIdent {
         self.iter()
     }
 }
+
+impl FromIterator<Ident> for VecIdent {
+    fn from_iter<T: IntoIterator<Item = Ident>>(iter: T) -> Self {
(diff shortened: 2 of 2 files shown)


idavis:
I'm concerned about the potential performance impact of these changes as it is changing the slowest part of the compiler.

The memory benchmark shows 700k of additional memory allocation. Have you run the benchmarks to see what the effect is on the compiler's compilation performance?

PR #1365, opened 2024-04-08
Add AST tab to playground.

idavis:


ScottCarda-MS (approved):
Looks great, thanks!

playground/src/editor.tsx
```diff
@@ -111,6 +117,15 @@ export function Editor(props: {
     setErrors(errList);
   }
 
+  astRef.current = async function updateAst() {
```
minestarks, line 120: Everything else looks ok to be copy-pasted, but I think I'd merge `updateAst` and `updateHir` into a single function (`updateAstOrHir`) to cut down on the noise a bit.

minestarks (approved):
I assume it works :)

billti (approved):

PR #1366, opened 2024-04-08
Update language service when manifest is saved. (Closes #1284)

### LINKED ISSUE #1284: Project isn't updated after editing qsharp.json
**Describe the bug**

The project settings aren't automatically updated when you edit the `qsharp.json` file. You need to go and edit a `.qs` file to trigger an update.

**To Reproduce**

1. Open a Q# project.
2. Try modifying the `qsharp.json` file, for example change the level of some lint as follows:
```json
{
    "lints": [
        {
            "lint": "divisionByZero",
            "level": "error"
        }
    ]
}
```
3. Problem: the lints won't update automatically on your open `.qs` files, you need to edit the file to trigger an update. 

**Expected behavior**

The expected behavior is that when we save the `qsharp.json file`, the Q# project should update automatically.

orpuente-MS:
This PR enables sending document-updates to a given project documents when its manifest is edited and saved.

https://github.com/microsoft/qsharp/assets/156957451/9d528103-97f3-4a53-858b-0b05b4b2b707

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -184,6 +184,36 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     ...registerQSharpNotebookCellUpdateHandlers(languageService),
   );
 
+  // Watch manifest changes and update each document in the same project as the manifest.
+  subscriptions.push(
+    vscode.workspace.onDidSaveTextDocument((manifest) => {
+      if (
+        manifest.languageId === "json" &&
+        manifest.uri.scheme === "file" &&
+        manifest.fileName.endsWith("qsharp.json")
+      ) {
+        const project_folder = manifest.fileName.slice(
+          0,
+          manifest.fileName.length - "qsharp.json".length,
+        );
+        vscode.workspace.textDocuments.forEach((document) => {
+          if (
+            !document.isClosed &&
+            isQsharpDocument(document) &&
+            // Check that the document is on the same project as the manifest.
+            document.fileName.startsWith(project_folder)
+          ) {
+            languageService.updateDocument(
+              document.uri.toString(),
+              document.version,
+              document.getText(),
+            );
+          }
+        });
+      }
+    }),
+  );
+
   // format document
   const isFormattingEnabled = getEnableFormating();
   const formatterHandle = {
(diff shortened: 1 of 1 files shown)


vscode/src/extension.ts
```diff
@@ -184,6 +184,36 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     ...registerQSharpNotebookCellUpdateHandlers(languageService),
   );
 
+  // Watch manifest changes and update each document in the same project as the manifest.
+  subscriptions.push(
+    vscode.workspace.onDidSaveTextDocument((manifest) => {
+      if (
+        manifest.languageId === "json" &&
```
minestarks, line 191: Language check seems redundant here, but harmless I suppose.

vscode/src/extension.ts
```diff
@@ -184,6 +184,36 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     ...registerQSharpNotebookCellUpdateHandlers(languageService),
   );
 
+  // Watch manifest changes and update each document in the same project as the manifest.
+  subscriptions.push(
+    vscode.workspace.onDidSaveTextDocument((manifest) => {
```
minestarks, line 189: Do you want to listen to `onDidDeleteFiles` as well? Similar to this, I think we'd want to trigger an update on all the child .qs files so they can get reparented to single-file-projects.

vscode/src/extension.ts
```diff
@@ -184,6 +184,36 @@ async function activateLanguageService(extensionUri: vscode.Uri) {
     ...registerQSharpNotebookCellUpdateHandlers(languageService),
   );
 
+  // Watch manifest changes and update each document in the same project as the manifest.
+  subscriptions.push(
```
minestarks, line 188: Can this code live inside `registerDocumentUpdateHandlers`? That function is meant to group together all the language service synchronization / update behavior for Q# documents. I think this logically belongs there.

minestarks (approved):
I'm amazed by the simplicity of this fix. I was getting all twisted up imagining what sorts of infrastructure changes we'd need to do in the language service... But this is very clever and to the point. Thanks!

vscode/src/extension.ts
orpuente-MS: Yes, you are right. I just moved it to `registerDocumentUpdateHandlers`.

vscode/src/extension.ts
orpuente-MS: Good observation. I just added it!

minestarks (approved):

PR #1368, opened 2024-04-08
Add the abilty to run compilations against the AST

idavis:


compiler/qsc/src/interpret/tests.rs
```diff
@@ -1617,5 +1630,177 @@ mod given_interpreter {
...
+                language_features,
+            );
+            for e in &errors {
+                eprintln!("{e:?}");
+            }
+            assert!(errors.is_empty(), "compilation failed: {}", errors[0]);
+            let package_id = store.insert(unit);
+
+            let mut interpreter =
+                Interpreter::from(true, store, package_id, capabilities, language_features)
```
swernli, line 1679: does passing `true` here mean the stdlib gets included twice? It was already added above on line 1662, but then `Interpreter::from` which calls `Compiler::from` which inserts it again. I feel like I must be missing something, because I wouldn've thought that would cause compilation errors for duplicate names.

idavis: Great catch, thanks.

compiler/qsc/src/incremental.rs
```diff
@@ -140,6 +179,52 @@ impl Compiler {
...
+    ///
+    /// This method calls an accumulator function with any errors returned
+    /// from each of the stages (parsing, lowering).
+    /// If the accumulator succeeds, compilation continues.
+    /// If the accumulator returns an error, compilation stops and the
+    /// error is returned to the caller.
+    pub fn compile_ast_fragments<F>(
+        &mut self,
+        source_name: &str,
+        source_contents: &str,
```
swernli, line 192: I think it's worth a clarifying comments here just to make sure it's understood that these are use for error reporting and not themselves compiled.

swernli (approved):

compiler/qsc/src/incremental.rs
idavis: Will update when the PR using the API is added.

PR #1372, opened 2024-04-10
Add `check_eq` for `StateDump` in Python

swernli:
This adds a utility to the `StateDump` object in Python to help with writing tests that verify quantum state. The check ignores global phase, so allows for passing in any dictionary where the states differ from the dump by a constant factor, including unnormalized states.


pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,27 @@ def __str__(self) -> str:
     def _repr_html_(self) -> str:
         return self.__data._repr_html_()
 
+    def check_eq(self, state: dict[int, complex]) -> bool:
```
idavis, line 352: use `Dict` and import it from `typing`

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,27 @@ def __str__(self) -> str:
...
+        # Filter out zero states
+        state = {k: v for k, v in state.items() if v != 0}
+        if len(state) != len(self):
+            return False
+        for key in state:
+            if key not in self.__inner:
+                return False
+            if phase is None:
+                phase = self.__inner[key] / state[key]
+            elif phase != self.__inner[key] / state[key]:
```
tcNickolas, line 369: Do we want to do comparisons here exactly, or with some tolerance? I ran into having to do them with tolerance pretty quickly, since real tests compare not hardcoded numbers but results of simulation which is not 100% precise

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,27 @@ def __str__(self) -> str:
     def _repr_html_(self) -> str:
         return self.__data._repr_html_()
```
tcNickolas, line 351: It would be very helpful to have another helper here to convert a dense array of amplitudes into a StateDump. Since Q# code itself cannot take a dictionary as a parameter, parameters of state preparation operations, for example, are arrays of amplitudes, and a tool to convert between an array and a sparse dict would help to use `check_eq` quite a lot

pip/qsharp/_qsharp.py
swernli: I adjusted the comparison to use a tolerance parameter that defaults to 1e-10 (a common value I've seen us use before). That should make it easy for the common case and configurable when needed. Updated the tests to verify tolerance behaves as expected.

pip/qsharp/_qsharp.py
swernli: I think I've got this working as expected, such that a list of real amplitudes can be passed in the same way it is passed into state preparation operations. Here's an example (including using a permissive tolerance):
image

tcNickolas: Why is the list of amplitudes limited to only real numbers? In the specific app where I used this so far it is, but generally (for example, in homework assignments) it can absolutely have complex numbers as well

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,34 @@ def __str__(self) -> str:
...
+        :param state: The state to check against, provided either as a dictionary of state indices to complex amplitudes,
+            or as a list of real amplitudes.
+        :param tolerance: The tolerance for the check. Defaults to 1e-10.
+        """
+        phase = None
+        # Convert a dense list of real amplitudes to a dictionary of state indices to complex amplitudes
+        if isinstance(state, list):
+            state = {i: complex(state[i], 0) for i in range(len(state))}
+        # Filter out zero states
+        state = {k: v for k, v in state.items() if v != 0}
```
tcNickolas, line 368: Should we have some logic here and around `if key not in self.__inner:` branch to account for amplitudes that have absolute values less than tolerance? If there's an amplitude 1e-15 in one of the dictionaries but it's 0 (missing) in others, it should be ok within tolerance, but the comparison will return false

pip/qsharp/_qsharp.py
swernli: Ah, good point. I was trying to mimic the inputs in Q# which takes just `Double` for state prep, but I forgot I can count on Python to coerce for me. I changed the type to `complex` in here and confirmed it accepts arrays of `float` just fine.

tcNickolas: There is another operation in Q# that takes array of Complex, so we need to mimic both I'm afraid :-) But yes, Python is very nice like that

pip/qsharp/_qsharp.py
swernli: Oh, I hadn't thought of that, good catch. I think I only need to update the check here, as that will ensure that keys that zero within tolerance will not be present. Then the only keys below will be those we expect to find in equivalent state dumps.

swernli: I'll add a test case for this and complex states as well.

tcNickolas: Then if self also has keys that are zero within tolerance, the `len(state) != len(self)` is going to fail... If both states are 0.99999|0> + 0.00001|1> (approximately speaking, this one is not normalized), state will have second amplitude dropped so it will have len 1, and self will not so it will have len 2

swernli: I see what you mean... tolerance is only applying to the input but not the state dump. It should apply to both.

pip/qsharp/_qsharp.py
swernli: Added new test cases to cover arrays of complex.

pip/qsharp/_qsharp.py
swernli: Added test case with a small rotation to show how tolerance works on the state dump itself instead of just the input.

tcNickolas: This should be fine - at least I cannot come up with a scenario where it wouldn't off the top of my head... Once this is released, I'll start using it and report back :-)

pip/tests/test_qsharp.py
```diff
@@ -98,6 +98,34 @@ def test_dump_machine() -> None:
     # Check that the state dump correctly supports iteration and membership checks
     for idx in state_dump:
         assert idx in state_dump
+    # Check that the sate dump is correct and equivalence check ignores global phase, allowing passing
```
billti, line 101: Typo on 'state dump'

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,35 @@ def __str__(self) -> str:
...
+            state = {i: state[i] for i in range(len(state))}
+        # Filter out zero states from the state dump and the given state based on tolerance
+        state = {k: v for k, v in state.items() if abs(v) > tolerance}
+        inner_state = {k: v for k, v in self.__inner.items() if abs(v) > tolerance}
+        if len(state) != len(inner_state):
+            return False
+        for key in state:
+            if key not in inner_state:
+                return False
+            if phase is None:
```
billti, line 375: So does `phase` accumlate here as you iterate over the keys? It doesn't seem to accumulate, as it's always either None (from line 363), or whatever the prior key computed on line 376.

I think a few explanatory comments might be good in this code, as it's not super-intuitive to read.

billti (approved):
A couple questions more than comments. If those are just my misunderstanding, then looks good. Thanks!

pip/tests/test_qsharp.py
swernli, line 101:
```suggestion
    # Check that the state dump is correct and equivalence check ignores global phase, allowing passing
```

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,35 @@ def __str__(self) -> str:
...
+                phase = inner_state[key] / state[key]
```
swernli, line 376:
```suggestion
                # Calculate the phase based on the first state pair encountered.
                # Every pair of states after this must have the same phase for the states to be equivalent.
                phase = inner_state[key] / state[key]
```

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,35 @@ def __str__(self) -> str:
...
+            elif abs(phase - inner_state[key] / state[key]) > tolerance:
+                return False
```
swernli, line 378:
```suggestion
                # This pair of states does not have the same phase,
                # within tolerance, so the the equivalence check fails.
                return False
```

pip/qsharp/_qsharp.py
```diff
@@ -349,6 +349,39 @@ def __str__(self) -> str:
...
+        for key in state:
+            if key not in inner_state:
+                return False
+            if phase is None:
+                # Calculate the phase based on the first state pair encountered.
+                # Every pair of states after this must have the same phase for the states to be equivalent.
+                phase = inner_state[key] / state[key]
+            elif abs(phase - inner_state[key] / state[key]) > tolerance:
+                # This pair of states does not have the same phase,
+                # within tolerance, so the the equivalence check fails.
```
swernli, line 381:
```suggestion
                # within tolerance, so the equivalence check fails.
```

pip/qsharp/_qsharp.py
swernli: Added clarifying comments. The phase is calculated once on the first state pair encountered, then confirmed to be the same for every other state pair after that.

PR #1375, opened 2024-04-10
Using new QIR codegen for adaptive profile in Rust/Python

idavis:
Support for new QIR codegen for adaptive profile in Python.

image

### DIFF
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -38,7 +38,7 @@ use qsc_circuit::{
     operations::entry_expr_for_qubit_operation, Builder as CircuitBuilder, Circuit,
     Config as CircuitConfig,
 };
-use qsc_codegen::qir_base::BaseProfSim;
+use qsc_codegen::{qir::fir_to_qir, qir_base::BaseProfSim};
 use qsc_data_structures::{
     functors::FunctorApp,
     language_features::LanguageFeatures,
@@ -97,6 +97,9 @@ pub enum Error {
     #[diagnostic(code("Qsc.Interpret.NotAnOperation"))]
     #[diagnostic(help("provide the name of a callable or a lambda expression"))]
     NotAnOperation,
+    #[error("partial evaluation error")]
+    #[diagnostic(transparent)]
+    PartialEvaluation(#[from] qsc_partial_eval::Error),
 }
 
 /// A Q# interpreter.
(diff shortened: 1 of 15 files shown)


compiler/qsc/src/interpret.rs
```diff
@@ -345,17 +348,50 @@ impl Interpreter {
...
-        let mut sim = BaseProfSim::new();
-        let mut stdout = std::io::sink();
-        let mut out = GenericReceiver::new(&mut stdout);
+            let val = self.run_with_sim(&mut sim, &mut out, expr)??;
 
-        let val = self.run_with_sim(&mut sim, &mut out, expr)??;
+            Ok(sim.finish(&val))
+        } else {
+            // compile the expression which will update the
+            // entry expression in the FIR store.
```
cesarzc, line 364:
```suggestion
            // Compile the expression. This operation will set the expression as
            // the entry-point in the FIR store.
```

compiler/qsc/src/interpret.rs
```diff
@@ -345,17 +348,50 @@ impl Interpreter {
...
+            let graph = self.compile_entry_expr(expr)?;
+            // The above line updated the entry expr in the FIR store, but we need to update
+            // the user package exec graph.
+            let package = self.fir_store.get_mut(self.package);
+            package.entry_exec_graph = graph.into();
+
+            // normal compilation has already been done
+            // analyze the capabilities before generate QIR
```
cesarzc, line 372:
```suggestion
            // Normal compilation has already been done, analyze the runtime capabilities
            // of the program since they are needed to generate QIR.
```

compiler/qsc/src/interpret.rs
```diff
@@ -345,17 +348,50 @@ impl Interpreter {
...
+            // This was already done as part of lowering, but we don't have the results here.
```
cesarzc, line 373: Do we plan to cache the calculated compute properties at some point in the future?

compiler/qsc/src/interpret.rs
```diff
@@ -345,17 +348,50 @@ impl Interpreter {
...
+                &self.fir_store,
+                self.package,
+                self.capabilities,
+            );
 
-        Ok(sim.finish(&val))
+            let Ok(compute_properties) = caps_results else {
+                // This should never happen, as the program should be checked for errors before trying to
+                // generate code for it. But just in case, simply report the failure.
+                return Err(vec![Error::UnsupportedRuntimeCapabilities]);
```
cesarzc, line 383: If this should never happen, should the program panic here instead of returning an error?

cesarzc (approved):

compiler/qsc/src/interpret.rs
```diff
@@ -345,17 +349,43 @@ impl Interpreter {
...
-        Ok(sim.finish(&val))
+            Ok(sim.finish(&val))
+        } else {
+            // Compile the expression. This operation will set the expression as
+            // the entry-point in the FIR store.
+            let (graph, compute_properties) = self.compile_entry_expr(expr)?;
+            // The above line updated the entry expr in the FIR store, but we need to update
+            // the user package exec graph.
+            let package = self.fir_store.get_mut(self.package);
+            package.entry_exec_graph = graph.into();
```
minestarks, line 370: This doesn't seem consistent with how we handle entry expressions elsewhere. It's also causing the `fir_store` to get "out of sync" with the HIR package store - the HIR package store doesn't have an entry point the FIR store does. Though this is probably harmless now, it's just one more thing to keep track of mentally that I'm worried might cause bugs as we refactor or add new features down the line.

I think `fir_to_qir` should take the entry exec graph, just like `eval` does. Then it can be plumbed all the way to the partial evaluator.  Easier to reason about IMO.

Later, if we ever want to tweak how entry expressions are handled, we don't have to worry about the evaluator and partial evaluator separately since they'll be explicit and consistent in their handling.

compiler/qsc/src/interpret.rs
```diff
@@ -97,6 +98,9 @@ pub enum Error {
     #[diagnostic(code("Qsc.Interpret.NotAnOperation"))]
     #[diagnostic(help("provide the name of a callable or a lambda expression"))]
     NotAnOperation,
+    #[error("partial evaluation error")]
+    #[diagnostic(transparent)]
+    PartialEvaluation(#[from] qsc_partial_eval::Error),
```
minestarks, line 103: Does this need to be a `WithSource` so that spans get mapped correctly?

compiler/qsc/src/interpret/tests.rs
```diff
@@ -712,6 +712,66 @@ mod given_interpreter {
             "#]].assert_eq(&res);
         }
 
+        #[test]
+        fn adaptive_qirgen() {
```
minestarks, line 716: Could you add a test case with an error? (Might help validate my other question about whether we need to wrap the errors with `WithSource`)

compiler/qsc/src/interpret.rs
idavis: PartialEval doesn't have the source. It can be updated in the future, but won't work right now.

compiler/qsc/src/interpret/tests.rs
idavis: Added test

minestarks (approved):

minestarks (approved):

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -34,12 +34,17 @@ use rustc_hash::FxHashMap;
 use std::{collections::hash_map::Entry, rc::Rc, result::Result};
 use thiserror::Error;
 
+pub struct EntryRequirements {
+    pub entry_exec_graph: Rc<[ExecGraphNode]>,
+    pub entry_expr_id: fir::StoreExprId,
+}
```
cesarzc, line 40:
```suggestion
pub struct ProgramEntry {
    pub exec_graph: Rc<[ExecGraphNode]>,
    pub expr: fir::StoreExprId,
}
```

cesarzc: This is just a suggestion, I don't think it's a blocker to merge this PR.

PR #1377, opened 2024-04-11
Refactor ShowQuantumStateComparison to minimize required specializations

### LINKED ISSUE #1207: Refactor KatasLibrary to minimize specialization requirements for reference/actual solutions
For example, we don't need specializations to show the expected/actual state for the solution if we change the code. (Found example in `multi_qubit_measurements` task `state_preparation`, and will be used in `superposition` as well.)

tcNickolas:
This closes #1207. 
This PR also renames and reorders the new tasks in Single-Qubit gates to make them more intuitive.


cesarzc (approved):
Looks great!

PR #1379, opened 2024-04-12
Add Q# Copilot to VS Code Extension

ScottCarda-MS:
Adds the Azure Quantum Copilot chat participant to the Copilot extension when both the Copilot extension and the QDK extension are installed. To select this chat participant in the chat window, use `@quantum` at the start of the prompt. The participant has two commands: `/quantumNotebook` for quickly creating a Q# notebook, and `/samples` followed by the name of a Q# sample to have the chat-bot explain the given sample (to list available samples, provide no argument).


vscode/src/chatParticipant.ts
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as vscode from "vscode";
+import { getRandomGuid } from "./utils";
+import { log } from "qsharp-lang";
+import { getAuthSession, scopes } from "./azure/auth";
+
+const chatUrl = "https://canary.api.quantum.microsoft.com/api/chat/completions";
```
billti, line 9: Did you chat with Zander about this? I assume we shoudln't be using a `canary*` endpoint when we ship.

vscode/src/chatParticipant.ts
```diff
@@ -0,0 +1,130 @@
...
+        "Can you list the names of the quantum samples you could write if asked?",
+        "The main samples I know how to write are Bell state, Grovers, QRNG, hidden shift, Bernstein-Vazarani, Deutsch-Jozsa, superdense coding, and teleportation",
+      );
+    }
+  } else if (request.command == "quantumNotebook") {
+    stream.progress("Opening a new Quantum Notebook...");
+    await vscode.commands.executeCommand("qsharp-vscode.createNotebook");
+    return Promise.resolve({});
+  } else {
+    response = await chatRequest(msaChatSession.accessToken, request.prompt);
```
billti, line 107: As discussed, our endpoint supports streaming now. Did you discuss with Zander how to use that? Looks like these are still awaiting for the full response before the user sees anything.

ScottCarda-MS: I reached out to Zander, but hadn't heard back from him yet. The behavior seems like it is streaming, so I thought that it might not be something that we'd need to do on our end. I could be mistaken about what "streaming" looks like though.

vscode/src/chatParticipant.ts
ScottCarda-MS: The endpoint has been updated appropriately. Thanks!

vscode/src/chatParticipant.ts
ScottCarda-MS: I've now updated the logic to handle the new streaming API using `fetchEventSource`.

vscode/package.json
```diff
@@ -447,7 +447,28 @@
...
       }
-    }
+    },
+    "extensionDependencies": [
+      "github.copilot-chat"
+    ],
+    "chatParticipants": [
+      {
+        "id": "quantum.copilot",
+        "name": "quantum",
```
billti, line 457: fullName should be provided too (see <https://code.visualstudio.com/api/extension-guides/chat>)

ScottCarda-MS: Added fullname: "Azure Quantum Copliot"

billti: Per the docs for fullName on the link, "...which is shown in the title area of a response from your participant". So it should be relatively brief, and "Copilot" is probably unnecessary and redundant if it's in the copilot chat window.

I suggest just `Azure Quantum` (maybe even just `Quantum` would be fine, but marketing would probably like us to have the `Azure` in there).

vscode/package.json
```diff
@@ -473,7 +473,29 @@
       "[qsharp]": {
         "debug.saveBeforeStart": "none"
       }
-    }
+    },
+    "extensionDependencies": [
+      "github.copilot-chat"
```
billti, line 478: Being this is a new dependency, do you know what the behavior is if we ship this in an update and the user doesn't already have (or can't install for some reason) this extension? Do we fail to install? Can it be optional if so?

ScottCarda-MS: I've tested it on VS Code where the github copilot is disabled, and the extension works fine. It just doesn't provide the quantum copilot experience (because that part needs the copilot extension).

vscode/package.json
ScottCarda-MS: Okay, I've changed it to "Azure Quantum".

ScottCarda-MS:
This PR has become stale. Will reopen in the future if this feature becomes viable.

PR #1380, opened 2024-04-12
Add Q# code generation from AST.

idavis:
This PR uses the formatter and a new code gen function to turn a parsed AST into the equivalent Q#. While the AST is a 1:1 representation of the code, it isn't exactly what is in the code. We don't keep number suffixes, scientific notation, readability separators, etc.

The test compile the sample Q#, generating new Q# from the AST and getting a despanned AST package. It then feeds the new Q# back into the compiler and repeats getting Q# and a despanned AST.

We verify that the Q# generated from the original and the Q# that had been generated from the original, match. This verifies that the original code and the output both generate the same output. This is not enough as they may change the semantics.

We then compare the despanned AST from the original compilation to the despanned AST from the generated Q#. This verifies that the semantics from the original program were maintained through the round tripping.

While the former may seem unnecessary, it is very helpful in debugging any failing tests as we get rich output of the expected Q# and expected AST.

This introduces a dev-dependency to give very detailed string comparison in tests which Rust's built-in `assert_eq` lacks. It would be great if `expect-test` exposed their error diff implementation, but they don't. It hasn't been updated in years, but appears feature complete and stable. We can vendor if need be, but I couldn't find a good alternative.

### DIFF
--- a/compiler/qsc_codegen/src/qsharp.rs
+++ b/compiler/qsc_codegen/src/qsharp.rs
@@ -1,0 +1,759 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#[cfg(test)]
+mod spec_decls;
+
+#[cfg(test)]
+mod tests;
+
+#[cfg(test)]
+mod test_utils;
+
+use std::io::Write;
+use std::vec;
+
+use qsc_ast::ast::{
+    self, Attr, BinOp, Block, CallableBody, CallableDecl, CallableKind, Expr, ExprKind, Functor,
+    FunctorExpr, FunctorExprKind, Ident, Item, ItemKind, Lit, Mutability, Pat, PatKind, Path,
+    Pauli, QubitInit, QubitInitKind, QubitSource, SetOp, SpecBody, SpecDecl, SpecGen, Stmt,
+    StmtKind, StringComponent, TernOp, TopLevelNode, Ty, TyDef, TyDefKind, TyKind, UnOp,
+    Visibility, VisibilityKind,
+};
+use qsc_ast::ast::{Namespace, Package};
+use qsc_ast::visit::Visitor;
+use qsc_formatter::formatter::format_str;
+use qsc_frontend::compile::PackageStore;
+
+fn write<W: Write>(output: W, packages: &[&Package]) {
+    let mut gen = QSharpGen::new(output);
+    for package in packages {
+        gen.visit_package(package);
+    }
+}
+
+pub fn write_store<W: Write>(output: W, store: &PackageStore) {
+    let mut gen = QSharpGen::new(output);
+    for (_, unit) in store {
+        gen.visit_package(&unit.ast.package);
+    }
+}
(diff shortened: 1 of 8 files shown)


compiler/qsc_codegen/src/qsharp.rs
```diff
@@ -0,0 +1,759 @@
...
+                unreachable!()
+            }
+            ItemKind::Callable(decl) => self.visit_callable_decl(decl),
+            ItemKind::Open(ns, alias) => {
+                self.write("open ");
+                self.visit_ident(ns);
+                for a in alias {
+                    self.write(" as ");
+                    self.visit_ident(a);
+                }
```
ScottCarda-MS, line 131: This for loop confused me into thinking that there could be multiple aliases. Could you use a more Option-specific syntax here like:
```suggestion
                if let Some(a) = alias {
                    self.write(" as ");
                    self.visit_ident(a);
                }
```

ScottCarda-MS: I guess I'm seeing this for loop pattern for options pretty consistently in this code. Is there a particular reason for doing that over the if-let-some pattern?

compiler/qsc_codegen/src/qsharp.rs
```diff
@@ -0,0 +1,759 @@
...
+                for t in ty {
+                    self.write(": ");
+                    self.visit_ty(t);
+                }
+            }
+            PatKind::Elided => {
+                self.write("...");
+            }
+            PatKind::Paren(pat) => {
+                self.write("(");
```
ScottCarda-MS, line 661: Same comment here as above about preserving unnecessary parens.

ScottCarda-MS (approved):

swernli (approved):

compiler/qsc_codegen/src/qsharp.rs
```diff
@@ -0,0 +1,759 @@
...
+            ExprKind::Lambda(kind, pat, expr) => {
+                self.visit_pat(pat);
+                match kind {
+                    CallableKind::Function => self.write(" -> "),
+                    CallableKind::Operation => self.write(" => "),
+                }
+                self.visit_expr(expr);
+            }
+            ExprKind::Paren(expr) => {
+                self.write("(");
```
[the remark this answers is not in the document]
idavis: With the AST, I think we should map 1:1 as much as possible, leaving optimizations for other tooling.

compiler/qsc_codegen/src/qsharp.rs
```diff
@@ -0,0 +1,759 @@
...
+impl<W> QSharpGen<W>
+where
+    W: Write,
+{
+    pub fn new(output: W) -> Self {
+        Self { output }
+    }
+
+    pub fn write(&mut self, args: &str) {
+        let _ = write!(&mut self.output, "{args}");
```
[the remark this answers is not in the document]
idavis: It depends on the writer implementation.

PR #1381, opened 2024-04-12
Add tasks 1.11, 1.12 to Superposition Kata

jkingdon-ms:


katas/content/superposition/zero_and_bitstring/Verification.qs
```diff
@@ -0,0 +1,31 @@
...
+                CNOT(qs[0], qs[i]);
+            }
+        }
+    }
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        let qubits = 3;
+        for i in 0 .. (2 ^ qubits) - 1 {
+            let bits = IntAsBoolArray(i, qubits);
```
tcNickolas, line 19: This approach generates some tests that don't fit the task description - "You are guaranteed that the first bit of the bit string is true.", but the very first test is [false, false, false]. We'll also need to check different array lengths, to make sure the loop conditions are specified correctly.

I would recommend having a list of ~6 hardcoded arrays, starting with [[true], [true, false], [true, true], ...], and a loop iterating through them and running the tests on just them. (This will also reduce the length of the log the learner gets if the solution is correct, since we report each success)

katas/content/superposition/zero_and_bitstring/Verification.qs
```diff
@@ -0,0 +1,31 @@
...
+            Message($"Testing for bits = {bits}...");
+            if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
+            Kata.ZeroAndBitstringSuperposition(_, bits),
+            ZeroAndBitstringSuperposition_Reference(_, bits),
+            qubits) {
```
tcNickolas, line 24: Indentation
```suggestion
                Kata.ZeroAndBitstringSuperposition(_, bits),
                ZeroAndBitstringSuperposition_Reference(_, bits),
                qubits) {
```

katas/content/superposition/two_bitstrings/Verification.qs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }  
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        let qubits = 3;
+        for i in 0 .. (2 ^ qubits) - 1 {
+            let bits1 = IntAsBoolArray(i, qubits);
+            // get unsigned inverse of bits1
+            let bits2 = IntAsBoolArray(~~~i &&& 0x7, qubits);
```
tcNickolas, line 46: Same comment here - this method of generation always creates bit strings that differ in all bits, so the tests cover only a small subset of all possible inputs (for example, they don't test what happens if only the last bit is different). I suggest using a hardcoded list of 6-7 bit string pairs of different lengths that differ in all bits, in all but one bits, etc.

katas/content/superposition/two_bitstrings/Verification.qs
```diff
@@ -0,0 +1,58 @@
...
+            Message($"Testing for bits1 = {bits1} and bits2 = {bits2}...");
+            if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
+            Kata.TwoBitstringSuperposition(_, bits1, bits2),
```
tcNickolas, line 49: Indentation

katas/content/superposition/two_bitstrings/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+We will start by allocating an auxiliary qubit and preparing it in the $\frac{1}{\sqrt2} (|0\rangle + |1\rangle)$ state using the $H$ gate. The overall state of the system will be 
+
+$$\frac{1}{\sqrt2} (|0\rangle + |1\rangle)_a \otimes |0 \dots 0\rangle_r = \frac{1}{\sqrt2} (|0\rangle_a \otimes |0 \dots 0\rangle_r + |1\rangle_a \otimes |0 \dots 0\rangle_r)$$
+
+At this point, we can prepare the two basis states of the target state separately, bit by bit, controlling the preparation of one of them on the $|0\rangle$ state of the auxiliary qubit and the preparation of the other one - on the $|1\rangle$ state. 
+If a bit in one of the bit strings is `true`, we will apply a controlled $X$ gate with the auxiliary qubit as control, the qubit in the corresponding position of the register as target, and control it on the $|0\rangle$ or the $|1\rangle$ state depending on which bit string we are considering at the moment. 
+Such controlled gate can be implemented using [`ControlledOnInt`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.canon.controlledonint) library function.
```
tcNickolas, line 9: Need to update to ApplyControlledOnInt, and use new link to docs https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonint

katas/content/superposition/two_bitstrings/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+After this the state of the system will be 
+$$\frac{1}{\sqrt2} (|0\rangle_a \otimes |bits_1\rangle_r + |1\rangle_a \otimes |bits_2\rangle_r)$$
+
+Finally, we will uncompute the auxiliary qubit by using [`ControlledOnBitString`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.canon.controlledonbitstring) library function with the second bit string and the `X` operation as arguments, the quantum register as the control, and the auxiliary qubit as the target.
```
tcNickolas, line 14: Same here, https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonbitstring

tcNickolas (changes requested):
Looks good, just a couple of suggestions on improving tests and formatting.

Thank you!

katas/content/superposition/zero_and_bitstring/Verification.qs
```diff
@@ -0,0 +1,31 @@
...
+                return false;
+            }
+        }
+
+        true
```
frtibble, line 29:
```suggestion
        return true;
```

tcNickolas: [Implicit return](https://github.com/microsoft/qsharp/wiki/Differences-from-the-previous-QDK#implicit-return) is new in modern QDK, so both syntaxes will work. It's a matter of personal preference and uniformity with other code in the kata :-)

frtibble: Nice! Good to know, thanks Mariia

katas/content/superposition/zero_and_bitstring/Verification.qs
jkingdon-ms: good point! will adjust.

katas/content/superposition/two_bitstrings/solution.md
```diff
@@ -0,0 +1,33 @@
...
+If a bit in one of the bit strings is `true`, we will apply a controlled $X$ gate with the auxiliary qubit as control, the qubit in the corresponding position of the register as target, and control it on the $|0\rangle$ or the $|1\rangle$ state depending on which bit string we are considering at the moment. 
+Such controlled gate can be implemented using [`ControlledOnInt`](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonint) library function.
+
+After this the state of the system will be 
+$$\frac{1}{\sqrt2} (|0\rangle_a \otimes |bits_1\rangle_r + |1\rangle_a \otimes |bits_2\rangle_r)$$
+
+Finally, we will uncompute the auxiliary qubit by using [`ControlledOnBitString`](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonbitstring) library function with the second bit string and the `X` operation as arguments, the quantum register as the control, and the auxiliary qubit as the target.
```
tcNickolas, line 14:
```suggestion
Finally, we will uncompute the auxiliary qubit by using [`ApplyControlledOnBitString`](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonbitstring) library function with the second bit string and the `X` operation as arguments, the quantum register as the control, and the auxiliary qubit as the target. 
```

katas/content/superposition/two_bitstrings/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+We will start by allocating an auxiliary qubit and preparing it in the $\frac{1}{\sqrt2} (|0\rangle + |1\rangle)$ state using the $H$ gate. The overall state of the system will be 
+
+$$\frac{1}{\sqrt2} (|0\rangle + |1\rangle)_a \otimes |0 \dots 0\rangle_r = \frac{1}{\sqrt2} (|0\rangle_a \otimes |0 \dots 0\rangle_r + |1\rangle_a \otimes |0 \dots 0\rangle_r)$$
+
+At this point, we can prepare the two basis states of the target state separately, bit by bit, controlling the preparation of one of them on the $|0\rangle$ state of the auxiliary qubit and the preparation of the other one - on the $|1\rangle$ state. 
+If a bit in one of the bit strings is `true`, we will apply a controlled $X$ gate with the auxiliary qubit as control, the qubit in the corresponding position of the register as target, and control it on the $|0\rangle$ or the $|1\rangle$ state depending on which bit string we are considering at the moment. 
+Such controlled gate can be implemented using [`ControlledOnInt`](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonint) library function.
```
tcNickolas, line 9:
```suggestion
Such controlled gate can be implemented using [`ApplyControlledOnInt`](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/applycontrolledonint) library function.
```

tcNickolas (approved):
Looks great, just two minor comments (updated the links to the docs but forgot to update the function names themselves), I might be able to apply the suggestions myself.

Thank you!

PR #1386, opened 2024-04-14
Partial evaluation support for calls with arguments

cesarzc:
This change implements support for calls with arguments in partial arguments.


compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -126,3 +156,61 @@ impl Scope {
...
+        let mut content_runtime_kind = RuntimeKind::Static;
+        for element in elements {
+            let element_value_kind = map_eval_value_to_value_kind(element);
+            if element_value_kind.is_dynamic() {
+                content_runtime_kind = RuntimeKind::Dynamic;
+                break;
+            }
+        }
+
+        // We assume the size of all arrays is static.
```
idavis, line 171: Can you please update the comment to include why this is assumed?

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -81,16 +81,46 @@ impl Scope {
     pub fn new(
         package_id: PackageId,
         callable: Option<(LocalItemId, FunctorApp)>,
-        args_runtime_properties: Vec<ValueKind>,
+        args: Vec<Arg>,
     ) -> Self {
+        // Determine the runtimne kind (static or dynamic) of the arguments.
```
idavis, line 86:
```suggestion
        // Determine the runtime kind (static or dynamic) of the arguments.
```

compiler/qsc_partial_eval/tests/intrinsics.rs
```diff
@@ -642,3 +642,259 @@ fn call_to_intrinsic_mresetz_adds_callable_and_generates_instruction() {
...
+
+#[test]
+fn call_to_intrinsic_begin_estimate_caching_with_classical_values_does_not_generate_instructions() {
+    let program = compile_and_partially_evaluate(indoc! {
+        r#"
+        namespace Test {
+            open Microsoft.Quantum.ResourceEstimation;
+            @EntryPoint()
+            operation Main() : Unit {
+                let _ = BeginEstimateCaching("test", 0);
```
swernli, line 654: Since `BeginEstimateCaching` is usually used to affect control flow, the test should verify that this usage works as expected (namely that the function always returns a hard-coded `true` value for branching). Something like this would be good:

```suggestion
                use q = Qubit();
                if BeginEstimateCaching("test", 0) {
                    X(q);
                }
```

compiler/qsc_partial_eval/tests/intrinsics.rs
```diff
@@ -642,3 +642,259 @@ fn call_to_intrinsic_mresetz_adds_callable_and_generates_instruction() {
...
+    let program = compile_and_partially_evaluate(indoc! {
+        r#"
+        namespace Test {
+            open Microsoft.Quantum.ResourceEstimation;
+            open QIR.Intrinsic;
+            @EntryPoint()
+            operation Main() : Unit {
+                use q = Qubit();
+                let i = __quantum__qis__m__body(q) == Zero ? 0 | 1;
+                let _ = BeginEstimateCaching("test", i);
```
swernli, line 681: Same thing here, an instruction inside an if-expression would confirm the control flow behavior.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -967,6 +996,39 @@ impl<'a> PartialEvaluator<'a> {
         Value::unit()
     }
 
+    fn resolve_args(&self, store_pat_id: StorePatId, value: Value) -> Vec<Arg> {
+        let pat = self.package_store.get_pat(store_pat_id);
+        match &pat.kind {
+            PatKind::Discard => vec![Arg::Discard(value)],
```
swernli, line 1002: I actually think you can just panic on this case... a discard in a argument list is only used for creating a new lambda via partial application, which will have been lifted by earlier parts of compilation. That also means that `Arg::Discard` can probably go away, leaving `Arg` to become a tuple-struct with what today is in `Arg::Variable`. Perhaps something like `struct Arg(pub LocalVariableId, Variable);` would be sufficient.

cesarzc: I think you can also have discards even in regular callables:
```qsharp
function Foo(_ : Int) : Int {
    0
}
```

Based on the HIR in the playground, the argument to `Foo` is a discard.

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -126,3 +156,62 @@ impl Scope {
...
+        | Value::Bool(_)
+        | Value::Closure(_)
+        | Value::Double(_)
+        | Value::Global(_, _)
+        | Value::Int(_)
+        | Value::Pauli(_)
+        | Value::Qubit(_)
+        | Value::Range(_)
+        | Value::String(_) => ValueKind::Element(RuntimeKind::Static),
+    }
```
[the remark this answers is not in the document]
cesarzc: Good catch! I'll add a few tests for this too.

compiler/qsc_partial_eval/tests/intrinsics.rs
cesarzc: Updated test.

compiler/qsc_partial_eval/tests/intrinsics.rs
cesarzc: Updated test.

compiler/qsc_partial_eval/src/evaluation_context.rs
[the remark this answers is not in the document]
cesarzc: Added tests.

swernli (approved):

PR #1387, opened 2024-04-15
Partial Evaluation Output Recording support

swernli:
This adds logic to insert output recording calls for recognized types into RIR programs generated by partial evaluation.

### DIFF
--- a/compiler/qsc_partial_eval/src/evaluation_context.rs
+++ b/compiler/qsc_partial_eval/src/evaluation_context.rs
@@ -72,6 +72,7 @@ pub struct Scope {
     pub callable: Option<(LocalItemId, FunctorApp)>,
     pub args_runtime_properties: Vec<ValueKind>,
     pub env: Env,
+    last_expr: Option<ExprId>,
     hybrid_exprs: FxHashMap<ExprId, Value>,
     hybrid_vars: FxHashMap<LocalVarId, Value>,
 }
@@ -87,12 +88,15 @@ impl Scope {
             callable,
             args_runtime_properties,
             env: Env::default(),
+            last_expr: None,
             hybrid_exprs: FxHashMap::default(),
             hybrid_vars: FxHashMap::default(),
         }
     }
 
-    pub fn get_expr_value(&self, expr_id: ExprId) -> &Value {
+    // Maybe this can be removed in favor of a single value getter, which would also eliminate the need
+    // for `hybrid_exprs` entirely?
+    pub fn _get_expr_value(&self, expr_id: ExprId) -> &Value {
         self.hybrid_exprs
             .get(&expr_id)
             .expect("expression value does not exist")
@@ -105,10 +109,21 @@ impl Scope {
     }
 
     pub fn insert_expr_value(&mut self, expr_id: ExprId, value: Value) {
+        self.last_expr = Some(expr_id);
         self.hybrid_exprs.insert(expr_id, value);
     }
 
     pub fn insert_local_var_value(&mut self, local_var_id: LocalVarId, value: Value) {
         self.hybrid_vars.insert(local_var_id, value);
+    }
+
+    pub fn clear_last_expr(&mut self) {
+        self.last_expr = None;
+    }
+
+    pub fn last_expr_value(&self) -> Value {
+        self.last_expr
+            .and_then(|expr_id| self.hybrid_exprs.get(&expr_id))
+            .map_or_else(Value::unit, Clone::clone)
     }
 }
--- a/compiler/qsc_partial_eval/src/evaluation_context.rs
+++ b/compiler/qsc_partial_eval/src/evaluation_context.rs
@@ -94,8 +94,7 @@ impl Scope {
         }
     }
 
-    // Maybe this can be removed in favor of a single value getter, which would also eliminate the need
-    // for `hybrid_exprs` entirely?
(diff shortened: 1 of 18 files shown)


compiler/qsc_partial_eval/tests/qubits.rs
```diff
@@ -36,20 +38,42 @@ fn qubit_ids_are_correct_for_allocate_use_release_one_qubit() {
...
-        &[
-            Instruction::Call(
-                op_callable_id,
-                vec![Operand::Literal(Literal::Qubit(0))],
-                None,
-            ),
-            Instruction::Return,
-        ],
-    );
+    expect![[r#"
```
swernli, line 41: Would it make sense to update all the partial eval tests to expect-macro patterns? In making this change, I found I'd have to manually update almost every test that generates RIR and checks the value against a constructed program. If the tests were string based like this one example shows, then a global expect-update would add the needed strings everywhere. It would also remove the need for the test utilities `assert_callable` and `assert_block_instructions` since all that same info (and more) is present in the strings.

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -87,12 +88,15 @@ impl Scope {
...
+            last_expr: None,
             hybrid_exprs: FxHashMap::default(),
             hybrid_vars: FxHashMap::default(),
         }
     }
 
-    pub fn get_expr_value(&self, expr_id: ExprId) -> &Value {
+    // Maybe this can be removed in favor of a single value getter, which would also eliminate the need
+    // for `hybrid_exprs` entirely?
+    pub fn _get_expr_value(&self, expr_id: ExprId) -> &Value {
```
swernli, line 99: We should talk through this idea... I'm more and more convinced that the expr-to-value hashmap shouldn't be needed.

compiler/qsc_partial_eval/tests/qubits.rs
swernli: Also, until the rest of the tests are updated, the CI build will fail, so we should decide one way or the other either as part of this PR or in a separate PR what pattern we'd like to follow.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -950,15 +969,181 @@ impl<'a> PartialEvaluator<'a> {
...
+                            variable_id: var.0.into(),
+                            ty: record_ty,
+                        }),
+                        Operand::Literal(Literal::Pointer),
+                    ],
+                    None,
+                ));
+            }
+
+            Value::BigInt(_)
```
idavis, line 1065: @swernli I thought that we implicitly wrapped single values of supported types in a tuple.

compiler/qsc_partial_eval/src/evaluation_context.rs
swernli: After some discussion, we landed at leaving the hashmap in right now to support ongoing work and reevaluating if ongoing work needs to use `get_expr_value` for more than the most recent expression result.

compiler/qsc_partial_eval/src/lib.rs
swernli: No, if only a single value is returned then it doesn't have any aggregate and will show up as it's own idividual item.

compiler/qsc_partial_eval/tests/qubits.rs
swernli: Trying out a targeted expect-test based approach, let me know what you think.

compiler/qsc_partial_eval/tests/branching.rs
```diff
@@ -236,41 +235,77 @@ fn if_expression_with_dynamic_condition() {
...
     let if_block_id = BlockId(2);
 
     // Verify the branch instruction in the initial-block.
-    let condition_var = Variable {
-        variable_id: 1.into(),
-        ty: Ty::Boolean,
-    };
-    let branch_inst = Instruction::Branch(condition_var, if_block_id, continuation_block_id);
-    assert_block_last_instruction(&program, initial_block_id, &branch_inst);
+    assert_block_last_instruction(
```
cesarzc, line 284: For branching tests, I think it is more readable to have just an `assert_blocks` function that compares against all the blocks than asserting each one of the blocks individually. IMO, it is easier to see where the branch and jump instructions point to if all the blocks are displayed at once.

swernli: Makes sense, I can add that!

cesarzc (approved):

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -963,15 +981,181 @@ impl<'a> PartialEvaluator<'a> {
...
+            | Value::Bool(_)
+            | Value::Closure(_)
+            | Value::Double(_)
+            | Value::Global(_, _)
+            | Value::Int(_)
+            | Value::Pauli(_)
+            | Value::Qubit(_)
```
idavis, line 1084: Add literal bool and int support

idavis (approved):

compiler/qsc_partial_eval/src/lib.rs
swernli: Good call, I'll add that and tests.

PR #1389, opened 2024-04-15
Add CZ section and CNOT and CZ tasks to Multi-Qubit Gates kata

WWhitedogi:


Manvi-Agrawal (commented):
Overall, it looks good. I left a few comments regarding typos, naming of tasks and ordering. Thank you for your contributions.

katas/content/multi_qubit_gates/index.md
Manvi-Agrawal, line 1: Unrelated to this PR. 
@tcNickolas, I understand that this tutorial follows the `Multi Qubit Gate` tutorial in the katas. I feel its a bit non cohesive and we can reorder the tasks a bit to make it more natural to follow along. How about this?

- CNOT, CZ, SWAP gates.
 - ket bra decomposition
 - Compound Gate
 - Controlled Gates and Multi-Controlled Gates
 - Gates on non adjacent qubits (Possibly add one line saying that in todays quantum computers, we don't have full qubit connectivity) to highlight practical relevance. Otherwise, it feels redundant to use a bunch of SWAP and CNOT gates, to implement CNOT on non-adjacent qubits, when QDK allows us to simulate it. xD

tcNickolas: Is there a particular reason why this flow feels more natural to you? The MultiQubitGates tutorial in the katas followed the flow that feels natural to me, and still does :-) 

* Compound gates are applying single-qubit gates to quantum systems, that's a more intuitive first step, since the learner is already familiar with the single-qubit gates and only needs to learn to apply them to multi-qubit systems. For example, Bell state prep task already relies on this, since it starts with applying an H gate.
* Gates acting on non-adjacent qubits are discussed here not in the context of physical implementation of gates (pretty much none of the gates we discuss in the tutorials are implemented as is), but in the context of teaching the learner to do the math for this scenario, how to do it with matrices and with Dirac notation. Q# allows us to simulate this, true, but first the learner needs to be able to figure out what gates to write, and they need math for this.

Let's take this discussion out of this PR, since it's much harder for me to keep track of this conversation in parallel with the discussion of the proposed changes.

Manvi-Agrawal: Works for me @tcNickolas .

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+    "qsDependencies": [
+        "../KatasLibrary.qs"
+    ]
+})
+@[section]({
+    "id": "multi_qubit_gates__cz_gate",
+    "title": "CZ Gate"
+})
+
+## CZ Gate (Controlled-Z Gate)
```
tcNickolas, line 138: Nice lesson! The lesson title is in the section name, so we don't need the second header here
```suggestion
```

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+
+The $CZ$ ("controlled-Z") gate is a two-qubit gate, with one qubit referred to as the **control** qubit, and the other as the **target** qubit. Typically, the first qubit is the control, and the second qubit is the target.
```
tcNickolas, line 140:
```suggestion
The $CZ$ ("controlled-Z") gate is a two-qubit gate, with one qubit referred to as the **control** qubit, and the other as the **target** qubit. Interestingly, for the $CZ$ gate it doesn't matter which qubit is control and which is target - the effect of the gate is the same either way!
```

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+
+The $CZ$ gate acts as a conditional gate: if the control qubit is in state $|1\rangle$, it applies the $Z$ gate to the target qubit, otherwise it does nothing.
+
+> If the system is in a superposition of several basis states, the effects of the gate will be a linear combination of the effects of it acting separately on each of the basis states.
+> This is a common feature for all quantum gates you'll encounter that are specified in terms of basis states: since all unitary gates are linear, it is sufficient to define their effect on the basis states, and use linearity to figure out their effect on any state.
+
```
tcNickolas, line 146: I don't think we need to repeat this note, since we mention for CNOT that this is a common feature
```suggestion
```

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+            $$CZ|11\rangle = -|11\rangle$$
+        </td>
+    </tr>
+</table>
+
+The $CZ$ gate is particularly useful for creating and manipulating entangled states where the phase of the quantum state is crucial. Consider the following separable state:
+
+$$\big(\alpha|0\rangle + \beta|1\rangle\big) \otimes \big(\gamma|0\rangle + \delta|1\rangle\big) = \alpha\gamma|00\rangle + \alpha\delta|01\rangle + \beta\gamma|10\rangle + \beta\delta|11\rangle$$
+
+If we apply the $CZ$ gate to it, with the first qubit as the control and the second as the target, we get the following state, which can no longer be easily separated:
```
tcNickolas, line 178:
```suggestion
If we apply the $CZ$ gate to it, with the first qubit as the control and the second as the target (or vice versa), we get the following state, which can no longer be separated:
```

katas/content/multi_qubit_gates/two_qubit_gate_1/index.md
```diff
@@ -0,0 +1,5 @@
+**Input:** Two unentangled qubits (stored in an array of length 2).
+The first qubit will be in state $|\psi\rangle = \alpha |0\rangle + \beta |1\rangle$, the second - in state $|0\rangle$
+(this can be written as two-qubit state $\big(\alpha |0\rangle + \beta |1\rangle \big) \otimes |0\rangle = \alpha |00\rangle + \beta |10\rangle$.
```
tcNickolas, line 3:
```suggestion
(this can be written as two-qubit state $\big(\alpha |0\rangle + \beta |1\rangle \big) \otimes |0\rangle = \alpha |00\rangle + \beta |10\rangle$).
```

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
 
 The $CNOT$ gate is self-adjoint: applying it for the second time reverses its effect.
 
+
+@[exercise]({
+    "id": "multi_qubit_gates__two_qubit_gate_1",
+    "title": "Two Qubit Gate 1",
```
tcNickolas, line 127: Originally the task names looked like this because there was no theory in the BasicGates kata, so part of the challenge was figuring out which gate was supposed to be used. Now there is no mystery (the only gate we learned so far is CNOT), so we can use more descriptive names.
How about we call this task "Entangle Qubits"? (And rename the folder too)

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+    "id": "multi_qubit_gates__two_qubit_gate_3",
+    "title": "Two Qubit Gate 3",
+    "path": "./two_qubit_gate_3/",
+    "qsDependencies": [
+        "../KatasLibrary.qs"
+    ]
+})
 @[exercise]({
     "id": "multi_qubit_gates__preparing_bell_state",
     "title": "Preparing a Bell State",
```
tcNickolas, line 202: Let's move the Bell State task up to the CNOT gate lesson, after "Entangle Qubits" and before CZ lesson

katas/content/multi_qubit_gates/two_qubit_gate_1/Placeholder.qs
```diff
@@ -0,0 +1,6 @@
+namespace Kata {
+    operation TwoQubitGate1 (qs : Qubit[]) : Unit is Adj + Ctl {
```
tcNickolas, line 2:
```suggestion
    operation EntangleQubits (qs : Qubit[]) : Unit is Adj + Ctl {
```
(and a matching change in other places)

katas/content/multi_qubit_gates/two_qubit_gate_1/solution.md
```diff
@@ -0,0 +1,11 @@
+Let's denote the first qubit in state $\alpha |0\rangle + \beta |1\rangle$ as A and the second qubit in state $|0\rangle$ as B.
+
+Compare our input state $\alpha |0_A0_B\rangle + \beta |1_A0_B\rangle$ with the goal state $\alpha |0_A0_B\rangle + \beta |1_A1_B\rangle$. 
+We want to pass our input qubit through a gate or gates (to be decided) that do the following. If qubit A is in the $|0\rangle$ state, then we want to leave qubit B alone (the first term of the superposition). 
+However, if A is in the $|1\rangle$ state, we want to flip qubit B from $|0\rangle$ into $|1\rangle$ state. In other words, the state of B is to be made contingent upon the state of A. 
+This gate exists and is called a CNOT (controlled-not) gate. Depending upon the state of the **control** qubit (A in our case), the value of the controlled or **target** qubit (B in our case) is inverted or unchanged.  Thus we get the goal state $\alpha |00\rangle + \beta |11\rangle$.
```
tcNickolas, line 6: This reads weird in a tutorial when we just told the learner all about this gate and now we tell them again that it exists... Let's rephrase a bit
```suggestion
This is exactly the effect of the $CNOT$ gate. Depending upon the state of the **control** qubit (A in our case), the value of the controlled or **target** qubit (B in our case) is inverted or unchanged. Thus, we get the goal state $\alpha |00\rangle + \beta |11\rangle$.  
```

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+
+If we apply the $CZ$ gate to it, with the first qubit as the control and the second as the target, we get the following state, which can no longer be easily separated:
+
+$$\alpha\gamma|00\rangle + \alpha\delta|01\rangle + \beta\gamma|10\rangle - \beta\delta|11\rangle$$
+
+The $CZ$ gate is also self-adjoint: applying it a second time reverses its effect, similar to the $CNOT$ gate.
+
+@[exercise]({
+    "id": "multi_qubit_gates__two_qubit_gate_2",
+    "title": "Two Qubit Gate 2",
```
tcNickolas, line 186: Let's call this task "Relative Phase -1" and make the folder relative_phase_minusone

katas/content/multi_qubit_gates/index.md
```diff
@@ -121,6 +121,82 @@ $$\alpha|00\rangle + \beta|11\rangle$$
...
+    "path": "./two_qubit_gate_2/",
+    "qsDependencies": [
+        "../KatasLibrary.qs"
+    ]
+})
+@[exercise]({
+    "id": "multi_qubit_gates__two_qubit_gate_3",
+    "title": "Two Qubit Gate 3",
```
tcNickolas, line 194: I spent some time staring at this task vs the existing qubit_swap task, and they look too similar to each other... I only checked that no tasks were identical when I was looking at the migration, but I didn't realize how similar they are while being slightly different :-( And this kata is shaping up to be pretty big, since we have one more batch of tasks to add to it...

Let's drop this task - apologies for the hassle!

katas/content/multi_qubit_gates/two_qubit_gate_1/Verification.qs
```diff
@@ -0,0 +1,25 @@
...
+    open Microsoft.Quantum.Katas;
+
+    operation  TwoQubitGate1 (qs : Qubit[]) : Unit is Adj + Ctl {
+        CNOT(qs[0], qs[1]);
+    }
+
+    operation CheckSolution() : Bool {
+        let solution = Kata. TwoQubitGate1;
+        let reference =  TwoQubitGate1;
+        let isCorrect = CheckOperationsEquivalence(solution, reference, 2);
```
tcNickolas, line 11: In this test, we don't need to check the equivalence of operations as unitaries, since the tasks asks us to prepare a state given another specific state. (For example, we can apply Z gate to the 2nd qubit before the CNOT, it would be a different unitary but a correct solution too.)

The test we need here is very similar to what I'm doing in https://github.com/microsoft/qsharp/blob/mariia/qec_shor/katas/content/qec_shor/bitflip_encode/Verification.qs using the operation `CheckOperationsEquivalenceOnInitialStateStrict` you created in your previous PR: compare the effect of the solution and the reference on several specific input states with different parameters. (I'm also refactoring the code to move it to the KatasLibrary.qs as part of that change - you can see the complete diff in https://github.com/microsoft/qsharp/compare/mariia/qec_shor?expand=1).

My change won't be ready for a few more days, though. For your PR, the easiest way to do is copy-paste CheckOperationsEquivalenceOnInitialStateStrict to this Verification.qs file and copy the test from my branch; I'll include that in the refactoring once it's merged and I'll be creating my own pull request.

katas/content/multi_qubit_gates/two_qubit_gate_2/Placeholder.qs
```diff
@@ -0,0 +1,7 @@
+namespace Kata {
+    operation TwoQubitGate2 (qs : Qubit[]) : Unit is Adj + Ctl {
+        // Implement your solution here...
+
```
tcNickolas, line 4:
```suggestion
```

katas/content/multi_qubit_gates/two_qubit_gate_2/Placeholder.qs
```diff
@@ -0,0 +1,7 @@
+namespace Kata {
+    operation TwoQubitGate2 (qs : Qubit[]) : Unit is Adj + Ctl {
```
tcNickolas, line 2:
```suggestion
    operation RelativePhaseMinusOne (qs : Qubit[]) : Unit is Adj + Ctl {
```
(and matching changes everywhere else)

katas/content/multi_qubit_gates/two_qubit_gate_2/Verification.qs
```diff
@@ -0,0 +1,25 @@
...
+    open Microsoft.Quantum.Katas;
+
+    operation  TwoQubitGate2 (qs : Qubit[]) : Unit is Adj + Ctl {
+        CZ(qs[0], qs[1]);
+    }
+
+    operation CheckSolution() : Bool {
+        let solution = Kata. TwoQubitGate2;
+        let reference =  TwoQubitGate2;
+        let isCorrect = CheckOperationsEquivalence(solution, reference, 2);
```
tcNickolas, line 11: Same verification comment here; we need to use CheckOperationsEquivalenceOnInitialStateStrict  with a single initial state prepared as qs => ApplyToEach(H, qs)

katas/content/multi_qubit_gates/two_qubit_gate_2/solution.md
```diff
@@ -0,0 +1,49 @@
+Firstly we notice that we are dealing with an unentangled pair of qubits. In vector form this is:
+In vector form the transformation we need is
+
```
tcNickolas, line 3:
```suggestion
Firstly we notice that we are dealing with an unentangled pair of qubits.
```

katas/content/multi_qubit_gates/two_qubit_gate_2/solution.md
```diff
@@ -0,0 +1,49 @@
...
+This gate is almost Pauli I, the identity gate, with the different in just the last column, showing what will happen to the $|11\rangle$ basis state. Applying it to our input state for $\alpha = \pi$, we'll get:
+
+$$
+\frac{1}{2}\begin{bmatrix}1 & 0 & 0 & 0\\\ 0 & 1 & 0 & 0\\\ 0 & 0 & 1 & 0\\\ 0 & 0 & 0 &\color{red}{e^{i\alpha}} \end{bmatrix}
+\begin{bmatrix}1\\\ 1\\\ 1\\\ 1\\\ \end{bmatrix}=
+\frac{1}{2}\begin{bmatrix}1\\\ 1\\\ 1\\\ 1\cdot\color{red}{e^{i\alpha}}\\ \end{bmatrix}=
+\frac{1}{2} \big( |00\rangle + |01\rangle + |10\rangle {\color{red}-} |11\rangle \big)
+$$
+
+The last thing we notice if we look through the [list of operations in the Microsoft.Quantum.Canon namespace](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.canon/cz) is the CZ (Controlled Z) gate, a special case of CR1 that implements exactly this gate.
```
tcNickolas, line 38: I'm not really sure why the explanation in the workbook is so convoluted when it leads to just using CZ... It's possible that we had in mind an extra task that would introduce a complex relative phase, but it's rather an overkill in this case with the CZ lesson added. Let's shorten it

```suggestion
We remember that the Pauli $Z$ gate flips signs in the single qubit case, and that $CZ$ is the 2-qubit version of this gate. And indeed, the effect of the $CZ$ gate is exactly the transformation we're looking for here.
```

tcNickolas (changes requested):
Looks good overall, and nice CZ lesson, that gate slipped our minds when we were creating the original tutorial!

I left some comments for tasks 2.1-2.3. Looking at the remaining tasks (for the next PR, not for this one, just saying this in case you work on the next one over the weekend),
* Let's call 2.4 "Amplitudes Swap" and add it as the third exercise in CNOT lesson
* 2.5 is good with current name, in the lesson Multi-controlled gates
* 2.6 is also good with current name, as the first exercise in the lesson Controlled Gates  (before Controlled Rotation)

And those three should be tested as unitaries indeed, since they accept generic input states, not specific ones.

Thank you!

katas/content/multi_qubit_gates/index.md
```diff
@@ -129,6 +131,14 @@ The $CNOT$ gate is self-adjoint: applying it for the second time reverses its ef
         "../KatasLibrary.qs"
     ]
 })
+@[exercise]({
+    "id": "multi_qubit_gates__two_qubit_gate_1",
+    "title": "Two Qubit Gate Controll Not",
```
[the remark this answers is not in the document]
tcNickolas: GitHub didn't show me this comment when I was looking at the files, sorry - I left a comment in my review

katas/content/multi_qubit_gates/index.md
```diff
@@ -259,7 +269,14 @@ The $SWAP$ gate acts on two qubits, and, as the name implies, swaps their quantu
         </td>
     </tr>
 </table>
-
+@[exercise]({
```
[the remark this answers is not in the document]
tcNickolas: After some time glaring at the new and the old tasks, I think we can just keep the old one and not add the new one. It shouldn't need the math of acting on non-adjacent qubits, since they're just swaps - at least I hope so. If we hear back from a confused learner, we'll revisit

WWhitedogi:
1. Delete swap task gate folder
2. Rename task
3. update verification.qs
4. update index.md

tcNickolas (approved):
Looks great! Just two minor comments.

Thank you!

PR #1390, opened 2024-04-15
Fix docs error on first load

billti:
The type assertion was incorrect, and props.documentation is not defined on first load, causing an error which caused later rendering to have issues.

Use a runtime check rather than a type assertion.


billti:
(Actually, this may have been caused by some local work in progress. Either way, the runtime check is safer (it fixed my runtime issue for example) and avoids the need for the type assertion).

idavis (approved):

PR #1393, opened 2024-04-15
Update profile configuration

idavis:
This updates the conditional compilation logic and target profile names. We need to take a good look at the conditional compilation attributes used to ensure that they are correct. This also updates the config attribute logic to deal with capabilities rather than profiles.

### DIFF
--- a/compiler/qsc_codegen/src/qsharp/spec_decls.rs
+++ b/compiler/qsc_codegen/src/qsharp/spec_decls.rs
@@ -36,14 +36,14 @@ fn attributes() {
         indoc! {r#"
             namespace Sample {
                 @EntryPoint()
-                @Config(Base)
+                @Config(Unrestricted)
                 operation Entry() : Unit {}
             }"#},
         None,
         &expect![[r#"
             namespace Sample {
                 @EntryPoint()
-                @Config(Base)
+                @Config(Unrestricted)
                 operation Entry() : Unit {}
             }"#]],
     );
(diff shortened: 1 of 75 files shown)


swernli (approved):

cesarzc (approved):

ScottCarda-MS (approved):
signing off on the changes to the language service components

compiler/qsc_codegen/src/qsharp/spec_decls.rs
```diff
@@ -36,14 +36,14 @@ fn attributes() {
         indoc! {r#"
             namespace Sample {
                 @EntryPoint()
-                @Config(Base)
+                @Config(Unrestricted)
```
minestarks, line 39: Why did this test have to change?

compiler/qsc_frontend/src/compile.rs
```diff
@@ -40,60 +40,33 @@ use thiserror::Error;
 bitflags! {
     #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct RuntimeCapabilityFlags: u32 {
```
minestarks, line 42: I think the name "target capability" would start to make more sense here, since it includes arbitrary  target hardware capabilities like Qubit Reset. It also sounds nicely similar to "target profile", which would be a group of "target capabilities".

minestarks (approved):
nitpicking, nitpicking...

compiler/qsc_codegen/src/qsharp/spec_decls.rs
idavis: Because a bug was fixed in contig attr logic.

PR #1397, opened 2024-04-17
Add RIR type check pass

swernli:
This change introduces a type check pass for RIR that verifies and panics on any mismatches found in the types. It adds the check to the default passes and the SSA transform tests.


compiler/qsc_rir/src/passes/ssa_transform/tests.rs
```diff
@@ -19,11 +21,14 @@ use super::transform_to_ssa;
 
 fn transform_program(program: &mut Program) {
```
idavis, line 22: Why are we duplicating `check_and_transform` in tests? The `transform_program` looks like a line for line copy of a public function.

swernli: My concern was that `check_and_transform` could be updated to have more passes that run after these that would not be relevant to these tests. Right now, that seems unlikely so I would be open to changing it to just use the top-level transform.

idavis: My main concern is what we are testing and what we are doing get out of sync.

idavis (approved):

compiler/qsc_rir/src/passes/ssa_transform/tests.rs
swernli: Fair enough. Given that right now they are equal, I'll update to call the check and transform. That will mean I also need to be sure to set the capabilities appropriately so that the right passes get run!

PR #1400, opened 2024-04-18
Use new QIR gen API for Base Profile

idavis:


compiler/qsc_codegen/src/qir.rs
```diff
@@ -546,9 +544,92 @@ impl ToQir<String> for rir::Program {
...
+
+    let mut index = 4;
+
+    // If we are not in the base profile, we need to add the capabilities
+    // associated with the adaptive profile.
+    if !program.config.is_base() {
+        // loop through the capabilities and add them to the metadata
+        // for values that we can generate.
+        for cap in program.config.capabilities.iter() {
+            let name = match cap {
```
swernli, line 580: nitpick: since this match expression is the same as the one below, I wonder if there is reason to have a utlity to turn capability flags into the matching QIR attributes. Not critical by any means.

swernli (approved):

cesarzc (approved):

DmitryVasilevsky (approved):

minestarks (approved):

PR #1401, opened 2024-04-18
Update doc comments in std library

DmitryVasilevsky:
This PR updates doc comments in the standard library. It also adds information to the readme file describing what to put in different sections of doc comments.
This is mostly random updates to a few spots in the library rather than a consistent sweep over all functions and operation. Mostly to solicit feedback on changes needed.


library/core/core.qs
```diff
@@ -3,20 +3,32 @@
...
     /// # Summary
-    /// Creates an array of given length with all elements equal to given value.
+    /// Creates an array of given `length` with all elements equal to given
+    /// `value`. `length` must be a non-negative integer.
+    ///
+    /// # Description
+    /// Use this function to create an array of length `length` where each
+    /// element is equal to `value`. This way of creating an array is preferred
+    /// over other methods if all elements of the array must be the same and
+    /// the length is known upfront.
```
ScottCarda-MS, line 31: Is this really the preferred way? I would have thought `[One, size = 3]` syntax would be preferred.

library/std/convert.qs
```diff
@@ -48,7 +62,8 @@ namespace Microsoft.Quantum.Convert {
     }
 
     /// # Summary
-    /// Produces a non-negative integer from a string of bits in little endian format.
+    /// Produces a non-negative integer from a string of bits in little-endian format.
+    /// bits[0] represents the least significant bit.
```
ScottCarda-MS, line 66:
```suggestion
    /// `bits[0]` represents the least significant bit.
```

library/std/math.qs
```diff
@@ -10,11 +10,16 @@ namespace Microsoft.Quantum.Math {
...
     /// # Summary
-    /// Represents the ratio of the circumference of a circle to its diameter.
+    /// Returns a double-precision approximation of the
+    /// matematical constant 𝝅 ≈ 3.14159265358979323846
     ///
-    /// # Output
-    /// A double-precision approximation of the circumference of a circle
-    /// to its diameter, π ≈ 3.14159265358979323846.
+    /// # Remarks
+    /// Matematical constant 𝝅 represents the ratio of the circumference
```
ScottCarda-MS, line 17:
```suggestion
    /// Mathematical constant 𝝅 represents the ratio of the circumference
```

library/core/core.qs
DmitryVasilevsky: @swernli may have more on this, but I believe we want to deprecate [One, size=3] syntax eventually.

swernli: Yeah, that was the main motivation behind introducing `Repeated` and putting it in the core. Hopefully we can discourage usage of `[One, size = 3]` enough to make it less painful in some future where it is removed.

library/std/canon.qs
```diff
@@ -432,8 +436,16 @@ namespace Microsoft.Quantum.Canon {
...
 
     /// # Summary
-    /// Applies a unitary operation on the target,
-    /// controlled on a state specified by a given bit mask.
+    /// Applies `oracle` on `target` when `controlRegister`
+    /// is in the state specified by `bits`.
+    ///
+    /// # Description
+    /// Applies a unitary operation `oracle` on the `target`, controlled
+    /// on a state specified by a given bit mask `bits`. Bit bits[i] corresponds to qubit controlRegister[i].
```
cesarzc, line 444:
```suggestion
/// on a state specified by a given bit mask `bits`.
/// Bit at `bits[i]` corresponds to qubit at `controlRegister[i]`.
```

library/std/canon.qs
```diff
@@ -432,8 +436,17 @@ namespace Microsoft.Quantum.Canon {
...
     /// # Summary
-    /// Applies a unitary operation on the target,
-    /// controlled on a state specified by a given bit mask.
+    /// Applies `oracle` on `target` when `controlRegister`
+    /// is in the state specified by `bits`.
+    ///
+    /// # Description
+    /// Applies a unitary operation `oracle` on the `target`, controlled
+/// on a state specified by a given bit mask `bits`.
+/// Bit at `bits[i]` corresponds to qubit at `controlRegister[i]`.
```
swernli, line 445:
```suggestion
    /// The bit at `bits[i]` corresponds to qubit at `controlRegister[i]`.
```

library/std/canon.qs
```diff
@@ -432,8 +436,17 @@ namespace Microsoft.Quantum.Canon {
...
 
     /// # Summary
-    /// Applies a unitary operation on the target,
-    /// controlled on a state specified by a given bit mask.
+    /// Applies `oracle` on `target` when `controlRegister`
+    /// is in the state specified by `bits`.
+    ///
+    /// # Description
+    /// Applies a unitary operation `oracle` on the `target`, controlled
+/// on a state specified by a given bit mask `bits`.
```
swernli, line 444:
```suggestion
    /// on a state specified by a given bit mask `bits`.
```

library/std/measurement.qs
```diff
@@ -13,32 +13,39 @@ namespace Microsoft.Quantum.Measurement {
...
     ///
     /// # Output
-    /// The result of measuring `Z ⊗ Z ⊗ ••• ⊗ Z`.
+    /// The result of measuring in the `Z ⊗ Z ⊗ ••• ⊗ Z` basis.
     ///
-    /// # Remarks
-    /// This operation does not reset the measured qubits to the |0⟩ state,
-    /// leaving them in the state that corresponds to the measurement result.
+    /// # See also
+    /// - Microsoft.Quantum.Measurement.MeasureEachZ
```
swernli, line 27: I'm glad to have a hint toward `MeasureEachZ` here, but I'm torn on whether it should be a "See Also" or part of remarks so that it can have some explanatory text of why they are related. Something like, "For an operation measures each qubit and returns an array of results, see `Microsoft.Quantum.Measurement.MeasureEachZ`."

library/std/measurement.qs
```diff
@@ -13,32 +13,39 @@ namespace Microsoft.Quantum.Measurement {
...
     /// ## targets
     /// An array of qubits to be measured.
     /// # Output
     /// An array of measurement results.
     ///
-    /// # Remarks
-    /// This operation does not reset the measured qubits to the |0⟩ state,
-    /// leaving them in the state that corresponds to the measurement results.
+    /// # See also
+    /// - Microsoft.Quantum.Measurement.MeasureAllZ
```
swernli, line 48: For this one, I think it's more relevant to have a hint toward `MResetEachZ`, but it could go either way.

swernli (approved):
This is great, thanks Dmitry!

library/std/canon.qs
DmitryVasilevsky: Updated. Somehow the previously committed suggestions didn't have proper indents so I'll update that too.

library/std/canon.qs
DmitryVasilevsky: Updated.

library/std/measurement.qs
DmitryVasilevsky: I added remarks section to MeasureEachZ describing the differences between MeasureEachZ, MeasureAllZ, and MResetEachZ. I also added/kept all 'see also' references.

PR #1403, opened 2024-04-18
Allow QIR tab in playground along with qir/ls profile configuration.

idavis:


idavis:
Waiting for #1400 before being ready.

playground/src/editor.tsx
```diff
@@ -56,9 +58,27 @@ function VSDiagsToMarkers(errors: VSDiagnostic[]): monaco.editor.IMarkerData[] {
...
+    const profile = params.get("profile");
+    if (profile === "base") {
+      return "base";
+    }
+    if (profile === "quantinuum") {
+      return "quantinuum";
+    }
+  }
+
+  return "unrestricted";
```
billti, line 75:
```suggestion
  return new URLSearchParams(window.location.search).get("profile") ?? "unrestricted";
```

playground/src/editor.tsx
```diff
@@ -116,38 +138,79 @@ export function Editor(props: {
...
+        clearTimeout(compilerTimeout);
+        props.setQir(qir);
+      } catch (e: any) {
+        if (timedOut) {
+          props.setQir("timed out");
+        } else {
+          props.setQir(e.toString());
+        }
+      } finally {
+        compiler.terminate();
```
billti, line 185: Is it benign to call this twice? The setTimeout body on line 172 may have called it, then the finally here will call it also. If you look at `webviewPanel.ts` in the VS Code source, it uses a flag on timeout to ensure that terminate isn't called twice. (I'm guessing I did that for reason, any probably tested at the time, but I don't recall now).

playground/src/main.tsx
```diff
@@ -70,6 +75,12 @@ function App(props: { katas: Kata[]; linkedCode?: string }) {
   const [compiler, setCompiler] = useState(() =>
     createCompiler(setCompilerState),
   );
+
+  const [compiler_worker_factory] = useState(() => {
```
billti, line 79: `createCompiler` on line 66 looks like it might be suitable, and is already created as state on line 75 here. Is that reusable, or not appropriate for some reason?

playground/src/editor.tsx
idavis: @billti This would allow arbitrary profile "names" to be set in the url, which violates the return type of TargetProfile. Is this something you'd like so that it 'just works' when we add new profiles? I can go either way as this is our playground.

playground/src/main.tsx
idavis: It isn't reusable. The qir compiler worker is expected to fail a lot and get runtime errors.

billti (approved):

playground/src/editor.tsx
billti: Assuming they code the code share link from the playground in the first place (where else?) I don't think we need to be overly cautious. Garbage in garbage out. (It may be equally confusing to 'think' you're passing a specific profile but it just keeps silently giving you 'unrestricted' behavior, as there is no error here).

I'm easy though. I wouldn't over think it :-)

playground/src/editor.tsx
idavis: @minestarks Do you know if it is safe to call terminate multiple times?

billti: Should be easy enough to call it twice in the 'happy path' (i.e. duplicate line 185) and see if "bad stuff" happens.

idavis: Seems to work fine. Called it 3x every time.

PR #1404, opened 2024-04-18
Support remapping of RIR qubit ids with branching

swernli:
This updates the RIR qubit reindexing pass to allow for running on RIR that includes branches so long as all paths result in the same remapping.

### DIFF
--- a/compiler/qsc_rir/src/passes/reindex_qubits.rs
+++ b/compiler/qsc_rir/src/passes/reindex_qubits.rs
@@ -4,109 +4,230 @@
 #[cfg(test)]
 mod tests;
 
+use std::collections::hash_map::Entry;
+
+use qsc_data_structures::index_map::IndexMap;
 use rustc_hash::FxHashMap;
 
-use crate::rir::{
-    Block, Callable, CallableId, CallableType, Instruction, Literal, Operand, Program, Ty,
+use crate::{
+    builder,
+    rir::{Block, BlockId, CallableId, CallableType, Instruction, Literal, Operand, Program, Ty},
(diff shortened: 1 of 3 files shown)


compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -4,110 +4,226 @@
...
+}
+
+impl ReindexQubitPass {
+    fn reindex_qubits_in_block(
+        &mut self,
+        program: &Program,
+        block: &mut Block,
+        qubit_map: &mut BlockQubitMap,
+    ) {
+        let instrs = block.0.drain(..).collect::<Vec<_>>();
```
minestarks, line 151:
```suggestion
        let instrs = std::mem::take(&mut block.0);
```

seems more to-the-point (and maybe more efficient?)

compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -157,15 +273,20 @@ fn add_mz(program: &mut Program) -> CallableId {
...
+fn add_cx(program: &mut Program) -> CallableId {
+    let cx_id = CallableId(
+        program
+            .callables
+            .iter()
+            .map(|(id, _)| id.0)
+            .max()
+            .expect("should be at least one callable")
+            + 1,
+    );
```
minestarks, line 289: Just noting that I spent about 10 minutes poking around thinking "there's got to be a better way".... but... no, there isn't.

minestarks (approved):

PR #1405, opened 2024-04-18
Truncate decimal to 4 places in alternate state visualization

Manvi-Agrawal:
## Code Path
`get_latex` -> `write_latex_for_term` -> `write_latex_for_real_number` -> `write_latex_for_decimal_number`

## Codespace Playground Link
https://miniature-telegram-j4vjv94qqv53pv6r-5555.app.github.dev/

## Changes Done
Truncating decimal output for state visualization for katas
image

## Scope: State Visualization in Katas
This flow doesn't impact any other outputs, like one from Python which shows 4 decimal digits accuracy.
image

### DIFF
--- a/compiler/qsc_eval/src/state.rs
+++ b/compiler/qsc_eval/src/state.rs
@@ -483,7 +483,8 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        write!(latex, "{}", (number.value * 10000.0).round() / 10000.0).expect("Expected to write decimal value.");
+        write!(latex, "{}", (number.value * 10000.0).round() / 10000.0)
+            .expect("Expected to write decimal value.");
     }
 }
 
--- a/compiler/qsc_eval/src/state.rs
+++ b/compiler/qsc_eval/src/state.rs
@@ -484,7 +484,7 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
         /// Using round() instead of neater string formatting({:.4})
-        /// because we do not want trailing zeros(We need 0.5 and not 0.5000)
+        /// because we do not want trailing zeros (we need 0.5 and not 0.5000)
         write!(latex, "{}", (number.value * 10000.0).round() / 10000.0)
             .expect("Expected to write decimal value.");
     }
--- a/compiler/qsc_eval/src/state.rs
+++ b/compiler/qsc_eval/src/state.rs
@@ -483,8 +483,8 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        /// Using round() instead of neater string formatting({:.4})
-        /// because we do not want trailing zeros (we need 0.5 and not 0.5000)
+        // Using round() instead of neater string formatting({:.4})
+        // because we do not want trailing zeros (we need 0.5 and not 0.5000)
         write!(latex, "{}", (number.value * 10000.0).round() / 10000.0)
             .expect("Expected to write decimal value.");
     }
--- a/compiler/qsc_eval/src/state.rs
+++ b/compiler/qsc_eval/src/state.rs
@@ -483,7 +483,7 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        write!(latex, "{:.4}", number.value).expect("Expected to write decimal value.");
+        write!(latex, "{}", (number.value * 10000.0).round() / 10000.0).expect("Expected to write decimal value.");
     }
 }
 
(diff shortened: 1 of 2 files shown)


compiler/qsc_eval/src/state/tests.rs
```diff
@@ -566,14 +566,14 @@ fn assert_latex_for_decimal(expected: &Expect, number: f64, render_one: bool) {
 fn check_get_latex_for_decimal() {
     assert_latex_for_decimal(
         &expect!([r#"
-            "0.25"
+            "0.2500"
         "#]),
         0.25,
```
Manvi-Agrawal, line 571: @tcNickolas , I am talking about these test cases. I tried out some basic superpositions, they are rendering fractions. 

Unrelated bug, for large states, it seems we omit the part of $|\psi\rangle = ... $.  

```qsharp
namespace Sample {

    @EntryPoint()
    operation Superposition() : Unit {
        // Qubits are only accesible for the duration of the scope where they
        // are allocated and are automatically released at the end of the scope.
        
        let N=8;
        use qubit = Qubit[N];

        for i in 0..N-1{
            H(qubit[i]);
        }

        DumpMachine();

        ResetAll(qubit);
    }
}
```

compiler/qsc_eval/src/state/tests.rs
```diff
@@ -647,9 +647,9 @@ fn check_get_latex_for_real() {
     );
     assert_latex_for_real(
         &expect!([r#"
-            "0.00025"
+            "0.0025"
         "#]),
-        1.0 / 4000.0,
+        1.0 / 400.0,
```
tcNickolas, line 652: Why do you change the value used in the test itself, not its output? The number 1/4000 should also have a chance to be printed

tcNickolas:
Looking at these test cases, I start to feel that it is indeed better to truncate the trailing zeros, showing 1 as 1.0000 does look odd. My main goal for this ask was to shorten the display of 15-digit floating-point numbers, not to lengthen the display of 1- or 2-digit ones. If the code that does that is a bit ugly, I think we can live with it - people look at the code results more often than at the code itself.

compiler/qsc_eval/src/state/tests.rs
```diff
@@ -594,7 +594,7 @@ fn check_get_latex_for_decimal() {
     );
     assert_latex_for_decimal(
         &expect!([r#"
-            "1"
+            "1.0000"
```
billti, line 597: If the value is 1, seems we'd want to show `1`, not `1.0000`.

Manvi-Agrawal: Changed, I was kind of trying to match it with Python's output when printing state for consistency in state visualization.
image

compiler/qsc_eval/src/state/tests.rs
Manvi-Agrawal: Done. Now this would be 0.0003 because of truncation to 4 decimal points. In general, I am not a big fan of testing with decimals, especially non-terminating.  I have seen weird test failures(passes on one, fails on other) due to slight change in decimal accuracy even with tolerances :-(

compiler/qsc_eval/src/state/tests.rs
Manvi-Agrawal: Indeed, you were right. I kind of digged into code to find this entry point :-). Thanks @tcNickolas for clearing my doubt.
https://github.com/microsoft/qsharp/blob/28817e401664ad2ab4f53f526f37ac00e310dd69/compiler/qsc_eval/src/state.rs#L344-L348

tcNickolas (approved):
Looks good to me! We still need a code owner approval, though.

Thank you!

compiler/qsc_eval/src/state.rs
```diff
@@ -483,7 +483,8 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        write!(latex, "{}", number.value).expect("Expected to write decimal value.");
+        write!(latex, "{}", (number.value * 10000.0).round() / 10000.0)
```
minestarks, line 486: Please refer to how the precision is set for formatting elsewhere in this file: 
https://github.com/microsoft/qsharp/blob/243ea46dd01bac3076f09a214ddb53782b09ce11/compiler/qsc_eval/src/state.rs#L27

You shouldn't have to round manually - Rust provides formatting specifiers for this purpose. `{:.4}` is the formatting specifier used to mean "4 digits after the decimal point". See Rust documentation to learn more: https://doc.rust-lang.org/std/fmt/#precision

minestarks (changes requested):

minestarks (approved):
Please add a line comment explaining why you had to write it this way instead `{.4}` (so that it doesn't confuse other naive readers like me ;))

Solution looks good to me. Thanks for the fix!

compiler/qsc_eval/src/state.rs
```diff
@@ -483,7 +483,10 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        write!(latex, "{}", number.value).expect("Expected to write decimal value.");
+        /// Using round() instead of neater string formatting({:.4})
+        /// because we do not want trailing zeros(We need 0.5 and not 0.5000)
```
tcNickolas, line 487:
```suggestion
        /// because we do not want trailing zeros (we need 0.5 and not 0.5000)
```

compiler/qsc_eval/src/state.rs
```diff
@@ -483,7 +483,10 @@ fn write_latex_for_real_number(latex: &mut String, number: &RealNumber, render_o
 /// 1 is only rendered if ``render_one`` is true.
 fn write_latex_for_decimal_number(latex: &mut String, number: &DecimalNumber, render_one: bool) {
     if render_one || is_significant(number.value - 1.0) {
-        write!(latex, "{}", number.value).expect("Expected to write decimal value.");
+        /// Using round() instead of neater string formatting({:.4})
+        /// because we do not want trailing zeros (we need 0.5 and not 0.5000)
```
Manvi-Agrawal, line 487:
```suggestion
        // Using round() instead of neater string formatting({:.4})
        // because we do not want trailing zeros (we need 0.5 and not 0.5000)
```

PR #1415, opened 2024-04-23
Use hierarchical namespaces in name resolution

### LINKED ISSUE #1336: Language Service no longer needs to auto-add `open Microsoft.Quantum.Measurement;`

sezna:
_This PR replaces #1364_

This PR introduces a feature called _hierarchical namespaces_, a module-like system that maintains backward compatibility. It is the foundational groundwork that will pave the way for:
1. Naked namespaces (files without namespace declarations)
2. Package publishing and sharing
3. Explicit item imports
4. Exports and re-exports
5. and the rest of the stuff [here](https://github.com/microsoft/qsharp/issues/493#issuecomment-1979722979)

Key changes include:
1. `VecIdent` type introduction: This abstraction over `Vec<Ident>` represents the new namespaces in the AST. Alternate names could be `DotIdent` or `Path`, but those names were already loaded with meaning in our codebase. I'm open to feedback on nomenclature here.
7. `NamespaceId` usage: This replaces our previous flat map-style `Rc<str> -> Res` item tracking.
8. `NamespaceTree`: This tracks the namespace hierarchy separately in low-cost data structure, and allows us to keep our basic `tys` and `terms` maps relatively untouched.
9. Name resolution algorithm update: The algorithm in `resolve.rs` received a heavy facelift. Not only did I try to simplify the existing code, but I also added the required functionality to check candidate partially-opened or re-opened namespaces. 

This PR maintains 100% backward compatibility with the existing resolution engine, preserving features like re-opening namespaces, shadowing semantics, merging namespaces via aliases, and absolute-path opens. It also introduces the ability to reference a function in a different namespace if its parent namespace is opened.

Additional notes:

1.  Katas used a form of shadowing that wasn't tested in our compiler. This PR [adds a test](https://github.com/microsoft/qsharp/pull/1312/files#diff-7a3aa16f18f6d132b9e54a6b6ebd53e6c3b41cea805c8f8db9c456b533a3c96bR2787) for that.
2. Pursuing #1321 could help differentiate user-facing and compiler-engineer-facing string representations of structs, which are increasingly conflated.
3. There's potential for refactoring `ast`, `fir`, and `hir` to use shared generic types, reducing duplicate `Ident`/etc types.
4. Completions functionality remains the same, but it's not yet aware of the ability to partially open namespaces (e.g. `open Microsoft.Quantum`. This will be addressed in future updates.
5. There is a performance regression, but we are performing a full tree-traversal namespace resolution algorithm in a place where we used to just do a constant table lookup. I have optimized this as much as I believe is reasonable for now, bringing the regression down from ~40%, where it was originally, to around 5-10% based on the benchmark run.

Also closes #1336


compiler/qsc_passes/src/invert_block.rs
```diff
@@ -329,14 +329,17 @@ fn make_range_field(range_id: NodeId, field: PrimField) -> Expr {
 }
 
 fn make_array_index_range_reverse(core: &Table, arr_id: NodeId, arr_ty: &Ty) -> Expr {
+    let ns = core
```
idavis, line 332: this vec can be a static array which is passed in by ref avoiding the allocation.

compiler/qsc_passes/src/loop_unification.rs
```diff
@@ -132,9 +132,13 @@ impl LoopUni<'_> {
         let Ty::Array(item_ty) = &array_id.ty else {
             panic!("iterator should have array type");
         };
+        let ns = self
```
idavis, line 135: Same as invert block. Create static array instead of new vecs.

compiler/qsc_ast/src/ast.rs
```diff
@@ -1306,6 +1306,119 @@ pub struct Ident {
     pub name: Rc<str>,
 }
 
+/// A [`VecIdent`] represents a sequence of idents. It provides a helpful abstraction
+/// that is more powerful than a simple `Vec<Ident>`, and is primarily used to represent
+/// dot-separated paths.
```
billti, line 1311: If perf continues to be a concern, there may be other structures we can look at. For example, VS Code uses implementations of `src/vs/base/common/ternarySearchTree.ts` and `src/vs/base/common/prefixTree.ts` to optimize some lookups (though they are a real pain when debugging)

compiler/qsc_fir/src/fir.rs
```diff
@@ -1423,6 +1423,89 @@ impl Display for PatKind {
     }
 }
 
+/// A [`VecIdent`] represents a sequence of idents. It provides a helpful abstraction
+/// that is more powerful than a simple `Vec<Ident>`, and is primarily used to represent
+/// dot-separated paths.
```
billti, line 1428: This looks familiar :-/

compiler/qsc_fir/src/global.rs
```diff
@@ -5,12 +5,15 @@ use crate::{
...
+use qsc_data_structures::{
+    index_map,
+    namespaces::{NamespaceId, NamespaceTreeRoot},
+};
 use rustc_hash::FxHashMap;
 use std::rc::Rc;
 
 pub struct Global {
-    pub namespace: Rc<str>,
+    pub namespace: Vec<Rc<str>>,
```
billti, line 16: A `NamespaceTreeRoot` _sounds like_ the kind of type this should be? I'm surprised to see a vector of strings here after the addition of other namespace types.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -17,7 +17,7 @@ mod tests;
 #[derive(PartialEq, Hash, Clone, Debug)]
 pub struct TrackedName {
     pub name: Rc<str>,
-    pub namespace: Rc<str>,
+    pub namespace: Vec<Rc<str>>,
```
billti, line 20: I'm a little confused as to when we use the new abstract types to track a namespace path, versus still using a list of strings.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -54,12 +54,22 @@ impl MutVisitor for Conditional {
...
+                                    .map(|x| Rc::from(x.to_string()))
+                                    .collect(),
                             });
                         }
                         ItemKind::Ty(ident, _) => self.included_names.push(TrackedName {
                             name: ident.name.clone(),
-                            namespace: namespace.name.name.clone(),
+                            namespace: namespace
+                                .name
+                                .clone()
```
billti, line 69: Same as above.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -69,12 +79,17 @@ impl MutVisitor for Conditional {
...
-                                namespace: namespace.name.name.clone(),
+                                namespace: (&namespace.name).into(),
                             });
                         }
                         ItemKind::Ty(ident, _) => self.dropped_names.push(TrackedName {
                             name: ident.name.clone(),
-                            namespace: namespace.name.name.clone(),
+                            namespace: namespace
+                                .name
+                                .clone()
```
billti, line 89: And again.

compiler/qsc_frontend/src/lower.rs
```diff
@@ -736,6 +736,10 @@ impl With<'_> {
             new_id
         })
     }
+
+    fn lower_vec_ident(&mut self, name: &ast::VecIdent) -> hir::VecIdent {
+        name.iter().cloned().map(|i| self.lower_ident(&i)).collect()
```
billti, line 741: Is the cloned needed here also? Or does `name.iter().map(|i| self.lower_ident(i)).collect()` work?

language_service/src/completion.rs
```diff
@@ -272,9 +277,9 @@ impl CompletionListBuilder {
     fn push_globals(
         &mut self,
         compilation: &Compilation,
-        opens: &[(Rc<str>, Option<Rc<str>>)],
+        opens: &[(NamespaceName, Option<NamespaceAlias>)],
         insert_open_range: Option<Range>,
-        current_namespace_name: &Option<Rc<str>>,
+        current_namespace_name: &Option<Vec<Rc<str>>>,
```
minestarks, line 282: AFAICT you're not changing any functionality in completions. Wouldn't it be much more straightforward and backwards-compatible if you left these as `Rc<str>`, and initialized them to the dotted names (`namespace.name.name()`)? `get_callables` doesn't need to have awareness of these `VecIdent`s since it's not doing anything special with them.

language_service/src/code_lens.rs
```diff
@@ -42,7 +42,7 @@ pub(crate) fn get_code_lenses(
                     .and_then(|parent_id| user_unit.package.items.get(parent_id))
                     .map(|parent| &parent.kind)
                 {
-                    let namespace = ns.name.to_string();
+                    let namespace = ns.name().to_string();
```
minestarks, line 45: Already a `String`, isn't it?

minestarks (commented):
Not done looking at all the files, but posting my first set of comments

compiler/qsc_ast/src/ast.rs
```diff
@@ -1306,6 +1306,122 @@ pub struct Ident {
...
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Default)]
+pub struct VecIdent(pub Vec<Ident>);
```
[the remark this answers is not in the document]
minestarks: This and the rest of my `VecIdent` related comments apply to both the AST and HIR versions

compiler/qsc_hir/src/hir.rs
```diff
@@ -1135,6 +1134,113 @@ impl Display for QubitInitKind {
...
+}
+
+impl FromIterator<Ident> for VecIdent {
+    fn from_iter<T: IntoIterator<Item = Ident>>(iter: T) -> Self {
+        VecIdent(iter.into_iter().collect())
+    }
+}
+
+impl Display for VecIdent {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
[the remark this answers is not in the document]
minestarks: Oh, I missed that - yes that's totally reasonable. Resolving

compiler/qsc_ast/src/ast.rs
```diff
@@ -164,7 +164,7 @@ pub struct Namespace {
     /// The documentation.
     pub doc: Rc<str>,
     /// The namespace name.
-    pub name: Box<Ident>,
+    pub name: VecIdent,
```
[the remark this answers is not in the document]
minestarks: OK. I can easily see it both ways, so I'll leave it up to you. I wanted to bring it up as a "have you thought about it this way?" thing. If it's not actually a helpful abstraction, there's no need for it.

On the one hand, I still think we can legitimately call it a Path - a Path is any identifier of the form `<optional.parent.namespace>.<ident>`. To me, the name in the namespace declaration fits this description exactly. The part before the last `.` is still a namespace, it's the _parent_ namespace.

On the other hand, `Path`s only refer to item identifiers so maybe it's best not to mix them up with namespaces. Would we ever want `visit_path` to cover namespace declarations? Probably not. They're semantically different enough. Fair enough :)

As for the "1-or-more" invariant - I don't see a need to codify it... I mainly brought that up to illustrate the similarities with `Path`. Thanks for explaining that bit.

compiler/qsc_ast/src/ast.rs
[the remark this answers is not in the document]
minestarks: > I'm curious why it conveys intent better? Lack of dynamic sizing?

Yeah, to me `Box<[Ident]>` gives "immutable array" vibes while `Vec<Ident>` gives "I might add to this later", since one of the core attributes of `Vec` is that it "capacity" and "length" are decoupled and over-allocating is the default.

compiler/qsc_fir/src/fir.rs
minestarks: Do we really need the vector representation in the FIR though? One advantage of lowering is that you get to abstract away details that don't matter at this level. Keeping the parts separately in a vector is useful when you're doing name resolution, but once we're in the FIR we're done with all that. We can keep the whole thing as a single string.

IMO this means you can revert all the changes to the `qsc_fir` folder and do away with `Idents` at the lowering step. Like this: https://github.com/microsoft/qsharp/commit/bed6b712133b0b687fd4cf34ac1ea5c20cda9858

compiler/qsc_frontend/src/compile.rs
```diff
@@ -54,6 +55,7 @@ pub struct AstPackage {
     pub tys: Table,
     pub names: Names,
     pub locals: Locals,
+    pub namespaces: NamespaceTreeRoot,
```
minestarks, line 58: This member doesn't seem to be used anywhere.

compiler/qsc_ast/src/ast.rs
```diff
@@ -1306,6 +1306,130 @@ pub struct Ident {
...
+        Span {
+            lo: self.0.first().map(|i| i.span.lo).unwrap_or_default(),
+            hi: self.0.last().map(|i| i.span.hi).unwrap_or_default(),
+        }
+    }
+
+    /// The stringified dot-separated path of the idents in this [`Idents`]
+    /// E.g. `a.b.c`
+    #[must_use]
+    pub fn name(&self) -> String {
```
minestarks, line 1421: Not sure how much it matters, but it would be great if this returned a `Rc<str>` since some callers store it as an `Rc<str>` anyway, and conversions incur a couple of unnecessary allocations.

Bonus: If the array only has 1 element, it's just an `Rc::clone()` - zero allocations!

compiler/qsc_parse/src/item.rs
```diff
@@ -136,15 +136,15 @@ fn parse_namespace(s: &mut ParserContext) -> Result<Namespace> {
...
+    let name = path(s)?;
     token(s, TokenKind::Open(Delim::Brace))?;
     let items = barrier(s, &[TokenKind::Close(Delim::Brace)], parse_many)?;
     recovering_token(s, TokenKind::Close(Delim::Brace));
     Ok(Namespace {
         id: NodeId::default(),
         span: s.span(lo),
         doc: doc.into(),
-        name,
+        name: (*name).into(),
```
minestarks, line 147: I admit I'm not a huge fan of having a `From<T>`  implementation for each type we may convert the `Idents` to, defined alongside `Idents`.

This is one example. I see this is coming from `impl From<Path>`  defined in `ast.rs`. It's three lines of code that's only used here, that could easily just live in this function. It makes the code less readable to see `into()` here, have to check the types, and then go search in a different file to understand what's actually going on (which is a nontrivial conversion).

compiler/qsc_hir/src/global.rs
```diff
@@ -125,16 +153,16 @@ impl PackageIter<'_> {
...
                     visibility: item.visibility,
                     status,
                     kind: Kind::Ty(Ty { id }),
                 })
             }
             (ItemKind::Namespace(ident, _), None) => Some(Global {
-                namespace: "".into(),
-                name: Rc::clone(&ident.name),
+                namespace: ident.into(),
+                name: "".into(),
```
ScottCarda-MS, line 165: Do we want a bit more logic here? I would think that we'd want to be more explicit about namespaces that are "items" of other namespaces, such that Namespace globals would have their `name` field be the namespace name, and their `namespace` field be their parent namespace name.

compiler/qsc_fir/src/mut_visit.rs
```diff
@@ -49,6 +50,8 @@ pub trait MutVisitor<'a>: Sized {
 
     fn visit_ident(&mut self, _: &'a mut Ident) {}
 
+    fn visit_idents(&mut self, _: &'a mut Idents) {}
```
ScottCarda-MS, line 53: Same comment here as in the ast and hir: the need to have a `walk_idents`.

compiler/qsc_fir/src/visit.rs
```diff
@@ -49,6 +50,8 @@ pub trait Visitor<'a>: Sized {
 
     fn visit_ident(&mut self, _: &'a Ident) {}
 
+    fn visit_idents(&mut self, _: &'a Idents) {}
```
ScottCarda-MS, line 53: Same comment here as in the ast and hir: the need to have a `walk_idents`.

.gitignore
```diff
@@ -17,3 +17,4 @@ __pycache__/
 /fuzz/Cargo.lock
 .mypy_cache/
 .pytest_cache/
+.idea/
```
ScottCarda-MS, line 20: Is this something on your local machine? You can use the .git\info\exclude file to exclude local things without effecting the repo.

compiler/qsc_parse/src/prim.rs
```diff
@@ -107,15 +89,17 @@ pub(super) fn path(s: &mut ParserContext) -> Result<Box<Path>> {
...
-        (Some(first), Some(last)) => {
-            let lo = first.span.lo;
-            let hi = last.span.hi;
-            Some(Box::new(Ident {
-                id: NodeId::default(),
-                span: Span { lo, hi },
-                name: join(parts.iter().map(|i| &i.name), ".").into(),
-            }))
-        }
+        (Some(_), Some(_)) => Some(
```
ScottCarda-MS, line 92: I'm not sure doing `(parts.first(), parts.last())` has the same purpose anymore. I think it would be clearer to have `parts.is_empty()`.

compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -283,17 +282,17 @@ fn call_generic_length() {
...
-            #8 59-68 "xs : 'T[]" : ?
-            #17 98-125 "Length([true, false, true])" : Int
-            #18 98-104 "Length" : (Bool[] -> Int)
-            #21 104-125 "([true, false, true])" : Bool[]
-            #22 105-124 "[true, false, true]" : Bool[]
-            #23 106-110 "true" : Bool
-            #24 112-117 "false" : Bool
-            #25 119-123 "true" : Bool
-        "#]],
+        &expect![[r##"
```
ScottCarda-MS, line 285: nit: The expects still occasionally put the double `##` in when they aren't needed. Please remove the extra `#` on the expect of these tests.

ScottCarda-MS (commented):
Still need to review the resolver changes and the new namespaces.rs file. Other than that, so far looks pretty good.

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+        // if it wasn't found, recurse into children
+        for (name, node) in &self.children {
+            let mut names = names_buf.to_vec();
+            names.push(name.clone());
+            let Some((names, node)) = node.borrow().find_namespace_by_id(id, &names) else {
+                continue;
+            };
+            if !names.is_empty() {
+                return Some((names, node));
+            }
```
ScottCarda-MS, line 299: Does this case ever catch? `names` is pushed to above, and I didn't see that there was anything that pops from names in recursion.

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -0,0 +1,365 @@
+use expect_test::expect;
+
+use super::*;
+
+#[allow(clippy::too_many_lines)]
+#[test]
+fn test_tree_construction() {
+    let mut root = NamespaceTreeRoot::default();
```
ScottCarda-MS, line 8: It might be better for testing purposes if there was a way to construct a default tree root without the prelude.

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -0,0 +1,365 @@
...
+                value: {},
+            },
+        }
+    "#]]
+    .assert_debug_eq(&root);
+}
+
+#[allow(clippy::too_many_lines)]
+#[test]
+fn test_find_id() {
```
ScottCarda-MS, line 187: Could you modify this test to also search for the intermediate-level namespaces: "ns0", "ns1", and "ns2"?

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -0,0 +1,365 @@
...
+                    },
+                },
+            ),
+        ]
+    "#]]
+    .assert_debug_eq(&result_buf);
+}
+// test that after inserting lots of namespaces, all ids are unique and sequential
+#[test]
+fn test_insert_or_find_namespace() {
```
ScottCarda-MS, line 335: This test makes me realize that `insert_or_find_namespace` can create multiple nodes, creating multiple ids, but only returns the id of the leaf node created. But I suppose a feature of this function is that the caller doesn't need to know if the namespace nodes exist or not prior to the call.

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+                let mut names = names_buf.to_vec();
+                names.push(name.clone());
+                return Some((names, node.clone()));
+            }
+        }
+
+        // if it wasn't found, recurse into children
+        for (name, node) in &self.children {
+            let mut names = names_buf.to_vec();
+            names.push(name.clone());
```
ScottCarda-MS, line 293: Just checking, does the `names_buf.to_vec()` create a clone of the names, such that the `names.push(name.clone())` won't pollute into the next iteration of the for loop?

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -0,0 +1,365 @@
+use expect_test::expect;
```
ScottCarda-MS, line 1: Could you add a test for `get_namespace_id`?

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+            } else {
+                return None;
+            }
+        }
+        Some(buf.map_or_else(|| self.id, |x| x.borrow().id))
+    }
+
+    /// Inserts a new namespace into the tree, if it does not yet exist.
+    /// Returns the ID of the namespace.
+    pub fn insert_or_find_namespace<I>(
```
ScottCarda-MS, line 239: It might be worth noting in the doc string that this function only returns `None` when given an empty `iter`.

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+        ns: impl IntoIterator<Item = Rc<str>>,
+    ) -> NamespaceId {
+        self.tree
+            .borrow_mut()
+            .insert_or_find_namespace(ns.into_iter().peekable(), &mut self.assigner)
+            .expect("namespace creation should not fail")
+    }
+
+    /// Create a new namespace node with the given children.
+    pub fn new_namespace_node(
```
ScottCarda-MS, line 115: This function creates a new node, but doesn't insert it anywhere into the tree, though it _does_ increment the assigner. This seems like dangerous behavior for a pub function.

ScottCarda-MS: I'm not seeing this used anywhere. Might be better to not have this function.

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    /// Insert a namespace into the tree. If the namespace already exists, return its ID.
+    pub fn insert_or_find_namespace(
+        &mut self,
+        ns: impl IntoIterator<Item = Rc<str>>,
+    ) -> NamespaceId {
+        self.tree
+            .borrow_mut()
+            .insert_or_find_namespace(ns.into_iter().peekable(), &mut self.assigner)
+            .expect("namespace creation should not fail")
```
ScottCarda-MS, line 111: I believe this will fail if given an empty ns iterator.

compiler/qsc_data_structures/src/namespaces.rs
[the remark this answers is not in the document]
ScottCarda-MS: I actually rather like `qsc_data_structures` for the location. It is a somewhat involved data structure that gets used in a variety of places to represent all the namespaces of a compilation.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -984,106 +1089,326 @@ fn decl_is_intrinsic(decl: &ast::CallableDecl) -> bool {
     }
 }
 
+/// Resolves a given symbol and namespace name, according to the Q# shadowing rules.
```
ScottCarda-MS, line 1092: I really appreciate this doc comment along with the example. Thanks!

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -984,106 +1089,326 @@ fn decl_is_intrinsic(decl: &ast::CallableDecl) -> bool {
...
-                // Explicit opens shadow prelude and unopened globals.
-                break;
-            }
-        }
-
-        if scope.kind == ScopeKind::Callable {
-            // Since local callables are not closures, hide local variables in parent scopes.
-            vars = false;
-        }
+    // let scopes = scopes.collect::<Vec<_>>();
```
ScottCarda-MS, line 1111: lingering commented-out code

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -756,15 +833,23 @@ impl GlobalTable {
         for (name, res) in builtins {
             core.insert(name, res);
         }
-        let mut tys: FxHashMap<Rc<str>, FxHashMap<Rc<str>, Res>> = FxHashMap::default();
-        tys.insert("Microsoft.Quantum.Core".into(), core);
+
+        let mut scope = GlobalScope::default();
+        let ns = scope.insert_or_find_namespace(vec![
```
ScottCarda-MS, line 838: It could be helpful if you could centralize the actual name of these prelude namespaces near, or in the prelude structure. That way if we ever want to change the name of the namespace (like remove the "Microsoft.Quantum" part), we'd only need to change it in the centralized location.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -463,20 +510,37 @@ impl Resolver {
...
+                    .map(|x| x.name.to_string())
+                    .collect::<Vec<_>>()
+                    .join("."),
+                name.span(),
+            ));
+            return;
+        };
+        let alias = alias
+            .as_ref()
+            .map_or(name.into(), |a| vec![Rc::clone(&a.name)]);
```
ScottCarda-MS, line 526: I'm noticing that the default value for aliases changed here. It use to default to an empty string, which makes sense because in that case there is no alias. Now, the default is to have the alias match the regular namespace name. Looking at how this is used, as a key to the opens entries map, I'm not actually sure how this worked before, do you know? It seems like whenever an open without an alias is used, it would _always_ push a new open?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -463,20 +510,37 @@ impl Resolver {
...
+        if self
+            .globals
+            .namespaces
+            .get_namespace_id(name.str_iter())
+            .is_some()
+        {
             self.current_scope_mut()
                 .opens
                 .entry(alias)
                 .or_default()
```
ScottCarda-MS, line 536: What is the purpose of the `or_default()` here? Doesn't that just create a mutable reference to an empty vector that will get thrown away at the end of the statement?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -984,106 +1089,326 @@ fn decl_is_intrinsic(decl: &ast::CallableDecl) -> bool {
...
 
-    if candidates.len() > 1 {
-        // If there are multiple candidates, remove unimplemented items. This allows resolution to
-        // succeed in cases where both an older, unimplemented API and newer, implemented API with the
-        // same name are both in scope without forcing the user to fully qualify the name.
-        let mut removals = Vec::new();
-        for res in candidates.keys() {
-            if let Res::Item(_, ItemStatus::Unimplemented) = res {
-                removals.push(*res);
+/// This function always returns an `Err` variant of `Result`. The error is of type `Error::Ambiguous` and contains
```
ScottCarda-MS, line 1295: It's a bit awkward to have a function that returns Result when it is intended to always return an Error. Since it is only ever called in one place, it might be better to have this function return `Error::Ambiguous` without the Result::Err wrapper, and have that wrapping take place at the call site.

compiler/qsc_lowerer/src/lib.rs
```diff
@@ -156,7 +156,19 @@ impl Lowerer {
     fn lower_item(&mut self, item: &hir::Item) -> fir::Item {
         let kind = match &item.kind {
             hir::ItemKind::Namespace(name, items) => {
-                let name = self.lower_ident(name);
+                let name = fir::Ident {
+                    // We don't have a node ID for the whole dotted name,
+                    // but we can use the node ID of the last part.
+                    // This ident is not a local anyway. (@swernli should check if this is harmless)
```
ScottCarda-MS, line 162: @swernli for the check, afterwards that part of the comment can be removed.

ScottCarda-MS (approved):

compiler/qsc_frontend/src/resolve.rs
ScottCarda-MS: I see, so it acts as a kind of find_or_insert.

compiler/qsc_fir/src/mut_visit.rs
ScottCarda-MS: removed

compiler/qsc_fir/src/visit.rs
ScottCarda-MS: removed

compiler/qsc_hir/src/mut_visit.rs
```diff
@@ -49,8 +49,15 @@ pub trait MutVisitor: Sized {
     }
 
     fn visit_span(&mut self, _: &mut Span) {}
+    fn visit_idents(&mut self, ident: &mut crate::hir::Idents) {
+        walk_idents(self, ident);
+    }
```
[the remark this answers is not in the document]
ScottCarda-MS: The thought was to put `Idents` next to `Ident`. The ordering for the rest doesn't need to change.

compiler/qsc_hir/src/mut_visit.rs
```diff
@@ -49,8 +49,15 @@ pub trait MutVisitor: Sized {
...
+}
+pub fn walk_idents(vis: &mut impl MutVisitor, ident: &mut crate::hir::Idents) {
```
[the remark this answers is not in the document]
ScottCarda-MS: This is to put it next to the logic for `Ident` and to have the order of the walk functions match the order of the visit functions.

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,304 @@
...
+/// ```
+/// This data structure is optimized for looking up namespace IDs by a given name. Looking up a namespace name by ID is
+/// less efficient, as it performs a breadth-first search. Because of this inefficiency, the results of this lookup are memoized.
+/// [`NamespaceTreeNode`]s are all stored in [`NamespaceTreeCell`]s, which are reference counted and support interior mutability for namespace
+/// insertions and clone-free lookups.
+#[derive(Debug, Clone)]
+pub struct NamespaceTreeRoot {
+    assigner: usize,
+    tree: NamespaceTreeCell,
+    memo: RefCell<FxHashMap<NamespaceId, MemoEntry>>,
```
minestarks, line 83: Not an objection, but just curious: Why is this inside a `RefCell`? My guess is "to avoid updating a bunch of existing places to be `&mut` references" - is that right?

compiler/qsc_passes/src/lib.rs
```diff
@@ -36,6 +36,9 @@ use qsc_rca::{PackageComputeProperties, PackageStoreComputeProperties};
 use replace_qubit_allocation::ReplaceQubitAllocation;
 use thiserror::Error;
 
+pub(crate) static CORE_NAMESPACE: &[&str] = &["Microsoft", "Quantum", "Core"];
+pub(crate) static QIR_RUNTIME_NAMESPACE: &[&str] = &["QIR", "Runtime"];
```
minestarks, line 40: I see there are a few spots where we do `.find_namespace(WELL_KNOWN_NAMESPACE.iter().copied())` . Feels like a lot of work when the namespace ID is constant each time. Can we maybe cache the found namespace ID with these statics?

minestarks: Alternatively, these namespaces could be extra special, and their IDs can be exposed directly from `Table` ?

compiler/qsc_ast/src/ast.rs
minestarks: Same comment for `hir.rs`

compiler/qsc_lowerer/src/lib.rs
swernli: Yes, this looks safe, because it effectively matches what we were doing before (creating a mapping from an `Ident` to a `LocalVarId`), and it's true that this one should not get used as a real local variable anywhere because of how name resolution works for namespaces (and items, as it turns out).

swernli: The comment can be removed.

compiler/qsc_parse/src/lib.rs
```diff
@@ -74,6 +74,9 @@ enum ErrorKind {
     #[error("missing entry in sequence")]
     #[diagnostic(code("Qsc.Parse.MissingSeqEntry"))]
     MissingSeqEntry(#[label] Span),
+    #[error("dotted namespace aliases are no longer allowed")]
```
minestarks, line 77: May be just me but I prefer this:

```suggestion
    #[error("dotted namespace aliases are not allowed")]
```

most likely, this error message will stick around forever and whoever sees it won't know these were ever allowed.

compiler/qsc_parse/src/item/tests.rs
```diff
@@ -1693,3 +1682,37 @@ fn namespace_with_conflicting_names() {
                         Type _id_ [87-90]: Path: Path _id_ [87-90] (Ident _id_ [87-90] "Int")"#]],
     );
 }
+
+// We technically broke this syntax. Although we don't think anybody was using it,
```
minestarks, line 1686: Can you be more specific as to when, e.g. "as of v1.3.1" or "May 2024".

This may be outdated thinking, but I've just read one too many comments like "we still need to support this for another few months!" and I had no way of knowing if that's from 2022 or 1996

compiler/qsc_hir/src/global.rs
ScottCarda-MS: I'm satisfied with your double-checking this logic in response to me comment. Thanks for double checking this.

compiler/qsc_passes/src/lib.rs
minestarks: Yeah... we're getting into diminishing returns. 

If we're doing _anything_ about it, I'd literally hardcode namespace IDs `1` and `2` for these well known namespaces, and just insert them into the `Table` at creation time. Anything more complex than that would be silly.

But even that.... probably not worth the trouble at the moment.

compiler/qsc_data_structures/src/namespaces/tests.rs
ScottCarda-MS: It's not absolutely necessary, so I'll go ahead an resolve this comment.

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -0,0 +1,566 @@
...
+
+use super::*;
+
+#[allow(clippy::too_many_lines)]
```
ScottCarda-MS, line 5: Looks like you'll need to add `#![allow(clippy::needless_raw_string_hashes)]` to make clippy happy with this file.

compiler/qsc_data_structures/src/index_map.rs
```diff
@@ -14,6 +14,23 @@ pub struct IndexMap<K, V> {
...
 
+impl<K, V> IndexMap<K, V>
+where
+    K: Into<usize>,
+    V: Default,
+{
+    pub fn get_mut_or_default(&mut self, key: K) -> &mut V {
+        let index: usize = key.into();
+        if index >= self.values.len() {
+            self.values.resize_with(index + 1, Default::default);
```
minestarks, line 25: I think we want `None` here not `Default`. If there are gaps in the vector before the index you're inserting, you want them initialized with `None` to indicate there's nothing at that index. Otherwise you're implicitly initializing the gaps.

```rust
let i: IndexMap<usize, usize> = IndexMap::default(); // no elements in the index map

let e = i.get_mut_or_default(33);  // only idx 33 should be initialized

i.contains_key(22); // should be false

for (k,v) in i {
  // should yield only one element
}
```

minestarks: You could just call `self.insert(...)` to be safe.

language_service/src/completion.rs
```diff
@@ -441,35 +450,35 @@ impl CompletionListBuilder {
...
-                                            if *name == namespace.name {
+                                            if *name == Into::<Vec<_>>::into(namespace) {
                                                 Some(alias)
                                             } else {
                                                 None
                                             }
                                         });
                                         qualification = match open {
-                                            Some(alias) => alias.clone(),
+                                            Some(alias) => alias.clone().map(|x| vec![x]),
```
minestarks, line 469: `clone()` seems needless now that there's a `map()`

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -463,20 +501,37 @@ impl Resolver {
...
-        if self.globals.namespaces.contains(&name.name) {
+    fn bind_open(&mut self, name: &Idents, alias: &Option<Box<Ident>>) {
+        let Some(id) = self.globals.find_namespace(name.str_iter()) else {
+            self.errors.push(Error::NotFound(
+                name.iter()
+                    .map(|x| x.name.to_string())
+                    .collect::<Vec<_>>()
+                    .join("."),
+                name.span(),
+            ));
```
minestarks, line 512: Use the `name.name()` implementation instead of manually joining?

```suggestion
            self.errors
                .push(Error::NotFound(name.name().to_string(), name.span()));
```

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -593,14 +648,23 @@ impl With<'_> {
...
+            .globals
+            .find_namespace(namespace.name.str_iter())
+            .expect("namespace should exist by this point");
+
+        let kind = ScopeKind::Namespace(ns);
         self.with_scope(namespace.span, kind, |visitor| {
+            // the below line ensures that this namespace opens itself, in case
+            // we are re-opening a namespace. This is important, as without this,
+            // a re-opened namespace would only have knowledge of its scopes.
+            visitor.resolver.bind_open(&namespace.name, &None);
```
minestarks, line 662: I don't follow - if this is important, how did we get by without it? Items in a namespace should already be available to the resolver.

FWIW, all the tests seem to pass if I comment out this line :P . If this is truly needed, maybe a unit test case demonstrating why?

minestarks: edited my above comment, you can ignore it.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -463,20 +501,37 @@ impl Resolver {
         }
     }
 
-    fn bind_open(&mut self, name: &ast::Ident, alias: &Option<Box<ast::Ident>>) {
-        let alias = alias.as_ref().map_or("".into(), |a| Rc::clone(&a.name));
-        if self.globals.namespaces.contains(&name.name) {
+    fn bind_open(&mut self, name: &Idents, alias: &Option<Box<Ident>>) {
+        let Some(id) = self.globals.find_namespace(name.str_iter()) else {
```
minestarks, line 505: Should the rules of resolution apply to the namespace name inside an `open` statement as well? I'd expect it to.

I'd expect this to work:

```qsharp

namespace Foo.Bar {
    open Bar; // should work, but gives error
}

namespace Foo {

}

```

compiler/qsc_frontend/src/resolve.rs
minestarks, line 1: Based on yesterday's discussion I would expect this to work:

```qsharp
namespace Foo {
    function Top() : Unit {}
}

namespace Foo.Bar {
    function Fun() : Unit {}
}

namespace Foo.Bar.Baz {
    function Bar() : Unit {
        Top(); // should resolve, doesn't
        Fun(); // should resolve, doesn't
        Foo.Bar.Fun();
    }
}

```

minestarks (commented):
Looked at all the files. Will revisit once you've had a chance to look at the comments.

language_service/src/completion.rs
minestarks: Gotcha, false alarm!

compiler/qsc_data_structures/src/index_map.rs
minestarks: AAAHHH I was definitely reading this as initializing this to `Some(T::default())`. 

Nah, I wouldn't say it's unclear, just a reading mistake on my part, it happens.

compiler/qsc_frontend/src/resolve.rs
minestarks: OK! SGTM if it's intentional

minestarks (approved):
Thank you!

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /// Get the namespace tree field. This is the root of the namespace tree.
+    #[must_use]
+    pub fn tree(&self) -> NamespaceTreeCell {
+        self.tree.clone()
+    }
+
+    /// Insert a namespace into the tree. If the namespace already exists, return its ID.
+    pub fn insert_or_find_namespace(
```
cesarzc, line 104: A must use here could be useful.

compiler/qsc_parse/src/item.rs
minestarks: Discussed offline. We agreed that we may not even need to use the `Path` parser for `Idents`, which will eliminate the need for a `From<Path>` impl, making this conversation moot. We'll take this in a future PR since it'll involve the parser changes from #1469 for parsing dotted identifiers.

cesarzc (approved):

PR #1416, opened 2024-04-23
Allow implicit top-level namespaces when parsing files

sezna:
This PR allows for implicit top-level namespaces.

Prior to this PR, all Q# had to be wrapped in namespace declarations:

```qsharp
namespace Foo {
  operation Main() : Unit {}
}
```

Now, users can optionally omit the top-level namespace declaration. This is nice, because it removes a level of necessary indentation.

If the explicit namespace declaration is omitted, the file path is used to determine the namespace name. For example, in `~/code/qsharp/project_dir/src/Foo/Bar/Baz.qs`, the file:
```qsharp
operation Quux() : Unit {}
```

defines a callable in the namespace `Foo.Bar.Baz`.

We detect the "root" of the project by calculating the largest common prefix of the file paths. This is the most versatile approach, in my opinion, without having to introduce project-based logic into the compiler itself (that is, add a `qsc_project` dependency to `qsc_frontend` -- which would be cyclical via `qsc_lint`). It allows us to have a perfect UX for the happy path following a canonical project layout, and reasonable UX for other situations (e.g., manually evoking `qsc` on two files in different parts of your computer -- what is the "root"?).


minestarks:
@sezna should we update our samples to eliminate namespace declarations, since that's the recommended syntax now?

swernli:
> @sezna should we update our samples to eliminate namespace declarations, since that's the recommended syntax now?

Samples are tricky, because they are effectively "live" in that we point people to them online as reference. Given that, I think it's better to have the samples match features that are already shipped. That way we avoid the user trap of trying the samples and getting failures until we officially release a new feature.

compiler/qsc_eval/src/tests.rs
```diff
@@ -127,7 +127,7 @@ fn check_partial_eval_stmt(
     let std_fir = qsc_lowerer::Lowerer::new().lower_package(&std.package);
     let std_id = store.insert(std);
 
-    let sources = SourceMap::new([("test".into(), file.into())], Some(expr.into()));
+    let sources = SourceMap::new([("".into(), file.into())], Some(expr.into()));
```
minestarks, line 130: Is this required to get the test passing? I think I'd rather we update the test case to accommodate the file name `test`, rather than set this slightly dangerous precedence for blank file names.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -88,8 +90,26 @@ impl SourceMap {
             offset_sources.push(source);
         }
 
+        // Each source has a name, which is a string. The project root dir is calculated as the
+        // common prefix of all of the sources.
+        // Calculate the common prefix.
+        let project_root_dir: String = longest_common_prefix(
```
minestarks, line 96: This isn't necessarily true - for example, all the source files could live under `project_root/src/Microsoft/Quantum/`. Best to pass in the project root explicitly from whatever point we're discovering the `qsharp.json`

compiler/qsc_frontend/src/compile.rs
```diff
@@ -123,6 +143,28 @@ impl SourceMap {
     pub fn iter(&self) -> impl Iterator<Item = &Source> {
         self.sources.iter()
     }
+
+    /// Returns the sources as an iter, but with the project root directory subtracted
+    /// from the individual source names.
+    pub(crate) fn localized_project_sources(&self) -> impl Iterator<Item = Source> + '_ {
```
minestarks, line 149: The word `localized` heavily connotes "localization" to me, is `relative` possibly a better term?

compiler/qsc_parse/src/item.rs
```diff
@@ -131,14 +132,86 @@ fn parse_top_level_node(s: &mut ParserContext) -> Result<TopLevelNode> {
...
+
+/// Given a file name, convert it to a namespace name.
+/// For example, `foo/bar.qs` becomes `foo.bar`.
+fn file_name_to_namespace_name(raw: &str, error_span: Span) -> Result<Idents> {
+    let path = std::path::Path::new(raw);
+    let mut namespace = Vec::new();
+    for component in path.components() {
+        match component {
+            std::path::Component::Normal(name) => {
+                // strip the extension off], if there is one
```
ScottCarda-MS, line 158:
```suggestion
                // strip the extension off, if there is one
```

compiler/qsc_parse/src/item.rs
```diff
@@ -131,14 +132,86 @@ fn parse_top_level_node(s: &mut ParserContext) -> Result<TopLevelNode> {
...
+    // validation rules
+    let _ = ident(&mut s).map_err(|_| Error(ErrorKind::InvalidFileName(error_span)))?;
+    if s.peek().kind != TokenKind::Eof {
+        return Err(Error(ErrorKind::InvalidFileName(error_span)));
+    }
+    Ok(())
+}
+
+// unit tests for file_name_to_namespace_name
+#[test]
```
ScottCarda-MS, line 197: Should this be moved to the item/test.rs file?

compiler/qsc_parse/src/item.rs
```diff
@@ -149,6 +222,14 @@ fn parse_namespace(s: &mut ParserContext) -> Result<Namespace> {
     })
 }
 
+/// Parses the contents of a namespace block, what is in between the open and close braces in an
+/// explicit namespace, and any top level items in an implicit namespace.
+#[allow(clippy::vec_box)]
+fn parse_namespace_block_contents(s: &mut ParserContext) -> Result<Vec<Box<Item>>> {
```
ScottCarda-MS, line 228: Can you just use `parse_many`? I assume the code was able to do this logic before your changes, as namespace blocks aren't new.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -88,8 +90,26 @@ impl SourceMap {
...
+        // Calculate the common prefix.
+        let project_root_dir: String = longest_common_prefix(
+            &offset_sources
+                .iter()
+                .map(|source| source.name.as_ref())
+                .collect::<Vec<_>>(),
+        )
+        .to_string();
+
+        let project_root_dir: Arc<str> = Arc::from(project_root_dir);
```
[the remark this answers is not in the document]
ScottCarda-MS: Clarity's good. I'm fine if that's the reason for this.

language_service/src/state/tests.rs
```diff
@@ -116,7 +116,7 @@ async fn close_last_doc_in_project() {
         .update_document(
             "project/src/this_file.qs",
             1,
-            "/* this should not show up in the final state */ we should not see compile errors",
+            "namespace Foo.Bar { operation Main() : Unit { 5 }  }",
```
minestarks, line 119: Why did you have to change this test string? I think I'd feel better knowing that we're still getting a parse error here...

compiler/qsc_fir/src/fir.rs
minestarks, line 1: Is this a merge error? Seems unrelated to the change.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -59,6 +59,8 @@ pub struct AstPackage {
 #[derive(Debug, Default)]
 pub struct SourceMap {
     sources: Vec<Source>,
+    /// if the sources are from a project, then this is the project root directory.
+    project_root_dir: Option<Arc<str>>,
```
minestarks, line 63: So, based on all the things we've discussed, I suggest naming this something other than "project root" to reduce confusion. It's not always the "project root" in the sense that most people would understand it (i.e. wherever the qsharp.json is).

What about one of:

- `common_prefix`
- `common_path` (inspired by Python having a `commonpath()` function
- `common_root_dir`
- `common_root`


And I'd clarify in the comments that this value doesn't always have to be equal to the project root.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -530,3 +571,20 @@ fn assert_no_errors(sources: &SourceMap, errors: &mut Vec<Error>) {
         panic!("could not compile package");
     }
 }
+
+#[must_use]
+pub fn longest_common_prefix<'a>(strs: &'a [&'a str]) -> &'a str {
```
minestarks, line 576: How does this work if it's not taking path separators into account? Won't the common prefix of `/foo/bar` and `/foo/baz` be `/foo/ba`?

compiler/qsc_parse/src/item.rs
```diff
@@ -131,14 +132,74 @@ fn parse_top_level_node(s: &mut ParserContext) -> Result<TopLevelNode> {
...
+                let mut name = name
+                    .to_str()
+                    .ok_or(Error(ErrorKind::InvalidFileName(error_span)))?;
+
+                if let Some(dot) = name.rfind('.') {
+                    name = name[..dot].into();
+                }
+                // verify that the component only contains alphanumeric characters, and doesn't start with a number
+
+                validate_namespace_name(error_span, name)?;
```
minestarks, line 168: Since you're parsing here anyway, why not use the `Ident` created in this function instead of creating a new one?

compiler/qsc_parse/src/lib.rs
```diff
@@ -108,10 +115,38 @@ impl<T, F: FnMut(&mut ParserContext) -> Result<T>> Parser<T> for F {}
 #[must_use]
 pub fn namespaces(
     input: &str,
+    file_name: Option<&str>,
```
minestarks, line 118: I think you really need some test cases where you pass in a value here and validate the AST (for both valid and invalid file names)

compiler/qsc_parse/src/item/tests.rs
```diff
@@ -41,6 +42,18 @@ fn adjoint_invert() {
...
+// unit tests for file_name_to_namespace_name
+#[test]
+fn test_file_name_to_namespace_name() {
+    let raw = "foo/bar.qs";
+    let error_span = Span::default();
+    let namespace =
+        source_name_to_namespace_name(raw, error_span).expect("test should not fail here");
+    assert_eq!(namespace.0.len(), 2);
+    assert_eq!(&*namespace.0[0].name, "foo");
+    assert_eq!(&*namespace.0[1].name, "bar");
```
minestarks, line 54: Just cause this is already the pattern in this file: this test would read clearer if you used an `expect![]` to validate the object returned. 

Also nitpicking: drop the `test_` prefix from the function name since none of the other tests in this file use it.

minestarks (commented):
Made a first pass, will take another look after you've had a chance to address comments

DmitryVasilevsky:
A few questions to the PR description. 

"We detect the "root" of the project by calculating the largest common prefix of the file paths."
1. What if we are compiling just one file and have no explicit project?
2. What if we have a project, but ALL files happen to be in "src/they/are/here"?

manually evoking qsc on two files in different parts of your computer -- what is the "root"?
1. What if we invoke qsc on two files that are on different drives/shares?
2. If we use a file `qs/x/y/z/a.qs` in two qsc invocations - with `qs/x/y/z1/b1.qs` and `qs/x/y2/b2.qs`, the namespace assumed for a.qs will be different?

compiler/qsc_frontend/src/compile.rs
```diff
@@ -530,3 +572,46 @@ fn assert_no_errors(sources: &SourceMap, errors: &mut Vec<Error>) {
...
+                return &common_prefix_so_far[0..i];
+            }
+        }
+    }
+
+    common_prefix_so_far
+}
+
+#[must_use]
+pub fn longest_common_prefix<'a>(strs: &'a [&'a str]) -> &'a str {
```
minestarks, line 594: If there's only 1 name in the list, without a directory, I'd expect this to return `""`, not the whole string. Otherwise `relative_project_sources()` returns empty string which is pretty counterintuitive.

It's also more consistent with the behavior when there are 2 or more files.

e.g. 

* `["a", "b"]` => `""`
* `["a"]` => `""`
* `["a/b", "a/c"]` => `"a/"`
* `["a/b"]` => `"a/"`

minestarks: In test form, I'd expect all these to pass, but none of them do. The 4th one panics.

```rs
#[test]
fn test_longest_common_prefix_two_relative_paths() {
    expect!["a/"].assert_eq(longest_common_prefix(&["a/b", "a/c"]));
}

#[test]
fn test_longest_common_prefix_one_relative_path() {
    expect!["a/"].assert_eq(longest_common_prefix(&["a/b"]));
}

#[test]
fn test_longest_common_prefix_one_file_name() {
    expect![""].assert_eq(longest_common_prefix(&["a"]));
}

#[test]
fn test_longest_common_prefix_only_root_common() {
    expect![""].assert_eq(longest_common_prefix(&["/a/b", "/b/c"]));
}
```

pip/src/interpreter.rs
```diff
@@ -141,7 +142,12 @@ impl Interpreter {
             )
             .load_project(&manifest_descriptor.0)
             .map_py_err()?;
-            SourceMap::new(project.sources, None)
+            SourceMap::new(
+                project.sources,
+                Some(Arc::from(
+                    manifest_descriptor.0.manifest_dir.to_string_lossy(),
+                )),
```
minestarks, line 149: Isn't this supposed to be the entry expression?

minestarks (changes requested):

minestarks (approved):

PR #1417, opened 2024-04-23
Include return type of entry point in RCA check

swernli:
This change adds new runtime feature flags for use of specific types in output from the entry expression/entry point that RCA can annotate a program with. This allows the later check pass to verify that the code will use types in output recording that are supported for the given target. For example, a target that does not support dynamic integers cannot accomodate recoding output with integer values, even if they are static.


swernli:
We found a bug that is better fixed as part of other ongoing work that causes some test cases with this change to fail. Keeping this in draft until that bug is fixed and this can rebased on top of it.

cesarzc (approved):

idavis (approved):

PR #1421, opened 2024-04-24
Partial evaluation support for return expressions

cesarzc:
This change implements support for return expressions in partial evaluation.

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -401,11 +401,7 @@ impl Env {
         variables_by_scope.into_iter().flatten().collect::<Vec<_>>()
     }
 
-    #[must_use]
-    pub fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
-
+    #[allow(clippy::len_without_is_empty)]
     #[must_use]
     pub fn len(&self) -> usize {
         self.0.len()
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -341,7 +341,7 @@ impl Env {
             .find_map(|scope| scope.bindings.get_mut(id))
     }
 
-    fn push_scope(&mut self, frame_id: usize) {
+    pub fn push_scope(&mut self, frame_id: usize) {
         let scope = Scope {
             frame_id,
             ..Default::default()
@@ -349,7 +349,7 @@ impl Env {
         self.0.push(scope);
     }
 
-    fn leave_scope(&mut self) {
+    pub fn leave_scope(&mut self) {
         // Only pop the scope if there is more than one scope in the stack,
         // because the global/top-level scope cannot be exited.
         if self.0.len() > 1 {
@@ -399,6 +399,16 @@ impl Env {
             })
             .collect();
         variables_by_scope.into_iter().flatten().collect::<Vec<_>>()
+    }
+
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    #[must_use]
+    pub fn len(&self) -> usize {
+        self.0.len()
     }
 }
 
--- a/compiler/qsc_partial_eval/src/evaluation_context.rs
+++ b/compiler/qsc_partial_eval/src/evaluation_context.rs
@@ -52,6 +52,7 @@ impl EvaluationContext {
 
     /// Pops the currently active block.
     pub fn pop_block_node(&mut self) -> BlockNode {
+        self.get_current_scope_mut().active_block_count -= 1;
         self.active_blocks
             .pop()
             .expect("there are no active blocks in the evaluation context")
@@ -67,6 +68,7 @@ impl EvaluationContext {
     /// Pushes a new active block.
     pub fn push_block_node(&mut self, b: BlockNode) {
         self.active_blocks.push(b);
+        self.get_current_scope_mut().active_block_count += 1;
     }
 
     /// Pushes a new (call) scope.
@@ -93,6 +95,8 @@ pub struct Scope {
     pub args_value_kind: Vec<ValueKind>,
     /// The classical environment of the callable, which holds values corresponding to local variables.
     pub env: Env,
+    /// Number of currently active blocks (starting from where this scope was created).
+    active_block_count: usize,
     /// Map that holds the values of local variables.
     hybrid_vars: FxHashMap<LocalVarId, Value>,
 }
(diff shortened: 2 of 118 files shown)


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -55,53 +56,17 @@ pub fn partially_evaluate(
...
-
-    #[error("failed to evaluate tuple element expression")]
-    #[diagnostic(code("Qsc.PartialEval.FailedToEvaluateTupleElementExpression"))]
-    FailedToEvaluateTupleElementExpression(#[label] Span),
+    EvaluationFailed(String, #[label] Span),
 
-    #[error("failed to evaluate binary expression operand")]
-    #[diagnostic(code("Qsc.PartialEval.FailedToEvaluateBinaryExpressionOperand"))]
-    FailedToEvaluateBinaryExpressionOperand(#[label] Span),
+    #[error("partial evaluation does not support early returns")]
```
cesarzc, line 63: Make these one `Unexpected` error that uses a `String` and a `Span`.

compiler/qsc_eval/src/lib.rs
```diff
@@ -400,6 +400,16 @@ impl Env {
             .collect();
         variables_by_scope.into_iter().flatten().collect::<Vec<_>>()
     }
+
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
```
swernli, line 405: I think `is_empty` can be removed. It's not called anywhere, and because of how `leave_scope` never pops the last scope created by the `default` implementation, `is_empty` actually always returns false, so it will likely never be useful.

cesarzc: I added it because of a clippy error but it makes more sense to not have it so I removed it.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -121,45 +123,70 @@ impl Scope {
...
+}
 
-    pub fn clear_last_expr(&mut self) {
-        self.last_expr = None;
+pub struct ValueControlFlow {
+    pub value: Value,
+    pub kind: ControlFlowKind,
+}
+
+impl ValueControlFlow {
```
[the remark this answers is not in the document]
cesarzc: Added a bunch of doc strings :)

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -121,45 +123,70 @@ impl Scope {
...
     pub fn get_local_var_value(&self, local_var_id: LocalVarId) -> &Value {
         self.hybrid_vars
             .get(&local_var_id)
             .expect("local variable value does not exist")
     }
 
-    pub fn insert_expr_value(&mut self, expr_id: ExprId, value: Value) {
-        self.last_expr = Some(expr_id);
-        self.hybrid_exprs.insert(expr_id, value);
+    pub fn has_returned(&self) -> bool {
```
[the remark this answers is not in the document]
cesarzc: Docstring added!

compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -121,45 +146,86 @@ impl Scope {
...
+    /// The block ID corresponding to a branch.
+    Block(BlockId),
+    /// The return value resulting from a branch.
+    Return(Value),
+}
 
-    pub fn clear_last_expr(&mut self) {
-        self.last_expr = None;
+/// An association between a value and a control flow option.
+pub struct ValueControlFlow {
```
cesarzc, line 198: Just use an enum.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1029,23 +1102,52 @@ impl<'a> PartialEvaluator<'a> {
...
+        let mut stmts_iter = block.stmts.iter();
+        for stmt_id in stmts_iter.by_ref() {
+            last_control_flow = self.try_eval_stmt(*stmt_id)?;
+            if matches!(last_control_flow.kind, ControlFlowKind::Return) {
+                return_stmt_id = Some(*stmt_id);
+                break;
+            }
+        }
+
+        // While we support multiple returns within a callable, disallow situations in which non qubit release
```
cesarzc, line 1120: Need to change this to still process qubit releases.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1029,23 +1102,52 @@ impl<'a> PartialEvaluator<'a> {
...
+        // statements are left unprocessed.
+        let non_qubit_relase_stmts_remain =
+            stmts_iter.any(|stmt_id| !self.is_qubit_release_stmt(*stmt_id));
+        if non_qubit_relase_stmts_remain {
+            let return_stmt =
+                self.get_stmt(return_stmt_id.expect("a return statement ID must have been set"));
+            Err(Error::Unexpected(
+                "early return".to_string(),
+                return_stmt.span,
+            ))
```
swernli, line 1130: Something seems odd to me here... shouldn't it only be early returns in a dynamic context that get treated as error? If the return is due to purely classical control flow, shouldn't it be allowed? For example:
```qsharp
operation XorMore(q : Qubit, cond: Bool) : Unit {
    if cond {
        X(q);
        return;
    }
    H(q);
    if M(q) == One {
        X(q);
    }
}
```
If `XorMore` is called with a classical true, it will perform just `X` and not include any control flow in the RIR, and with a classical false it will have the `H` and one branch in the RIR based on `M`. Both of those could be handled with the infra you have here, but it looks like that would be rejected by this logic.

swernli: It's possible I'm just missing the intent here, so happy to sync up more on it Monday.

compiler/qsc_partial_eval/src/evaluation_context.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: This is a good point. This implementation was overly restrictive.

I've changed it such that we only raise an error when we are performing evaluation within a branch (at the callable scope) level. I did this with simple reference counting for now, but it gives us the points where we should handle early returns in a later change.

Happy to sync up on the details :)

swernli (approved):

PR #1427, opened 2024-04-24
Playground Tabs

ScottCarda-MS:
Small thing, but wanted to change the order of the playground tabs so that "AST" is before "HIR" when reading left-to-right.

Also drops the unused "Logs" tab.


minestarks (approved):
Thanks for the attention to detail!

PR #1430, opened 2024-04-25
Remove stray $ in kata

billti:
Currently Polar Coordinates kata has a note which renders below (note the second paragraph LaTeX). This fixes it.

image


tcNickolas (approved):
Good catch, thank you!

PR #1437, opened 2024-04-26
Avoid RCA panic when last stmt is binding

swernli:
RCA had an assumption that a block with a non-Unit type must end with an expression, but as it turns out it's valid to have the outermost block of a callable with non-unit value end in an unreachable binding (a binding that appears after the return expresion, for example). If that code was submitted, the panic in RCA would cause a the language service to stop responding. This fixes the issue by treating the check for a terminating expression as best effort, since other logic already handles properly annotating callables that have return expressions.

### DIFF
--- a/language_service/src/compilation.rs
+++ b/language_service/src/compilation.rs
@@ -72,6 +72,21 @@ impl Compilation {
             language_features,
         );
 
+        // Compute new lints and append them to the errors Vec.
+        // Lints are only computed if the erros vector is empty. For performance
+        // reasons we don't want to waste time running lints every few keystrokes,
+        // if the user is in the middle of typing a statement, for example.
+        if errors.is_empty() {
+            let lints = qsc::linter::run_lints(&unit, Some(lints_config));
+            let lints: Vec<_> = lints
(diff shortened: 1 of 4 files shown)


language_service/src/compilation.rs
```diff
@@ -100,12 +85,19 @@ impl Compilation {
             unit,
         );
 
-        let lints = qsc::linter::run_lints(unit, Some(lints_config));
-        for lint in lints {
-            errors.push(WithSource::from_map(
-                &unit.sources,
-                qsc::compile::ErrorKind::Lint(lint),
-            ));
+        // Compute new lints and append them to the errors Vec.
```
swernli, line 88: The changes in this file were to ensure RIR passes happen before lints. Since lints can add warnings and errors, and RIR will only run when errors list is empty, a lint warning would prevent RIR checks, which is not desireable.

swernli: Looks like this is already getting fixed in #1439, so I can remove this part of the changes.

cesarzc (approved):

PR #1439, opened 2024-04-26
Fix duplicated lints

orpuente-MS:
Lints were being generated twice for `qsharp` files. This PR fixes that issue. It also changes the order of the passes, now the linter-passes happen after the fir-passes.

### DIFF
--- a/language_service/src/compilation.rs
+++ b/language_service/src/compilation.rs
@@ -289,10 +289,9 @@ fn run_linter_passes(
 ) {
     if errors.is_empty() {
         let lints = qsc::linter::run_lints(unit, Some(config));
-        let lints: Vec<_> = lints
+        let lints = lints
             .into_iter()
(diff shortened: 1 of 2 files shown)


swernli (approved):

language_service/src/compilation.rs
```diff
@@ -312,6 +278,25 @@ fn run_fir_passes(
     }
 }
 
+/// Compute new lints and append them to the errors Vec.
+/// Lints are only computed if the erros vector is empty. For performance
+/// reasons we don't want to waste time running lints every few keystrokes,
+/// if the user is in the middle of typing a statement, for example.
+fn run_linter_passes(
+    config: Option<&[LintConfig]>,
```
idavis, line 286: @orpuente-MS  We are only ever passing `Some(..)` into this call. Can we drop the option type here?

orpuente-MS: Good catch, already fixed. Thanks!

idavis (approved):

PR #1442, opened 2024-04-29
Track nested scopes during debugging

### LINKED ISSUE #1436: Evaluator produces confusing locals view while debugging
The evaluator refactoring in #1261 had an unexpected side effect of leaving variables in scope too long, which can cause confusing debugger output:
image
The `y: 1` entry should only be visible while inside the block for the if-expression. The control flow graph will need to be updated to include scope entry and exit markers when running in debug mode.

swernli:
This adds new execution graph nodes for pushing and popping variable scopes that are used only while debugging (like statement nodes) so that the evaluator can correctly respond to local variable requests during debugging.

Fixes #1436

### DIFF
--- a/compiler/qsc_lowerer/src/lib.rs
+++ b/compiler/qsc_lowerer/src/lib.rs
@@ -274,6 +274,9 @@ impl Lowerer {
 
     fn lower_block(&mut self, block: &hir::Block) -> BlockId {
         let id = self.assigner.next_block();
+        if self.enable_debug {
+            self.exec_graph.push(ExecGraphNode::PushScope);
+        }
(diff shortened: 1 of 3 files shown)


compiler/qsc_lowerer/src/lib.rs
```diff
@@ -274,6 +274,9 @@ impl Lowerer {
 
     fn lower_block(&mut self, block: &hir::Block) -> BlockId {
         let id = self.assigner.next_block();
```
idavis, line 276: Please document this function for what is going on in detail.

idavis (approved):

PR #1444, opened 2024-04-29
Fix lint message formatting

### LINKED ISSUE #1443: Bad formatting in lint reports
The message when hovering over a lint squiggle in VS Code is ill formatted.

image

orpuente-MS:
When hovering over the squiggle generated by a Lint, the message in the first line appears twice. This PR removes the duplicated message. Fixes #1443 

Currently, the error message is shown in the first line and in the last line, which is redundant:
image

This is how it looks after the change:
image

### DIFF
--- a/compiler/qsc_linter/src/linter.rs
+++ b/compiler/qsc_linter/src/linter.rs
@@ -43,7 +43,7 @@ pub struct Lint {
 
 impl std::fmt::Display for Lint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.message)
+        write!(f, "{}", self.level)
     }
 }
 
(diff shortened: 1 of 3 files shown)


swernli:
FYI: the "fixes #1443" should be in the description rather than the title

orpuente-MS:
> FYI: the "fixes #1443" should be in the description rather than the title

Thanks!

compiler/qsc_linter/src/linter.rs
```diff
@@ -43,7 +43,7 @@ pub struct Lint {
 
 impl std::fmt::Display for Lint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.message)
+        write!(f, "{}", self.level)
```
billti, line 46: So the `Display` string rendering for a Lint is just its level? That doesn't sound right.

orpuente-MS: Yes, you are right. I changed it to something that makes more sense. Thanks!

swernli (approved):

PR #1453, opened 2024-04-30
introduce import and export statements

sezna:
This PR introduces exports and imports. A namespace can export any symbol it has access to. External packages (consumers of this package) will only have access to explicitly exported items.

## How it works

Exports are a type of `Item`, and they appear inside of namespaces (implicit or explicit). Exports are paths, which are resolved to local item ids in the resolver.  They are then [bound to the scope](https://github.com/microsoft/qsharp/pull/1453/files#diff-58ecc2f18ce2a352b89fdb047252d971d27a29bdacb49f377dbba36f744b8d1bR638) of that namespace. After that, the exported item is [added to the api surface](https://github.com/microsoft/qsharp/pull/1453/files#diff-58ecc2f18ce2a352b89fdb047252d971d27a29bdacb49f377dbba36f744b8d1bR644) of the namespace.

When we support external package dependencies, we only want to allow importing of explicitly exported items. At that time, we will add a flag to the items to denote if they originate from an export.

This PR also supports export aliases: `export Foo as Bar;`

Glob imports work similarly to other languages: `Foo.*` brings in everything from `Foo`, while `Foo.* as Bar` brings in all of `Foo`, but aliased under a namespace `Bar`. For example, if `function F() : Unit {}` exists in `Foo`, then after importing via `Foo.* as Bar`, it would be available as `Bar.F()`.

### DIFF
--- a/compiler/qsc_frontend/src/resolve/tests.rs
+++ b/compiler/qsc_frontend/src/resolve/tests.rs
@@ -3690,67 +3690,6 @@ namespace Main {
             }"#]],
     );
 }
-#[test]
-fn multiple_exports() {
-    check(
-        indoc! {"
-            namespace Foo {
-                operation ApplyX() : Unit {}
-                operation ApplyY() : Unit {}
-                export ApplyX, ApplyY;
-            }
-            namespace Main {
-                open Foo;
-                operation Main() : Unit {
-                    ApplyX();
-                    ApplyY();
-                }
-            }
-        "},
-        &expect![[r#"
-            namespace namespace7 {
-                operation item1() : Unit {}
-                operation item2() : Unit {}
-                export item1, item2;
-            }
-            namespace namespace8 {
-                open namespace7;
-                operation item4() : Unit {
-                    item1();
-                    item2();
-                }
-            }
-        "#]],
-    );
-}
-
-#[test]
-fn no_exports() {
-    check(
-        indoc! {"
-            namespace Foo {
-                operation ApplyX() : Unit {}
-            }
-            namespace Main {
-                open Foo;
-                operation Main() : Unit {
-                    ApplyX();
-                }
-            }
-        "},
-        &expect![[r#"
-            namespace namespace7 {
-                operation item1() : Unit {}
-            }
-            namespace namespace8 {
-                open namespace7;
-                operation item3() : Unit {
-                    item1();
-                }
-            }
-        "#]],
-    );
-}
 
 #[test]
 fn export_non_existent_symbol() {
@@ -3850,36 +3789,6 @@ fn export_non_item() {
             }
 
             // ExportedNonItem(Span { lo: 80, hi: 84 })
-        "#]],
-    );
-}
-
-#[test]
-fn export_udt() {
-    check(
-        indoc! {"
-            namespace Foo {
-                newtype Pair = (First: Int, Second: Int);
-                export Pair;
-            }
-            namespace Main {
-                open Foo;
-                operation Main() : Unit {
-                    Pair(1, 2);
-                }
-            }
-        "},
-        &expect![[r#"
-            namespace namespace7 {
-                newtype item1 = (First: Int, Second: Int);
-                export item1;
-            }
-            namespace namespace8 {
-                open namespace7;
-                operation item3() : Unit {
-                    item1(1, 2);
-                }
-            }
         "#]],
     );
 }
@@ -4890,7 +4799,7 @@ fn disallow_glob_export() {
                 operation ApplyY() : Unit {}
             }
             namespace Bar {
-                export Foo.*;resou
+                export Foo.*;
             }
         "},
         &expect![[r#"


compiler/qsc_frontend/src/resolve.rs
```diff
@@ -569,6 +582,68 @@ impl Resolver {
...
+                        Res::Item(
+                            ItemId {
+                                package: Some(_), ..
+                            },
+                            _,
+                        ) => {
+                            self.errors.push(Error::ExportedExternalItem(item.span));
+                            continue;
+                        }
+                        Res::Item(id, _) => id,
```
minestarks, line 627: Feels like you should use the `ItemStatus` in some way to determine whether the item is available in the current configuration.

e.g. the error message in the below export could be a little more helpful (just like it is in the function call)

```qsharp
namespace Foo {
    export { BaseOp }; // "this is not a callable or type", no help
             ^^^^^^

    @EntryPoint()
    operation Main() : Unit {
        BaseOp(); // help: found a matching item `Foo.BaseOp` that is not available for the current compilation configuration
        ^^^^^^
    }

    @Config(Base)
    operation BaseOp() : Unit {

    }
}
```

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -569,6 +582,68 @@ impl Resolver {
...
+                            ) {
+                                Ok(res) => res,
+                                Err(err) => {
+                                    self.errors.push(err);
+                                    continue;
+                                }
+                            }
+                        }
+                    };
+
```
minestarks, line 614: Shouldn't you just call `resolve_path` here (first with `NameKind::Ty`, then with `NameKind::Term`)? It does all the work for you (and some good error checking)

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -569,6 +582,68 @@ impl Resolver {
...
+                        Res::Item(id, _) => id,
+                        _ => {
+                            self.errors.push(Error::ExportedNonItem(item.span));
+                            continue;
+                        }
+                    };
+
+                    scope
+                        .terms
+                        .insert(Rc::clone(&item.name.name), resolved_item_id);
```
minestarks, line 636: Why is this needed? The terms table should have been updated when the item was bound in the first place.

compiler/qsc_frontend/src/resolve.rs
minestarks, line 1: Are multiple exports allowed for the same item? `export { Foo, Foo, Foo }` doesn't seem to produce an error

compiler/qsc_frontend/src/resolve.rs
minestarks: Looks like this makes it so that if you're exporting an item from a _different_ namespace, the name is made available in the current namespace.

Not sure if this is intentional behavior, but it at least needs to be tweaked a little, e.g. to avoid the ambiguous resolution issue from my other comment.

There should probably also be a unit test for it, so that _something_ fails when I comment out this line

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -569,6 +582,68 @@ impl Resolver {
                 scope.tys.insert(Rc::clone(&name.name), id);
                 scope.terms.insert(Rc::clone(&name.name), id);
             }
+            ast::ItemKind::Export(export) => {
```
minestarks, line 585: This kind of lumps the bind and resolve steps together which isn't a great idea IMO. Ideally, we bind all names first in one pass, then we resolve all references in another pass, which eliminates ordering issues. 

In trying to find a good example to articulate why I think this is a problem, I came across this one:

```qsharp
namespace Foo {
    export { Bar.Baz };
    export { Baz };
}

namespace Bar {
    operation Baz(): Unit {
    }
}
```

The first `export` statement imports `Baz` into the namespace, and the second export statement references it. Should this compile?

And if yes, should the following compile too? Declaration order doesn't matter for any of the other items, and I don't think it should matter for exports either.

```qsharp
namespace Foo {
    export { Baz };
             ^^^ name error
    export { Bar.Baz };
}

namespace Bar {
    operation Baz(): Unit {
    }
}
```

How about this?

```qsharp
namespace Foo {
    operation Main() : Unit {
        Baz(); // can reference Baz no problem
    }

    export { Baz };
             ^^^ name error

    open Bar;
    
}

namespace Bar {
    operation Baz(): Unit {
    }
}
```

minestarks (commented):
Will continue after you've had a chance to review this batch of comments.

compiler/qsc_parse/src/item.rs
minestarks, line 1: Doesn't have to be this PR, but just so I don't forget, we should make a note to update the syntax highlighting grammars (codemirror and TextMate) with the new keywords.

compiler/qsc_frontend/src/compile.rs
```diff
@@ -514,8 +514,14 @@ fn resolve_all(
         dropped_names.extend(unit.dropped_names.iter().cloned());
     }
 
+    // bind all symbols in `add_local_package`
     let mut errors = globals.add_local_package(assigner, package);
     let mut resolver = Resolver::new(globals, dropped_names);
+
+    // bind all exported symbols in a follow-on step
+    resolver.resolve_exports(package);
```
minestarks, line 522: I think the behavior is ok here, but the method name and comment confused me.

Would "resolve and bind all imports and exports" be a more complete description of what's happening here?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -108,6 +113,18 @@ pub(super) enum Error {
     #[diagnostic(help("this item is not implemented and cannot be used"))]
     #[diagnostic(code("Qsc.Resolve.Unimplemented"))]
     Unimplemented(String, #[label] Span),
+
+    #[error("this export is not a callable or type")]
+    #[diagnostic(code("Qsc.Resolve.ExportedNonItem"))]
+    ExportedNonItem(#[label] Span),
+
+    #[error("exporting external items is not yet supported")]
```
minestarks, line 121: I'm going to continue my crusade against the word "yet" in errors... 

```suggestion
    #[error("exporting external items is not supported")]
```

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -108,6 +113,18 @@ pub(super) enum Error {
...
+    #[diagnostic(code("Qsc.Resolve.ExportedExternalItem"))]
+    ExportedExternalItem(#[label] Span),
+
+    #[error("export statements are only allowed in a namespace scope")]
+    #[diagnostic(code("Qsc.Resolve.ExportFromNonNamespaceScope"))]
+    ExportFromNonNamespaceScope(Span),
```
minestarks, line 127: We may want to tweak the error message here, for the user's sake. The word "namespace" won't make sense to them as we don't want them using the `namespace` keyword anymore, and they won't know what "implicit namespaces" are.

Maybe just state what we _don't_ want to happen, like:

```suggestion
    #[error("export statements are not allowed in a local scope")]
    #[diagnostic(code("Qsc.Resolve.ExportFromLocalScope"))]
    ExportFromLocalScope(Span),
```

or

```suggestion
    #[error("export statements are not allowed in a callable body")]
    #[diagnostic(code("Qsc.Resolve.ExportFromCallableBody"))]
    ExportFromCallableBody(Span),
```

or maybe something better if you can think of it.

compiler/qsc_ast/src/ast.rs
```diff
@@ -262,6 +272,8 @@ pub enum ItemKind {
     Open(Idents, Option<Box<Ident>>),
     /// A `newtype` declaration.
     Ty(Box<Ident>, Box<TyDef>),
+    /// An export declaration
+    Export(ExportDecl),
```
minestarks, line 276: How do we handle item properties on export and import items?

```qsharp
/// doc comment!
@Config(Base)
internal import Bar.BarOperation;
```

I guess the current answer is that they don't cause any errors and have no effect, but is this desirable?

compiler/qsc_frontend/src/resolve.rs
minestarks, line 1: The fact that we can export namespaces opens up some ambiguities:

```qsharp
namespace Foo.Bar {
    operation Baz() : Unit {
        Message("i'm in Foo.Bar.Baz")
    }
}

namespace Foo.Bar2 {
    operation Baz() : Unit {
        Message("i'm in Foo.Bar2.Baz")
    }
}

namespace Foo {
    operation Bar() : Unit {
        Message("i'm in Foo.Bar")
    }
}

namespace Main {
    export { Foo.Bar as Qux };
    export { Foo.Bar2 as Qux };

    @EntryPoint()
    operation Main() : Unit {
        Qux();
        Qux.Baz();
        // What is Qux, exactly?
    }
}
```

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -566,10 +712,236 @@ impl Resolver {
...
+                }
+            }
+
+            self.names.insert(item.path.id, resolved_item);
+        }
+    }
+
+    /// This function performs the logic to see if a path is a ty, term, or namespace, and then
+    /// performs the appropriate logic to bind the path.
+    fn bind_exported_path_of_unknown_name_kind(
```
minestarks, line 829: Isn't this method doing basically the same thing as `bind_import`? Can we make them into one method?

In general, I can't understand why export and import implementations are so divergent in this PR, when an `export` is basically the same thing as an `import` (with a few minor differences like `ItemSource` and lack of support for globs). The resolution and binding to be almost exactly the same code in both, no? Is there any conceivable case where an `import`ed item should successfully resolve, but a re-`export` of the same shouldn't? Or vice versa?

compiler/qsc_parse/src/item.rs
```diff
@@ -518,3 +525,202 @@ pub(super) fn check_input_parens(inputs: &Pat) -> Result<()> {
...
+/// ```qsharp
+/// import Foo.Bar;
+/// import Foo.{Baz, Quux};
+/// import Foo.{Bar.{Baz, Quux}, Corge as Grault};
+/// ```
+fn parse_import(s: &mut ParserContext) -> Result<ImportDecl> {
+    let lo = s.peek().span.lo;
+    let _doc = parse_doc(s);
+    token(s, TokenKind::Keyword(Keyword::Import))?;
+    let (base_path, _) = seq(s, ident, TokenKind::Dot)?;
```
minestarks, line 577: Is a `seq` parser with a `.` separator meaningfully different than a `Path` parser? Why go through the trouble of making `seq` generic instead of using `Path`?

`seq` allows an optional trailing separator. Are all these syntaxes allowed on purpose?

```qsharp
import Foo.{Bar, Bar2};
import Foo{Bar, Bar2};
import Foo.;
```

compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -72,9 +72,7 @@ fn one_callable() {
                             ),
                         ),
                         opens: {
-                            [
-                                "Foo",
-                            ]: [
+                            []: [
```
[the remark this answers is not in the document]
minestarks: I think so!

minestarks (approved):

compiler/qsc_ast/src/ast.rs
```diff
@@ -1684,3 +1734,117 @@ pub enum SetOp {
...
+            span,
+            items,
+            is_export,
+        }
+    }
+
+    /// Returns true if this is an export declaration.
+    #[must_use]
+    pub fn is_export(&self) -> bool {
+        self.is_export
```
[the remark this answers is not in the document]
minestarks: I think `ImportOrExportItem` should stay a single type but `ImportOrExportDecl` could be separate enums (though I don't mind the way it is right now)

swernli (approved):

PR #1454, opened 2024-04-30
Update RCA handling of array dynamism

swernli:
This change updates the assumptions in RCA to be less restrictive on assuming the dynamic size of arrays. Before, generating the `ValueKind` for arrays would always assume both dynamic content and dynamic size, which could cause failures in checks when, for example, a tuple of static sized arrays is returned from the entry point. Instead, we rely on the propagation of the runtime feature flags to track whether a dynamically sized array is used. This is a precursor to later simplifying the `QuantumProperties` struct by removing `ValueKind` in favor of just tracking dynamic vs static.

### DIFF
--- a/compiler/qsc_rca/src/lib.rs
+++ b/compiler/qsc_rca/src/lib.rs
@@ -599,8 +599,10 @@ impl ValueKind {
             Self::Element(RuntimeKind::Static)
         } else {
             match ty {
-                // For a dynamic array, both contents and size are dynamic.
-                Ty::Array(_) => ValueKind::Array(RuntimeKind::Dynamic, RuntimeKind::Dynamic),
+                // For a dynamic array, the contents dynamic and size is static.
+                // We assume this because the source of the array produces something with dynamic length,
+                // that source should have already added the runtime feature flag for dynamic arrays.
+                Ty::Array(_) => ValueKind::Array(RuntimeKind::Dynamic, RuntimeKind::Static),
                 // For every other dynamic type, we use the element variant with a dynamic runtime value.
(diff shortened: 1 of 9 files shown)


compiler/qsc_rca/src/lib.rs
```diff
@@ -599,8 +599,10 @@ impl ValueKind {
             Self::Element(RuntimeKind::Static)
         } else {
             match ty {
-                // For a dynamic array, both contents and size are dynamic.
-                Ty::Array(_) => ValueKind::Array(RuntimeKind::Dynamic, RuntimeKind::Dynamic),
+                // For a dynamic array, the contents dynamic and size is static.
```
cesarzc, line 602:
```suggestion
                // For a dynamic array, the content is dynamic and the size is static.
```

compiler/qsc_rca/src/lib.rs
```diff
@@ -659,7 +661,7 @@ impl ValueKind {
                 }
                 ValueKind::Element(self_runtime_kind) => {
                     *content_runtime_kind = self_runtime_kind;
-                    *size_runtime_kind = self_runtime_kind;
+                    *size_runtime_kind = RuntimeKind::Static;
```
cesarzc, line 664:
```suggestion
                    // When we project from an element variant to an array variant, we assume the size of the
                    // array is static because we rely on the dynamically-sized arrays runtime feature flag to
                    // detect such cases.
                    *size_runtime_kind = RuntimeKind::Static;
```

cesarzc (approved):

PR #1402, opened 2024-04-18
Update READMEs to add details on building playground and katas

Manvi-Agrawal:
Add instructions about building katas locally by building playground

PR #1455, opened 2024-04-30
[Node Watch] Fix node watch error windows

Manvi-Agrawal:
### Issue
```
Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at spawn (node:child_process:761:9)
    at runWatcher (file:///C:/Users/manvi/Documents/qsharp/watch.mjs:108:22)
    at file:///C:/Users/manvi/Documents/qsharp/watch.mjs:121:1
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12) {
  errno: -4071,
  code: 'EINVAL',
  syscall: 'spawn'
}
```
### Fix
As per https://github.com/nodejs/node/issues/52554#issuecomment-2060026269, we should set `{shell: true}` because
spawning .cmd files is blocked.

`const npmCmd = isWin ? "npm.cmd" : "npm";`

### Verification
- Watch is working on Windows after my change.
- Tested on codespaces(Linux environment), and observed no side effect.

### Possible Side Effects
Need to investigate if `shell:True` can cause issues on Linux.

### DIFF
--- a/watch.mjs
+++ b/watch.mjs
@@ -55,7 +55,7 @@ function buildRust() {
   const result = spawnSync(
     "wasm-pack",
     ["build", "--dev", "--no-pack", "--target", "web", "--out-dir", buildDir],
-    { cwd: wasmDir },
+    { cwd: wasmDir, shell: true },
   );
   console.log("wasm-pack done! ", result.stderr.toString());
 
@@ -105,7 +105,7 @@ onRustChange();
  */
 function runWatcher(dir, name, watchTask = "tsc:watch") {
   console.log(`Spawning tsc:watch for ${name} in ${dir}`);
-  const npmWatcher = spawn(npmCmd, ["run", watchTask], { cwd: dir });
+  const npmWatcher = spawn(npmCmd, ["run", watchTask], { cwd: dir, shell: true });
   npmWatcher.stdout.on("data", (data) =>
     console.log(`tsc:watch ${name}: ${data}`),
   );
--- a/watch.mjs
+++ b/watch.mjs
@@ -105,7 +105,10 @@ onRustChange();
  */
 function runWatcher(dir, name, watchTask = "tsc:watch") {
   console.log(`Spawning tsc:watch for ${name} in ${dir}`);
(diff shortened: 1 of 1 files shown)


Manvi-Agrawal:
@tcNickolas , @minestarks , I was trying to use watch command on my windows machine(https://github.com/microsoft/qsharp/pull/1402#discussion_r1571199619) and realised that I couldnt use it. So, sending a small fix :-)

billti:
Wow. I tested this when I checked it in and was working fine, even did some special handling for Windows already. Reading through the linked Node.js issue, crazy that they broke this in a minor patch release 2 weeks ago and then just locked the conversation :-/

We'll need to test on macOS and Linux that the change works fine. Did you try it on any other OS?

watch.mjs
```diff
@@ -55,7 +55,7 @@ function buildRust() {
   const result = spawnSync(
     "wasm-pack",
     ["build", "--dev", "--no-pack", "--target", "web", "--out-dir", buildDir],
-    { cwd: wasmDir },
+    { cwd: wasmDir, shell: true },
```
Manvi-Agrawal, line 58: Need to investigate if `shell: true` can cause issues in Linux

Manvi-Agrawal:
> Wow. I tested this when I checked it in and was working fine, even did some special handling for Windows already. Reading through the linked Node.js issue, crazy that they broke this in a minor patch release 2 weeks ago and then just locked the conversation :-/

> We'll need to test on macOS and Linux that the change works fine. Did you try it on any other OS?

@billti I have only Windows :-). Maybe I can use codespaces/WSL for linux testing. I realised after raising PR, and updated description main comment and added a new comment to realise that you have this doubt :)

minestarks (approved):
LGTM but I haven't validated on macOS

Issue #1388, opened 2024-04-15
Review algorithm samples and see if anything suspicious with global phase shows up.


swernli:
See #1317 for more context.

Some of our samples might use patterns that interact poorly with global phase in a way that can confuse readers of the state dumps. We should review the samples and see if any changes like the qubit scoping for #1370 can mitigate that confusion. In general, samples should avoid nuanced global phase if possible.

This review may result in sub-issues for relevant sample updates or just be closed once samples have been sufficiently reviewed.

tcNickolas:
I skimmed through our algorithms samples, the next most confusing one is Teleportation: it says teleporting one-qubit state but DumpMachine has two qubits. We should be able to use DumpRegister instead to have one-qubit states.

swernli:
Given we have not observed any new phase issues with samples, I'm going to close this issue. We can either reopen or file fresh if a new phase related issue is seen.

PR #1456, opened 2024-05-01
switched to dumpregister

goshua13:
Switched dumpMachine to dumpRegister to have one-qubit states #1388 @cesarzc

### DIFF
--- a/samples/algorithms/Teleportation.qs
+++ b/samples/algorithms/Teleportation.qs
@@ -16,9 +16,6 @@ namespace Sample {
 
     @EntryPoint()
     operation Main() : Result[] {
-        // Allocate the message and target qubits.
-        use (message, target) = (Qubit(), Qubit());
-
         // Use the `Teleport` operation to send different quantum states.
         let stateInitializerBasisTuples = [
             ("|0〉", I, PauliZ),
@@ -29,6 +26,9 @@ namespace Sample {
 
         mutable results = [];
         for (state, initializer, basis) in stateInitializerBasisTuples {
+            // Allocate the message and target qubits.
+            use (message, target) = (Qubit(), Qubit());
+
             // Initialize the message and show its state using the `DumpMachine`
             // function.
             initializer(message);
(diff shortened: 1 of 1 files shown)


tcNickolas (approved):
Looks good to me - definitely less confusing than printing two qubits when talking about the state of just one!

Thank you!

samples/algorithms/Teleportation.qs
```diff
@@ -33,13 +33,13 @@ namespace Sample {
             // function.
             initializer(message);
             Message($"Teleporting state {state}");
-            DumpMachine();
```
cesarzc, line 36: Thanks for the changes @goshua13!

One thing that we @swernli and I noticed is that `DumpRegister` output for teleportation of state `|-〉` because it shows a global phase.

In order to reduce global phase confusion, could you please also make the following changes to this sample:
- Move the allocation of the `message` and `target` qubits inside the loop. This will deallocate them in every iteration and will remove any global phase accumulated from previous teleportations because all qubits are released for every iteration.
- Inside the `Teleport` operation, switch the ordering of the corrections. The current order introduces a global phase that makes things a bit more confusing.

tcNickolas: Does that help? I tried that but still got -1 phase on |-> half of the times

cesarzc: It helps for the global state. With these changes the output of `DumpMachine` is consistent. However, as part of looking into this, we found a potential bug in `DumpRegister` and that's why sometimes we get a -1 phase and sometimes not. We specifically get the -1 phase when we apply a `Z` correction.

swernli: Yup, like Cesar said these changes make the internal state more consistent, which only highlights how odd it is that `DumpRegister` specifically adds an unexpected phase.

We tried this out with just the minus state and printing what corrections are applied. With the updates Cesar suggested, we can see that when no corrections are applied the before and after of both `DumpMachine` (shown first) and `DumpRegister` (shown second) agree:
image

But when a Z correction is applied, the message qubit is left in the one state, and somehow that causes `DumpRegister` to report the target qubit as having a phase of -1:
image

Given that the underlying amplitudes are the same, the "other" qubit being in the one state should not cause `DumpRegister` on the target qubit to include a phase.

cesarzc (approved):

PR #1457, opened 2024-05-01
Adds Superposition Task 2.6

frtibble:
Part of #1185

### DIFF
--- a/katas/content/index.json
+++ b/katas/content/index.json
@@ -12,5 +12,6 @@
   "oracles",
   "deutsch_algo",
   "deutsch_jozsa",
-  "qec_shor"
+  "qec_shor",
+  "superposition"
 ]
--- a/katas/content/index.json
+++ b/katas/content/index.json
@@ -12,6 +12,5 @@
   "oracles",
   "deutsch_algo",
   "deutsch_jozsa",
-  "qec_shor",
-  "superposition"
+  "qec_shor"
 ]
--- a/katas/content/superposition/wstate_power_of_two/Verification.qs
+++ b/katas/content/superposition/wstate_power_of_two/Verification.qs
@@ -21,7 +21,7 @@ namespace Kata.Verification {
         }
     }
 
-        function FindFirstDiff (bits1 : Bool[], bits2 : Bool[]) : Int {
+    function FindFirstDiff (bits1 : Bool[], bits2 : Bool[]) : Int {
         for i in 0 .. Length(bits1) - 1 {
             if bits1[i] != bits2[i] {
                 return i;
@@ -31,21 +31,16 @@ namespace Kata.Verification {
     }
 
     operation TwoBitstringSuperposition_Reference (qs : Qubit[], bits1 : Bool[], bits2 : Bool[]) : Unit is Adj + Ctl {
-        // find the index of the first bit at which the bit strings are different
         let firstDiff = FindFirstDiff(bits1, bits2);
 
-        // Hadamard corresponding qubit to create superposition
         H(qs[firstDiff]);
 
-        // iterate through the bit strings again setting the final state of qubits
         for i in 0 .. Length(qs) - 1 {
             if bits1[i] == bits2[i] {
-                // if two bits are the same, apply X or nothing
                 if bits1[i] {
                     X(qs[i]);
                 }
             } else {
-                // if two bits are different, set their difference using CNOT
                 if i > firstDiff {
                     CNOT(qs[firstDiff], qs[i]);
                     if bits1[i] != bits1[firstDiff] {
--- a/katas/content/superposition/wstate_power_of_two/Verification.qs
+++ b/katas/content/superposition/wstate_power_of_two/Verification.qs
@@ -6,7 +6,6 @@ namespace Kata.Verification {
         let N = Length(qs);
 
         if N == 1 {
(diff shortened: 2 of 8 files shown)


katas/content/index.json
```diff
@@ -12,5 +12,6 @@
   "oracles",
   "deutsch_algo",
   "deutsch_jozsa",
-  "qec_shor"
+  "qec_shor",
+  "superposition"
```
frtibble, line 16: TODO: remove before merge

katas/content/superposition/wstate_power_of_two/solution.md
```diff
@@ -0,0 +1,48 @@
...
+
+2. The next instance, $N = 2$, requires preparing $|W_2\rangle = \frac{1}{\sqrt2}\big(|10\rangle + |01\rangle\big)$. It matches one of the Bell states we've seen earlier, but preparing it will be more interesting (and more useful for the next steps!) if we think of it in recursive terms. Let's see how to express $|W_2\rangle$ in terms of $|W_1\rangle$:
+
+$$|W_2\rangle = \frac{1}{\sqrt2}\big(|10\rangle + |01\rangle\big) = \frac{1}{\sqrt2}\big(|W_1\rangle \otimes |0\rangle + |0\rangle \otimes |W_1\rangle\big)$$
+
+This representation suggests us a solution: "split" the starting state $|00\rangle$ in two terms, prepare $|W_1\rangle$ on the first qubit for the first term and on the second qubit - for the second term.
+To do this, we can again use an auxiliary qubit prepared in the $|+\rangle$ state and control the preparation of $|W_1\rangle$ state on the first or the second qubit based on the state of the auxiliary qubit:
+
+$$|0\rangle_{aux} |00\rangle_{reg} \overset{H}{\longrightarrow}\frac{1}{\sqrt2}(|0\rangle + |1\rangle)_{aux} \otimes |00\rangle_{reg} = \frac{1}{\sqrt2}(|0\rangle_{aux} |00\rangle_{reg} + |1\rangle_{aux} |00\rangle_{reg})\overset{CNOT_0}{\longrightarrow}$$
```
frtibble, line 12: I have been struggling with getting this formatted correctly, something is still wrong though I'm not sure where

frtibble: This change fixed it, but needs checking it's still equivalent: https://github.com/microsoft/qsharp/pull/1457/commits/821913a601913a2a929af0bd425b38165eac395f

katas/content/superposition/wstate_power_of_two/Verification.qs
```diff
@@ -0,0 +1,84 @@
...
+                    X(qs[i]);
+                }
+            } else {
+                if i > firstDiff {
+                    CNOT(qs[firstDiff], qs[i]);
+                    if bits1[i] != bits1[firstDiff] {
+                        X(qs[i]);
+                    }
+                }
+            }
```
tcNickolas, line 50: This approach works nicely for N=1 and N=2, but if the solution happens to be correct for these two but incorrect for one of the larger Ns, instead of a nice readable message we get a crash because of the use of an auxiliary qubit in the reference solution (if the solution is incorrect, applying adjoint of the reference solution ends with this aux qubit in a non-zero state and an error about that). 

This is very confusing, so I think we need to sneak a peek at the next task and use the reference solution from it that uses arbitrary rotations but not extra qubits - this way we always get readable results.

```suggestion
    open Microsoft.Quantum.Math;

    operation WState_PowerOfTwo_Reference (qs : Qubit[]) : Unit is Adj + Ctl {
        let N = Length(qs);
        Ry(2.0 * ArcSin(Sqrt(1.0/IntAsDouble(N))), qs[0]);
        for i in 1 .. N - 1 {
            ApplyControlledOnInt(0, Ry(2.0 * ArcSin(Sqrt(1.0/IntAsDouble(N - i))), _), qs[0 .. i-1], qs[i]);
```

katas/content/superposition/wstate_power_of_two/Verification.qs
```diff
@@ -0,0 +1,84 @@
...
+
+        Message($"Testing for N = 2...");
+        if not CheckOperationsEquivalenceOnZeroStateWithFeedback(
+            Kata.WState_PowerOfTwo,
+            TwoBitstringSuperposition_Reference(_, [false, true], [true, false]),
+            2) {
+            return false;
+        }
+
+        for n in [4, 8, 16] {
```
tcNickolas, line 72: And this change allows us to shorten the test quite a bit:
```suggestion
        for n in [1, 2, 4, 8, 16] {
```

tcNickolas (approved):
Looks great! My only concern is about the exception that happens in verification if the solution is correct for N=1 and 2 but incorrect for larger ones, so I suggest using a different reference solution, without extra qubits. All my comments are suggestions, I'll apply them and merge so that you don't have to :-)

Thank you!

katas/content/superposition/wstate_power_of_two/solution.md
tcNickolas: Looks great to me, though I'm looking at the finished version already :-)

katas/content/index.json
tcNickolas, line 16:
```suggestion
  "qec_shor"
```

PR #1459, opened 2024-05-01
LaTeX state representation in Jupyter

minestarks:
Implement `_repr_latex_` as per IPython spec (see https://ipython.readthedocs.io/en/stable/config/integrating.html) so that Jupyter editors can provide the option of displaying the output as LaTeX .

latex


pip/qsharp/_qsharp.py
```diff
@@ -343,6 +343,9 @@ def __str__(self) -> str:
     def _repr_html_(self) -> str:
         return self.__data._repr_html_()
 
+    def _repr_latex_(self) -> str:
+        return self.__data._repr_latex_()
```
DmitryVasilevsky, line 347: Make sure that there's something displayed when LaTeX representation isn't available.

DmitryVasilevsky, line 347: Make sure that there's something displayed when LaTeX representation isn't available.

DmitryVasilevsky (approved):

billti (approved):

swernli (approved):

idavis (approved):

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -309,7 +309,7 @@ export function onCompilerEvent(msg: string, eventTarget: IQscEventTarget) {
     case "DumpMachine":
       qscEvent = makeEvent("DumpMachine", {
         state: qscMsg.state,
-        stateLatex: qscMsg.stateLatex,
+        stateLatex: qscMsg.stateLatex, // can be null or undefined or something, check
```
minestarks, line 312:
```suggestion
        stateLatex: qscMsg.stateLatex,
```

pip/qsharp/_qsharp.py
minestarks: Fixed, tested and confirmed that jupyter falls back to HTML when this returns `None`. Thanks!

Issue #1317, opened 2024-03-26
Mitigating confusing global phase


tcNickolas:
Discussion landed at a few concrete tasks to help with mitigating how confusing global phase in simulation can be:

- [x] #1370
- [x] #1388
- [x] #1450
- [x] https://github.com/qir-alliance/qir-runner/issues/162
- [x] Pull in fix for Rx/Ry phase by updating simulator tag (#1461)
- [x] Add equivalence check member function for StateDump in Python to make checking states easier with global phase (#1372)
- [x] Ensure `DumpRegister` does not add phase on display (#1461)
- [x] https://github.com/qir-alliance/qir-runner/pull/173
- [x] Pull in fix for Rx/Ry of 2 Pi by updating simulator tag (#1551)
- [x] https://github.com/qir-alliance/qir-runner/pull/186
- [x] Pull in fix for rotations by 0.0 by updating simulator tag (#1641)

**Describe the bug**

QDK sparse simulator introduces a global phase at random that shows up in `DumpMachine` output (and the outputs derived from it, such as `dump_operation`).

This is very confusing for several ways:
1. Pedagogically, global phase is hard enough to explain on its own, and if it appears randomly in contexts where there's nowhere for it to come from, this is very confusing for the learner (in the katas output, in assignments, in experimentation they do on their own, and so on.)
2. Writing tests becomes a lot more elaborate compared, say, to Qiskit: where in Qiskit I would just write `assert state_vector == pytest.approx(a)` to compare the state vectors, in Q# I have to first detect the global phase difference between the expected state and the actual state and then as I loop over the elements to remember to include it in every element comparison.

**To Reproduce**

Several examples:
* If you run our sample BellState.qs several times, sometimes the last two states have an extra -1 phase.
* The test [TestPreparationCompletion](https://github.com/microsoft/qsharp/blob/29a1039154726438ced8dcbbdbda76c9aadff2e0/library/src/tests/resources/state_preparation.qs#L63) in our libraries prepares states with real amplitudes, but the expected DumpMachine output (starting with states that have both positive and negative) [has complex amplitudes](https://github.com/microsoft/qsharp/blob/29a1039154726438ced8dcbbdbda76c9aadff2e0/library/src/tests/state_preparation.rs#L138).
* In the katas, phase oracles that are different by a global phase produce the same output when applied to the same state.
[...]

PR #1461, opened 2024-05-01
Fix global phase for `PauliI` rotation and `DumpRegister`

### LINKED ISSUE #1450: Use global phase intrinsic for `R` with `PauliI` to make application of phase more consistent
The current decompositions treat `R` with `PauliI` as identity, which can create confusion when comparing to mathematical behavior where Ri is expected to introduce a global phase. Instead, we should introduce a global phase intrinsic that is used during simulation to apply expected global phase changes and incorporate that into the behavior of the `R1` gate as well. This global phase intrinsic should be treated as a no-op during compilation for hardware, so it is not a `__quantum__qis__` intrinsic but rather a simulation specific intrinsic.

swernli:
This change introduces a new, stdlib-only intrinsic for applying global phase during simulation to make phase behavior more consisten and fixes #1450. This intrinsic only applies the global phase during simulation; the new `GlobalPhase` intrinsic is a no-op for QIR generation, circuit generation, and resource estimation. The change also includes updates to the stdlib and associated tests to account for the corrected global phase behavior.

This also fixes a bug in `DumpRegister` related to iteration through the state vector that ensures the "base" amplitude used for separating the state is consistently chosen, which avoids corner cases that could introduce a phase in the displayed output that wasn't present in the underlying state.


swernli:
@tcNickolas Ok, with the combination of these changes and the teleport sample update from #1456, I've confirmed a few things of note:
 - The teleportation sample now always returns consistent before and after dumps with the expected phases.
 - `R(PauliI)` behaves as an application of global phase consistent with how it is documented.
 - `R1` no longer introduces an extra global phase and is consistent with the other rotations like `Z` and `S` rather than being off by a phase (and introduced unit tests to make sure it stays that way).
 - Application of controlled global phase behave as expected in simulation and introduce calls to `Rz` where appropriate so that execution on physical systems is consistent.
 - The updated state preparation tests confirm that calls `PreparePureStateD` now results in states without complex phase that match the inputs provided.
 - The fixes to `R1` exposed a bug in the `ApplyQFT` test cases that, when corrected, confirms the Fourier transform is being performed as expected.

There is still one other issue remaining from #1317 that will require changes to the internals of the simulator, which will be in another PR (hopefully out soon).

compiler/qsc_eval/src/backend.rs
```diff
@@ -264,8 +264,28 @@ impl Backend for SparseSim {
...
+                let [ctls_val, theta] = &*arg.unwrap_tuple() else {
+                    panic!("tuple arity for GlobalPhase intrinsic should be 2");
+                };
+                let ctls = ctls_val
+                    .clone()
+                    .unwrap_array()
+                    .iter()
+                    .map(|q| q.clone().unwrap_qubit().0)
+                    .collect::<Vec<_>>();
+                let q = self.sim.allocate();
```
minestarks, line 281: Uh oh.... this additional allocation breaks a delicate assumption. Can't yet figure out the concrete ramifications of it though.

https://github.com/microsoft/qsharp/blob/baa771fd4744d52013af268ef4d16dd1ee61e1ea/compiler/qsc_eval/src/backend.rs#L427-L431

minestarks: I guess my question is: does `sim.release()` below decrement the next qubit ID so that the _next_ call to `sim.allocate()` will return the same qubit ID as last time? If the answer is yes, then I think we're in the clear.

minestarks (approved):
The actual global phase fix is beyond my knowledge, but the evaluator/backend bits LGTM

compiler/qsc_eval/src/backend.rs
swernli: Yes, good double check. Short answer, you are correct that once a qubit is released it is available to be reused by later calls to allocate. So as you say, we should be safe here.

The longer answer is interesting because it shows how we are relying on some inside knowledge of the sparse simulator to know this short-cut for global phase is actually quite efficient. In the sparse simulator, allocating a new qubit is almost free (some small id maps are upated, but the quantum state itself doesn't have to change at all). Since the qubit starts in the zero state, no bits in the state indices need to change. Then applying Rz with the angle multiplied by -2 causes every state to be multiplied by the desired value that induces the global state without changing the size of the state vector. Finally, releasing the qubit normally requires a collapse of the state, but because the qubit has no entries in the state vector that have a "1" in the corresponding bit, so the release is nearly a no-op (just the map updates). These tricks mean we effectively short-cut to just multiplying the whole state vector by the desired value.

ivanbasov (approved):

idavis (approved):

cesarzc (approved):

DmitryVasilevsky (approved):

PR #1463, opened 2024-05-02
Handle impossible factories in RE API

msoeken:
Currently, the factory builder API returns a vector of factories for space-time evaluation. However, returning no factory may mean that the algorithm was too fast, and the optimizer will continue to search for some. But there are some cases in which no factory can be found given the input parameters, no matter of the duration of the algorithm.

This PR changes the API to allow to indicate if no magic factories can be computed by returning `Option<Vec<...>>` instead of `Vec<...>`.

### DIFF
--- a/resource_estimator/src/estimates/error.rs
+++ b/resource_estimator/src/estimates/error.rs
@@ -37,6 +37,12 @@ pub enum Error {
     #[error("No solution found for the provided maximum number of physical qubits.")]
     #[diagnostic(code("Qsc.Estimates.MaxPhysicalQubitsTooSmall"))]
     MaxPhysicalQubitsTooSmall,
+    /// Resource estimation configuration can never produce T states
+    ///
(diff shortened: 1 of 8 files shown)


ivanbasov:
Is there any semantic difference in the case between returning an empty array vs returning None?

msoeken:
> Is there any semantic difference in the case between returning an empty array vs returning None?

Yes, there is. In case of an empty array, we would increase the number of cycles and continue trying, see here:

https://github.com/microsoft/qsharp/blob/95019fd19483697037e24c19cfcbc6cead13a7f9/resource_estimator/src/estimates/physical_estimation.rs#L639C17-L641C18

and

https://github.com/microsoft/qsharp/blob/main/resource_estimator/src/estimates/physical_estimation.rs#L674

But, there are situations in which there aren't just any factories for the provided output error rate, then the factory has the chance to return `None`.

ivanbasov (approved):

PR #1466, opened 2024-05-02
Documentation in the VSCode - core, std, and current project

DmitryVasilevsky:
This PR adds "Q#: Show API documentation" command to the command palette (F1). This command opens WebView window with concatenated documentation autogenerated from all doc comments for the standard library and currently opened project. It is searchable via FindWidget (Ctrl+F). This PR also enables FindWidget for all other WebView windows.

Future work:
- Documentation is in random order. We may consider ordering it by namespace then function names
- We may consider adding filter to the documentation - show only pieces that contain typed word
- Telemetry


npm/qsharp/src/compiler/compiler.ts
```diff
@@ -253,6 +254,17 @@ export class Compiler implements ICompiler {
     return this.wasm.generate_docs();
   }
 
+  async getCombinedDocumentation(): Promise<string> {
+    const docFiles: IDocFile[] = this.wasm.generate_docs();
+    let content = "";
+    for (const file of docFiles) {
+      if (file.metadata.indexOf("qsharp.name:") >= 0) {
```
billti, line 261: I don't implicitly understand why this block is here or what problem it's solving. Can you add some comments. Thanks.

DmitryVasilevsky: Added comments.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -277,6 +278,22 @@ export class Compiler implements ICompiler {
...
+    const docFiles: IDocFile[] = this.wasm.generate_docs();
+    // Create combined documentation in the worker
+    let content = "";
+    for (const file of docFiles) {
+      // Some files may contain information other than documentation
+      // For example, table of content is a separate file in a special format
+      // We check presence of qsharp.name in metadata to make sure we take
+      // only files that contain documentation from some qsharp object.
+      if (file.metadata.indexOf("qsharp.name:") >= 0) {
+        content += file.contents + "\n---\n";
```
billti, line 291: It might be helpful to pass a collection of documents, rather than one giant concatenated string. I can imaging (especially once we have docs for user code or references packages) that we might want to sort, filter, collapse, etc. by namespace. If it's all one long string we just have to parse it and break it all apart again somehow.

vscode/package.json
```diff
@@ -316,6 +320,11 @@
         "title": "Show circuit",
         "category": "Q#"
       },
+      {
+        "command": "qsharp-vscode.showDocumentation",
+        "title": "Show documentation",
```
billti, line 325: Maybe `Open API documentation` would be more informative. This could read like you're going to show documentation for the Q# language or the QDK docs.

DmitryVasilevsky: I don't like "open" as it suggests we open something that's already exists. Like open a file. I can thing of "Show auto-generated API documentation" or simply "Generate API documentation".

billti: We're not really 'generating' anything (as far as the user is concerned), just showing them docs that already exist on the code. `Show API documentation` is fine if you don't like the word `Open`.

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: Updated. I wanted to put most processing in the worker. But I moved concatenation out of it now.

vscode/package.json
DmitryVasilevsky: OK. Show API documentation it is.

vscode/src/documentation.ts
```diff
@@ -0,0 +1,54 @@
...
+  const documentation = await worker.getDocumentationContent(
+    program.sources,
+    targetProfile,
+    program.languageFeatures,
+  );
+
+  // Concatenate all documentation.
+  // The following adds an empty line and a horizontal line
+  // between documentation for different functions.
+  const content = documentation.join("\n<pre>\n\n</pre>\n---\n");
```
billti, line 42: Just use a `<br>` tag in html to insert a line break, don't `pre` format regular line breaks.

vscode/src/webview/docview.tsx
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Markdown } from "qsharp-lang/ux";
+
+export function DocumentationView(props: { contentToRender: string }) {
```
billti, line 6: If this component isn't going to add any functionality or formatting, just use the `Markdown` component directly. This component doesn't add anything.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -93,7 +93,12 @@ export interface ICompiler {
     operation?: IOperationInfo,
   ): Promise<CircuitData>;
 
-  getDocumentation(): Promise<IDocFile[]>;
+  getDocumentationFiles(): Promise<IDocFile[]>;
+  getDocumentationContent(
+    additionalSources?: Array<any>[],
```
billti, line 98: Is this really an array of arrays of any? Seems like this could have a better type.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -273,10 +278,41 @@ export class Compiler implements ICompiler {
...
+  ): Promise<string[]> {
+    // Get documentation from wasm layer
+    const docFiles: IDocFile[] = this.wasm.generate_docs(
+      additionalSources,
+      targetProfile,
+      languageFeatures,
+    );
+
+    const result: string[] = [];
+    for (const file of docFiles) {
```
minestarks, line 304: FWIW I think this is too much business logic for this layer. As much as possible, `compiler.ts` should call the wasm methods and pass the result through without modification, otherwise it becomes really hard to follow what's happening where.

I'm in favor of returning to having just one `getDocumentation()` method (with the new, optional parameters) , and moving this filtering to `vscode/src/documentation.ts`, the only current place it's used.

Keeping this layer simple will also minimize API breaking changes (though I don't think this is used by QCOM or anything today).

minestarks (approved):
Only request is my comment in compiler.ts about having a single `getDocumentation()` method. Assuming the CSS is being worked on, LGTM otherwise. Thanks!

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: I was thinking about unifying these two get functions but wanted to put some logic in there... Getting "files" from this API isn't convenient for anything other than just placing these files in the file system (which generation of the actual documentation does). I need to thing about this a bit more...

playground/build.js
```diff
@@ -58,7 +58,7 @@ function copyLibs() {
   mkdirSync(monacoDest, { recursive: true });
   cpSync(monacoBase, monacoDest, { recursive: true });
 
-  copyKatex(join(thisDir, "public/libs/katex"), true);
+  copyKatex(join(thisDir, "public/libs/katex"));
```
billti, line 61: Changed this as we just copy both light and dark CSS files for both playground and VS Code now.

vscode/src/documentation.ts
```diff
@@ -0,0 +1,54 @@
...
+import { isQsharpDocument } from "./common";
+import { getTarget } from "./config";
+import { Uri, window } from "vscode";
+import { loadProject } from "./projectSystem";
+import { sendMessageToPanel } from "./webviewPanel";
+
+export async function showDocumentationCommand(extensionUri: Uri) {
+  // Reveal panel an show 'Loading...' for immediate feedback.
+  sendMessageToPanel(
+    "documentation", // This is needed to route the message to the proper panel
```
billti, line 14: I changed the string here to be consistent across the panel and the calling code, else reloading the VS Code window doesn't deserialize and reload the state (i.e. old content) correctly.

vscode/src/documentation.ts
```diff
@@ -0,0 +1,54 @@
...
+  const documentation = await worker.getDocumentationContent(
+    program.sources,
+    targetProfile,
+    program.languageFeatures,
+  );
+
+  // Concatenate all documentation.
+  // The following adds an empty line and a horizontal line
+  // between documentation for different functions.
+  const content = documentation.join("<br/>\n\n---\n\n");
```
billti, line 42: Changed this to resolve the prior comment I left about not using `<pre>` with line breaks.

vscode/src/webview/webview.tsx
```diff
@@ -15,12 +15,13 @@ import {
...
 } from "qsharp-lang/ux";
 import { HelpPage } from "./help";
+import { DocumentationView } from "./docview";
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore - there are no types for this
 import mk from "@vscode/markdown-it-katex";
 import markdownIt from "markdown-it";
-const md = markdownIt("commonmark", { html: true, breaks: true });
+const md = markdownIt("commonmark");
```
billti, line 24: The options were the default and redundant, so I removed them for consistency with other usage in the code.

vscode/src/webview/webview.tsx
```diff
@@ -49,19 +50,78 @@ type CircuitState = {
...
   | HistogramState
   | EstimatesState
-  | CircuitState;
+  | CircuitState
+  | DocumentationState;
 const loadingState: State = { viewType: "loading" };
 const helpState: State = { viewType: "help" };
 let state: State = loadingState;
 
+const themeAttribute = "data-vscode-theme-kind";
```
billti, line 69: Hopefully the comments are explanatory, but this is what watches for theme changes in VS Code and switches the stylesheet as needed.

vscode/src/webview/webview.tsx
```diff
@@ -178,6 +246,12 @@ function App({ state }: { state: State }) {
       return <CircuitPanel {...state.props}></CircuitPanel>;
     case "help":
       return <HelpPage />;
+    case "documentation":
+      // Ideally we'd have this on all web views, but it makes the font a little
```
billti, line 250: Per the comments, only added this class to the docs now, not the webview in general, as I don't want to impact rendering of other existing views.

vscode/src/documentation.ts
DmitryVasilevsky: You already fixed this.

vscode/src/webview/docview.tsx
DmitryVasilevsky: This is the component that will be used to filter content later. I placed content concatenation here as this is the only processing that we have at the moment.

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: I changed it to [string, string][], like on other similar functions that take sources.

npm/qsharp/src/compiler/compiler.ts
DmitryVasilevsky: I decided to keep single getDocumentation function like it used to be, only add three optional parameters to it.

vscode/src/documentation.ts
DmitryVasilevsky: Updated PR description and removed this as not yet implemented.

playground/public/index.html
```diff
@@ -16,7 +16,7 @@
       rel="icon"
       href="data:image/gif;base64,R0lGODlhEAAQAAAAACwAAAAAEAAQAAACDvAxdbn9YZSTVntx1qcAADs="
     />
-    <link rel="stylesheet" href="libs/katex/github-markdown.css" />
+    <link rel="stylesheet" href="libs/katex/github-markdown-light.css" />
```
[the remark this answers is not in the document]
billti: It was intentional. DO NOT change it back.

vscode/build.mjs
```diff
@@ -78,12 +77,13 @@ export function copyKatex(destDir, useLightTheme) {
...
-    : "github-markdown.css";
   copyFileSync(
-    join(libsDir, `github-markdown-css/${cssFileName}`),
-    join(katexDest, "github-markdown.css"),
+    join(libsDir, `github-markdown-css/github-markdown-light.css`),
+    join(katexDest, "github-markdown-light.css"),
+  );
+  copyFileSync(
+    join(libsDir, `github-markdown-css/github-markdown-dark.css`),
+    join(katexDest, "github-markdown-dark.css"),
```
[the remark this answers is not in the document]
billti: It is intended. It's part of the change I mentioned with switching the style sheet based on the VS Code theme (rather than using the one style sheet which uses the OS theme).

vscode/src/documentation.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export async function showDocumentationCommand(extensionUri: Uri) {
+  // Reveal panel and show 'Loading...' for immediate feedback.
+  sendMessageToPanel(
+    "documentation", // This is needed to route the message to the proper panel
+    true, // Reveal panel
+    null, // With no message
+  );
+  const editor = window.activeTextEditor;
+  if (!editor || !isQsharpDocument(editor.document)) {
```
billti, line 19: You should probably do this check _before_ you open the panel in line 13 above.

billti (approved):

vscode/src/documentation.ts
DmitryVasilevsky: Yes! Updated.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -28,9 +29,38 @@ struct Compilation {
 
 impl Compilation {
     /// Creates a new `Compilation` by compiling sources.
-    pub(crate) fn new() -> Self {
+    pub(crate) fn new(
+        additional_sources: Option<SourceMap>,
```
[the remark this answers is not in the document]
DmitryVasilevsky: This function was used to generate API documentation for the standard library only. Now it can generate documentation for the standard library and additional sources. But I see how this name can be confusing. I don't want to name them just 'source' because the whole standard library is always included. I added comment to the generate_docs function below in this file.

PR #1467, opened 2024-05-03
[Samples-Array] Fix typo

Manvi-Agrawal:
Noticed a small typo in `Arrays` sample

### DIFF
--- a/samples/language/Array.qs
+++ b/samples/language/Array.qs
@@ -27,7 +27,7 @@ namespace MyQuantumApp {
         let stringArray = ["a", "string", "array"];
 
         // A new array expression creating the array `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`
-        let intArray = Repeated(0, 10);
+        let repeatedArray = Repeated(0, 10);
 
         // Arrays can be sliced with ranges.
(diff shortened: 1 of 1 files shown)


samples/language/Array.qs
```diff
@@ -27,7 +27,7 @@ namespace MyQuantumApp {
         let stringArray = ["a", "string", "array"];
 
         // A new array expression creating the array `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`
-        let intArray = Repeated(0, 10);
+        let repeatedArray = Repeated(0, 10);
```
Manvi-Agrawal, line 30: Callout: We are redefining `intArray` to `Repeated(0,10)`, due to which slicing will results in elements being zeros. Hence, changed `intArray = Repeated(0,10)` to `repeatedArray = Repeated(0,10)`, so that outputs for array slice is what is mentioned in the sample.

tcNickolas (approved):
Ah, great catch! I wonder if we should rewrite the language samples to actually print what the values should be - that's what I do in my samples for Q# Pocket Guide, and it's a lot easier to verify them by just running them. Let me follow up.

Thank you!

PR #1479, opened 2024-05-04
Superdense coding kata finish

Manvi-Agrawal:
Part of #1185.

Changes Done:
- Revamped Task 1:  Using `qAlice` and `qBob` for better clarity. Replaced `\rangle` with `\ket`
- Added tasks 2-4.
- Using a tuple of boolean variables instead of `ProtocolMessage` to improve learner and developer productivity.


katas/content/superdense_coding/Common.qs
```diff
@@ -0,0 +1,67 @@
+namespace Kata.Verification {
+    open Kata;
```
Manvi-Agrawal, line 2: Oddly enough, removing this gives me : `main.qs@(1,1): type error: type ProtocolMessage cannot be converted into a string`

katas/content/superdense_coding/alice_sends_message/Solution.qs
```diff
@@ -0,0 +1,14 @@
+namespace Kata {
+    newtype ProtocolMessage = (Bit1 : Bool, Bit2 : Bool);
+
```
Manvi-Agrawal, line 3:
```suggestion
```

katas/content/index.json
```diff
@@ -9,6 +9,7 @@
   "preparing_states",
   "single_qubit_measurements",
   "multi_qubit_measurements",
+  "superdense_coding",
   "random_numbers",
```
tcNickolas, line 13: Let's move this one after QRNG, that's usually how these applications are taught, and that matches the order in the classic katas. (Teleportation will be between these two)
```suggestion
  "random_numbers",
  "superdense_coding",
```

katas/content/superdense_coding/Common.qs
tcNickolas: Makes sense, the namespaces right now are not hierarchical, so Kata.Verification doesn't have access to Kata namespace. (That's why we call solutions with Kata. prefix)

katas/content/superdense_coding/alice_sends_message/Placeholder.qs
```diff
@@ -0,0 +1,12 @@
+namespace Kata {
+    newtype ProtocolMessage = (Bit1 : Bool, Bit2 : Bool);
+
+    operation EncodeMessageInQubit(qAlice : Qubit, message : ProtocolMessage) : Unit {
```
tcNickolas, line 4: Looking at this kata in the modern system, I think we should drop ProtocolMessage type altogether, and rewrite it as just a tuple of two bits, like this:
```suggestion
    operation EncodeMessageInQubit(qAlice : Qubit, (bit1, bit2) : (Int, Int)) : Unit {
```

It will make the code much shorter and more straightforward, for example, we won't have to teach bit access in the kata. And in this task it makes things confusing, actually, since we're asking them to implement solution within an if that is supposed to teach bit access, but the solution also needs to include another, independent if outside of this one... Plus then we can drop protocol notes in task description

tcNickolas (changes requested):
I took a quick look, things are looking good overall, but I think we need to get rid of `ProtocolMessage` and just use tuples. The custom types don't really work great with new infrastructure, with us having to redefine them each time, and since it's not structured as a multi-file Q# project, IntelliSense fails to recognize them too... I think using tuples will improve the experience both for the learner and for developer.

I'll take a more detailed pass once you remove `ProtocolMessage`, it will be a lot easier at that point.

Thank you!

katas/content/index.json
Manvi-Agrawal: Done. Thanks for your suggestion.

katas/content/superdense_coding/alice_sends_message/Placeholder.qs
Manvi-Agrawal: Reverted to tuple of Booleans by checking out to previous commit.

katas/content/superdense_coding/index.md
```diff
@@ -18,10 +18,41 @@ We split the superdense coding protocol into several steps:
...
+    "path": "bob_decodes_message",
+    "qsDependencies": [
+        "../KatasLibrary.qs",
+        "./Common.qs"
+    ]
+})
+
+@[exercise]({
+    "id": "superdense_coding__protocol_e2e",
+    "title": "Superdense coding protocol: End to End",
```
Manvi-Agrawal, line 52:
```suggestion
    "title": "Superdense Coding Protocol: End to End",
```

katas/content/superdense_coding/index.md
```diff
@@ -18,10 +18,41 @@ We split the superdense coding protocol into several steps:
...
+    "path": "alice_sends_message",
+    "qsDependencies": [
+        "../KatasLibrary.qs",
+        "./Common.qs"
+    ]
+})
+
+@[exercise]({
+    "id": "superdense_coding__bob_decodes_message",
+    "title": "Bob's task: Decode the Message and Reset the qubits",
```
Manvi-Agrawal, line 42:
```suggestion
    "title": "Bob's task: Decode the Message and Reset the Qubits",
```

Manvi-Agrawal:
> I'll take a more detailed pass once you remove ProtocolMessage, it will be a lot easier at that point.
@tcNickolas , could you please take a look? I have removed `ProtocolMessage` class. Thank you.

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,33 @@
+Recall that we learnt how to prepare all Bell states in "Preparing Quantum States" kata. This is slightly advanced version of that task demonstrating the "power of entanglement", where operations applied to only one qubit affect the state of the system in interesting way.
+
+Let's use the below Bell state we prepared in the previous task:
+
+$$\frac{1}{\sqrt{2}} \big(\ket{00} + \ket{11} \big)$$
+
+We can transform it according to the value of bits:
```
tcNickolas, line 7:
```suggestion
We can transform it into every other Bell state according to the value of `message`:
```

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+- `(0, 0)`: $\ket{\Phi^{+}} = \frac{1}{\sqrt{2}} (\ket{00} + \ket{11})$
```
tcNickolas, line 9: These two lists are rendered as one 8-point list, which looks a bit odd. I would merge the pairs of bullet points that describe the same scenario into one point each, so that each value of message is described and handled right away

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,33 @@
...
+- `(0, 1)`: $\ket{\Psi^{+}} = \frac{1}{\sqrt{2}} (\ket{01} + \ket{10})$
+- `(1, 0)`: $\ket{\Phi^{-}} = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$
+- `(1, 1)`: $\ket{\Psi^{-}} = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10})$
+
+- If `bits == (0, 0)`, we do nothing - the prepared state is already $\ket{\Phi^{+}}$.
+
+- If `bits ==  (0, 1)`, we need to change the second qubit in both $\ket{00}$ and $\ket{11}$ terms. Observe that applying an $X$ gate to Alice's qubit does exactly that:
+  $$X \otimes I (\ket{\Phi^{+}}) = \frac{1}{\sqrt{2}} (\ket{10} + \ket{01})$$
```
tcNickolas, line 17: Here and in the next formula we need brackets around (X \otimes I) rather than around the \ket for Bell state (just like you did in the last formula) - \ket is already a single entity, but getting a two-qubit gate out of the tensor product needs emphasizing

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+- If `bits == (1, 0)`, we need to add a relative phase of $-1$ to the $\ket{11}$ term. Observe that applying $Z$ gate to Alice's qubit does exactly that:
+  $$Z \otimes I (\ket{\Phi^{+}}) = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$$
+
+- If `bits = (1, 1)`, we use the same logic to realize that we need to apply both the $Z$ and $X$ corrections to get $\ket{\Psi^{-}}$ state.
+  $$ (Z \otimes I).(X \otimes I) (\ket{\Psi^{+}}) = (Z \otimes I) \frac{1}{\sqrt{2}} (\ket{10} + \ket{01}) = \frac{1}{\sqrt{2}} (-\ket{10} + \ket{01}) = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10}) $$
```
tcNickolas, line 23:
```suggestion
  $$ (Z \otimes I) \cdot (X \otimes I) \ket{\Psi^{+}} = (Z \otimes I) \frac{1}{\sqrt{2}} (\ket{10} + \ket{01}) = \frac{1}{\sqrt{2}} (-\ket{10} + \ket{01}) = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10}) $$
```

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,33 @@
...
+
+The final sequence of steps is as follows:
+
+1. Apply the $X$ gate to the alice's qubit if `bit2 == 1`.
+2. Apply the $Z$ gate to the alice's qubit if `bit1 == 1`.
```
tcNickolas, line 28:
```suggestion
1. Apply the $X$ gate to Alice's qubit if `bit2 == 1`.
2. Apply the $Z$ gate to Alice's qubit if `bit1 == 1`.
```

katas/content/superdense_coding/bob_decodes_message/Placeholder.qs
```diff
@@ -0,0 +1,14 @@
+namespace Kata {
+    operation DecodeMessageFromQubits(qAlice : Qubit, qBob : Qubit) : (Bool, Bool) {
+        // Implement your solution here...
+
+        // Returning the qubit states after measuring
+        return (MResetZ(qAlice) == One, MResetZ(qBob) == One);
+    }
+
+    // You might find this helper operation from an earlier task useful.
+    operation CreateEntangledPair(qAlice : Qubit, qBob : Qubit) : Unit is Adj {
```
tcNickolas, line 10: I thought we were going to drop this helper from this task and assume that the learner will just type in CNOT and H?

katas/content/superdense_coding/bob_decodes_message/Placeholder.qs
```diff
@@ -0,0 +1,14 @@
+namespace Kata {
+    operation DecodeMessageFromQubits(qAlice : Qubit, qBob : Qubit) : (Bool, Bool) {
+        // Implement your solution here...
+
+        // Returning the qubit states after measuring
+        return (MResetZ(qAlice) == One, MResetZ(qBob) == One);
```
tcNickolas, line 6: I would do a placeholder with `return (false, false)`, and let the learner do the rest - this way we're giving away most of the solution already :-)

katas/content/superdense_coding/bob_decodes_message/Solution.qs
```diff
@@ -0,0 +1,12 @@
+namespace Kata {
+    operation DecodeMessageFromQubits(qAlice : Qubit, qBob : Qubit) : (Bool, Bool) {
+        Adjoint CreateEntangledPair(qAlice, qBob);
```
tcNickolas, line 3: Same here, we can type CNOT and H and drop CreateEntangledPair altogether

katas/content/superdense_coding/bob_decodes_message/solution.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `(0, 0)`: $\ket{\Phi^{+}} = \frac{1}{\sqrt{2}} (\ket{00} + \ket{11})$
+- `(0, 1)`: $\ket{\Psi^{+}} = \frac{1}{\sqrt{2}} (\ket{01} + \ket{10})$
+- `(1, 0)`: $\ket{\Phi^{-}} = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$
+- `(1, 1)`: $\ket{\Psi^{-}} = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10})$
+
+To get our state back, we can undo the entanglement by applying $CNOT$ and $H$ gate.
+Notice that it's important to keep the order right. The qubits that are subject to the Hadamard transform and the $CNOT$ gate in the preparation of the pair have to match the operations below, or the order of the data bits will get flipped.
+
+Notice that [`Adjoint`](https://learn.microsoft.com/azure/quantum/user-guide/language/expressions/functorapplication#adjoint-functor) functor in Q# does exactly that.
```
tcNickolas, line 11: This is confusing to me - we're looking to get a message rather than a state, and it's unclear Adjoint of what operation we're discussing. I'd rephrase as something like this:

```suggestion
To read out the encoded message, Bob needs to figure out which of the four Bell states he has. We can map the Bell states to basis states by applying a $CNOT$ gate with the first qubit as control and the second qubit as target, followed by an $H$ gate on the first qubit. (Notice that this is exactly what [`Adjoint`](https://learn.microsoft.com/azure/quantum/user-guide/language/expressions/functorapplication#adjoint-functor) of the state preparation operation in the first task does.)
```

katas/content/superdense_coding/protocol_e2e/Placeholder.qs
```diff
@@ -0,0 +1,33 @@
...
+
+    // You might find this helper operation from an earlier task useful.
+    operation CreateEntangledPair(qAlice : Qubit, qBob : Qubit) : Unit is Adj {
+        H(qAlice);
+        CNOT(qAlice, qBob);
+    }
+
+    // You might find this helper operation from an earlier task useful.
+    operation EncodeMessageInQubit(qAlice : Qubit, message : (Bool, Bool)) : Unit {
+        let (cbit1, cbit2) = message;
```
tcNickolas, line 16: Same suggestion cbit -> bit here

katas/content/superdense_coding/protocol_e2e/Placeholder.qs
```diff
@@ -0,0 +1,33 @@
+namespace Kata {
+    operation SuperdenseCodingProtocol(message : (Bool, Bool)) : (Bool, Bool) {
+        // Implement your solution here...
+
+        return (false, false);
+    }
+
+    // You might find this helper operation from an earlier task useful.
```
tcNickolas, line 8: I'd use plural here
```suggestion
    // You might find these helper operations from earlier tasks useful.
```
And then we can remove the same comment from the next two tasks, to make them more compact

katas/content/superdense_coding/protocol_e2e/index.md
```diff
@@ -0,0 +1,9 @@
+**Input:** Two classical bits to be transmitted.
+
+**Goal:**
```
tcNickolas, line 3:
```suggestion
**Goal:**
```

katas/content/superdense_coding/protocol_e2e/index.md
```diff
@@ -0,0 +1,9 @@
+**Input:** Two classical bits to be transmitted.
+
+**Goal:**  
+Put together the steps performed in previous tasks to implement the full superdense coding protocol:
+
+- Prepare an EPR Pair
```
tcNickolas, line 6: We've not used this term here before, so let's stick to terms we used
```suggestion
- Prepare a Bell state.
```

katas/content/superdense_coding/protocol_e2e/index.md
```diff
@@ -0,0 +1,9 @@
...
+- Encode the two classical bits in the state of the pair by applying quantum gates to one member of the pair,
```
tcNickolas, line 7:
```suggestion
- Encode the two classical bits in the state of the qubits by applying quantum gates to one qubit of the Bell state,
```

katas/content/superdense_coding/protocol_e2e/index.md
```diff
@@ -0,0 +1,9 @@
...
+- Decode the two classical bits from the state of the pair.
```
tcNickolas, line 8:
```suggestion
- Decode the two classical bits from the resulting state.
```

katas/content/superdense_coding/protocol_e2e/solution.md
```diff
@@ -0,0 +1,11 @@
+We can put together the operations defined in the previous tasks to run superdense protocol E2E.
+
+- Prepare an EPR Pair
```
tcNickolas, line 3: Same update here to match the task description

katas/content/superdense_coding/protocol_e2e/solution.md
```diff
@@ -0,0 +1,11 @@
+We can put together the operations defined in the previous tasks to run superdense protocol E2E.
```
tcNickolas, line 1:
```suggestion
We can put together the operations defined in the previous tasks to run superdense coding protocol end-to-end.
```

tcNickolas (changes requested):
Looks great! I know I left a lot of comments, but they're mostly polish to keep things nice and uniform. One last round of edits, and this is ready to ship :-)

Thank you!

katas/content/superdense_coding/alice_sends_message/Placeholder.qs
```diff
@@ -0,0 +1,14 @@
...
+
+    operation EncodeMessageInQubit(qAlice : Qubit, message : (Bool, Bool)) : Unit {
+        // Get the bits from the message
+        let (bit1, bit2) = message;
+
+        // Accessing bit1
+        if bit1 {
+            // Implement your solution here...
+
+        }
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done, I wanted to do it this way since its cleaner.

I dont know if something changed. Earlier, it was complaining about some return type (Bool, Bool) issue and adding if block suppresed that error :-)

katas/content/superdense_coding/bob_decodes_message/Placeholder.qs
Manvi-Agrawal: I thought of keeping it for the convenience of the learner, in case they might  want to use it. I was more of favor of dropping it here, in case we drop the last End-To-End task, which requires us to copy-paste all the 3 solutions. 

@tcNickolas , Kindly let me know if you think, this should be dropped here. In that case, I was thinking to update the solution text to avoid mention of `Adjoint`  functor.

katas/content/superdense_coding/alice_sends_message/solution.md
Manvi-Agrawal: Done

katas/content/superdense_coding/alice_sends_message/solution.md
Manvi-Agrawal: Nice catch. Updated solution.

katas/content/superdense_coding/index.md
```diff
@@ -18,10 +18,41 @@ We split the superdense coding protocol into several steps:
 - Finally, we compose those steps into the complete superdense coding protocol.
 
 @[exercise]({
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done

katas/content/superdense_coding/alice_sends_message/Solution.qs
```diff
@@ -0,0 +1,15 @@
+namespace Kata {
+
+    operation EncodeMessageInQubit(qAlice : Qubit, message : (Bool, Bool)) : Unit {
+        // Get the bits from the message
+        let (cbit1, cbit2) = message;
```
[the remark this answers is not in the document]
Manvi-Agrawal: Done, initially I wanted to call them cbit in placeholder and solution to denote that they are classical bits. And it explaination I was bit1 with supplementary info in text that said classical bits, etc. It seems that might confuse the learner, so using `bit1` everywhere

katas/content/superdense_coding/bob_decodes_message/Placeholder.qs
Manvi-Agrawal: Thanks for the suggestion. I updated the testing harness to do a reset of the qubits.

I was getting the below error without it, so I used reset operation to get away with that. 

```text
[ERR_ASSERTION]: Exercise "superdense_coding__bob_decodes_message" has compilation or runtime errors
        //  when using the placeholder as solution. Compilation and runtime errors:
        //   Message = runtime error: Qubit1 released while not in |0⟩ state
       ```

**Comment by Manvi-Agrawal** [comment]
> Looks great! I know I left a lot of comments, but they're mostly polish to keep things nice and uniform. One last round of edits, and this is ready to ship :-)
@tcNickolas , thanks for your review. I have addressed most of your comments. Could you please take a look again?

File: katas/content/superdense_coding/index.md [hunk 27]
```diff
@@ -18,10 +18,41 @@ We split the superdense coding protocol into several steps:
...
+    "path": "alice_sends_message",
+    "qsDependencies": [
+        "../KatasLibrary.qs",
+        "./Common.qs"
+    ]
+})
+
+@[exercise]({
+    "id": "superdense_coding__bob_decodes_message",
+    "title": "Bob's task: Decode the Message and Reset the Qubits",
```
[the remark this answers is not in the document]
Manvi-Agrawal: Changed. This was in the previous kata:-)

katas/content/superdense_coding/bob_decodes_message/Placeholder.qs
tcNickolas: I would drop it here (but keep it in the e2e task, since you're right that we need all 3 solutions in it anyways). We can mention the Adjoint in the solution text even if we're not using it in the code - it's a useful bit of knowledge that using adjoint of state prep allows us to do measurement in Bell state even if we choose to implement adjoint by hand

Manvi-Agrawal: @tcNickolas, done.

Manvi-Agrawal:
@tcNickolas , thanks for your review. I have addressed all your comments. Could you please take a look again?

katas/content/superdense_coding/alice_sends_message/index.md
```diff
@@ -0,0 +1,18 @@
...
+<details>
+  <summary><b>Need a hint?</b></summary>
+Manipulate Alice's half of the entangled pair to change the joint state of the two qubits to one of the following four states based on the value of message:
+
+- `(0, 0)`: $\ket{\Phi^{+}} = \frac{1}{\sqrt{2}} (\ket{00} + \ket{11})$
+- `(0, 1)`: $\ket{\Psi^{+}} = \frac{1}{\sqrt{2}} (\ket{01} + \ket{10})$
+- `(1, 0)`: $\ket{\Phi^{-}} = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$
+- `(1, 1)`: $\ket{\Psi^{-}} = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10})$
+
+</details>
```
tcNickolas, line 18: I just realized that we don't actually explain here how we test the solution. In all the previous katas, if the outcome of a task was expected to be a state, like here, the results would be a state comparison, but here for some reason we're talking about transfer. (And this was a problem in the classic kata too, I just never looked at it with a fresh eye! Revisiting everything as part of this migration is not a bad thing :-))

I suggest we rephrase a bit, to give more context on testing and to move expected encoding from hint to main task - strictly speaking, it's possible to implement superdense coding with a different mapping of messages to Bell states, it's just the customary approach:
```suggestion
Superdense coding protocol changes the joint state of the two qubits to one of the following four states based on the value of message:

- `(0, 0)`: $\ket{\Phi^{+}} = \frac{1}{\sqrt{2}} (\ket{00} + \ket{11})$
- `(0, 1)`: $\ket{\Psi^{+}} = \frac{1}{\sqrt{2}} (\ket{01} + \ket{10})$
- `(1, 0)`: $\ket{\Phi^{-}} = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$
- `(1, 1)`: $\ket{\Psi^{-}} = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10})$

Note that your solution is tested as part of an end-to-end implementation of the protocol; the goal is to get the message transmitted correctly.
```

katas/content/superdense_coding/alice_sends_message/solution.md
```diff
@@ -0,0 +1,35 @@
...
+$$\frac{1}{\sqrt{2}} \big(\ket{00} + \ket{11} \big)$$
+
+We can transform it into every other Bell state according to the value of `message`:
+
+- `(0, 0)`: $\ket{\Phi^{+}} = \frac{1}{\sqrt{2}} (\ket{00} + \ket{11})$
+- `(0, 1)`: $\ket{\Psi^{+}} = \frac{1}{\sqrt{2}} (\ket{01} + \ket{10})$
+- `(1, 0)`: $\ket{\Phi^{-}} = \frac{1}{\sqrt{2}} (\ket{00} - \ket{11})$
+- `(1, 1)`: $\ket{\Psi^{-}} = \frac{1}{\sqrt{2}} (\ket{01} - \ket{10})$
+
+### Detailed Explaination
```
tcNickolas, line 14: We use subheaders as headers, when there's a different solution or a different topic raised. Here, the goal of the solution is already to provide a detailed explanation, so I'd rephrase
```suggestion
Here is how we can perform this transformation:
```

katas/content/superdense_coding/bob_decodes_message/index.md
```diff
@@ -0,0 +1,6 @@
+**Inputs:**
+
+1. `qAlice` : Qubit received from Alice.
+2. `qBob` : Bob's part of the entangled pair.
+
+**Goal** :  Decode the message using the qubit received from Alice and reset both qubits to a $\ket{00}$ state. For this, retrieve two bits of classic data from the qubits and return a tuple of the form `(Bool, Bool)` to represent `message`. The state of the qubits in the end of the operation should be $\ket{00}$.
```
tcNickolas, line 6: We need a note on testing here as well. Also, since we've modified the testing harness, there is no requirement to reset qubits to 0, the solution passes without that

```suggestion
**Goal** :  Decode the message using the qubit received from Alice. For this, retrieve two bits of classic data from the qubits and return a tuple of the form `(Bool, Bool)` to represent `message`. The state of the qubits in the end of the operation does not matter.

Note that your solution is tested as part of an end-to-end implementation of the protocol; the goal is to get the message transmitted correctly.
```

katas/content/superdense_coding/Common.qs
```diff
@@ -0,0 +1,79 @@
...
+        CreateEntangledPairWrapper_Reference([qAlice, qBob]);
+
+        encodeOp(qAlice, message);
+
+        let (bit1, bit2) = decodeOp(qAlice, qBob);
+
+        // We get this error if dont call Reset
+        // [ERR_ASSERTION]: Exercise "superdense_coding__bob_decodes_message" has compilation or runtime errors
+        //  when using the placeholder as solution. Compilation and runtime errors:
+        //   Message = runtime error: Qubit1 released while not in |0⟩ state
```
tcNickolas, line 47: You don't have to explain the Reset. It's a known behavior that qubits have to be in 0 state before release, you'll notice that, for example, in state preparation tasks, if the state prepared is not what we want (and thus after adjoint is applied the state is not 0), we reset the state before returning false, without much emphasis really:
```suggestion
```

tcNickolas (approved):
Looks great! Of course, I had a realization that we're not explaining why the messages from testing harness look the way they do, so I think we need to add notes about that to two tasks... But that's the final round of suggestions, promise!

Thank you!

katas/content/superdense_coding/index.md
```diff
@@ -18,10 +18,38 @@ We split the superdense coding protocol into several steps:
...
+})
+
+@[exercise]({
+    "id": "superdense_coding__protocol_e2e",
+    "title": "Superdense Coding Protocol: End to End",
+    "path": "protocol_e2e",
+    "qsDependencies": [
+        "./Common.qs"
     ]
 })
```
tcNickolas, line 55: And another thing I just realized - we try to add a little conclusion to the end of the katas even if they didn't have them before
```suggestion
})


@[section]({
    "id": "superdense_coding__conclusion",
    "title": "Conclusion"
})

Congratulations! In this kata you learned to implement superdense coding protocol.
```

katas/content/superdense_coding/Common.qs
Manvi-Agrawal: I thought of adding here in case someone ends up getting confused of the difference from original harness. But probably we can omit that, since its us developers only who would be referring old katas for short time :-)

katas/content/superdense_coding/alice_sends_message/index.md
Manvi-Agrawal: Nice catch, Done :)

katas/content/superdense_coding/index.md
```diff
@@ -18,10 +18,46 @@ We split the superdense coding protocol into several steps:
...
+
+@[exercise]({
+    "id": "superdense_coding__protocol_e2e",
+    "title": "Superdense Coding Protocol: End to End",
+    "path": "protocol_e2e",
+    "qsDependencies": [
+        "./Common.qs"
+    ]
+})
+
```
Manvi-Agrawal, line 56:
```suggestion
```

tcNickolas (approved):
Looks great, thank you!

PR #1480, opened 2024-05-04
Partial evaluation support for assignment expressions and multiple array related expressions

cesarzc:
This change adds support in partial evaluation for the following expressions:
- Assignment.
- Array repeat.
- Array in-place concatenation.
- Array in-place update of value at index.
- Array access to value at index.

### DIFF
--- a/compiler/qsc_partial_eval/src/lib.rs
+++ b/compiler/qsc_partial_eval/src/lib.rs
@@ -440,7 +440,7 @@ impl<'a> PartialEvaluator<'a> {
         };
 
         // If this was an assign expression, update the bindings in the hybrid side to keep them in sync and to insert
-        // store instructions if needed.
+        // store instructions for variables of type `Bool`, `Int` or `Double`.
         if let Ok(EvalControlFlow::Continue(_)) = eval_result {
             let expr = self.get_expr(expr_id);
             if let ExprKind::Assign(lhs_expr_id, _)
@@ -571,7 +571,7 @@ impl<'a> PartialEvaluator<'a> {
             ));
         };
 
-        self.update_hybrid_bindings(lhs_expr_id, rhs_value)?;
+        self.update_bindings(lhs_expr_id, rhs_value)?;
         Ok(EvalControlFlow::Continue(Value::unit()))
     }
 
@@ -613,7 +613,7 @@ impl<'a> PartialEvaluator<'a> {
             .expect("could not convert array index into usize");
         let mut array = (*array_value.unwrap_array()).clone();
         let _ = std::mem::replace(&mut array[index], replace_value);
-        self.update_hybrid_bindings(array_expr_id, Value::Array(array.into()))?;
+        self.update_bindings(array_expr_id, Value::Array(array.into()))?;
         Ok(EvalControlFlow::Continue(Value::unit()))
     }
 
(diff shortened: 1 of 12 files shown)


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -230,23 +282,146 @@ impl<'a> PartialEvaluator<'a> {
...
+        // This is a workaround since evaluating sections of the execution graph that correspond to local variable
+        // expressions does not yield the expected results.
+        let expr = self.get_expr(expr_id);
+        if let ExprKind::Var(Res::Local(local_var_id), _) = &expr.kind {
+            let value = self
+                .eval_context
+                .get_current_scope()
+                .get_classical_local_value(*local_var_id);
+            return Ok(EvalControlFlow::Continue(value.clone()));
+        }
```
idavis, line 406: We should talk about this on Monday.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -370,32 +526,111 @@ impl<'a> PartialEvaluator<'a> {
...
+            return Err(Error::Unexpected(
+                "embedded return in array".to_string(),
+                value_expr.span,
+            ));
+        };
+        let size_control_flow = self.try_eval_expr(size_expr_id)?;
+        let EvalControlFlow::Continue(size) = size_control_flow else {
+            let size_expr = self.get_expr(size_expr_id);
+            return Err(Error::Unexpected(
+                "embedded return in array".to_string(),
```
idavis, line 548:
```suggestion
                "embedded return in array size".to_string(),
```

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1107,12 +1380,117 @@ impl<'a> PartialEvaluator<'a> {
     }
 
     fn try_eval_stmt(&mut self, stmt_id: StmtId) -> Result<EvalControlFlow, Error> {
-        // If the statement is classical, we can just evaluate it.
-        if self.is_classical_stmt(stmt_id) {
-            self.eval_classical_stmt(stmt_id)
+        let stmt = self.get_stmt(stmt_id);
+        match stmt.kind {
+            StmtKind::Expr(expr_id) => self.try_eval_expr(expr_id),
```
idavis, line 1385: Why doesn't handling of `StmtKind::Expr` have to handle control flow adjustments? Please add a comment.

idavis (approved):

compiler/qsc_partial_eval/src/lib.rs
swernli: Is there a test case associated with this? I commented it out and everything still passes, so maybe it's not needed (or there is a test missing)?

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Comment added.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Added some more test cases that without the fix or my initial workaround should fail (now they don't 😄)

swernli (approved):

Issue #1460, opened 2024-05-01
Replace CheckOperationsEquivalence with CheckOperationsAreEqual


tcNickolas:
**Is your feature request related to a problem? Please describe.**
It's confusing that KatasLibrary redefines an operation that is included in M.Q.Diagnostics, and if a person reads into the code, they can  deduce that we don't have this functionality in M.Q.Diagnostics (in fact, I only discovered that we do have it today...)

**Describe the solution you'd like**
* We need to remove `CheckOperationsEquivalence` from KatasLibrary.qs and replace it with `CheckOperationsAreEqual`. 
* We should probably also rename `CheckOperationsEquivalenceStrict` from KatasLibrary.qs to `CheckOperationsAreEqualStrict` so that the naming logic matches.

PR #1481, opened 2024-05-04
Change operation from CheckOperationsEquivalence to CheckOperationsAreEqual

goshua13:
#1460


tcNickolas (changes requested):
I'm afraid this is not the solution we're looking for. Q# standard library defines an operation [CheckOperationsAreEqual](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.diagnostics/checkoperationsareequal) that has the same effect as this `CheckOperationsEquivalence`, so we're looking to remove the code of `CheckOperationsEquivalence` and replace all calls to it from all Verification.qs files with calls to the standard library `CheckOperationsAreEqual`.

The part where you rename `CheckOperationsEquivalenceStrict` to `CheckOperationsAreEqualStrict` looks right, but you'll also need to do this name switch everywhere `CheckOperationsEquivalenceStrict` was used. I kicked the CI so you'll see it failing because it won't be able to find operations the verification is using without rename. (And if you set up the playground build locally, you'll be able to see those failures in more detail.)

Thank you for picking this up!

katas/content/KatasLibrary.qs
```diff
@@ -6,43 +6,21 @@ namespace Microsoft.Quantum.Katas {
...
-        apply {
-            op(target);
-            Adjoint reference(target);
-        }
-
-        let areEquivalent = CheckAllZero(control + target);
-        ResetAll(control + target);
-        areEquivalent
-    }
+    open Microsoft.Quantum.Diagnostics;
```
tcNickolas, line 9: You shouldn't need to re-open this namespace, it's already open in line 6

tcNickolas (changes requested):
This is headed in the right direction, but there are quite a few build failures. 
1. There are a few occurrences of `CheckOperationsEquivalence` itself in the code which need to be updated, we don't always use it wrapped in `CheckOperationsEquivalenceStrict`
2. The order of arguments in `CheckOperationsEquivalence` and `CheckOperationsAreEqual` is different, in the latter the number of qubits is first, so you need to update that too (that's what's causing the errors "expected Int, found (Qubit[] => Unit is f?0)")
3. There might be cases when the M.Q.Diagnostics namespace is not open in the Verification.qs, though I haven't scoured the code for those. In this case, the error would be that `CheckOperationsAreEqual` is not found

Have you tried building this changes in the playground locally? These build issues should be a lot easier to track down locally than using the CI.

Thank you!

katas/content/KatasLibrary.qs
```diff
@@ -6,43 +6,21 @@ namespace Microsoft.Quantum.Katas {
...
 
     /// # Summary
     /// Given two operations, checks whether they act identically (including global phase) for all input states.
     /// This is done through controlled versions of the operations instead of plain ones which convert the global phase
     /// into a relative phase that can be detected.
-    operation CheckOperationsEquivalenceStrict(
+    operation CheckOperationsAreEqualStrict(
```
tcNickolas, line 15: Now that I think about it, we should probably change the order of arguments here as well for the inputSize to go first, to match the library `CheckOperationsAreEqual`. It can be jarring if two similar operations use different order of arguments, even if they technically come from different libraries

katas/content/KatasLibrary.qs
goshua13: I will setup my local environment soon! I haven't done that yet

goshua13: When running my build I am getting this error. have you seen this before? all of my checks are passing. I am also on main.

```
Checking formatting...
All matched files use Prettier code style!
Running the cargo fmt and clippy checks
    Checking qsc_frontend v0.0.0 (/Users/aragon/Projects/qsharp/compiler/qsc_frontend)
error: you should put bare URLs between `<`/`>` or make a proper Markdown link
   --> compiler/qsc_frontend/src/resolve.rs:130:35
    |
130 |     /// it is missed in the list. https://github.com/microsoft/qsharp/issues/897
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

swernli: Hi @goshua13, looks like the most recent Rust version enhanced some clippy lints, and our scripts treat warnings as errors by default to help keep the code warning clean. I'll open a PR with fixes for this soon! In the meantime, you can build locally with `--no-checks` to skip this clippy check step of the build script.

swernli: See #1492

goshua13: got it! thanks

tcNickolas (approved):
Looks great, thank you!

PR #1484, opened 2024-05-06
Add `CodeGenIntrinsic` attribute for simulator overrides

### LINKED ISSUE #1483: Provide a mechanism for defining simulator-only implementations of callables designed to be intrinsic for code generation
Today, a user can provide an implementation (or decomposition) for an operation that they want to simulate and get that same implementation at code generation time (assuming it follows the restrictions for their chosen target) OR they can define the callable as `body intrinsic` and have it generated as an extern that must be supported by the chosen codegen target. If the user chooses the latter, they get their desired codegen behavior but cannot simulate their program, while the former works for simulation but may prevent QIR codegen or include undesired decomposition.

Given that, it has come up across several discussions that it would be convenient to have a mechanism whereby users could denote a callable as intrinsic for the purpose of codegen but still provide an implementation that gets used during simulation. What form this takes and how it is used in the language are up for discussion, though one possible approach could be a specialized attribute that denotes a callable that should be treated this way.

swernli:
Fixes #1483

### DIFF
--- a/compiler/qsc_partial_eval/tests/intrinsics.rs
+++ b/compiler/qsc_partial_eval/tests/intrinsics.rs
@@ -1031,3 +1031,35 @@ fn call_to_length_in_inner_function_succeeds() {
                 Return"#]],
     );
 }
+
+#[test]
+fn call_to_operation_with_codegen_intrinsic_override_should_skip_impl() {
+    let program = get_rir_program(indoc! {"
+        namespace Test {
+            operation Op1() : Unit {
+                body intrinsic;
+            }
+            @CodeGenIntrinsic()
+            operation Op2() : Unit {
+                Op1();
+            }
+            operation Op3() : Unit {
+                Op1();
+            }
+            @EntryPoint()
+            operation Main() : Unit {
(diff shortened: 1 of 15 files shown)


minestarks (approved):

compiler/qsc_partial_eval/tests/intrinsics.rs
```diff
@@ -1054,3 +1054,39 @@ fn call_to_pauli_i_rotation_for_global_phase_is_noop() {
...
+            }
+            @EntryPoint()
+            operation Main() : Unit {
+                Op1();
+                Op2();
+                Op3();
+            }
+        }
+    "});
+
```
cesarzc, line 1080: It would be good to use `assert_callable` like in other tests in this file to verify that the call IDs match with the correct callables.

cesarzc (approved):

compiler/qsc_partial_eval/tests/intrinsics.rs
swernli: Added!

PR #1485, opened 2024-05-06
Partial evaluation support for unary operator expressions

cesarzc:
This change implements support in partial evaluation for unary operator expressions (leading positive, leading negative, bitwise not and logical not).

### DIFF
--- a/compiler/qsc_partial_eval/src/lib.rs
+++ b/compiler/qsc_partial_eval/src/lib.rs
@@ -26,7 +26,7 @@ use qsc_fir::{
         self, BinOp, Block, BlockId, CallableDecl, CallableImpl, ExecGraphNode, Expr, ExprId,
         ExprKind, Global, Ident, LocalVarId, Mutability, PackageId, PackageStore,
         PackageStoreLookup, Pat, PatId, PatKind, Res, SpecDecl, SpecImpl, Stmt, StmtId, StmtKind,
-        StoreBlockId, StoreExprId, StoreItemId, StorePatId, StoreStmtId,
+        StoreBlockId, StoreExprId, StoreItemId, StorePatId, StoreStmtId, UnOp,
     },
     ty::{Prim, Ty},
 };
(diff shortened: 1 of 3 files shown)


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1104,6 +1106,78 @@ impl<'a> PartialEvaluator<'a> {
...
+            ty: rir_variable_type,
+        };
+
+        // Generate the instruction depending on the unary operator.
+        let value_operand = map_eval_value_to_rir_operand(&value);
+        let instruction = match un_op {
+            UnOp::Neg => {
+                let constant = match rir_variable_type {
+                    rir::Ty::Integer => Operand::Literal(Literal::Integer(-1)),
+                    rir::Ty::Double => Operand::Literal(Literal::Double(-1.0)),
```
swernli, line 1153: turns out there is a floating point negation operation in LLVM, though that probably gets inserted by optimization. We should consider whether we want to use `fneg` and add it into RIR or keep this as a multiplication for simplicity. See [fneg](https://llvm.org/docs/LangRef.html#fneg-instruction) in the LLVM docs.

Relatedly, this made me notice that we are always emitting the QIR with text for LLVM integer operations, and technically those need to be the floating point alternatives if we expect these to work properly, ie: `mul` for integer operands and `fmul` for double operands. @idavis FYI, so we can decide if we should support this in QIR codegen or treat it as an error for now.

swernli: Looking again at the tests, we actually aren't testing this functionality right now. The best thing might be treating a negation of a dynamic double as a failure (maybe `Error::Unimplemented`) for now and punting on putting these instructions into the RIR.

swernli (approved):

compiler/qsc_partial_eval/src/lib.rs
cesarzc: I was thinking about supporting doubles in partial evaluation even if they are currently not supported in RIR to QIR. The reason being that it's almost the same amount of work to support them, and we also would rely on a verification pass after partial evaluation is done that would report an error if we produced something that is not supported by the intended target so it wouldn't go to the RIR to QIR stage.

PR #1487, opened 2024-05-07
[Playground][Watch] Reload katas and samples while local development

Manvi-Agrawal:
### Context
Previously, katas and samples in playground didnt change as a result of changes in katas and samples directory. Hence I am following a similar pattern to the rust builds. 

### Verification
Playground gets updated during local development.

@tcNickolas , @minestarks could you please look into this small PR.


billti:
I had this on my TODO list also, so thanks for beating me to it! :-)

billti (approved):

PR #1493, opened 2024-05-08
Ensure variable types are supported by target in partial eval

swernli:
This change adds logic to verify that RIR variables that are created by partial eval are supported by the currently configured target capabilities.


swernli:
We came up with an alternative approach that we prefer, and it overrides the changes in this PR.

PR #1494, opened 2024-05-08
Fix RCA for repeat and while loops

cesarzc:
This change fixes Runtime Capabilities Analysis (RCA) for repeat and while loops.

### DIFF
--- a/compiler/qsc_rca/tests/loops.rs
+++ b/compiler/qsc_rca/tests/loops.rs
@@ -53,7 +53,50 @@ fn check_rca_for_dynamic_for_loop() {
 }
 
 #[test]
-fn check_rca_for_dynamic_repeat_until_loop_with_initial_classical_condition() {
+fn check_rca_for_classical_repeat_until_loop() {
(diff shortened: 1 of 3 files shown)


compiler/qsc_rca/tests/loops.rs
```diff
@@ -51,3 +51,261 @@ fn check_rca_for_dynamic_for_loop() {
...
+    let mut compilation_context = CompilationContext::default();
+    compilation_context.update(
+        r#"
+        use q = Qubit();
+        mutable r = Zero;
+        let loop = repeat { }
+        until r == One
+        fixup {
+            set r = MResetZ(q);
+        };
```
swernli, line 155: nitpick: using different indentation here would make it easier to quickly parse the differences between these test cases.

swernli (approved):

compiler/qsc_rca/tests/loops.rs
cesarzc: Updated.

PR #1495, opened 2024-05-08
Add support for CodeActions in the Language Service

orpuente-MS:
This PR adds support for CodeActions in the Language Service. QuickFixes are one kind of CodeActions available in VS Code.

Below is a demo of QuickFixes for lints:

https://github.com/microsoft/qsharp/assets/156957451/e0c6ba8b-3a67-42af-b9f5-87a755efa2a0


language_service/src/code_action.rs
```diff
@@ -0,0 +1,147 @@
+use miette::Diagnostic;
```
billti, line 1: Add the license header

vscode/src/codeActions.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ILanguageService } from "qsharp-lang";
+import * as vscode from "vscode";
+import { ICodeAction } from "../../npm/qsharp/lib/web/qsc_wasm";
```
billti, line 6: Don't import via a relative path into the npm package. Import from 'qsharp-lang'.

vscode/src/codeActions.ts
```diff
@@ -0,0 +1,73 @@
...
+export function createCodeActionsProvider(languageService: ILanguageService) {
+  return new QSharpCodeActionProvider(languageService);
+}
+
+class QSharpCodeActionProvider implements vscode.CodeActionProvider {
+  constructor(public languageService: ILanguageService) {}
+  async provideCodeActions(
+    document: vscode.TextDocument,
+    range: vscode.Range | vscode.Selection,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
billti, line 18: I haven't checked out the branch to test, but I _think_ you can just remove these two params if not used, rather than suppress the warnings.

vscode/src/common.ts
```diff
@@ -3,6 +3,8 @@
 
 import { TextDocument, Uri, Range, Location } from "vscode";
 import { ILocation, IRange } from "qsharp-lang";
+import { IWorkspaceEdit } from "../../npm/qsharp/lib/web/qsc_wasm";
```
billti, line 6: Again, just import from the 'qsharp-lang' package (i.e. add to the imports in the line above). If you need to re-export more things in that package, then do so.

language_service/src/code_action.rs
orpuente-MS: Added

vscode/src/codeActions.ts
orpuente-MS: Fixed, thanks

vscode/src/codeActions.ts
orpuente-MS: Yes, it worked. Thank you!

language_service/src/code_action.rs
```diff
@@ -0,0 +1,153 @@
...
+    let diagnostics = compilation
+        .errors
+        .iter()
+        .filter(|error| is_error_relevant(error, source_name, range, encoding));
+
+    // An example of what quickfixes could look like if they were generated here.
+    // The other option I considered was generating the quickfixes when the errors
+    // are initially issued. But that has two problems:
+    //  1. It does unnecesary computations at compile time, that would go to waste if using the CLI compiler.
+    //  2. The quickfix logic would be spread across many crates in the compiler.
```
minestarks, line 45: Yep, I think this approach works best. If we ever hit perf issues as the number of available code actions grows, VS Code provides an opportunity to defer the computation of edits even further (using the `resolveCodeAction` request). But I don't see this being an issue anytime soon.

compiler/qsc_data_structures/src/line_column.rs
```diff
@@ -124,6 +136,26 @@ impl Range {
...
+        let start = self.start.max(other.start);
+        let end = self.end.min(other.end);
+
+        if start <= end {
+            Some(Self { start, end })
+        } else {
+            None
+        }
+    }
 }
```
minestarks, line 159: I like these methods, but per my first comment, I think we shouldn't need them. As much as possible we should use compiler `Span`s within the implementations of language service features.

vscode/src/common.ts
```diff
@@ -37,3 +38,17 @@ export function toVscodeRange(range: IRange): Range {
 export function toVscodeLocation(location: ILocation): any {
   return new Location(Uri.parse(location.source), toVscodeRange(location.span));
 }
+
+export function toVscodeWorkspaceEdit(
```
minestarks, line 42: I like this! I'd love it if you went a step further and refactored `rename.ts` to call into this as well when it's constructing `WorkspaceEdit`s.

minestarks (changes requested):
Looks great! Few requested changes around range/span handling.

minestarks:
Forgot to mention, you'll want to wire this up in the Playground using the Monaco editor as well. Its APIs are very similar to VS Code's, so hopefully it shouldn't be too complicated.

This is where you'd want to register it, alongside other language feature providers:

https://github.com/microsoft/qsharp/blob/ec6df4a317344ae672151f8a4485bb0e7f494e96/playground/src/main.tsx#L241


https://microsoft.github.io/monaco-editor/typedoc/interfaces/languages.CodeActionProvider.html

vscode/src/common.ts
orpuente-MS: Done!

language_service/src/code_action.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::{
+    compilation::Compilation,
+    protocol::{CodeAction, CodeActionKind, TextEdit, WorkspaceEdit},
+};
+
+pub(crate) fn get_code_actions(
+    compilation: &Compilation,
+    source_name: &str,
+    range: &Range,
+    encoding: Encoding,
```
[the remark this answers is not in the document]
orpuente-MS: Yes, you were right. That made things a lot easier. Thanks!

compiler/qsc_data_structures/src/line_column.rs
orpuente-MS: Yes, after your suggestion these methods are not necessary anymore inside Range.

Instead, I went ahead and added them as an implementation to Span. I will refactor the code using `utils/span_contains(...)` to use `Span::contains(...)` in a follow up PR, to not bloat this one.

compiler/qsc_data_structures/src/span.rs
```diff
@@ -16,6 +16,28 @@ pub struct Span {
     pub hi: u32,
 }
 
+impl Span {
+    /// Returns true if the position is within the range.
+    #[must_use]
+    pub fn contains(&self, offset: u32) -> bool {
+        (self.lo..self.hi).contains(&offset)
```
minestarks, line 23: Maybe the compiler is smart enough to optimize this away, but I think a comparison (`<` etc) with the `lo` and `hi` offsets is more efficient than comparing against each member of the range.

minestarks (approved):

compiler/qsc_data_structures/src/span.rs
orpuente-MS: Oh, if you do it manually that actually issues the following clippy lint warning: https://rust-lang.github.io/rust-clippy/rust-1.78.0/index.html#/manual_range_contains. That's why started using `Range::contains`.

orpuente-MS: Just to verify I went to Compiler Explorer and they generate the same assembly even with opt-level=1 (basic optimizations). So I think it's fine.

minestarks: Lints like these are why Clippy drives me crazy. "More readable" according to who?! Come on!

But in any case, following convention is good. Compiler optimizations are good. Thanks for checking!

swernli (approved):

PR #1497, opened 2024-05-08
Check unresolved callables during partial eval

swernli:
This change updates the RCA check pass to not treat unresolved callables as errors and instead relies on a new check at partial eval time that checks the resolved callable with the known arguments to return RCA check errors at that point.


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -748,30 +743,121 @@ impl<'a> PartialEvaluator<'a> {
             panic!("global is not a callable");
         };
 
+        // Set up the scope for the call, which allows additional error checking if the callable was
+        // previusly unresolved.
```
cesarzc, line 747:
```suggestion
        // previously unresolved.
```

cesarzc (approved):

idavis (approved):

PR #1499, opened 2024-05-09
Partial evaluation support for binary operations

cesarzc:
This change implements support for binary operations and binary assign operations in partial evaluation.


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -367,6 +372,210 @@ impl<'a> PartialEvaluator<'a> {
...
+        bin_op: BinOp,
+        lhs_bool: bool,
+        rhs_expr_id: ExprId,
+    ) -> Result<EvalControlFlow, Error> {
+        let value = match (bin_op, lhs_bool) {
+            // Handle short-circuiting for logical AND and logical OR.
+            (BinOp::AndL, false) => Value::Bool(false),
+            (BinOp::OrL, true) => Value::Bool(true),
+            // The other possible cases.
+            (BinOp::AndL | BinOp::OrL | BinOp::Eq | BinOp::Neq, _) => {
```
swernli, line 435: I think a similar thing can happen here... if the case is `(BinOp::AndL, true)` then the result is just the rhs, so rather than emitting a comparison it can be a store of rhs into the rir variable. Same for `(BinOp::OrL, false)`, `(BinOp::Eq, true)` and `(BinOp::Neq, false)`.

compiler/qsc_partial_eval/tests/operators.rs
```diff
@@ -300,3 +303,2065 @@ fn logical_not_unary_operator_generates_logical_not_rir_instruction() {
...
+fn comparing_measurement_result_against_result_literal_for_equality_adds_read_result_and_comparison_instructions(
+) {
+    let program = get_rir_program(indoc! {
+        r#"
+        namespace Test {
+            @EntryPoint()
+            operation Main() : Bool {
+                use q = Qubit();
+                let r = MResetZ(q);
+                r == One
```
swernli, line 463: With the suggestion about optimizing some comparisons, this case will become a store, which unfortunately means we'd want a `r == Zero` case to confirm that the right comparison operation emits for that. We'd also need an `r != One` case to complement the one below this.

compiler/qsc_partial_eval/tests/operators.rs
```diff
@@ -300,3 +303,2065 @@ fn logical_not_unary_operator_generates_logical_not_rir_instruction() {
...
+#[test]
+fn integer_div_with_lhs_classical_integer_and_rhs_dynamic_integer() {
+    let program = get_rir_program(indoc! {
+        r#"
+        namespace Test {
+            @EntryPoint()
+            operation Main() : Int {
+                use q = Qubit();
+                let i = MResetZ(q) == Zero ? 0 | 1;
+                1 / i
```
swernli, line 1259: Could you include a test case that will trigger the division by zero error? Something like this:
```qsharp
use q = Qubit();
let z = 0;
let i = MResetZ(q) == Zero ? 0 | 1;
i / z;
```

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1912,6 +2081,223 @@ impl<'a> PartialEvaluator<'a> {
...
+                ConditionCode::Sle,
+                lhs_operand,
+                rhs_operand,
+                bin_op_rir_variable,
+            );
+            (bin_op_rir_variable, bin_op_rir_ins)
+        }
+        _ => panic!("unsupported binary operation for integers: {bin_op:?}"),
+    };
+    Ok((rir_variable, rir_ins))
```
swernli, line 2215: nitpick: this could be a wrapping `Ok()` around the match expression, and then you don't need to let-bind the variables. But that may just be a personal taste thing, so feel free to ignore.

swernli (commented):
Looking good so far!

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -367,6 +372,210 @@ impl<'a> PartialEvaluator<'a> {
...
+            BinOp::Eq => ConditionCode::Eq,
+            BinOp::Neq => ConditionCode::Ne,
+            _ => {
+                return Err(Error::Unexpected(
+                    format!("invalid binary operator for Result operands: {bin_op:?})"),
+                    bin_op_expr_span,
+                ))
+            }
+        };
+        let instruction = Instruction::Icmp(condition_code, lhs_operand, rhs_operand, rir_variable);
```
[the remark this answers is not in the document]
cesarzc: Good point. Updated this logic and the tests to reflect this optimization.

compiler/qsc_partial_eval/tests/operators.rs
cesarzc: Updated the tests and added tests for when both operands are dynamic.

compiler/qsc_partial_eval/tests/operators.rs
cesarzc: Added test case for division by zero.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: I usually prefer it the other way 😅. `match` expressions, especially long ones, binding to a variable before wrapping it in an `Ok()`. I'll leave this one as is 😄

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated :)

compiler/qsc_partial_eval/tests/misc.rs
```diff
@@ -176,3 +179,98 @@ fn unitary_call_within_an_if_with_classical_condition_within_a_repeat_until_loop
...
+            Block 1:Block:
+                Call id(3), args( Variable(0, Integer), Pointer, )
+                Return
+            Block 2:Block:
+                Variable(0, Integer) = Store Integer(5)
+                Jump(1)"#]],
+    );
+}
+
+#[ignore = "WIP"]
```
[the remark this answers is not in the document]
cesarzc: Updated.

PR #1500, opened 2024-05-09
Fix incomplete string escape in circuit.ts

minestarks:
This issue was found by CodeQL (thanks to @idavis for setting this up and raising the alert). `replace` with a string argument only replaces the first occurrence of `\n`.

### DIFF
--- a/vscode/src/circuit.ts
+++ b/vscode/src/circuit.ts
@@ -315,7 +315,7 @@ function errorsToHtml(
 
     const location = documentHtml(document, diag.range);
     const message = escapeHtml(`(${diag.code}) ${diag.message}`).replace(
-      "\n",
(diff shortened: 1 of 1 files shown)


vscode/src/circuit.ts
```diff
@@ -315,7 +315,7 @@ function errorsToHtml(
 
     const location = documentHtml(document, diag.range);
     const message = escapeHtml(`(${diag.code}) ${diag.message}`).replace(
```
billti, line 317: replaceAll has also been common for about 4 years to address this common mistake <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll>

billti (approved):

PR #1502, opened 2024-05-09
Distinguishing Unitaries kata: add first task

tcNickolas:
Testing harness doesn't enforce limitation on the number of unitary calls, but I still pass that parameter to simplify implementing this limitation later

### DIFF
--- a/katas/content/index.json
+++ b/katas/content/index.json
@@ -9,7 +9,6 @@
   "preparing_states",
   "single_qubit_measurements",
   "multi_qubit_measurements",
-  "distinguishing_unitaries",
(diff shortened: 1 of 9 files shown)


cesarzc (approved):

katas/content/index.json
```diff
@@ -9,6 +9,7 @@
   "preparing_states",
   "single_qubit_measurements",
   "multi_qubit_measurements",
+  "distinguishing_unitaries",
```
tcNickolas, line 12: Rookie mistake, forgot to remove from the list...
```suggestion
```

Issue #1470, opened 2024-05-03
Make code examples in /samples/language folder more user-friendly


tcNickolas:
**Is your feature request related to a problem? Please describe.**
Currently the code samples in the /samples/language folder don't produce any output; they evaluate some expressions, but the results are only listed in the comments, not printed for the user to see. This makes any mistakes (such as https://github.com/microsoft/qsharp/pull/1467) hard to spot, and the experience of using them not intuitive.

**Describe the solution you'd like**
We can add some Message commands to print the results of expression evaluation instead of assigning them to variables in most cases (unless, of course, assigning a value to a variable is the point of the example). This will let user compare the results with what they should be easily.

tcNickolas:
Taking Array.qs for example:

* It mentions syntax elements that are not illustrated in the sample, `[0, size = 10]` syntax and Length function, we should add them as well.
* We should print the results of variable assignments: after `let intArray : Int[] = [1, 2, 3, 4]` we should add `Message($"{intArray}");`, after each `let slice = ...` - `Message($"{slice}");`, and so on.

goshua13:
Working on this

PR #1504, opened 2024-05-10
Make Array.qs in /samples/language more user-friendly

goshua13:
#1470

### DIFF
--- a/samples/language/Array.qs
+++ b/samples/language/Array.qs
@@ -23,16 +23,26 @@ namespace MyQuantumApp {
 
         // A basic Int Array literal
         let intArray : Int[] = [1, 2, 3, 4];
+        Message($"{intArray}");
+        Message($"{intArray} of length {Length(intArray)}");
+
         // A basic String Array literal
         let stringArray = ["a", "string", "array"];
+        Message($"{stringArray}");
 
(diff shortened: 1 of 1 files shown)


samples/language/Array.qs
```diff
@@ -23,16 +23,26 @@ namespace MyQuantumApp {
 
         // A basic Int Array literal
         let intArray : Int[] = [1, 2, 3, 4];
+        Message($"{intArray}");
+        Message($"{intArray} of length {Length(intArray)}");
```
Manvi-Agrawal, line 27:
```suggestion
        Message($"Integer Array : {intArray} of length {Length(intArray)}");
```

Nice addition of showing length :-)

[Minor suggestion] We can get rid of the first message. I would recommending adding some text to make it slightly more user friendly :-)

tcNickolas: +1 on printing this array only once.

I wouldn't go all out to add text to every message. The goal of adding printing is to be able to see the values easily and to confirm that they match the comments, but the comments take the brunt of communication about types and language nuances here, so we don't need to duplicate that.

tcNickolas (approved):
Looks great, thank you!

PR #1503, opened 2024-05-10
Use \ket and \bra macros instead of \rangle and \langle

tcNickolas:
Closes #1448. 

* Mostly automatically generated, had to fix a few formulas manually (one had a missing bracket before | which is now a parsing issue)
* The remaining \rangle and \langle are in linear algebra and qubit topics, where they denote inner product rather than bra-ket notation.

PR #1507, opened 2024-05-10
MathJax preview fix and improvements

billti:
This fixes an issue with `\braket` where it was doubling up the entries. The preview today shows:

image

The fix was to replace physics with braket package. (Not sure why, will log a MathJax bug for that also).

The below two items I was playing with a week ago, but including here:

- Made the page render one kata at a time, and supports linking to a kata via the URL and forward/back navigation between katas without a page reload. (As on slow machines putting every kata on the one page was too slow)
- Converted to Preact to match the rest of our code, instead of manipulating DOM elements directly.


billti:
This seems to be showing all of the changes from #1503 for some reason too. Will fix and repush.

billti:
I'll just open a new PR on a new branch. Something's messed up.

PR #1508, opened 2024-05-10
MathJax kata preview fixes

billti:
This fixes an issue with `\braket` where it was doubling up the entries. The preview today shows:

image

The fix was to replace physics with braket package. (Not sure why, will log a MathJax bug for that also).

The below two items I was playing with a week ago, but including here:

- Made the page render one kata at a time, and supports linking to a kata via the URL and forward/back navigation between katas without a page reload. (As on slow machines putting every kata on the one page was too slow)
- Converted to Preact to match the rest of our code, instead of manipulating DOM elements directly.


playground/src/kataViewer.tsx
```diff
@@ -0,0 +1,206 @@
...
+  // The below are added by the MathJax and Highlight.js scripts
+  interface Window {
+    MathJax: any;
+    hljs: any;
+  }
+}
+
+window.MathJax = {
+  loader: {
+    load: ["[tex]/color", "[tex]/braket"],
```
billti, line 27: The change from `physics` to `braket` here is what fixed the braket rendering issue (for some reason).

DmitryVasilevsky: I suggest we add a comment here describing a difference between 'bracket' in physics and 'bracket' in bracket. Also probably worth mentioning this as a guidance on your LaTeX authoring page.

Manvi-Agrawal:
> Made the page render one kata at a time, and supports linking to a kata via the URL and forward/back navigation between katas without a page reload.

Thanks @billti for doing this to improve dev productivity. I was really wishing this feature last week when working on migration of "Superdense Coding" kata because it was slightly annoying to go down and select the kata each time with local changes :)

tcNickolas:
Looks great! Now I can use this preview on my slow machine :-)

playground/public/katas.html
```diff
@@ -4,9 +4,18 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Katas MathJax review</title>
+    <link
```
DmitryVasilevsky, line 7: I think we should have a link to this page from the main playground.

DmitryVasilevsky (approved):

playground/public/katas.html
billti: Maybe later. It's for internal use. We don't want it getting a general audience.

playground/src/kataViewer.tsx
billti: `color` and `braket` auto-load, so are only explicitly loaded here as an optimization. `physics` doesn't auto-load and just should never have been added. (See the list of packages in __bold__ [at this link](https://docs.mathjax.org/en/latest/input/tex/macros/index.html), which are those which auto-load).

I actually figured out the issue by using the MathJax demo page and then seeing that it didn't load __any__ packages. So technically the fix is just use MathJax with the default packages and don't add anything.

minestarks (approved):

PR #1510, opened 2024-05-12
Fix SSA panic due to unpropagated variable mapping

swernli:
This fixes a panic in SSA that results from variable mapping updates not propagating to a block that has a single predecesor. This is corrected by treating that single predecessor block variable map as the update map for the current block during the transformation step.


compiler/qsc_rir/src/passes/ssa_transform.rs
```diff
@@ -48,7 +48,15 @@ pub fn transform_to_ssa(program: &mut Program, preds: &IndexMap<BlockId, Vec<Blo
             .expect("block should have at least one predecessor");
 
         // The block is only a candidate for phi nodes if it has multiple predecessors.
-        if !rest_preds.is_empty() {
+        if rest_preds.is_empty() {
+            // If the block has only one predecessor, track any updates to the variable map from that
+            // predecessory to ensure any phi values that may have been added or inherited in the predecessor
```
swernli, line 53:
```suggestion
            // predecessor to ensure any phi values that may have been added or inherited in the predecessor
```

idavis (approved):

PR #1514, opened 2024-05-13
Fix RIR transform bugs

swernli:
This fixes multiple bugs in RIR transformations:

 - Updates variable remapping logic during SSA pass to check whether the proposed replacement itself has a pending remap, fixing scenario where mutable variables are mapped to other mutable variables.
 - Includes variables created by `Store` instructions in the count for identifying highest variable id to use for inserted phi-nodes, fixing issues where a phi variable may overlap with an existing remapped mutable variable.
 - Avoids wrapping subtraction in qubit remapping mass when there are no qubits in the program.
 - Fixes type check pass to validate that comparison instruction produces a Boolean value rather than the same value as arguments.

This change adds some new test cases and expands test coverage significatly by running all programs produced in partial eval tests through RIR transforms in the test utility.

### DIFF
--- a/compiler/qsc/src/codegen/tests.rs
+++ b/compiler/qsc/src/codegen/tests.rs
@@ -625,8 +625,8 @@ mod adaptive_ri_profile {
             block_2:
               br label %block_3
             block_3:
-              %var_2 = phi i64 [0, %block_1], [1, %block_2]
-              call void @__quantum__rt__integer_record_output(i64 %var_2, i8* null)
+              %var_3 = phi i64 [0, %block_1], [1, %block_2]
+              call void @__quantum__rt__integer_record_output(i64 %var_3, i8* null)
               ret void
             }
 
--- a/compiler/qsc_rir/src/passes/reindex_qubits.rs
+++ b/compiler/qsc_rir/src/passes/reindex_qubits.rs
@@ -4,7 +4,7 @@
 #[cfg(test)]
 mod tests;
 
-use std::collections::hash_map::Entry;
+use std::{collections::hash_map::Entry, ops::Sub};
 
 use qsc_data_structures::index_map::IndexMap;
 use rustc_hash::FxHashMap;
@@ -47,7 +47,7 @@ pub fn reindex_qubits(program: &mut Program) {
         used_cx,
         cx_id,
         mresetz_id,
-        highest_used_id: program.num_qubits - 1,
+        highest_used_id: program.num_qubits.max(1).sub(1),
     };
 
     let pred_map = build_predecessors_map(program);
--- a/compiler/qsc_rir/src/passes/reindex_qubits.rs
+++ b/compiler/qsc_rir/src/passes/reindex_qubits.rs
@@ -47,6 +47,10 @@ pub fn reindex_qubits(program: &mut Program) {
         used_cx,
         cx_id,
         mresetz_id,
+        // For this calculation, qubit IDs can never be lower than zero but a program may not use
+        // any qubits. Since `highest_used_id` is only needed for remapping pass and a program without any
+        // qubits won't do any remapping, it's safe to treat this as 1 and let `highest_used_id` default
+        // to zero.
         highest_used_id: program.num_qubits.max(1).sub(1),
     };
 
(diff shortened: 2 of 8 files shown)


cesarzc (commented):
nit: adding a small comment here about this being needed for cases in which no qubits are used would be useful because it isn't evident just by looking at the use of `max`.

compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -47,7 +47,7 @@ pub fn reindex_qubits(program: &mut Program) {
         used_cx,
         cx_id,
         mresetz_id,
-        highest_used_id: program.num_qubits - 1,
+        highest_used_id: program.num_qubits.max(1).sub(1),
```
cesarzc, line 50: nit: adding a small comment here about this being needed for cases in which no qubits are used would be useful because it isn't evident just by looking at the use of `max`.

swernli: Added comment here to clarify, thanks!

compiler/qsc_rir/src/passes/ssa_transform.rs
```diff
@@ -239,17 +239,25 @@ impl VariableMapper for FxHashMap<VariableId, Operand> {
...
     }
 
     fn to_variable(&self, var: Variable) -> Variable {
-        self.get(&var.variable_id)
-            .copied()
-            .map_or(var, |operand| match operand {
-                Operand::Literal(_) => panic!("literal not supported in this context"),
-                Operand::Variable(var) => var,
-            })
+        let mut var = var;
```
cesarzc, line 254: nit: this looks like a `get_originating_variable` method or `get_original_variable` method. The name `to_variable` is a bit confusing because it implies a conversion, but the method is not doing that exactly.

cesarzc: It seems like the previous implementation was a more straight-forward conversion, but that is no longer the case.

cesarzc (approved):

compiler/qsc_rir/src/passes/ssa_transform.rs
swernli: That's a good point... I think this can be cleaned up with a different pattern.

compiler/qsc/src/codegen/tests.rs
```diff
@@ -625,8 +625,8 @@ mod adaptive_ri_profile {
             block_2:
               br label %block_3
             block_3:
-              %var_2 = phi i64 [0, %block_1], [1, %block_2]
-              call void @__quantum__rt__integer_record_output(i64 %var_2, i8* null)
+              %var_3 = phi i64 [0, %block_1], [1, %block_2]
```
idavis, line 628: why isn't this `%var_2`?

compiler/qsc_rir/src/passes/ssa_transform.rs
swernli: I simplified this to be more like the `impl` on `Operand` above, and I think it's more clear that it's a mapping that is occurring using the map passed as argument.

compiler/qsc/src/codegen/tests.rs
swernli: Part of the bug fix is to start the index for inserted phi node variables at the highest variable including stores, rather than without including stores. I think this is needed to play nice with the fix for properly mapping mutable variables that get stored to other mutable variables, but there is one other approach I could try to see if it maintains both the reuse logic and fix for mapping.

swernli: Yeah, trying to reuse variable IDs during mapping for phi nodes in SSA creates a cycle in the variable mapping that will cause this test case to hang. The bug is meant to fix scenarios where a store instruction uses a variable that itself was created via a store instruction (see the test case `ssa_transform_maps_store_instrs_that_use_values_from_other_store_instrs` in the ssa_transform/tests.rs file). This was the case that motivated the update that checks if the variable we are mapping to has it's own mapping and repeating that search until we find the actual source value. But if variable IDs are reused, it will overlap with other mappings that may not have been applied yet. I can't remove the reused variable from the map otherwise it will leave behind orphaned variables, but I can't skip putting variables created for phi into the map either because then those uses won't be updated to use the phi created value. So I changed the logic to not reuse variable IDs, which leaves the gap `var_0`, `var_1`, `var_3`.

If this becomes an issue for understanding the QIR later we could add another pass that tries to consolidate variable IDs, but since these are named `%var_0` instead of unnamed `%0` variables, LLVM doesn't care what they are called and does not force them to be sequential.

idavis (approved):

PR #1517, opened 2024-05-13
Fix integer output recording name

cesarzc:


swernli (approved):
Looks like there are still more tests to update, but the production change looks accurate. Signing off assuming test updates are coming!

PR #1519, opened 2024-05-14
Fix for logic based on `CallToUnresolvedCallee` flag

swernli:
The check for the dynamism of the return from unresolved callees incorrectly ended up applying to calls where the flag had propagated from another call site, including the entry point callable itself. This change introduces new infra to track the call expression that introduced the flag as opposed to all calls the flag had propagated to, which keeps the desired protection but avoids rejecting other call expressions.


swernli:
FYI: I confirmed the new test case fails on main but passes with the fix, so we know it is testing the specific flag propagation the change is intended to side-step.

PR #1526, opened 2024-05-15
Additional RIR block remapping, Error string updates

swernli:
This change fixes another issue in block id remapping in RIR that can lead to degenerate performance cases. The corresponding test case is updated to exercise deeper nesting and verify the fix.

This also updates the error string used for dynamic values encountered by partial evaluation based on feedback.

### DIFF
--- a/compiler/qsc_rir/src/passes/remap_block_ids.rs
+++ b/compiler/qsc_rir/src/passes/remap_block_ids.rs
@@ -37,7 +37,7 @@ pub fn remap_block_ids(program: &mut Program) {
         // This effectively remaps all the blocks in the list and updates the mapped id of the current block.
         // This is only safe without cycles, so on a cyclic graph the node is skipped and not remapped.
         if is_acyclic {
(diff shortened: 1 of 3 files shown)


compiler/qsc_rir/src/passes/remap_block_ids.rs
```diff
@@ -54,6 +54,7 @@ pub fn remap_block_ids(program: &mut Program) {
             // the same blocks back-to-back.
             continue;
         }
+        blocks_to_visit.retain(|id| !successors.contains(id));
```
cesarzc, line 57: A comment here would be useful. Something like: "Since we are going to extend the blocks to visit using the successors of the block we are currently visiting, remove the successor blocks already in the blocks to visit.".

cesarzc (approved):

compiler/qsc_rir/src/passes/remap_block_ids.rs
swernli: Added comment!

PR #1528, opened 2024-05-15
Fix debugger scope handling during early returns

swernli:
Changes introduced in #1442 to enable debugger tracking of variable scopes caused a bug if a program used callables with early returns, namely those returns would not pop any added scopes and later execution would read/write the wrong variable states. This fixes the issue by introducing a debug-only return node that triggers a frame leave handling in the environment. This needs to be debug only because it should not be used during partial evaluation.

### DIFF
--- a/compiler/qsc/src/interpret/tests.rs
+++ b/compiler/qsc/src/interpret/tests.rs
@@ -1581,6 +1581,99 @@ mod given_interpreter {
         }
 
         #[test]
+        fn debugger_simple_execution_succeeds() {
+            let source = indoc! { r#"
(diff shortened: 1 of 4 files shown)


swernli:
Regarding testing, I'm thinking it would make sense to add some targeted unit tests to the debugger tests in this PR and then add execution of all the samples to integration tests. We'll have to decide how to handle the resource estimation samples so the test execution is not too long from trying to simulate a scalable algorithm, but I don't want to delay this PR while we work that out.

idavis (approved):

PR #1530, opened 2024-05-15
Superdense coding sample works in Adaptive Profile

DmitryVasilevsky:
Changed Superdense coding example to work with adaptive profile
* Use quantum random bit generator instead of one provided in simulator
* Removed unnecessary check that wasn't compatible with the profile
* Changed return value so that correctness of the algorithm can be verified.
Now it generates QIR in adaptive profile.

CHECK: Not sure if this return type from Main is supported. Need to check.

### DIFF
--- a/samples/algorithms/SuperdenseCoding.qs
+++ b/samples/algorithms/SuperdenseCoding.qs
@@ -9,10 +9,6 @@
 /// This Q# program implements superdense coding to send two classical bits of
 /// information.
 namespace Sample {
-    open Microsoft.Quantum.Measurement;
-    open Microsoft.Quantum.Diagnostics;
-    open Microsoft.Quantum.Random;
-
     @EntryPoint()
     operation Main() : ((Bool, Bool), (Bool, Bool)) {
         use (aliceQubit, bobQubit) = (Qubit(), Qubit());
@@ -37,7 +33,7 @@ namespace Sample {
     operation DrawRandomBit() : Bool {
         use q = Qubit();
         H(q);
-        MResetZ(q) == One
+        return MResetZ(q) == One;
     }
 
     // Prepares an entangled state: 1/sqrt(2)(|00〉 + |11〉)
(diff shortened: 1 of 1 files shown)


swernli (approved):

samples/algorithms/SuperdenseCoding.qs
```diff
@@ -9,37 +9,39 @@
 /// This Q# program implements superdense coding to send two classical bits of
 /// information.
 namespace Sample {
+    open Microsoft.Quantum.Measurement;
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Random;
```
billti, line 14: I just removed all 3 namespaces locally and it still worked fine. I believe Measurement is automatically included, and there are no Diagnostics or Random calls with these changes.

samples/algorithms/SuperdenseCoding.qs
```diff
@@ -9,37 +9,39 @@
...
         ResetAll([aliceQubit, bobQubit]);
-        return (decodedBit1, decodedBit2);
+        return ((sourceBit1, sourceBit2), (decodedBit1, decodedBit2));
+    }
+
+    // Generates random bit using an auxilliary qubit
+    operation DrawRandomBit() : Bool {
+        use q = Qubit();
+        H(q);
+        MResetZ(q) == One
```
cesarzc, line 40:
```suggestion
        return MResetZ(q) == One;
```

cesarzc: nit: our samples do not generally use implicit returns (at least currently).

swernli: aww, I liked having the implicit return in there. But yes, you are right that it's not what we've done so far. When would be the right time to start using that language feature?

billti (approved):

samples/algorithms/SuperdenseCoding.qs
DmitryVasilevsky: Added return.

PR #1532, opened 2024-05-16
GHZ and CAT samples work in Adaptive and Base Profiles

DmitryVasilevsky:
This updates GHZ state and CAT state sample to work in base profile and adaptive profile.
- Removed check for zero state
- updated comments
- Used library functions MResetEachZ() and Rest()
- Made preparation function specific to GHZ in GHZ sample

### DIFF
--- a/samples/algorithms/CatState.qs
+++ b/samples/algorithms/CatState.qs
@@ -7,9 +7,7 @@
 ///
 /// This Q# program implements a cat state of 5 qubits.
 namespace Sample {
-    open Microsoft.Quantum.Arrays;
     open Microsoft.Quantum.Diagnostics;
-    open Microsoft.Quantum.Measurement;
 
     @EntryPoint()
     operation Main() : Result[] {
@@ -35,6 +33,6 @@ namespace Sample {
         // Set the first qubit in the register into a (|0〉 + |1〉) / √2 superposition.
         // Then apply a CNOT to the remaining qubits using the first qubit as control.
         H(register[0]);
-        ApplyToEach(CNOT(register[0], _), Rest(register));
+        ApplyToEach(CNOT(register[0], _), register[1...]);
     }
 }
--- a/samples/algorithms/CatState.qs
+++ b/samples/algorithms/CatState.qs
@@ -7,6 +7,7 @@
 ///
 /// This Q# program implements a cat state of 5 qubits.
 namespace Sample {
+    open Microsoft.Quantum.Arrays;
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Measurement;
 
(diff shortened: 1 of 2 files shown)


samples/algorithms/CatState.qs
```diff
@@ -21,20 +22,19 @@ namespace Sample {
...
     operation PrepareCatState(register : Qubit[]) : Unit {
         Fact(Length(register) > 0, "Qubit register must not be empty.");
-        Fact(CheckAllZero(register), "Qubits are not in the |0〉 state.");
 
-        // Set the first qubit in the register into a 1/sqrt(2)(|0〉 + |1〉) superposition.
+        // Set the first qubit in the register into a (|0〉 + |1〉) / √2 superposition.
         // Then apply a CNOT to the remaining qubits using the first qubit as control.
         H(register[0]);
-        ApplyToEach(CNOT(register[0], _), register[1...]);
+        ApplyToEach(CNOT(register[0], _), Rest(register));
```
swernli, line 38: What was wrong with the `[1...]` syntax? That seems to work fine for generating QIR and avoids the need to open the arrays namespace.

samples/algorithms/GHZ.qs
```diff
@@ -30,23 +31,18 @@ namespace Sample {
...
     operation PrepareGHZState(qs : Qubit[]) : Unit {
-        Fact(Length(qs) > 0, "`qs` length must be greater than zero");
-        Fact(CheckAllZero(qs), "All qubits must be in the |0〉 state");
+        Fact(Length(qs) == 3, "`qs` length be 3.");
 
         H(qs[0]);
-        ApplyToEach(CNOT(qs[0], _), qs[1...]);
+        CNOT(qs[0], qs[1]);
+        CNOT(qs[0], qs[2]);
     }
```
swernli, line 47: Rather than make this specific to the length 3, this could remain generalized and avoid `ApplyToEach` like this:
```qsharp
    operation PrepareGHZState(qs : Qubit[]) : Unit {
        Fact(Length(qs) >= 2, "`qs` length should be at least 2.");

        H(qs[0]);
        for q in qs[1...] {
            CNOT(qs[0], q);
        }
    }
```

samples/algorithms/GHZ.qs
```diff
@@ -18,6 +18,7 @@
 namespace Sample {
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Measurement;
+    open Microsoft.Quantum.Arrays;
```
swernli, line 21: I think this was left behind from an earlier draft, it doesn't seem to be used now.

samples/algorithms/GHZ.qs
cesarzc: I think it's ok to leave it in 3 since GHZ is specifically a three-qubit state. The more general case is covered by the cat state sample.

samples/algorithms/GHZ.qs
```diff
@@ -30,23 +31,18 @@ namespace Sample {
...
-    /// ## qs
-    /// The given register of qubits to be transformed into the GHZ state. It is assumed
-    /// that these qubits are in their default |0〉 state.
+    /// Prepares state (|000〉 + |111〉) / √2 (GHZ state) across a register
+    /// of three qubits `qs`.
+    /// All qubits are assumed to be in |0〉 state on input.
     operation PrepareGHZState(qs : Qubit[]) : Unit {
-        Fact(Length(qs) > 0, "`qs` length must be greater than zero");
-        Fact(CheckAllZero(qs), "All qubits must be in the |0〉 state");
+        Fact(Length(qs) == 3, "`qs` length be 3.");
```
cesarzc, line 42:
```suggestion
Fact(Length(qs) == 3, "`qs` length should be 3.");
```

samples/algorithms/CatState.qs
DmitryVasilevsky: Yes, I like [1...] and I changed the code back.
Funny, just a few months ago I was advocating the same thing - if we have a concise way of expressing something in a language we should prefer it over the library function. I'm still advocating it now. It's just I was looking at library code, library documentation and the set of function a lot recently that it got into my brain. :)

samples/algorithms/GHZ.qs
DmitryVasilevsky: Yes, just like Cesar mentioned I prefer to have specific implementation for 3 qubits here in GHZ sample because Cat sample has general implementation.

samples/algorithms/GHZ.qs
DmitryVasilevsky: Removed unnecessary open statements here and in Cat sample. Do we have lints yet?

samples/algorithms/GHZ.qs
DmitryVasilevsky: A word dropped out on the way from my brain to keyboard...

swernli (approved):

PR #1533, opened 2024-05-16
Test algorithm samples with and without debugging

swernli:
This change introduces a scenario test under qsc that verifies the behavior of the algorithm samples with normal and debug execution. It uses fixed seeds to ensure results are predictable and restricts the slowest test (Shors.qs) to be release only.

Included is some cleanup of output from the samples so that compatible characters are used and no trailing spaces are present.


minestarks:
I don't understand how there can be such a dramatic difference in time taken to run all samples twice vs the existing "build all samples" logic.

running all samples twice: 1.88s
image

existing "Building qsharp samples" step: 109.929s
image

If there's some black magic that makes the first one run so fast, maybe we can remove the existing step?!?

swernli:
> I don't understand how there can be such a dramatic difference in time taken to run all samples twice vs the existing "build all samples" logic.
> ...
> If there's some black magic that makes the first one run so fast, maybe we can remove the existing step?!?

Ha, yeah, it's quite dramatic, isn't it? I'll go into more detail below for posterity, but the short answer is that it does things in memory, in parallel, rather than sequentially with disk reads. That said, I think the other test is still worthwhile (and maybe can be moved to integration tests?), in part because it builds all samples including language samples rather than just the algorithm samples. As soon as a new qs file or qsharp.json project is added to the samples folders, it will get picked up automatically by this part of the build. If after this PR (and ongoing work to add verification output to the language samples) we want to pull those into a test pattern like this, then we can revisit the question of whether the older style is still needed. Longer explanation below...

When compiling these as unit tests, the test files are brought in using the `include_str!` macro, which tells Rust to include the contents of that file at compile time so that the running test can read the contents from the data section of the process memory rather than needing to load it from file. Rust also handily keeps track of any files pulled in this way, so that when the file is changed on disk it triggers the relevant rebuilds even if there are no other changes to the code. And during runtime, the tests are run in parallel, so can take advantage of as many cores as are available on the machine. Compare this to how the tests run using qsc in the build script: the folders are searched on disk, each file path is passed as an argument to a separate process invocation, which needs to read the file from disk in order to compile it. All that sequential process start and disk access overwhelms the actual runtime of the compiler. The tradeoff is needing to add each sample we want to test individually as test case rather than have it be auto-detected from disk. In my opinion, the scale of the perf difference makes that additional maintenance burden worthwhile.

minestarks:
Can we maybe add a build script to iterate through the samples and include _all_ of them in the Rust tests? The fact that each sample is hardcoded sadly takes away from the usefulness of these tests... I can guarantee that the next person to add a sample will _not_ think of adding them to this list :/

swernli:
> Can we maybe add a build script to iterate through the samples and include _all_ of them in the Rust tests? The fact that each sample is hardcoded sadly takes away from the usefulness of these tests... I can guarantee that the next person to add a sample will _not_ think of adding them to this list :/

Not while still using `expect![]` macros to verify the output inline, no. Each one really does need to be individually added. If we switched to a mode where the output we compare to was stored in files, then maybe... but it would definitely get more complicated. The samples testing is kind of a twist on the old "iron triangle" adage where in this case the trade-off is test speed vs test simplicity vs test automation, and we only get to pick two.

minestarks:
We _could_ have one big giant `expect![]` that verifies the result of dynamically concatenated program outputs, right? 

It's not obviously clear to me which way's better at the moment, you're right, each approach has its drawbacks - I'll let it sit in the back of my mind for a bit.

swernli:
I have a couple crazy ideas on how we could add a bit more automation to this, but I need to experiment with them more before I know how bad/good the ideas are...

samples_test/Cargo.toml
minestarks, line 1: Any reason `samples_test` is a  top level folder and not a subfolder of `samples` (e.g. `samples/tests`)?

swernli: I debated on that one. My thinking was that right now the samples folder is pretty cleanly for folks looking to find Q# samples. We link to it from other places, like QCOM. If there were Rust tests mixed in, especially slightly arcane ones like these that are generated from a build script, it makes the folder confusing for newcomers.

On a related note, I also debated about keeping these as a test under "compiler/qsc" where I had originally put them. But it occurred to me that with the way the build script is set up to watch the samples folder, any change to samples would cause qsc to be rebuilt, which is not ideal. This way there is a clearer division: samples_test depends on samples and qsc, so changes to either of those will rebuild samples_test, but qsc won't depend on the samples folder.

tcNickolas:
We'll also need a user-facing folder samples/testing when we get to adding
samples of how to test Q# code. So, without looking at the code, I'm in
favor of keeping Rust test code that does not target users outside of the
samples folder so as not to confuse these two.

samples_test/Cargo.toml
minestarks: All the reasoning is sound, it's just that if the conclusion is a new top-level folder... I'm worried @billti 's eye might start twitching :D

minestarks: Maybe we need that top-level `cargo` folder after all to corral everything under.

swernli: I assume @billti will rearrange the folders at some point, yeah ;)

billti: I did npm. @minestarks will do the Python packages and you'll take care of the Rust ones is my recollection. No rush though, next week is fine ;)

samples_test/build.rs
```diff
@@ -0,0 +1,62 @@
...
+            .file_name()
+            .expect("file name should be separable")
+            .to_str()
+            .expect("file name should be valid");
+        let file_stem = path
+            .file_stem()
+            .expect("file name should be separable")
+            .to_str()
+            .expect("file name should be valid");
+        let file_stem_upper = file_stem.to_uppercase();
```
minestarks, line 32: Can we check for spaces in the file name? Sounds dumb, but the first thing I tried is to copy-paste `BellState.qs` and ended up with `BellState copy.qs` in my samples folder. Might be confusing when you're just playing around with samples, unaware that it's part of the build process.

image

samples_test/build.rs
```diff
@@ -0,0 +1,62 @@
...
+                    vec![("{file_name}".into(), include_str!("../../../../../samples/algorithms/{file_name}").into())],
+                    None,
+                )
+            }}
+
+            #[allow(non_snake_case)]
+            #[test]
+            fn run_{file_stem}() {{
+                let output = compile_and_run({file_stem}_src());
+                {file_stem_upper}_EXPECT.assert_eq(&output);
```
minestarks, line 49: Consider writing a comment here like:

```suggestion
                // This constant must be defined in `samples_test/src/tests.rs` and
                // must contain the output of the sample {file_name}
                {file_stem_upper}_EXPECT.assert_eq(&output);
```

I ended up in this line when I simply added another sample, which generated this test error: 

image

Having a comment telling me exactly how to continue would be so helpful!

minestarks (approved):
Great! Now maybe we can add the language samples too and get rid of the step in JS? ;)

samples_test/build.rs
swernli: Good idea, added!

samples_test/build.rs
swernli: Added an assert, so now it fails early with the offending file name:
image

cesarzc (approved):

swernli:
> Great! Now maybe we can add the language samples too and get rid of the step in Python? ;)

Finally put the ideas this question triggered into a PR: https://github.com/microsoft/qsharp/pull/1739. Is the shift away from Python, qsc-based tests worth the build.rs complexity? We can let the team discuss. But at least we know it's possible!

PR #1534, opened 2024-05-16
Three qubit repetition code sample works in Adaptive Profile

DmitryVasilevsky:
Three Qubit Repetition Code sample suitable for adaptive profile.
Based on https://github.com/microsoft/Quantum/blob/main/samples/azure-quantum/three-qubit-repetition-code/ThreeQubitRepetitionCode.qs

### DIFF
--- a/samples/algorithms/ThreeQubitRepetitionCode.qs
+++ b/samples/algorithms/ThreeQubitRepetitionCode.qs
@@ -62,30 +62,21 @@ namespace Microsoft.Quantum.Samples {
     }
 
     operation RevertBitFlip(register : Qubit[], parity01 : Result, parity12 : Result) : Bool {
-        if parity01 == One {
-            if parity12 == Zero {
-                X(register[0]);
-            }
-        }
+        mutable result = true;
         if parity01 == One {
             if parity12 == One {
                 X(register[1]);
+            } else {
+                X(register[0]);
             }
-        }
-        if parity01 == Zero {
+        } else {
             if parity12 == One {
                 X(register[2]);
-            }
-        }
-
-        mutable result = true;
-        if parity01 == Zero {
-            if parity12 == Zero {
+            } else {
                 set result = false;
             }
         }
-
-        result
+        return result;
     }
 
     operation Encode(register : Qubit[]) : Unit is Adj {
--- a/samples/algorithms/ThreeQubitRepetitionCode.qs
+++ b/samples/algorithms/ThreeQubitRepetitionCode.qs
@@ -5,13 +5,12 @@ namespace Microsoft.Quantum.Samples {
 
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Math;
-    open Microsoft.Quantum.Measurement;
+    open Microsoft.Quantum.Diagnostics;
 
     @EntryPoint()
     operation ThreeQubitRepetitionCode() : (Bool, Int) {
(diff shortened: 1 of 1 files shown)


samples/algorithms/ThreeQubitRepetitionCode.qs
```diff
@@ -0,0 +1,108 @@
...
+        let theta = PI() * 0.5;
+        for i in 1..4 {
+            for qubit in register {
+                Rx(theta, qubit);
+            }
+        }
+    }
+
+    operation RevertBitFlip(register : Qubit[], parity01 : Result, parity12 : Result) : Bool {
+        if parity01 == One {
```
swernli, line 66: This patten of branching seems to be done to avoid the unimplemented LHS dynamic Boolean And support. Rather than tailor the sample to our current intermediate state and then need to remember to come back and clean it up later, we should write the sample the way we intend for it to work and let the generation features catch up. So I'd suggest choosing one of these options:

- Write the if-expressions using And between the conditions that are currently nested, even if that doesn't work right now (it will soon). Worth noting, it won't be as efficient in the QIR produced, but should still work. 
- Convert the two results into a single integer and branch off of that instead.
- Nest using if-else on the first result and have the second parity check inside those branches. This might be the most efficient in terms of QIR instructions produced.

samples/algorithms/ThreeQubitRepetitionCode.qs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    operation Encode(register : Qubit[]) : Unit is Adj {
+        CNOT(register[0], register[1]);
+        CNOT(register[0], register[2]);
+    }
+
+    operation MeasureBitFlipSyndrome(encodedRegister : Qubit[], auxiliaryRegister : Qubit[]) : (Result, Result) {
+        // Measure the bit flip syndrome by checking the parities between qubits 0 and 1, and between qubits 1 and 2.
+        ResetAll(auxiliaryRegister);
```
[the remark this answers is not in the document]
DmitryVasilevsky: Removed reset and made qubit allocation local.

samples/algorithms/ThreeQubitRepetitionCode.qs
DmitryVasilevsky: I did minimal changes from the original sample just to satisfy the compiler, but I actually like your last suggestion! It makes code cleaner. I updated the code.

samples/algorithms/ThreeQubitRepetitionCode.qs
```diff
@@ -0,0 +1,111 @@
...
+                if (bitFlipReverted) {
+                    set bitFlipCount += 1;
+                }
+            }
+        }
+
+        // Transform the qubit to the |1〉 state and measure it in the computational basis.
+        H(encodedRegister[0]);
+        let result = MResetZ(encodedRegister[0]) == One;
+        ResetAll(encodedRegister);
```
swernli, line 44: Nitpick: not really important since it's right at the end of execution, but the first qubit here was already reset by the line above, so this could be:
```suggestion
        ResetAll(encodedRegister[1...]);
```

swernli (approved):
I like the updated branching strategy!

samples/algorithms/ThreeQubitRepetitionCode.qs
DmitryVasilevsky: Updated and added comment.

PR #1535, opened 2024-05-16
Bump version to 1.5 and remove Fact in HiddenShift

billti:


idavis (approved):

samples/algorithms/HiddenShift.qs
```diff
@@ -34,14 +34,11 @@ namespace Sample {
                 nQubits
             );
             let hiddenShift = ResultArrayAsInt(hiddenShiftBitString);
-            Fact(
```
DmitryVasilevsky, line 37: I think we should add a comment to indicate what is expected.

swernli (approved):

DmitryVasilevsky (approved):

samples/algorithms/HiddenShift.qs
billti: I did.

PR #1538, opened 2024-05-16
Fixing LaTeX in the doc comment for Measure and Exp

DmitryVasilevsky:
I updated doc comment for Measure and Exp function.
* Moved text from Remarks to Description as it is specific to the function.
* Simplified formula for probability of getting `Zero` as the outcome of measurement.
* Fixed LaTeX.

### DIFF
--- a/library/std/intrinsic.qs
+++ b/library/std/intrinsic.qs
@@ -258,6 +258,17 @@ namespace Microsoft.Quantum.Intrinsic {
     /// Performs a joint measurement of one or more qubits in the
     /// specified Pauli bases.
     ///
+    /// # Description
+    /// The probability of getting `Zero` is
+    /// $\langle \psi | \frac{I + P_0 \otimes \ldots \otimes P_{N-1}}{2} | \psi \rangle$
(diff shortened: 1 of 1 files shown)


library/std/intrinsic.qs
```diff
@@ -258,6 +258,17 @@ namespace Microsoft.Quantum.Intrinsic {
     /// Performs a joint measurement of one or more qubits in the
     /// specified Pauli bases.
     ///
+    /// # Description
+    /// The probability of getting `Zero` is
+    /// $\langle \psi | \frac{I + P_0 \otimes \ldots \otimes P_{N-1}}{2} | \psi \rangle$
```
billti, line 263: Can we use `\bra` and `\ket` rather than pipes and `\langle` and `\rangle`? We already changed the katas to be consistent here.

DmitryVasilevsky: Was trying to remember what we were supposed to use and failed. Fixed now.

swernli (approved):

PR #1540, opened 2024-05-17
Partial evaluation support for hybrid Boolean operations with a dynamic LHS expression

cesarzc:
This change implements partial evaluation support for hybrid Boolean operations with a dynamic LHS expression, including short-circuiting and optimization logic.

### DIFF
--- a/compiler/qsc_partial_eval/src/lib.rs
+++ b/compiler/qsc_partial_eval/src/lib.rs
@@ -577,84 +577,109 @@ impl<'a> PartialEvaluator<'a> {
             BinOp::AndL => {
                 // ...
                 let lhs_rir_var = map_eval_var_to_rir_var(lhs_eval_var);
-
-                // ... continuation block.
-                let current_block_node = self.eval_context.pop_block_node();
-                let continuation_block_id = self.create_program_block();
-                let continuation_block_node = BlockNode {
-                    id: continuation_block_id,
-                    successor: current_block_node.successor,
-                };
-                self.eval_context.push_block_node(continuation_block_node);
-
-                // ... variable.
-                let variable_id = self.resource_manager.next_var();
-                let rir_variable = rir::Variable {
-                    variable_id,
-                    ty: rir::Ty::Boolean,
-                };
-
-                // ... true block.
-                let true_block_id = self.create_program_block();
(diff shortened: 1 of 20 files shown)


compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -525,6 +525,159 @@ impl<'a> PartialEvaluator<'a> {
...
+            }
+            (_, _) => panic!("invalid Boolean comparison operator {bin_op:?}"),
+        };
+        Ok(result_var)
+    }
+
+    fn eval_logical_bool_bin_op(
+        &mut self,
+        default_result: bool,
+        rhs_eval_condition_var: rir::Variable,
```
cesarzc, line 616: Consider changing the name to something that conveys that this comes from the LHS of the expression.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -525,6 +525,159 @@ impl<'a> PartialEvaluator<'a> {
...
+                // the branch that is evaluated when LHS is false. We cannot use LHS directly as the condition because
+                // the condition to go into the evaluating branch is 'LHS == false', so we need a comparison instruction
+                // whose bound variable will be used as the condition to go into the evaluating branch.
+                let lhs_rir_var = map_eval_var_to_rir_var(lhs_eval_var);
+                let rhs_eval_condition_var_id = self.resource_manager.next_var();
+                let rhs_eval_condition_rir_var = rir::Variable {
+                    variable_id: rhs_eval_condition_var_id,
+                    ty: rir::Ty::Boolean,
+                };
+                let rhs_eval_condition_ins = Instruction::Icmp(
```
cesarzc, line 555: We could not generate this instruction.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Not needed anymore.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Updated with optimization for ORs.

swernli (approved):

PR #1547, opened 2024-05-21
Use package spans in partial evaluation

cesarzc:


swernli (approved):

compiler/qsc_partial_eval/tests/calls.rs
```diff
@@ -1285,7 +1285,7 @@ fn call_to_unresolved_callee_producing_dynamic_value_fails() {
 
     assert_error(
         &error,
-        &expect!["UnexpectedDynamicValue(Span { lo: 298, hi: 305 })"],
+        &expect!["UnexpectedDynamicValue(PackageSpan { package: PackageId(2), span: Span { lo: 298, hi: 305 } })"],
```
swernli, line 1288: Could you add a test case there that calls into something from stdlib that will fail with `PackageId(1)` to confirm that behavior?

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -74,29 +76,49 @@ pub enum Error {
...
     #[error("an unexpected error occurred related to: {0}")]
     #[diagnostic(code("Qsc.PartialEval.Unexpected"))]
     #[diagnostic(help(
         "this is probably a bug, please consider reporting this as an issue to the development team"
     ))]
-    Unexpected(String, #[label] Span),
+    Unexpected(String, #[label] PackageSpan),
 
-    #[error("failed to evaluate: {0} not yet implemented")]
+    #[error("failed to evaluate: {0} is not supported yet")]
```
minestarks, line 99: Nit: I don't like the word "yet" in error messages. Never promise features in an error message :)

minestarks:
```suggestion
    #[error("failed to evaluate: {0} is not supported")]
```

compiler/qsc_partial_eval/src/lib.rs
minestarks, line 1: I may have missed them, but are there any tests for partial eval errors returned from the standard library?

minestarks (approved):

compiler/qsc_partial_eval/src/lib.rs
cesarzc: I think there is not, I'll add one

compiler/qsc_partial_eval/tests/calls.rs
cesarzc: I'll add one :)

cesarzc: Added test.

compiler/qsc_partial_eval/src/lib.rs
cesarzc: Added test.

PR #1548, opened 2024-05-21
Updated UX for VSCode QIR generation

swernli:
This change updates the extension behavior for QIR code generation to give more granular error messages and guidance when possible. We used to always report an error with "Submitting to Azure" and recommend switching to Base profile. Now the message and guidance will differentiate between a few different scenarios:
 - Generating QIR with profile set to unrestricted will use codegen specific message and prompt for Base Profile.
 - Submitting to Azure target with profile set to unrestricted will recommend either Base or Adaptive_RI depending on what the target supports.
 - Submitting to Azure target that only supports Base profile with Adaptive_RI configured will give specific error message and recommend Base.

 This also fixes a minor bug with a missing await that could cause an error message to crop up even if the user clicked "Set" to take the recommended change.


vscode/src/qirGeneration.ts
```diff
@@ -30,25 +32,43 @@ export async function getQirForActiveWindow(): Promise<string> {
...
-  // Check that the current target is base or adaptive_ri profile, and current doc has no errors.
   const targetProfile = getTarget();
-  if (targetProfile === "unrestricted") {
+  const is_unrestricted = targetProfile === "unrestricted";
+  const is_base = targetProfile === "base";
+
+  let error_msg =
+    supports_adaptive === undefined
+      ? "Generating QIR "
+      : "Submitting to Azure ";
```
billti, line 42: I'm assuming from the logic that the parameter is only passed when submitted to Azure? Maybe a comment to that effect.

billti (approved):

vscode/src/qirGeneration.ts
swernli: Yes, that's correct. I added some comments here and in the signature to clarify the intent.

PR #1551, opened 2024-05-21
Update sparse simulator revision, cargo updates

swernli:
This updates the revision of the sparse simulator used to pull in the fix from https://github.com/qir-alliance/qir-runner/pull/173. The change also includes the results from `cargo update` to get dependent packages up the latest minor/rev version compatible with the repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -136,21 +137,15 @@ dependencies = [
...
 [[package]]
 name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
-name = "bitflags"
-version = "2.4.2"
+version = "2.5.0"
```
billti, line 140: There's quite a lot of package bumps here, including to some major package versions. We feel pretty confident in bumping all these right now?

billti (approved):

Cargo.lock
swernli: Confident enough… this is all from cargo update.

PR #1552, opened 2024-05-21
Surface partial evaluation errors in VS Code

minestarks:
- Defines a struct `IQSharpError` at the WASM boundary to contain full error information from Q# interpreter error calls. Currently this is only used in QIR generation and circuit visualization, but it can and should eventually be wired up for all `ICompiler` methods that can return errors.

- We create a new diagnostic collection (a set of errors to be shown in the Problems view in VS Code) that shows errors from the last run command. These errors persist in the editor, but can be dismissed by the user or when another command is run.

partial-eval-errors

### DIFF
--- a/vscode/src/diagnostics.ts
+++ b/vscode/src/diagnostics.ts
@@ -238,6 +238,7 @@ function reportIfQSharpErrors(e: unknown) {
       }
     } catch (_) {
       // Couldn't parse the error as JSON.
+      log.warn(`could not parse error string ${e}`);
     }
   }
 
--- a/vscode/src/diagnostics.ts
+++ b/vscode/src/diagnostics.ts
@@ -275,8 +275,12 @@ function reportIfQSharpErrors(e: unknown) {
  * @returns A VS code URI that's okay to use in a Diagnostic object
  */
 function getSourceUri(maybeUri: string): vscode.Uri {
+  // An error without a span (e.g. "no entrypoint found") gets reported as a "project-level" error.
+  // See: https://github.com/microsoft/qsharp/blob/f8d344b32a1f1f918f3c91edf58c975db10f4370/wasm/src/diagnostic.rs#L191
+  // Ideally this would be a proper URI pointing to the project root or root document.
+  // For now, make up a fake file path for display purposes.
   if (maybeUri === "<project>") {
-    return vscode.Uri.parse(maybeUri, false); // strict=false ignores lack of scheme
+    return vscode.Uri.file("Q# project");
   }
 
   try {
--- a/vscode/src/diagnostics.ts
+++ b/vscode/src/diagnostics.ts
@@ -3,6 +3,7 @@
 
 import {
   ILanguageService,
+  IQSharpError,
   VSDiagnostic,
   log,
   qsharpLibraryUriScheme,
@@ -16,6 +17,7 @@ export function startCheckingQSharp(
   return [
     ...startLanguageServiceDiagnostics(languageService),
     ...startQsharpJsonDiagnostics(),
+    ...startCommandDiagnostics(),
   ];
 }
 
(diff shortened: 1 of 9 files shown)


vscode/src/diagnostics.ts
```diff
@@ -130,3 +140,151 @@ export function updateQSharpJsonDiagnostics(
...
+
+  let qsharpErrors: IQSharpError[] | undefined;
+  if (typeof e === "string") {
+    try {
+      const errors = JSON.parse(e);
+      // Check for the shape of IQSharpError[]
+      if (Array.isArray(errors) && errors.length > 0 && errors[0].document) {
+        qsharpErrors = errors;
+      }
+    } catch (_) {
```
billti, line 239: Should there maybe be a log.warn here or something? Is this expected to happen ever?

vscode/src/diagnostics.ts
```diff
@@ -130,3 +140,151 @@ export function updateQSharpJsonDiagnostics(
...
+
+/**
+ * This is temporary until we're able to report proper stdlib and project URIs from
+ * the wasm layer. See https://github.com/microsoft/qsharp/blob/f8d344b32a1f1f918f3c91edf58c975db10f4370/wasm/src/diagnostic.rs
+ *
+ * @param maybeUri A source name returned from a Q# diagnostic
+ * @returns A VS code URI that's okay to use in a Diagnostic object
+ */
+function getSourceUri(maybeUri: string): vscode.Uri {
+  if (maybeUri === "<project>") {
```
billti, line 278: When is the maybeUri `<project>`? Worth a note.

swernli (approved):

vscode/src/diagnostics.ts
```diff
@@ -130,3 +140,151 @@ export function updateQSharpJsonDiagnostics(
...
+    return vscode.Uri.parse(maybeUri, false); // strict=false ignores lack of scheme
+  }
+
+  try {
+    return vscode.Uri.parse(maybeUri, true);
+  } catch (e) {
+    // Not a URI, assume it's a filename from the stdlib
```
billti, line 285: Maybe not urgent for this PR, but `Not a URI, assume it's a filename from the stdlib` seems dangerous. Are we sure no other errors might get here? Is there some other condition we could check to be sure? (e.g. would it contain `library/std` in the path in that case?)

vscode/src/diagnostics.ts
minestarks: Added comments and improved the handling slightly.

billti (approved):

vscode/src/diagnostics.ts
```diff
@@ -130,3 +140,151 @@ export function updateQSharpJsonDiagnostics(
...
+/**
+ * This is temporary until we're able to report proper stdlib and project URIs from
+ * the wasm layer. See https://github.com/microsoft/qsharp/blob/f8d344b32a1f1f918f3c91edf58c975db10f4370/wasm/src/diagnostic.rs
+ *
+ * @param maybeUri A source name returned from a Q# diagnostic
+ * @returns A VS code URI that's okay to use in a Diagnostic object
+ */
+function getSourceUri(maybeUri: string): vscode.Uri {
+  if (maybeUri === "<project>") {
+    return vscode.Uri.parse(maybeUri, false); // strict=false ignores lack of scheme
```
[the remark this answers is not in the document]
minestarks: Added comments and improved the handling slightly.

cesarzc (approved):

vscode/src/diagnostics.ts
minestarks: Definitely dangerous. _Right now_ I'm sure, by virtue of having written the code myself, but it's delicate for sure. This would be part of the broader improvements I'd like to make in error handling.

There's pretty much no meaningful indicator we can check. "arrays.qs" is an example of what can end up here.

vscode/src/diagnostics.ts
minestarks: Right now, no. But there are some functions in the WASM layer that just throw message `string`s as errors. It's conceivable that those may come through this code path in the future. I guess warning in that case would be a good thing... throwing raw strings is not a great idea in general.

vscode/src/diagnostics.ts
minestarks: Will address this in a future change.

PR #1559, opened 2024-05-22
Drop FIR increment if FIR passes fail

swernli:
This fixes an issue with incremental compilation where an increment with a failure could still have that FIR present in the package during later compilation, making errors "sticky" instead of associated with just the increment. This achieves it by tracking what the most recently lowered additions to the package were and removing them if there were pass failures like those from RCA check.

### DIFF
--- a/compiler/qsc/src/interpret/tests.rs
+++ b/compiler/qsc/src/interpret/tests.rs
@@ -517,22 +517,7 @@ mod given_interpreter {
         }
 
         #[test]
-        fn callables_failing_profile_validation_are_still_registered() {
-            fn verify_same_error<E>(result: &Result<Value, Vec<E>>, output: &str)
-            where
-                E: Diagnostic,
-            {
-                is_only_error(
-                    result,
-                    output,
-                    &expect![[r#"
-                    cannot use a dynamic integer value
-                       [line_0] [set x = 2]
-                    cannot use a dynamic integer value
-                       [line_0] [x]
-                "#]],
-                );
-            }
+        fn callables_failing_profile_validation_are_not_registered() {
             let mut interpreter = get_interpreter_with_capbilities(TargetCapabilityFlags::Adaptive);
             let (result, output) = line(
                 &mut interpreter,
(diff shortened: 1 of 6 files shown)


pip/tests/test_interpreter.py
```diff
@@ -284,20 +283,31 @@ def test_callables_failing_profile_validation_are_still_registered() -> None:
     assert "Qsc.CapabilitiesCk.UseOfDynamicDouble" in str(excinfo)
     with pytest.raises(Exception) as excinfo:
         e.interpret("Foo()")
-    assert "Qsc.CapabilitiesCk.UseOfDynamicDouble" in str(excinfo)
+    assert "Qsc.Eval.UnboundName" in str(excinfo)
```
minestarks, line 286: If I understand correctly, the old behavior is more desirable, but this is the best we can do for now. Could you add a comment explaining why? For the next time we encounter these quirks after encountering a total memory loss? ;)

compiler/qsc/src/interpret/tests.rs
```diff
@@ -517,69 +517,63 @@ mod given_interpreter {
...
+                &result,
+                &output,
+                &expect![[r#"
+                cannot use a dynamic integer value
+                   [line_0] [set x = 2]
+                cannot use a dynamic integer value
+                   [line_0] [x]
+            "#]],
+            ); // do something innocuous
             let (result, output) = line(&mut interpreter, indoc! {r#"Foo()"#});
```
minestarks, line 538: This tests that the evaluator fails gracefully, but what about one for the partial evaluator? What happens if we codegen for `Foo()`?

minestarks (approved):

compiler/qsc/src/interpret/tests.rs
swernli: that test is in the Python, pip/tests/test_interpreter.py. I could reproduce it here too.

pip/tests/test_interpreter.py
swernli: I'll comment some more, yeah. The previous behavior was a bit misleading in the unit test... the error was coming from the definition in the previous interpret call in addition to the call the site, so two errors were being reported. Now it's more consistent with the behavior you'd see when there was an HIR pass failure: the callable is not defined, but the name is still hanging around.

compiler/qsc/src/interpret/tests.rs
swernli: added!

samples/notebooks/azure_submission.ipynb
```diff
@@ -243,7 +231,7 @@
       "name": "python",
       "nbconvert_exporter": "python",
       "pygments_lexer": "ipython3",
-      "version": "3.11.7"
+      "version": "3.11.6"
```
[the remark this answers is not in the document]
swernli: Not intentional, just inherited from my environment. I'll revert this part of the change to avoid confusion.

cesarzc (approved):

PR #1562, opened 2024-05-23
Added dot product via iterative phase estimation sample

DmitryVasilevsky:
This sample includes two version. The version for Unrestricted profile computes actual dot product using floating point numbers and is not suitable for current hardware. The version for Adaptive profile doesn't include floating point calculations and only provides estimation for the dot product. Multiple shots can be used to calculate it more precisely.
Original code is Copyright (c) 2023 KPMG Australia.


samples/algorithms/DotProductViaPhaseEstimation.qs
```diff
@@ -0,0 +1,165 @@
...
+        TargetReg : Qubit,
+        AncilReg : Qubit,
+        theta1 : Double,
+        theta2 : Double
+    ) : Unit is Adj + Ctl {
+
+        H(AncilReg);
+        // Arbitrary controlled rotation based on theta. This is vector v.
+        Controlled R([AncilReg], (PauliY, theta1, TargetReg));
+        // X gate on ancilla to change from |+> to |->.
```
swernli, line 141:
```suggestion
        // X gate on ancilla to change from |+〉 to |-〉.
```

swernli (approved):
Looks good, thanks!

PR #1563, opened 2024-05-23
Use __quantum__qis__m__body instead of __quantum__qis__mz__body in QIR generation

cesarzc:
This change updates QIR generation to use `__quantum__qis__m__body` instead of `__quantum__qis__mz__body`.


compiler/qsc_rir/src/passes/reindex_qubits/tests.rs
```diff
@@ -236,7 +236,7 @@ fn qubit_reindexed_multiple_times_with_mz_inserts_multiple_cx() {
     let mut program = Program::new();
     program.num_qubits = 1;
     program.callables.insert(X, x_decl());
-    program.callables.insert(MZ, mz_decl());
+    program.callables.insert(MZ, m_decl());
```
swernli, line 239:
```suggestion
    program.callables.insert(M, m_decl());
```

swernli, line 239: This symbol didn't get updated like the rest. Up to you if you want to do it now or not.

swernli (approved):

compiler/qsc_rir/src/passes/reindex_qubits/tests.rs
cesarzc: Updated.

billti (approved):

PR #1564, opened 2024-05-23
Support logical `not` in `@Config` attribute

swernli:
This change adds support for logical `not` inside of `@Config` attributes and makes it work as expected with the additive native of the attributes. So to specify an item that should only be available when compiling with `Adaptive` and `IntegerComputations`, but not with `HigherLevelConstructs` like dynamic strings, you can write:
```qsharp
@Config(Adaptive)
@Config(IntegerComputations)
@Config(not HigherLevelConstructs)
operation Foo() : Unit { }
```


compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -2033,6 +2040,263 @@ fn dropped_callable() {
...
+                function B() : Unit {
+                    Dropped();
+                }
+            }
+        "},
+        TargetCapabilityFlags::Adaptive,
+        &expect![[r#"
+            namespace namespace7 {
+                @Config(Adaptive)
+                @Config(not HigherLevelConstructs)
```
DmitryVasilevsky, line 2289: Is it possible to test from something other than HigherLevelConstructs?

swernli: Good idea, I'll add some of those.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -131,7 +131,11 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
         return true;
     }
     let mut found_capabilities = TargetCapabilityFlags::empty();
+    let mut disallowed_capabilities = TargetCapabilityFlags::empty();
 
+    // When checking attributes, anything we don't recognize (invalid form or invalid capability) gets
+    // left in the compilation by returning true. This ensures that later compilation steps, specifically lowering
+    // from AST to HIR, can check the attributes and return errors as appropriate.
     for attr in attrs {
```
idavis, line 139: We need to handle `not Base` and `not Unrestricted`.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -143,17 +147,31 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
...
+                    }
+                }
                 _ => return true, // Unknown config attribute, so we assume it matches
             }
         } else {
             // Something other than a parenthesized expression, so we assume it matches
             return true;
         }
     }
-    if found_capabilities == TargetCapabilityFlags::empty() {
```
idavis, line 153: Should we update the existing `@Config(Adaptive)` to be `@Config(not Base)`?

swernli: I don't think so... since `Unrestricted` is a special case meaning "all capabilities" it includes "adaptive." But I could be convinced otherwise if you think it's clearer in the code base.

compiler/qsc_frontend/src/compile/preprocess.rs
swernli: Good catch! I'll fix the handling there and get test cases added.

compiler/qsc_frontend/src/compile/preprocess.rs
swernli: I'll take a look through the standard library and update any markers. I think we did expect that the code could use different attributes once the new RCA check pass was default and base profile check was removed.

compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -2033,6 +2069,462 @@ fn dropped_callable() {
...
+            namespace A {
+                @Config(Adaptive)
+                function Dropped() : Unit {}
+
+                function B() : Unit {
+                    Dropped();
+                }
+            }
+        "},
+        TargetCapabilityFlags::all(),
```
idavis, line 2255: Test name and flags don't match

swernli: Fixed!

idavis (approved):

library/std/internal.qs
```diff
@@ -163,7 +163,7 @@ namespace Microsoft.Quantum.Intrinsic {
         }
     }
 
-    @Config(Base)
+    @Config(not Adaptive)
```
DmitryVasilevsky, line 166: Is this change to contrast the other 'Adaptive' version? Base seems a bit easier to understand - not requiring any special capabilities.

DmitryVasilevsky (approved):

library/std/internal.qs
swernli: In this case, it's meant to be more specific. `@Config(Base)` would mean it matches only the empty set of capabilities, whereas this version of the callable should be used if the target doesn't support the `Adaptive` capability. So if the target capabilities were for some reason configured with just `QubitReset` and nothing else, this is the version that should be used rather than the one that required adaptivity.

PR #1565, opened 2024-05-23
Respect configured target profile for histogram in VS Code

swernli:
This updates the call to `run` to that is used by the histogram feature to take the current target profile, such that histograms in the VS Code extension correctly compile the code with the configured target.

### DIFF
--- a/npm/qsharp/src/compiler/compiler.ts
+++ b/npm/qsharp/src/compiler/compiler.ts
@@ -240,6 +240,7 @@ export class Compiler implements ICompiler {
     eventHandler: IQscEventTarget,
   ): Promise<void> {
     let sources;
+    let profile: TargetProfile = "unrestricted";
     let languageFeatures: string[] = [];
 
     if (Array.isArray(sourcesOrConfig)) {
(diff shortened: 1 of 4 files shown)


wasm/src/lib.rs
```diff
@@ -353,6 +354,7 @@ pub fn run(
     event_cb: &js_sys::Function,
     shots: u32,
     language_features: Vec<String>,
+    profile: &str,
```
billti, line 357: This would likely be a breaking change for QCOM I guess if we change the API to run code.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -240,6 +240,7 @@ export class Compiler implements ICompiler {
     eventHandler: IQscEventTarget,
```
billti, line 240: Actually, maybe not if this API didn't change. We should check they are only using the ICompiler interface, and not calling any wasm exports directly (which they shouldn't be)

wasm/src/lib.rs
swernli: Actually, I don't think it is, as this is the wasm but the npm already handles overloading to two versions:
https://github.com/microsoft/qsharp/blob/swernli/histogram-respects-target/npm/qsharp/src/compiler/compiler.ts#L52-L64

billti (approved):

npm/qsharp/src/compiler/compiler.ts
swernli: Yeah, that's what I was assuming. You are right that it could be a breaking change for anything calling directly into WASM though.

swernli: I double checked the calls there, and nothing goes directly to WASM, it all uses the compiler from the npm package. So we should be good!

PR #1566, opened 2024-05-23
Introduce glob imports in import statements

sezna:
_branched off of #1469_

This PR mainly introduces glob imports in import statements, e.g.:

 ```
import Foo.*;
import Foo.Bar.* as Baz;
```
et cetera.

Glob imports work similarly to other languages: `Foo.*` brings in everything from `Foo`, while `Foo.* as Bar` brings in all of `Foo`, but aliased under a namespace `Bar`. For example, if `function F() : Unit {}` exists in `Foo`, then after importing via `Foo.* as Bar`, it would be available as `Bar.F()`.

Other incidental changes include:
- Improving the `Display` impl for `ImportOrExportItem`, for easier-to-read tests

### DIFF
--- a/compiler/qsc_parse/src/item/tests.rs
+++ b/compiler/qsc_parse/src/item/tests.rs
@@ -1853,7 +1853,7 @@ fn parse_glob_import() {
     check(
         parse_import_or_export,
         "import Foo.*;",
-        &expect![[r#"ImportOrExportDecl [0-13]: [Path _id_ [7-12] (Ident _id_ [7-10] "Foo").*]"#]],
+        &expect![[r#"ImportOrExportDecl [0-13]: [Path _id_ [7-10] (Ident _id_ [7-10] "Foo").*]"#]],
     );
 }
 
@@ -1863,7 +1863,7 @@ fn parse_glob_import_in_list() {
         parse_import_or_export,
         "import Foo.Bar, Foo.Baz.*;",
         &expect![
-            r#"ImportOrExportDecl [0-26]: [Path _id_ [7-14] (Ident _id_ [7-10] "Foo") (Ident _id_ [11-14] "Bar"), Path _id_ [16-25] (Ident _id_ [16-19] "Foo") (Ident _id_ [20-23] "Baz").*]"#
+            r#"ImportOrExportDecl [0-26]: [Path _id_ [7-14] (Ident _id_ [7-10] "Foo") (Ident _id_ [11-14] "Bar"), Path _id_ [16-23] (Ident _id_ [16-19] "Foo") (Ident _id_ [20-23] "Baz").*]"#
         ],
     );
 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: compiler/qsc_parse/src/prim.rs
Comment (minestarks) [inline compiler/qsc_parse/src/prim.rs:1]: Surely this is a spurious change that snuck in?

File: compiler/qsc_parse/src/item.rs [hunk 1]
```diff
@@ -561,12 +562,31 @@ fn parse_import_or_export(s: &mut ParserContext) -> Result<ImportOrExportDecl> {
...
     // an import item is a path followed by an optional alias,
-    let path = *(path(s)?);
+    let mut idents = Vec::default();
+    let mut is_glob = false;
+    loop {
+        idents.push(*ident(s)?);
+        if s.peek().kind == TokenKind::Dot {
+            s.advance();
+        }
+        if s.peek().kind == TokenKind::ClosedBinOp(ClosedBinOp::Star) {
```
minestarks, line 572: I leave the solution up to your judgment, but it sort of bothers me that `*` is still considered a `ClosedBinOp`. I don't know if this has any practical implications in the code, or if it's just a naming thing.

compiler/qsc_parse/src/item.rs
```diff
@@ -561,12 +562,31 @@ fn parse_import_or_export(s: &mut ParserContext) -> Result<ImportOrExportDecl> {
...
 fn parse_import_or_export_item(s: &mut ParserContext) -> Result<ImportOrExportItem> {
     // an import item is a path followed by an optional alias,
-    let path = *(path(s)?);
+    let mut idents = Vec::default();
+    let mut is_glob = false;
+    loop {
+        idents.push(*ident(s)?);
+        if s.peek().kind == TokenKind::Dot {
+            s.advance();
+        }
```
minestarks, line 571: So what if the next token isn't a dot?

The below parses as an import item currently :)

```qsharp
import Foo Bar Baz *;
```

You may want to use a loop like `while token(s, TokenKind::Dot).is_ok()`

compiler/qsc_parse/src/item.rs
minestarks, line 572:
```
if token(s, SomeTokenKind).is_ok() {
    is_glob = true;
    break;
}
```

matches existing code patterns better

compiler/qsc_parse/src/item.rs
```diff
@@ -561,12 +562,31 @@ fn parse_import_or_export(s: &mut ParserContext) -> Result<ImportOrExportDecl> {
...
+        if s.peek().kind == TokenKind::ClosedBinOp(ClosedBinOp::Star) {
+            is_glob = true;
+            s.advance();
+            break;
+        }
+        if s.peek().kind != TokenKind::Ident {
+            break;
+        }
```
minestarks, line 579: I think you'll want to avoid breaking here so we loop back around to consume an ident (now that we've seen a dot).

Overall, I think this loop could be cleaned up a bit - maybe just copy the `path` parser and add the check for the `*`?

minestarks (approved):

PR #1571, opened 2024-05-24
Update RCA for mutable `Result`, array, and tuple variables

swernli:
This change strengthens RCA checks around use of mutable `Result`, array, and tuple variables in a dynamic context to make them consistent with how we verify the use of other dynamic value types like `Int` and `Double`. This gives users compile time feedback when updating a mutable value in a dynamic context will not produce QIR, rather than a codegen failure with `an unexpected error occurred related to: re-assignment within a dynamic branch is unsupported for type <type>` error.


compiler/qsc_rca/src/errors.rs
```diff
@@ -51,6 +51,20 @@ pub enum Error {
     #[diagnostic(code("Qsc.CapabilitiesCk.UseOfDynamicQubit"))]
     UseOfDynamicQubit(#[label] Span),
 
+    #[error("cannot use a dynamic result")]
+    #[diagnostic(help(
+        "using a mutable Result variable whose value depends on a measurement result is not supported by the current target"
```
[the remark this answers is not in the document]
swernli: that's a good point... at first this error was limited to mutable Q# variables, but it turns out it also needs to apply to code with immutable binding, like:
```qsharp
let result = if M(q) == Zero { One } else { M(q) };
```
I'll update the text to not use that term, since it's no longer specific to that.

compiler/qsc_rca/src/core.rs
```diff
@@ -684,12 +696,24 @@ impl<'a> Analyzer<'a> {
...
                 derive_runtime_features_for_value_kind_associated_to_type(
                     dynamic_value_kind,
                     expr_type,
                 );
+            if is_any_result(expr_type) {
+                dynamic_runtime_features |= RuntimeFeatureFlags::UseOfDynamicResult;
+            }
+            if matches!(expr_type, Ty::Tuple(tup) if !tup.is_empty()) {
+                dynamic_runtime_features |= RuntimeFeatureFlags::UseOfDynamicTuple;
+            }
```
swernli, line 716: There's a bug here... this is rejecting use of the `Measure` operation, even though the condition there is classical.

swernli: Whew, found it!

idavis (approved):

PR #1575, opened 2024-05-28
Support error Uri, update RCA error strings

swernli:
This change adds support for error Uris by hooking the miette error `.url()` to the VS Code and Monaco `target` field. It also updates the wording on RCA check pass errors to use "configured target profile" instead of "current target" for more clarification. A follow up PR can update the error link(s) to point to more specific documentation pages.

In VS Code, the error links show up as a clickable error code:
image

In Python/Jupyter, the miette visualization is used which puts clickable links next to the error code:
image

In Monaco (used in the playground) links work like the VS Code version:
image


idavis (approved):

minestarks (approved):
So clean!

PR #1578, opened 2024-05-29
Add images to Deutsch algorithm kata

### LINKED ISSUE #1156: Migrate ExploringDeutschJozsaAlgorithm kata to quantum.microsoft.com
Current [ExploringDeutschJozsaAlgorithm tutorial](https://github.com/microsoft/QuantumKatas/tree/main/tutorials/ExploringDeutschJozsaAlgorithm) can not be used with the modern QDK. It would be nice to have it available as part of the [new katas experience](https://quantum.microsoft.com/en-us/experience/quantum-katas).

It has _a lot_ of content (4 separate notebooks!)

Steps:

* Split the tutorial in two: Deutsch algorithm and Deutsch-Jozsa algorithm
* From current part I, I would drop the tasks on implementing classical tasks, and keep just the problem definition for Deutsch-Jozsa algorithm tutorial.
* [x] Current part II would become Deutsch algorithm tutorial, I'd keep the oracle implementation examples but add demos to show what these examples do to states, using 0.6|0> + 0.8|1> (currently the oracles are just compiled, not executed).
* [x] Current part III would become Deutsch-Jozsa algorithm tutorial. I would drop the videos and the section that explores inputs other than constant and balanced functions, it seems like too much. I would also convert some of the oracle examples into exercises to give the learner more practice.
* Drop current part IV, since it focuses on Azure Quantum job submission, and we cannot do that from the website.
* [ ] The images (Deutsch algorithm) have to be redesigned to match the overall theme of the site.
* The solution descriptions need to be written, since the original tutorials don't have them.
* [ ] Ideally we'd need AllowAtMostNCallsCA equivalent to verify how many oracle calls the solution did (https://github.com/microsoft/qsharp/issues/1154), but we can have the first version without this.

**Open question: how to merge this with the [old DeutschJozsaAlgorithm kata](https://github.com/microsoft/QuantumKatas/blob/main/DeutschJozsaAlgorithm/DeutschJozsaAlgorithm.ipynb)?** 

DeutschJozsaAlgorithm kata predates ExploringDeutschJozsaAlgorithm tutorial, so it uses different notations (marking oracles instead of phase oracles). It also predates the Oracles tutorial, so a lot of the material is already covered there, including conversion from marking to phase oracle. However, it has some unique content - Bernstein-Vazirani algorithm (this one has a lot fewer resources covering it than DJ, so it's good to keep it) and nice marking oracles. 

I would suggest the following:

* [x] Add Bernstein-Vazirani algorithm as the last task to Deutsch-Jozsa algorithm tutorial (the math is exactly the same, only the interpretation differs, so it makes sense to keep them together rather than doing a standalone tutorial) - need to pull in some math about applying Hadamard gates to an arbitrary state for this, since we're taking a shortcut in DJ tutorial looking at just one state
* [ ] Add marking in tasks 1.5, 1.6, 1.8 to the future MarkingOracles kata
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/qsharp/pull/1228
- [ ] https://github.com/microsoft/qsharp/pull/1231
```

tcNickolas:
This closes #1156.


tcNickolas:
I redrew images compared to what we have in the classic katas (to remove column vectors and only use ket notations), and that sped up image load significantly compared to the old images - I don't see a noticeable delay in kata load now

billti:
Do we ever expect katas to support 'themes' or different color backgrounds? These images look fine on a light background, e.g.,

image

But if I switch to a dark background some of the content disappears

image

The SVG is also still a little verbose, but we can optimize that later. (The 3 images are about 150kb, so we wouldn't want to add too many. Looking at the complexity of the image I'm guessing it could be done in about 10kb).

tcNickolas:
I haven't heard anything about themes support in the katas (personally I don't use dark themes outside of console, since they hurt my eyes). This image is based on the "official" one provided by our designer team (same style, different notations), so I guess we don't have themes at the moment.

I'm happy to accept optimization suggestions, as long as they come with specific commands to do them that can be easily followed by non-designers :-) These images, I got by saving PowerPoint diagram in svg and then scaling them using Figma, and they're a lot smaller than the previous iteration with column vectors, so I'm happy with them. I don't anticipate having more than 5-ish images per kata, and that's only for Grover, most are going to be image-free, so I hope it won't be a problem.

tcNickolas:
I used Inkscape to optimize the images automatically, so the three of them are ~68k now.

cesarzc (approved):

PR #1579, opened 2024-05-29
[Unitary-Hack] Lint rule: Use functions

### LINKED ISSUE #1471: Lint rule: `operation`s that don't manipulate quantum state can be `function`s instead
We should add a rule to the linter that will emit a warning when an `operation` does not manipulate quantum state. The warning should suggest declaring the callable as a `function` instead.

Context:

In Q#, callables declared as `operation` can manipulate quantum state (allocate qubits, apply gates, call other `operation`s etc) while callables declared as `function` must be pure functions and cannot manipulate state.

Currently, a callable such as the one below can be declared as an `operation` or `function`. 

```qsharp
operation Add(a: Int, b: Int) : Int {
    return a + b;
}

function Add(a: Int, b: Int) : Int {
    return a + b;
}
```
[Playground link](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACp2NvQoCMRCE%2BzzF2CUoHFqejRYKFoKFPsDmRwzo5sglhci9uyR3gli6xcwwMN8yPVzfkXHYh4CXAIAqmx2n%2BDwFz0mq2oTORUo%2BMI7kWSq0uLBP02gQ1ZoG55vvYUK%2BW2gHwjWzKbPZD2Vr7VISWhw4LaDHoEaboOWiSzkyCHPo9ferD7eAVn%2BCBvEG4%2Fwj7gEBAAA%3D&profile=unrestricted)

Since this is a pure function, `function` is a more appropriate designation for it.

Declaring the proper callable type not only helps prevent mistakes, it may also allow runtime capability analysis to run faster since the compiler can rely on the knowledge that `function`s don't manipulate quantum state.

----

**CONTRIBUTORS PLEASE READ**

**Getting started**

Welcome! Please take a look through our [README](https://github.com/microsoft/qsharp/blob/main/README.md) to orient yourself in the repo and find instructions on how to build.

For this issue you'll want to have a working knowledge of Rust and compilers.

For documentation on how to add lints, see:
https://github.com/microsoft/qsharp/blob/main/compiler/qsc_linter/src/lib.rs

For examples of existing lints, see:

https://github.com/microsoft/qsharp/blob/ed2a8fcab06acea8e2b8fcaf64f109bbd979f403/compiler/qsc_linter/src/lints/ast.rs#L10-L12

https://github.com/microsoft/qsharp/blob/ed2a8fcab06acea8e2b8fcaf64f109bbd979f403/compiler/qsc_linter/src/lints/hir.rs#L8

**Testing**

You can demonstrate that the lint works by running the playground locally (see the code example in the issue description).

Please add unit tests verifying the functionality you implemented.

Before you submit a pull request please run `python ./build.py` to ensure the project builds cleanly. See README for details.

**Reviews**

Once you have published your PR, the codeowners will automatically get tagged and we'll review shortly.

If you need clarification on an issue please tag @minestarks with your questions.

Manvi-Agrawal:
Fixes #1471.

### Future TODO
- https://github.com/microsoft/qsharp/issues/1585
- https://github.com/microsoft/qsharp/issues/1593

image
image

Based on https://github.com/microsoft/qsharp/blob/main/compiler/qsc_passes/src/callable_limits.rs

### Note to reviewers
Main changes are in `qsc_linter` folder. Changes in other tests are due to this lint rule complaining about NeedlessOperation, so I converted to function and it changes span accordingly.


samples/language/GettingStarted.qs
```diff
@@ -1,13 +0,0 @@
...
-/// # Description
-/// This is a minimal Q# program that can be used to start writing Q# code.
-namespace MyQuantumProgram {
-
-    @EntryPoint()
-    operation Main() : Result[] {
-        // TODO: Write your Q# code here.
-        return [];
-    }
-}
```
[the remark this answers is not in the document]
minestarks: That's a great question, I hadn't considered it. The sample is still useful since we use it as the front page at https://microsoft.github.io/qsharp/ and it's nice to be able to just go in and start typing in a blank(-ish) template.

For now, I think it's ok to change it to the following:

```qsharp
@EntryPoint()
operation Main() : Unit {
    use q = Qubit();
}
```

samples/samples.mjs
```diff
@@ -7,7 +7,6 @@
 
 /** @type {Array<{title: string; file: string; shots: number; omitFromTests?: boolean}>} */
 export default [
-    { title: "Minimal", file: "./language/GettingStarted.qs", shots: 100 },
```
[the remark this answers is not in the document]
minestarks: Yeah, I think it's ok to fix up the samples in a different PR.

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,80 @@
...
+impl Visitor<'_> for OperationLimits {
+    fn visit_callable_decl(&mut self, decl: &CallableDecl) {
+        if decl.kind == CallableKind::Operation {
+            // Checking for special Ctl and Adj implementations
+            // if decl.adj.is_some() || decl.ctl.is_some() || decl.ctl_adj.is_some() {
+            //     self.op_char.push(decl.span);
+            // }
+            // if decl.functors != FunctorSetValue::Empty {
+            //     self.op_char.push(decl.name.span);
+            // }
```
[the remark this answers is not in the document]
minestarks: I think it's OK not to show the lint if you have an `operation` that defines explicit specializations. 

Most importantly, this is so that our own code in the standard library compiles without warnings:
https://github.com/microsoft/qsharp/blob/9aa8aaff64e08d481171b71245d5a072350f0adf/library/std/intrinsic.qs#L184-L193

But we can show the lint if you have _implicit_ specializations. Funny enough, this means that our own specialization sample would need to be fixed.
https://github.com/microsoft/qsharp/blob/9aa8aaff64e08d481171b71245d5a072350f0adf/samples/language/Specializations.qs#L12-L16

A user who wants to deliberately override the lint can define an explicit specialization. Maybe that can be mentioned in the error help text.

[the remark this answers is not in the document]
minestarks: Hmm, I think you're correct in your understanding, is this still giving you trouble? Maybe the visitor skipped some blocks for some reason?

minestarks:
Thanks @Manvi-Agrawal ! Would you mind resolving the conflicts in the branch so I can kick off the checks for this PR? I'll give your questions some more thought once I get the feedback from the unit tests in the CI.

compiler/qsc_linter/src/lints/hir.rs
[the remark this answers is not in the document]
minestarks: Thanks for thinking of these edge cases, you're uncovering some stuff that hadn't occurred to me before!

After some discussion within the team, we realized that "empty operations" could also be a special case that we exclude from the lint. That takes care of `I` and `DoNothing`.

So the overall set of rules would be (unless I'm missing something)

- empty operations: no lint
- operations with errors (e.g. partially typed code): no lint*
- non-empty operations, with specializations, and no quantum operations: show lint, but don't offer quickfix (to avoid deleting user code in any explicit specializations)
- non-empty operations with _no_ specializations, and no quantum operations: show lint, offer quickfix to convert to function

*I don't think the linter runs at all if there are errors, so I don't think this needs special consideration.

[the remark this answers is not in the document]
minestarks: Oh! I did not realize this was happening - the implicit specializations are generated in the [`spec_gen` HIR pass](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_passes/src/spec_gen.rs). These passes happen before the linter runs. So by the time we see the HIR, all the specializations look the same.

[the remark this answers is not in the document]
minestarks: Sorry, one more to check! Lambdas can be declared as `operation`s or `function`s, differentiated by the arrow syntax.


```qsharp
let a = (q) => CNOT(q); // operation
let a = (q) -> CNOT(q); // function (error)
let a = (a) => a + 1;   // declared as operation, but could be function
let a = (a) -> a + 1; 
```

Ideally the 3rd one would be a lint. (Even if we can't support these fully, at the very least, we want to verify the behavior is reasonable.)

[Playground](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACm2NQQ6CMBRE9z3FLNsoIW4lGBPjEsVED%2FAp1ZBAC6VdEMLdDV1AjPzln5n3NDWqb0kqZMPDk3a%2Bya35WGowMgYA56t2dshNpR0X4WNaZclVRiOjSnOBI166chhDOl%2BtHAgpeCeQnnC53Z%2B8EwnieB1vlqPf8ttrGURcWWus%2BN9QEBB2OCTAvCmVrMmqEtSvsj0K7yCNr0sUauFu8qKFF%2BKJTewLZ4XROCYBAAA%3D&profile=unrestricted)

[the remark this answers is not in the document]
minestarks: I'd consider quick fix out of scope for the purposes of the bounty, but it certainly would be nice to have to make the feature feel complete!

minestarks:
@Manvi-Agrawal reran the build - looks like you have an outstanding clippy warning. Please let us know when this is ready for final review (all known issues addressed). I can't tell whether you're still planning to push more changes.

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,104 @@
...
 use crate::linter::hir::declare_hir_lints;
 
 use super::lint;
 
 declare_hir_lints! {
-    (Placeholder, LintLevel::Allow, "this a placeholder", "remove after addding the first HIR lint"),
+    (NeedlessOperation, LintLevel::Warn, "unnecessary operation declaration", "convert to function")
+}
+
+// Helper to check if a operation has desired operation characteristics
```
orpuente-MS, line 18: Typo
```suggestion
// Helper to check if a operation has desired operation characteristics
```
```suggestion
// Helper to check if an operation has desired operation characteristics
```

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,104 @@
...
+//
+// empty operations: no lint, special case of `I` operation used for Delay
+// operations with errors (e.g. partially typed code): no lint because linter does not run
+// non-empty operations, with specializations, and no quantum operations: show lint, but don't offer quickfix (to avoid deleting user code in any explicit specializations)
+// non-empty operations with no specializations, and no quantum operations: show lint, offer quickfix to convert to function
+#[derive(Default)]
+struct OperationLimits {
+    // Operation Characteristics
+    op_char: bool,
 }
```
orpuente-MS, line 28: Maybe we could name this struct something like `IsQuantumOperation` to convey its meaning better (or something along those lines). I didn't really get what "limits" or "op_char" meant when reading the code the first time.

```suggestion
struct OperationLimits {
    // Operation Characteristics
    op_char: bool,
}
```
```suggestion
struct IsQuantumOperation(bool);
```

orpuente-MS (commented):
Looks great so far! The code looks neat, and the intent was easy to understand. I left a couple comments, mostly nitpicks.

One general comment is that, when possible (in the rust files), is better to write the documentation as doc strings with triple slashes instead of as comments.

`// This is a comment.`

`/// This is a docstring.`

minestarks:
@Manvi-Agrawal Could you please run your local playground with the browser Developer Tools enabled, and watch the JavaScript console logs? It's reporting some panics from the Rust layer when you trigger the lint. Please ensure no panics / error messages show up in regular testing in the playground.

These panics also seem to be showing up in VS Code integration tests , but unfortunately they're not causing the tests to fail, so we didn't catch them earlier :(
https://github.com/microsoft/qsharp/actions/runs/9354102807/job/25746572167#step:9:798

compiler/qsc_linter/src/lints/hir.rs
orpuente-MS: Yes, that sounds good!

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,104 @@
...
+    // Operation Characteristics
+    is_op: bool,
 }
 
-impl HirLintPass for Placeholder {
-    fn check_expr(&self, expr: &qsc_hir::hir::Expr, buffer: &mut Vec<Lint>) {
-        if let qsc_hir::hir::ExprKind::Lit(Lit::Int(42)) = &expr.kind {
-            buffer.push(lint!(self, expr.span));
+impl IsQuantumOperation {
+    // Checks for empty declarations
```
orpuente-MS, line 31:
```suggestion
    /// Returns `true` if the declaration is empty.
```

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,104 @@
...
+        match spec_decl {
+            None => true,
+            Some(decl) => match &decl.body {
+                SpecBody::Gen(_) => true,
+                SpecBody::Impl(_, block) => block.stmts.is_empty(),
+            },
+        }
+    }
+
+    // Empty operation means no code for body and specializations.
```
orpuente-MS, line 42:
```suggestion
    /// Returns `true` if the operation is empty.
    /// An operation is empty if there is no code for its body and specializations.
```

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,104 @@
...
+
+/// Helper to check if an operation has desired operation characteristics
+///
+/// empty operations: no lint, special case of `I` operation used for Delay
+/// operations with errors (e.g. partially typed code): no lint because linter does not run
+/// non-empty operations, with specializations, and no quantum operations: show lint, but don't offer quickfix (to avoid deleting user code in any explicit specializations)
+/// non-empty operations with no specializations, and no quantum operations: show lint, offer quickfix to convert to function
+#[derive(Default)]
+struct IsQuantumOperation {
+    // Operation Characteristics
```
orpuente-MS, line 26:
```suggestion
    /// This field is set to `true` after calling `Visitor::visit_callable_decl(...)`
    /// if the operation satisfies the characteristics described above.
```

orpuente-MS (commented):
A few suggestions to convert some more comments to docstrings.

samples/language/GettingStarted.qs
orpuente-MS, line 1: @minestarks I don't remember what the agreement on the minimal sample was. This looks good to me, but could you double check this?

orpuente-MS (approved):
@Manvi-Agrawal looks good! I will approve.

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,106 @@
...
+/// We use [`IsQuantumOperation`] helper to check if a operation has desired operation characteristics
+impl HirLintPass for NeedlessOperation {
+    fn check_callable_decl(&self, decl: &CallableDecl, buffer: &mut Vec<Lint>) {
+        if decl.kind == CallableKind::Operation {
+            let mut op_limits = IsQuantumOperation::default();
+
+            op_limits.visit_callable_decl(decl);
+
+            if !op_limits.is_op {
+                buffer.push(lint!(self, decl.span));
```
minestarks, line 102: Just to make the squiggles a little less obtrusive while I'm still typing up the operation, can we use `decl.name.span` for this? 
```suggestion
                buffer.push(lint!(self, decl.name.span));
```

image

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -1,17 +1,106 @@
...
+    visit::{self, Visitor},
+};
 
 use crate::linter::hir::declare_hir_lints;
 
 use super::lint;
 
 declare_hir_lints! {
-    (Placeholder, LintLevel::Allow, "this a placeholder", "remove after addding the first HIR lint"),
+    (NeedlessOperation, LintLevel::Warn, "unnecessary operation declaration", "convert to function")
```
minestarks, line 15: Let's make the error message a bit more user-friendly. Here's my attempt:

```suggestion
    (NeedlessOperation, LintLevel::Warn, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
```

language_service/src/state/tests.rs
minestarks, line 1: Many of the changes in these tests apply to empty operations... can they be reverted now that we allow empty operations?

npm/qsharp/test/basics.js
minestarks, line 1: You should be able to revert this now because the operations are empty, right?

minestarks (changes requested):
Great work, thanks. Minor cosmetic improvements and then we should be good to go.

compiler/qsc_linter/src/lints/hir.rs
minestarks: You're right. VS Code does support clickable error codes and we've started to add help documentation for some of them. We could do this for lint warnings as well. I'll see if I can set up a page like that.

image

https://github.com/microsoft/qsharp/wiki/QIR#runtime-capabilities-check-errors

minestarks (approved):

samples/language/GettingStarted.qs
minestarks: Yeah I think this is good for the purposes of this PR (just making it lint compliant).

(We could possibly remove the comment & now even the EntryPoint attribute to make it even more minimal, but let's leave that for a different change)

minestarks:
Just needs one more signoff from @DmitryVasilevsky, @cesarzc, @sezna, @swernli, @tcNickolas (probably due to the small sample update) .

swernli (approved):

PR #1580, opened 2024-05-29
[Unitary hack] Add doc for internal AND

### LINKED ISSUE #1212: Add three qubit `AND` operation to public API
The change in https://github.com/microsoft/qsharp/pull/1202 adds an `AND` internal helper for use with decomposition of multi-controlled operations that is more efficient than `PhaseCCX` and is very similar to `ApplyAndAssuming0Target` internal helper from the unstable arithmetic namespace. Both of these are capturing a common pattern, which is to use an optimized decomposition of a reversible AND on three qubits that is better for resource estimation and simulation. Further, there is interest in making AND a `body intrinsic` operation that generates to a gate in QIR so that backend providers are free to implement the operation in whatever way is most efficient on their hardware.  We need to coordinate the effort here and make sure we have a cohesive plan for how to expose AND publicly from the standard library and how to expand the set of QIR intrinsics to include the notion of AND as distinct from a Toffoli/CCNOT.

Let's use this issue to discuss requirements, concerns, and staging for the change.

Manvi-Agrawal:
Fixes #1212.
image


swernli (commented):
Thanks for the contribution! An additional recommendation for the API that should be included in the documentation is the assumption that when the normal/non-adjoint specialization is called the `target` qubit is expected to be in the zero or ground state. I'm not quite sure how to phrase that guidance, exactly, but I think it's worth finding a way to include that in the Summary section.

library/std/intrinsic.qs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Quantum.Intrinsic {
 
     /// # Summary
     /// Applies the AND gate that is more efficient for use with decomposition of multi-controlled operations.
+    /// Please note that target qubit must be in $\ket{0}$ state.
```
swernli, line 13: I believe we need to avoid markdown in `Summary` and `Input` sections for compatibility with VS Code. You can use unicode here instead, and don't need the "Please" at the front:
```suggestion
    /// Note that target qubit must be in |0⟩ state.
```

swernli (approved):

swernli:
Looks like the change to the stdlib made some debugger tests fail because they confirm the line number that the debugger breaks into. Since the file contents changed, the line numbers have to be updated. A little annoying but an expected consequence of the change.

swernli:
To run the integration tests, you should be able to use the same python build script, but pass an additional flag:
`python build.py --integeration-tests`

swernli (approved):

minestarks (approved):

minestarks:
@Manvi-Agrawal I will merge this PR as soon as the `main` branch is reopened after our next release (soon)!

Issue #1333, opened 2024-03-28
Q# Lints


orpuente-MS:
This issue is a place to gather ideas/proposals for new lints.

Future lints:
 - Dead Code

Current Lints:
 - `DivisionByZero`
 - `NeedlessParens`
 - `RedundantSemicolons`

saram-aman:
@orpuente-MS please assign the task to me!

PR #1584, opened 2024-05-29
When `@EntryPoint` missing assume `Main` is entry

swernli:
This change updates the entry point logic to assume that in the absence of an explicit `@EntryPoint` attributed callable, a unique `Main` is the entry point.
image

### DIFF
--- a/compiler/qsc_passes/src/entry_point.rs
+++ b/compiler/qsc_passes/src/entry_point.rs
@@ -21,7 +21,7 @@ use thiserror::Error;
 #[derive(Clone, Debug, Diagnostic, Error)]
 pub enum Error {
     #[error("duplicate entry point callable `{0}`")]
-    #[diagnostic(help("only one callable should be annotated with the entry point attribute"))]
+    #[diagnostic(help("only one callable named `Main` or one callable with the `@EntryPoint()` attribute must be present if no entry expression is provided"))]
     #[diagnostic(code("Qsc.EntryPoint.Duplicate"))]
     Duplicate(String, #[label] Span),
 
@@ -34,7 +34,7 @@ pub enum Error {
     BodyMissing(#[label("cannot have specialization implementation")] Span),
 
     #[error("entry point not found")]
-    #[diagnostic(help("a single callable with the `@EntryPoint()` attribute must be present if no entry expression is provided"))]
+    #[diagnostic(help("a single callable with the `@EntryPoint()` attribute must be present if no entry expression is provided and no callable named `Main` is present"))]
     #[diagnostic(code("Qsc.EntryPoint.NotFound"))]
     NotFound,
 }
@@ -124,13 +124,19 @@ fn create_entry_from_callables(
 fn get_callables(package: &Package) -> Vec<(&CallableDecl, LocalItemId)> {
     let mut finder = EntryPointFinder {
         callables: Vec::new(),
+        main: Vec::new(),
     };
     finder.visit_package(package);
-    finder.callables
+    if finder.callables.is_empty() {
+        finder.main
+    } else {
+        finder.callables
+    }
 }
 
 struct EntryPointFinder<'a> {
     callables: Vec<(&'a CallableDecl, LocalItemId)>,
+    main: Vec<(&'a CallableDecl, LocalItemId)>,
 }
 
 impl<'a> Visitor<'a> for EntryPointFinder<'a> {
@@ -138,6 +144,9 @@ impl<'a> Visitor<'a> for EntryPointFinder<'a> {
         if let ItemKind::Callable(callable) = &item.kind {
             if item.attrs.iter().any(|a| a == &Attr::EntryPoint) {
                 self.callables.push((callable, item.id));
+            }
+            if callable.name.name.as_ref().to_lowercase() == "main" {
+                self.main.push((callable, item.id));
             }
         }
     }
--- a/compiler/qsc_passes/src/entry_point.rs
+++ b/compiler/qsc_passes/src/entry_point.rs
@@ -145,7 +145,7 @@ impl<'a> Visitor<'a> for EntryPointFinder<'a> {
             if item.attrs.iter().any(|a| a == &Attr::EntryPoint) {
                 self.callables.push((callable, item.id));
             }
-            if callable.name.name.as_ref().to_lowercase() == "main" {
+            if callable.name.name.as_ref() == "Main" {
                 self.main.push((callable, item.id));
             }
         }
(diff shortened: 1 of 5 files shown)


compiler/qsc_passes/src/entry_point.rs
```diff
@@ -124,13 +124,19 @@ fn create_entry_from_callables(
 fn get_callables(package: &Package) -> Vec<(&CallableDecl, LocalItemId)> {
     let mut finder = EntryPointFinder {
         callables: Vec::new(),
+        main: Vec::new(),
     };
     finder.visit_package(package);
-    finder.callables
+    if finder.callables.is_empty() {
+        finder.main
+    } else {
```
idavis, line 132: If you have something marked as `@EntryPoint` and a main operation, should we issue a lint?

compiler/qsc_passes/src/entry_point.rs
```diff
@@ -139,6 +145,9 @@ impl<'a> Visitor<'a> for EntryPointFinder<'a> {
             if item.attrs.iter().any(|a| a == &Attr::EntryPoint) {
                 self.callables.push((callable, item.id));
             }
+            if callable.name.name.as_ref().to_lowercase() == "main" {
```
idavis, line 148: Should we validate that the callable is an operation?

idavis: If so, we should have another test as the tests in the PR currently always are operations.

compiler/qsc_passes/src/entry_point.rs
```diff
@@ -139,6 +145,9 @@ impl<'a> Visitor<'a> for EntryPointFinder<'a> {
...
+                self.main.push((callable, item.id));
```
idavis, line 149: Should we validate that the main has no params?

swernli: That happens automatically above; this part just collects candidates, the other part checks whether they are unique and valid.

compiler/qsc_passes/src/entry_point.rs
swernli: No, a function can be an entry point today. Granted that entry point can't really do anything interesting as it can't generate any dynamic values and as such is reducible to a constant... but it would be a new restriction that didn't exist before.

compiler/qsc_passes/src/entry_point.rs
swernli: Maybe... that could be considered for follow up. Today if you have an `@EntryPoint` and an entry expression nothing gets flagged. But this is a little more static than that. So I could see a lint warning about ambiguous entry point if there is a callable named `Main` and a separate callable with the attribute.

compiler/qsc_passes/src/entry_point.rs
```diff
@@ -21,7 +21,7 @@ use thiserror::Error;
 #[derive(Clone, Debug, Diagnostic, Error)]
 pub enum Error {
     #[error("duplicate entry point callable `{0}`")]
-    #[diagnostic(help("only one callable should be annotated with the entry point attribute"))]
+    #[diagnostic(help("only one callable named `Main` or one callable with the `@EntryPoint()` attribute must be present if no entry expression is provided"))]
```
DmitryVasilevsky, line 24: Can I have multiple mains if I define the entry point?

compiler/qsc_passes/src/entry_point.rs
minestarks, line 1: It should also be in the `Main` namespace... right?

orpuente-MS: Adding to this. Does it need to be in a file called `Main.qs`?

swernli: I was actually intentionally ignoring those restrictions. While we could use those to try and disambiguate, we already have a mechanism for disambiguation: `@EntryPoint()`. So I see it this way:
- if you have a single callable named `Main` then that is your entry point, regardless of where it is,
- if for some reason you need/want to have more than one callable named `Main`, identify the single entry point via the `@EntryPoint()` attribute.

This sort of matches the way `Main` works in C#. You can have `Main` in any class, as long as there is only one. If you have more than one class with a `Main` method, you have to provide additional configuration to the compiler to identify which one is the one to use as the entry point (see https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/main-command-line).

compiler/qsc_passes/src/entry_point.rs
swernli: Added it as an idea to #1333

compiler/qsc_passes/src/entry_point.rs
billti, line 148: Why the 'to_lowercase'? Is 'mAiN' valid (Or even just `main`)?

compiler/qsc_passes/src/entry_point/tests.rs
```diff
@@ -51,13 +51,43 @@ fn test_entry_point_attr_to_expr() {
...
         "",
+        &expect![[r#"
+            Expr 12 [0-0] [Type Int]: Call:
+                Expr 11 [32-36] [Type Int]: Var: Item 1
+                Expr 10 [36-38] [Type Unit]: Unit"#]],
+    );
+}
+
+#[test]
+fn test_entry_point_attr_missing_implies_main_alernate_casing() {
```
billti, line 69: Looking here, maybe it is allowing any casing. Seems like an odd choice, being we are a case sensitive language and 'opinionated' in much of our coding guidelines. Any reason we don't require `Main` casing?

swernli: I could go either way on this. On the one hand, the fact that items are pascal cased and variables are camel cased is in the style guide but not enforced at compiling. This would be the first place where we are requiring a user follow that guidance. On the other hand, it's more consistent to require it rather than treat all casings of `Main` as equal. It's an easy thing to remove, we just need to decide on a preference for behavior.

swernli: Looking around, it seems both Rust and C# are strongly opinionated on this point and accept only `main` and `Main` respectively, with any other casing being an error. That seems like good enough reason to me to make ours case sensistive too. I'll update.

compiler/qsc_passes/src/entry_point.rs
swernli: Responded in other comment. I'll remove the case insensitivity.

billti (approved):

idavis (approved):

PR #1589, opened 2024-05-30
Add unpublished katas to playground and tests

### LINKED ISSUE #1419: Work in Progress kata list
 I would appreciate a WIP list(Work in Progress) of katas that get tested in CI and local but not get published. That will give us immunity from accidentally publishing katas to the main website.

A `work_in_progress.json` file that builds the katas locally in playground and tests it in CI.

Adding kata to index.json and removing it before merging PR.

**Additional context**
https://github.com/microsoft/qsharp/pull/1408#discussion_r1575595598

tcNickolas:
Unpublished katas are listed after published ones in alphabetic order.

This closes #1419.

### DIFF
--- a/npm/qsharp/generate_katas_content.js
+++ b/npm/qsharp/generate_katas_content.js
@@ -15,7 +15,13 @@
  * Each Kata is organized in a directory where an index.md file provides a description on how the kata must be composed.
  */
 
-import { existsSync, mkdirSync, readFileSync, writeFileSync, readdirSync } from "node:fs";
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  writeFileSync,
+  readdirSync,
+} from "node:fs";
 import { basename, dirname, join, relative, sep } from "node:path";
 import { fileURLToPath } from "node:url";
 
@@ -534,7 +540,14 @@ function tryCreateMarkdownSegment(text) {
   return null;
 }
 
-function createKata(kataPath, id, title, segments, globalCodeSources, published) {
+function createKata(
+  kataPath,
+  id,
+  title,
+  segments,
+  globalCodeSources,
+  published,
+) {
   // Validate that the kata has at least one segment.
   if (segments.length === 0) {
     throw new Error(`Kata '${id}' does not have any segments`);
@@ -578,7 +591,7 @@ function createKata(kataPath, id, title, segments, globalCodeSources, published)
     id,
     title,
     sections,
-    published
+    published,
   };
 }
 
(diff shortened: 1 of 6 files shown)


npm/qsharp/src/katas.ts
```diff
@@ -66,14 +66,17 @@ export type Kata = {
   id: string;
   title: string;
   sections: KataSection[];
+  published: boolean;
 };
 
-export async function getAllKatas(): Promise<Kata[]> {
-  return katasContent.katas as Kata[];
+export async function getAllKatas(includeUnpublished = false): Promise<Kata[]> {
```
minestarks, line 72: In case we add a bunch more optional switches to this method in the future, can we make this a named argument?

```suggestion
export async function getAllKatas({
  includeUnpublished = true,
}): Promise<Kata[]> {
```

And at the call site, it would say: `getAllKatas({ includeUnpublished: true })`


This is just setting ourselves up to avoid code like `getAllKatas(true, true, false, false, true)` down the line.

billti:
I did have an alternate approach for this which I've been meaning to discuss with you, and might be a little cleaner than this. Happy to discuss tomorrow or next week if you have time to see if it might be preferable.

npm/qsharp/generate_katas_content.js
```diff
@@ -649,10 +672,16 @@ function generateKatasContent(katasPath, outputPath) {
     indexPath,
     "Could not read the contents of the katas index file",
   );
-  const katasDirs = tryParseJSON(
-    indexJson,
-    `Invalid katas index at ${indexPath}`,
+  const publishedKatasDirs = Object.values(
+    tryParseJSON(indexJson, `Invalid katas index at ${indexPath}`),
   );
```
minestarks, line 677: Unless I'm missing something, I don't think `Object.values()` is needed here? `index.json` seems to declare a string array, so the type of `publishedKatasDirs` should just be a string array without you having to do anything.

minestarks (approved):

npm/qsharp/generate_katas_content.js
tcNickolas: You're right. Odd, at one point I couldn't make it work without that... Apparently, I changed something afterwards :-)

npm/qsharp/src/katas.ts
tcNickolas: Sounds good, I can do that (I'm assuming you're suggesting to change just the syntax, not the default value from false to true)

swernli (approved):

PR #1592, opened 2024-05-30
Reset zoom level when circuit window is resized

minestarks:
Fixes a minor UI annoyance where the circuit shows up too small due to the narrow window width, and you have to click on the little Zoom button to make it bigger. Now, we resize automatically when the window itself is resized.

window-resize-circuit


npm/qsharp/ux/circuit.tsx
```diff
@@ -83,8 +89,25 @@ function ZoomableCircuit(props: { circuit: qviz.Circuit }) {
...
 
+  /**
+   * Window resize handler to recalculate and set the zoom level
+   * based on the new window width.
+   */
+  function onResize() {
+    const [container, svg] = [circuitDiv.current, currentSvg()];
+    if (container && svg) {
+      // Recalculate the zoom level based on the container width
+      const initialZoom = calculateZoomToFit(container, svg);
```
billti, line 100: Does this mean any zoom level I've already set is lost? i.e. If I've already zoomed in, and then resize the window, it will zoom all the way back out again?

minestarks: Yes, that's what will happen. I convinced myself that this was the desirable behavior. But if we want to remember that the user touched the zoom level, we may want to keep around a specific "user set the zoom level" flag, which only gets cleared if a new circuit is drawn.

minestarks: @billti could you sign off if you're ok with the current version of the PR? We can always improve later :)

billti (approved):

PR #1598, opened 2024-05-31
Reduce katas bundle sizes

billti:
This change removes the duplication of the katas content in the generated bundle. The default bundle now only includes the HTML content (which is what QCOM uses), and generates a separate bundle for the Markdown version (which the main playground page uses - as some point we may just want to remove the Markdown version).

Testing this change with a pending katas PR (which adds some more SVG images), and the generated katas bundle size went from 1.77MB to 1.14MB.

I also extracted the Katas into their own module import, and out of the main "qsharp-lang" package import. This way they won't _accidentally_ get pulled into any bundles just trying to use the compiler (you need to import from `qsharp-lang/katas` now).

I won't merge this until I've coordinated with QCOM on the breaking changes.


playground/src/kata.tsx
```diff
@@ -40,7 +44,7 @@ function ExplainedSolutionElem(props: { solution: ExplainedSolution }) {
               </pre>
             );
           case "text-content":
-            return <Markdown markdown={item.asMarkdown}></Markdown>;
+            return <Markdown markdown={item.content}></Markdown>;
```
cesarzc, line 47: I can't recall, why do we use markdown in the playground instead of html? If the playground could use the same setup as the viewer we could stop producing the markdown version.

cesarzc (approved):

playground/src/kata.tsx
billti: I cover that in the first paragraph of the PR. We may want to look at doing that later.

minestarks (approved):

PR #1599, opened 2024-05-31
Added all teleportation tasks except two

devikamehra:
Added all the tasks of teleportation kata except 1.7 and 1.8

### DIFF
--- a/katas/content/teleportation/Common.qs
+++ b/katas/content/teleportation/Common.qs
@@ -1,4 +1,5 @@
 namespace Kata.Verification {
+    open Microsoft.Quantum.Convert;
     open Microsoft.Quantum.Diagnostics;
     open Microsoft.Quantum.Math;
 
@@ -102,7 +103,9 @@ namespace Kata.Verification {
     // Also repeats for each state several times as
     // code is expected to take different paths each time because
     // measurements done by Alice are not deterministic.
-    operation TeleportTestLoop(teleportOp : ((Qubit, Qubit, Qubit) => Unit)) : Bool {
+    operation TeleportTestLoop(teleportOp : (
+        (Qubit, Qubit, Qubit) => Unit)
+        ) : Bool {
         // Define setup operations for the message qubit
         // on which to test teleportation: |0⟩, |1⟩, |0⟩ + |1⟩, unequal superposition.
         let setupPsiOps = [(I, "|0⟩"), (X, "|1⟩"), (H, "|+⟩"), (Ry(ArcCos(0.6) * 2.0, _), "0.6|0⟩ + 0.8|1⟩")];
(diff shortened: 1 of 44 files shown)


devikamehra:
@tcNickolas For task 1.7, what is the ideal way to add such a task which does not require unit testing? Can I remove the verification in any way or should this be under the example folder? As per my understanding, the example folder contains the code where no intervention from the user is required but this task does require the user to perform changes. Please advise!

For task 1.8, I was able to perform all the steps but getting this error which I am not able to resolve and thus not included this task in index.md of kata.

Exercise "teleportation__entanglement_swapping" has compilation or runtime errors when using the placeholder as solution. Compilation and runtime errors:Message = type error: expected (((Qubit, Qubit) => Int is f?0), ((Qubit, Int) => Unit is f?1)), found (Unit => (((Qubit, Qubit) => Int), ((Qubit, Int) => Unit)))

Error is coming from defining the function prototype in Common.qs

I have the code checked in for this one. Can you please help?

tcNickolas:
For 1.7, we can do such open-ended tasks as examples/demos. We don't have them in the existing katas yet, since this is the first protocol kata being migrated, but the last lesson in qec_shor kata is an example of such approach: we write down the task description in the body of the kata index.md, as it needs to be framed a "lesson" (`@[section]` tag), and then embed the runnable code using `@[example]` tag.

For 1.8, you're passing the operation `Kata.EntanglementSwapping` itself to `TeleportEntanglementSwappingTestLoop`, but you need to pass its return values, so if you add () after `Kata.EntanglementSwapping` it should work.

devikamehra:
> For 1.7, we can do such open-ended tasks as examples/demos. We don't have them in the existing katas yet, since this is the first protocol kata being migrated, but the last lesson in qec_shor kata is an example of such approach: we write down the task description in the body of the kata index.md, as it needs to be framed a "lesson" (`@[section]` tag), and then embed the runnable code using `@[example]` tag.
> 
> For 1.8, you're passing the operation `Kata.EntanglementSwapping` itself to `TeleportEntanglementSwappingTestLoop`, but you need to pass its return values, so if you add () after `Kata.EntanglementSwapping` it should work.

Thank you @tcNickolas! I have updated all the tasks now.

tcNickolas:
Wow, I didn't realize at first that this PR is actually all the remaining tasks of the kata. A pull request of this size is very hard to review and edit, especially since I think there are going to be quite a few edits... Can you please split it in two or three parts of more manageable size, each with logically related tasks? 

A few things I noticed at a glance:
* Putting entanglement swapping before standard teleport testing looks odd, since it's a different algorithm. You can define a separate lesson for it instead
* Common.qs serves as a library for routines that are used across multiple tasks verification; if some operation is used only in one task, it's better to put it in Verification.qs of that file. (Better code organization also simplifies the review, since each operation will be as close to its use as necessary)
* The solution explanations typically need to be a lot more detailed than these, especially for the later tasks. For example, for entangled_trio, the text explains the code of the solution in English, but it doesn't explain the logic of it, the person still needs to do the math themselves. If you're not comfortable writing more detailed explanations, please mark those explanations as TODO and I'll get to them in the final cleanup of the kata
* Please check that you're using proper whitespace everywhere, for example, in if statements before {

Please split this PR and address these high-level comments, and I'll take a closer look at the first of the partial PRs on Monday!
Thank you!

devikamehra:
> Wow, I didn't realize at first that this PR is actually all the remaining tasks of the kata. A pull request of this size is very hard to review and edit, especially since I think there are going to be quite a few edits... Can you please split it in two or three parts of more manageable size, each with logically related tasks?
> 
> A few things I noticed at a glance:
> 
> * Putting entanglement swapping before standard teleport testing looks odd, since it's a different algorithm. You can define a separate lesson for it instead
> * Common.qs serves as a library for routines that are used across multiple tasks verification; if some operation is used only in one task, it's better to put it in Verification.qs of that file. (Better code organization also simplifies the review, since each operation will be as close to its use as necessary)
> * The solution explanations typically need to be a lot more detailed than these, especially for the later tasks. For example, for entangled_trio, the text explains the code of the solution in English, but it doesn't explain the logic of it, the person still needs to do the math themselves. If you're not comfortable writing more detailed explanations, please mark those explanations as TODO and I'll get to them in the final cleanup of the kata
> * Please check that you're using proper whitespace everywhere, for example, in if statements before {
> 
> Please split this PR and address these high-level comments, and I'll take a closer look at the first of the partial PRs on Monday! Thank you!

@tcNickolas Sure Mariia, will break this PR down into 3-4 PRs. For the rest of the comments, let me take a stab at it.

devikamehra:
Duplicate. Created multiple pull requests.

PR #1600, opened 2024-05-31
Fix CodeActions not working on multifile projects

orpuente-MS:
This PR fixes CodeActions not working on multifile projects.

### DIFF
--- a/compiler/qsc_data_structures/src/span.rs
+++ b/compiler/qsc_data_structures/src/span.rs
@@ -17,19 +17,21 @@ pub struct Span {
 }
 
 impl Span {
-    /// Returns true if the position is within the range.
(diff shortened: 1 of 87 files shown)


minestarks (approved):
Does this mean we're returning a "global" `SourceMap` span now, instead of a file offset? If so, maybe consider removing the `source_name` comparison to simplify the code even further. The global `span`s should be all you need to compare, since they're unique across sources.

orpuente-MS:
> Does this mean we're returning a "global" `SourceMap` span now, instead of a file offset? If so, maybe consider removing the `source_name` comparison to simplify the code even further. The global `span`s should be all you need to compare, since they're unique across sources.

Yes, you are right! And I can avoid String comparisons that way.

PR #1602, opened 2024-06-01
Read correct field in QEC scheme

msoeken:
The wrong field is read to calculate the physical number of qubits for a QEC scheme.


swernli (approved):
Are there any test updates that should come with this?

billti (approved):

msoeken:
> Are there any test updates that should come with this?

No test needed to be updated.

PR #1604, opened 2024-06-01
[Unitary Hack] Save Python qsharp-widgets RE output to PNG

### LINKED ISSUE #1063: Add ability to save the Resource Estimator space-time diagram as a file (png/jpg/svg)
This would not work in VS Code but we can consider Jupyter notebook and other scenarios

nirajvenkat:
Submission for Unitary Hack issue: fixes #1063 

This change adds the ability to save the resource estimates diagram locally as a PNG. For this I added a new dependency called `html2canvas` that takes a DOM and renders it to an image. This currently works only on IPython `qsharp-widgets`, could be extended to the VSCode version. 

Additionally we take into account the color theme within VSCode (which is provided in the stylesheet `qsharp-ux.css`) and this is passed in as the background color for rendering. Tested on (dark/light) Modern VSCode themes.

https://github.com/microsoft/qsharp/assets/3744263/898de6d9-fba4-4fea-b729-a3bb698dc51c

~Current issues:~
- ~This button is placeholder, nicer version: https://github.com/microsoft/vscode-notebook-renderers/pull/118~
- ~Saved imaged is slightly truncated at the bottom~

### DIFF
--- a/npm/qsharp/ux/estimatesOverview.tsx
+++ b/npm/qsharp/ux/estimatesOverview.tsx
@@ -5,6 +5,8 @@
 // The results table is also a legend for the scatter chart.
 
 import { useState } from "preact/hooks";
+import { createRef } from "preact";
+import html2canvas from 'html2canvas';
 import { ColorMap } from "./colormap.js";
 import {
   CreateSingleEstimateResult,
@@ -187,6 +189,29 @@ export function EstimatesOverview(props: {
   const [selectedRow, setSelectedRow] = useState<string | null>(null);
   const [selectedPoint, setSelectedPoint] = useState<[number, number]>();
 
+  const printRef = createRef();
+
+  const handleDownloadImage = async () => {
+    const element = printRef.current;
+    const canvas = await html2canvas(element, {
+      backgroundColor: getComputedStyle(element).getPropertyValue("--main-background")
+    });
+
+    const data = canvas.toDataURL('image/png');
+    const link = document.createElement('a');
+
+    if (typeof link.download === 'string') {
+      link.href = data;
+      link.download = 'image.png';
+
+      document.body.appendChild(link);
+      link.click();
+      document.body.removeChild(link);
+    } else {
+      window.open(data);
+    }
+  };
+
   const runNameRenderingError =
     props.runNames != null &&
     props.runNames.length > 0 &&
@@ -306,8 +331,13 @@ export function EstimatesOverview(props: {
         </>
       ) : (
         <>
-          {getResultTable()}
-          {getScatterChart()}
+          <button type="button" onClick={handleDownloadImage}>
+            Save as Image
+          </button>
+          <div ref={printRef}>
+            {getResultTable()}
+            {getScatterChart()}
+          </div>
         </>
       )}
     </div>
--- a/npm/qsharp/ux/estimatesOverview.tsx
+++ b/npm/qsharp/ux/estimatesOverview.tsx
@@ -5,7 +5,6 @@
 // The results table is also a legend for the scatter chart.
 
 import { useState } from "preact/hooks";
-import { createRef } from "preact";
 import { ColorMap } from "./colormap.js";
 import {
   CreateSingleEstimateResult,
@@ -15,7 +14,6 @@ import {
 } from "./data.js";
 import { ResultsTable, Row } from "./resultsTable.js";
 import { Axis, PlotItem, ScatterChart, ScatterSeries } from "./scatterChart.js";
-import { saveToPng } from "./saveImage.js";
 
 const columnNames = [
   "Run name",
@@ -189,27 +187,6 @@ export function EstimatesOverview(props: {
   const [selectedRow, setSelectedRow] = useState<string | null>(null);
   const [selectedPoint, setSelectedPoint] = useState<[number, number]>();
 
-  const printRef = createRef();
-
-  const handleSaveImage = async () => {
-    const element = printRef.current;
-    const backgroundColor =
-      getComputedStyle(element).getPropertyValue("--main-background");
-    const data = await saveToPng(element, backgroundColor);
-
-    const link = document.createElement("a");
-    if (typeof link.download === "string") {
-      link.href = data;
-      link.download = "image.png";
-
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
-    } else {
-      window.open(data);
-    }
-  };
-
   const runNameRenderingError =
     props.runNames != null &&
     props.runNames.length > 0 &&
@@ -329,31 +306,8 @@ export function EstimatesOverview(props: {
         </>
       ) : (
         <>
-          <button
-            role="button"
-            onClick={handleSaveImage}
-            className={"qs-estimatesOverview-saveIcon"}
-            style="position: absolute; top: 5px; right: 5px;"
-          >
-            <span>
-              <svg
-                width="75%"
-                height="75%"
-                viewBox="0 0 16 16"
-                fill="none"
-                xmlns="http://www.w3.org/2000/svg"
-              >
-                <path
-                  className={"qs-estimatesOverview-saveIconSvgPath"}
-                  d="M12.0147 2.8595L13.1397 3.9845L13.25 4.25V12.875L12.875 13.25H3.125L2.75 12.875V3.125L3.125 2.75H11.75L12.0147 2.8595ZM3.5 3.5V12.5H12.5V4.406L11.5947 3.5H10.25V6.5H5V3.5H3.5ZM8 3.5V5.75H9.5V3.5H8Z"
-                />
-              </svg>
-            </span>
-          </button>
-          <div ref={printRef}>
-            {getResultTable()}
-            {getScatterChart()}
-          </div>
+          {getResultTable()}
+          {getScatterChart()}
         </>
       )}
     </div>
--- a/npm/qsharp/ux/estimatesOverview.tsx
+++ b/npm/qsharp/ux/estimatesOverview.tsx
@@ -6,7 +6,6 @@
 
 import { useState } from "preact/hooks";
 import { createRef } from "preact";
-import html2canvas from 'html2canvas';
 import { ColorMap } from "./colormap.js";
 import {
   CreateSingleEstimateResult,
@@ -192,26 +191,9 @@ export function EstimatesOverview(props: {
   const printRef = createRef();
 
   const handleDownloadImage = async () => {
-    const element = printRef.current;
-    const canvas = await html2canvas(element, {
-      backgroundColor: getComputedStyle(element).getPropertyValue("--main-background")
-    });
-
-    const data = canvas.toDataURL('image/png');
-    const link = document.createElement('a');
-
-    if (typeof link.download === 'string') {
-      link.href = data;
-      link.download = 'image.png';
-
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
-    } else {
-      window.open(data);
-    }
-  };
-
+    console.log("DATA");
+  }
+  
   const runNameRenderingError =
     props.runNames != null &&
     props.runNames.length > 0 &&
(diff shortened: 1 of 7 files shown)


nirajvenkat:

billti:
We are _very_ conservative in this project on taking new dependencies. I see you added one called 'html2canvas'. Looking at that project I can see it has over 900 open issues, hasn't had a commit for 2 years, and the README states 

  "_The script is still in a very experimental state, so I don't recommend using it in a production environment nor start building applications with it yet, as there will be still major changes made_".

Based on the above, I'm hesitant to take a change which adds this dependency.

nirajvenkat:
Would it be ok to add a different dependency? [This one](https://github.com/bubkoo/html-to-image) for example, is more active and has less outstanding issues.

Also the way I have been doing development is to first run the watcher using `npm run dev`, make some changes, after which I switch to a jupyter notebook (also in VSCode) and must reload the kernel to get the newest version built by the watcher, then re-run the cell. Is there any way to attach a debugger or to view print logs?

billti:
We use SVG a lot, as well as dynamic DOM nodes (via Preact) for our widgets, so issues such as https://github.com/bubkoo/html-to-image/issues/425, https://github.com/bubkoo/html-to-image/issues/463, and https://github.com/bubkoo/html-to-image/issues/392 could come into play (I've been optimizing some SVG sizes via def/use lately, so likely to be even more of an issue going forward).

Sadly again, the project seems unmaintained. The last release was 18 months ago, and all issues since just have an automated response and no follow up (including this one 🤣 https://github.com/bubkoo/html-to-image/issues/455)

nirajvenkat:
> Also the way I have been doing development is to first run the watcher using npm run dev, make some changes, after which I switch to a jupyter notebook (also in VSCode) and must reload the kernel to get the newest version built by the watcher, then re-run the cell. Is there any way to attach a debugger or to view print logs?

@minestarks @ivanbasov Do you have any pointers on a faster way to work on `qsharp-widgets`?

nirajvenkat:
Fixed prior issues, ready for review.

npm/qsharp/ux/saveImageUtil.tsx
minestarks, line 1: Is the point of this file to render the HTML elements inside a SVG, and then render that inside a Canvas so that we can save it as PNG? 

If that's the case, I'm not sure all this is necessary. We actually just want an image of the ScatterChart image (see scatterChart.tsx),  and that's already a SVG. 

I think it's ok if the "save as image" button only applies to the ScatterChart and only captures that SVG as image. Then hopefully you can remove all this code which renders HTML as a foreignObject. 

Hopefully, not having to take a screenshot of the whole document will also take care of the "extra whitespace on the sides" problem as well.

minestarks (changes requested):
Thank you!! This works great functionally but I have some style / UI requests.

npm/qsharp/ux/estimatesOverview.tsx
```diff
@@ -306,8 +329,31 @@ export function EstimatesOverview(props: {
         </>
       ) : (
         <>
-          {getResultTable()}
-          {getScatterChart()}
+          <button
+            role="button"
+            onClick={handleSaveImage}
+            className={"qs-estimatesOverview-saveIcon"}
+            style="position: absolute; top: 5px; right: 5px;"
```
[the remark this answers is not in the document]
nirajvenkat: 1. Resolved.
2. Removed the extra whitespace by moving button from `estimatesOverview` to the `scatterChart` component, where it is able to reference the inner svg div.

npm/qsharp/ux/estimatesOverview.tsx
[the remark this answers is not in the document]
nirajvenkat: 1. Resolved.
2. Removed the extra whitespace by moving button from estimatesOverview to the scatterChart component, where it is able to reference the inner svg div.

npm/qsharp/ux/saveImageUtil.tsx
```diff
@@ -0,0 +1,72 @@
...
+    .then(encodeURIComponent)
+    .then((html) => `data:image/svg+xml;charset=utf-8,${html}`);
+}
+
+export async function nodeToDataURI(
+  node: HTMLElement,
+  width: number,
+  height: number,
+): Promise<string> {
+  const xmlns = "http://www.w3.org/2000/svg";
```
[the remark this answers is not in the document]
nirajvenkat: Using HTTPS is not possible with this link. I assume that is the proposed fix.

npm/qsharp/ux/saveImageUtil.tsx
nirajvenkat: I am making a test commit to show my progress. We have two paths, SVG and PNG, toggled by `isVSCodeEnv`. Here are my thoughts.

SVG:
image
- Does not suffer from truncation issues, and we do not pass in a background color. But we suffer from styling issues (click the above image for example).
- Works great on regular Jupyter notebook in the browser, but I am running into an issue on VS Code where the button does nothing. I cannot figure out how to attach a debugger to inspect further ([these steps](https://github.com/microsoft/qsharp/pull/1604#issuecomment-2145692207) did not work for me).
- I know saving to SVG is possible, it works well with [this extension](https://github.com/Microsoft/vscode-notebook-renderers), where we get the option to save as SVG or PDF. Not sure how their code works yet. Here is a snippet that uses their extension:
```python
from IPython.display import SVG
SVG(url='https://upload.wikimedia.org/wikipedia/commons/8/84/Example.svg')
```

PNG:
image
- Involves rendering first to a Canvas. But works with all themes and both browser + VS Code.
- Image is getting truncated at the bottom, trying to figure out why. My current workaround is to pass in a padded height into nodeToDataURI. If we use svgToDataURI directly, image gets truncated again.

minestarks: @nirajvenkat thanks for investigating this. Just letting you know I'll get back to you by tomorrow on this so you can finish by the hackathon deadline.

My first instinct is to say "go with PNG and apply the padded height hack", but I want to test a little further. Thanks.

nirajvenkat:
@minestarks Not sure if I can finish up in time for UH submission, so to speed thing up I have proposed commit 6ef794fa396d0faae8b117056a1de59e8cde3f53. This would be saving to PNG, and a simpler API for the UI side that takes in the React RefObject and optionally a filename and does the rest.

npm/qsharp/ux/saveImageUtil.tsx
nirajvenkat: Going to resolve this to allow the automated workflow.

npm/qsharp/ux/scatterChart.tsx
```diff
@@ -105,7 +107,7 @@ export function ScatterChart(props: {
     const centerX = (pointRect.left + pointRect.right) / 2;
     const divRect = topDiv.getBoundingClientRect();
     tooltip.style.left = `${centerX - divRect.left - halfWidth}px`;
-    tooltip.style.top = `${centerY - divRect.top + 12}px`;
+    tooltip.style.top = `${centerY - divRect.top}px`; // + 12}px`;
```
minestarks, line 110: What was this change for?

npm/qsharp/ux/scatterChart.tsx
```diff
@@ -154,6 +156,12 @@ export function ScatterChart(props: {
   }
   const selectedPoint = getSelectedPointData();
 
+  const saveRef = createRef();
+
+  const handleSaveImage = async () => {
+    saveRefToImage(saveRef);
+  };
```
minestarks, line 163: You'll probably want to use `useRef` here instead (as we do in other places) .

```suggestion
  const saveRef = useRef<SVGSVGElement>(null);

  const handleSaveImage = async () => {
    saveToImage(saveRef.current!);
  };
```

Also would recommend calling  `.current` here at the call site, so that `saveToImage()` takes the element instead of the ref. Makes the code boundaries cleaner, IMO. (I renamed `saveRefToImage() -> saveToImage()` to make it clear it doesn't take a ref)

npm/qsharp/ux/scatterChart.tsx
nirajvenkat: Apologies, forgot to revert. Was trying to find a way around the padding hack.

minestarks:
We should be good on the hackathon deadline. This is so close, but it doesn't work in VS Code for me. Probably because VS Code webview doesn't support `window.open`? Feel free to take a stab at implementing proper VS Code extension support (possibly using `showSaveDialog` ?) but that may be tricky and not required for this PR. 

In the meantime please disable it for VS Code (consider taking [this commit to pass an `allowSaveImage` flag ](https://github.com/microsoft/qsharp/commit/dab55c386bcc3708bb6cf12bf3a4fe45d8912630)) . 

Once VS Code is good to go and the other comments are resolved I'll merge this. We're OK on the deadline.

npm/qsharp/ux/scatterChart.tsx
nirajvenkat: I tried `useRef` instead of `createRef` earlier and ran into an error. So I looked around and found [this post](https://stackoverflow.com/a/61638527) which shows how they are related. This is why I chose to use `createRef` which would also avoid type casting which we would need to do otherwise. Quick and ugly way to do this (at the call site or somewhere else:
```
    saveToImage(saveRef.current! as unknown as HTMLElement);
```
We cannot cast directly from `SVGSVGElement` as it does not inherit from `HTMLElement`. However, this code smells so I found an overly complicated route that uses a `DOMParser` to convert to string first, then appends to a newly created `HTMLElement`.

Overall I suggest we continue to use `createRef`, it does not have these issues.

nirajvenkat:
Is it possible for you to merge those changes on your end?

npm/qsharp/ux/estimatesPanel.tsx
```diff
@@ -14,6 +14,7 @@ export function EstimatesPanel(props: {
   runNames: string[];
   calculating: boolean;
   onRowDeleted: (rowId: string) => void;
+  allowSaveImage?: boolean;
```
nirajvenkat, line 17: Here we have a `?` on `allowSaveImage`. Should this be added to `allowSaveImage` in the other files?

nirajvenkat:
@minestarks I was thinking that based on the below setting, it should have been okay for you to commit directly to this branch. Not sure what the convention is.
image
In any case this merge was possible on my end. Tested it out by building VSIX package. The button does not show up at all, is this expected?

npm/qsharp/ux/estimatesPanel.tsx
minestarks: I think this is alright - I like optional properties when there's more than one caller (so that we don't have to update all callers) , but with these components there's really only one caller, so there's not much point to it.

npm/qsharp/ux/scatterChart.tsx
minestarks: Couldn't `saveToImage` just take a `SVGSVGElement` directly? There's really not a big need to make it that generic. I'm still a bit iffy about using `createRef` (personally having a strong bias towards consistency).

I agree about wanting to avoid the cast. If there's no other way to make it work it's ok.

minestarks:
> @minestarks I was thinking that based on the below setting, it should have been okay for you to commit directly to this branch. Not sure what the convention is.

@nirajvenkat sorry, it just didn't occur to me that I could push to your branch. Glad you were able to pull the changes regardless

minestarks (approved):
Please consider the `useRef` suggestion if you can make it work - otherwise looks good. Thanks for the contribution!

npm/qsharp/ux/scatterChart.tsx
nirajvenkat: I don't know of another way, tried casting `SVGSVGElement` to `Element` but it lacks a `style` prop which belongs to `HTMLElement`. This prop is required when we clone the style.

npm/qsharp/ux/saveImageUtil.tsx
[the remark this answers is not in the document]
billti: That's an automated warning. Just ignore it. Obviously doesn't apply when it's just a namespace and not a link.

nirajvenkat:
I am investigating an issue with this branch where the UI is flickering when trying to interact with the points on the graph. Specifically the tooltip that appears when you hover over points keeps flickering. I have a hunch on what it may be but will continue to investigate tomorrow.

PR #1605, opened 2024-06-03
Create template for MarkingOracles kata

tcNickolas:
This includes the testing operation and the four introductory tasks 1.3-1.7 from old DeutschJozsaAlgorithm kata.

### DIFF
--- a/katas/content/marking_oracles/index.md
+++ b/katas/content/marking_oracles/index.md
@@ -5,7 +5,7 @@
     "title": "Overview"
 })
 
-Quantum oracles are a key part of many quantum algorithms that rely on quantum implementation of a classical function. The algorithms' discussions often assume that the quantum oracle that implements the function of interest is provided.
+Quantum oracles are a key part of many quantum algorithms that rely on a quantum implementation of a classical function. The algorithms' discussions often assume that the quantum oracle that implements the function of interest is provided.
 However, in practice implementing a quantum oracle for a specific problem can be quite challenging.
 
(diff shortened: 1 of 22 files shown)


katas/content/marking_oracles/index.md
```diff
@@ -0,0 +1,64 @@
+# Marking Oracles
+
+@[section]({
+    "id": "marking_oracles__overview",
+    "title": "Overview"
+})
+
+Quantum oracles are a key part of many quantum algorithms that rely on quantum implementation of a classical function. The algorithms' discussions often assume that the quantum oracle that implements the function of interest is provided.
```
cesarzc, line 8:
```suggestion
Quantum oracles are a key part of many quantum algorithms that rely on a quantum implementation of a classical function. The algorithms' discussions often assume that the quantum oracle that implements the function of interest is provided.
```

katas/content/marking_oracles/kth_bit/solution.md
```diff
@@ -0,0 +1,6 @@
+Since the effect of this oracle depends only on the value of the $k$-th qubit, we can ignore the rest of the qubits and focus on just `x[k]`. We need to flip the state of the target qubit if the input qubit `x[k]` is in the $\ket{1}$ state and leave it unchanged otherwise - this is exactly the effect of the $CNOT$ gate gate with the qubit `x[k]` used as control and the qubit `y` used as target.
```
cesarzc, line 1:
```suggestion
Since the effect of this oracle depends only on the value of the $k$-th qubit, we can ignore the rest of the qubits and focus on just `x[k]`. We need to flip the state of the target qubit if the input qubit `x[k]` is in the $\ket{1}$ state and leave it unchanged otherwise - this is exactly the effect of the $CNOT$ gate with the qubit `x[k]` used as control and the qubit `y` used as target.
```

cesarzc (approved):

PR #1609, opened 2024-06-05
LS: Q# Structs Support

ScottCarda-MS:
Language Server support for Q# Structs. This also revises some of the hover-card info to ensure that the header tells the user about what kind of identifier is being hovered over and where it is defined.


language_service/src/hover.rs
```diff
@@ -158,12 +188,24 @@ impl<'a> Handler<'a> for HoverGenerator<'a> {
         &mut self,
         path: &'a ast::Path,
         _: &'_ hir::ItemId,
-        _: &'a hir::Package,
+        item: &'a hir::Item,
```
minestarks, line 191: Is there really a need to add this new parameter here? We can easily use the existing `ItemId` passed in right?

```rust
        let item = package
            .items
            .get(item_id.item)
            .expect("item id should exist");
```

Sure it may mean that we're doing one extra lookup, but I think I'd rather trade that off for keeping these signatures clean. They're pretty subtle as they are.

language_service/src/hover.rs
```diff
@@ -158,12 +188,24 @@ impl<'a> Handler<'a> for HoverGenerator<'a> {
...
+        let ns = item
+            .parent
+            .and_then(|parent_id| package.items.get(parent_id))
+            .map_or_else(
+                || Rc::from(""),
+                |parent| match &parent.kind {
+                    hir::ItemKind::Namespace(namespace, _) => namespace.name(),
+                    _ => Rc::from(""),
+                },
+            );
```
minestarks, line 205: Can we refactor this into `get_namespace_name` or something , and make the one in `at_callable_ref` call that too?

language_service/src/hover.rs
```diff
@@ -307,6 +358,16 @@ fn is_param(param_pats: &[&ast::Pat], node_id: ast::NodeId) -> bool {
...
 
+fn is_struct(ty_def: &ast::TyDef) -> bool {
+    match ty_def.kind.as_ref() {
+        ast::TyDefKind::Paren(inner) => is_struct(inner),
+        ast::TyDefKind::Tuple(fields) => fields
+            .iter()
+            .all(|field| matches!(field.kind.as_ref(), ast::TyDefKind::Field(Some(_), _))),
+        ast::TyDefKind::Err | ast::TyDefKind::Field(..) => false,
+    }
+}
```
minestarks, line 369: I kind of expected this to be a method of `ast::TyDef`, just like `ty::Udt::is_struct()` is a method too. Would that make sense?

language_service/src/hover.rs
```diff
@@ -146,8 +146,38 @@ impl<'a> Handler<'a> for HoverGenerator<'a> {
...
+            contents,
+            span: self.range(type_name.span),
+        });
+    }
+
+    fn at_struct_def(
+        &mut self,
+        context: &LocatorContext<'a>,
+        type_name: &'a ast::Ident,
+        def: &'a ast::StructDecl,
```
minestarks, line 172: Love how neatly this fits into the set of existing methods!

minestarks (approved):

language_service/src/hover.rs
ScottCarda-MS: It's just that we already do that exact lookup (and have to do it) prior to calling these functions: https://github.com/microsoft/qsharp/blob/6ff58ff16ee80285522e9856132e186db2cff54c/language_service/src/name_locator.rs#L393

language_service/src/hover.rs
ScottCarda-MS: good idea, done

language_service/src/hover.rs
ScottCarda-MS: Seeing as how all the other features' uses of these functions only need the Item Id, it would be great to eliminate the extra params.

ScottCarda-MS: done

compiler/qsc_doc_gen/src/display.rs
```diff
@@ -335,25 +345,105 @@ struct IdentTyDef<'a> {
...
-            f,
-            "newtype {} = {}",
-            self.ident.name,
-            TyDef { def: self.def }
-        )
+        if let Some(fields) = as_struct(self.def) {
+            write!(f, "struct {} ", self.ident.name)?;
+
+            write!(f, "{{ ")?;
+            if let Some((last, most)) = fields.split_last() {
```
DmitryVasilevsky, line 354: Looks like this split_last() is used not to print that comma after the last element. Is this how this is typically done in Rust? My concern is that the code to represent the element as a string is repeated twice. That way changes to the visual representation will require two changes in the code. Consider looking at alternatives. There's old-fashioned approach (mutable variable to control the printing of the comma) - fast code. There's functional approach via map/join with a lambda - concise code.
Also feel free to ignore this comment here and in all other places.

compiler/qsc_doc_gen/src/display.rs
```diff
@@ -589,6 +692,38 @@ fn eval_functor_expr(expr: &ast::FunctorExpr) -> ty::FunctorSetValue {
     }
 }
 
+fn as_struct(ty_def: &ast::TyDef) -> Option<Vec<ast::FieldDef>> {
+    match ty_def.kind.as_ref() {
+        ast::TyDefKind::Paren(inner) => as_struct(inner),
```
DmitryVasilevsky, line 697: There're some recursion calls to remove parens here. We don't expect enormously large recursion here, do we?

DmitryVasilevsky (approved):

compiler/qsc_doc_gen/src/display.rs
ScottCarda-MS: It's only as large as the user's code is silly 😊.
```qsharp
newtype Foo = ((((First: Int))), Second: Int);
```
This example will recurse over 3+1 calls to this function.

compiler/qsc_doc_gen/src/display.rs
ScottCarda-MS: I wrote a `fmt_brace_seq` function to abstract over this logic (it was used in several places). Note that this function can't replace the similar `display_type_params` (because that needs to return a String instead of write to output) or the `fmt_tuple` (because that needs to support trailing comma for single-element tuples).

PR #1624, opened 2024-06-10
Migrate the KeyDistribution_BB84 kata

tcNickolas:
I call it "Quantum Key Distribution" rather than name after a specific protocol in case we get to adding E91 protocol soon - they're small enough that they don't warrant separate katas for each one. I also trimmed the tasks a bit to streamline the discussion, they looked a bit repetitive in the old kata.


katas/content/key_distribution/Common.qs
```diff
@@ -0,0 +1,31 @@
...
+
+    operation BasisToString(base : Bool) : String {
+        base ? "Hadamard" | "computational"
+    }
+
+    operation StateToString(base : Bool, bit : Bool) : String {
+        if base {  // ∣+⟩ / ∣-⟩
+            return bit ? "|-⟩" | "|+⟩";
+        } else {  // ∣0⟩ / ∣1⟩
+            return bit ? "|1⟩" | "|-⟩";
```
cesarzc, line 17:
```suggestion
            return bit ? "|1⟩" | "|0⟩";
```

cesarzc: My suggestion assumes that the original return string when the bit is not set was unintended.

katas/content/key_distribution/index.md
```diff
@@ -0,0 +1,136 @@
+# Quantum Key Distribution
+
+@[section]({
+    "id": "key_distribution__overview",
+    "title": "Overview"
+})
+
+Quantum key distribution is a type of quantum communication protocols that allow two parties to generate shared secret keys - random strings of bits known only to those two parties. These shared keys can then be used for a variety of different classical cryptographic protocols like encryption or authentication.
```
cesarzc, line 8:
```suggestion
Quantum key distribution is a type of quantum communication protocol that allow two parties to generate shared secret keys - random strings of bits known only to those two parties. These shared keys can then be used for a variety of different classical cryptographic protocols like encryption or authentication.
```

katas/content/key_distribution/index.md
```diff
@@ -0,0 +1,136 @@
...
+  <tr>
+    <td style="text-align:center">Hadamard basis</td>
+    <td style="text-align:center">$\ket{+}$</td>
+    <td style="text-align:center">$\ket{-}$</td>
+  </tr>
+</table>
+
+The bases used in the protocol are selected such that if an eavesdropper tries to measure a qubit in transit and chooses the wrong basis, then they just get a 0 or 1 measurement result with equal probability.
+
+Alice has to make two random choices per qubit she prepares, one for which basis to prepare in, and the other for what bit value she wants to send, choosing each option with $50\%$ probability.
```
cesarzc, line 53:
```suggestion
Alice has to make two random choices for each qubit she prepares, one for which basis to prepare in, and the other for what bit value she wants to send, choosing each option with $50\%$ probability.
```

katas/content/key_distribution/index.md
tcNickolas:
```suggestion
Quantum key distribution is a type of quantum communication protocol that allows two parties to generate shared secret keys - random strings of bits known only to those two parties. These shared keys can then be used for a variety of different classical cryptographic protocols like encryption or authentication.
```

tcNickolas: Good point, but then we need to change the verb too

PR #1627, opened 2024-06-11
Disable needless operation lint for Katas

minestarks:
The katas contain a lot of placeholder code like this:

```qsharp
    operation ZZMeasurement(qs : Qubit[]) : Int {
        // Implement your solution here...

        return -1;
    }
```

And the "operation can be a function instead" lint, while generally helpful, can distract a learner.

In this PR:
- Add the ability to set the lint configuration per language service instance. (Anything in qsharp.json will still override these settings).
- Make needlessOperation "allow" by default (so that we don't need to make a QCOM change)
- Explicitly set needlessOperation to "warn" everywhere but the katas.

### DIFF
--- a/compiler/qsc_linter/src/lints/hir.rs
+++ b/compiler/qsc_linter/src/lints/hir.rs
@@ -12,7 +12,7 @@ use crate::linter::hir::declare_hir_lints;
 use super::lint;
 
 declare_hir_lints! {
-    (NeedlessOperation, LintLevel::Warn, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
+    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
 }
 
 /// Helper to check if an operation has desired operation characteristics
--- a/language_service/src/protocol.rs
+++ b/language_service/src/protocol.rs
@@ -2,15 +2,18 @@
 // Licensed under the MIT License.
 
 use qsc::line_column::Range;
-use qsc::{compile::Error, target::Profile, LanguageFeatures, PackageType};
+use qsc::PackageType;
+use qsc::{compile::Error, target::Profile, LanguageFeatures};
+use qsc_linter::LintConfig;
 use qsc_project::Manifest;
 
 /// A change to the workspace configuration
-#[derive(Clone, Debug, Default, Copy)]
+#[derive(Clone, Debug, Default)]
 pub struct WorkspaceConfigurationUpdate {
     pub target_profile: Option<Profile>,
     pub package_type: Option<PackageType>,
     pub language_features: Option<LanguageFeatures>,
+    pub lints_config: Option<Vec<LintConfig>>,
 }
 
 #[derive(Debug)]
--- a/language_service/src/protocol.rs
+++ b/language_service/src/protocol.rs
@@ -2,10 +2,10 @@
 // Licensed under the MIT License.
 
 use qsc::line_column::Range;
-use qsc::PackageType;
-use qsc::{compile::Error, target::Profile, LanguageFeatures};
-use qsc_linter::LintConfig;
-use qsc_project::Manifest;
+use qsc::{
+    compile::Error, linter::LintConfig, project::Manifest, target::Profile, LanguageFeatures,
+    PackageType,
+};
 
 /// A change to the workspace configuration
 #[derive(Clone, Debug, Default)]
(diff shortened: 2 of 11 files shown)


language_service/src/protocol.rs
```diff
@@ -2,15 +2,18 @@
 // Licensed under the MIT License.
 
 use qsc::line_column::Range;
-use qsc::{compile::Error, target::Profile, LanguageFeatures, PackageType};
+use qsc::PackageType;
+use qsc::{compile::Error, target::Profile, LanguageFeatures};
```
swernli, line 6: nitpick: why did this one `use` get separated from the bunch?

```suggestion
use qsc::{compile::Error, target::Profile, LanguageFeatures, PackageType};
```

language_service/src/protocol.rs
```diff
@@ -2,15 +2,18 @@
...
+use qsc_linter::LintConfig;
```
swernli, line 7: Would it make sense to re-export this via a `pub use` in qsc to keep with the pattern of everything outside of the compiler accessing things only via that crate? I know we haven't been vigilant about keeping that up (case in point, `qsc_project` is directly referenced below, but maybe that is a special case), so feel free to defer, but just thought I'd mention it.

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -12,7 +12,7 @@ use crate::linter::hir::declare_hir_lints;
 use super::lint;
 
 declare_hir_lints! {
-    (NeedlessOperation, LintLevel::Warn, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
+    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
```
swernli, line 15: On the one hand, this seems like an elegant solution to our problem of wanting different behavior in the extension language service (and playground) vs QCOM. But on the other hand, it seems a little unfortunate that now someone needs to look in multiple places to see what the default level of a lint is (here and in the overrides passed by the callers). I'm not quite sure how to reconcile that in my head... we get what we want, but at the cost of a potential maintenance burden in the future whose size I'm unsure of.

Since QCOM still uses `checkCode` could we instead go the other way where this spot stays as a warning and `checkCode` uses the override to turn this (and perhaps other future lints) off for that scenario?

minestarks: QCOM doesn't use `checkCode` for squiggles (though you're right, it does use `checkCode` in some places, and I hadn't even considered that. But I think this change is fine in that regard).

I've resigned myself to just having overrides in multiple places. I don't think we can get around that. What you're suggesting would just shift the overrides to a different spot, but not get rid of them completely.

Mayyybe eventually we can simplify a little bit by having all _these_ hardcoded defaults be the same, e.g. make them all `Allow` (opt-in) or `Warn` (opt-out). Then, at least, when you look at an `update_configuration` call in the source, you can rest assured that's the definitive list for that scenario, and not something to be merged with hardcoded defaults elsewhere.

I can't do that right now though because I'm trying to avoid a QCOM change....

language_service/src/protocol.rs
minestarks: This cheeky little import tried to make a run for it! Don't worry, I caught it and put it back in its braces.

language_service/src/protocol.rs
swernli: Great! Yeah, things have to be public from the underlying qsc_* crates to be visible in qsc, so nothing stops something outside of the compiler folder from importing from there, nor could we reasonably enforce such a limit with anything other than code review and periodic refactoring. Sounds like you already have the types you need for this spot, so no other changes should be needed.

swernli (approved):

ScottCarda-MS (approved):

language_service/src/state/tests.rs
```diff
@@ -1601,6 +1598,89 @@ async fn lints_update_after_manifest_change() {
...
+    let mut updater = new_updater(&received_errors);
+    updater.update_configuration(WorkspaceConfigurationUpdate {
+        lints_config: Some(vec![LintConfig {
+            kind: LintKind::Ast(AstLint::DivisionByZero),
+            level: LintLevel::Warn,
+        }]),
+        ..WorkspaceConfigurationUpdate::default()
+    });
+
+    // Triger a document update.
```
cesarzc, line 1616:
```suggestion
    // Trigger a document update.
```

cesarzc (approved):

PR #1635, opened 2024-06-12
Clean unsupported chars from implicit namespace name

### LINKED ISSUE #1582: Skipping top-level namespace in single file can produce errors if the file isn't saved or has non-Ident chars
When working in a single file that hasn't been saved yet, VS Code gives it a boilerplate name like "untitled1-1" which may include special characters that the `Ident` parser used for namespaces doesn't allow. This can cause that single file to be treated as invalid until it is saved:
image
Likewise, if the single file has characters like spaces in the name after saving to disk it will also fail:
image
This failure seems correct for multi-file, where names of the files in the src folder must conform to the structure of the project when implied namespace names are used. But it would be helpful to special case single file and use some kind of default name if the file name parsing fails.

swernli:
This change adds some simple logic to clean up unsupported characters from file names before they are used as implicit namespace names. It is by no means exhaustive, but at least takes any `-` replaces them with `_`. This will applied for all files, including unsaved ones, so a file saved to disk as `MyCool-File.qs` would become the namespace `MyCool_File` and VS Code's somewhat quirky names for new unsaved file buffers like `untitled:Untitled-1` becomes `Untitled_1`. This allows skipping explicit namespaces in unsaved new files, fixes #1582.

### DIFF
--- a/compiler/qsc_frontend/src/compile.rs
+++ b/compiler/qsc_frontend/src/compile.rs
@@ -602,7 +602,10 @@ pub fn longest_common_prefix<'a>(strs: &'a [&'a str]) -> &'a str {
 }
 
 fn truncate_to_path_separator(prefix: &str) -> &str {
-    let last_separator_index = prefix.rfind('/').or_else(|| prefix.rfind('\\'));
+    let last_separator_index = prefix
+        .rfind('/')
+        .or_else(|| prefix.rfind('\\'))
+        .or_else(|| prefix.rfind(':'));
     if let Some(last_separator_index) = last_separator_index {
         // Return the prefix up to and including the last path separator
         return &prefix[0..=last_separator_index];
--- a/compiler/qsc_frontend/src/compile/tests.rs
+++ b/compiler/qsc_frontend/src/compile/tests.rs
@@ -1330,9 +1330,10 @@ fn implicit_namespace_basic() {
 }
 
 #[test]
-fn reject_bad_filename_implicit_namespace() {
+fn bad_filename_implicit_namespace_best_effort_fixup() {
     let sources = SourceMap::new(
         [
+            // Rejected for starting with number
             (
                 "123Test.qs".into(),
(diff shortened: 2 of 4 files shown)


minestarks:
@swernli In `untitled:Untitled-1`, the `:` isn't part of the file name, it's the URL scheme separator (same as the colon in `http://foo`). To handle that, I think a better fix would be to update `truncate_to_path_separator` so it truncates back to the `:` if it can't find a slash. (Bonus: This would handle the technically valid `C:foo.qs` Windows path too, though I don't know what kind of crazy use case would land you in _that_ scenario).

I think `-` -> `_` is a good conversion. I can see how that's necessary for the Untitled file case.

But otherwise, can we keep it very conservative? I'm not as sold on converting spaces. If we open up the door to more filename collisions, it can get confusing IMO. I don't feel too strongly about this though.

swernli:
> @swernli In `untitled:Untitled-1`, the `:` isn't part of the file name, it's the URL scheme separator (same as the colon in `http://foo`). To handle that, I think a better fix would be to update `truncate_to_path_separator` so it truncates back to the `:` if it can't find a slash. (Bonus: This would handle the technically valid `C:foo.qs` Windows path too, though I don't know what kind of crazy use case would land you in _that_ scenario).
> 
> I think `-` -> `_` is a good conversion. I can see how that's necessary for the Untitled file case.
> 
> But otherwise, can we keep it very conservative? I'm not as sold on converting spaces. If we open up the door to more filename collisions, it can get confusing IMO. I don't feel too strongly about this last point though.

I can make the change to `truncate_to_path_separator`, thanks for pointing to that! Should I also drop the handling for ` ` (space) turning into `-`?

swernli:
I've updated the logic to only convert `-` -> `_` which solves the main case I was concerned with. We'll avoid any other special handling until it's necessary.

minestarks (approved):

PR #1643, opened 2024-06-17
Controlled rotation gates don't check that target and controls are distinct qubits

### LINKED ISSUE #1339: Controlled rotation gates don't check that target and controls are distinct qubits
**Describe the bug**

It is possible to apply Controlled Ry and Rz gates with 2 controls and target that is the same as one of the controls, and have no exception happen.

**To Reproduce**

The smallest repro I could find was:

```qsharp
operation BugDemo() : Unit {
    use qs = Qubit[2];
    ApplyToEach(H, qs);
    Controlled Ry(qs, (0.5, qs[1]));
    ResetAll(qs);
}

BugDemo()
```

Using only one control doesn't trigger the bug (the exception is thrown correctly), and using X instead of Ry also doesn't trigger it.

**Expected behavior**

Controlled gates are not defined for targets that are the same as controls, so I expect an exception for any combination of gates and control/targets that have overlapping controls and targets. (This was the case in Classic QDK too.)

**System information**

- qsharp 1.3.1

swernli:
Fixes #1339


compiler/qsc_eval/src/val.rs
```diff
@@ -381,6 +381,31 @@ impl Value {
             Value::Var(_) => "Var",
         }
     }
+
+    #[must_use]
+    pub fn to_qubits(&self) -> Vec<Qubit> {
```
minestarks, line 386: The name of this made me think it's some kind of conversion operator, which confused me. It's more of a getter. `qubits()` would be a better name IMO (as in `get_qubits`, but Rust doesn't like prefixes in getter names)

minestarks (approved):

compiler/qsc_eval/src/val.rs
swernli: Updated the name and added a comment to help explain the usage.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1181,12 +1182,16 @@ impl<'a> PartialEvaluator<'a> {
...
-        );
+        let (args, ctls_arg) = self
+            .resolve_args(
+                (store_item_id.package, callable_decl.input).into(),
+                args_value.clone(),
+                ctls,
+                fixed_args,
+            )
+            .map_err(|()| {
+                Error::EvaluationFailed("qubis in invocation are not unique".to_string(), args_span)
```
idavis, line 1193: Qubits

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1181,12 +1182,16 @@ impl<'a> PartialEvaluator<'a> {
...
-            fixed_args,
-        );
+        let (args, ctls_arg) = self
+            .resolve_args(
+                (store_item_id.package, callable_decl.input).into(),
+                args_value.clone(),
+                ctls,
+                fixed_args,
+            )
+            .map_err(|()| {
```
idavis, line 1192: Feels odd that we are returning an error of `()` rather than creating a real error kind, then having to map the empty error here

swernli: The alternative is to pass `args_span` through to the underlying `resolve_args` function. I'll try that version to see if the ergonomics are better.

compiler/qsc_partial_eval/src/lib.rs
swernli: Fixed by relying on the same string from evaluator.

PR #1650, opened 2024-06-17
Improve linter documentation

orpuente-MS:
This PR doesn't change any code. It adds a new docstring to improve the experience of a contributor adding a new lint.

It also fixes an error in the crate-level documentation.


ScottCarda-MS (approved):

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -6,6 +6,23 @@ use crate::linter::ast::declare_ast_lints;
 use qsc_ast::ast::{BinOp, ExprKind, Lit, StmtKind};
 use qsc_data_structures::span::Span;
 
+// Read Me:
```
minestarks, line 9: Should these be `///` docstrings?

minestarks (approved):
Thank you for making the codebase more developer friendly!

compiler/qsc_linter/src/lints/ast.rs
orpuente-MS: No, unfortunately I get a compiler warning if I try to make those `///` docstrings.

PR #1654, opened 2024-06-18
Sccarda/struct lints

ScottCarda-MS:


compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -122,3 +121,65 @@ fn precedence(expr: &qsc_ast::ast::Expr) -> u8 {
...
+    }
+}
+
+impl DeprecatedFunctionConstructor<'_> {
+    fn resolve_item_relative_to_user_package(
+        &self,
+        item_id: &hir::ItemId,
+    ) -> (&hir::Item, &hir::Package, hir::ItemId) {
+        self.resolve_item(self.user_package_id, item_id)
+    }
```
orpuente-MS, line 140: This is starting to look more like an HIR-related lint, maybe it should go in the lints/hir.rs file?

ScottCarda-MS: Yes I agree, it is better in the HIR as it save some of the look ups and can reuse infrastructure that other HIR lints have to use.

ScottCarda-MS:
Closed in favor of #1660

PR #1656, opened 2024-06-20
Packages for integration tests are reliably installed

idavis:


build.py
```diff
@@ -271,6 +243,34 @@ def install_qsharp_python_package(cwd, wheelhouse, interpreter):
...
+        command_args = [
+            interpreter,
+            "-m",
+            "pip",
+            "install",
+            requirement,
+            "--only-binary",
+            "qirrunner",
+            "--only-binary",
+            "pyqir",
```
swernli, line 269: nitpick: since I doubt we want any source packages in this part of the build, would it make sense to collapse this to the `:all:` special case?
```suggestion
            "--only-binary",
            ":all:",
```

swernli (approved):

build.py
```diff
@@ -271,6 +243,34 @@ def install_qsharp_python_package(cwd, wheelhouse, interpreter):
...
 
 
+# If any package fails to install when using a requirements file, the entire
+# process will fail with unpredicatable state of installed packages. To avoid
+# this, we install each package individually from the requirements file.
+#
+# The reason we allow failures is that tooling for integration tests may not
+# be available on all platforms, so we don't want to fail the build if we can't
+# run the tests. The CI will run the tests on the platforms where the tooling
+# is available giving us the confidence that the tests pass on those platforms.
```
minestarks, line 253: I may be missing something, but this sounds antithetical to having reliable tests? Which problem are we trying to solve?

minestarks (approved):

PR #1660, opened 2024-06-23
Lints and Code Actions for Deprecated UDT Syntax

ScottCarda-MS:
Expands the linting framework and adds lints for `newtype` declarations, references to UDT constructor functions, use of the `::` field accessor operator, and use of the `w/` and `w/=` operators for structs. A code action is added to convert uses of `w/` and `w/=` operators with structs to copy constructor syntax, and another code action is added to convert uses of `::` to `.`.


compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -29,7 +35,9 @@ use super::lint;
 //  For more details on how to add a lint, please refer to the crate-level documentation
 //  in qsc_linter/lib.rs.
 declare_hir_lints! {
-    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
+    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead"),
+    (DeprecatedFunctionConstructor, LintLevel::Warn, "deprecated function constructors", "function constructors for struct types are deprecated, use `new` instead"),
+    (DeprecatedWithOperator, LintLevel::Warn, "deprecated `w/` and `w/=` operators for structs", "`w/` and `w/=` operators for structs are deprecated, use `new` instead", Option<WithOperatorLint>),
```
orpuente-MS, line 40: As we discussed yesterday, we shouldn't put implementation details of the lints such as `WithOperatorLint` here. This macro is intended to hold metadata about the lints such as their names, default levels and messages.

Fitting this implementation detail into the macro also introduces an extra layer of indirection to access this auxiliary information that is not evident that exists, `self.lint_info`.

ScottCarda-MS: I wanted to use the macro to remove boilerplate code for each of the lints. This change allows the structs to be declared by the macro and includes the necessary `level` field. Only lints that need lint-specific information will have the additional `info` field whose type corresponds to the given last parameter of the macro.

These changes make it easier to write new lints, both those that don't need extra data and those that do. If it is a matter of communicating that in code, the description in the comments can be updated to make that clearer. I also think that this is a smoother experience than requiring the declaration of structs that have a required `level` field, as when they don't have that there are obfuscated error messages due to the macro.

If you feel strongly about this change, then I will defer to you as the code-owner, but I _do_ think that these changes make for a better coding experience and ease of maintenance.

ScottCarda-MS: I've reverted the commit in question, removing the changes to the macro.

language_service/src/state/tests.rs
```diff
@@ -711,6 +711,7 @@ fn notebook_document_lints() {
                                 kind: Ast(
                                     RedundantSemicolons,
                                 ),
+                                code_action_edits: [],
```
minestarks, line 714: Shouldn't these older lints now be updated to return code action edits as well? It's confusing that some of the quickfix implementations are defined along with the lints, and some of them are inline in the language service.

ScottCarda-MS: We'd need to plumb through the Compilation through to the AST lints, as we did with the HIR lints. @orpuente-MS and I had discussed this and wanted to avoid doing that until necessary. I agree that the current way where some code actions come from two different sources is confusing though, so I'd be fine to do that work. Do you think that that would be worth it?

ScottCarda-MS: Actually, strike that. Since these code actions only remove characters, I don't think we'll need to query the source for anything, meaning we don't need the compilation.

ScottCarda-MS: Okay I've wired up the AST lints to also use the `code_action_edits` field.

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -29,7 +35,10 @@ use super::lint;
 //  For more details on how to add a lint, please refer to the crate-level documentation
 //  in qsc_linter/lib.rs.
 declare_hir_lints! {
-    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead")
+    (NeedlessOperation, LintLevel::Allow, "operation does not contain any quantum operations", "this callable can be declared as a function instead"),
+    (DeprecatedFunctionConstructor, LintLevel::Warn, "deprecated function constructors", "function constructors for struct types are deprecated, use `new` instead"),
+    (DeprecatedWithOperator, LintLevel::Warn, "deprecated `w/` and `w/=` operators for structs", "`w/` and `w/=` operators for structs are deprecated, use `new` instead"),
+    (DeprecatedDoubleColonOperator, LintLevel::Warn, "deprecated `::` for field access", "`::` operator is deprecated, use `.` instead"),
```
minestarks, line 41: Admittedly I haven't looked into the implementation yet, but in general, why do these need to be HIR lints? My first thought was that these are syntactic concerns so you should only need the AST?

billti:
Will the linting be 'off' by default? Is this is brand new syntax, seems like we should give people a couple releases to migrate before we start populating the problem list with warnings.

compiler/qsc_linter/src/lints/hir.rs
ScottCarda-MS: The new lints all use the compilation to do lookups for types, name references, or source code. The AST lints don't make use of the compilation, so it is not plumbed through the AST lints.

compiler/qsc_linter/src/lints.rs
```diff
@@ -9,9 +9,20 @@ macro_rules! lint {
         Lint {
             span: $span,
             level: $lint.level,
-            message: $lint.message,
-            help: $lint.help,
-            kind: $lint.kind,
+            message: $lint.message(),
+            help: $lint.help(),
+            kind: $lint.lint_kind(),
+            code_action_edits: vec![],
```
minestarks, line 15: I can kind of see why this is tricky to achieve with the current macros, but it'd be great if we made `code_action_edits()` into a method so that we can lazily compute the edits. I doubt we need to worry about perf any time soon though.

minestarks (approved):

compiler/qsc_linter/src/lints.rs
ScottCarda-MS: The way an edit is constructed can vary wildly from one lint to another, and also require all kinds of various information. I'm not seeing how to make a single function type for this field that doesn't result in doing all the heavy lifting at the linter still anyways and just having a getter function here that takes no args. At that point the complexity isn't worth it.

Maybe you had something different in mind?

ScottCarda-MS:
> Will the linting be 'off' by default? Is this is brand new syntax, seems like we should give people a couple releases to migrate before we start populating the problem list with warnings.

I've set the default lint level to "Allow" for the new lints, effectively disabling them for anyone who doesn't specifically overwrite the lint levels in their project config file.

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -139,3 +163,12 @@ fn precedence(expr: &qsc_ast::ast::Expr) -> u8 {
         _ => 0,
     }
 }
+
+/// Crates a lint for deprecated user-defined types declarations using `newtype`.
```
orpuente-MS, line 167:
```suggestion
/// Creates a lint for deprecated user-defined types declarations using `newtype`.
```

orpuente-MS (approved):
I like how moving the code actions logic to the linter reduced the complexity of retrieving the original source code. As Mine said, at some point in the future we will need to compute the code actions on demand, but we are good for now.

PR #1663, opened 2024-06-24
Project system refactor in preparation for packages work

minestarks:
This is a large refactor of project system code in preparation for the packages work.

This aims to collapse a lot of types across the JS/VSCode/WASM/compiler layers that are similar-but-not-quite the same. I did end up getting rid of a lot of types, but in some places confusion is unavoidable (e.g. `ProjectConfig` vs `ProgramConfig`). I tried to document those places as much as possible.

**Highlights**
- 🎉 Get rid of the deprecated overloads in `ICompiler`
- 🎊 Replace a bunch of separately-passed-in callbacks in the WASM layer with a trait and eliminate a _lot_ of layers to plumb those callbacks through `ProjectLoader`, `LanguageService`, etc. This also eliminates the need for some complex macros and manually-defined types such as `type AsyncFunction<'a, Arg, Return> = Box<dyn Fn(Arg) -> PinnedFuture<Return> + 'a>`
- 🍾 Define a `ProgramConfig` type to group the compiler-related parameters that are commonly passed into the WASM functions, so we don't have to update 1,000,000 spots when we add a new compiler setting, or, say, package support.
- 🙌 Consolidate `FileSystem` and `FileSystemAsync`
- 🥳 The manifest is now consistently parsed in the native layer, rather than the parsing sometimes being done in JS.

Changes by layer:

- `qsc_project`
  - Consolidate `FileSystem` and `FileSystemAsync` implementations by making one call the other. Previously, we had two copies of the implementation, one sync and one async - and they're about to get a lot more complicated with packages work. So I tried an approach where the sync implementation just wraps the `async` one.
  - This also means that the `async` feature is no longer optional in this crate.
  - Define a trait, `JSProjectHost`, to define the manifest/filesystem functions, instead of using a struct with callbacks. This helped get rid of a lot of boxed async function types we were having to store in different places, and replaced them with more intuitive trait impls.
  - The manifest is now being parsed in the native layer instead of JS. Hopefully this will centralize our error handling a bit and make behavior more consistent across scenarios (Python, JS). It also lays the groundwork for package dependencies, where we'll have to parse dependency manifests in the native layer anyway.
  - The project loading code now returns the `name` and `manifest_path`, values which are used in various extension features. Previously these values were being generated JS-side, but it makes sense to populate them at the point we're loading the project.

- `language_service` 
  - Call the new `find_manifest_directory` host function instead of `get_manifest` which used to do the manifest parsing JS-side.
  - Remove the project system callbacks in favor of the unified `JSProjectHost` trait.
  - Remove `LoadManifestResult` in favor of the `Project` type already defined in `qsc_project`
  - Tests: Split out the in-memory FS used in a few different tests, to a shared module `test_fs`.

- `npm`
  - Get rid of the deprecated overloads in `ICompiler`. QCOM has already been updated to use the new signatures.
  - Where applicable, `ICompiler` and `IDebugService` methods all take a consistent `ProgramConfig` type, which contains the sources and compiler settings, which also has a straightforward conversion to the type used in the WASM crate.
  - Verified backcompat with QCOM. (`IDebugService` is a breaking change but it's not used anywhere but VS Code).

- `pip`
  - Only change here is to trivially plumb through the new `resolve_path` filesystem method.

- `vscode`
  - Fix up the various spots where the `ICompiler` methods were updated to take `ProgramConfig`.
  - We don't parse the manifest in the VS Code layer anymore. Instead we just discover it (in `findManifestDirectory`), and the native code does all the rest.

- `wasm`
  - Consolidate parameters such as `sources`, `language_features`, `target_profile` into one `ProgramConfig` type. This is _almost_ the same as the `ProgramConfig` in compiler.ts , except this one is more strictly typed since backcompat isn't a concern.
  - Remove tons of code that was needed make the async JS callbacks work, replacing it with a `wasm-bindgen` attributes where applicable, and using a `trait` instead of separate callbacks.


compiler/qsc_project/src/fs.rs
```diff
@@ -91,4 +91,8 @@ impl FileSystem for StdFs {
             .collect::<Result<_, _>>()
             .map_err(crate::Error::from)?)
     }
+
+    fn resolve_path(&self, base: &Path, path: &Path) -> miette::Result<PathBuf> {
+        Ok(base.join(path))
```
swernli, line 96: Will this leave behind the `..` and `.` that you want to eliminate?

pip/src/interpreter.rs
```diff
@@ -138,6 +139,9 @@ impl Interpreter {
                 list_directory.expect(
                     "file system hooks should have been passed in with a manifest descriptor",
                 ),
+                resolve_path.expect(
```
swernli, line 142: Perhaps these three things that all need to be passed together could be a tuple or struct?

vscode/src/projectSystem.ts
```diff
@@ -104,9 +70,31 @@ async function findManifestDocument(
...
+
+  let listing: { uri: vscode.Uri; content: string } | null = null;
+  try {
+    listing = await readFileUri(potentialManifestLocation);
+  } catch (err) {
+    log.error("Error thrown when reading file: ", err);
+  }
+  return listing;
+}
+
```
swernli, line 86: Could this just become an exists check of some kind, rather than reading the contents?

swernli (approved):
Looks good, a few minor comments!

compiler/qsc_project/src/project.rs
```diff
@@ -135,54 +174,49 @@ pub trait FileSystem {
...
+
+        let fs = ToFileSystemAsync { fs: self };
+
+        // WARNING: This will panic if there are *any* await points in the
+        // load_project implementation. Right now, we know that will never be the case
+        // because we just passed in our synchronous FS functions to the project loader.
+        // Proceed with caution if you make the `FileSystemAsync` implementation any
+        // more complex.
+        FutureExt::now_or_never(fs.load_project(manifest)).expect("load_project should never await")
+    }
```
minestarks, line 194: Looking for @sezna 's blessing and/or comments on this crate in particular - and especially this method. This was the only way I could find to combine the sync/async implementations, but happy to discuss if this is concerning

vscode/src/projectSystem.ts
minestarks: Simplified to not read contents.

pip/src/interpreter.rs
minestarks: I tried to figure out how to make the tuple - or better yet, named parameters - work across the Python interop boundary here, but got lost pretty quickly. I think that'll have to happen another time, maybe when we touch this code again soon for packages.

For the time being, I at least got rid of the triplicated `expect`s by making the pattern matching a little nicer.

compiler/qsc_project/src/fs.rs
minestarks: You're right, I need to do that explicitly here. Fixed.

This has no effect in this PR, but will become relevant later when we add relative project references (e.g. `../MyDepeendency`)

compiler/qsc_project/src/project.rs
```diff
@@ -135,54 +176,49 @@ pub trait FileSystem {
...
+}
 
-        let qs_sources = qs_files.map(|path| self.read_file(&path));
+/// Trivial wrapper to turn a `FileSystem` into a `FileSystemAsync`
+struct ToFileSystemAsync<'a, FS>
+where
+    FS: ?Sized,
```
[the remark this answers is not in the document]
minestarks: I was excited about this suggestion, but sadly it conflicts with another blanket implementation I have in `js.rs`. As far as I know, negative trait bounds aren't a thing, so we can't make these impls mutually exclusive?

https://github.com/microsoft/qsharp/blob/8a8a131b3edef9d1d57b61c05ab375102eb2b7c6/compiler/qsc_project/src/js.rs#L36-L45

PR #1664, opened 2024-06-24
Adding M.Q.Convert.DoubleAsStringWithPrecision function

### LINKED ISSUE #1406: Add M.Q.Convert.DoubleAsStringWithFormat
I print arbitrary floating-point numbers that are arguments to the tests in the katas quite often, and printing them with the default ~15 digits of precision doesn't look particularly good, especially if there are multiple numbers and a narrow space to display the text. Here is the example from the published Single-Qubit Gates kata, task "Prepare Arbitrary State":

image

Bring back the function [M.Q.Convert.DoubleAsStringWithFormat](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.convert.doubleasstringwithformat) that used to be available in the classic QDK.

I could modify some of the tests to use "neat" numbers, but that makes the code much more complicated unnecessarily, and sometimes I need specific numbers that I cannot round to 3-4 digits of precision.

JPark1023:
Resolves #1406


library/std/convert.qs
```diff
@@ -264,4 +264,16 @@ namespace Microsoft.Quantum.Convert {
...
 
+    /// # Summary
+    /// Produces string representation of given double with desired precision
+    ///
+    /// # Input
+    /// ## input
+    /// Double to be converted
+    /// ## precision
+    /// Number of decimal digits
+    function DoubleAsStringWithPrecision(input: Double, precision: Int) : String {
```
msoeken, line 275: The PR title does not match with the function name.

library/std/convert.qs
```diff
@@ -264,4 +264,16 @@ namespace Microsoft.Quantum.Convert {
         );
     }
 
+    /// # Summary
```
tcNickolas, line 267: This change fails the CI check on Q# file formatting.  The error is "Run the formatter with the `--write` option to correct formatting for the above files." - can you give it a try? At a glance, sentences need to end with periods, same as in other API docs.

I would also add an `# Example` section, similar to that for `ResultArrayAsInt`, to clarify the use of the function.

tcNickolas:
We need to include a unit test for the new function. I think a similar test is in compiler\qsc_eval\src\intrinsic\tests.rs, for example, `int_as_double()` test

compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -36,6 +37,15 @@ pub(crate) fn call(
         #[allow(clippy::cast_precision_loss)]
         "IntAsDouble" => Ok(Value::Double(arg.unwrap_int() as f64)),
         "IntAsBigInt" => Ok(Value::BigInt(BigInt::from(arg.unwrap_int()))),
+        "DoubleAsStringWithPrecision" => {
+            let [input, prec_val] = unwrap_tuple(arg);
+            let precision = usize::try_from(prec_val.unwrap_int()).expect("integer value");
```
swernli, line 42: I think this `expect` can fail in practice, since Q# `Int` is signed and nothing prevents a user from passing a negative number to it. You'll need to check for negative values and return the appropriate error before converting to `usize` here (and add a unit test that confirms the behavior).

compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -36,6 +37,15 @@ pub(crate) fn call(
...
+            Ok(Value::String(Rc::from(format!(
+                "{:.*}",
+                precision,
+                input.unwrap_double()
+            ))))
```
swernli, line 47: This may require a little extra handling for round numbers as `input` as well as the `precision` zero case. Since the only thing that differentiates the display of an integer and round numbered double in Q# is the presence of a decimal point, you'll want to make sure that no matter values are passed in the given number always has at least one decimal place. For example, `DoubleAsStringWithPrecision(1.0, 2)` needs to return the string `"1.00"` and `DoubleAsStringWithPrecision(1.1, 0)` should either return an error or always use a single zero decimal, so return `"1.0"`.

swernli (changes requested):
This is a good start, thanks for working on this! I left a few comments, and we'll need to make sure this new function is covered by Rust unit tests as well.

compiler/qsc_eval/src/intrinsic.rs
swernli: It could be, though that would be a new convention for Q#, there isn't anything else that displays a double without something after the `.`.

JPark1023: As-is looks like `DoubleAsStringWithPrecision(1.1, 0)` returns "1". Would it be better to leave it as Rust implements it or return "1.0"?

swernli: I think it at least needs a decimal point to differentiate it from an integer. So either "1." or "1.0" would be preferred.

swernli (commented):
Looks like there are still some unit test failures due to a span change. Otherwise the code is looking good!

library/std/convert.qs
```diff
@@ -264,4 +264,20 @@ namespace Microsoft.Quantum.Convert {
...
+    /// Produces string representation of given double with desired precision (performs rounding).
```
tcNickolas, line 268: Looking at the similar API documents at https://github.com/microsoft/qsharp-runtime/blob/da1872db685826861313ef078acab615e04f1334/src/Simulation/QSharpFoundation/Convert/Convert.qs#L197, I'd suggest to phrase this a bit smoother:
```suggestion
    /// Converts a given double-precision floating-point number to a string representation with desired precision, rounding if required.
```

library/std/convert.qs
```diff
@@ -264,4 +264,20 @@ namespace Microsoft.Quantum.Convert {
...
+    ///
+    /// # Input
+    /// ## input
+    /// Double to be converted.
+    /// ## precision
+    /// Number of decimal digits.
+    /// # Example
+    /// ```qsharp
+    /// Message($"{DoubleAsStringWithPrecision(0.354, 2)}"); // Prints 0.35
```
tcNickolas, line 277: Since the examples are the only source of information for the reader on padding with zeroes and corner cases such as rounding and precision=0, I'd recommend including samples for padding, precision=0, and rounding, one sample each

library/std/convert.qs
```diff
@@ -264,4 +264,20 @@ namespace Microsoft.Quantum.Convert {
...
     }
 
+    /// # Summary
+    /// Produces string representation of given double with desired precision (performs rounding).
+    ///
+    /// # Input
+    /// ## input
+    /// Double to be converted.
+    /// ## precision
+    /// Number of decimal digits.
```
tcNickolas, line 274: It's important to clarify that the number is non-negative, since otherwise the user doesn't know that. (We could, for example, replace last digits before the decimal point with 0 for negative precision)
```suggestion
    /// Non-negative number of digits after the decimal point.
```

tcNickolas (changes requested):
I took a look at the API docs side of things and left some comments. Once you resolve conflicts and make sure all tests pass, we'll take another look, and this should be good to go. 

Thank you!

library/std/convert.qs
JPark1023: As in add those cases to the # Example section?

tcNickolas: Yes.

tcNickolas: You can just add them as lines to the same Q# code snippet

tcNickolas:
Looks like there are Q# formatting issues and failing tests, though I'm not sure why the unrelated tests fail. @swernli Any ideas?

swernli:
> Looks like there are Q# formatting issues and failing tests, though I'm not sure why the unrelated tests fail. @swernli Any ideas?

They are related... the formatting or addition/removal of comments can change the contents of the file and makes the spans different that the test relies on. In addition the Q# file formatting, you can run `env UPDATE_EXPECT=1 cargo test` which will update the expect tests. Take note that sometimes if the input Q# string in the unit test has trailing spaces on any lines, the Rust formatting often removes those and can cause variations in the spans, so make sure those nested strings in the unit tests don't have any trailing spaces.

tcNickolas:
> They are related... the formatting or addition/removal of comments can change the contents of the file and makes the spans different that the test relies on. In addition the Q# file formatting, you can run `env UPDATE_EXPECT=1 cargo test` which will update the expect tests. Take note that sometimes if the input Q# string in the unit test has trailing spaces on any lines, the Rust formatting often removes those and can cause variations in the spans, so make sure those nested strings in the unit tests don't have any trailing spaces.

I don't think this is documented in the repo - this is only of the only two mentions of `UPDATE_EXPECT`, and the other one is in a PR as well. Could you please document this for the contributors unfamiliar with how our tests work (including me :-) )?

swernli:
> I don't think this is documented in the repo - this is only of the only two mentions of `UPDATE_EXPECT`, and the other one is in a PR as well. Could you please document this for the contributors unfamiliar with how our tests work (including me :-) )?

I don't think it makes sense to document this in the repo since it's in the error message from the test failure (see https://github.com/microsoft/qsharp/actions/runs/9899922588/job/27378539797?pr=1664#step:9:1218) and shows up locally when running `cargo test` or `python build.py` as well. It's a feature of the Rust test library we are using ([expect_test](https://docs.rs/expect-test/latest/expect_test/)), so any time an expect test fails it tells users in the test output how to automatically update:
```
  You can update all `expect!` tests by running:
  
      env UPDATE_EXPECT=1 cargo test
  
  To update a single test, place the cursor on `expect` token and use `run` feature of rust-analyzer.
```

JPark1023:
Where do I run `env UPDATE_EXPECT=1 cargo test` ? It doesn't recognise 'env' when running it in the root repo

swernli:
> Where do I run `env UPDATE_EXPECT=1 cargo test` ? It doesn't recognise 'env' when running it in the root repo

`env` is the Linux command line mechanism for setting environment variables. If you are Windows, you'll want to use `set UPDATE_EXPECT=1` followed by `cargo test` instead. Note that on Windows environment variables remain set for the duration of that terminal instance when set this way.

JPark1023:
Couldn't find any trailing spaces in the unit tests and formatted the Q# files but those 2 tests are still failings. I'm not sure where to go from here, any suggestions?

tcNickolas:
Can you push your latest changes to this pull request, so that we are looking at the same code as you are?

library/qs_source/src/std/convert.qs
```diff
@@ -264,5 +264,37 @@ namespace Microsoft.Quantum.Convert {
...
-    export IntAsDouble, IntAsBigInt, ResultAsBool, BoolAsResult, BoolArrayAsInt, IntAsBoolArray, BoolArrayAsBigInt, BigIntAsBoolArray, ResultArrayAsInt, ResultArrayAsBoolArray, BoolArrayAsResultArray, ComplexAsComplexPolar, ComplexPolarAsComplex;
+    /// # Summary
+    /// Converts a given double-precision floating-point number to a string representation with desired precision, rounding if required.
+    ///
+    /// # Input
+    /// ## input
+    /// Double to be converted.
+    /// ## precision
+    /// Non-negative number of digits after the decimal point.
+    /// # Example
```
tcNickolas, line 275: The only theory I have about formatting is adding `///` before the `/// # Example` line. Running formatter from VS Code doesn't do this, but it's the only thing I can see that's different from other lines.

tcNickolas:
When I tried running `cargo run --release --bin qsc_formatter -- ./library/ ./samples/ -r --write`, the change it did was to rearrange the lines in the export statement back into a single line.

library/qs_source/src/std/convert.qs
tcNickolas, line 275: This is not a formatter issue, but I would add an empty line here for readability, like the other operations in this file do
```suggestion
    ///
    /// # Example
```

tcNickolas:
@swernli Can you help us understand the expected behavior of the `env UPDATE_EXPECT=1 cargo test` command? 

When I run it (on Windows, so splitting it in two, `set UPDATE_EXPECT=1` then `cargo test`), it shows me the failing test and the diff between expected and actual, but it doesn't update the compiler\qsc_eval\src\tests.rs file itself. If I change `LocalItemId` from 123 to 124 by hand in the two failing tests, this fixes them, but that's not quite what I expected from "automatically update". 

Is this the expected behavior, or did I set something up wrong?

swernli:
Since I'm on macOS, `env UPDATE_EXPECT=1 cargo test` works as expected from the cmdline. On Windows you can also try using PowerShell, which makes persisting environment variables easier (see [about Environment Variables](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.4#use-the-variable-syntax)). You can also use several alternatives built-into VS Code via the rust-analyzer extension. With your cursor in one of the `expect!` macros, you can use the command palette to bring up the `rust-analyzer: Run` command:
image

It will bring up the same options for running tests that it usually does (running the individual test, running the tests in the module) but with the additional "expect" options automatically set:
image

Selecting one of those will run the test with the option set to regenerate the content to match what is returned. So you can do this on a per-function or per-module. There are some short video walkthroughs of this feature are on the expect-test repo: https://github.com/rust-analyzer/expect-test

@JPark1023, could you try one of these methods and see if it works for you?

swernli (approved):

JPark1023:
> When I tried running `cargo run --release --bin qsc_formatter -- ./library/ ./samples/ -r --write`, the change it did was to rearrange the lines in the export statement back into a single line.

Running `Format Document` with Azure Quantum Development Kit rearranged it with each function being on a separate line. What's the difference between those two ways of formatting?

tcNickolas:
I _suspect_ that "Format Document" comes from the installed extension, the latest released version, and "cargo run" comes from the code in the branch, so the latest code in main. If changes were done to that logic since the last release, the results will change. I haven't tracked that down to confirm, though

PR #1666, opened 2024-06-25
Fix minor formatting issues in Distinguishing States/Unitaries katas

tcNickolas:
Also includes an improvement for flaky test in BB84 kata

### DIFF
--- a/katas/content/key_distribution/random_array/Verification.qs
+++ b/katas/content/key_distribution/random_array/Verification.qs
@@ -6,8 +6,8 @@ namespace Kata.Verification {
     operation CheckSolution() : Bool {
         // The test only checks that the operation returns an array of correct length
         // and that it's not always the same. It doesn't analyze the distribution of true and false elements.
-        let N = 10;
-        let randomArrays = ForEach(Kata.RandomArray, [N, N, N]);
+        let N = 20;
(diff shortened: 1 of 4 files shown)


katas/content/key_distribution/random_array/Verification.qs
```diff
@@ -6,8 +6,8 @@ namespace Kata.Verification {
     operation CheckSolution() : Bool {
         // The test only checks that the operation returns an array of correct length
         // and that it's not always the same. It doesn't analyze the distribution of true and false elements.
-        let N = 10;
-        let randomArrays = ForEach(Kata.RandomArray, [N, N, N]);
+        let N = 20;
+        let randomArrays = ForEach(Kata.RandomArray, [N, size = 10]);
```
billti, line 10: Did you mean to change N from 10 to 20, but then pass `size = 10`.

billti (approved):

katas/content/key_distribution/random_array/Verification.qs
tcNickolas: N is length of individual array, size is the number of arrays we sample. They don't have to be the same.

cesarzc (approved):

PR #1668, opened 2024-06-25
Migrate to 1ES templates

idavis:


.ado/publish.yml
```diff
@@ -16,398 +73,408 @@ variables:
...
+        templateContext:
+          type: releaseJob
+          isProduction: true
+          inputs:  # All input build artifacts must be declared here
+          - input: pipelineArtifact
+            artifactName: VSIX
+            targetPath: $(System.DefaultWorkingDirectory)/target/vscode
+        steps:
+        - script: |
+            npm install -g @vscode/vsce
```
DmitryVasilevsky, line 370: Sounds silly, but I suggest moving this to pre-approval if possible. I think we should have as few steps for actual publishing as possible so that we have fewer failures after approval. I don't see why this should fail, but it would be beneficial to catch failures in the part that executes often.
(Feel free to ignore.)

idavis: This can't be moved as we are on a fresh image for the job and no data from the previous jobs is available except that which is fed as a pipeline artifact.

billti (approved):

swernli (approved):

PR #1669, opened 2024-06-26
Parameter to set maximum number of distillation rounds in default search

msoeken:
This adds a parameter `maxDistillationRounds` to the `constraints` group in resource estimation, which allows to modify the allowed number of distillation rounds in the first search for T factories. The default value is 3. Making this number larger will slow down resource estimation due to the larger exploration space of possible T factories.


resource_estimator/src/system/constants.rs
```diff
@@ -5,10 +5,10 @@
 pub const MAX_CODE_DISTANCE: u64 = 50;
 
 /// Maximum number of distillation rounds
-pub const MAX_DISTILLATION_ROUNDS: usize = 3;
+pub const MAX_DISTILLATION_ROUNDS: u64 = 3;
```
ivanbasov, line 8: Not sure if this is necessary. Usize seems to be more appropriate.

ivanbasov (approved):

resource_estimator/src/system/constants.rs
msoeken: This change increases consistency with other constants. Also, we are using `u64` mostly for job parameters.

PR #1672, opened 2024-06-26
Skip prereqs if already run, and skip wasm checks/install if not needed

billti:
As per description. This should reduce build times slightly, and avoid the wasm-pack failures on aarch64 where it's not needed anyway.

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -118,7 +118,7 @@ extends:
           # build steps
 
           - script: |
-              python build.py --npm --wasm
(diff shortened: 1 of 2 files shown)


minestarks (approved):
Curious how much time this saves in practice?

billti:
> Curious how much time this saves in practice?

Very little. Its primary goal is to fix the flaky ADO failure running wasm-pack checks on aarch64 (where it’s not needed). Any incidental time savings are a bonus.

idavis (approved):

PR #1673, opened 2024-06-27
Add noisy simulator backend to qsharp

orpuente-MS:
This PR does two things:
 - [x] Adds a noisy simulator backend to the qsharp repo.
 - [x] Exposes a thin PyO3 interface around the simulator.


noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,326 @@
...
+    handle_error, instrument::Instrument, kernel::apply_kernel, operation::Operation,
+    ComplexVector, Error, SquareMatrix, TOLERANCE,
+};
+use num_complex::Complex;
+
+/// A vectorized density matrix.
+#[derive(Debug, Clone)]
+pub struct DensityMatrix {
+    /// Dimension of the matrix. E.g.: If the matrix is 5 x 5, then dim is 5.
+    dim: usize,
```
DmitryVasilevsky, line 20: Should we rename this to dimension as other variable names aren't shorted?

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+fn one_qubit() {
+    let h = h();
+    let mz = mz();
+
+    for _ in 0..10 {
+        let mut sim = DensityMatrixSimulator::new(1);
+        sim.apply_operation(&h, &[0])
+            .expect("operation should succeed");
+        let measurement = sim
+            .sample_instrument_with_distribution(&mz, &[0], 0.3)
```
idavis, line 60: Is there something special about 0.3 and 0.7 here?

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    for _ in 0..10 {
+        let mut sim = DensityMatrixSimulator::new(1);
+        sim.apply_operation(&h, &[0])
+            .expect("operation should succeed");
+        let measurement = sim
+            .sample_instrument_with_distribution(&mz, &[0], 0.7)
+            .expect("measurement should succeed");
+        assert_eq!(measurement, 1);
+    }
```
idavis, line 73: We have 20 different assertion failure points. How do I know where my failures are if I only ever see one at a time?

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+#[test]
+fn bell_pair_sampling() {
```
idavis, line 77: If I see a test failure with this name, how does it help me know what was expected, what failed, in what context, so that I can narrow my search to find the bug?

noisy-simulator/src/state_vector_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+    ($left:expr, $right:expr $(,)?) => {
+        if !approx_eq($left, $right) {
+            panic!("aprox_equal failed, left = {}, right = {}", $left, $right);
+        }
+    };
+}
+
+fn approx_eq(a: f64, b: f64) -> bool {
+    (a - b).abs() <= TOLERANCE
+}
```
idavis, line 22: saw this pattern in another file. Push into higher level tests.rs

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,320 @@
...
+
+#[pyclass]
+pub(crate) struct DensityMatrixSimulator(noisy_simulator::DensityMatrixSimulator);
+
+#[pymethods]
+impl DensityMatrixSimulator {
+    #[new]
+    #[pyo3(signature = (number_of_qubits, seed=42))]
+    #[allow(unused_variables)]
+    pub fn new(number_of_qubits: usize, seed: usize) -> Self {
```
idavis, line 150: remove default value, make seen an option, where is seed consumed?

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,320 @@
...
+#[pyclass]
+pub(crate) struct StateVectorSimulator(noisy_simulator::StateVectorSimulator);
+
+#[pymethods]
+impl StateVectorSimulator {
+    #[new]
+    #[pyo3(signature = (number_of_qubits, seed=42))]
+    #[allow(unused_variables)]
+    pub fn new(number_of_qubits: usize, seed: usize) -> Self {
+        Self(noisy_simulator::StateVectorSimulator::new(number_of_qubits))
```
idavis, line 261: Same as above RE seed

Cargo.toml
```diff
@@ -55,6 +55,7 @@ libfuzzer-sys = "0.4"
 log = "0.4"
 miette = "5.10"
 thiserror = "1.0"
+nalgebra = { version = "0.33.0", features = ["rayon"] }
```
idavis, line 58: Do we need all of its default features plus rayon?

pip/src/noisy_simulator.rs
orpuente-MS: This was to be able to use this package as a drop-in replacement for the Cpp simulator. Since we don't have that requirement anymore, I removed `seed`.

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: Yes, you are right.

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+};
+use num_complex::Complex;
+
+macro_rules! assert_approx_eq {
+    ($left:expr, $right:expr $(,)?) => {
+        if !approx_eq($left, $right) {
+            panic!("aprox_equal failed, left = {}, right = {}", $left, $right);
+        }
+    };
+}
```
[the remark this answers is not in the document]
orpuente-MS: The one appeal here is that the panic happens in the same stack frame where the assert is written. I am not sure how beneficial that is, so I changed it to a function.

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+fn mz() -> Instrument {
+    let m0 = operation!([1., 0.;
+                         0., 0.;]);
+    let m1 = operation!([0., 0.;
+                         0., 1.;]);
+    Instrument::new(vec![m0, m1])
+}
+
+#[test]
+fn constructor() {
```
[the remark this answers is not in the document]
orpuente-MS: Removed that test.

noisy-simulator/src/density_matrix_simulator/tests.rs
orpuente-MS: Not really, one should be less than 0.5 and another greater than 0.5.

noisy-simulator/src/density_matrix_simulator/tests.rs
orpuente-MS: Removed the loop and split the test in two.

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,255 @@
...
+    let _sim = DensityMatrixSimulator::new(1);
+}
+
+#[test]
+fn one_qubit() {
+    let h = h();
+    let mz = mz();
+
+    for _ in 0..10 {
```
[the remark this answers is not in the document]
orpuente-MS: Removed the loop.

noisy-simulator/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+nalgebra = { workspace = true }
+num-complex = { workspace = true }
+rand = { workspace = true }
+rayon = { workspace = true }
```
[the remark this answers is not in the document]
orpuente-MS: From rayon's GitHub repo:

> "By default, when building to WebAssembly, Rayon will treat it as any other platform without multithreading support and will fall back to sequential iteration. This allows existing code to compile and run successfully with no changes necessary, but it will run slower as it will only use a single CPU core."

noisy-simulator/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "noisy-simulator"
+version = "0.1.0"
```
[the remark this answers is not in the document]
orpuente-MS: @billti @swernli What are your opinions on this?

Cargo.toml
orpuente-MS: Yes. I just checked this.

noisy-simulator/src/density_matrix_simulator/tests.rs
orpuente-MS: Documented this test.

noisy-simulator/Cargo.toml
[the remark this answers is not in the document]
orpuente-MS: We will do versioning like the compiler.

pip/qsharp/noisy_simulator/_noisy_simulator.pyi
```diff
@@ -0,0 +1,224 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from typing import Any, Callable, ClassVar, Optional, Dict, List
```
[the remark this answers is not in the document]
orpuente-MS: Nice catch, I just removed them.

pip/qsharp/noisy_simulator/__init__.py
```diff
@@ -0,0 +1,15 @@
+from ._noisy_simulator import (
```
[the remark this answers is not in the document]
orpuente-MS: Added

noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    /// Applies the operation matrix to the target qubits.
+    fn apply_operation_matrix(
+        &mut self,
+        operation_matrix: &SquareMatrix,
+        qubits: &[usize],
+    ) -> Result<(), Error> {
+        // TODO [Research]: Figure out why they do this qubits_expanded thing.
+        let mut qubits_expanded = Vec::with_capacity(2 * qubits.len());
+        for id in qubits {
```
idavis, line 148: What was the result of the research on this TODO?

noisy-simulator/src/operation/tests.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use crate::{
+    operation::{operation, Operation},
+    tests::assert_approx_eq,
+    SquareMatrix,
+};
+
+/// Check that the inner matrices of the instrument are constructed correctly.
+#[test]
+fn constructor() {
```
idavis, line 15: What in our code is being tested here?

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+
+#[pymethods]
+impl DensityMatrixSimulator {
+    #[new]
+    pub fn new(number_of_qubits: usize, seed: Option<u64>) -> Self {
+        let mut simulator = noisy_simulator::DensityMatrixSimulator::new(number_of_qubits);
+
+        if let Some(seed) = seed {
+            simulator.set_rng_seed(seed);
+        }
```
idavis, line 176: Should we pass this seed into the new fn instead of calling set_rng_seed manually?

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+
+#[pymethods]
+impl StateVectorSimulator {
+    #[new]
+    pub fn new(number_of_qubits: usize, seed: Option<u64>) -> Self {
+        let mut simulator = noisy_simulator::StateVectorSimulator::new(number_of_qubits);
+
+        if let Some(seed) = seed {
+            simulator.set_rng_seed(seed);
+        }
```
idavis, line 306: Same question as above regarding seed in new call

noisy-simulator/src/operation/tests.rs
orpuente-MS: That the inner matrices of an instrument are constructed correctly. Namely, the "effect" matrix and the "operation" matrix.

noisy-simulator/src/operation.rs
```diff
@@ -0,0 +1,121 @@
...
+        let (dim, _) = kraus_operators
+            .first()
+            .expect("there should be at least one Kraus Operator")
+            .shape();
+
+        let number_of_qubits = (dim as f64).log2() as usize;
+        assert!(
+            1 << number_of_qubits == dim,
+            "kraus operators should have dimensions 2^k x 2^k"
+        );
```
idavis, line 62: From an API perspective, do we want to crash the Python process when bad data is supplied? Should this return `Result<Self, crate::Error>` instead?

noisy-simulator/src/operation.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        for kraus_operator in kraus_operators.iter() {
+            let (rows, cols) = kraus_operator.shape();
+            assert!(
+                rows == dim && cols == dim,
+                "kraus operators should be square matrices and have the same dimensions"
+            );
```
idavis, line 69: same as above regarding error vs panic

noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,346 @@
...
+            trace_change: 1.0,
+            data,
+        }
+    }
+
+    /// Builds a `DensityMatrix` from its raw fields. Returns `None` if
+    ///  the provided args don't represent a valid `DensityMatrix`.
+    ///
+    /// This method is to be used from the PyO3 wrapper.
+    pub fn try_from(
```
idavis, line 47: The try_from pattern uses `Result<T, E>` whereas this is using `Option<T>`.

noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,346 @@
...
+    fn trace(&self) -> f64 {
+        let mut trace: Complex<f64> = Complex::ZERO;
+        for idx in 0..self.dimension {
+            trace += self.data[(self.dimension + 1) * idx];
+        }
+        assert!(
+            trace.im <= TOLERANCE,
+            "state trace is not real, imaginary part is {}",
+            trace.im
+        );
```
idavis, line 111: Why would we crash the user in this case?

noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,346 @@
...
+            if outcome_trace >= TOLERANCE {
+                last_non_zero_trace_outcome = outcome;
+                last_non_zero_trace = outcome_trace;
+            }
+        }
+
+        if summed_probability + TOLERANCE <= random_sample || last_non_zero_trace < TOLERANCE {
+            let err = Error::FailedToSampleInstrumentOutcome;
+            handle_error!(self, err);
+        }
```
idavis, line 286: In many of these cases I see handle_error! used where I would normally expect to see map_err. Was this an intentional design choice?

pip/src/noisy_simulator.rs
orpuente-MS: I am not sure. Do you mean passing an `Option<u64>` to `new`? Rather than having an argument that is `None` some of the times, I would preffer having `fn with_rng_seed(self, seed: u64) -> Self` such that we can do:

```rust
let sim = DensityMatrixSimulator::new(number_of_qubits)
    .with_rng_seed(seed);
```

noisy-simulator/src/density_matrix_simulator.rs
```diff
@@ -0,0 +1,346 @@
...
+        if !new_state.is_normalized() {
+            return Err(Error::InvalidState(format!(
+                "`state` is not normalized, trace is {}",
+                new_state.trace()
+            )));
+        }
+        if !new_state.is_hermitian() {
+            return Err(Error::InvalidState("`state` is not Hermitian".to_string()));
+        }
+        self.state = Ok(new_state);
```
idavis, line 327: Why is state a result?

noisy-simulator/src/operation.rs
```diff
@@ -0,0 +1,121 @@
...
+
+impl Operation {
+    /// Construct an operation from a list of Kraus operators.
+    /// Matrices must be of dimension 2^k x 2^k, where k is an integer.
+    /// Returns `None` if the kraus matrices are ill formed.
+    pub fn new(kraus_operators: Vec<SquareMatrix>) -> Self {
+        let (dim, _) = kraus_operators
+            .first()
+            .expect("there should be at least one Kraus Operator")
+            .shape();
```
idavis, line 56: Same regarding panic with input. Expected bad input shouldn't crash.

noisy-simulator/src/instrument.rs
```diff
@@ -0,0 +1,117 @@
...
+    summed_effect: SquareMatrix,
+    summed_effect_transpose: SquareMatrix,
+    summed_kraus_operators: Vec<SquareMatrix>,
+}
+
+impl Instrument {
+    /// Creates a new instrument.
+    pub fn new(operations: Vec<Operation>) -> Self {
+        let summed_operation: SquareMatrix = operations.iter().map(|op| op.matrix()).sum();
+        let summed_effect: SquareMatrix = operations.iter().map(|op| op.effect_matrix()).sum();
```
idavis, line 25: Might be premature optimization, but 
```
let (summed_operations, summed_effect) = operations.iter().fold((<zero_mat>,<zero_mat>), |(sum_op, sum_eff), op| {
        (sum_op + op.matrix(), sum_eff + op.effect_matrix())
    });
```

noisy-simulator/src/instrument.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+fn summed_kraus_operators(operations: &[Operation]) -> Vec<SquareMatrix> {
+    let choi_matrix: SquareMatrix = operations
+        .iter()
+        .map(|op| {
+            op.kraus_operators()
+                .iter()
+                .map(|m| {
+                    // choi_matrix += vec(K) * vec(K)^dagger
```
idavis, line 80: Is this left over code or a comment of what the code is doing?

noisy-simulator/src/instrument.rs
```diff
@@ -0,0 +1,117 @@
...
+                    let dim = m.shape().0.pow(2);
+                    let mut choi = SquareMatrix::zeros(dim, dim);
+                    for row in 0..dim {
+                        for col in 0..dim {
+                            choi[(row, col)] += m[m.vector_to_matrix_index(col)]
+                                * m[m.vector_to_matrix_index(row)].conj();
```
idavis, line 86: On the surface this calcuation looks like it is doing a lot of extra work that can be cached.

noisy-simulator/src/instrument.rs
```diff
@@ -0,0 +1,117 @@
...
+            let (krows, kcols) = operations[0].kraus_operators()[0].shape();
+            let mut kraus_operator = SquareMatrix::zeros(krows, kcols);
+            let sqrt_eigenvalue = eigenvalue.sqrt();
+            for row in 0..choi_dim {
+                let idx = kraus_operator.vector_to_matrix_index(row);
+                kraus_operator[idx] = sqrt_eigenvalue * eigenvectors[(row, col)];
+            }
+            summed_kraus_operators.push(kraus_operator);
+        } else if *eigenvalue < -TOLERANCE {
+            panic!("failed to decompose Choi matrix, lambda_i = {eigenvalue}");
```
idavis, line 112: Should this be an error rather than crash?

noisy-simulator/src/kernel.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/// Construct a mask that has 1s at locations given by the target `qubits` ids.
+fn make_mask(state: &ComplexVector, qubits: &[usize]) -> usize {
+    // Number of elements in the density matrix.
+    let num_elements = state.len();
+    let mut mask: usize = 0;
+    for id in qubits {
+        let id_mask: usize = 1 << id;
+        assert!(id_mask < num_elements, "invalid qubit id: {id}");
```
idavis, line 87: Do we want to crash or return error?

noisy-simulator/src/operation.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// Return the number of qubits that the operation acts on.
+    pub fn number_of_qubits(&self) -> usize {
+        self.number_of_qubits
+    }
+}
+
+#[cfg(test)]
+pub(crate) use operation;
```
idavis, line 121: In all of our other code the use statemements are at the top of the file.

noisy-simulator/src/state_vector_simulator.rs
```diff
@@ -0,0 +1,359 @@
...
+            trace_change: 1.0,
+            data: state_vector,
+        }
+    }
+
+    /// Builds a `StateVector` from its raw fields. Returns `None` if
+    ///  the provided args don't represent a valid `StateVector`.
+    ///
+    /// This method is to be used from the PyO3 wrapper.
+    pub fn try_from(
```
idavis, line 46: Same as comment on density matrix try_from

noisy-simulator/src/state_vector_simulator.rs
```diff
@@ -0,0 +1,359 @@
...
+                last_non_zero_probability_index = i;
+                if summed_probability > random_sample {
+                    self.data = state_copy;
+                    self.renormalize_with_norm_squared(norm_squared)?;
+                    return Ok(());
+                }
+            }
+        }
+        if summed_probability + TOLERANCE > random_sample && last_non_zero_probability >= TOLERANCE
+        {
```
idavis, line 152: I think cargo fmt needs to be run on this.

noisy-simulator/src/state_vector_simulator.rs
```diff
@@ -0,0 +1,359 @@
...
+        if let Err(err) = self.state.as_mut()?.sample_kraus_operators(
+            operation.kraus_operators(),
+            qubits,
+            renormalization_factor,
+            self.rng.gen(),
+        ) {
+            handle_error!(self, err);
+        };
+
+        Ok(())
```
idavis, line 223: Mentioned in another comment, but I would have expected map_err to handle this.

noisy-simulator/src/operation.rs
orpuente-MS: Yes, nice catch. That was my intention when returning Result from the methods in both simulators but forgot to do the same for `Operation`.

noisy-simulator/src/operation/tests.rs
idavis: Should the test just do that then? It looks like is it doing a lot more than verifying that a macro constructs a matrix properly.

pip/src/noisy_simulator.rs
swernli: an alternative option could be to have `new` stay as is, taking only number of qubits, and a second constructor `new_with_seed` that takes an optional seed. Then the first `new` is just a thin wrapper around `new_with_seed` and majority of callers can avoid having to pass explicit `None`.

noisy-simulator/src/operation.rs
orpuente-MS: Moved it to be just below the macro, since a `pub <MACRO_NAME>;` can't be before the macro declaration.

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: Yes, this is an intentional design choice. Because depending on the kind of error the simulator state becomes invalid, and we need to mutate `self.state` to be that last error, then `return err` from the function.

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: This is an intentional design choice. The simulator state becomes invalid on some cases (when things go wrong numerically). The alternative is to have state be a `DensityMatrix` and have another field called `last_error` that is an `Option<Error>`. But this is prone to errors, because you could accidentally have `last_error` be Some(...) after the simulator has recovered.

noisy-simulator/src/instrument.rs
orpuente-MS: A comment of what the code is doing

noisy-simulator/src/kernel.rs
orpuente-MS: At this point this should never fail. I am returning an error before getting to this point. This should be a panic.

noisy-simulator/src/state_vector_simulator.rs
orpuente-MS: Fixed

noisy-simulator/src/operation.rs
orpuente-MS: Fixed

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: Yes you are right, fixed this

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: Fixed

pip/src/noisy_simulator.rs
orpuente-MS: Yes, I like that alternative, because what I suggest is usually used when you have many optional arguments.

noisy-simulator/src/instrument.rs
orpuente-MS: Make the comment more clear

noisy-simulator/src/state_vector_simulator.rs
orpuente-MS: I tried that but then the formatter that we have on our repo (what formats the code when you hit Ctrl + S) automatically formats it that way each time I save the file.

noisy-simulator/src/density_matrix_simulator/tests.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::DensityMatrixSimulator;
+use crate::tests;
+
+#[test]
+fn measure_0() {
```
cesarzc, line 8: One of the points we talked about offline was naming tests. In general, it's helpful for reviewers and maintainers if tests convey what they are checking in the name, even if it is verbose.

For example, this test could be named something like `measuring_plus_state_yields_zero_with_50_percent_probability`.

noisy-simulator/src/tests.rs
```diff
@@ -0,0 +1,324 @@
...
+
+/// Assert that two f64 are equal up to a `TOLERANCE`.
+pub fn assert_approx_eq(left: f64, right: f64) {
+    if (left - right).abs() > TOLERANCE {
+        panic!("aprox_equal failed, left = {left}, right = {right}");
+    }
+}
+
+/// Returns an H gate.
+fn h() -> Operation {
```
cesarzc, line 19: It seems like the operations defined for these tests are modelling a noisless system. It would be good to add a comment or convey that in the name of the operations because it could be unexpected that the test cases for a noisy simulator use noiseless operations.

noisy-simulator/src/state_vector_simulator/tests.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::StateVectorSimulator;
+use crate::tests;
```
cesarzc, line 5: It seems like all the tests so far for both noisy simulators are done using noiseless operators. It would be good to have a few tests that use operators with a simple noise model (e.g. X gate with a 0.02 probability of not affecting the state or something similar).

noisy-simulator/src/state_vector_simulator/tests.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[test]
+fn measure_1() {
+    tests::measure_1::<StateVectorSimulator>();
+}
+
+#[test]
+fn bell_pair_sampling() {
+    // We perform the test 100 times because of the probabilistic nature of the MZ measurement.
+    for seed in 0..100 {
```
cesarzc, line 20: From what I can tell, the measurement should always yield the same result for the same seed, and in this case in particular the test is only asserting that both qubits are measured in the same state. Given this, what is the reasoning behind testing with different seeds?

noisy-simulator/src/instrument.rs
orpuente-MS: Yes, this should be an error. I just updated it.

noisy-simulator/src/density_matrix_simulator/tests.rs
orpuente-MS: I changed all of the unit tests names to be more descriptive.

noisy-simulator/src/state_vector_simulator/tests.rs
orpuente-MS: Since the MZ has randomness associated with it, after you measure the first qubit you will either see a 0 or a 1 with 50% probability. After that the second measurement should yield the same result as the first one.

If you only perform one test with a single seed, you will either be checking that both are 0 or 1, but not both cases. The reasoning behind multiple seeds is to test both cases.

We actually don't need 100 shots to make sure we are checking both cases. With only 20 shots we will test both scenarios with 99.999905% probability. So, I will lower that number to 20.

noisy-simulator/src/density_matrix_simulator.rs
orpuente-MS: Still haven't figured out

noisy-simulator/src/tests.rs
orpuente-MS: Yes, you are right, I changed those to `noiseless_h`, etc.

noisy-simulator/src/instrument.rs
orpuente-MS: Why is this faster? Is it because it iterates once? Since these iterators usually have less than 10 elements, I don't think it will impact the performance.

Also, I don't know if the improvement would be noticeable compared to the matrix sums that are much slower.

noisy-simulator/src/instrument.rs
orpuente-MS: Do you mean the `m.vector_to_matrix_index(...)` method? That does an integer division and a modulo internally. Do you know if the hash function in a hash-map is cheaper than that?

orpuente-MS: Ohh, or since `m` is a square matrix maybe I could cache it in a constant length array.

orpuente-MS: I just implemented that.

noisy-simulator/src/state_vector_simulator/tests.rs
orpuente-MS: I added a noisy test with a noisy identity test, just to check the noisy nature of the simulator works. Not sure if it is valuable adding more noisy tests for three reasons:
 1. The noisy behavior is already tested with that one test and the math behind the simulator is already tested in the noisy tests.
 2. The noisy tests are really slow to run. The one I added takes ~5s to run on each simulator (~10s in total).

noisy-simulator/src/instrument/tests.rs
```diff
@@ -0,0 +1,61 @@
...
+};
+use nalgebra::ComplexField;
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
+/// Seed for the random number generators.
+const SEED: u64 = 42;
+
+/// Check that the inner matrices of the instrument are constructed correctly.
+#[test]
+fn check_non_slective_operation_matrix_is_computed_correctly() {
```
idavis, line 18: selective

noisy-simulator/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+pub(crate) use handle_error;
+
+pub use {
+    density_matrix_simulator::{DensityMatrix, DensityMatrixSimulator},
+    instrument::Instrument,
+    operation::Operation,
+    state_vector_simulator::{StateVector, StateVectorSimulator},
+};
```
idavis, line 198: These should be at the top

orpuente-MS: Moved to the top, `pub(crate) handle_error` needs to stay there because it needs to be after the macro definition. Else it does not compile.

noisy-simulator/src/state_vector_simulator.rs
orpuente-MS: Explained in the other comment.

pip/qsharp/noisy_simulator/_noisy_simulator.pyi
```diff
@@ -0,0 +1,238 @@
...
+
+class Operation:
+    """
+    EXPERIMENTAL:
+
+    This struct represents a quantum operation. A quantum operation is a linear
+    transformation that maps a valid density matrix to another valid density matrices.
+    """
+
+    def __init__(self, kraus_operators: List[List[List[complex]]]) -> None:
```
billti, line 23: As I mentioned in the meeting, these types give an error when passing a NumPy matrix, even though it does work. No need to fix this PR, but open an issue to fix in a follow up. Thanks.

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,375 @@
...
+
+//! This module is a thin `PyO3` wrapper around the rust `noisy_simulator` crate.
+
+use noisy_simulator::{ComplexVector, NoisySimulator, SquareMatrix};
+use num_complex::Complex;
+use pyo3::{exceptions::PyException, prelude::*};
+type PythonMatrix = Vec<Vec<Complex<f64>>>;
+
+pub(crate) fn register_noisy_simulator_submodule(py: Python, m: &PyModule) -> PyResult<()> {
+    // let m = PyModule::new(py, "noisy_simulator")?;
```
billti, line 12: Delete this commented out line?

pip/src/noisy_simulator.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+    // Move matrix into a linear container.
+    let mut data = Vec::with_capacity(nrows * ncols);
+    for mut row in matrix {
+        data.append(&mut row);
+    }
+    Ok(SquareMatrix::from_row_iterator(nrows, ncols, data))
+}
+
+/// Performance Warning:
```
billti, line 44: Is there any guidance to go with this performance warning (here and above)? I see it used in one place, which seems unavoidable and not on a perf critical path. Is this here to stop folks accidentally using it later in other paths where it might be a concern?

billti (approved):

pip/qsharp/noisy_simulator/_noisy_simulator.pyi
orpuente-MS: Thanks! Will open the issue after merging. I found this, seems there is hope of writing a correct type hint:

https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing

Apparently you can just take a dependency on `numpy.typing` without taking a dependency on the entire `numpy` library.

pip/src/noisy_simulator.rs
orpuente-MS: Yes, this is just used in a non-critical path for performance. But I wanted to avoid any perf accidents in the future.

pip/src/noisy_simulator.rs
orpuente-MS: Deleted.

pip/src/noisy_simulator.rs
orpuente-MS: I will add that to the comment.

billti:
I just noticed the top-level folder name has a dash in (`noisy-simulator`), whereas our others (`language_server`, `resource_estimator`) use underscores. Maybe change this for consistency also @orpuente-MS .

orpuente-MS:
> I just noticed the top-level folder name has a dash in (`noisy-simulator`), whereas our others (`language_server`, `resource_estimator`) use underscores. Maybe change this for consistency also @orpuente-MS .

Thanks, I changed the folder name and also the name of the crate in the Cargo.toml to use underscores.

idavis (approved):

PR #1675, opened 2024-06-27
Improve CI build times

swernli:
This change makes two updates that have major benefit for overall build times:

 - Convert Rust integration tests into unit tests, as this keeps compilation to one executable instead of many.
 - Use cmdline overrides in build.py when running unit tests to avoid LTO.

Between these two changes, the build portion of `cargo test` on Windows drops from ~17 min to ~3 min, and overall build times across all platforms is improved. Note that this does NOT change the optimization levels for builds for other components like WASM and Python, so scenario test and production optimization and binary performance should be unaffected.


build.py
```diff
@@ -214,18 +214,22 @@ def use_python_env(folder):
...
         if build_type == "release":
             cargo_test_args.append("--release")
+            # Disable LTO for release tests to speed up compilation
+            cargo_test_args.append("--config")
+            cargo_test_args.append('profile.release.lto="off"')
         subprocess.run(cargo_test_args, check=True, text=True, cwd=root_dir)
+        step_end()
+
+    step_start("Building the command line compiler")
+    cargo_build_args = ["cargo", "build", "--bin", "qsc"]
```
billti, line 229: Do we need `qci`?

swernli: No, nothing actually uses it in CI or production. At this point it's still around just because some folks like using it locally during dev/test, so they won't let me delete it ;)

billti (approved):

build.py
minestarks: I don't love this middle ground - if the build script doesn't build it, it's effectively dead code.... the people who _don't_ use it are bound to break it for the next person who _does_ use it.

If it's a maintenance headache at this point, I'd be ok deleting it, but that's easy for me to say right now 😄

idavis (approved):

PR #1676, opened 2024-06-27
Package dependency resolution

minestarks:
We now resolve dependencies from `qsharp.json`.

Example manifest:

```json
{
  "dependencies": {
    "MyLocalDep": {
      "path": "../some-local-folder"
    },
    "MyRemoteDep": {
      "github": {
        "owner": "some-github-user",
        "repo": "some-github-repo",
        "ref": "commit-or-tag",
      }
    }
  }
}
```

📦 Packages are recursively resolved and their sources added to the current compilation.

📢 Any errors that occur during package resolution is reported through the language service and extension commands.

📝 For remote code, a special text content provider is added so we can show GitHub content as read-only in the editor during debugging and code navigation.


compiler/qsc_project/src/project.rs
```diff
@@ -164,6 +518,65 @@
...
+pub struct PackageGraphSources {
+    pub root: PackageInfo,
+    pub packages: FxHashMap<PackageKey, PackageInfo>,
+}
+
+impl PackageGraphSources {
+    /// Temporary implementation which just concatenates all sources.
+    /// This will be replaced soon to build and return all the dependencies
+    /// along with the root package sources.
+    /// TODO(packages): delete this method and use the package graph instead
```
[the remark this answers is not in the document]
minestarks: Known, will address in next PR

language_service/src/state.rs
```diff
@@ -262,28 +258,13 @@
 
         let removed_compilation = self.remove_open_document(uri);
 
+        // TODO: consume dependency sources as well
```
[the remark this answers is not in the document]
minestarks: Known, will address in next PR

pip/tests/test_project.py
```diff
@@ -75,6 +87,34 @@
             },
             "qsharp.json": "{}",
         },
+        "with_deps": {
+            "src": {
+                # TODO(packages) When package aliases are taken into account,
```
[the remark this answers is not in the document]
minestarks: Known, will address in next PR

npm/qsharp/src/browser.ts
```diff
@@ -172,7 +173,22 @@ export type {
...
+  type Example,
+  type Exercise,
+  type ExplainedSolution,
+  type ExplainedSolutionItem,
+  type Kata,
+  type KataSection,
+  type Lesson,
+  type LessonItem,
+  type Question,
+} from "./katas.js";
```
swernli, line 190: This seems unrelated, could it be from a merge artifact?

npm/qsharp/src/language-service/language-service.ts
```diff
@@ -117,6 +118,7 @@ export class QSharpLanguageService implements ILanguageService {
       readFile: async () => null,
       listDirectory: async () => [],
       resolvePath: async () => null,
+      fetchGithub: async () => "",
```
swernli, line 121: Note from Mine's past self: would it make sense to provide a default implementation that the playground can also use?

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -144,7 +153,6 @@ export class Compiler implements ICompiler {
       profile ?? "adaptive_ri",
     );
   }
```
swernli, line 155: nitpick: newline is gone?

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -274,3 +296,18 @@ export const compilerProtocol: ServiceProtocol<ICompiler, QscEventData> = {
...
+    ? {
+        root: {
+          sources: program.sources,
+          languageFeatures: program.languageFeatures || [],
+          dependencies: {},
+        },
+        packages: {},
+      }
+    : program.packageGraphSources;
+}
```
swernli, line 313: This appears to be dead code.

```suggestion
```

pip/tests/test_project.py
```diff
@@ -51,11 +51,23 @@ def test_project_unreadable_source(qsharp) -> None:
...
+    assert result == 4
+
+
+def test_project_circular_dependency_error(qsharp) -> None:
+    with pytest.raises(Exception) as excinfo:
+        qsharp.init(project_root="/circular")
+    assert str(excinfo.value).find("Circular dependency detected between") != -1
+
+
 memfs = {
```
swernli, line 66: Would it make sense to have some fake github sources here to test that functionality in Python too?

vscode/qsharp.schema.json
```diff
@@ -33,6 +33,9 @@
           }
         }
       }
+    },
+    "dependencies": {
```
swernli, line 37: There might be more fields that are needed here.

vscode/src/projectSystem.ts
```diff
@@ -132,14 +133,16 @@ async function readFileUri(
...
     // `readFile` should throw the below if the file is not found
     if (
-      !(_err instanceof vscode.FileSystemError && _err.code === "FileNotFound")
+      !(err instanceof vscode.FileSystemError && err.code === "FileNotFound")
     ) {
-      log.error("Unexpected error trying to read file", _err);
+      log.error("Unexpected error trying to read file", err);
+      return null;
+    } else {
+      throw err;
```
swernli, line 144: Maybe it's ok to always throw here? Makes it less likely errors will be missed and require looking at the logs.

vscode/test/suites/language-service/language-service.test.ts
```diff
@@ -174,4 +202,79 @@ suite("Q# Language Service Tests", function suite() {
...
+      vscode.languages.getDiagnostics(projectMainDocUri);
+
+    assert.isEmpty(
+      actualDiagnostics,
+      `Expected no diagnostics, but got ${JSON.stringify(actualDiagnostics)}`,
+    );
+
+    // Sanity check the test setup - is this the correct position?
+    const text = doc.getText(
+      new vscode.Range(new vscode.Position(3, 16), new vscode.Position(3, 26)),
```
swernli, line 221: Looks like these offsets need to get updated to match the below.

language_service/src/compilation.rs
```diff
@@ -90,8 +91,9 @@ impl Compilation {
         Self {
             package_store,
             user_package_id: package_id,
-            errors,
+            compile_errors: errors,
```
swernli, line 94: nitpick (from @ScottCarda-MS): rename `errors` to `compile_errors`

language_service/src/tests.rs
```diff
@@ -185,6 +188,7 @@ async fn document_in_project() {
                             ),
                         ),
                     ],
+                    [],
```
swernli, line 191: Is it worth adding a test case that propagates up a project error?

wasm/src/project_system.rs
```diff
@@ -141,36 +188,110 @@ impl ProjectLoader {
...
+        project_config.try_into()
     }
 }
 
-impl From<ProjectSources> for Vec<(String, String)> {
-    fn from(sources: ProjectSources) -> Self {
-        serde_wasm_bindgen::from_value(sources.into())
-            .expect("sources object should be an array of string pairs")
-    }
+#[allow(clippy::needless_pass_by_value)]
```
swernli, line 208: It looks like this can be removed

swernli (approved):

npm/qsharp/src/language-service/language-service.ts
minestarks: Reply from my current self: Yes, that would be useful, but I don't have the time right now :)

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -74,11 +75,18 @@ export interface ICompiler {
  * Type definition for the configuration of a program.
  * If adding new properties, make them optional to maintain backward compatibility.
  */
-export type ProgramConfig = {
-  /** An array of source objects, each containing a name and contents. */
-  sources: [string, string][];
-  /** An array of language features to be opted in to in this compilation. */
-  languageFeatures?: string[];
+export type ProgramConfig = (
```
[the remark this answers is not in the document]
minestarks: Well, I'd describe it a bit more differently, and I'd say the type _is_ fulfilling its job. It:

- Encapsulates all the program-related parameters so we don't have to add backcompat overloads or _every single function_ that takes a program. We just need to update this type
- Gives us the flexibility to reorder parameters by making everything named (e.g. so we don't end up with code like `compile(undefined, undefined, someparam)` in QCOM

Also, I don't really consider single-file mode "legacy" - for example, the Playground will be single file mode for the foreseeable future. I think there's value in allowing a simpler interface where you can pass in sources directly like this.

I honestly think this type is fine, and it might be TypeScript itself that's bothering you? It's an input object that can either have `sources` or `packageGraphSources` and I think that's just fine.

PR #1678, opened 2024-06-28
Run unit and samples tests in parallel

swernli:
This change tweaks the CI workflow to run `--cli` and `--samples` in a separate, parallel set of tasks, and the remainder in the existing "Build and test" task.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,40 @@ jobs:
       - name: Prereqs
         run: python ./prereqs.py --install
       - name: Build and Test
-        run: python ./build.py --no-check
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -116,7 +116,40 @@ jobs:
       - name: Prereqs
         run: python ./prereqs.py --install
       - name: Build and Test
```
billti, line 118: Maybe we should remove the `and test` from the name of this task?

billti: Oh, I guess it still runs the tests (I always get mixed up with --no-check and --no-test).

billti (approved):

PR #1681, opened 2024-06-28
Optimize playground publish task

swernli:
Skip doing a full build and test for the playground, since merge queues protect main.


billti (approved):

swernli:
Merging with admin privileges since this is a playground publish only change and doesn't affect (or get tested by) any other CI

PR #1682, opened 2024-06-29
Mitigate race condition in completion list triggering

minestarks:
Relevant to upcoming field access completion work - @ScottCarda-MS 

I won't be here to see this PR to completion, so please merge when/if appropriate.


ScottCarda-MS (approved):
It is just adding a 50 ms delay to avoid the race condition.

PR #1687, opened 2024-06-29
Fix needless operation

### LINKED ISSUE #1585: [FollowUp] NeedlessOperation lint cleanup in samples not in list
Fix the samples where the outer callable should be a function because it doesnt manipulate quantum state. See https://github.com/microsoft/qsharp/pull/1579#discussion_r1618402799 for more details

Manvi-Agrawal:
fixes #1585.

### DIFF
--- a/samples/language/LambdaExpression.qs
+++ b/samples/language/LambdaExpression.qs
@@ -9,7 +9,7 @@
 namespace MyQuantumApp {
     open Microsoft.Quantum.Arrays;
     @EntryPoint()
-    operation Main() : Unit {
+    function Main() : Unit {
         // A lambda function is defined with an arrow `->`.
         // The below function takes two inputs and adds them.
         let add = (x, y) -> x + y;
--- a/samples/language/LambdaExpression.qs
+++ b/samples/language/LambdaExpression.qs
@@ -9,7 +9,7 @@
 namespace MyQuantumApp {
     open Microsoft.Quantum.Arrays;
     @EntryPoint()
-    function Main() : Unit {
+    operation Main() : Unit {
         // A lambda function is defined with an arrow `->`.
         // The below function takes two inputs and adds them.
         let add = (x, y) -> x + y;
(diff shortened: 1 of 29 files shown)


DmitryVasilevsky:
I would prefer to limit this PR to operation -> function replacements only. Maybe with additional change to the output strings in one of the functions. I'd prefer other questions not in code comments, but rather in a GitHub issue. @swernli, what do you think?

swernli:
> I would prefer to limit this PR to operation -> function replacements only. Maybe with additional change to the output strings. I'd prefer other questions not in code comments, but rather in a GitHub issue. @swernli, what do you think?

Yes, I agree. The questions in the comments can be removed and the PR focused on cleaning up the existing samples. We've had some debates about reducing some redundant samples, but this PR doesn't seem like the right time to take that on. @Manvi-Agrawal you are definitely asking the right questions though!

DmitryVasilevsky (changes requested):
Please leave only operation -> function changes and output string improvements. Open issues for questions instead of adding comments (or add to existing ones).

swernli (approved):
Thanks for splitting the general samples clean up into separate issues. The focused changes here look good!

samples/language/LambdaExpression.qs
```diff
@@ -9,7 +9,7 @@
 namespace MyQuantumApp {
     open Microsoft.Quantum.Arrays;
     @EntryPoint()
-    operation Main() : Unit {
+    function Main() : Unit {
```
swernli, line 12: Whoops, looks like this one should remain operation.

swernli: Actually, I only noticed this because the CI failed (see https://github.com/microsoft/qsharp/actions/runs/9848645768/job/27191881736). Running `python build.py` with default settings includes a samples check that compiles all the algorithm and language samples, and integration tests verify the notebooks. So thankfully there is some coverage to help with these!

swernli: Oh, I misunderstood your original feedback, sorry about that! You are right, samples do not get checked in CI with qsc for lints, just compilation. That's definitely a gap we should close. I've been working on turning samples tests into Rust unit tests (#1533), and with the incoming package support we have to chance to convert the rest. I think as part of that will be our best bet for getting that coverage.

cesarzc (approved):
Thank you for addressing this lints!

DmitryVasilevsky (approved):

PR #1689, opened 2024-07-01
More precise logical volume calculation

msoeken:
This change allows a more precise logical volume calculation for resource estimation, by not accounting for un-allocated qubits in error correction. This is a feature from the logical counter that is used to compute the space-time overhead from a Q# evaluation.

This feature cannot yet be accessed from the job parameters passed to resource estimation. This feature may cause the logical counter to behave differently for a subset of configurations in the job parameters. So far, the logical counter is only executed once for all configurations.


resource_estimator/src/counts.rs
```diff
@@ -37,6 +37,12 @@ pub struct LogicalCounter {
     /// Global allocation barrier (when calling global barrier this is advanced
     /// to allocate new qubits after the barrier)
     allocation_barrier: usize,
+    /// If set, enables more precise volume computation: in this mode,
+    /// un-allocated qubits do not contribute to the logical volume, as they do
+    /// not need to be error corrected. The vector contains for each qubit the
+    /// data to compute the logical depth collected while the qubit was
+    /// allocated.
+    precise_volume: Option<Vec<(VolumeEntry, VolumeEntry)>>,
```
ivanbasov, line 45: The tuple here hides meanings of its elements. This is especially dangerous because types are the same. Could we please have a struct with clearly named volume entries instead?

swernli:
@msoeken and @ivanbasov is this still under review or needed to merge? It's been open for a while.

msoeken:
@swernli I changed it to draft because some work remains to be done before this can be merged.

swernli:
closing this for now, @msoeken will keep the branch to revive this functionality when needed.

PR #1692, opened 2024-07-01
support imports and exports in the formatter

sezna:
This is one of the changes required for the imports/exports feature work, and this is particularly important for the standard library, which will make extensive use of long export lists and needs to be formatted well.

This PR adds a new context, similar to the existing open/close delim contexts, that is defined by `import` or `export`...`;`. 

If there is a newline after the `import` or `export` keyword, then we format with newlines. Otherwise, puts it all on one line. See the tests for examples.

before:
```qsharp

export
Foo,
Bar,
Baz;
```

after:
```qsharp
export
    Foo,
    Bar,
    Baz;
```
OR, in single-line mode,

```qsharp
export Foo, Bar, Baz;
```

### DIFF
--- a/compiler/qsc_formatter/src/formatter.rs
+++ b/compiler/qsc_formatter/src/formatter.rs
@@ -184,7 +184,7 @@ enum Delimiter {
 
 impl Delimiter {
     /// Constructs a Delimiter from a token, given the current type-parameter state.
-    fn from_concrete_toke_kind(
+    fn from_concrete_token_kind(
         kind: &ConcreteTokenKind,
         type_param_state: TypeParameterListState,
         import_export_state: ImportExportState,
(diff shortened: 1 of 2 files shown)


compiler/qsc_formatter/src/formatter.rs
```diff
@@ -490,10 +526,10 @@ impl<'a> Formatter<'a> {
         }
     }
 
-    /// Updates the type_param_state of the FormatterState based
+    /// Updates the type_param_state and import_export_state of the FormatterState based
     /// on the left and right token kinds. Returns the delimiter
     /// state of the left and right tokens.
-    fn update_type_param_state(
+    fn update_formatter_state(
```
ScottCarda-MS, line 532: We talked about this change a bit in-person, but looking at it now I really don't think this function should handle the update to both of the state fields. Most of the logic in this function still pertains to `TypeParameterListState` and the match statement that you introduced tries to handle both states simultaneously, which is confusing and might be error prone as we maintain this going forward.
From a naming perspective, I would expect `update_formatter_state` to handle updating all of the state of the formatter, but it just handles two of the state fields, which is a bit awkward. Though, I'm not suggesting to merge in the logic to update `indent_level`, `delim_newlines_stack`, and `spec_decl_state` fields also, as I think that function would be way too complex.

My suggestion here is to move the calls to `Delimiter::from_concrete_token_kind` (I just realized that's misspelled in the code) to outside of this function, have this function just updated the type_param_state as before, add a new function that just updates the import_export state:

```qsharp
fn update_import_export_state(&mut self, left_kind: &ConcreteTokenKind) {
    use qsc_frontend::keyword::Keyword;
    use ConcreteTokenKind::*;
    use TokenKind::*;
    match left_kind {
        Syntax(Semi) => {
            self.import_export_state = ImportExportState::NoState;
        }
        Syntax(Keyword(Keyword::Import | Keyword::Export)) => {
            self.import_export_state = ImportExportState::HandlingImportExportStatement;
        }
        _ => (),
    }
}
```

compiler/qsc_formatter/src/formatter.rs
```diff
@@ -490,29 +536,51 @@ impl<'a> Formatter<'a> {
...
-
-        // If we are leaving a type param list, reset the state
-        if matches!(left_kind, Syntax(Gt))
-            && matches!(
-                self.type_param_state,
-                TypeParameterListState::InTypeParamList
-            )
-        {
-            self.type_param_state = TypeParameterListState::NoState
+        // Based on the left-hand-side token, we determine if:
```
ScottCarda-MS, line 569: This comment is no longer correct, now that the import_export logic was separated out. I think this chunk of code could go back to the way it was before, unless you think the match statement is better. Either way, the comment should be corrected.

compiler/qsc_formatter/src/formatter.rs
ScottCarda-MS: Thanks for making that change!

ScottCarda-MS (approved):
Looks great!

PR #1694, opened 2024-07-02
Skip commenting memory/perf benches on draft PRs

sezna:
This should help reduce some of the comment noise until we have time to customize that action some more.

I tested that this works on this PR -- the benches are not run when the PR is a draft.

I had to pull the two relevant jobs out into a different workflow file, because the triggers are different. For the reporting benches stage (the memory and perf comments that show up on PRs), they'll only run if the PR is in a draft. But, and this is the different part that required them being in a separate file, they also need to trigger when a PR is marked "ready for review".

### DIFF
--- a/.github/workflows/bench-reports.yml
+++ b/.github/workflows/bench-reports.yml
@@ -1,4 +1,4 @@
-name: CI Build and Test
+name: Benchmark Reports
 
 on:
   pull_request:


.github/workflows/bench-reports.yml
```diff
@@ -0,0 +1,90 @@
+name: CI Build and Test
```
swernli, line 1: This should use a different name to differentiate between this and the normal CI Build and Test.

swernli:
It looks like the tests for the npm package got stuck for the last two hours on macos-14... @billti have you seen anything like this before?

swernli (approved):

idavis (approved):

PR #1698, opened 2024-07-03
Q# Package System

### LINKED ISSUE #883: Initial Q# Package System
# Q# Package System

## Definitions
* **package**: an exported or imported Q# project. 
* **package system**: the canonical and conventional way to export or import Q# projects, as packages.

Note the above constrained definitions -- we are not trying to build a registry or robust SemVer-based system at this point. 

## Objectives
* Unblock code-sharing for library authors.
* Provide alternative canonical Microsoft libraries that are not the standard library
* Provide a conventional system that supports the common use case of sharing publicly available code
* Don't immediately support private libraries, that can be a follow-on task
* Be able to construct a complete list of sources and fetch them before compilation begins

## User Stories
* As a hardware partner, I would like to publish an algorithm, as a Q# library, to showcase my hardware's capabilities.
* As a Q# author, I would like to import code without having to copy-paste it into my project.

## Potential Design
We initially only support shared code on GitHub. We use the GitHub REST API, which supports listing directories and cloning subfolders. This helps us avoid `libgit2`, which is impermissively licensed, and any native git implementations which may not work on all of our scenarios.

To import a Q# package, the user will add an entry to their manifest file's `dependencies` property. This will look similar to the `dependencies` property in a package.json. [We could even use their GitHub URL shorthand.](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#github-urls), or we could just use GitHub URLs.

We will browse to that URL, look for a `qsharp.json` in the specified subfolder, or the root if no subfolder is specified, and parse the dependencies of that package. We will continue this process, browsing to them and parsing their manifests, with all dependencies until we have a complete graph of direct dependencies and transitive dependencies. At this stage, we will load all of these sources into memory and begin compilation as normal. In non-web scenarios, we may be able to cache these dependencies to disk, but in web scenarios, they may be cached in the state of the language service.

Related: #73 
```[tasklist]
### Tasks
- [ ] Tests: test dependency ordering edge cases with different language features enabled, etc
- [ ] Tests: test that entry points in libraries are not used
- [x] wire up all scenarios
- [x] work out TODOs on alex's branch
- [ ] telemetry
- [ ] Load transitive deps for qsc
- [ ] Load transitive deps for qsi
- [x] go-to-def and breaking _into_ things don't work
- [ ] https://github.com/microsoft/qsharp/pull/1651#discussion_r1646331561 change this test
```

sezna:
This PR:

1. Refactors the compiler API such that the caller passes in a constructed package store, allowing for dependencies to be pre-compiled and inserted before compiling user code
2. Introduces the `BuildableProgram` abstraction, which encapsulates the notion of "some user code that is ready to compile, along with a package store that has all of its dependencies compiled"
3. As a drive-by fix, fixes circuits for internal operations
4. Removes the notion of `Visibility` from the AST, since visibility is calculated via exports now. `internal` is still parsed without error for backwards compatibility, but is a no-op because `internal` is the default now.
5. Enforces visibility in the HIR, where items must be `Visibility::Public` to show up across packages as a `GlobalItem`
6. Adds export statements to the standard library to preserve the existing API with the new internal-by-default semantics
7. Inserts all packages into the user code's namespace tree with the prefix defined as the package alias in the `qsharp.json`

closes #883 

The Q# formatting CI stage will fail until #1692 goes in

This PR supersedes #1693

### DIFF
--- a/pip/tests/test_project.py
+++ b/pip/tests/test_project.py
@@ -94,13 +94,13 @@ memfs = {
         },
         "compile_error": {
             "src": {
-                "test.qs": "namespace Test { operation ReturnsFour() : Int { 4.0 } export ReturnsFour; }",
+                "test.qs": "namespace Test { operation ReturnsFour() : Int { 4.0 } }",
             },
             "qsharp.json": "{}",
         },
         "with_deps": {
             "src": {
-                "test.qs": "namespace Test { operation CallsDependency() : Int { return Foo.Test.ReturnsFour(); } export CallsDependency; }",
+                "test.qs": "namespace Test { operation CallsDependency() : Int { return Foo.Test.ReturnsFour(); } }",
             },
             "qsharp.json": """
                 {
@@ -146,17 +146,15 @@ memfs = {
 
 
 def fetch_github_test(owner: str, repo: str, ref: str, path: str):
-    if (owner, repo, ref, path) == ("test-owner", "test-repo", "12345", "/qsharp.json"):
-        return """{ "files" : ["src/test.qs"] }"""
-    elif (owner, repo, ref, path) == (
-        "test-owner",
-        "test-repo",
-        "12345",
-        "/src/test.qs",
-    ):
-        return "namespace Test { operation ReturnsTwelve() : Int { 12 }; export ReturnsTwelve; }"
-    else:
-        raise Exception(f"Unexpected fetch_github call: {owner}, {repo}, {ref}, {path}")
+    match (owner, repo, ref, path):
+        case ("test-owner", "test-repo", "12345", "/qsharp.json"):
+            return """{ "files" : ["src/test.qs"] }"""
+        case ("test-owner", "test-repo", "12345", "/src/test.qs"):
+            return "namespace Test { operation ReturnsTwelve() : Int { 12 } export ReturnsTwelve;}"
+        case _:
+            raise Exception(
+                f"Unexpected fetch_github call: {owner}, {repo}, {ref}, {path}"
+            )
 
 
 def read_file_memfs(path):


compiler/qsc_frontend/src/compile/tests.rs
```diff
@@ -962,7 +968,7 @@ fn unimplemented_call_from_dependency_produces_error() {
         [
```
swernli, line 968: This doesn't seem right, but revising `@Unimplemented` relative to exports seems out of scope, especially since it is not used anywhere today (or even documented). Can you either comment out the test attribute (or skip) and then file a follow up issue on me?

language_service/src/state/tests.rs
```diff
@@ -1848,8 +1848,8 @@ async fn error_from_dependency_reported() {
                                     ExpectedItem(
                                         Ident,
                                         Span {
-                                            lo: 26,
-                                            hi: 26,
+                                            lo: 0,
```
swernli, line 1851: This seems odd, might be worth digging into why spans from dependencies are lost.

compiler/qsc_linter/src/linter/ast.rs
```diff
@@ -62,7 +62,6 @@ pub(crate) trait AstLintPass {
     fn check_stmt(&self, _stmt: &Stmt, _buffer: &mut Vec<Lint>) {}
     fn check_ty(&self, _ty: &Ty, _buffer: &mut Vec<Lint>) {}
     fn check_ty_def(&self, _ty_def: &TyDef, _buffer: &mut Vec<Lint>) {}
```
swernli, line 64: It occurs to me that one side effect of removing visibility from the AST is that we won't be able to add a lint warning in the future to tell people it's unnecessary and should be removed. Is the possibility of a future lint enough of a reason to keep it in?

pip/tests/test_project.py
```diff
@@ -150,6 +146,27 @@ def test_github_dependency(qsharp) -> None:
 
 
 def fetch_github_test(owner: str, repo: str, ref: str, path: str):
+<<<<<<< HEAD
```
swernli, line 149: Looks like a merge conflict got mistakenly checked in here?

samples/language/DemoProject/qsharp.json
```diff
@@ -0,0 +1,14 @@
+{
+    "dependencies": {
+        "Foo": {
+            "path": "../DependencyA"
+        },
+        "GitHub": {
+            "github": {
+                "owner": "sezna",
+                "repo": "qsharp-project-template",
+                "ref": "b4575725349f01183c98858c6ce8342cf490c436"
```
swernli, line 10: do we want to check this in with a dependency to a personal repo? Would it make sense to have it in this repo, or does that get weird?

samples/language/DemoProject/src/Main.qs
```diff
@@ -0,0 +1,27 @@
...
+/// # Description
+/// Organizing code into multiple Q# source files is an important part of
+/// writing readable and maintainable code. In this project, we have `Main.qs`,
+/// and `Particle.qs`, which defines a new namespace for particle operations.
+/// The presence of a Q# manifest file (`qsharp.json`) tells the compiler
+/// to include all Q# files under `src/`.
+namespace MyQuantumApp {
+    open Particle;
+    @EntryPoint()
+    operation Main() : Unit {
```
swernli, line 13: This sample triggers the "operation should be a function" lint, so probably worth making it `function` to avoid confusion.

wasm/src/project_system.rs
```diff
@@ -355,25 +356,50 @@ impl From<PackageInfo> for qsc_project::PackageInfo {
...
     let capabilities = qsc::target::Profile::from_str(&program.profile())
         .unwrap_or_else(|()| panic!("Invalid target : {}", program.profile()))
         .into();
 
     let pkg_graph: PackageGraphSources = program.packageGraphSources().into();
     let pkg_graph: qsc_project::PackageGraphSources = pkg_graph.into();
 
-    let (sources, language_features) = pkg_graph.into_sources_temporary();
+    // This builds all the dependencies
+    let buildable_program = BuildableProgram::new(capabilities, pkg_graph);
```
swernli, line 382: It struck me how usefly this comment is to remind the reader that `BuildableProgram::new` actually does quite a bit of compiling... at the very least it's worth putting a description into the doc comment on `new` so folks can get that reminder on hover. As food for thought for future work: is there another name besides `new` that can convey how heavyweight this is without being too different from idomatic Rust?

wasm/src/language_service.rs
```diff
@@ -29,6 +29,7 @@ impl LanguageService {
         LanguageService(qsls::LanguageService::new(Encoding::Utf16))
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub fn start_background_work(
```
swernli, line 33: This looks like it's leftover from an earlier iteration, it can be removed.

wasm/src/diagnostic.rs
```diff
@@ -51,9 +51,9 @@ serializable_type! {
     /// in the editor or to populate the Problems view in VS Code.
     QSharpError,
     {
-        document: String,
-        diagnostic: VSDiagnostic,
-        stack: Option<String>,
+        pub document: String,
+        pub diagnostic: VSDiagnostic,
+        pub stack: Option<String>,
```
swernli, line 56: this change doesn't seem required and can be dropped.

vscode/test/suites/language-service/language-service.test.ts
```diff
@@ -204,12 +217,11 @@ suite("Q# Language Service Tests", function suite() {
   });
 
   test("Package dependencies", async () => {
-    const doc = await vscode.workspace.openTextDocument(projectMainDocUri);
-    vscode.commands.executeCommand("workbench.action.problems.focus");
+    const doc = await vscode.workspace.openTextDocument(mainPackageMainQs);
```
swernli, line 220: If the removed `executeCommand("workbench.action.problems.focus")` wasn't needed here, is it still needed in the test case below (line 262)?

library/qs_source/core/core.qs
```diff
@@ -56,4 +56,7 @@ namespace Microsoft.Quantum.Core {
 
         output
     }
+
+    export Length,
+    Repeated;
```
swernli, line 61: I think once your formatting change arrives this might need to be indented. Also, is it worth considering moving these to the top of the namespaces so when the file is opened fresh someone can see all the exports? That might get distracting for the longer lists, but would still be informative. Your call.

library/qs_source/src/std/core.qs
```diff
@@ -106,4 +106,6 @@ namespace Microsoft.Quantum.Core {
         }
         return true;
     }
+    export RangeStart, RangeEnd, RangeStep, RangeReverse, IsRangeEmpty;
```
swernli, line 109: nitpick: newline before this rather than after.

library/qs_source/src/std/internal.qs
```diff
@@ -247,4 +247,5 @@ namespace Microsoft.Quantum.Intrinsic {
 
         targets
     }
+    export CH, CCH, ApplyGlobalPhase, ControllableGlobalPhase, GlobalPhase, CRz, CS, CT, MapPauli, EntangleForJointMeasure, CollectControls, AdjustForSingleControl, AND, PhaseCCX, CCZ, CCY, CRxx, CRyy, CRzz, IndicesOfNonIdentity, RemovePauliI, SpreadZ;
```
swernli, line 250: nitpick: missing newline

library/qs_source/src/std/diagnostics.qs
```diff
@@ -149,4 +149,6 @@ namespace Microsoft.Quantum.Diagnostics {
         areEqual
     }
 
+    export DumpMachine, DumpRegister, CheckZero, CheckAllZero, Fact, CheckOperationsAreEqual;
```
swernli, line 152: nitpick: extra newline

library/qs_source/core/core.qs
swernli: Actually, never mind on moving these to the top. I realize now the exports are effectively for the purpose of making the callables "public" and later we'll have a std somewhere that re-exports things.

swernli (commented):
Finished reviewing, changes look good, a few minor things called out. The only major thing is a possible side effect of removing visibility from the AST precluding future lints, but even that can be something we add back in later in a limited fashion if it makes sense to do so.

swernli (approved):

compiler/qsc/src/bin/qsc.rs
```diff
@@ -95,64 +96,91 @@ enum Emit {
...
-        if let Some(qsharp_json) = cli.qsharp_json {
-            if let Some(dir) = qsharp_json.parent() {
-                let project = match fs.load_project(dir, None) {
-                    Ok(project) => project,
-                    Err(errs) => {
-                        for e in errs {
-                            eprintln!("{e:?}");
-                        }
-                        return Ok(ExitCode::FAILURE);
+    let (mut store, dependencies, source_map) = if let Some(qsharp_json) = cli.qsharp_json {
```
cesarzc, line 112: This block of logic grew considerably in this PR. It would be worth breaking it down in different functions so it can be more easily seen where `store`, `dependencies` and `source_map` come from.

compiler/qsc/src/bin/qsi.rs
```diff
@@ -81,54 +85,91 @@ impl Receiver for TerminalReceiver {
...
-        if let Some(qsharp_json) = cli.qsharp_json {
-            if let Some(dir) = qsharp_json.parent() {
-                let project = match fs.load_project(dir, None) {
-                    Ok(project) => project,
-                    Err(errs) => {
-                        for e in errs {
-                            eprintln!("{e:?}");
-                        }
-                        return Ok(ExitCode::FAILURE);
+    let (store, dependencies, source_map) = if let Some(qsharp_json) = cli.qsharp_json {
```
cesarzc, line 93: Similarly to my other comment on `qsc`, it would be worth breaking down the logic here into multiple fucntions such that it's easier to see where the values of `store`, `dependencies` and `source_map` come from.

compiler/qsc/src/packages/tests.rs
```diff
@@ -0,0 +1,291 @@
...
+        .root
+        .dependencies
+        .insert("NonExistent".into(), "nonexistent-dep-key".into());
+
+    let buildable_program = super::prepare_package_store(
+        TargetCapabilityFlags::default(),
+        program.package_graph_sources,
+    );
+
+    expect![[r"
```
cesarzc, line 128: This test is called `missing_dependency`, but it doesn't seem to assert any errors so I think I'm not getting something. How can I see the missing dependency in this test?

compiler/qsc_project/src/project.rs
```diff
@@ -593,19 +593,77 @@ pub struct PackageGraphSources {
...
-        (sources, self.root.language_features)
+        let mut sorted_packages = self.packages.into_iter().collect::<Vec<_>>();
+        sorted_packages.sort_by_key(|(a_key, _pkg)| {
+            sorted_keys
+                .iter()
+                .position(|key| key.as_str() == &**a_key)
+                .unwrap_or_else(|| panic!("package {a_key} should be in sorted keys list"))
+        });
+
+        log::debug!("build plan: {:#?}", sorted_keys);
```
cesarzc, line 649: Was this log macro intentionally left behind?

compiler/qsc_frontend/src/compile.rs
```diff
@@ -175,6 +183,9 @@ pub type SourceName = Arc<str>;
 
 pub type SourceContents = Arc<str>;
 
+// the arc<str> is only `None` for the legacy stdlib, core, and an interpreter special case
```
cesarzc, line 186: Are the stdlib and core considered legacy after these changes? What makes them legacy?

compiler/qsc_passes/src/capabilitiesck/tests_common.rs
```diff
@@ -49,12 +49,16 @@
 
 impl CompilationContext {
     fn new(source: &str) -> Self {
+        let mut store = qsc::PackageStore::new(qsc::compile::core());
+        // TODO(alex) ask cesarzc if std is wanted here
```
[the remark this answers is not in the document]
cesarzc: Looking at the code, std is used in these tests so I would be surprised if the tests passed without it.

katas/src/lib.rs
```diff
@@ -28,13 +28,18 @@ pub fn check_solution(
     receiver: &mut impl Receiver,
 ) -> Result<bool, Vec<Error>> {
     let source_map = SourceMap::new(exercise_sources, Some(EXERCISE_ENTRY.into()));
+    let mut store = qsc::PackageStore::new(qsc::compile::core());
```
cesarzc, line 31: After seeing this code repeated multiple times, I wonder if it would make sense to have utility functions like `qsc::PackageStore::with_core_lib` and `qsc::PackageStore::with_std_lib` to reduce the boilerplate code a bit.

cesarzc (approved):
I finished reviewing and the changes look good. I left a few comments with mostly non-blocking things, but it would be good to address them before merging.

PR #1700, opened 2024-07-05
Added summary section in chemistry.py

devikamehra:
Added summary table for multiple qubit parameters in chemistry.py

### DIFF
--- a/samples/estimation/df-chemistry/chemistry.py
+++ b/samples/estimation/df-chemistry/chemistry.py
@@ -516,11 +516,6 @@ res = qsharp.estimate(
 with open("resource_estimate.json", "w") as f:
     f.write(res.json)
 
-if isinstance(res,dict):
-    result = [res]
-else:
-    result = res
-
 data = {
     'Run name': [],
     'T factory fraction': [],
@@ -529,7 +524,12 @@ data = {
     'rQOPS': []
 }
 
-for item in result:
+result_obj = json.loads(res.json)
+
+if isinstance(result_obj,dict):
+    result_obj = [ result_obj ]
+
+for item in result_obj:
     # Run name
     data['Run name'].append(item["jobParams"]["qubitParams"]["name"])
     # T factory fraction and Runtime
@@ -552,18 +552,9 @@ for item in result:
     else:
         data['Physical qubits'].append("-")
         data['rQOPS'].append("-")
-       
-data = {
-    'Run name': ['qubit_maj_ns_e6'],
-    'T factory fraction': ['15.26 %'],
-    'Runtime': ['19 mins'],
-    'Physical qubits': [207604],
-    'rQOPS': [343333334]
-}
  
 df = pd.DataFrame(data)
-pd.set_option('display.colheader_justify', 'center')
-print(df.to_string(index=False, justify='center'))
+print(df.to_markdown(index=False))
 
 # Print high-level resource estimation results
 # if "physicalCountsFormatted" in res:
--- a/samples/estimation/df-chemistry/chemistry.py
+++ b/samples/estimation/df-chemistry/chemistry.py
@@ -552,14 +552,9 @@ for item in result_obj:
     else:
         data['Physical qubits'].append("-")
         data['rQOPS'].append("-")
- 
-df = pd.DataFrame(data)
+
+# Print high-level resource estimation results 
+df = pd.DataFrame(data, columns=["Run name", "T factory fraction", "Runtime", "Physical qubits", "rQOPS"])
 print(df.to_markdown(index=False))
 
-# Print high-level resource estimation results
-# if "physicalCountsFormatted" in res:
-#     print(f"Algorithm runtime: {res['physicalCountsFormatted']['runtime']}")
-#     print(
-#         f"Number of physical qubits required: {res['physicalCountsFormatted']['physicalQubits']}"
-#     )
 print("For more detailed resource counts, see file resource_estimate.json")
--- a/samples/estimation/df-chemistry/chemistry.py
+++ b/samples/estimation/df-chemistry/chemistry.py
@@ -492,11 +492,6 @@ if args.paramsfile is None:
             "errorBudget": 0.01,
             "qubitParams": {"name": "qubit_maj_ns_e6"},
             "qecScheme": {"name": "floquet_code"},
-        },
-        {
-            "errorBudget": 0.01,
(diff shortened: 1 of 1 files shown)


samples/estimation/df-chemistry/chemistry.py
```diff
@@ -10,6 +10,7 @@
 from scipy import linalg as LA
 from urllib.parse import urlparse
 from urllib.request import urlretrieve
+from qsharp_widgets import EstimatesPanel
```
msoeken, line 13: It seems that this is not used in this change. Would it be possible to use the widgets instead of manually creating the table in text? Otherwise, maybe a pandas DataFrame could be the better option.

msoeken (changes requested):

samples/estimation/df-chemistry/chemistry.py
devikamehra: Widgets were not a good choice for commandline. I can try DataFrame and remove this import statement.

devikamehra: @msoeken I have added data using dataframes but it does not have the grid for better visuals. In case it is ok to add one more library tabulate, the visuals would be as follows:
image

Please let me know if you find this one better.

msoeken: It's not a very common library. How about the markdown output for DataFrames: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_markdown.html

devikamehra: Sure, I have updated the code.

samples/estimation/df-chemistry/chemistry.py
```diff
@@ -513,10 +516,50 @@ def ndarray2d_to_string(arr):
...
+        data['Physical qubits'].append(item["physicalCounts"]["physicalQubits"])
+        data['rQOPS'].append(item["physicalCounts"]["rqops"])
+    elif "frontierEntries" in item:
+        data['Physical qubits'].append(item["frontierEntries"][0]["physicalCounts"]["physicalQubits"])
+        data['rQOPS'].append(item["frontierEntries"][0]["physicalCounts"]["rqops"])
+    else:
+        data['Physical qubits'].append("-")
+        data['rQOPS'].append("-")
+ 
+df = pd.DataFrame(data)
```
msoeken, line 556: I suggest to explicitly list the column names in the `columns` keyword parameter to `DataFrame`, because the order in the `dict` may not be deterministic.

samples/estimation/df-chemistry/chemistry.py
```diff
@@ -513,10 +516,50 @@ def ndarray2d_to_string(arr):
...
-if "physicalCountsFormatted" in res:
-    print(f"Algorithm runtime: {res['physicalCountsFormatted']['runtime']}")
-    print(
-        f"Number of physical qubits required: {res['physicalCountsFormatted']['physicalQubits']}"
-    )
+# if "physicalCountsFormatted" in res:
+#     print(f"Algorithm runtime: {res['physicalCountsFormatted']['runtime']}")
+#     print(
+#         f"Number of physical qubits required: {res['physicalCountsFormatted']['physicalQubits']}"
+#     )
```
msoeken, line 564: These lines can be reomved.

```suggestion
```

samples/estimation/df-chemistry/chemistry.py
```diff
@@ -513,10 +516,50 @@ def ndarray2d_to_string(arr):
...
+        data['Physical qubits'].append(item["frontierEntries"][0]["physicalCounts"]["physicalQubits"])
+        data['rQOPS'].append(item["frontierEntries"][0]["physicalCounts"]["rqops"])
+    else:
+        data['Physical qubits'].append("-")
+        data['rQOPS'].append("-")
+ 
+df = pd.DataFrame(data)
+print(df.to_markdown(index=False))
+
 # Print high-level resource estimation results
```
msoeken, line 559: This line should go before calling `to_markdown`.

msoeken (changes requested):
The changes look great. Just a few more comments. Thank you!

msoeken (approved):
Looks good to me. Thanks @devikamehra

samples/estimation/df-chemistry/chemistry.py
tcNickolas: When I try to run this code, it requires pandas (expected) and then tabulate. This conversation sounded like we don't want to add tabulate as a dependency; what's the final verdict on this?

tcNickolas: @devikamehra Have you seen the discussion about the tabulate dependency that has crept into the PR?

devikamehra: @tcNickolas No Mariia, Thank you driving my attention on this. Will verify if tabulate required for installation because I did not include it as per the discussion with Mathias.

devikamehra: @msoeken @tcNickolas Got it, I have updated the code such that we do not require pandas at all. Let me know if this plain vanilla code is ok or not.

billti (approved):

samples/estimation/df-chemistry/chemistry.py
```diff
@@ -488,11 +487,18 @@ def ndarray2d_to_string(arr):
...
-        "qubitParams": {"name": "qubit_maj_ns_e6"},
-        "qecScheme": {"name": "floquet_code"},
-    }
+    params = [
+        {
+            "errorBudget": 0.01,
+            "qubitParams": {"name": "qubit_maj_ns_e6"},
+            "qecScheme": {"name": "floquet_code"},
+        },
+        {
```
tcNickolas, line 496: These two sets of parameters look to be the same, it would make sense to make them different if we want to illustrate running resource estimates on different parameters at once

devikamehra: Removed the extra example. I think the original file had only one.
In case you have any specific second example in mind, I can add that here.

tcNickolas: I don't have a specific example in mind, but it makes sense that if we're updating the code to run multiple estimates at once and print the summary table, that table has several rows. Any other built-in qubit platform would work here

devikamehra: Sure, I updated the code with another example.

sezna (approved):

PR #1703, opened 2024-07-08
include packageType in qsharp.schema.json

sezna:
Tested with https://www.jsonschemavalidator.net

### DIFF
--- a/vscode/qsharp.schema.json
+++ b/vscode/qsharp.schema.json
@@ -18,7 +18,7 @@
       }
     },
     "packageType": {
-      "title": "Package Type",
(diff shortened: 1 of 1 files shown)


vscode/qsharp.schema.json
```diff
@@ -17,6 +17,11 @@
         "type": "string"
       }
     },
+    "packageType": {
+      "title": "Package Type",
```
billti, line 21: `Package type` (notice how `Language features` above is not Title Case).

billti (approved):

ScottCarda-MS (approved):

PR #1707, opened 2024-07-08
Allow the running of unsaved files

swernli:
The additional check for the ".qs" extension in the program config prevented using single file, unsaved buffers with language mode set to Q# in run and debug. Removing this fixes the issue.

### DIFF
--- a/vscode/src/programConfig.ts
+++ b/vscode/src/programConfig.ts
@@ -3,7 +3,6 @@
 
 import { IProjectConfig, ProgramConfig } from "qsharp-lang";
 import * as vscode from "vscode";
-import { Utils } from "vscode-uri";
 import { isQsharpDocument } from "./common";
 import { getTarget } from "./config";
 import { loadProject } from "./projectSystem";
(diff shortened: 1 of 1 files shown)


vscode/src/programConfig.ts
```diff
@@ -76,10 +75,6 @@ export function getActiveQSharpDocumentUri(): vscode.Uri | undefined {
 export async function getProgramForDocument(
   docUri: vscode.Uri,
 ): Promise<FullProgramConfigOrError> {
-  if (Utils.extname(docUri) !== ".qs") {
-    return { success: false, errorMsg: `${docUri.fsPath} is not a Q# file` };
```
billti, line 80: A lot of other paths call getActiveProgram (esp. commands like RE and circuits), which ends with `return await getProgramForDocument(docUri);`, and check `success` on the result to see if they should error or not. Did you check those code paths to make sure this change is OK?

billti: (they seem OK at first glance to me - just double checking)

billti (approved):

vscode/src/programConfig.ts
swernli: If you look farther up in getActiveProgram, it checks the language setting for the document. I'm of the opinion that the setting should supersede extension. But there is one other place, the debugger, that calls this function directly. I'm going to do some more manual tests to confirm desired behavior.

swernli: I was able to verify behavior, and this works as expected for code lenses and play button. Run menu is still not working, but that is a separate issue I'll address in follow up.

PR #1711, opened 2024-07-09
Enable Run menu in VS Code

swernli:
The Run menu options for "Start Debugging" and "Run without Debugging" were not working, in part to how the validation checking for the presence of a launch.json. This reverses the if-statement and drops the else condition to get recognition of single file scenarios without launch.json. Confirmed via manual testing.

### DIFF
--- a/vscode/src/debugger/activate.ts
+++ b/vscode/src/debugger/activate.ts
@@ -113,19 +113,7 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
     config: vscode.DebugConfiguration,
     _token?: vscode.CancellationToken | undefined,
   ): vscode.ProviderResult<vscode.DebugConfiguration> {
-    // if launch.json is missing or empty
-    if (!config.type && !config.request && !config.name) {
-      const docUri = getActiveQSharpDocumentUri();
-      if (docUri) {
-        config.type = "qsharp";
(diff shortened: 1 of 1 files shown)


ScottCarda-MS (approved):

vscode/src/debugger/activate.ts
```diff
@@ -113,19 +113,7 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
...
-        config.name = "Launch";
-        config.request = "launch";
-        config.programUri = docUri.toString();
-        config.shots = 1;
-        config.noDebug = "noDebug" in config ? config.noDebug : false;
-        config.stopOnEntry = !config.noDebug;
-      }
-    } else if (config.program && folder) {
+    if (config.program && folder) {
       // A program is specified in launch.json.
```
billti, line 117: Would be good to test with a `launch.json` present, and also maybe have @idavis give the thumbs up before merging.

billti (approved):

vscode/src/debugger/activate.ts
swernli: Will do!

swernli: Confirmed it behaves as expected with a launch.json (uses the file specified in that configuration, along with relevant settings).

PR #1712, opened 2024-07-09
Workarounds for IPython cell timing and output buffering

### LINKED ISSUE #1557: Notebook cells running `qsharp` calls always report 0.0s execution time
(Note: I haven't tested in JupyterLab, only in VS Code, so it's possible that this a bug specific to that environment)

If I create a cell with qsharp that takes a long time to execute, something like:
```qsharp
operation Test() : Unit {
    use q = Qubit();
    for _ in 1..10000000 {
        X(q);
    }
    Reset(q);
}
```
and then try to run it using:
```python
qsharp.run("Test()", shots=1);
```
The cell will show as "waiting" while the native simulation is running and only report any execution time once the operation has completed, making the reported time 0.0s. Perhaps somehow the state transition for going into the native code isn't being seen by Python and only the time after we cross back from that boundary is counted?

This has another negative side effect, which is that clicking the "stop" button on the cell doesn't actually attempt any cancellation and leaves the notebook in a bad state. Trying to run any later cells are stuck waiting on the long running cell, which means the only way to cancel is to restart the kernel.

swernli:
This change adds some workarounds that help with accurately reporting timing and avoiding output buffering when running either `%%qsharp` cell magic or `qsharp.eval` (and similar calls) in Python cells.

Fixes #1557

### DIFF
--- a/pip/qsharp/_ipython.py
+++ b/pip/qsharp/_ipython.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from IPython.display import display, Javascript, Pretty
+from IPython.display import display, Javascript, clear_output
 from IPython.core.magic import register_cell_magic
 from ._native import QSharpError
 from ._qsharp import get_interpreter
@@ -12,9 +12,16 @@ def register_magic():
     @register_cell_magic
     def qsharp(line, cell):
         """Cell magic to interpret Q# code in Jupyter notebooks."""
+        # This effectively pings the kernel to ensure it recognizes the cell is running and helps with
+        # accureate cell execution timing.
+        clear_output()
 
         def callback(output):
             display(output)
+            # This is a workaround to ensure that the output is flushed. This avoids an issue
+            # where the output is not displayed until the next output is generated or the cell
+            # is finished executing.
+            display(display_id=True)
 
         try:
             return get_interpreter().interpret(cell, callback)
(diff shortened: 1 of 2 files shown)


pip/qsharp/_ipython.py
```diff
@@ -12,9 +12,16 @@ def register_magic():
...
+        # This effectively pings the kernel to ensure it recognizes the cell is running and helps with
+        # accureate cell execution timing.
+        clear_output()
 
         def callback(output):
             display(output)
+            # This is a workaround to ensure that the output is flushed. This avoids an issue
+            # where the output is not displayed until the next output is generated or the cell
+            # is finished executing.
+            display(display_id=True)
```
billti, line 24: Interesting. So this doesn't impact the actual content rendering at all? (i.e. doesn't clear the display, or add a newline, etc.)

billti (approved):

pip/qsharp/_ipython.py
swernli: That's right. It's just enough to force an update of the output area, including showing any buffered output, without changing anything visible.

Before I merge, I'd like to verify behavior with JupyterLab too, just to make sure it doesn't have any unforeseen side effects in that environment (VS Code is where I've tested so far, and it behaves nicely).

swernli: Confirmed the fix in JupyterLab too! Before the change, output is buffered the same way as in VS Code, with the fix the output is effectively streamed and no additional UI artifacts show up.

idavis (approved):

PR #1713, opened 2024-07-09
Fix rendering read-only source buffer for Github dependencies that are in subdirectories of repos

sezna:
There was just one missing slash! Tested locally: go-to-def,  debugging, stepping into, etc all work.

### DIFF
--- a/compiler/qsc_project/src/project.rs
+++ b/compiler/qsc_project/src/project.rs
@@ -414,8 +414,15 @@ pub trait FileSystemAsync {
         let manifest = serde_json::from_str::<Manifest>(&manifest_content).map_err(|e| {
             Error::GitHubManifestParse {
                 path: format!(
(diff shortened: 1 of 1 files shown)


swernli (approved):

ScottCarda-MS (approved):

compiler/qsc_project/src/project.rs
```diff
@@ -437,7 +437,7 @@ pub trait FileSystemAsync {
             // and open them using cached contents.
             sources.push((
                 format!(
-                    "qsharp-github-source:{}/{}/{}{path}",
+                    "qsharp-github-source:{}/{}/{}/{path}",
```
ScottCarda-MS, line 440: From the integration test failures, it looks like `path` can sometimes have a leading `/` which messes up this string.

PR #1714, opened 2024-07-09
Fix panic in hover on local variables of notebook cell

swernli:
There was an assumption baked into the hover support that locals will always come from a callable scope, but in notebooks they can come from fragments with top-level statements. This avoids the panic by treating the callable name as optional until it is required for display.

### DIFF
--- a/language_service/src/hover/tests.rs
+++ b/language_service/src/hover/tests.rs
@@ -1596,3 +1596,16 @@ fn notebook_callable_defined_in_later_cell() {
         ("cell2", "operation Callee() : Unit {}"),
     ]);
 }
+
+#[test]
+fn notebook_local_definition() {
+    check_notebook(
+        &[("cell1", "let x = 3;"), ("cell2", "let ◉↘y◉ = x + 1;")],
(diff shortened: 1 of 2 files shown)


language_service/src/hover/tests.rs
```diff
@@ -1596,3 +1596,16 @@ fn notebook_callable_defined_in_later_cell() {
         ("cell2", "operation Callee() : Unit {}"),
     ]);
 }
+
+#[test]
+fn notebook_local_definition() {
```
ScottCarda-MS, line 1601: What happens if the hover is on the reference to `x` in the second cell?

ScottCarda-MS (approved):
Good catch!

language_service/src/hover/tests.rs
swernli: The same thing that happens with hover on any local variable use with files: nothing. I'm not sure if that is intentional, but it is consistent at least. Is that worth filing an issue for follow up across both scenarios?

billti (approved):

PR #1715, opened 2024-07-09
include package alias in auto-opens for completions

sezna:
auto-opens:

https://github.com/microsoft/qsharp/assets/12157751/a0446fd8-0da4-4553-80ef-6f3a8c87d0c2

fully qualified namespace paths:

https://github.com/microsoft/qsharp/assets/12157751/39ddebb6-de57-41e5-9da0-b12f0ce2a344

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -19,6 +19,7 @@ use qsc::{
 };
 use rustc_hash::FxHashSet;
 use std::rc::Rc;
+use std::sync::Arc;
 
 type NamespaceName = Vec<Rc<str>>;
 type NamespaceAlias = Rc<str>;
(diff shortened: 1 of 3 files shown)


billti:
Nice! Videos of experience looks great. I'll let @swernli  and @ScottCarda-MS give the thumbs up on the code changes. Thanks!

language_service/src/completion.rs
```diff
@@ -471,7 +480,15 @@ impl CompletionListBuilder {
                                                 Some(start) => {
                                                     additional_edits.push(TextEdit {
                                                         new_text: format!(
-                                                            "open {};{indent}",
+                                                            "open {}{};{indent}",
```
ScottCarda-MS, line 483: Is it doable to have the package alias just be part of the stored namespace for items? I ask because prefixing with the package name should really happen everywhere the namespace appears, right? I can see from trying it out that the alias does _not_ appear when hovering over items from external code, which seems erroneous. It would be good if we could solve all these issues centrally by making the package alias part of the stored namespace; otherwise we'll need to do this prefixing logic a lot, right?

swernli: I think it's worth filing a follow up issue for this and doing a deeper investigation. To your point, Scott, there are other places in language service like hover that don't have the alias. I think we'd want to double check stuff like type errors as well (I think those don't use fully qualified names at all, but it's worth a double check, because if they do and alias is missing it could cause confusion due to seeming collisions). It's possible this is already well plumbed for those type of compiler error messages, but either way it's worth combing through and making sure we are consistent throughout.

swernli (approved):
The added functionality and implementation look good. I think the feedback from @ScottCarda-MS is worth capturing into an issue for follow up.

ScottCarda-MS (approved):

PR #1717, opened 2024-07-09
Languages Service loads project and dependencies in Notebooks

swernli:
This updates the compilation logic for the language service when the active document is a notebook to include any specificed project root folder such that dependencies are correctly resolved. This avoids spurious error squiggles on items that come from dependencies and even enables go to definition and hover for those items.

### DIFF
--- a/language_service/src/compilation.rs
+++ b/language_service/src/compilation.rs
@@ -15,7 +15,7 @@ use qsc::{
     CompileUnit, LanguageFeatures, PackageStore, PackageType, PassContext, SourceMap, Span,
 };
 use qsc_linter::{LintConfig, LintLevel};
-use qsc_project::PackageGraphSources;
+use qsc_project::{PackageGraphSources, Project};
 use std::mem::take;
 use std::sync::Arc;
 
@@ -44,7 +44,7 @@ pub(crate) enum CompilationKind {
     /// A Q# notebook. In a notebook compilation, the user package
     /// contains multiple `Source`s, with each source corresponding
     /// to a cell.
-    Notebook,
+    Notebook { project: Option<Project> },
 }
 
 impl Compilation {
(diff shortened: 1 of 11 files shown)


swernli (commented):
I'm missing tests, which I'll add soon.

swernli:
Tests added, and good thing too, caught a bug reporting errors from dependencies of dependencies.

language_service/src/compilation.rs
```diff
@@ -113,21 +113,81 @@ impl Compilation {
...
+            &dependencies,
+        ) {
+            Ok(compiler) => compiler,
+            Err(errors) => match &project {
+                Some(p) => {
+                    trace!("compilation failure for sources: {:?}", errors);
+                    let (_, store) = qsc::compile::package_store_with_stdlib(target_profile.into());
+                    return Self {
+                        package_store: store,
+                        user_package_id: 2.into(), // We don't have a legitimate user package id here, so fall back to a guess.
```
ScottCarda-MS, line 180: I thought the user's package id was standardized to a particular value, right?

language_service/src/state/tests.rs
```diff
@@ -605,20 +605,22 @@ async fn target_profile_update_causes_error_in_stdlib() {
     );
 }
 
-#[test]
-fn notebook_document_no_errors() {
+#[tokio::test]
```
ScottCarda-MS, line 608: Are the tokio tests needed because these are async tests?

language_service/src/state/tests.rs
```diff
@@ -924,6 +936,207 @@ fn close_notebook_clears_errors() {
...
+    expect_errors(
+        &errors,
+        &expect![[r#"
+        []
+    "#]],
+    );
+}
+
+#[tokio::test]
+async fn update_notebook_reports_errors_from_dependencies() {
```
ScottCarda-MS, line 985: Is this test saying that if there are errors in a dependency, we'd see those errors when we open the associated namespace in the notebook?

ScottCarda-MS: Maybe I'm not understanding the relationship between the project and the notebook in these tests. It doesn't seem right to me that people would get errors from external code like that.

language_service/src/compilation.rs
swernli: No, it's not, unfortunately. Even without external depednencies, there was the option for the user to skip the stdlib (though technically we haven't exposed it out of the compiler yet). So core is always 0, if stdlib is included it's 1, and the user codee is then either 1 or 2. But with external dependencies, those must get compiled and added to the store before the user code, and there can be any number of those (since it also includes dependencies of dependencies, and so on). So the user code could have any id greater than 0 now. To make things more confusing, when the user code fails to compile, `Compiler::new` does not add it to the store AND consumes the store, which is why I have to generate a new one here. Since it was never added to the store, it doesn't have any id at this point in the code.

language_service/src/state/tests.rs
swernli: Yup. The standard `#[test]` doesn't support async code, but `#[tokio::test]` does. I hadn't actually seen that before digging into the tests in this file, so it's new to me as well.

language_service/src/state/tests.rs
swernli: If a dependency can't compile, the user code can't compile either. Since dependencies are brought in as source and compiled here rather than distributed as precompiled binaries (the same mechanism as Rust's source distribution), all dependencies must be compiled in the right order before the user code. So yes, it's expected and necessary for the user to see failures from dependencies to know there is a bug or misconfiguration in one of them that will prevent any other code from working. These errors show up in the problems view in VS Code, and clicking them will jump to the offended dependency file, which can help with reporting bugs in published code.

One thing I should do here is mimic the pattern Alex set up that does a best effort compile of the user code even when dependencies fail... it will show errors on everything the user code tries to pull from that or any dependency, but that can be a helpful signal to the user that things have gone wrong.

language_service/src/compilation.rs
ScottCarda-MS: Maybe we should change the logic to always include a package 1, and in normal compilation that will be the stdlib, but if that is skipped in a compilation, then we put an empty package there. That way the user code is _always_ 2. Does this seem like a good idea? If so, I could make a git issue for this.

language_service/src/state/tests.rs
ScottCarda-MS: Okay, I can see the reasoning behind that. Thanks for the explanation!

ScottCarda-MS (approved):

language_service/src/compilation.rs
swernli: No, that wouldn't work either because of dependencies. If the user code depends on Package A, which in turn depends on Package B, B must be compiled first, then A, then the user code. So the user code will be 4 by necessity. Have more dependencies, and the numbers just get higher.

ScottCarda-MS: I see, thanks for the explanation.

language_service/src/compilation.rs
```diff
@@ -120,23 +120,71 @@ impl Compilation {
...
             store,
-            &[(std_id, None)],
-        )
-        .expect("expected incremental compiler creation to succeed");
+            &dependencies,
+        ) {
+            Ok(compiler) => compiler,
+            Err(user_errors) => {
+                errors.extend(user_errors);
+                // Because there were errors in the leaf project, we need to create a new compiler with no sources
```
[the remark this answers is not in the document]
swernli: I can make that change, but one thing that I was struggling to describe was the difference between the user code in a project and the user code in notebook cells. At this point in the compilation, we are handling errors from the user code project, but from the perspective of the notebook user, that might not be seen as their code but rather their first dependency. Not sure if we are ever going to change that, but in working on this I realized there are two mental models there and we happen to be encoding one vs the other.

pip/qsharp/_qsharp.py
```diff
@@ -52,6 +53,8 @@ def __init__(
 
         self._config["languageFeatures"] = language_features
         self._config["manifest"] = manifest
+        if project_root:
+            self._config["projectRoot"] = "file://" + project_root
```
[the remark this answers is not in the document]
swernli: Mostly confident... the URI string we output here is for consumption of the language service, which is running in VS Code and needs to use this URI to access the corresponding folder on disk. If we are in the web, the URI would be wrong because it doesn't refer to the virtual file system... but the web + virtual file system scenario doesn't support Jupyter notebooks. If/when we have pure web hosted Jupyter we'll need to revisit this, but other things in our Python would need to change too.

Writing this up tells me I should add a comment here to capture this assumption :)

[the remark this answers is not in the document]
swernli: Added comment.

language_service/src/compilation.rs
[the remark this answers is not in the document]
swernli: Updated terminology.

PR #1719, opened 2024-07-10
Remove `newtype`, `::`, and `!` From Samples, Libraries, and Katas

ScottCarda-MS:
This PR replaces the `::` field operator with the new `.` field operator and removes uses of the `!` unwrap operator in our Samples, Libraries, and Katas code. It also replaces UDT definitions with struct definitions.


billti (changes requested):
Leaving this review to block accidental early merging. We want to ensure this doesn't go in until the language changes are released in 'stable', syntax highlighting for 3rd parties are updated (GitHub, Monaco, etc.) and the official documentation updated.

katas/content/oracles/index.md
```diff
@@ -229,7 +229,6 @@ $$\ket{\xi} = \Big[\frac{1}{\sqrt{2}}\big(\ket{110} - \ket{111}\big)\Big] \otime
 
 We can see that these two equations are identical, except for the $-1$ phase that appeared on the $\ket{111}$ basis state - our marked state.  This is a specific example of the phase kickback effect, as the phase from $\ket{-}$ has been *kicked back* into $\ket{x}$.
 
-
```
tcNickolas, line 232: Can we exclude the files with whitespace-only changes from the PR? It does not make sense to do them in files that are otherwise completely unrelated to user-defined types

billti: Do they impact any rendering? If not, seems fine to fix while in the files.

If it's just 'noisy' when reviewing, you can suppress by putting `?w=1` on the end of the PR url (e.g. try <https://github.com/microsoft/qsharp/pull/1719/files?w=1>)

samples/language/TypeDeclarations.qs
```diff
@@ -2,33 +2,25 @@
 /// Type Declarations
 ///
 /// # Description
-/// User-defined types, or UDTs as they are commonly called, are supported
-/// in Q#. They are immutable but support a copy-and-update construct.
+/// User-defined struct types are supported in Q#.
+// They are immutable but support a copy-constructor syntax.
```
billti, line 6: Dropped a `/` here, making it a double instead of triple, so it's not a doc comment.

katas/content/oracles/index.md
ScottCarda-MS: I can revert those changes if we really want to, but the warnings in those files will then persist and just get fixed again the next time someone saves the file.

samples/language/TypeDeclarations.qs
ScottCarda-MS: Good catch! Fixed

ScottCarda-MS:
Superseded by #1797

PR #1727, opened 2024-07-10
Introduce the modern Stdlib API; Fix re-exports

sezna:
This PR fixes all the rough edges around re-exports and re-exports with aliases, and specifically re-exporting namespaces.

## Summary of Changes

* The package store has to be available to the HIR and FIR, because a re-export can refer to an item from another package. To resolve any `ItemId` where `package` is not `None`, we need to query the `PackageStore`. So that got plumbed all the way through.
* An export can be its own `ItemKind`. It serves as a "reference" to another item in another package if it is present. This is used to ensure that items defined in package A will show up in package B's items, if it is re-exported from package B.
* In both the resolver and the type checker, we follow the "chain of re-exports" until we find the original item that is being exported. There are multiple places that this happens. [Here is an example](https://github.com/microsoft/qsharp/pull/1727/files#diff-d98d007bbe0a3fce75f8d49278e0b566c9f8d91731f3e451296db14bebff001fR403).
* Lots of tests.
* The most nuanced thing is the re-creation of the namespace tree when adding an external package in the resolver. This was actually the bug that took me the longest to sort out. I'll break out of this list for the sake of formatting, as this one requires more explanation:

### The Nuanced Bit

Namespace trees are not acyclic, and the same underlying node can show up in different places of the tree. For example, in the new standard library, `Microsoft.Quantum.Arrays` is accessible via both `Microsoft.Quantum.Arrays` _and_ `Std.Arrays`.  When we add an external package to the local package, we need to recreate this structure. 

To accomplish this, I wrote an iterator over the namespace tree which returns all names for which each namespace can be referred to. For example, the above namespace would show up in the iterator as `[Microsoft.Quantum.Arrays, Std.Arrays]`. This allows the resolver to reconstruct the exact same tree structure and preserve the aliased/equality property of the two namespaces. And, this is all done without having to know the underlying `NamespaceId`s. This is important, because `NamespaceId`s start from 0 for each package compilation -- it'd be an issue if we had to rely on the IDs of one package to resolve namespaces in another package. 

## Other Notes

With re-exports working as designed, the entire stdlib diff for re-exporting all items under the new name `Std` is:

```qsharp
namespace Std {
    operation TestFunc() : Unit { Message ("test func"); }
    export TestFunc, Microsoft.Quantum.Arrays, Microsoft.Quantum.Canon, Microsoft.Quantum.Convert, Microsoft.Quantum.Core, Microsoft.Quantum.Diagnostics, Microsoft.Quantum.Logical, Microsoft.Quantum.Intrinsic, Microsoft.Quantum.Math, Microsoft.Quantum.Measurement, Microsoft.Quantum.Random, Microsoft.Quantum.ResourceEstimation, Microsoft.Quantum.Unstable;
}

```

When we support glob exports, in addition to glob imports, this will be even shorter:

```


// This file re-exports the standard library under the name `Std`, which will be the preferred standard library API going forward.

namespace Std {
    export Microsoft.Quantum.*;
}

```
----

## Example

Feel free to test the Bernstein Vaszirani example with the new API:

```qsharp


import Std.Arrays.IndexRange;
import Std.Convert.ResultArrayAsInt;
import Std.Diagnostics.Fact;
import Std.Math.BitSizeI;

@EntryPoint()
operation Main() : Int[] {
    let nQubits = 10;

    let integers = [127, 238, 512];
    mutable decodedIntegers = [];
    for integer in integers {
        let parityOperation = EncodeIntegerAsParityOperation(integer);

        let decodedBitString = BernsteinVazirani(parityOperation, nQubits);
        let decodedInteger = ResultArrayAsInt(decodedBitString);
        Fact(
            decodedInteger == integer,
            $"Decoded integer {decodedInteger}, but expected {integer}."
        );

        Message($"Successfully decoded bit string as int: {decodedInteger}");
        set decodedIntegers += [decodedInteger];
    }

    return decodedIntegers;
}

operation BernsteinVazirani(Uf : ((Qubit[], Qubit) => Unit), n : Int) : Result[] {
    use queryRegister = Qubit[n];
    use target = Qubit();

    X(target);

    within {
        ApplyToEachA(H, queryRegister);
    } apply {
        H(target);
        Uf(queryRegister, target);
    }

    let resultArray = MResetEachZ(queryRegister);

    Reset(target);

    return resultArray;
}

operation ApplyParityOperation(
    bitStringAsInt : Int,
    xRegister : Qubit[],
    yQubit : Qubit
) : Unit {
    let requiredBits = BitSizeI(bitStringAsInt);
    let availableQubits = Length(xRegister);
    Fact(
        availableQubits >= requiredBits,
        $"Integer value {bitStringAsInt} requires {requiredBits} bits to be represented but the quantum register only has {availableQubits} qubits"
    );

    for index in IndexRange(xRegister) {
        if ((bitStringAsInt &&& 2^index) != 0) {
            CNOT(xRegister[index], yQubit);
        }
    }
}

function EncodeIntegerAsParityOperation(bitStringAsInt : Int) : (Qubit[], Qubit) => Unit {
    return ApplyParityOperation(bitStringAsInt, _, _);
}

```

----

## Description from #1736 

Yes, that title is a mouthful. This is _quite_ the edge case identified by @ScottCarda-MS  -- thank you for bug bashing!

Re-exports are nuanced. They're basically pointers to `ItemId`s which point to another package. As a refresher, if an `ItemId` has a package id (`item_id.package.is_some()`), then it refers to an item outside of the current package, so it is thusly a reexport. If `item_id.package.is_none()`, then the item is local to the current package.

Based on this preexisting notion, re-exports were supported by inserting `ItemKind::ExportedItem` for re-exports which point to another package. These are identified in `add_external_package` and inserted into the names of the user code for resolution. And if an item was local, then no `ExportedItem` is needed -- we just mark the item's `Visibility` as public. 

But what, what if that local item is exported with an alias? Then, we need some way to mark it as public _and_  tell the HIR that it has a different name. To handle this, we insert an `ExportedItem` if an item is local and the export has an alias.

So, to recap, an `ItemKind::ExportedItem` is inserted if the item being exported is either from another package (a re-export) or from the local package, but aliased. Everything mentioned up until now is handled in #1727.

This PR fixes a bug where a re-export of a local item which is introduced to the scope via an import alias loses its alias upon exporting. Wow, what a mouthful. Basically, in order to insert an `ItemKind::ExportedItem` for an export with an alias, we just checked if `export.alias.is_some()`. But, actually, you can also basically perform an aliased export like this: `import Foo as Bar; export Bar;`. In this case, there is indeed no export on the alias. But we do still need an `ItemKind::ExportedItem`, because we need to track that this same `ItemId` is accessible via both names -- the original declared name and the re-exported aliased name.

### DIFF
--- a/compiler/qsc_frontend/src/resolve.rs
+++ b/compiler/qsc_frontend/src/resolve.rs
@@ -162,6 +162,16 @@ pub(super) enum Error {
     #[error("glob exports are not supported")]
     #[diagnostic(code("Qsc.Resolve.GlobExportNotSupported"))]
     GlobExportNotSupported(#[label] Span),
+
+    #[error("aliasing a glob import is invalid")]
+    #[diagnostic(help("try `import {namespace_name} as {alias}` instead"))]
+    #[diagnostic(code("Qsc.Resolve.GlobImportAliasNotSupported"))]
+    GlobImportAliasNotSupported {
+        namespace_name: String,
+        alias: String,
+        #[label]
+        span: Span,
+    },
 }
 
 #[derive(Debug, Clone)]
@@ -1029,9 +1039,19 @@ impl Resolver {
                 .push(Error::GlobExportNotSupported(item.path.span));
             return;
         }
+
+        if let Some(alias) = &item.alias {
+            self.errors.push(Error::GlobImportAliasNotSupported {
+                span: item.span(),
+                namespace_name: Into::<Idents>::into(item.path.clone()).name().to_string(),
+                alias: alias.name.to_string(),
+            });
+            return;
+        }
+
         let items = Into::<Idents>::into(item.path.clone());
         let ns = self.globals.find_namespace(items.str_iter());
-        let alias = item.alias.as_ref().map(|x| Box::new(x.clone()));
+
         let Some(ns) = ns else {
             self.errors.push(Error::GlobImportNamespaceNotFound(
                 item.path.name.to_string(),
@@ -1040,7 +1060,7 @@ impl Resolver {
             return;
         };
         if !is_export {
-            self.bind_open(&items, &alias, ns);
+            self.bind_open(&items, &None, ns);
         }
     }
 
--- a/compiler/qsc_frontend/src/resolve/tests.rs
+++ b/compiler/qsc_frontend/src/resolve/tests.rs
@@ -5094,3 +5094,23 @@ fn import_newtype() {
             }"#]],
     );
 }
+
+#[test]
+fn disallow_glob_alias_import() {
+    check(
+        indoc! {r#"
+                namespace Bar {}
+                namespace Main {
+                    import Bar.* as B;
+                }
+                "#},
+        &expect![[r#"
+            namespace namespace7 {}
+            namespace namespace8 {
+                import namespace7;
+            }
+
+            // GlobImportAliasNotSupported { namespace_name: "Bar", alias: "B", span: Span { lo: 45, hi: 55 } }
+        "#]],
+    );
+}


compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -329,3 +329,190 @@ fn test_get_namespace_id() {
...
+                vec![Rc::from(format!("ns{i}")), Rc::from(format!("ns{j}"))].into_iter(),
+            );
+        }
+    }
+
+    let result = root.iter().collect::<Vec<_>>();
+    expect![[r#"
+        [
+            [
+                [],
```
cesarzc, line 349: Is this an implicit "root" nampespace?

compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -329,3 +329,190 @@ fn test_get_namespace_id() {
...
     .assert_debug_eq(&result_buf);
 }
+
+#[allow(clippy::too_many_lines)]
+#[test]
+fn test_tree_iter() {
+    let mut root = NamespaceTreeRoot::default();
+    for i in 0..3 {
+        for j in 'a'..'d' {
+            let _ = root.insert_or_find_namespace(
```
cesarzc, line 339: Given the changes in `namespaces.rs`, was this test meant for `insert_or_find_namespace_from_root_with_id`?

compiler/qsc_hir/src/hir.rs
```diff
@@ -355,6 +359,7 @@ impl Display for ItemKind {
                 }
             }
             ItemKind::Ty(name, udt) => write!(f, "Type ({name}): {udt}"),
+            ItemKind::Export(_, export) => write!(f, "Export: {export}"),
```
swernli, line 362: nitpick: is it worth including the ident in the pretty print, similar to how it is for `Type` above?

compiler/qsc_hir/src/mut_visit.rs
```diff
@@ -72,6 +72,7 @@ pub fn walk_item(vis: &mut impl MutVisitor, item: &mut Item) {
         ItemKind::Callable(decl) => vis.visit_callable_decl(decl),
         ItemKind::Namespace(name, _) => vis.visit_idents(name),
         ItemKind::Ty(name, _) => vis.visit_ident(name),
+        ItemKind::Export(_, _) => (),
```
swernli, line 75: I think this should visit the ident:

```suggestion
        ItemKind::Export(name, _) => vis.visit_ident(name),
```

compiler/qsc_hir/src/visit.rs
```diff
@@ -65,6 +65,7 @@ pub fn walk_item<'a>(vis: &mut impl Visitor<'a>, item: &'a Item) {
         ItemKind::Callable(decl) => vis.visit_callable_decl(decl),
         ItemKind::Namespace(name, _) => vis.visit_idents(name),
         ItemKind::Ty(name, _) => vis.visit_ident(name),
+        ItemKind::Export(_, _) => {}
```
swernli, line 68: here too, it should probably visit the ident:

```suggestion
        ItemKind::Export(name, _) => vis.visit_ident(name),
```

compiler/qsc_lowerer/src/lib.rs
```diff
@@ -153,8 +157,6 @@ impl Lowerer {
         }
 
         fir_package.entry = entry;
-
-        qsc_fir::validate::validate(fir_package);
```
swernli, line 157: Hmm, this is an unfortunate removal, but has no effect on behavior since validation is just for checking and catching compiler bugs. I see why it was needed, since validate now requires the store as a parameter, and we don't have one here. It's highly unlikely we'll introduce bugs of the kind this validate would check (including references to ids that don't get added to the corresponding maps). I'll resolve this comment, just wanted note the change.

compiler/qsc_rca/src/core.rs
```diff
@@ -1927,10 +1927,27 @@ impl<'a> Visitor<'a> for Analyzer<'a> {
...
+                _,
+                qsc_fir::fir::ItemId {
+                    package: Some(package),
+                    item,
+                },
+            ) => {
+                let package = self.package_store.get(*package);
+                let Some(item) = package.items.get(*item) else {
+                    return;
+                };
```
swernli, line 1940: I actually think it makes sense to turn this into an `.expect` since an item should be found in the package it's listed with. That would help catch any (admittedly unlikely) bugs.

compiler/qsc_passes/src/capabilitiesck/tests_common.rs
```diff
@@ -76,7 +85,8 @@ impl CompilationContext {
     }
 
     fn new_for_exe(source: &str) -> Self {
-        let (std_id, store) = qsc::compile::package_store_with_stdlib(TargetCapabilityFlags::all());
+        let mut store = qsc::PackageStore::new(qsc::compile::core());
```
swernli, line 88: Is it possible this was a bad merge?

swernli: To clarify, I can't tell from the diff why the utility function you introduced won't work here. Did something change or was this unintentional?

compiler/qsc_data_structures/src/namespaces.rs
```diff
@@ -196,7 +196,75 @@ impl NamespaceTreeRoot {
...
+            let prefix_id = root_contents
+                .borrow_mut()
+                .insert_or_find_namespace(prefix.into_iter().peekable(), &mut self.assigner)
+                .expect("empty name checked for above");
+
+            self.insert_with_id(Some(prefix_id), base_id, &suffix);
+        }
+    }
+
+    /// Each item in this iterator is one namespace. The reason there are multiple paths for it,
```
swernli, line 229: nitpick: for clarity I might phrase this as:

```suggestion
    /// Each item in this iterator is the same, single namespace. The reason there are multiple paths for it,
```

compiler/qsc_frontend/src/compile.rs
```diff
@@ -516,18 +516,23 @@ fn resolve_all(
     assigner: &mut HirAssigner,
     package: &ast::Package,
     mut dropped_names: Vec<TrackedName>,
-) -> (Names, Locals, Vec<resolve::Error>) {
+) -> (
+    Names,
+    Locals,
+    Vec<resolve::Error>,
+    qsc_data_structures::namespaces::NamespaceTreeRoot,
```
swernli, line 523: nitpick: can the errors stay at the end? It looks worse for the diff but something in my head likes the errors list to be the last or the first thing in a tuple. Not a big deal; I'm fine if you shrug and leave it as-is.

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -1500,7 +1624,53 @@ fn bind_global_item(
...
+            } else {
+                for decl_item in decl.items.iter() {
+                    // if the item is a namespace, bind it here as an item
+                    let Some(ns) = scope
+                        .namespaces
+                        .get_namespace_id(Into::<Idents>::into(decl_item.path.clone()).str_iter())
+                    else {
+                        continue;
+                    };
+                    //                    bind_global_item(names, scope, namespace, next_id, item)?;
```
swernli, line 1639: This line looks like itmight be leftover from an earlier implementation.

compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -4742,11 +4746,11 @@ fn export_namespace() {
                 operation item2() : Unit {}
             }
             namespace namespace8 {
-                export namespace7;
+                export item4;
```
swernli, line 4749: Wait a minute, does exporting a namespace actually do anything? I was under the impression you could only export callables and types.

compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -273,8 +273,9 @@ fn conditional_compilation_not_available() {
 
 #[test]
 fn errors_across_multiple_lines() {
-    let mut store = PackageStore::new(crate::compile::core());
-    let std_id = store.insert(crate::compile::std(&store, TargetCapabilityFlags::all()));
+    let mut store = PackageStore::new(compile::core());
+    let std = compile::std(&store, TargetCapabilityFlags::all());
```
swernli, line 277: Out of curiosity, why split these apart?

compiler/qsc_frontend/src/compile/tests.rs
```diff
@@ -1267,7 +1267,7 @@ fn accept_use_qubit_block_syntax_if_preview_feature_is_off() {
         &store,
         &[(std, None)],
         sources,
-        TargetCapabilityFlags::empty(),
+        TargetCapabilityFlags::all(),
```
swernli, line 1270: This change doesn't seem like it would be needed... does something fail without it?

swernli (approved):
A few minor issues and questions, the biggest one is making sure the RCA tests are turned back on and passing. Otherwise, looks good!

compiler/qsc/src/interpret/package_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+use qsc_project::{PackageGraphSources, PackageInfo};
+use rustc_hash::FxHashMap;
+
+#[test]
+fn import_and_call_reexport() {
+    let pkg_graph: PackageGraphSources = PackageGraphSources {
+        root: PackageInfo {
+            sources: vec![(
+                "PackageB.qs".into(),
+                indoc! {"
```
cesarzc, line 20: nit: indentation in some of this test cases is a bit odd and makes it slightly harder to follow what the test case is doing.

compiler/qsc/src/interpret/package_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    let BuildableProgram {
+        store,
+        user_code,
+        user_code_dependencies,
+        ..
+    } = buildable_program;
+
+    let user_code = SourceMap::new(user_code.sources, None);
+
+    let _ = Interpreter::new(
```
cesarzc, line 73: Since this is an interpreter test, would it make sense to also call `eval_entry()` in this test cases to verify that the evaluator runs the correct item?

compiler/qsc_codegen/src/qir.rs
```diff
@@ -20,7 +20,7 @@ use qsc_rir::{
 fn lower_store(package_store: &qsc_frontend::compile::PackageStore) -> qsc_fir::fir::PackageStore {
     let mut fir_store = qsc_fir::fir::PackageStore::new();
     for (id, unit) in package_store {
-        let package = qsc_lowerer::Lowerer::new().lower_package(&unit.package);
+        let package = qsc_lowerer::Lowerer::new().lower_package(&unit.package, &fir_store);
```
cesarzc, line 23: Is there a reason for lowering and updating the package in two separate calls instead of using the `lower_and_update_package` function?

compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -469,6 +470,176 @@ fn continue_after_lower_error() {
...
+                Ok(())
+            },
+        )
+        .expect("compile_fragments should succeed");
+
+    expect!["[]"].assert_eq(&format!("{errors:#?}"));
+}
+
+#[test]
+fn reexports() {
```
cesarzc, line 561: nit: I would rename this to be more explicit, something like: "package_reexports_operation_from_a_dependency".

compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -469,6 +470,176 @@ fn continue_after_lower_error() {
...
+    );
+    assert!(package_a.errors.is_empty(), "{:#?}", package_a.errors);
+
+    let package_a = store.insert(package_a);
+
+    let package_b = SourceMap::new(
+        [(
+            "PackageB.qs".into(),
+            indoc! {"
+                import A.PackageA.Foo;
```
cesarzc, line 506: nit: IMO this test is doing a lot. I think it could be broken down into multiple tests:
- Importing with alias.
- Re-exporting with alias.

It also seems like importing with no alias is a missing test case in this file.

A lot of the code for these test cases would be shared so it would also be good to create a few utility functions to make the code of each test more concise.

compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -582,6 +588,13 @@ impl<'a> Context<'a> {
                         .expect("local should have type")
                         .clone(),
                 ),
+                Some(Res::ExportedItem(item, _)) => {
+                    // get the underlying item this refers to
+                    let item_scheme = self.globals.get(item).expect("item should have scheme");
```
cesarzc, line 593: I'm a bit unfamiliar with the term "item scheme". I couldn't find a reference to it in the functions used in this part of the logic. What do you mean by it?

compiler/qsc_frontend/src/resolve.rs
```diff
@@ -930,30 +967,56 @@ impl Resolver {
...
-                    .insert(local_name.clone(), ScopeItemEntry::new(id, item_source));
+                scope.tys.insert(
+                    local_name.clone(),
+                    ScopeItemEntry::new(id, item_source.clone()),
+                );
             }
 
             let res = match (term_result, ty_result) {
-                (Ok(res @ Res::Item(..)), _) | (_, Ok(res @ Res::Item(..))) => res,
+                // This is kind of a messy match, it is merged and formatted this way
```
cesarzc, line 977: nit: I would place the "... is kind of a messy match ..." part of the comment before the `match` expression, leaving the explanation for the first case here. That way each relevant case gets its own comment :)

compiler/qsc_rca/src/lib.rs
```diff
@@ -6,9 +6,6 @@
 //! execution on a quantum kernel and does not consider these elements when determining the capabilities. Additionally,
 //! this implementation also provides details on why the program requires each capability.
 
-#[cfg(test)]
```
cesarzc, line 9: I think this might be unintended since it disables RCA tests.

cesarzc (approved):
The only major thing is re-enabling RCA tests, all the other comments are non-blockers. Everything else looks good!

compiler/qsc_hir/src/mut_visit.rs
swernli: We discussed this together and found a solution that uses the assigner from lowering to generate new ids for the cloned `Ident`, which seems to have done the trick!

compiler/qsc_codegen/src/qir.rs
cesarzc: I don't think you should refactor in this PR. After going through all the changes there are many places where using `lower_and_update_package` could be used but that can be done later. Closing this comment.

compiler/qsc_frontend/src/typeck/rules.rs
cesarzc: Thanks for the explanation!

ScottCarda-MS: I think this framework is a bit out-of-date. It comes from when all namespace items could be used as a callable, including UDT's (referring to their constructor function). Moving forward, I don't think items other than callables should have schemes (and UDT's until we remove them).

ScottCarda-MS: The scheme describes the callable type associated with the item.

Issue #1268, opened 2024-03-18
Add samples of testing Q# code


tcNickolas:
Testing Q# code is an important part of developing a quantum project with QDK. We do most of our own testing (for example, for libraries) using Rust, so an external user working on their own project has very little information on how they can test their project using Python. 

Some things can be very challenging to figure out, for example, `dump_operation`: https://learn.microsoft.com/en-us/azure/quantum/user-guide/testing-debugging does not have information on it, and the only place I could find info on it was https://github.com/microsoft/qsharp/blob/d6a8f8a71a71bf80168e846c940674595c413a5b/pip/tests/test_qsharp.py#L103 which is helpful but 1) hard to discover without knowing upfront about `dump_operation` and 2) insufficient for developing practical tests (for example, it doesn't handle the global phase that can show up in the output).

We need to add tests for different scenarios to our samples folder (I'd create a separate folder for it called "testing", to match the naming logic used in "estimation" folder), to help users develop their own tests. Here are some examples I can think of right away:
* [x] Tests done mostly in Q# with a thin Python wrapper to integrate with pytest
   * verifying that measurement results of basis states are as expected
   * verifying classical computation
   * verifying that two operations are the same up to a global phase using `CheckOperationsAreEqual`
* [ ] Test that validates that Q# state is as expected using Python API `dump_machine`
* [x] Test that validates that Q# operation is as expected using Python API `dump_operation`
* We'll also need to make sure these tests are executed as part of the CI build - #1859

Manvi-Agrawal:
@tcNickolas , I would like to work on this issue

tcNickolas:
Sounds great!

There is a bit of information on testing in https://learn.microsoft.com/en-us/azure/quantum/user-guide/testing-debugging, but it's missing `dump_operation` (and is somewhat outdated, I think - `dump_machine` returns complex numbers for the past few releases). 

I also have some examples of pytest integration and using `dump_machine` and `dump_operation` for test at https://github.com/tcNickolas/quantum-programming-in-depth (qsharp-v1 folders), those use QDK 1.3.

Pulkit1822:
Hey @tcNickolas , I would also like to work on this issue.

tcNickolas:
Sorry, I think I missed the last comment on this issue. @Manvi-Agrawal, @Pulkit1822 - what's the status of this issue? Are either of you or both working on it? 

[...]

PR #1732, opened 2024-07-11
[Samples] Add samples for testing operations in Q#

Manvi-Agrawal:
Fixes part of #1268 
- Added samples for testing operations using `dump_operation`. 
- Also, using `CheckOperationsAreEqual` inside Fact function to demonstrate another way

- Hopefully, the new testing samples should be executed in CI. @tcNickolas could you please start a CI run. Snapshot from locally running `build.py`. 
image

- Running in CI:
image


samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
+**Copyright (c) Microsoft Corporation. All rights reserved.**  
+**Licensed under the MIT License.**
+
+# Testing Operations in Modern QDK
+This sample project demonstrates different approaches to testing operations in Modern QDK, which unlike Classic QDK, doesn't support native Q# tests.
```
swernli, line 5: So far we've avoided specific references to the Classic QDK in our guidance here, since this QDK is expected to stand on it's own and not require folks to be familiar with the Classic QDK. I'd rephrase this as:
```suggestion
# Testing Operations in the QDK
This sample project demonstrates different approaches to testing operations in the QDK, both via Python and within Q# itself.
```

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Testing Methods
+
+There are two primary ways to test operations in Modern QDK:
```
swernli, line 9: Same here, just refer to it as the QDK.
```suggestion
There are two primary ways to test operations in the QDK:
```

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Installation
+- Install the `qsharp` python package by following the instructions mentioned [here](https://learn.microsoft.com/azure/quantum/install-overview-qdk#add-support-for-python-and-jupyter-notebooks).
+- Install `pytest` python package.
+
+## Running the sample
+Open the `samples/testing/dump_operation` directory, and run `pytest` command.
+
+## Reference Links:
+- [Q# Testing guide](https://learn.microsoft.com/azure/quantum/user-guide/testing-debugging).
+- [Getting started with modern QDK](https://learn.microsoft.com/azure/quantum/install-overview-qdk)
```
swernli, line 36: Here too, this can just be the QDK.
```suggestion
- [Getting started with the QDK](https://learn.microsoft.com/azure/quantum/install-overview-qdk)
```

samples/testing/dump_operation/README.md
swernli: I don't think that is needed, the guidance you have put together here should be enough for anyone searching for information on testing.

samples/testing/dump_operation/src/SWAP.qs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    // # Summary
+    /// Operation for testing with `TestEquivalence` operation.
+    ///
+    /// # Input
+    /// ## qs
+    /// Input qubit register
+    operation ApplySWAP2(qs : Qubit[]) : Unit is Ctl + Adj {
+        SWAP(qs[0], qs[1]);
```
swernli, line 33: Perhaps it would be worth having the second SWAP implementation be the CNOT based decomposition? I'm not sure if this overcomplicates or distracts from the goal of the sample, so I'm fine if it stays the same for simplicity.

samples/testing/dump_operation/src/SWAP.qs
```diff
@@ -0,0 +1,35 @@
...
+///
+/// # Description
+/// Organizing code into multiple Q# source files is an important part of
+/// writing readable and maintainable code. In this project, we have `BellState.qs`, `SWAP.qs`,
+/// and `OperationEquivalence.qs`, which contain the operation to be tested.
+/// The presence of a Q# manifest file (`qsharp.json`) tells the compiler
+/// to include all Q# files under `src/`.
+/// These will be tested by Python wrapper, `test_dump_operation.py`
+
+namespace SWAP {
```
swernli, line 15: This is a bit of a nitpick, but it can be a little confusing when the namespace has the same name as an intrinsic gate. Maybe `Test_SWAP` could avoid any conceptual overlap with the operation `SWAP`?

swernli (approved):
A few more minor comments, but otherwise looks good!

build.py
```diff
@@ -363,6 +363,43 @@ def run_python_integration_tests(cwd, interpreter):
     )
     step_end()
 
+if build_samples:
```
swernli, line 366: As of https://github.com/microsoft/qsharp/pull/1739 all samples testing now occurs via `cargo test` and auto-generated Rust unit tests, so this section was deleted. I think the automatic merge may have reintroduced it. You can delete it here and confirm the added project is automatically picked up and compiled as part of the Rust tests.

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Testing Operations in the QDK
+This sample project demonstrates different approaches to testing operations in the QDK, both via Python and within Q# itself.
+
+## Testing Methods
+
+There are two primary ways to test operations in the QDK:
+
+1. **Dumping Operation Output:**
+   - Use the `dump_operation` Python API to retrieve the operation's representation and compare it against the expected output.
```
tcNickolas, line 12: Technically quantum operations don't have output, so this phrasing is confusing. I'd clarify that it's matrix representation instead
```suggestion
1. **Using Operation Matrix Representation:**
   - Use the `dump_operation` Python API to retrieve the operation's representation as a matrix and compare it against the expected matrix.
```

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+2. **Q# `Fact` Assertions:**
+   - Define a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical. The `Fact` function asserts that the check returns `true`.
```
tcNickolas, line 15:
```suggestion
   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical. The `Fact` function asserts that the check returns `true`.
```

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Project Structure
+This sample project is a multi-file Q# project that showcases both testing methods. The project structure is as follows:
+
+- src
+    - "BellState.qs": Q# file containing the `AllBellStates` operation to be tested
+    - "Test_SWAP.qs": Q# file containing the `ApplySWAP1` and `ApplySWAP2` operations to be tested
+    - "OperationEquivalence.qs": Q# file containing the `TestEquivalence` operation to be called in python wrapper
+- "qsharp.json": Q# project configuration file
+- test_dump_operation.py: "Python wrapper containing tests"
```
tcNickolas, line 25: This list uses inconsistent formatting for file names (using quotes or not) and text (last line has the next explanation in quotes). Let's unify; I don't think we need quotes for file names, if we want to highlight them, ` works better, and we don't need quotes around text

samples/testing/dump_operation/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Installation
+- Install the `qsharp` python package by following the instructions mentioned [here](https://learn.microsoft.com/azure/quantum/install-overview-qdk#add-support-for-python-and-jupyter-notebooks).
+- Install `pytest` python package.
+
+## Running the sample
+Open the `samples/testing/dump_operation` directory, and run `pytest` command.
```
tcNickolas, line 32: The directory name emphasizes one of the operations used in testing, but since the project has grown to show two approaches, I would rename it to emphasize the kind of thing we test instead - call it just "operations", so that the whole path says "testing operations"

samples/testing/dump_operation/src/Test_SWAP.qs
```diff
@@ -0,0 +1,37 @@
...
+///
+/// # Description
+/// Organizing code into multiple Q# source files is an important part of
+/// writing readable and maintainable code. In this project, we have `BellState.qs`, `Test_SWAP.qs`,
+/// and `OperationEquivalence.qs`, which contain the operation to be tested.
+/// The presence of a Q# manifest file (`qsharp.json`) tells the compiler
+/// to include all Q# files under `src/`.
+/// These will be tested by Python wrapper, `test_dump_operation.py`
+
+namespace Test_SWAP {
```
tcNickolas, line 15: This file contains the main code that is tested rather than the test code, so I would not call the namespace and the file starting with Test. It's really confusing that it uses Test in the name and the file that has the test logic does not!

samples/testing/dump_operation/test_dump_operation.py
```diff
@@ -0,0 +1,119 @@
...
+def test_single_qubit_not_gate() -> None:
+    qsharp.init(target_profile=qsharp.TargetProfile.Unrestricted)
+
+    res = dump_operation("qs => X(qs[0])", 1)
+    assert res == [
+        [complex(0.0, 0.0), complex(1.0, 0.0)],
+        [complex(1.0, 0.0), complex(0.0, 0.0)],
+    ]
+
+def test_single_qubit_superposition() -> None:
```
tcNickolas, line 26: This test is named differently from the others, I'd suggest unified naming scheme
```suggestion
def test_single_qubit_hadamard_gate() -> None:
```

samples/testing/dump_operation/src/OperationEquivalence.qs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OperationEquivalence {
+    open Microsoft.Quantum.Diagnostics;
+    open Test_SWAP;
+    /// # Summary
+    /// Verifies the equivalence of quantum operations using `Fact` function
+    /// and the `CheckOperationsAreEqual` operation.
+    operation TestEquivalence() : Unit {
+        let actual = qs => Test_SWAP.ApplySWAP1(qs);
+        let expected = qs => Test_SWAP.ApplySWAP2(qs);
```
tcNickolas, line 25: You shouldn't need lambdas here, since you're not defining new operations using partial application or applying the operations to a part of an array, just saying actual = ApplySWAP1 should work

samples/testing/dump_operation/src/OperationEquivalence.qs
```diff
@@ -0,0 +1,28 @@
...
+        Fact(CheckOperationsAreEqual(2, actual, expected) == true, "Actual and expected operation should be same");
```
tcNickolas, line 26: You don't need to compare a Boolean to true to convert it to a Boolean

samples/testing/dump_operation/src/OperationEquivalence.qs
```diff
@@ -0,0 +1,28 @@
...
+///
+/// # Description
+/// This code builds upon the concepts explained in the README file,
+/// demonstrating how to organize Q# code into multiple files for testing.
+/// Here, we have separate files (`BellState.qs`, `Test_SWAP.qs`, and `OperationEquivalence.qs`)
+/// containing individual operations under test.
+
+/// The presence of a Q# manifest file (`qsharp.json`) instructs the compiler
+/// to include all Q# files within the `src` directory. These operations are tested
+/// by the Python wrapper script, `test_dump_operation.py`.
```
tcNickolas, line 15: For a multi-file project, I would keep the description of the project file structure to the README file, and use summary of the specific files within it to describe just the content of that file. This way, there is no information duplication, the files look cleaner, and it's easier to update them if we rename a file or add an extra one

samples/testing/dump_operation/test_dump_operation.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
tcNickolas, line 1: To match the project renaming suggestion, I'd rename this file to "test_operations.py", since it showcases both approaches

samples/testing/dump_operation/test_dump_operation.py
```diff
@@ -0,0 +1,119 @@
...
+# Licensed under the MIT License.
+
+import qsharp
+from qsharp.utils import dump_operation
+
+def test_empty_operation() -> None:
+    qsharp.init(target_profile=qsharp.TargetProfile.Unrestricted)
```
tcNickolas, line 8: Do we need to set this profile explicitly every time? I think it's the default setting, my tests get away with not using it, just saying qsharp.init()

samples/testing/dump_operation/test_dump_operation.py
```diff
@@ -0,0 +1,119 @@
...
+    
+    res = dump_operation("qs => ()", 1)
+    assert res == [
+        [complex(1.0, 0.0), complex(0.0, 0.0)],
```
tcNickolas, line 12: Do you need to call complex explicitly, or can you compare the numbers with just [1, 0], [0, 1]? I think Python is nice enough about type casting to allow to compare floating point with complex directly, I do that in my tests

tcNickolas (changes requested):
Looks good! I like the variety of tests you have, including the ones showing partial application to test operations with parameters!

I left suggestions for code and docs readability.

samples/testing/operations/src/Test_SWAP.qs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+
+/// # Sample
+/// Multi File Testing Project
+///
+/// # Description
+/// This code builds upon the concepts explained in the README file,
+/// demonstrating how to organize Q# code into multiple files for testing.
+
+namespace Test_SWAP {
```
tcNickolas, line 11: The previous comment chain disappeared, so here's a new one: I agree that it's good to name namespaces with names distinct from the operations defined in them, but it's also good not to put non-test logic in namespace named "test". We can use "CustomOperation" or "CustomSWAP" to satisfy both

swernli:
Thanks for merging main and confirming things work well with the new sample testing. I can see the auto-generated tests run successfully as well as the new tests added here. It looks great!

samples/testing/operations/test_operations.py
```diff
@@ -0,0 +1,127 @@
...
+import qsharp
+from qsharp.utils import dump_operation
+
+@pytest.fixture(autouse=True)
+def setup():
+    """Fixture to execute asserts before a test is run"""
+    # You can set the target profile here if you want
+    # By default, it uses "Unrestricted" profile
+    qsharp.init(target_profile=qsharp.TargetProfile.Base)
+    yield # this is where the testing happens
```
tcNickolas, line 14: Several questions on this piece:
* The doc comment mentions asserts but the code doesn't actually include asserts
* I understand that this executes qsharp.init before every test, correct? But all tests except the first one call qsharp.init anyways, re-initializing qsharp again and not benefitting from this fixture?
* You shouldn't need to specify profile at all, the tests can run without that, and here you want to have the minimal test code that achieves the goal.
* I would swap the init argument for `project_root='.'`, since you do actually need that for the later tests and they don't work without it. Right now, the later tests call init three times, one from setup, one with setting target profile, and one with setting project root, and I think only the last one is actually used

samples/testing/operations/src/BellState.qs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// # Sample
+/// Multi File Testing Project (Refer to README for Project Overview)
+///
+/// # Description
+/// This code builds upon the concepts explained in the README file,
+/// demonstrating how to organize Q# code into multiple files for testing.
+
```
tcNickolas, line 10: We do not need to repeat the reference to the README file in every Q# file. A README file in the root folder implies that it has the descriptions that are relevant to all files, and each file needs to include only its own documentation. Please remove this summary from this file and the other two Q# files in this folder

samples/testing/operations/src/BellState.qs
```diff
@@ -0,0 +1,42 @@
...
+    /// # Summary
+    /// Operation that generates all bell states for testing with `dump_operation.py`.
+    ///
+    /// # Input
+    /// ## qs
+    /// Input qubit register
+    ///
+    /// ## choice
+    /// Bell state to construct.
+    /// 0: |Φ+〉(PhiPlus)
```
tcNickolas, line 21: I think 〉 is the ket symbol that looks weird in some environments (https://github.com/microsoft/QuantumKatas/issues/4 was actually one of the first issues on the original katas!), and we try to use ⟩ instead.

If you want to go all the way to fancy with Unicode ket notations, Unicode has subscripts for + and - (https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts), Bell state notations look better with them, since regular plus and minus look a bit like tensor products of |Φ⟩ with |+⟩ and |-⟩

samples/testing/operations/src/OperationEquivalence.qs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// # Description
+/// This code builds upon the concepts explained in the README file,
+/// demonstrating how to organize Q# code into multiple files for testing.
+
+namespace OperationEquivalence {
+    open Microsoft.Quantum.Diagnostics;
+    open CustomOperation;
+    /// # Summary
+    /// Verifies the equivalence of quantum operations using `Fact` function
```
tcNickolas, line 15: Here, we need to mention that this check compares two operations up to a global phase. This is important, for example, if the user tries to hunt down a global phase issue. The other approach to testing doesn't have this issue

samples/testing/operations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+2. **Q# `Fact` Assertions:**
+   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical. The `Fact` function asserts that the check returns `true`.
+
+## Project Structure
+This sample project is a multi-file Q# project that showcases both testing methods. The project structure is as follows:
+
+- src
+    - `BellState.qs`: Q# file containing the `AllBellStates` operation to be tested
+    - `Test_SWAP.qs`: Q# file containing the `ApplySWAP1` and `ApplySWAP2` operations to be tested
```
tcNickolas, line 19: Need to update doc to match the code

samples/testing/operations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Testing Methods
+
+There are two primary ways to test operations in the QDK:
+
+1. **Using Operation Matrix Representation:**
+   - Use the `dump_operation` Python API to retrieve the operation's representation as a matrix and compare it against the expected matrix.
+
+2. **Q# `Fact` Assertions:**
+   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical. The `Fact` function asserts that the check returns `true`.
```
tcNickolas, line 12: Need to mention the "up to the global phase" part

samples/testing/operations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Project Structure
+This sample project is a multi-file Q# project that showcases both testing methods. The project structure is as follows:
+
+- src
+    - `BellState.qs`: Q# file containing the `AllBellStates` operation to be tested
+    - `Test_SWAP.qs`: Q# file containing the `ApplySWAP1` and `ApplySWAP2` operations to be tested
+    - `OperationEquivalence.qs`: Q# file containing the `TestEquivalence` operation to be called in python wrapper
```
tcNickolas, line 20: Python should be capitalized when used in a sentence (as opposed to python command line)

tcNickolas (changes requested):
This looks a lot better, I like how the code got more concise and matrices in the test are human-readable!

I left some more comments, with them fixed this will be good to go!

samples/testing/operations/src/BellState.qs
tcNickolas: That sample is actually a great example of why + and - in Bell state variables should be subscripts rather than regular symbols - it uses |Φ+⟩ (not a tensor product) side by side with |+0〉 (tensor product).

I'm happy to table this discussion for now, though. I've started an internal discussion about unifying Unicode in the samples

samples/testing/operations/README.md
```diff
@@ -9,21 +9,21 @@ There are two primary ways to test operations in the QDK:
    - Use the `dump_operation` Python API to retrieve the operation's representation as a matrix and compare it against the expected matrix.
 
 2. **Q# `Fact` Assertions:**
-   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical. The `Fact` function asserts that the check returns `true`.
+   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical upto a global phase. The `Fact` function asserts that the check returns `true`.
```
tcNickolas, line 12:
```suggestion
   - Use a `Fact` function in your Q# code that uses the `CheckOperationsAreEqual` operation to verify if two operations are identical up to a global phase. The `Fact` function asserts that the check returns `true`.
```

samples/testing/operations/README.md
```diff
@@ -9,21 +9,21 @@ There are two primary ways to test operations in the QDK:
...
 
 ## Project Structure
 This sample project is a multi-file Q# project that showcases both testing methods. The project structure is as follows:
 
 - src
     - `BellState.qs`: Q# file containing the `AllBellStates` operation to be tested
-    - `Test_SWAP.qs`: Q# file containing the `ApplySWAP1` and `ApplySWAP2` operations to be tested
-    - `OperationEquivalence.qs`: Q# file containing the `TestEquivalence` operation to be called in python wrapper
+    - `CustomOperation.qs`: Q# file containing the `ApplySWAP1` and `ApplySWAP2` operations to be tested
```
tcNickolas, line 19:
```suggestion
    - `CustomOperation.qs`: Q# file containing the `ApplySWAP` operation to be tested
```

samples/testing/operations/src/OperationEquivalence.qs
```diff
@@ -1,20 +1,13 @@
...
-/// # Description
-/// This code builds upon the concepts explained in the README file,
-/// demonstrating how to organize Q# code into multiple files for testing.
-
 namespace OperationEquivalence {
     open Microsoft.Quantum.Diagnostics;
     open CustomOperation;
     /// # Summary
-    /// Verifies the equivalence of quantum operations using `Fact` function
+    /// Verifies the equivalence of quantum operations upto a global phase using `Fact` function
```
tcNickolas, line 8:
```suggestion
    /// Verifies the equivalence of quantum operations up to a global phase using `Fact` function
```

samples/testing/operations/test_operations.py
```diff
@@ -66,7 +57,6 @@ def test_custom_operation() -> None:
     ]
 
 def test_operation_no_args_in_qsharp_file() -> None:
-    qsharp.init(target_profile=qsharp.TargetProfile.Unrestricted)
     qsharp.init(project_root='.')
```
tcNickolas, line 60: Same comment, since we have a fixture that does exactly the same thing, we don't need to repeat .init in the individual tests. Please remove duplicate inits

tcNickolas (changes requested):
Looks good. Hopefully one last round of suggestions!

Thank you!

tcNickolas (approved):
Looks great! Thank you so much for working on this!

PR #1736, opened 2024-07-12
Allow re-exporting an import which has an alias to an item defined within the same package.

sezna:
Yes, that title is a mouthful. This is _quite_ the edge case identified by @ScottCarda-MS  -- thank you for bug bashing!

Re-exports are nuanced. They're basically pointers to `ItemId`s which point to another package. As a refresher, if an `ItemId` has a package id (`item_id.package.is_some()`), then it refers to an item outside of the current package, so it is thusly a reexport. If `item_id.package.is_none()`, then the item is local to the current package.

Based on this preexisting notion, re-exports were supported by inserting `ItemKind::ExportedItem` for re-exports which point to another package. These are identified in `add_external_package` and inserted into the names of the user code for resolution. And if an item was local, then no `ExportedItem` is needed -- we just mark the item's `Visibility` as public. 

But what, what if that local item is exported with an alias? Then, we need some way to mark it as public _and_  tell the HIR that it has a different name. To handle this, we insert an `ExportedItem` if an item is local and the export has an alias.

So, to recap, an `ItemKind::ExportedItem` is inserted if the item being exported is either from another package (a re-export) or from the local package, but aliased. Everything mentioned up until now is handled in #1727.

This PR fixes a bug where a re-export of a local item which is introduced to the scope via an import alias loses its alias upon exporting. Wow, what a mouthful. Basically, in order to insert an `ItemKind::ExportedItem` for an export with an alias, we just checked if `export.alias.is_some()`. But, actually, you can also basically perform an aliased export like this: `import Foo as Bar; export Bar;`. In this case, there is indeed no export on the alias. But we do still need an `ItemKind::ExportedItem`, because we need to track that this same `ItemId` is accessible via both names -- the original declared name and the re-exported aliased name.

### DIFF
--- a/compiler/qsc_frontend/src/lower.rs
+++ b/compiler/qsc_frontend/src/lower.rs
@@ -145,9 +145,9 @@ impl With<'_> {
             .iter()
             .filter_map(|(res, alias)| match res {
                 resolve::Res::ExportedItem(id, hir_alias) => {
-                    Some((*id, hir_alias.clone().or(alias.clone())))
+                    Some((*id, hir_alias.as_ref().or(alias.as_ref())))
                 }
-                resolve::Res::Item(id, _) => Some((*id, alias.clone())),
+                resolve::Res::Item(id, _) => Some((*id, alias.as_ref())),
                 _ => None,
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


compiler/qsc_frontend/src/lower.rs
```diff
@@ -144,7 +144,10 @@ impl With<'_> {
         let exported_hir_ids = exports
             .iter()
             .filter_map(|(res, alias)| match res {
-                resolve::Res::ExportedItem(id) | resolve::Res::Item(id, _) => Some((*id, alias)),
+                resolve::Res::ExportedItem(id, hir_alias) => {
+                    Some((*id, hir_alias.clone().or(alias.clone())))
```
swernli, line 148: I haven't tried this locally, but I'm wondering if this could be adjusted to avoid the appearance of a double clone with something like `hir_alias.or(alias).clone()`

PR #1737, opened 2024-07-12
Use `PackageType::Lib` by default

swernli:
In most places, we already used the `PackageType::Lib` by default, but not in the language service. With more library authoring expected in the future, this tweaks the user experience to avoid errors for missing entry point until the Q# program as run. As part of these changes, the behavior of the entry point check is updated to always provide feedback for other entry point related errors (like duplicates) but only return an error on missing entry point if the compilation is for an executable.

Of note, this does NOT remove the configuration option for `packageType` in the qsharp.json, which we can do as a follow up if needed.

This does update the error string that pops up when failing to run/debug the Q#. If the program is missing an entry point, a more specific error message is returned:
image

and the generic compilation failure message is also updated:
image

### DIFF
--- a/compiler/qsc_passes/src/entry_point.rs
+++ b/compiler/qsc_passes/src/entry_point.rs
@@ -4,6 +4,8 @@
 #[cfg(test)]
 mod tests;
 
+use crate::PackageType;
+
 use super::Error as PassErr;
 use miette::Diagnostic;
 use qsc_data_structures::span::Span;
@@ -44,13 +46,14 @@ pub enum Error {
 pub(super) fn generate_entry_expr(
     package: &mut Package,
     assigner: &mut Assigner,
+    package_type: PackageType,
 ) -> Vec<super::Error> {
     if package.entry.is_some() {
         return vec![];
     }
     let callables = get_callables(package);
 
-    match create_entry_from_callables(assigner, callables) {
+    match create_entry_from_callables(assigner, callables, package_type) {
         Ok(expr) => {
             package.entry = Some(expr);
             vec![]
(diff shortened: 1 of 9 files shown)


compiler/qsc_passes/src/entry_point.rs
```diff
@@ -44,13 +46,14 @@ pub enum Error {
...
     assigner: &mut Assigner,
+    package_type: PackageType,
 ) -> Vec<super::Error> {
     if package.entry.is_some() {
         return vec![];
     }
     let callables = get_callables(package);
 
-    match create_entry_from_callables(assigner, callables) {
+    match create_entry_from_callables(assigner, callables, package_type) {
```
billti, line 56: I'm surprised to see PackageType wired through in more places now. I had expected PackageType would just go away, and we now just have projects, and in a project you _may_ add an entry point attribute to an operation, and if you try to directly run a project with a) No entrypoint attribute, b)  No 'Main' operation, and c) No entry expression provided (such as in a launch.json or such), then you get a runtime/launch error.

Or is my overly simplified view of the world an unrealistic fantasy :-)

billti: (I guess for compilation you need to locate it also)

swernli: We still need to have the concept internally of "compile and require an entry point, give me a nice error if you can't find one" for run and debug in VS Code and execution in playground/QCOM editors. Everything else can use Lib, but those are still Exe. So the result of this work is the change in user experience and opening the door to removing the setting from the qsharp.json so users never need to know the concept, but the compiler still needs the distinction.

For clarity, we could in a future refactoring change the name of the variables and the enum to indicate the simple thing this represents: is entry point required. But with this proposed change that refactor becomes an internal design question rather than a user experience question.

billti (approved):

PR #1738, opened 2024-07-12
Playground doesn't support implicit namespaces

### LINKED ISSUE #1653: Playground doesn't support implicit namespaces
It appears that  the playground is assigning numbers to sources even though the name `"code"` is passed in as the source name? I haven't dug in too deeply.

image

Pretty simple repro, just use an implicit namespace on the playground.

swernli:
Fixes #1653

### DIFF
--- a/playground/src/editor.tsx
+++ b/playground/src/editor.tsx
@@ -241,7 +241,16 @@ export function Editor(props: {
     });
 
     editor.current = newEditor;
-    const srcModel = monaco.editor.createModel(props.code, "qsharp");
+    const srcModel =
+      monaco.editor.getModel(
+        monaco.Uri.parse(props.kataExercise?.id ?? "main.qs"),
+      ) ??
+      monaco.editor.createModel(
+        "",
+        "qsharp",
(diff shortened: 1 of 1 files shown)


ScottCarda-MS (approved):
Seem to work for me when I tested it out.

playground/src/editor.tsx
```diff
@@ -241,7 +241,16 @@ export function Editor(props: {
...
     editor.current = newEditor;
-    const srcModel = monaco.editor.createModel(props.code, "qsharp");
+    const srcModel =
+      monaco.editor.getModel(
+        monaco.Uri.parse(props.kataExercise?.id ?? "main.qs"),
+      ) ??
+      monaco.editor.createModel(
+        "",
+        "qsharp",
+        monaco.Uri.parse(props.kataExercise?.id ?? "main.qs"),
```
billti, line 251: Look at you with all your fancy JS operators. We'll make a TypeScript dev out of you yet! :wink:

billti (approved):

PR #1739, opened 2024-07-15
Move all samples testing to Rust unit tests

swernli:
This expands the unit tests in the samples_test folder to include testing of all samples, including compilation of resource estimation samples, compilation of dynamically discovered qsharp.json projects, and compile and run of language samples. With these additions, the build script no longer needs to compile the qsc cmdline tool and can skip the cmdline based compilation of samples, so the build.py file is updated here to remove those.


idavis:
@swernli How do I debug a failing test?

swernli:
> @swernli How do I debug a failing test?

You can debug each module, which will launch the debugger for the tests covered by that module (each test in it's own thread):
image

You can also ctl/cmd + click on the file name in the include macro to jump to the build-time generated file and set breakpoints inside of individual test cases.

swernli:
I changed the file generation strategy to split the difference between these being pure build artifacts and being checked in source files. Now the test files are generated to the src folder so they can used and debugged like normal tests, but they are added to the gitignore so that aren't part of the source-controlled content.

idavis (approved):

PR #1740, opened 2024-07-15
VSCode shows documentation by package

DmitryVasilevsky:
Changed the way generated API documentation is shown in vscode. Also affects published documentation.
* Items are sorted by package (current project first, then direct dependencies, then standard library, then core library), then by namespace, then by item name.
* Fully qualified names are shown for items instead of just item names. Fully qualified names include package alias (if any), namespace and item name. For example, Length is shown as Microsoft.Quantum.Core.Length.
* As namespace and package alias is now part of the title, a separate line with the namespace is removed.

This change does NOT address https://github.com/microsoft/qsharp/issues/1823


compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -140,51 +158,110 @@ pub fn generate_docs(
...
+        (
+            file.0.package.clone(),
+            file.0.namespace.clone(),
+            file.0.name.clone(),
+        )
+    });
+
+    let mut result: Files = files
+        .into_iter()
+        .map(|(metadata, name, content)| (name, Arc::from(metadata.to_string().as_str()), content))
```
ScottCarda-MS, line 227: Files and FilesWithMetadata are both maintained only in this file. In this conversion, I see the ordering of their fields differs. It would be good if they had the same ordering of fields, as that would make reasoning between the two a little more straightforward.

ScottCarda-MS (approved):

billti:
Playing around with some sample projects, and I'm not sure we need the type to be part of the header (e.g. 'operation' or 'user defined type'). Especially as we show the definition immediately below the header anyway

image

billti:
p.s. All the docs still use the `Microsoft.` namespace also. Where did we land on either adding or migrating to `std` for the docs? CC @sezna

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -140,51 +158,110 @@ pub fn generate_docs(
...
         compilation: &compilation,
     };
 
     let mut toc: FxHashMap<Rc<str>, Vec<String>> = FxHashMap::default();
-    for (_, unit) in &compilation.package_store {
+
+    for (package_id, unit) in &compilation.package_store {
+        let is_current_package = compilation.current_package_id == Some(package_id);
+        let package_kind;
+        match package_id {
```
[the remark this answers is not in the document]
DmitryVasilevsky: I don't like to mix assignment with the 'continue' even if it's allowed. I'll change match to if-else chain.

compiler/qsc_doc_gen/src/generate_docs.rs
DmitryVasilevsky: Updated.

compiler/qsc_doc_gen/src/generate_docs/tests.rs
```diff
@@ -27,9 +27,7 @@ fn docs_generation() {
         qsharp.summary: "Returns the number of elements in the input array `a`."
         ---
 
-        # Length function
-
-        Namespace: Microsoft.Quantum.Core
+        # Microsoft.Quantum.Core.Length function
```
billti, line 30: Why did we change the primary heading to include the full namespace? The main heading reads cleaner when shorter, and including the namespace makes it a little awkward when shortly we want to add another namespace in the docs (e.g. mention it is also available in the `std.` namespace).

billti: With the docs popping up in a side pane, making the header really long by including the namespace can be pretty ugly.

image

We can live with it for now to get 1.8 out, but please open a tracking issue to improve by changing it back the way it was for a follow up release.

DmitryVasilevsky: I believe this was the result of the discussion on what to do with the namespaces and package aliases. @swernli , if I'm not mistaken, you were the main proponent of this, but maybe I'm wrong. I'm don't have strong opinion if that should be included in the title or not. But if we don't include fully qualified name in the title, we should include it somewhere else in the description - and that part we couldn't agree on. Any suggestions?

billti: > But if we don't include fully qualified name in the title, we should include it somewhere else in the description - and that part we couldn't agree on.

The namespace used to be right underneath it. You can see it was deleted in the diff.

Anyways, as mentioned, just resolve this comment and merge for now. There's some other clean-up in this area and the 'std' addition to figure out, so we'll revisit later. Thanks.

DmitryVasilevsky: > The namespace used to be right underneath it. You can see it was deleted in the diff.

Yes. But now it's not just namespace, it's also the package alias (when it is present). And if you use the same library in two different projects with two different aliases, saying just 'namespace' could be confusing. Now we could say something like 'fully qualified name', but that includes the name as well.

It's actually easy to change this now and still open an issue.

DmitryVasilevsky: I updated the code to include only the item name in the title and fully qualified name in the body of the doc.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -211,17 +283,23 @@ fn get_namespace(package: &Package, item: &Item) -> Option<Rc<str>> {
...
     let doc = increase_header_level(&item.doc);
     let title = &metadata.title;
+    let fqn = &metadata.fully_qualified_name();
     let sig = &metadata.signature;
 
     let content = format!(
         "# {title}
 
-Namespace: {ns}
+Fully qualified name: {fqn}
```
sezna, line 302: approving for now, but did we decide to also include "also available as Std.___" in a future change?

sezna (approved):

compiler/qsc_doc_gen/src/generate_docs.rs
DmitryVasilevsky: Yes. I will open an issue for that.

DmitryVasilevsky: https://github.com/microsoft/qsharp/issues/1866

PR #1744, opened 2024-07-16
Noisy Simulator: fix clippy lints, better docs, add unit tests, bug fix

orpuente-MS:
This PR contains some general maintenance for the `noisy_simulator` which is still in an experimental state. It contains:
 - Clippy lints fixes.
 - Better docstrings.
 - More unit tests.
 - A bug fix in `noisy_simulator/src/state_vector_simulator.rs: Line 58`.
 - Changes the Operation __init__ type hint in the python API to take `Any` instead of `List[List[List[complex]]]`.


noisy_simulator/src/instrument.rs
```diff
@@ -102,7 +108,8 @@ fn summed_kraus_operators(operations: &[Operation]) -> Result<Vec<SquareMatrix>,
                 .iter()
                 .map(|k| {
                     // This code is doing the equivalent to:
-                    // choi_matrix += vec(K) * vec(K)^dagger
+                    // choi_matrix += vec(K) * vec(K*)
+                    // where * denotes the complex conjugate
```
swernli, line 112: Nitpick: it seems a little odd to use `*` for both multiplication and conjugate. Another option to consider would be `Kdg` or `Kadj` for dagger and adjoint respectively. I've seen both used in a lot of code, and I tend toward the adjoint variation myself, though either would be fine.

swernli (approved):
Changed look good. One minor quirk: I think this needs a rebase on main that drops the individual commits from the initial noisy simulator work. The PR shows 78 commits and many of them are from 20 days ago in your previous branch.

orpuente-MS:
> Changes look good. One minor quirk: I think this needs a rebase on main that drops the individual commits from the initial noisy simulator work. The PR shows 78 commits and many of them are from 20 days ago in your previous branch.

Thanks for the help with this!

noisy_simulator/src/instrument.rs
```diff
@@ -102,7 +108,9 @@ fn summed_kraus_operators(operations: &[Operation]) -> Result<Vec<SquareMatrix>,
                 .iter()
                 .map(|k| {
                     // This code is doing the equivalent to:
-                    // choi_matrix += vec(K) * vec(K)^dagger
+                    // choi_matrix += vectorized(K) * vectorized(K).adjoint()
+                    // Note that if you multiply a row vector times a column
```
billti, line 112: Is this back-to-front? I thought if you multiple a column vector by a row vector you get a matrix (i.e, a density matrix is `|ψ><ψ|`), but a row vector times a column vector is like an inner product and gives you a scalar.

noisy_simulator/src/state_vector_simulator.rs
```diff
@@ -55,7 +55,7 @@ impl StateVector {
             )));
         }
 
-        if data.len() != dimension * dimension {
+        if data.len() != dimension {
```
billti, line 58: Was this just a bug before then being fixed here? How did it get missed (and found)?

billti (approved):

noisy_simulator/src/state_vector_simulator.rs
orpuente-MS: It was missed because this path only gets hit when you are using it from Python and save the state to a variable, and then set back that state to the simulator again. And I was missing a unit test for that.

noisy_simulator/src/instrument.rs
orpuente-MS: Ah, yes, you are right! Nice catch.

PR #1747, opened 2024-07-17
Fire telemetry event when a github dependency is fetched

sezna:
As the title says -- simple enough.

### DIFF
--- a/vscode/src/projectSystem.ts
+++ b/vscode/src/projectSystem.ts
@@ -248,6 +248,9 @@ export async function fetchGithubRaw(
   const uri = `${githubEndpoint}/${owner}/${repo}/${ref}/${pathNoLeadingSlash}`;
   log.info(`making request to ${uri}`);
   const response = await fetch(uri);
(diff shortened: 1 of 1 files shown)


vscode/src/projectSystem.ts
```diff
@@ -247,6 +248,11 @@ export async function fetchGithubRaw(
   const uri = `${githubEndpoint}/${owner}/${repo}/${ref}/${pathNoLeadingSlash}`;
   log.info(`making request to ${uri}`);
   const response = await fetch(uri);
```
billti, line 250: Network errors (including CORS) cause the `await fetch` to throw, in which case it will never get to the sendTelemetryEvent line.

That said, we only support GitHub (which should never see a CORS error), and if there are network issues telemetry probably wont send either, so may be fine. Maybe worth a comment to note for future readers though.

billti (approved):

ScottCarda-MS (approved):

PR #1754, opened 2024-07-19
Avoid spurious errors in Q# git diffs

swernli:
This change updates the logic in `isQsharpDocument` to make sure the language service doesn't generate spurious errors from treating the left-hand side of the diff as a Q# file.

### DIFF
--- a/vscode/src/common.ts
+++ b/vscode/src/common.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { TextDocument, Uri, Range, Location } from "vscode";
+import { Utils } from "vscode-uri";
 import { ILocation, IRange, IWorkspaceEdit } from "qsharp-lang";
 import * as vscode from "vscode";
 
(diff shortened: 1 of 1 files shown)


vscode/src/common.ts
```diff
@@ -2,14 +2,19 @@
...
 
 export const qsharpLanguageId = "qsharp";
 
 // Returns true for all Q# documents, including unsaved files, notebook cells, etc.
 export function isQsharpDocument(document: TextDocument): boolean {
-  return document.languageId === qsharpLanguageId;
+  return (
+    document.languageId === qsharpLanguageId &&
+    (Utils.extname(document.uri) === ".qs" || document.isUntitled) &&
+    document.uri.scheme !== "git"
```
billti, line 16: In many language service the language service does work fine in a diff view, e.g. Rust as shown below. I use this quite a bit when looking at PRs to navigate through code. Does this last line break that?

image

swernli: Thankfully, it does. I'm not 100% sure how VSCode identifies the difference between the two sides, but I believe it has to do with the `git` scheme that I check for which lets each side be treated as a separate buffer (the right side does not appear to have this scheme and is instead a view onto the on-disk file). The change made it so that the left-hand side no longer produces the errors while the right-hand side behaves like a normal editor (including run/debug support, but without code lenses, which matches my experience with Rust).

Before the change:

image

After the change:

image

I tested in the local web editor environment (the one you get when running `npm start` in the vscode folder) and the behavior seemed good. What I'm not sure of is how this behaves on github.dev; hopefully that environment doesn't play fast and loose with that `git` URI scheme.

billti (approved):

PR #1756, opened 2024-07-19
Completions use `import` instead of `open`; Make completions `Main`-shortcut aware

sezna:
In the below video, you'll notice:
1. The namespace `Main` is omitted from auto-imports when it is in a dependency
2. We use auto-imports instead of auto-opens
3. I also found a bug where completions wouldn't populate if your namespace is the same as the namespace you're importing from. This made sense before we had external packages, but now, it is very common that you'll be in `Main` and importing something from another namespace called `Main`

https://github.com/user-attachments/assets/0508ceef-2b20-4ffb-a52d-5417cc348bc2

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -615,12 +615,14 @@ fn format_external_name(
         vec![]
     };
 
-    if package_alias_from_manifest.is_some()
+    // if this comes from an external project's Main, then the path does not include Main
+    let item_comes_from_main_of_external_project = package_alias_from_manifest.is_some()
         && qualification.len() == 1
-        && qualification.first() == Some(&"Main".into())
-    {
-        fully_qualified_name.append(&mut qualification[1..].to_vec());
-    } else {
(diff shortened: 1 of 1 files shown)


language_service/src/completion.rs
```diff
@@ -597,22 +589,48 @@ impl CompletionListBuilder {
...
+        vec![Rc::from(&*alias.clone())]
+    } else {
+        vec![]
+    };
+
+    if package_alias_from_manifest.is_some()
+        && qualification.len() == 1
+        && qualification.first() == Some(&"Main".into())
+    {
+        fully_qualified_name.append(&mut qualification[1..].to_vec());
```
ScottCarda-MS, line 622: Since this is already predicated on qualification.len() == 1, this line will do nothing. I think you can just append the full qualification slice when the above condition is false.

ScottCarda-MS (approved):

swernli (approved):

PR #1759, opened 2024-07-19
Katas UI string review - QRNG

SoniaLopezBravo:


katas/content/random_numbers/random_number/solution.md
```diff
@@ -1,6 +1,6 @@
-We can reuse the `RandomBit` and `RandomNBits` operations from earlier exercises.
+You can reuse the `RandomBit` and `RandomNBits` operations from earlier exercises.
 
-We'll generate an $N$-bit random number by calling the `RandomNBits` operation, where $N$ is the bitsize of $max - min$. We can repeat this process until the result is less than or equal than $max - min$, and return that number plus $min$.
+YOU'll generate an $N$-bit random number by calling the `RandomNBits` operation, where $N$ is the bitsize of $max - min$. You can repeat this process until the result is less than or equal than $max - min$, and return that number plus $min$.
```
billti, line 3: Why is `YOU` capitalized? Typo maybe.

SoniaLopezBravo:
```suggestion
You'll generate an $N$-bit random number by calling the `RandomNBits` operation, where $N$ is the bitsize of $max - min$. You can repeat this process until the result is less than or equal than $max - min$, and return that number plus $min$.
```

katas/content/random_numbers/index.md
```diff
@@ -16,14 +16,14 @@ The quantum algorithm for random number generation is one of the simplest applic
 
 **What you should know to start working on this kata:**
 
-- The concept of qubit and measurement
-- Single-qubit gates
+- Basic knowledge of qubits and measurements. If you need a refresher, you can check the Qubit kata.
+- Basic knowledge of single-qubit gates. If you need a refresher, you can check the Single-Qubit Gates kata.
```
tcNickolas, line 20: Are we updating the format of how we list prerequisites to include pointers to other refresher katas? I'm reluctant to do this, since this will lead to really long lists of prerequisites, especially in the later katas that assume pretty much all single-qubit and multi-qubit katas as a given.

In this particular case, measurements needed for this kata are covered in "Measurement in Single-Qubit Systems" kata, so even for this very basic kata there are three separate prerequisites.

katas/content/random_numbers/random_bit/solution.md
```diff
@@ -1,6 +1,6 @@
-The state of single qubit can be represented as a two-dimensional column vector $\begin{bmatrix} \alpha \\ \beta \end{bmatrix}$, where $\alpha$ and $\beta$ are complex numbers that satisfy $|\alpha|^2 + |\beta|^2 = 1$. When we measure the qubit, we get either 0 with probability $|\alpha|^2$ or 1 with probability $|\beta|^2$. Essentially we can control probability of measurement outcome by setting the right amplitudes of basis states. 
+The state of single qubit can be represented as a two-dimensional column vector $\begin{bmatrix} \alpha \\ \beta \end{bmatrix}$, where $\alpha$ and $\beta$ are complex numbers that satisfy $|\alpha|^2 + |\beta|^2 = 1$. When you measure the qubit, you get either 0 with probability $|\alpha|^2$ or 1 with probability $|\beta|^2$. Essentially we can control probability of measurement outcome by setting the right amplitudes of basis states.
```
tcNickolas, line 1: There is still a leftover we in the last sentence

katas/content/random_numbers/random_bit/solution.md
```diff
@@ -1,6 +1,6 @@
...
+The state of single qubit can be represented as a two-dimensional column vector $\begin{bmatrix} \alpha \\ \beta \end{bmatrix}$, where $\alpha$ and $\beta$ are complex numbers that satisfy $|\alpha|^2 + |\beta|^2 = 1$. When you measure the qubit, you get either 0 with probability $|\alpha|^2$ or 1 with probability $|\beta|^2$. Essentially we can control probability of measurement outcome by setting the right amplitudes of basis states. 
 
-When we allocate the qubit in Q#, amplitudes $\alpha$ and $\beta$ are 1 and 0, respectively. Now our goal is set equal amplitudes for $\alpha$ and $\beta$ for absolute randomness. We can achieve that by simply applying Hadamard gate to the initial state $\ket{0}$:
+When you allocate the qubit in Q#, amplitudes $\alpha$ and $\beta$ are 1 and 0, respectively. Now your goal is set equal amplitudes for $\alpha$ and $\beta$ for absolute randomness. You can achieve that by simply applying Hadamard gate to the initial state $\ket{0}$:
```
tcNickolas, line 3:
```suggestion
When you allocate the qubit in Q#, amplitudes $\alpha$ and $\beta$ are 1 and 0, respectively. Now your goal is to set equal amplitudes for $\alpha$ and $\beta$ for absolute randomness. You can achieve that by simply applying Hadamard gate to the initial state $\ket{0}$:
```

katas/content/random_numbers/weighted_random_bit/solution.md
```diff
@@ -7,24 +7,24 @@ $$
 
 Here, $\theta$ is the angle between the state vector and the $Z$-axis, and $\phi$ is the longitude angle with respect to the $X$-axis on the Bloch sphere.
 
-Our goal is to generate $0$ or $1$ with the probability of generating a $0$ equal to $x$ and the probability of generating a $1$ equal to $1 - x$. This means that the qubit state should look like
+Your goal is to generate $0$ or $1$ with the probability of generating a $0$ equal to $x$ and the probability of generating a $1$ equal to $1 - x$. This means that the qubit state should look like
```
tcNickolas, line 10:
```suggestion
Your goal is to generate $0$ or $1$ with the probability of generating a $0$ equal to $x$ and the probability of generating a $1$ equal to $1 - x$. This means that the qubit state should look like this:
```

tcNickolas (changes requested):

katas/content/random_numbers/random_bit/solution.md
SoniaLopezBravo:
```suggestion
The state of single qubit can be represented as a two-dimensional column vector $\begin{bmatrix} \alpha \\ \beta \end{bmatrix}$, where $\alpha$ and $\beta$ are complex numbers that satisfy $|\alpha|^2 + |\beta|^2 = 1$. When you measure the qubit, you get either 0 with probability $|\alpha|^2$ or 1 with probability $|\beta|^2$. Essentially, you can control probability of measurement outcome by setting the right amplitudes of basis states. 
```

katas/content/random_numbers/index.md
SoniaLopezBravo: I see, each kata is a consequence of the previous katas. Would you be ok with a general sentence at the end of prerequisites like " if you need a refresher of these topics, check out the previous katas"?

tcNickolas: Yes. This seems intuitive when going through a sequence of tutorials, but one general sentence is better than long lists of pointers to the same katas

tcNickolas (approved):

cesarzc (approved):

DmitryVasilevsky (approved):

PR #1764, opened 2024-07-20
Noisy simulator performance optimization

orpuente-MS:
This PR introduces two main performance optimizations to the noisy simulator:
 - An `get_unchecked` access in the hot loop of the simulator to avoid bound checks.
 - A cache locality optimization: Because `nalgebra` stores its matrices in column major form, it is very inefficient performing Matrix x Vector multiplications, but fast performing Matrix.transpose() x Vector multiplications. So, if we transpose all Kraus operators when they enter the simulator, we can take advantage of this.

There are other minor perf improvements measured with `iai-callgrind` that give us ~3% better performance each. Mostly operations that were being done manually and now are being done with `nalgebra`.

Here is a screenshot of the total improvement in a 10-qubits and 10 operations circuit measured with `iai-callgrind`:
image

### DIFF
--- a/noisy_simulator/src/instrument.rs
+++ b/noisy_simulator/src/instrument.rs
@@ -110,15 +110,19 @@ fn summed_kraus_operators(operations: &[Operation]) -> Result<Vec<SquareMatrix>,
         if *eigenvalue > 0. {
             let mut kraus_operator = SquareMatrix::zeros(krows, kcols);
             let sqrt_eigenvalue = eigenvalue.sqrt();
+
+            // Perf transformation note: for performance reasons we transpose the
+            // Kraus operators before storing them. We want to store each column
+            // of the eigenvector matrix as a Kraus matrix in row major order, and
+            // then transpose it. But because nalgebra stores and iterates over
+            // its matrices in column major order, we can skip the transposition,
+            // since it is already implicit.
+            //
+            // See noisy_simulator/src/operation.rs/Operation::new for more details.
             for row in 0..choi_dim {
                 let idx = kraus_operator.vector_to_matrix_index(row);
                 kraus_operator[idx] = sqrt_eigenvalue * eigenvectors[(row, col)];
             }
-
-            // Perf transformation note: for performance reasons we transpose the
-            // Kraus operators before storing them.
-            // See noisy_simulator/src/operation.rs/Operation::new for more details.
-            // kraus_operator.transpose_mut();
 
             summed_kraus_operators.push(kraus_operator);
         } else if *eigenvalue < -TOLERANCE {
@@ -151,6 +155,8 @@ fn compute_choi_matrix(operations: &[Operation]) -> SquareMatrix {
 /// Perf transformation note: Typically vectorization stacks the
 /// rows of a matrix into a single column vector. But since we
 /// transposed all matrices until now, we stack the columns instead.
+///
+/// See `noisy_simulator/src/operation.rs/Operation::new` for more details.
 fn vectorize<T: nalgebra::Scalar + Copy>(matrix: &DMatrix<T>) -> DVector<T> {
     let mut vectorized_matrix = DVector::<T>::from_vec(Vec::<T>::new());
     for column in matrix.column_iter() {
--- a/noisy_simulator/src/instrument.rs
+++ b/noisy_simulator/src/instrument.rs
@@ -8,6 +8,7 @@
 mod tests;
 
 use crate::{operation::Operation, Error, SquareMatrix, TOLERANCE};
+use nalgebra::{DMatrix, DVector};
 
 /// An instrument is the means by which we make measurements on a quantum system.
 pub struct Instrument {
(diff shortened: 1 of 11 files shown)


noisy_simulator/src/instrument.rs
```diff
@@ -151,19 +135,32 @@ fn summed_kraus_operators(operations: &[Operation]) -> Result<Vec<SquareMatrix>,
...
+/// Perf transformation note: Typically vectorization stacks the
+/// rows of a matrix into a single column vector. But since we
+/// transposed all matrices until now, we stack the columns instead.
+///
+/// See `noisy_simulator/src/operation.rs/Operation::new` for more details.
+fn vectorize<T: nalgebra::Scalar + Copy>(matrix: &DMatrix<T>) -> DVector<T> {
+    let mut vectorized_matrix = DVector::<T>::from_vec(Vec::<T>::new());
+    for column in matrix.column_iter() {
+        vectorized_matrix.extend(column.iter().copied());
+    }
```
idavis, line 164: Instead of extending, can we initialize the `vectorized_matrix` as we should know its final size before the loop

noisy_simulator/src/kernel.rs
```diff
@@ -54,22 +60,44 @@ pub fn apply_kernel(
     let mut new_entries = ComplexVector::zeros(num_elements);
     for s in 0..state.len() {
         if (s & mask) == 0 {
-            new_entries.fill(Complex::ZERO);
-
             // Extract relevant entries into a vector to make the gate application easier.
             for k in 0..num_elements {
```
idavis, line 64: Need to double check, but if you were to use `index_offsets.len()` the compiler should be able to infer get_unchecked without resorting to explicitly calling unsafe code.

noisy_simulator/src/kernel.rs
```diff
@@ -54,22 +60,44 @@ pub fn apply_kernel(
...
-                state[idx] = new_entries[k];
+                // SAFETY: index_offsets and new_entries have size `num_elements`,
+                // so those get_unchecked are safe. state.get_unchecked(idx) is safe because:
+                //  1. s has total_qubits_in_system bits
+                //  2. index_offset has total_qubits_in_system bits
+                //  3. Therefore, idx = s | index_offset also has total_qubits_in_system bits.
+                //  4. idx < (1 << total_qubits_in_system) = state.len() because
+                //     it has (total_qubits_in_system + 1) bits.
+                //  5. Therefore state.get_unchecked(idx) is safe.
+                let idx = s | unsafe { index_offsets.get_unchecked(k) };
```
idavis, line 97: Same as above

noisy_simulator/src/kernel.rs
orpuente-MS: Indeed, I was able to remove the first `get_unchecked(k)` while still skipping the bound check. Just doing `for k in 0..index_offsets.len()` didn't work. I had to do `(0..index_offsets).for_each(|k| { ... })` to be able to remove the bound check there.

I tried several approaches to try to remove the `extracted_entries.get_unchecked_mut(k)` on the second unsafe block, but nothing worked. Here is a list of the things I tried:
 - An assert `assert!(index_offsets.len() == extracted_entries.len())` at the beginning of the loop.
 - An assert at the beginning of the outer loop.
 - `for _ in _` zipped iteration.
 - `for_each` zipped iteration.

noisy_simulator/src/instrument.rs
```diff
@@ -97,49 +98,32 @@ impl Instrument {
...
-                .sum::<SquareMatrix>()
-        })
-        .sum();
-
+    let choi_matrix = compute_choi_matrix(operations);
     let (choi_dim, _) = choi_matrix.shape();
     let eigen_decomposition = choi_matrix.symmetric_eigen();
     let eigenvectors = eigen_decomposition.eigenvectors;
     let eigenvalues = eigen_decomposition.eigenvalues;
     let mut summed_kraus_operators = Vec::new();
```
idavis, line 106: Given that an eigenvalue <= 0 is an error case, would it make sense to preallocate the summed_kraus_operateors vector to the length of the eigenvalues?

noisy_simulator/src/instrument.rs
```diff
@@ -97,49 +98,32 @@ impl Instrument {
...
 
+    let (krows, kcols) = operations[0].kraus_operators()[0].shape();
```
idavis, line 108: Do we have an invariant somewhere that makes both of these index operations safe?

noisy_simulator/src/instrument.rs
orpuente-MS: Yes, you are right.

noisy_simulator/src/instrument.rs
orpuente-MS: Yes, we have two invariants:
 1. When the `kraus_operators: Vec<SquareMatrix>` enters `Operation::new` we check that it contains at least one kraus operator, else we return an `Err`.
 2. When `operations: Vec<Operation>` enters `Instrument::new` we check that it contains at least one `Operation`, else we return an `Err`.

idavis (approved):

noisy_simulator/src/instrument.rs
```diff
@@ -97,49 +98,32 @@ impl Instrument {
...
+    let mut summed_kraus_operators = Vec::with_capacity(eigenvalues.len());
 
+    let (krows, kcols) = operations[0].kraus_operators()[0].shape();
     for (col, eigenvalue) in eigenvalues.iter().enumerate() {
         if *eigenvalue > 0. {
-            let (krows, kcols) = operations[0].kraus_operators()[0].shape();
             let mut kraus_operator = SquareMatrix::zeros(krows, kcols);
             let sqrt_eigenvalue = eigenvalue.sqrt();
+
+            // Perf transformation note: for performance reasons we transpose the
```
DmitryVasilevsky, line 114: Here and in other places. I suggest we drop 'transformation' and just say 'Perf note'. Not sure what you mean by transformation and if it makes things clearer.

orpuente-MS: Thanks for the feedback. I changed "Perf transformation note" to "Performance note", to keep the trend of not shortening words.

billti (approved):

PR #1766, opened 2024-07-22
fix panic in typeck which was causing language service to go into a bad state

### LINKED ISSUE #1762: Random refcell failures in language service
After some investigation, it seems like this could get triggered for a number of conditions (thanks @sezna for digging into this). If a language service worker thread panics or is killed while it holds onto a refcell, later invocations of the worker will fail with borrow errors. This can happen if compilation panics, but also seems to be occurring with some frequency when a project has erroneous references to another project in a local build of the extension.

sezna:
As the title says, this fixes a panic that was causing the language service to go into a bad state. We were assuming that all paths have type schemes, but that's not the case as they can point to namespaces. This should close #1762. Those failures in #1762 were arising from a compilation update worker crashing (due to the aforementioned panic) while holding a lock to the refcell.

### DIFF
--- a/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/compiler/qsc_frontend/src/typeck/rules.rs
@@ -576,7 +576,9 @@ impl<'a> Context<'a> {
             None => match self.names.get(path.id) {
                 None => converge(Ty::Err),
                 Some(Res::Item(item, _)) => {
-                    let scheme = self.globals.get(item).expect("item should have scheme");
+                    let Some(scheme) = self.globals.get(item) else {
+                        return converge(Ty::Err);
+                    };
                     let (ty, args) = self.inferrer.instantiate(scheme, expr.span);
                     self.table.generics.insert(expr.id, args);
                     converge(Ty::Arrow(Box::new(ty)))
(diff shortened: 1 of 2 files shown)


compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -576,7 +576,9 @@ impl<'a> Context<'a> {
             None => match self.names.get(path.id) {
                 None => converge(Ty::Err),
                 Some(Res::Item(item, _)) => {
-                    let scheme = self.globals.get(item).expect("item should have scheme");
+                    let Some(scheme) = self.globals.get(item) else {
```
swernli, line 579: Could this be covered by a unit test, or does that stop working when #1761 merged because the behavior of the repro changes?

swernli (approved):

compiler/qsc_frontend/src/typeck/rules.rs
sezna: I originally had a unit test for it, and then backed it out because the behavior was about to change. But good point, it can still be covered, and you're the one who will have to resolve the conflict 😉

sezna: added in https://github.com/microsoft/qsharp/pull/1766/commits/da9854d2748022260314891cc635fb59ea7ea983

idavis (approved):

PR #1767, opened 2024-07-22
Auto-inserted `import` in notebooks should include project aliases

### LINKED ISSUE #1749: Auto-inserted `open` in notebooks does not include project aliases
The logic for inserting the dependency alias from the qsharp.json on auto-opens works as expected in files but not in notebooks. It still inserts the namespace without the alias which doesn't actually help resolve items from within:
image

swernli:
Simple oversight in not including the dependencies when returning the compilation object for the notebook case.

Fixes #1749


ScottCarda-MS (approved):

sezna (approved):
context: stefan and I discussed this and this fix is exactly what we expected

PR #1769, opened 2024-07-22
Fix accumulating perf impact of `qsharp.run` with may shots in Python

swernli:
The existing imlementation of `qsharp.run` would pass the entry expression for every shot executed, causing it to be recompiling and building up large internal state due to incrementing node ids and the like. This updates the patterns to have the interpreter track the last execution graph it compiled such that a new API in Rust can re-`eval` that last expression without incurring any additional compilation.


compiler/qsc_fir/src/fir.rs
```diff
@@ -894,6 +894,9 @@ impl Display for SpecDecl {
     }
 }
 
+/// An execution graph represented by a reference counted vector of nodes.
+pub type ExecGraph = Rc<[ExecGraphNode]>;
```
billti, line 898: I was worried when I saw all the "cloning" of a graph above until I got to this. Maybe the type alias could reflect that it's a reference count (e.g. `RcExecGraph` or something).

swernli: When you are in the editor with the benefit of rust-analyzer it shows the definition and the comment, does that help?
image

pip/qsharp/_qsharp.py
```diff
@@ -215,10 +218,20 @@ def on_save_events(output: Output) -> None:
...
+    results.append({"result": None, "events": []})
+    run_results = get_interpreter().run(
+        entry_expr, on_save_events if save_events else print_output
+    )
+    results[-1]["result"] = run_results
+    if on_result:
+        on_result(results[-1])
+
+    # Run the remaining shots using `rerun`, which will reuse the compiled entry expression
+    for _ in range(shots - 1):
```
billti, line 231: Rather than duplicate all the logic and vars, why not have it all in the for loop, and just call `run` if iteration 1, else `rerun`?

pip/src/interpreter.rs
```diff
@@ -207,6 +207,17 @@ impl Interpreter {
         }
     }
 
+    fn rerun(&mut self, py: Python, callback: Option<PyObject>) -> PyResult<PyObject> {
```
billti, line 210: It feels a little ugly to keep duplicating code for the == 1 or > 1 case. Might it be possible to just make `entry_expr` on `run` optional, and if you don't pass it we just try to re-run the last graph (i.e. call `run_last_expr` if not provided).

swernli: I debated about that and was on the fence. I don't know why I lean away from having lots of optional parameters to a function, but as you point out this case could really benefit from simplifying. I'll make the switch!

pip/qsharp/_qsharp.py
swernli: Using your suggestion on making the expression optional in `run` should simplify this.

pip/qsharp/_native.pyi
```diff
@@ -84,6 +84,18 @@ class Interpreter:
         """
         ...
 
+    def rerun(self, output_fn: Callable[[Output], None]) -> Any:
```
billti, line 87: Should this be deleted now?

billti (approved):

idavis (approved):

PR #1771, opened 2024-07-23
Add a qsharp.json context menu and palette command to add a project reference

billti:
On invocation, a list of available projects to reference from the local workspace and GitHub (the GitHub list to be populated as they become available), or just a template for the user to complete. 

On selection, the changes are made to the qsharp.json for the user to review and save.

After invocation the drop-down appears similar to the below:

image

### DIFF
--- a/vscode/src/createProject.ts
+++ b/vscode/src/createProject.ts
@@ -289,14 +289,22 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
           });
         });
 
-        // Note: As some point we may want to detect/avoid duplicate names, e.g. if the user already
-        // references a project via 'foo', and they add a reference to a 'foo' on GitHub or in another dir.
         Object.keys(githubProjects).forEach((name) => {
           projectChoices.push({
             name: name,
             ref: githubProjects[name],
           });
         });
+
+        // Convert any spaces, dashes, dots, tildes, or quotes in project names
+        // to underscores. (Leave more 'exotic' non-identifier patterns to the user to fix)
+        //
+        // Note: At some point we may want to detect/avoid duplicate names, e.g. if the user already
+        // references a project via 'foo', and they add a reference to a 'foo' on GitHub or in another dir.
+        projectChoices.forEach(
+          (val, idx, arr) =>
+            (arr[idx].name = val.name.replace(/[- "'.~]/g, "_")),
+        );
 
         const folderIcon = new vscode.ThemeIcon("folder");
         const githubIcon = new vscode.ThemeIcon("github");
--- a/vscode/src/createProject.ts
+++ b/vscode/src/createProject.ts
@@ -187,7 +187,7 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
       github: {
         owner: "swernli",
         repo: "MyQsLib",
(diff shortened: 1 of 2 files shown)


vscode/src/createProject.ts
```diff
@@ -149,4 +165,191 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+
+  type Dependency = LocalProjectRef | GitHubProjectRef;
+
+  // TODO: Replace with a list of legitimate known Q# projects on GitHub
+  const githubProjects: { [name: string]: GitHubProjectRef } = {
+    assert: {
+      github: {
+        owner: "swernli",
+        repo: "MyQsLib",
+        ref: "6ebc707",
```
swernli, line 190:
```suggestion
        ref: "29dc67f",
```

vscode/src/createProject.ts
```diff
@@ -149,4 +165,191 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+          await vscode.workspace.findFiles("**/qsharp.json")
+        ).filter((file) => file.toString() !== qsharpJsonUri.toString());
+
+        const projectChoices: Array<{ name: string; ref: Dependency }> = [];
+
+        projectFiles.forEach((file) => {
+          const dirName = file.path.slice(0, -"/qsharp.json".length);
+          const relPath = getRelativeDirPath(qsharpJsonDir.path, dirName);
+          projectChoices.push({
+            name: dirName.slice(dirName.lastIndexOf("/") + 1),
```
swernli, line 285: Isn't there a chance this could add libraries with directory names with special characters or whitespace and then try to use those as aliases? I haven't actually tried that, and I'm not sure what we do to sanitize the name chosen in the json. It's possible this could cause unintended behavior, so we should test what the failure handling in the dependency path is for spaces.

billti: Yeah, I was thinking we might need to do some 'normalizing'. Even without whitespace or "special characters", it's quite common to have a dash in a directory name (it added `df-chemistry` for example, which is invalid).

We could expect the user to fix up in truly oddball cases, but at least handle the common case of dashes and spaces (maybe just replace with a `_`).

swernli (approved):
I like the functionality, and the folder name clean-up is a reasonable place to start. We just need to remember to pull out the example git library before shipping!

vscode/src/createProject.ts
```diff
@@ -149,4 +165,199 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+      owner: string;
+      repo: string;
+      ref: string;
+      path?: string; // Optional, defaults to the root of the repo
+    };
+  };
+
+  type Dependency = LocalProjectRef | GitHubProjectRef;
+
+  // TODO: Replace with a list of legitimate known Q# projects on GitHub
```
sezna, line 184: Should we remove this before merging it into `main`?

sezna (approved):

vscode/src/createProject.ts
billti: Before release, but not before DEV build. (That was also Stefan's comment I dismissed).

PR #1772, opened 2024-07-23
Gracefully fail in compiler if dependency cycle exists

sezna:
We currently assume/hope the front-end scenario will catch dependency cycles (the JS, Python, or whatever scenario is feeding the dependencies in) and panic otherwise.

Upon retrospection, this was naive, and we should handle dependency cycles gracefully if they're passed through to the back-end. It certainly doesn't hurt. @idavis found a panic which was not detected by the front-end, and therefore caused the backend to panic, and the language service to crash.

This PR updates the compiler to report a `DependencyCycle` error, which ultimately shows up in the _Problems_ tab, if one is detected in the backend.

### DIFF
--- a/compiler/qsc_project/src/project.rs
+++ b/compiler/qsc_project/src/project.rs
@@ -618,7 +618,7 @@ pub type OrderedDependencies = Vec<(Arc<str>, PackageInfo)>;
 
 impl PackageGraphSources {
     /// Produces an ordered vector over the packages in the order they should be compiled
-    pub fn compilation_order(self) -> Result<(OrderedDependencies, PackageInfo), DependencyCycle> {
+    pub fn compilation_order(self) -> (Result<OrderedDependencies, DependencyCycle>, PackageInfo) {
         // The order is defined by which packages depend on which other packages
         // For example, if A depends on B which depends on C, then we compile C, then B, then A
         // If there are cycles, this is an error, and we will report it as such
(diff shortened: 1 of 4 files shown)


compiler/qsc_project/src/project.rs
```diff
@@ -657,16 +657,24 @@ impl PackageGraphSources {
         }
 
         let mut sorted_packages = self.packages.into_iter().collect::<Vec<_>>();
+        let mut cycle_detected = false;
```
idavis, line 660: How does this differentiate between package missing and cycle detected?

sezna: I will admit, this algorithm is a little terse/loosely commented -- I'll update that.

On line 644, this algorithm filters out any dependencies which have greater than 0 dependencies, since we know they won't be able to be the first element in the build order. So, only dependency-less packages are in the `queue`. Every package should be reachable from this `queue`, as they are all inserted as neighbors in the initial construction of `graph`. The only situation in which a package is not present in `sorted_keys` but _is_ present in `sorted_packages` is if it is not reachable via a path which starts with a package that has zero dependencies. In other words, if all possible build orders that reach this package contain a cycle, then there is no possible package to begin the build order with, as every package on the path has at least one dependency it must compile first.

sezna: I just documented this more thoroughly in this commit: d8dc9b1a

idavis (approved):

swernli (approved):
Does this remove the need for the dependency cycle checks in the client layers? If so, it might be good to remove those so there is one place that does cycle checks.

sezna:
@swernli the front-end errors provide more information about the cycle itself, so I think it is good to keep them. This just removes the panic case from the back end in case the cycle makes it through to the compiler.

PR #1780, opened 2024-07-23
remove erroneous export statement for internals of stdlib

### LINKED ISSUE #1774: Intrinsic APIs are marked as internal should not be exported
Operations in `library/qs_source/src/std/internal.qs` that were previously marked as internal visibility to avoid making them part of the public API should not be exported.

_Edited for clarity after discussion, since the expectation is not to export these items_

sezna:
Closes #1774


ScottCarda-MS (approved):

swernli:
It looks like there are a couple other ones from other files to prune too...

library/qs_source/src/std/unstable_table_lookup.qs
```diff
@@ -274,5 +274,5 @@ namespace Microsoft.Quantum.Unstable.TableLookup {
         }
     }
 
-    export Select, WriteMemoryContents, Unlookup, MustBeFixed, EncodeUnary;
+    export Select, WriteMemoryContents, Unlookup;
```
swernli, line 277: I mistyped on this one, both `WriteMemoryContents` and `Unlookup` are meant to be internal, so only `Select` needs to be exported.
```suggestion
    export Select;
```

swernli (approved):
One last tweak needed, otherwise good.

PR #1786, opened 2024-07-24
Noisy simulator: return 2d python lists instead of 1d python lists

orpuente-MS:
This PR changes the output of the Python wrapper to return 2D python lists instead of 1D python lists for density matrices, Kraus operators and other inner matrices of the simulator.


pip/tests/test_noisy_simulator.py
```diff
@@ -13,6 +13,17 @@
...
 # Tests for the Q# noisy simulator.
 
+
+def test_matrices_are_not_transposed_when_returned_back_to_python():
+    """
+    This test is due to the performance optimization we make in
+    `noisy_simulator/src/operation.rs/Operation::new`, we want to check that
+    we are reversing the transpose we made there before returning to Python.
+    """
+    op = Operation([[[1, 2], [3, 4]]])
```
DmitryVasilevsky, line 23: If you ever implement validity checks in Operation constructor, this may stop working. Operators should preserve normalization.

orpuente-MS: Thank you, I used one of the Kraus operators of the depolarizing channel instead.

pip/src/noisy_simulator.rs
```diff
@@ -44,20 +44,22 @@ fn python_to_nalgebra_matrix(matrix: PythonMatrix) -> PyResult<SquareMatrix> {
     Ok(SquareMatrix::from_row_iterator(nrows, ncols, data))
 }
 
-fn nalgebra_matrix_to_python_list(matrix: &SquareMatrix) -> Vec<Complex<f64>> {
-    let (nrows, ncols) = matrix.shape();
-    let mut list = Vec::with_capacity(nrows * ncols);
-
+fn nalgebra_matrix_to_python_matrix(matrix: &SquareMatrix) -> PythonMatrix {
```
billti, line 47: Does the .pyi file need any updates for the changed signature?

orpuente-MS: It doesn't need any updates. It was giving a 2D matrix as type hint already. I guess nobody catch it before because types in Python don't trigger any kind of error.

billti (approved):

DmitryVasilevsky (approved):

PR #1787, opened 2024-07-24
Avoid errors in read-only GitHub sources

swernli:
This is an incremental improvement for viewing GitHub sources that avoids treating them as open documents since they are read-only buffers. In the future, we can do the work to associate them with the existing compilation taht contains those sources so that things like hover and goto-def work, but for now those code paths are user-code specific and need some reworking to handle navigation in a compile unit for a dependency.

### DIFF
--- a/language_service/src/state.rs
+++ b/language_service/src/state.rs
@@ -139,13 +139,6 @@ impl<'a> CompilationStateUpdater<'a> {
         let doc_uri: Arc<str> = Arc::from(uri);
         let text: Arc<str> = Arc::from(text);
 
-        if doc_uri.starts_with(qsc_project::GITHUB_SCHEME) {
-            // Since this is a read only GitHub document provided by our own compilation,
-            // we don't need to track it in the open documents map. As a later feature,
-            // we could consider tracking these documents to provide a better user experience.
-            return;
-        }
-
         let project = match self.load_manifest(&doc_uri).await {
             Ok(Some(p)) => p,
             Ok(None) => Project::from_single_file(doc_uri.clone(), text.clone()),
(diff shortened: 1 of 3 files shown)


language_service/src/state.rs
```diff
@@ -139,6 +139,13 @@ impl<'a> CompilationStateUpdater<'a> {
         let doc_uri: Arc<str> = Arc::from(uri);
         let text: Arc<str> = Arc::from(text);
 
+        if doc_uri.starts_with(qsc_project::GITHUB_SCHEME) {
+            // Since this is a read only GitHub document provided by our own compilation,
+            // we don't need to track it in the open documents map. As a later feature,
+            // we could consider tracking these documents to provide a better user experience.
+            return;
```
billti, line 146: Currently when I open a document from a referenced project, I can still hover on items to get info, goto def on them (at least for standard library stuff)? Will that stop working if just return here?

swernli: Good question. The stdlib has a different schema so is not affected by this change. This only affects resources that are loaded from a GitHub dependency. Before this change the source falls back to single file and shows errors. With this change it gets syntax highlighted but language features are disabled. @sezna and I looked at this for a while today and have some ideas on how to build it up going forward, but we are limited by the fact that most language features assume the current document is either part of the user project or part of stdlib, and now we have another category that needs to be handled throughout.

I have a prototype of the next step that shows how far we can get quickly, and how many things break (hover starts showing incorrect types or values or not working at all, go-to-def is patchy). We have good ideas to build on, but it's not a small change to update every language service feature.

billti: I wonder if it wouldn't be simpler/possible to just filter out the errors from being returned for these docs. It's def handy to still be able to hover items from std library or other locations in the same doc and get something (which I think still work if the document is added).

Are there still negative implications of adding the doc to the language service but filter out any errors for it?

swernli: I think you might be right. As an incremental step toward a full solution, this is the right spot, but for better user experience in a short term workaround your suggestion is probably the way to go. Let me try that.

swernli: Yup, that ends up being a simpler temporary workaround! It gives the same desired behavior (avoids spurious errors) while still supporting a subset of useful features. Some things are broken-ish, like hover on symbols that we don't recognize because they come from elsewhere in the project, but perhaps that is a reasonable middle ground for now.

billti (approved):

sezna (approved):

Issue #481, opened 2023-07-20
Bounded polymorphism


samarsha:
From microsoft/qsharp-language#149:

> **Is your feature request related to a problem? Please describe.** When writing a polymorphic function in Q#, the type variables are fully general. They are opaque types with no properties that the function can rely on.
> 
> But there are many cases where a function needs a type to have certain properties. For example, a function may need its type variables to be equatable or comparable to store/retrieve them from a list or map. Arithmetic functions may also require that the types are numeric.
> 
> **Describe the solution you'd like** Q# needs some form of bounded polymorphism to be able to write functions that are valid for only certain types. I believe type classes are the best solution for this: they are very flexible and fit with Q#'s functional style.
> 
> **Describe alternatives you've considered** Overloading functions based on type (microsoft/qsharp-language#29) would solve the naming problem (e.g. `PlusI`, `PlusD`, `PlusL` could be overloaded to `Plus`), but this solution doesn't scale. Because the numeric types are not formally grouped, there is no way to write a single polymorphic function that calls the right `Plus` automatically - it similarly needs to be overloaded for each numeric type, creating unnecessary code duplication.

sezna:
It is really quite difficult to write most abstractions without bounded polymorphism. I think this task can be split up into smaller tasks:

1. Allow specifying of bounds for existing built-in classes
2. Allow definition of custom classes
3. Allow implementing custom classes for arbitrary types

Number 1 gets us a lot of functionality, and may be enough for a while. For example, being able to constrain inputs to being `Eq + Show` would unblock things like [a testing framework](github.com/sezna/qtest). Numbers 2 and 3 could be done eventually.

PR #1791, opened 2024-07-25
Support call tracking within Q# code

### LINKED ISSUE #1154: Support AllowAtMostNCallsCA operation to limit the number of operation calls that can be done in the code
Classic QDK supported [AllowAtMostNCallsCA](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.allowatmostncallsca) operation that restricted the number of times a certain operation would be called in the code. 

This functionality was broadly used in the Quantum Katas, for example, to limit the number of oracle calls in oracular algorithms, to prohibit using measurements in tasks focused on reversible computing, to limit the use of certain gates in tasks that asked to implement these gates in a specific way, etc. This is especially important in automatically graded programming assignments in which I really want to check whether the solution violates the requirements imposed by the task.

A similar functionality in the modern QDK.

I don't have an alternative implementation for this functionality. In the classic QDK, there used to be a way to write a [custom simulator that could do that](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Common/CounterSimulator.cs) that predated AllowAtMostNCallsCA, but this is not supported in the modern QDK.

swernli:
This change adds callable tracking APIs to the Diagnostics namespace that allow Q# code to register for tracking of specific callables and getting the count of times that callable was invoked. This is useful in katas and auto-grading scenarios to verify that exercise restrictions are being appropriately followed.

Fixes #1154


library/qs_source/src/std/diagnostics.qs
```diff
@@ -149,5 +149,25 @@ namespace Microsoft.Quantum.Diagnostics {
...
     }
 
-    export DumpMachine, DumpRegister, CheckZero, CheckAllZero, Fact, CheckOperationsAreEqual;
+    @Config(Unrestricted)
+    operation StartCountingOperation<'In, 'Out>(callable : 'In => 'Out) : Unit {
+        body intrinsic;
+    }
+
+    @Config(Unrestricted)
+    operation StopCountingOperation<'In, 'Out>(callable : 'In => 'Out) : Int {
```
swernli, line 158: I went back and forth on whether to call these `StopCountingOperation` or `GetOperationCounts`, but to me the latter does not imply tracking stops while the former can imply the final count is returned. If there is a strong preference for "get" instead, then I think we'll need three APIs: start, get, and stop. I prefer to keep it to two and minimize the API surface, personally.

library/qs_source/src/std/diagnostics.qs
```diff
@@ -149,5 +149,25 @@ namespace Microsoft.Quantum.Diagnostics {
...
+        body intrinsic;
+    }
+
+    @Config(Unrestricted)
+    operation StartCountingFunction<'In, 'Out>(callable : 'In -> 'Out) : Unit {
```
billti, line 163: Do we need distinct APIs for operations and functions, versus just having one Start/Stop pair for any callable?

compiler/qsc_eval/src/lib.rs
```diff
@@ -1448,6 +1462,31 @@ impl State {
...
+            }
+            "StopCountingOperation" | "StopCountingFunction" => {
+                let count = self.call_counts.remove(&callable).unwrap_or(-1);
+                Value::Int(count)
+            }
+            _ => panic!("unknown counting call"),
+        }
+    }
+
+    fn increment_call_count(&mut self, callee_id: StoreItemId, functor: FunctorApp) {
```
billti, line 1485: Did you test if this has any notable overhead - esp. something like resource estimation where call counts could be very high with no actual quantum simulation overhead? I can think of some trivial optimizations if needed, but if no noticable impact then doesn't matter.

library/src/tests/diagnostics.rs
```diff
@@ -40,3 +40,119 @@ fn check_operations_are_equal() {
...
+    test_expression(
+        "{
+            import Microsoft.Quantum.Diagnostics.StopCountingFunction;
+
+            function f1() : Unit {}
+            StopCountingFunction(f1)
+        }",
+        &Value::Int(-1),
+    );
+}
```
billti, line 158: Should there be some tests that cover the built-in intrinsics as well, like `X`, etc.?

library/qs_source/src/std/diagnostics.qs
swernli: Yeah, because they are distinct types in Q#. A generic operation is `'In => 'Out` while a generic function is `'In -> 'Out`. There is no way to express something that is any generic callable (something something type classes grumble grumble). I could have one Start/Stop that takes generic `'T` but then you could pass anything and passing the wrong type would result in a runtime failure, which is not a good experience and is inconsistent with our goals of type safety for the language. See our old friend https://github.com/microsoft/qsharp/issues/481.

library/src/tests/diagnostics.rs
swernli: I don't think it should be needed, from the perspective of the evaluator every callable is just a resolution (item id plus package id). The evaluator infra makes sure all callables are treated equal.

compiler/qsc_eval/src/lib.rs
swernli: Good question. I ran the Deutsch-Jozsa benchmark on main, then compared it to this branch in two ways:

1. No operations counted, so just checking the overhead of the checks when no callables are in the tracking hashmap.
2. With `X`, `H`, `CX`, `CNOT`, `Controlled X` all registered, such that 5 operations are in the tracking hashmap that are checked for every call and sometimes hit.

For both versions, the benchmark showed no change in performance (behavior was within noise, such that 1500 iterations ran and some were faster and some were slower, but all by less than +/- 1%). That matches intuition: there would have to be a very high number of callables registered in the hashmap for checking membership to trigger any measurable performance difference.

billti (approved):

compiler/qsc_eval/src/lib.rs
```diff
@@ -1448,6 +1462,31 @@ impl State {
...
         }
     }
+
+    fn counting_call(&mut self, name: &str, arg: Value) -> Value {
+        let callable = if let Value::Closure(closure) = arg {
+            (closure.id, closure.functor)
+        } else {
+            arg.unwrap_global()
+        };
+        match name {
```
cesarzc, line 1472: If we are just using the callable name to determine whether it's a counting call, it is possible that a user creates a callable with the same name, the callable would not behave as the user expects it, right?

It is arguably a very rare case, but maybe it could be even more rare if we also checked more properties like making sure `CallableImpl` is the `Intrinsic` variant, `CallableKind` is the `Operation` variant and possibly also the `input` and `output` properties. This would make it extremely unlikely that someone would implement operations with the exact name and properties as these ones.

cesarzc: I just realized that this comment would be better placed at the `is_counting_call` function!

cesarzc (approved):

compiler/qsc_eval/src/lib.rs
swernli: This is an interesting question. The way the match is used up on line 979 already only checks the name of the callable if it is intrinsic, so that part is covered. At first I figured that would be sufficient because the compiler enforces that all intrinsic callables are globally unique (see `Qsc.Resolve.DuplicateIntrinsic` error in the resolver). But this does change a bit when it comes to dependencies on external projects, as a transitive dependency's globals are not included in the global table for a compilation such that duplicate intrinsic functions can exist. To put it another way, if A depends only on B, and B depends only on C, A can declare an intrinsic with the same name as one declared in C without it triggering this error because global items in C are not visible in A. This gets further complicated by simulatable intrinsics which are not covered by this check. I'll file a follow up issue to cover closing that gap and checking whether there is a reasonable way to ensure globally unique intrinsics even across transitive dependencies.

The good news is that in practice this won't cause a problem for any of these callables, as they are in the standard library and are currently included in every compilation, so anyone declaring an intrinsic callable with the same will get an error. But if/when we make skipping the stdlib an option we have to consider how that will interact with some of these other intrinsics. Like some of the type conversion intrinsics, I think this might make sense to include in the non-quantum part of the stdlib (core or otherwise) that cannot be disabled. For now, the check is safe as there can only be one callable that is both intrinsic and has these names.

swernli: Filed https://github.com/microsoft/qsharp/issues/1817

library/std/src/diagnostics.qs
```diff
@@ -149,5 +149,73 @@ namespace Microsoft.Quantum.Diagnostics {
         areEqual
     }
 
-    export DumpMachine, DumpRegister, CheckZero, CheckAllZero, Fact, CheckOperationsAreEqual;
+    /// # Summary
+    /// Starts counting the number of times the given operation is called. The counter is reset to zero.
+    ///
+    /// # Description
+    /// This operation allows you to count the number of times a given operation is called. As part of
+    /// starting teh counting, the counter is reset to zero, which may override existing counts for the same operation.
```
tcNickolas, line 157:
```suggestion
    /// starting the counting, the counter is reset to zero, which may override existing counts for the same operation.
```

library/std/src/diagnostics.qs
```diff
@@ -149,5 +149,73 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    /// Counting is based on the specific specialization of the operation invoked, so `X` and `Adjoint X` are counted separately.
+    /// Likewise `Controlled X`, `CNOT`, and `CX` are independent operations that are counted separately.
```
tcNickolas, line 159: What about Adjoint Controlled and Controlled Adjoint, are they the same or different?

library/std/src/diagnostics.qs
```diff
@@ -149,5 +149,73 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    /// # Input
+    /// ## callable
+    /// The operation to be counted.
+    @Config(Unrestricted)
+    operation StartCountingOperation<'In, 'Out>(callable : 'In => 'Out) : Unit {
+        body intrinsic;
+    }
+
+    /// # Summary
+    /// Stops counting the number of times the given operation is called and returns the count. Returns -1
```
tcNickolas, line 170: Same question, if the operation is not counted, why -1 instead of 0?

library/std/src/diagnostics.qs
```diff
@@ -149,5 +149,73 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    ///
+    /// # Description
+    /// This operation allows you to stop counting the number of times a given operation is called and returns the count.
+    /// If the operation was not being counted, it returns -1.
+    ///
+    /// # Input
+    /// ## callable
+    /// The operation whose count will be returned.
+    /// # Output
+    /// The number of times the operation was called since the last call to `StartCountingOperation`.
```
tcNickolas, line 181: Same as in counting qubits, it would be good to have a Q# code snippet in API docs for this

library/src/tests/diagnostics.rs
```diff
@@ -40,3 +40,119 @@ fn check_operations_are_equal() {
...
+    test_expression(
+        "{
+            import Microsoft.Quantum.Diagnostics.StartCountingOperation;
+            import Microsoft.Quantum.Diagnostics.StopCountingOperation;
+
+            operation op1() : Unit is Adj + Ctl {}
+            StartCountingOperation(op1);
+            StartCountingOperation(Adjoint op1);
+            StartCountingOperation(Controlled op1);
+            op1();
```
tcNickolas, line 103: Adjoint Controlled and Controlled Adjoint need to be added here, I think

library/src/tests/diagnostics.rs
```diff
@@ -40,3 +40,119 @@ fn check_operations_are_equal() {
...
+            Adjoint op1(); Adjoint op1();
+            Controlled op1([], ()); Controlled op1([], ()); Controlled op1([], ());
+            (StopCountingOperation(op1), StopCountingOperation(Adjoint op1), StopCountingOperation(Controlled op1))
+        }",
+        &Value::Tuple([Value::Int(1), Value::Int(2), Value::Int(3)].into()),
+    );
+}
+
```
tcNickolas, line 111: In the chat we discussed the subtleties of tracking if we define "let my_x = X" and track my_x, vs if we define "let my_x = q => X(q)" and track my_x. I'll need a test case for this

tcNickolas: And mentioning this in API docs would be helpful, I'm bound to forget it otherwise

library/std/src/diagnostics.qs
swernli: Those are the same as each other but different from the others. So you must count body, adjoint, controlled, and controlled adjoint (which is the same as adjoint controlled) separately. But this comment made me realize that I should not treat `Controlled Controlled` and different than `Controlled`, so I will make that change and add a unit test.

library/std/src/diagnostics.qs
swernli: I wanted to differentiate between "You asked me to count this and it was never called" and "You never asked me to count this" which I thought would be helpful. I'm not wedded to it, though.

tcNickolas: The incorrect "You never asked me to count this" scenario in user's code won't be caught by returning either of these values in most cases - the checks typically check that the value is greater than a threshold (0 or a positive number of allowed calls) to say "you were not supposed to use that many", they almost never check for "wait, you called too few!". I like to think that this scenario will be caught by the user trying this code on an incorrect solution to see it fail, but I can't think of a way to automate it...

At the same time, there is a check that does arithmetic with numbers of calls to something (if you're allowed to use one or the other but not both), and I'm worried that returning a negative value in this case will mess the arithmetic up.

Actually, this feels like getting an element from a map when it's not there - some languages return 0, some throw exceptions. Should we maybe just throw an exception? This will happen when the teacher writes the testing harness, not when the student writes their code, so it's a good way to bring it to their attention without impacting student experience. As long as the student doesn't reset the counter in their code...

swernli: That makes sense. I'll turn it into a runtime error instead (both for this and qubit tracking), that way when someone starts tracking after it has already started or stops when it hasn't started they get immediate failures that help them know they are doing something wrong. It will also simplify the doc comments :)

library/src/tests/diagnostics.rs
swernli: Added a test case. I'll be doing some big updates to the doc comments to cover all the feedback.

library/std/src/diagnostics.qs
swernli: Added some samples and clarifications.

tcNickolas (approved):

PR #1792, opened 2024-07-25
Fix bug for max_t_factories constraint

msoeken:
A computation to determine the number of T factories was numerically imprecise leading to results in which the constraint was not satisfied.

The computation has been fixed. Also the chemistry RE benchmark has been updated for which this problem has been discovered in the first place. Corresponding test cases for the chemistry RE benchmark are updated accordingly.

### DIFF
--- a/resource_estimator/src/system/tests.rs
+++ b/resource_estimator/src/system/tests.rs
@@ -472,21 +472,21 @@ pub fn test_hubbard_e2e_increasing_max_num_qubits() -> Result<()> {
 fn prepare_chemistry_estimation_with_expected_majorana(
 ) -> PhysicalResourceEstimation<Protocol, TFactoryBuilder, LogicalResourceCounts> {
     let ftp = Protocol::floquet_code();
-    let qubit = Rc::new(PhysicalQubit::qubit_maj_ns_e4());
+    let qubit = Rc::new(PhysicalQubit::qubit_maj_ns_e6());
 
     let value = r#"{
         "numQubits": 1318,
-        "tCount": 55321680,
-        "rotationCount": 205612244,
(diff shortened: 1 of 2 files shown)


resource_estimator/src/system/tests.rs
```diff
@@ -472,21 +472,21 @@ pub fn test_hubbard_e2e_increasing_max_num_qubits() -> Result<()> {
 fn prepare_chemistry_estimation_with_expected_majorana(
 ) -> PhysicalResourceEstimation<Protocol, TFactoryBuilder, LogicalResourceCounts> {
     let ftp = Protocol::floquet_code();
-    let qubit = Rc::new(PhysicalQubit::qubit_maj_ns_e4());
+    let qubit = Rc::new(PhysicalQubit::qubit_maj_ns_e6());
```
swernli, line 475: Out of curiosity, why change this to the more precise Majorana qubit? Likewise the error budget below is more tolerant. Is there value in keeping both test cases or is the updated the one the only one worth testing?

swernli (approved):

resource_estimator/src/system/tests.rs
msoeken: I wanted to add a scenario in which the error with the factory calculation is observed. The other related test cases that also use a chemistry benchmark just need _some_ benchmark, so it might as well be this one. I don't think there is value in having both. Therefore, for brevity, I decided to use this one for the other test cases as well.

swernli:
@ivanbasov Can you take a look too? The Rust code seems fine to me, but given how much the test case and the numbers validated changed I'm not sure how to verify the results are reasonable.

ivanbasov (approved):

PR #1793, opened 2024-07-25
Sample hardware gateset project

cesarzc:
This change creates a small sample project that provides operations corresponding to a native gateset for a hypothetical hardware device.


samples/projects/quantum_backend/qsharp.json
```diff
@@ -0,0 +1,4 @@
+{
+    "files" : [ "src/GateSet.qs" ],
+    "packageType": "lib"
```
swernli, line 3: nitpick: the `packageType` is no longer needed.

samples/projects/application_for_backend/qsharp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "packageType": "exe",
```
billti, line 2: We got rid of packageType

samples/projects/application_for_backend/src/Main.qs
```diff
@@ -0,0 +1,16 @@
+import Std;
```
billti, line 1: This should be either `open Std;` or `import Std.*;` (or import the individual items you want to use)

samples/projects/application_for_backend/src/Main.qs
```diff
@@ -0,0 +1,16 @@
...
+import ContosoBackend;
```
billti, line 2: As above

samples/projects/application_for_backend/src/Main.qs
sezna: Just to add further clarification in case it isn't already known -- `import Std;` does nothing, since it brings in the "`Std`" name as "`Std`", but it is already available as that name.

samples/projects/application_for_backend/qsharp.json
```diff
@@ -0,0 +1,8 @@
...
+    "dependencies": {
+        "ContosoBackend": {
+            "path": "../quantum_backend"
```
sezna, line 5: Maybe this should be a git dependency? It'll build then in isolation without cloning the entire "samples" dir.

If you use your own commit as the `ref` parameter, it'll build now. We can update that ref later.

Or, we could leave it as is and assume all consumers will only build this in the context of the entire `samples` dir.

sezna:
Do we want the library projects in `samples`?  I was thinking we'd perhaps want to use `library` for our libraries, where we have `std` currently.

samples/projects/quantum_backend/qsharp.json
cesarzc: Removed.

samples/projects/application_for_backend/qsharp.json
cesarzc: Removed.

samples/projects/application_for_backend/src/Main.qs
cesarzc: Changed to a mix of importing everything from the ContosoBackend and an individual item from Std.

samples/projects/application_for_backend/src/Main.qs
cesarzc: Updated.

samples/projects/quantum_backend/src/GateSet.qs
```diff
@@ -0,0 +1,40 @@
+operation Rx(theta : Double, q : Qubit) : Unit {
```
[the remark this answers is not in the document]
cesarzc: Good question, the idea is that when you are developing a program for a specific quantum backend, you only use the gates provided by the backend's library. We still need to think what other features we can create to make this more evident for the user (e.g. lints).

I added both a couple of READMEs and doc comments in the Q# files explaining the philosophy and some of the details.

samples/projects/application_for_backend/qsharp.json
cesarzc: Good point, I'll make it a git dependency.

cesarzc: Updated.

cesarzc:
> Do we want the library projects in `samples`? I was thinking we'd perhaps want to use `library` for our libraries, where we have `std` currently.
> 
> Edit: #1801 sets us up if we want to go this route.

Since this a sample library and not something we want people to rely on, I think it makes sense to keep it in the samples folder.

samples/projects/quantum_backend_application/qsharp.json
```diff
@@ -0,0 +1,11 @@
+{
+    "dependencies": {
+        "ContosoBackend": {
+            "github": {
+                "owner": "microsoft",
+                "repo": "qsharp",
+                "ref": "4e7863b7e31926f3bcd5c4afeaed721d0fb0a652/samples/projects/quantum_backend_library"
```
sezna, line 7: I know we recommend using hashes, but maybe it makes sense to just force it in with `main` as the ref here, so you don't have to update this every time you touch the backend? I can imagine that getting old...

sezna (approved):

samples/projects/quantum_backend_application/qsharp.json
cesarzc: I reverted this to a local dependency to be compatible with the way we test samples.

samples/projects/quantum_backend/src/GateSet.qs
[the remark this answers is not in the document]
billti: Unresolving this, as I kind of agree with @Manvi-Agrawal . If all the 'custom' gates are just calling the already existing intrinsic of the same name, it's not adding anything new. And if it's not stopping the user for using the other available 'canonical' gates, then it's not taking anything away.... so I'm not really sure of the value of this project of what it's trying to show.

sezna (commented):
Quick question -- what about this project is specifically adding value around hardware gate sets? The functions look like they're just passthroughs for intrinsics, just wondering what the use case is.

sezna (changes requested):
requesting changes pending comments around this project's use case -- see above

cesarzc:
> requesting changes pending comments around this project's use case -- see above

The use case for this kind of library is for someone developing a program for a particular target. The library project exposes the native gate set for a particular target in a way that an application that references this library can be simulated and also submitted to the target. This user would then only use quantum operations from the gate set project instead of operations from that are part of the Q# standard library.

I think there is a feature that might be missing to make this kind of library more useful:
- Have a way to enforce no quantum gates/operations from the standard library are used.
    - One option that we can consider for this is to formally introduce the concept of gate set/instruction set.
    - Ideally, we should allow the use of callables from the standard library that are not and do not depend on the Q# gate set/instruction set.

Having said this, I think it would still be valuable to merge a sample like this one since it can be used as a reference for hardware providers to create their own gate set libraries.

sezna (approved):

cesarzc:
Closing for now until we have an agreement on how to handle this kind of scenario.

PR #1797, opened 2024-07-26
Update samples to reflect latest 1.7 changes; Update katas and stdlib to use structs

### LINKED ISSUE #1709: [Question] Samples CI checks and redundancy
**Generic questions**
- Edited because of sample validation changes: Samples in `samples.mjs` are checked for lint, hence in #1579 "Getting Started" sample had to be updated for PR to go in. Should we have similar sort of gatekeeping for the **samples not in `samples.mjs`** list? Then we can just rid of duplicate check?

- It might be worth publishing language samples to the same/separate website, if possible, for greater discovery. 

- If option 2 isnt feasible, can we move samples not in samples.mjs list to a new repo? I liked classic QDK samples because they were in a separate repo. @tcNickolas has told me previously that monolith helps to simplify the dependencies. I highly doubt that somebody would likely visit this massive codebase to see these curated samples. Currently, they serve little purpose for both for Q# compiler team and people onboarding to Q#. Maybe add new samples repo as a submodule here if folks think it might be useful, although I know git submodules are a bit annoying.

**Specific questions**
- : Is `QuantumHelloWorld.qs` sample needed when we have Getting started(aka minimal) and superposition sample?
Also, this is **not in `samples.mjs`** list. Interestingly, it does pop up on [samples website](https://quantum.microsoft.com/en-us/experience/quantum-coding) which seems to have been updates a long time back.
image

- Is `RandomBit.qs` sample needed when code is very similar to superposition sample? Currently, this is in **samples list** and might be okay to have this duplicate to reiterate that random number can be generated using superposition.

sezna:
This PR:
- Updates katas to use structs instead of newtypes
- Performs some minor katas formatting updates (consistency on spacing; removing trailing whitespace; etc)
- Checks our samples for lint warnings or errors in the CI
- updates our samples to use the latest style from Q# 1.7
  - imports instead of opens
  - implicit entrypoint via `Main`
  - implicit namespaces
- Adds stricter lint configs to projects within this repo

Closes #1709 

Thanks @ScottCarda-MS and @swernli for collaborating on this PR!


library/std/src/unstable_state_preparation.qs
```diff
@@ -150,7 +150,7 @@ namespace Microsoft.Quantum.Unstable.StatePreparation {
 
         // Continue recursion while there are control qubits.
         if (IsRangeEmpty(rngControl)) {
-            let (abs, arg) = newCoefficients[0]!;
```
DmitryVasilevsky, line 153: Is this no longer possible?

katas/content/complex_arithmetic/complex_powers_real/Solution.qs
```diff
@@ -1,14 +1,13 @@
-namespace Kata { 
+namespace Kata {
     open Microsoft.Quantum.Math;
 
     function ComplexExpReal(r : Double, x : Complex) : Complex {
         if AbsD(r) < 1e-9 {
-            return Complex(0.0, 0.0);
```
DmitryVasilevsky, line 6: Is this no longer recommended? I liked it.

katas/content/complex_arithmetic/index.md
```diff
@@ -234,35 +233,33 @@ Another way to think about this is that you're taking a point that is $1$ unit a
...
 
-> In Q#, complex numbers in polar form are represented as user-defined type `ComplexPolar` from the `Microsoft.Quantum.Math` namespace. 
-> 
-> You can convert a complex number $x = r \cdot e^{i\theta}$ into a tuple of two `Double` numbers using unwrap operator and tuple deconstruction: `let (r, theta) = x!;`, 
-> or access its magnitude and phase using their names: `let (r, theta) = (x::Magnitude, x::Argument);`.
-> 
-> You can construct a complex number from its magnitude and phase as follows: `let x = ComplexPolar(r, theta);`.
+> In Q#, complex numbers in polar form are represented as user-defined struct type `ComplexPolar` from the `Microsoft.Quantum.Math` namespace.
+>
+> Given a complex number $x = r \cdot e^{i\theta}$, you can access its magnitude and phase using their names: `let (r, theta) = (x.Magnitude, x.Argument);`.
```
DmitryVasilevsky, line 238: This example should probably change. There's no need to access both fields at the same time now.

katas/content/complex_arithmetic/index.md
```diff
@@ -234,35 +233,33 @@ Another way to think about this is that you're taking a point that is $1$ unit a
...
+>
+> You can construct a complex number from its magnitude and phase as follows: `let x = new ComplexPolar{ Magnitude = r, Argument = theta };`.
```
DmitryVasilevsky, line 240: Is this the preferred way? Other places use ComplexPolar(a, b).

katas/content/complex_arithmetic/complex_powers_real/Solution.qs
sezna: I agree that the new way is more verbose, but that's how structs ended up, since the fields are named. I think we will want to discuss adding shorthand syntax.

The old UDT syntax does still work, so feel free to continue using it.

sezna: We have a newer way to do it, so I was updating to that new syntax. However, yeah, it might be good to keep the katas with the older, more concise syntax until a later date. What do you think?

katas/content/complex_arithmetic/index.md
sezna: I've updated this to use the older syntax for now, since it is more concise.

katas/content/complex_arithmetic/index.md
sezna: Updated. Thanks!

library/std/src/unstable_state_preparation.qs
sezna: This is still possible!

katas/content/complex_arithmetic/complex_conjugate/Solution.qs
```diff
@@ -1,7 +1,7 @@
-namespace Kata {    
-    open Microsoft.Quantum.Math;    
+namespace Kata {
+    open Microsoft.Quantum.Math;
 
     operation ComplexConjugate(x : Complex) : Complex {
```
swernli, line 4: This is showing up in my editor with the lint warning that it should be a function. That's not something that intended to be fixed here?

katas/content/complex_arithmetic/complex_exponents/Placeholder.qs
```diff
@@ -1,8 +1,8 @@
 namespace Kata {
     open Microsoft.Quantum.Math;
-    
+
     function ComplexExponent(x : Complex) : Complex {
         // Implement your solution here...
-        return Complex(0.0, 0.0);
+        return Complex(0., 0.);
```
swernli, line 6: Is changing `0.0` to `0.` really needed? I'm curious why the preference for this style. I tend to lean the other way, personally.

swernli: Oh, never mind. I see other katas already do this, so this change is making it consistent.

samples/language/EntryPoint.qs
```diff
@@ -1,15 +1,12 @@
...
-/// the entry point of a Q# program rather than requiring entry points to have
+// # Sample
+// EntryPoint
+//
+// # Description
+// The `@EntryPoint()` attribute is used to designate a particular operation as
+// the entry point of a Q# program rather than requiring entry points to have
 // a particular name such as `main`, `Main`, or `__main__`.
-namespace MyQuantumApp {
+// The Q# compiler identifies `Main` as the entry point operation because it is marked with the `@EntryPoint()` attribute.
```
swernli, line 8: Missing comment update:

```suggestion
// The Q# compiler identifies `Main` as the entry point operation by its name.
// An entry point can also have any name if it is also marked with the `@EntryPoint()` attribute.
```

samples/language/Namespaces.qs
```diff
@@ -2,19 +2,17 @@
...
 
-    @EntryPoint()
-    operation Main() : Result[] {
-        // `DumpMachine` is in the Microsoft.Quantum.Diagnostics namespace
-        DumpMachine();
-        return [];
-    }
-}
\ No newline at end of file
+operation Main() : Result[] {
```
swernli, line 14: This should be a function. I'm surprised this lint doesn't cause failure in CI...

swernli (approved):

samples_test/src/tests/language.rs
```diff
@@ -336,13 +334,11 @@ pub const VARIABLES_EXPECT: Expect = expect![[r#"
     Immutable Int: 42
     Mutable Int: 43
     Mutable Int after mutation: 42
-    Shadowed Immutable Int: 0
```
swernli, line 339: Why delete the shadowing example from this file?

katas/content/complex_arithmetic/complex_conjugate/Solution.qs
sezna: We were just focusing on structs and namespaces, as lints are disabled for katas. That being said, the placeholder is a function, so there's no reason for the solution to be an operation. I've updated it

samples/language/Namespaces.qs
Manvi-Agrawal: Unfortunately, it seems that only samples in `samples.mjs` are checked for lint. See point 1 in https://github.com/microsoft/qsharp/issues/1709 filed earlier by me.

sezna: Fixed, but also looked into why this is happening. It seems we set the default lint level to Allow for that lint: https://github.com/microsoft/qsharp/blob/main/compiler/qsc_linter/src/lints/hir.rs#L38

I didn't realize this, as I thought I was seeing this lint show up in VS Code. Maybe we should consider changing it to warn?

@Manvi-Agrawal  -- I saw that issue 😄  this PR introduces lint checks for all samples. See the functions [here](https://github.com/microsoft/qsharp/pull/1797/files#diff-16dc338c9c40d978fff7f2459d5ea48f1746c215d2608f635a83ae5ba0442c91R299) and [here](https://github.com/microsoft/qsharp/pull/1797/files#diff-e93cacfe927ccbf5b05cd6b97272ffc28432426179e07631686eed2a2cc1f6fdR153). It was a good call, thanks for filing that issue!

katas/content/complex_arithmetic/complex_conjugate/Solution.qs
Manvi-Agrawal: Unnecessary operation lint is disabled in the katas because sometimes the actual solution is a operation and placeholder is not. Prolly, we can enable lint rule for solutions written in katas to account for some of these. 

However, there are certain anomalities that might happen and  might be hard to automate. Also, not sure if they are worth the time and effort.

- If somebody accidently mentions something as operation in Placeholder which can be function instead(can get info from solution file). Additionally, typically 2-3 tasks are added at a time in katas mostly, so most of these should be easy to catch manually(though I know its good to automate as much as possible)

- With regards to above point, I dont think we can enforce if the learner decides to change function to operation while implementing something.

samples/language/Namespaces.qs
Manvi-Agrawal: Earlier, it was set to `Warn` by me and it was changed to `Allow` in https://github.com/microsoft/qsharp/pull/1627. I think it was done because katas might have Placeholders code as operations which doesnt fullfill operation requirements, but the actual solution is an operation.

sezna: I see -- I missed the fact that we override it to warn everywhere else. I think the status quo is fine here, but I'll manually set the configuration for the `samples_test` crate to match how we have it elsewhere.

sezna: Updating the lint config caught four more needless operations in samples! Updating now.

Manvi-Agrawal: Nice, do you think that the below test that executes samples in `samples.mjs` is relevant now? Given that we are now checking for lint and sample execution for all samples? Prolly getting rid of this test in this/other PR might be a good idea if its not needed.
https://github.com/microsoft/qsharp/blob/d1fb2a164a3ec4db73506beda3a70e097ddaacc9/npm/qsharp/test/basics.js#L350-#L370

swernli: Worth some double checking, but I think you are probably right, the newer tests make this testing infra redundant. Good candidate for follow up PR!

samples/language/CopyAndUpdateOperator.qs
```diff
@@ -1,36 +1,24 @@
...
-        // `new_array` is an array with values `[10, 100, 12, 200]`.
-        // `array` is unchanged.
-        let new_array = array
-            w/ 1 <- 100
-            w/ 3 <- 200;
-        Message($"Updated array: {new_array}");
-
-        // `new_pair` is a Pair with value `Pair(20, 100)`.
-        // `pair` is unchanged.
-        let new_pair = pair w/ second <- 100;
```
tcNickolas, line 33: It took me a while to find a sample that would cover this functionality with structs, should we add a pointer to that here?

samples/language/Unit.qs
```diff
@@ -1,17 +1,20 @@
...
-/// The `Unit` type is the singleton type whose only value is ().
-/// Functions implicitly return `Unit` if no explicit or implicit
-/// return is specified.
-namespace MyQuantumApp {
+// # Sample
+// Unit
+//
+// # Description
+// The `Unit` type is the singleton type whose only value is ().
+// Functions implicitly return `Unit` if no explicit or implicit
```
tcNickolas, line 6: Functions won't return Unit except for in these samples, since a function returning Unit would not be useful (no side effects like operation can have). I'd generalize
```suggestion
// Callables implicitly return `Unit` if no explicit or implicit
```

samples/language/Variables.qs
```diff
@@ -1,44 +1,36 @@
...
+    Message($"Mutable Int: {mutableInt}");
+
+    // Mutable variables can be mutated with the `set` keyword:
+    set mutableInt -= 1;
+    Message($"Mutable Int after mutation: {mutableInt}");
+
+    // This is not mutation, rather, this is declaring a new variable
+    // entirely.	    mutable point = new Point3d { X = 0.0, Y = 0.0, Z = 0.0 };
+    let immutableInt = 43;
+    let immutableInt = 0;	    // The below line copies `point`, moves the X coordinate by +1.0,
```
tcNickolas, line 23: Looks like these lines were added below but also copy-pasted here accidentally?
```suggestion
    // entirely.
    let immutableInt = 43;
    let immutableInt = 0;
```

samples/language/CopyAndUpdateOperator.qs
sezna: Added, good catch!

samples/language/MultiFileProject/qsharp.json
```diff
@@ -1 +1,16 @@
-{}
+{
+    "lints": [
+        {
+            "lint": "deprecatedFunctionConstructor",
```
ScottCarda-MS, line 4: I thought we weren't going through with this change?

ScottCarda-MS: I was actually thinking of removing this lint altogether since we are going to support both for the near future.

ScottCarda-MS (approved):

samples/language/MultiFileProject/qsharp.json
sezna: Good point -- I left it in just so we'd get warnings, but I'll remove that for now, now that we've decided that lint will go away.

PR #1798, opened 2024-07-26
Support themes in kata previews

billti:
Cleaned up the kata svg images and content to reduce css duplication, updated the katas preview page to support themes (based on OS preference - same as QCOM), and explicitly set the playground to light theme so images render correctly.

Example of theme switch when changing OS preference below.

https://github.com/user-attachments/assets/cc1df760-3959-4a33-b409-f3b3e3a21edc


DmitryVasilevsky:
Is it normal that my preview looks like this:
image

billti:
Yes, as the standalone image doesn't have any CSS. The other option is to copy all the CSS into every image, which means to remember to update every image to match if the styles change and also increases file sizes a bit.

DmitryVasilevsky:
Somehow that Apply_hadamard image still renders with a little glitch for me on a monitor with lower DPI. The vertical line is almost invisible.
image

DmitryVasilevsky (approved):

billti:
Updated the way CSS is used to enable easier editing and previewing, and documented this in the readme. (If your OS theme is set to light (or just NOT to dark), the GitHub preview should also look better now).

sezna (approved):

PR #1800, opened 2024-07-29
Support qubit count tracking within Q#

### LINKED ISSUE #1155: Support AllowAtMostNQubits operation to limit the number of qubits allocated in the code
Classic QDK supported [AllowAtMostNQubits](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.allowatmostnqubits) operation that restricted the number of qubits that could be allocated in the code.

This functionality was broadly used in the Quantum Katas, for example, to make sure the phase oracle is implemented without the phase kickback trick, or to limit the number of auxiliary qubits allocated by a reversible computing task. This is especially important in automatically graded programming assignments in which I really want to check whether the solution violates the requirements imposed by the task.

A similar functionality in the modern QDK.

I don't have an alternative implementation for this functionality. In the classic QDK, there used to be a way to write a [custom simulator that could do that](https://github.com/microsoft/QuantumKatas/blob/main/utilities/Common/CounterSimulator.cs) that predated AllowAtMostNQubits, but this is not supported in the modern QDK.

swernli:
This introduces qubit count tracking APIs to the Diagnostics namespace that allow Q# code to get counts for how many unique qubits are used during a section of execution. This is useful in katas and auto-grading scenarios to verify that exercise restrictions are being appropriately followed.

Fixes #1155

### DIFF
--- a/compiler/qsc_eval/src/lib.rs
+++ b/compiler/qsc_eval/src/lib.rs
@@ -117,6 +117,16 @@ pub enum Error {
     #[diagnostic(code("Qsc.Eval.QubitUniqueness"))]
     QubitUniqueness(#[label] PackageSpan),
 
+    #[error("qubits already counted")]
+    #[diagnostic(help("counting for qubits must be stopped before it can be started again"))]
+    #[diagnostic(code("Qsc.Eval.QubitsAlreadyCounted"))]
+    QubitsAlreadyCounted(#[label] PackageSpan),
+
+    #[error("qubits not counted")]
+    #[diagnostic(help("counting for qubits must be started before it can be stopped"))]
+    #[diagnostic(code("Qsc.Eval.QubitsNotCounted"))]
+    QubitsNotCounted(#[label] PackageSpan),
+
     #[error("qubits are not separable")]
     #[diagnostic(help("subset of qubits provided as arguments must not be entangled with any qubits outside of the subset"))]
     #[diagnostic(code("Qsc.Eval.QubitsNotSeparable"))]
@@ -174,6 +184,8 @@ impl Error {
             | Error::InvalidNegativeInt(_, span)
             | Error::OutputFail(span)
             | Error::QubitUniqueness(span)
+            | Error::QubitsAlreadyCounted(span)
+            | Error::QubitsNotCounted(span)
             | Error::QubitsNotSeparable(span)
             | Error::RangeStepZero(span)
             | Error::ReleasedQubitNotZero(_, span)
@@ -463,6 +475,7 @@ pub struct State {
     current_span: Span,
     rng: RefCell<StdRng>,
     call_counts: FxHashMap<CallableCountKey, i64>,
+    qubit_counter: Option<QubitCounter>,
 }
 
 impl State {
@@ -484,6 +497,7 @@ impl State {
             current_span: Span::default(),
             rng,
             call_counts: FxHashMap::default(),
+            qubit_counter: None,
         }
     }
 
@@ -1010,6 +1024,14 @@ impl State {
                         callee_span,
                     ));
                 }
+
+                // If qubit counting is enabled, update the qubit counter when the intrinsic for allocation is called.
+                if let (Some(counter), "__quantum__rt__qubit_allocate", Value::Qubit(q)) =
+                    (&mut self.qubit_counter, name.as_ref(), &val)
+                {
+                    counter.allocated(q.0);
+                }
+
                 self.set_val_register(val);
                 self.leave_frame();
                 Ok(())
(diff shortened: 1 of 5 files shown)


compiler/qsc_eval/src/lib.rs
```diff
@@ -1972,5 +1996,25 @@ fn is_counting_call(name: &str) -> bool {
...
+
+#[derive(Default)]
+struct QubitCounter {
+    seen: FxHashSet<usize>,
+    count: i64,
+}
+
+impl QubitCounter {
+    fn allocated(&mut self, qubit: usize) {
+        if self.seen.insert(qubit) {
```
billti, line 2012: Does this mean qubit reuse still counts as the one qubit (which makes sense).

swernli: Yes, that's correct. If qubit A is allocated, used, released, then allocated again, that still counts as just one. I added some unit tests to make sure of this.

library/std/src/diagnostics.qs
```diff
@@ -217,5 +217,32 @@ namespace Microsoft.Quantum.Diagnostics {
...
-    export DumpMachine, DumpRegister, CheckZero, CheckAllZero, Fact, CheckOperationsAreEqual, StartCountingOperation, StopCountingOperation, StartCountingFunction, StopCountingFunction;
+    /// # Summary
+    /// Starts counting the number of qubits allocated. The counter is reset to zero.
+    ///
+    /// # Description
+    /// This operation allows you to count the number of qubits allocated until `StopCountingQubits` is called.
+    /// As part of starting the counting, the counter is reset to zero, which may override a previous count.
+    /// The counter is incremented only when a new unique qubit is allocated, so reusing the same qubit multiple times
+    /// across separate allocations does not increment the counter.
+    @Config(Unrestricted)
```
billti, line 228: Is this not interesting/useful for other profiles? For example, our BellState.qs sample uses only 2 qubits in 'unrestricted', but uses 8 if run under 'base profile'

swernli: If this was allowed in Base profile you could never call it, because it returns a runtime dynamic integer, which is not allowed in Base profile.

billti (approved):

library/src/tests/diagnostics.rs
```diff
@@ -156,3 +156,117 @@ fn check_stop_counting_function_without_start() {
...
+
+#[test]
+fn check_stop_counting_qubits_without_start_returns_negative_one() {
+    test_expression(
+        "{
+            import Microsoft.Quantum.Diagnostics.StopCountingQubits;
+
+            StopCountingQubits()
+        }",
+        &Value::Int(-1),
```
tcNickolas, line 270: This looks odd... I feel that if we don't count something, the natural count of that is 0, why a negative value?

library/src/tests/diagnostics.rs
```diff
@@ -156,3 +156,117 @@ fn check_stop_counting_function_without_start() {
...
+            }
+            {
+                use q = Qubit();
+            }
+            StopCountingQubits()
+        }",
+        &Value::Int(1),
+    );
+}
+
```
tcNickolas, line 239: It would be helpful to include a test that would call an operation between starting and stopping counting qubits. The most common pattern of using AllowAtMostNQubits I can find is 

                within {
                    AllowAtMostNQubits(2*N, "You are not allowed to allocate extra qubits");
                } apply {
                    AssertOperationsEqualReferenced(N,
                                                    ArbitraryBitPattern_Oracle_Challenge(_, pattern),
                                                    ArbitraryBitPattern_Oracle_Challenge_Reference(_, pattern));
                }

So a test could replicate it with a comparison of some library operations using CheckOperationsAreEqual, and checking that 2N qubits are a result

tcNickolas:
One more thing that needs a test case - in the chat we discussed "because multi-controlled gates are decomposed during simulation the same way they are for hardware, they will allocate additional auxiliary qubits proportional to the number of control qubits, which will show up in the count." - there is no test case with a controlled functor right now

library/src/tests/diagnostics.rs
swernli: Per conversation on the other PR, this is now an error instead.

library/std/src/diagnostics.qs
```diff
@@ -217,5 +217,32 @@ namespace Microsoft.Quantum.Diagnostics {
...
+
+    /// # Summary
+    /// Stops counting the number of qubits allocated and returns the count.
+    ///
+    /// # Description
+    /// This operation allows you to stop counting the number of qubits allocated and returns the count since the
+    /// last call to `StartCountingQubits`.
+    ///
+    /// # Output
+    /// The number of unique qubits allocated since the last call to `StartCountingQubits`.
```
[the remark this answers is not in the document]
swernli: I added a simple sample in the remarks section for `StartCountingQubits`

library/src/tests/diagnostics.rs
swernli: I added `check_counting_qubits_works_with_allocation_in_operation_calls` at the end of the file that shows how this pattern could work now.

tcNickolas (approved):

PR #1804, opened 2024-07-29
Migrate Graph Coloring kata, part 2

tcNickolas:
This finishes the migration and publishes the kata

### DIFF
--- a/katas/content/solving_graph_coloring/weak_coloring_classical/index.md
+++ b/katas/content/solving_graph_coloring/weak_coloring_classical/index.md
@@ -5,7 +5,7 @@
 Each tuple gives the indices of the start and the end vertices of the edge.
 The vertices are numbered $0$ through $V - 1$.
 3. An array of $V$ integers, representing the vertex coloring of the graph. 
-$i$-th element of the array is the color of the vertex number $i$.
+The $i$-th element of the array is the color of the vertex number $i$.
 
 For example, consider the triangular graph `0 -- 1 -- 2 -- 0` with $V = 3$ and `edges = [(0, 1), (1, 2), (0, 2)]`. 
 The coloring `[0, 1, 0]` is a valid weak coloring for it despite the fact that the connected vertices $0$ and $2$ are assigned the same color, since each of them is also connected to the vertex $1$ of a different color.
--- a/katas/content/solving_graph_coloring/weak_coloring_classical/index.md
+++ b/katas/content/solving_graph_coloring/weak_coloring_classical/index.md
@@ -1,0 +1,14 @@
+**Inputs:**
+
+1. The number of vertices in the graph $V$ ($V \leq 6$).
+2. An array of $E$ tuples of integers, representing the edges of the graph ($E \leq 12$).
+Each tuple gives the indices of the start and the end vertices of the edge.
+The vertices are numbered $0$ through $V - 1$.
+3. An array of $V$ integers, representing the vertex coloring of the graph. 
+$i$-th element of the array is the color of the vertex number $i$.
+
+For example, consider the triangular graph `0 -- 1 -- 2 -- 0` with $V = 3$ and `edges = [(0, 1), (1, 2), (0, 2)]`. 
+The coloring `[0, 1, 0]` is a valid weak coloring for it despite the fact that the connected vertices $0$ and $2$ are assigned the same color, since each of them is also connected to the vertex $1$ of a different color.
+
+**Output:**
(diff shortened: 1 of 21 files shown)


cesarzc (approved):
Looks good!

katas/content/solving_graph_coloring/weak_coloring_classical/index.md
```diff
@@ -0,0 +1,14 @@
+**Inputs:**
+
+1. The number of vertices in the graph $V$ ($V \leq 6$).
+2. An array of $E$ tuples of integers, representing the edges of the graph ($E \leq 12$).
+Each tuple gives the indices of the start and the end vertices of the edge.
+The vertices are numbered $0$ through $V - 1$.
+3. An array of $V$ integers, representing the vertex coloring of the graph. 
+$i$-th element of the array is the color of the vertex number $i$.
```
ScottCarda-MS, line 8: nit
```suggestion
The $i$-th element of the array is the color of the vertex number $i$.
```

katas/content/solving_graph_coloring/weak_coloring_one_vertex/solution.md
```diff
@@ -0,0 +1,12 @@
+To implement this check, you need to start by counting the vertices connected to the given vertex 
+and allocating an array of auxiliary qubits, one for each neighboring vertex. These qubits will have their states flipped if the vertices connected by the corresponding edge have the same color. 
+This can be done easily using with the operation `Oracle_ColorEquality` from an earlier task, similarly to how you did in when validating the vertex coloring of the graph. 
+
+Now, for the coloring of the vertex to be valid, it needs to either have no neighbors, or to have at least one neighbor of a different color, so the array of auxiliary qubits should be in any state but $\ket{1...1}$. To implement this, you can flip the target qubit $\ket{y}$ unconditionally using the $X$ gate, and then flip it again only if qubits are in state $\ket{1...1}$.
```
ScottCarda-MS, line 5: The references to state $\ket{1...1}$ are confusing to me here because the solution uses `ApplyControlledOnInt(0, X, x1, y);` which I take to mean that it will flip it again only when the qubits are in state $\ket{0...0}$. Am I misunderstanding this?

ScottCarda-MS (commented):
There's some confusion on my part about the references to the $\ket{1...1}$ state.

katas/content/solving_graph_coloring/weak_coloring_one_vertex/solution.md
tcNickolas: The part of the solution this paragraph talks about is in `Oracle_WeakColoring_OneVertex`:
```
            X(y);
            if nNeighbors > 0 {
                Controlled X(sameColorChecks, y);
            }
```
`ApplyControlledOnInt(0)` shows up in `Oracle_ColorEquality` that is subroutine from an earlier exercise, so I don't take about it here

ScottCarda-MS: Yes, as a subroutine of the solution, it seems like the solution is only flipping the output when the input qubits are in state $\ket{0...0}$.

tcNickolas: Sorry, I'm not sure I follow...
When we compare two colors, we use a different logic indeed: since we're computing XORs of pairs of bits to check that they're all 0, we get the results by checking that all qubits are 0. But this was discussed earlier, in exercise "Are Colors Equal?", and I'm assuming that the reader is already familiar with this

ScottCarda-MS: Oh if flips the solution's output when all of the individual color checks return $\ket{1}$. Is that right?

tcNickolas: Yes, with a twist: if all of the individual color checks return $\ket{1}$, that's the only scenario when weak coloring is *in*valid (meaning all neighbors' colors are equal; at least one of color pairs has to be different and return $\ket{0}$ for it to be valid). 

So the complete solution flips the output twice: once unconditionally, and once if all of the individual color checks return $\ket{1}$. This means that all the individual color checks return $\ket{1}$ is the only scenario in which the result remains unflipped, indicating invalid coloring

ScottCarda-MS: That makes sense. Thanks!

ScottCarda-MS (approved):

PR #1805, opened 2024-07-30
Update rust version to 1.80

orpuente-MS:
This PR updates the versions in our build and release pipelines to use Rust 1.80.

When fixing the new clippy lints to satisfy 1.80, we get some errors in the build pipeline.

All of the errors are related to the fact that rust 1.80 implements Iterator for `&Box<[...]>`. So, what before had to be written as `for elt in container.iter()` now can be written as `for elt in container` which is more concise, and clippy enforces this new style.
 
`for elt in &**container` happens to be equivalent and it works both in rust 1.78 and rust 1.80, and clippy doesn't complain about it in either.
 
This PR changes all the instances of `for elt in container.iter()` where container is a `&Box<[...]>` to `for elt in &**container` so we can update the rust version to 1.80. There will be a subsequent PR that changes all the instances of `for elt in &**container` to `for elt in container`.


ScottCarda-MS:
Based on the description of the PR, why not just updated all instances of `for elt in container.iter()` where container is a `&Box<[...]>` to `for elt in container` instead of `for elt in &**container`?

orpuente-MS:
> Based on the description of the PR, why not just updated all instances of `for elt in container.iter()` where container is a `&Box<[...]>` to `for elt in container` instead of `for elt in &**container`?

Because our current pipeline is built using rust 1.78 and `impl<T> Iterator for Box<[T]>` was introduced in rust 1.80. So `for elt in container` issues a compile error in the current build pipeline.

idavis (approved):

PR #1810, opened 2024-07-31
Improve ApplyQFT API docs

tcNickolas:


library/qs_source/src/std/canon.qs
```diff
@@ -489,17 +489,22 @@ namespace Microsoft.Quantum.Canon {
...
     /// least significant bit xₙ. The state of qs[0] becomes
     /// (|0⟩+𝑒^(2π𝑖[0.xₙ])|1⟩)/sqrt(2) after the operation.
     ///
     /// # Input
     /// ## qs
-    /// Quantum register in a little-endian format to which the QFT is applied.
+    /// Quantum register in a little-endian format to which the rotations are applied.
+    ///
+    /// # Remarks
+    /// Note that this operation applies only the rotations part of the QFT.
```
DmitryVasilevsky, line 505: We agreed that Remarks section would contain general information (in contrast to information that is only relevant to this specific operation you are commenting on.) So maybe we can include a link to a common knowledge that defines Quantum QFT? A Wikipedia article maybe?

DmitryVasilevsky (approved):

library/qs_source/src/std/canon.qs
tcNickolas: There is a link to wiki article on QFT in `# Reference` right below, so I don't think we need to duplicate it. I wanted to call out how to implement a complete QFT using this operation; is there a better section do to that?

DmitryVasilevsky: Yes, it's fine. But that Wikipedia article doesn't have the swaps in the end. Is there a good reference that has those swaps?

tcNickolas: I think this all goes back to N&C which had a circuit without swaps (and said swaps should follow in the text). Wikipedia mentions the need for swaps somewhere in the text. Can't think of a good source with swaps, sorry; a lot of sources copy the same circuit over and over from Wikipedia

sezna (approved):

PR #1811, opened 2024-08-01
Added a sample function with an implicit return

filipw:
I did not see this in the samples, and to me it makes sense to have it together with other "return" samples. 
It's quite a significant feature (especially to programmers used to Python 😄)

### DIFF
--- a/samples/language/ReturnStatement.qs
+++ b/samples/language/ReturnStatement.qs
@@ -21,6 +21,13 @@ namespace MyQuantumApp {
         // Any code following a return statement is inaccessible.
     }
 
+    function Returns42Implicit() : Int {
+        // If the last expression is not followed by a semicolon,
+        // its value is implicitly returned, in which case the return keyword is not necessary.
+        42 // This literal value is implicitly returned.
+        // Any code following an implicit return would cause the compilation to fail.
+    }
+
     function ReturnsUnit() : Unit {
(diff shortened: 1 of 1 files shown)


sezna (approved):
Thank you, looks good! And solid choice on numeric value as well.

filipw:
> And solid choice on numeric value as well.

I wish I could take credit for it - the example above already used it 😄

samples/language/ReturnStatement.qs
```diff
@@ -21,6 +21,13 @@ namespace MyQuantumApp {
...
     }
 
+    function Returns42Implicit() : Int {
+        // If the last expression is not followed by a semicolon,
+        // its value is implicitly returned, in which case the return keyword is not necessary.
+        42 // This literal value is implicitly returned.
+        // Any code following an implicit return would cause the compilation to fail.
+    }
+
     function ReturnsUnit() : Unit {
```
Manvi-Agrawal, line 31: Probably, its worth having an example explicitly returning Unit here for completeness? And remove `samples/language/Unit.qs` example? I remember coming across that sample and thought it was mentioned here but it isnt. Thoughts?

```qsharp
function ReturnsUnitExplicit() : Unit {
        return ();
    }
 ```

swernli (approved):

PR #1814, opened 2024-08-02
Finish migration of QFT kata and publish it

tcNickolas:
I dropped part 3 of the original kata deliberately; the two tasks it includes are from a competition and are not used anywhere I'm aware of.


katas/content/qft/alternating_amplitudes/solution.md
```diff
@@ -0,0 +1,10 @@
+Since the amplitudes of this state have no imaginary parts, it is clear that the value of exponents for an arbitrary amplitude $e^{2\pi i \cdot \frac{Fk}{2^n}}$ must be either $2\pi i$ to get an amplitude equal $1$ or $\pi i$ for an amplitude of $-1$. 
+This means that only $j_1$ can be $1$: all other bits of $F$ must be $0$. 
+Indeed, you can check that using $F = 10...0 = 2^{n - 1}$ with the solution of the previous task yields the required state.
+
+You can simplify preparing this state: instead of using the general encoding of $F$ in the register, you can apply an $X$ gate to the first qubit of the register.
```
cesarzc, line 5:
```suggestion
You can simplify preparing this state, instead of using the general encoding of $F$ in the register, you can apply an $X$ gate to the first qubit of the register.
```

cesarzc (approved):

katas/content/qft/alternating_amplitudes/solution.md
tcNickolas: This doesn't look right: the second part of the sentence is an explanation of the first, so : is appropriate

ScottCarda-MS (approved):

PR #1820, opened 2024-08-07
Fix duplicate imports being generated when using completions

### LINKED ISSUE #1775: Auto-Imports will Trigger when Unnecessary
**Describe the bug**

When all the items of a namespace are imported with the `*` blob, auto-completes of these items erroneously create unnecessary insert import statements for that specific item. This seems to be the behavior when importing from a remote package.

**To Reproduce**

1. Import a namespace from a remote package using the `*` blob.
2. Begin typing out an item from the namespace.
3. Auto-complete an item from the namespace.

You will see that an auto-import will be inserted into the file for the specific item. This is not necessary as the item is already available via the blob import.

**Edit**

This behavior also seems to repo when importing from the standard library, such as the Math namespace's PI function.

### LINKED ISSUE #1777: Auto-Imports Cause Duplicate Import Statements
**Describe the bug**

When an item is imported from a namespace, two copies of the item will appear in the completions list. If the user completes with the second option, a duplicate import statement will be automatically inserted. This seems to be the behavior when importing from a remote package.

**To Reproduce**

1. Import an item from a namespace from a remote package.
2. Begin typing out an item from the namespace.
3. Note that there are two instances of the item in the completion list.
4. Auto-complete the item using the second option in the completion list.

You will see that an additional auto-import will be inserted into the file for the specific item. There should only be one option for this item in the completion list, and auto-imports should not create an erroneous duplicate import.

**Edit**

This behavior also seems to repo when importing from the standard library, such as the Math namespace's PI function.

sezna:
Closes #1777
Closes #1775 

This PR updates completions in the following ways:
- Organizes `completions.rs` a bit -- it had some very, very nested code that was hard to make sense of.
- Doesn't insert an import if a symbol has already been imported via a glob import or open
- Doesn't insert an import if a symbol has already been imported by name exactly
- Recognizes if the parent namespace has already been imported via an alias
- Removes the `ignore_internal_callable` test, which was testing `internal` as if it would make things private to a single namespace, when actually `internal` means private to a package.

Recognizing a preexisting glob import:
 

https://github.com/user-attachments/assets/2732da7a-8211-4bb3-927e-1af234b81fb8

Recognizing a preexisting alias:

https://github.com/user-attachments/assets/ca5e99f4-294a-4eed-8b91-d9133a08d124


language_service/src/completion/tests.rs
```diff
@@ -174,84 +174,6 @@ fn ignore_unstable_callable() {
     );
 }
 
-#[test]
-fn ignore_internal_callable() {
```
ScottCarda-MS, line 178: Do we no longer ignore internal callables? That seems like an intended feature, no?

ScottCarda-MS: Ah I see this is explained in your description, my bad. Is there a similar test somewhere that ensures that internal items in _other packages_ are not discoverable during completions?

ScottCarda-MS (approved):

language_service/src/completion/tests.rs
sezna: Yeah, that's verified via HIR tests [here](https://github.com/microsoft/qsharp/blob/7ae7a6f1b15f33aaf1c1fe6e81beaeb2fd55372c/compiler/qsc_frontend/src/compile/tests/multiple_packages.rs#L184). We don't have a test for that specifically in completions, though.

language_service/src/completion.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(warnings)]
```
DmitryVasilevsky, line 4: Maybe a comment why this is needed?

sezna: This was actually a mistake, I meant to remove it. Whoops! Will fix.

sezna: Fixed in cd458b68

language_service/src/completion/tests.rs
```diff
@@ -174,84 +174,6 @@ fn ignore_unstable_callable() {
     );
 }
```
billti, line 175: Why is the only test change _removing_ tests. Isn't there any change in behavior to verify here?

sezna: Added tests in b0f7de3

language_service/src/completion.rs
```diff
@@ -749,3 +713,157 @@ impl Visitor<'_> for ContextFinder {
...
+        if let Some(parent) = package.items.get(item_id) {
+            if let ItemKind::Namespace(callable_namespace, _) = &parent.kind {
+                // filter out internal packages that are not from the user's
+                // compilation
+                if matches!(i.visibility, Visibility::Internal) && !is_user_package {
+                    return None; // ignore item if not in the user's package
+                }
+
+                match &i.kind {
+                    ItemKind::Callable(callable_decl) => {
```
orpuente-MS, line 740: Could we change `i` to `item` here? It's just three more letters and it is more readable.

```suggestion
fn package_item_to_completion_item(
    item: &qsc::hir::Item,
    package: &qsc::hir::Package,
    is_user_package: bool,
    current_namespace_name: Option<&[Rc<str>]>,
    display: &CodeDisplay,
    package_alias_from_manifest: &Option<Arc<str>>,
    imports: &[ImportItem],
    insert_open_at: Option<Range>,
    indent: &String,
) -> Option<(CompletionItem, SortPriority)> {
    // We only want items whose parents are namespaces
    if let Some(item_id) = item.parent {
        if let Some(parent) = package.items.get(item_id) {
            if let ItemKind::Namespace(callable_namespace, _) = &parent.kind {
                // filter out internal packages that are not from the user's
                // compilation
                if matches!(item.visibility, Visibility::Internal) && !is_user_package {
                    return None; // ignore item if not in the user's package
                }

                match &item.kind {
                    ItemKind::Callable(callable_decl) => {
```

orpuente-MS (approved):
Approved with minor comment.

Issue #591, opened 2023-08-16
Kata examples should be editable and runanble


DmitryVasilevsky:
Currently katas in the playground have no way to edit or run examples (in the main index file or in the answer/solution file).

**To Reproduce**

Open any Kata with example "multi-qubit systems" and scroll to any example (for example, Lesson: Multi-qubit systems in Q#). The code will be shown read-only without any buttons to run.

**Expected behavior**

Kata code examples should be editable and runnable just like the default sample in the playground.

sezna:
@DmitryVasilevsky is this done now? It seems katas are editable now on microsoft.github.io/qsharp

Issue #1577, opened 2024-05-28
Phantom "entry point not found" error in the Playground


orpuente-MS:
**Describe the bug**

I found some bad behavior where the `LanguageService` issues an "entry point not found" error even though the `@EntryPoint` is present.

**To Reproduce**

1. Open the Playground.
2. Click on "Getting Started" (the first tutorial).
3. Click back on "Minimal" (the first sample).
4. You will see the diagnostic in the editor.

**Expected behavior**
After switching back to "Minimal" there shouldn't be any errors.

**Screenshots**

![image](https://github.com/microsoft/qsharp/assets/156957451/6a530dc5-c8b2-4f32-9c33-ae41a7014740)

**System information**

- Version of the package or extension for which the problem occurs: latest main branch
- Your operating system: Windows 11
- For issues with extensions, the version of the IDE for which the problem occurs: Playground

**Additional context**

swernli:
The spot fix that I thought resolved this actually broke other language service features in the playground, so we need to revert it.

swernli:
This is fixed as of recent updates to how we track editor instances with Monaco in the playground.

PR #1821, opened 2024-08-07
Add 'Get a Link' to kata examples

ggridin:
An alternative approach to address "Kata examples should be editable and runnable" https://github.com/microsoft/qsharp/issues/591

Kata example window has "Get a link to this code" icon and is scrollable (saves physical space in kata lesson for long examples).

### DIFF
--- a/playground/src/kata.tsx
+++ b/playground/src/kata.tsx
@@ -6,6 +6,7 @@ import { QscEventTarget } from "qsharp-lang";
 import { Editor, getProfile } from "./editor.js";
 import { OutputTabs } from "./tabs.js";
 import { Markdown } from "qsharp-lang/ux";
+import { Viewer } from "./viewer.js";
 
 import type {
   CompilerState,
@@ -93,11 +94,7 @@ function LessonElem(props: Props & { section: KataSection }) {
         {lesson.items.map((item) => {
           switch (item.type) {
             case "example":
-              return (
-                <pre>
-                  <code>{item.code}</code>
-                </pre>
-              );
+              return <Viewer code={item.code}></Viewer>;
             case "text-content":
               return <Markdown markdown={item.content}></Markdown>;
             case "question":
--- a/playground/src/main.css
+++ b/playground/src/main.css
@@ -7,3 +7,13 @@ pre:has(code) {
   margin: 5px 0px 5px 0px;
   padding: 0em 0em 0em 0.5em;
 }
+.viewer-column {
+  grid-area: editor;
+  margin: 16px;
+  margin-top: 32px;
+}
+.example-viewer {
+  height: 40vh;
+  min-height: 400px;
+  border: 1px solid var(--border-color);
+}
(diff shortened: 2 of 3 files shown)


ggridin:
Screenshot of the window:
Kata-examples-Get-a-Link

Kata editor in function LessonElem is kata.tsx is causing stable repro of "Entry point not found" error https://github.com/microsoft/qsharp/issues/1577 in "Deutsch Algo..." & "Deutsch-Josza..." katas

kata.tsx file with kata examples editor is attached (rename it to kata.tsx if you'd like to test)

[kata-example-editor.txt](https://github.com/user-attachments/files/16518530/kata-example-editor.txt)

tcNickolas:
I'm not sure I follow the proposed change, sorry.

* The change as proposed by the PR makes the demos scrollable in the playground, but not editable or runnable. Do I understand correctly that the one benefit it offers is space saving?
* Using the kata-example-editor.txt instead of kata.tsx makes the demos editable but not runnable - they have controls to select profile and number of shots, but the Run and Cancel buttons are greyed out and not clickable. Plus, all the kata editors start throwing the missing entry point error, so that doesn't look like a good idea.

ggridin:
@tcNickolas 
kata-example-editor.txt -> kata.tsx has greyed Run & Cancel button due to #1577 
- If you check kata exercises on the same page - they report "Entry point not found" error. Changing profile to "base" and back to "unrestricted" allows editing and running of the example.
- Proposed/then reverted fix in #1577 ("code.ts") (seems to) fixes the problem. However, I do not know why the fix was reverted.
The root problem might be in language-service layer.

To process forward without language-service fix (I do not know how to do that yet) proposed "Get a link" workaround might be used. The benefits are:
1. Single click option to open example in edit and run environment (vs current copy-paste flow)
2. Space saving

ggridin:
@orpuente-MS @swernli could you comment on why the "Phantom error" fix was reverted?

swernli:
> @orpuente-MS @swernli could you comment on why the "Phantom error" fix was reverted?

The fix from PR https://github.com/microsoft/qsharp/pull/1628 was reverted because while it seemed to make the phantom error go away it actually broke all other language features (hover, rename, go-to-def, and other error feedback stopped working). The real fix ended up coming in PR https://github.com/microsoft/qsharp/pull/1738, which fixed the phantom entry point error as a side effect of repairing the way file names are inferred from the Monaco editor instances so that implicit namespaces worked. I'll update the original issue to link the real fix.

Of note in that fix, if the properties submitted to the editor had a kata exercise id, that was used for the editor name, otherwise it fell back to the file name "main.qs" under the assumption that everything without an exercise id is a single editor page with one example. This change breaks that assumption, because now there can again be multiple editors on a page with the same file name, interfering with language service document tracking logic.

Given that all this is for our developer playground and not for customer facing katas, I don't think having to copy-paste the examples is that big of a burden. I personally would be in favor of leaving the code as-is and closing the issue as "Not Planned." That said, I don't feel strongly about this and I'm fine if you or others feel motivated to find a way to make this work.

PR #1824, opened 2024-08-07
Start migration of Phase Estimation kata

tcNickolas:
* First two exercises match the last two exercises from classic Linear Algebra kata, and follow a similar approach for migrating it to Q#, replacing Python code with a hardcoded exercise
* Third exercise is 1.3 from classic PhaseEstimation kata, rewritten to work with Boolean checks rather than assertions.

Part 2 will include exercises 2.1 and 1.4 from classic PhaseEstimation kata and new (to katas) QPE algorithm theory and demo of its success probability

### DIFF
--- a/katas/content/phase_estimation/eigenvectors_x/Verification.qs
+++ b/katas/content/phase_estimation/eigenvectors_x/Verification.qs
@@ -1,0 +1,32 @@
+namespace Kata.Verification {
+    open Microsoft.Quantum.Math;
+
+    @EntryPoint()
+    operation CheckSolution() : Bool {
+        let actual = Kata.EigenvectorsX();
+        if Length(actual) != 2 {
+            Message("The array of eigenvectors should have exactly two elements.");
+            return false;
+        }
+        for i in 0 .. 1 {
+            if Length(actual[i]) != 2 {
+                Message("Each eigenvector should have exactly two elements.");
+                return false;
+            }
+            if AbsD(actual[i][0]) + AbsD(actual[i][1]) < 1E-9 {
+                Message("Each eigenvector should be non-zero.");
+                return false;
+            }
+        }
+
+        // One eigenvector has to have equal components, the other one - opposite ones
+        if AbsD(actual[0][0] - actual[0][1]) < 1e-9 and AbsD(actual[1][0] + actual[1][1]) < 1e-9 or 
+            AbsD(actual[0][0] + actual[0][1]) < 1e-9 and AbsD(actual[1][0] - actual[1][1]) < 1e-9 {
+                Message("Correct!");
+                return true;
+            }
+
+        Message("Incorrect value for one of the eigenvectors.");
+        return false;
+    }
+}
--- a/katas/content/phase_estimation/index.md
+++ b/katas/content/phase_estimation/index.md
@@ -1,0 +1,108 @@
+# Phase Estimation
+
+@[section]({
+    "id": "phase_estimation__overview",
+    "title": "Overview"
+})
+
+This kata introduces you to the phase estimation algorithm - an important building block in more advanced quantum algorithms such as integer factoring.
+
+**This kata covers the following topics:**
+
+- The definition of eigenvalues and eigenvectors
+- The phase estimation problem
+- The quantum phase estimation algorithm based on quantum Fourier transform
+
+**What you should know to start working on this kata:**
+
+- Basic quantum gates and measurements.
+- Quantum Fourier transform.
+
+@[section]({
+    "id": "phase_estimation__eigen",
+    "title": "Eigenvectors, Eigenvalues, and Eigenphases"
+})
+
+An *eigenvector* of a matrix $A$ is a non-zero vector that, when multiplied by that matrix, changes by a scalar factor:
+
+$$A \ket{v} = \lambda \ket{v}$$
+
+The number $\lambda$ is called an *eigenvalue* that corresponds to this eigenvector. In general, eigenvalues of matrices can be complex numbers.
+
+Recall that all quantum gates are unitary matrices, for which their inverse equals their adjoint ($U^{-1} = U^\dagger$). This means that the eigenvalues of their eigenvectors have the property that their modulus equals $1$: 
+
+$$|\lambda| = 1$$
+
+Thus, they can be written in the following form:
+$$\lambda = e^{i\theta}$$
+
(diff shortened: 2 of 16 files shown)


katas/content/phase_estimation/index.md
```diff
@@ -0,0 +1,124 @@
...
+> $$(|\lambda| \cdot |\ket{v}|)^2 = |\ket{v}|^2$$
+> And then, finally:
+> $$|\lambda| = 1$$
+
+If the quantum gate is self-adjoint, that is, its matrix equals its inverse $U^{-1} = U$, the eigenvalues of this matrix can only be $+1$ and $-1$, with eigenphases $0$ and $\pi$, respectively.
+
+> You can prove this in a similar manner, using the defintion of an eigenvalue:
+> $$U^2 \ket{v} = U(U \ket{v}) = U(\lambda \ket{v}) = \lambda U \ket{v} = \lambda^2 \ket{v}$$
+> At the same time,
+> $$U^2 \ket{v} = UU \ket{v} = U^(-1)U \ket{v} = I \ket{v} = \ket{v}$$
```
ScottCarda-MS, line 59:
```suggestion
> $$U^2 \ket{v} = UU \ket{v} = U^{-1}U \ket{v} = I \ket{v} = \ket{v}$$
```

katas/content/phase_estimation/index.md
```diff
@@ -0,0 +1,124 @@
...
+The phase estimation problem is formulated as follows. 
+
+You are given a unitary operator $U$ and its eigenvector $\ket{\psi}$. The eigenvector is given as a way to prepare it,
+for example, a unitary $P$ that, when applied to the $\ket{0}$ state, converts it to the state $\ket{\psi}$.
+
+Your goal is to find the eigenvalue $\lambda$ associated with this eigenvector, or, in a more common formulation, the corresponding eigenphase $\theta$:
+
+$$U\ket{\psi} = e^{2 \pi i \theta} \ket{\psi}, \theta = ?$$
+
+The value of $\theta$ is defined to be between $0$ and $1$, since any value outside of this range has an equivalent value within it. Instead of discussing the floating-point number $\theta$, sometimes its representation as a binary fraction with $n$ digits is used:
```
ScottCarda-MS, line 100: nit:
```suggestion
The value of $\theta$ is defined to be between $0$ and $1$, since any value outside of this range has an equivalent value within it. Instead of representing $\theta$ as a decimal, sometimes it is represented as a binary fraction with $n$ digits:
```

katas/content/phase_estimation/state_eigenvector/index.md
```diff
@@ -0,0 +1,12 @@
...
+1. A single-qubit unitary $U$.
+2. A single-qubit state $\ket{\psi}$, given as a unitary $P$ that prepares it from the $\ket{0}$ state. In other words, the result of applying the unitary $P$ to the state $\ket{0}$ is the $\ket{\psi}$ state:
+   $$P\ket{0} = \ket{\psi}$$
+
+**Output:** Return true if the given state is an eigenstate of the given unitary, and false otherwise.
+
+<details>
+  <summary><b>Need a hint?</b></summary>
+  The library operation <code>CheckZero</code> allows you to check whether the state of the given qubit is $\ket{0}$.
+</details>
```
ScottCarda-MS, line 12: There needs to be some extra spacing here, otherwise the parser gets confused and doesn't render the $\ket{0}$ correctly.
```suggestion
<details>
  <summary><b>Need a hint?</b></summary>

The library operation <code>CheckZero</code> allows you to check whether the state of the given qubit is $\ket{0}$.

</details>
```

katas/content/phase_estimation/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+@[section]({
+    "id": "phase_estimation__problem",
+    "title": "Phase Estimation Problem"
+})
+
+The phase estimation problem is formulated as follows. 
+
+You are given a unitary operator $U$ and its eigenvector $\ket{\psi}$. The eigenvector is given as a way to prepare it,
+for example, a unitary $P$ that, when applied to the $\ket{0}$ state, converts it to the state $\ket{\psi}$.
```
ScottCarda-MS, line 94:
```suggestion
You are given a unitary operator $U$ and its eigenvector $\ket{\psi}$. The eigenvector is given as a unitary operator $P$ that, when applied to $\ket{0}$, results in the state $\ket{\psi}$.
```

ScottCarda-MS (approved):
Just a few suggestions, otherwise looks good.

katas/content/phase_estimation/index.md
tcNickolas: Technically the eigenstate doesn't have to be a unitary - it can be the result of a more complicated process than that. But I guess for the sake of streamlining the explanation it makes sense, we're not going to see any other scenarios in this kata

katas/content/phase_estimation/eigenvectors_x/Verification.qs
```diff
@@ -0,0 +1,32 @@
...
+        if Length(actual) != 2 {
+            Message("The array of eigenvectors should have exactly two elements.");
+            return false;
+        }
+        for i in 0 .. 1 {
+            if Length(actual[i]) != 2 {
+                Message("Each eigenvector should have exactly two elements.");
+                return false;
+            }
+            if AbsD(actual[i][0]) + AbsD(actual[i][1]) < 1E-9 {
```
DmitryVasilevsky, line 16: There's 0.001 tolerance a bit earlier. Should those be unified or there's a reason to keep them different?

DmitryVasilevsky (approved):

katas/content/phase_estimation/eigenvectors_x/Verification.qs
tcNickolas: In both tasks the return values are hardcoded rather than computed, so the check tolerance is driven by gut feeling rather than by exact science. 

The 0.001 tolerance was for the eigenvalues task (it's the same one used in complex arithmetic kata); there, the answer is specific values, and we're checking to see that those values were entered (the other option is completely incorrect values, rather than values that are slightly off; the learner is extremely unlikely to enter something like 1.005). 

The 1E-9 tolerance is for the task that involves any non-zero vectors; vectors like [0.0001, 0.0001] are valid eigenvectors, so with the check I want to make sure that the learner actually entered a vector that satisfies the condition and didn't leave [0, 0] in there (that vector also happens to satisfy the condition, but it's not an eigenvector because it's zero).

PR #1825, opened 2024-08-07
Fixed some broken latex in the Measure doc

### LINKED ISSUE #1429: Doc comment is not rendered correctly for Measure operation
Playground doesn't render Measure doc comment correctly.
Also \boldone is not supported.

- Make sure that the doc comment for Measure operation is rendered correctly in the playground.
- Remove \boldone from doc comments.

As a result of discussion with @vadym-kl we arrived at the following formula

$\langle\psi| \frac{I + P_0 \otimes \ldots \otimes P_{N-1}}{2} |\psi\rangle$

We'll use it in the doc comment for the Measure operation. We'll also drop the non-standard notation for probability

$\Pr(\texttt{Zero} \mid \ket{\psi})$

ScottCarda-MS:
What the Remarks section looked like before:
image

What the Remarks section looks like now:
image

Fixes #1429

### DIFF
--- a/library/std/src/intrinsic.qs
+++ b/library/std/src/intrinsic.qs
@@ -304,14 +304,6 @@ namespace Microsoft.Quantum.Intrinsic {
     /// Performs a joint measurement of one or more qubits in the
     /// specified Pauli bases.
     ///
-    /// # Description
-    /// The probability of getting `Zero` is
-    /// $\bra{\psi} \frac{I + P_0 \otimes \ldots \otimes P_{N-1}}{2} \ket{\psi}$
-    /// where $P_i$ is the $i$-th element of `bases`, and where
-    /// $N$ is the `Length(bases)`.
-    /// That is, measurement returns a `Result` $d$ such that the eigenvalue of the
-    /// observed measurement effect is $(-1)^d$.
-    ///
     /// If the basis array and qubit array are different lengths, then the
     /// operation will fail.
     ///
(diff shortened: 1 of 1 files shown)


DmitryVasilevsky (approved):

library/std/src/intrinsic.qs
```diff
@@ -368,17 +368,12 @@ namespace Microsoft.Quantum.Intrinsic {
     /// The output result is given by the distribution:
     /// $$
     /// \begin{align}
-    ///     \Pr(\texttt{Zero} | \ket{\psi}) =
```
tcNickolas, line 371: I think this line got dropped, so the rewrite is not just replacing boldone with I, it changes the formula? And it doesn't look like it should be the breaking one, worst case scenario we can replace \Pr with text

library/std/src/intrinsic.qs
```diff
@@ -368,17 +368,12 @@ namespace Microsoft.Quantum.Intrinsic {
...
-    ///         \frac12 \braket{
-    ///             \psi \mid|
-    ///             \left(
-    ///                 \boldone + P_0 \otimes P_1 \otimes \cdots \otimes P_{N-1}
-    ///             \right) \mid|
-    ///             \psi
-    ///         },
+    ///     \langle\psi|
```
tcNickolas, line 371: Do our docs understand \bra and \ket macros instead of \langle and \rangle? I know the katas do and I'm trying to use them for readability, but I haven't had the chance to check the docs

library/std/src/intrinsic.qs
DmitryVasilevsky: Some time ago we decided that \Pr(\texttt{Zero} | \ket{\psi}) is not a standard notation so we left this out of the formula, but we should say this in the text I was under impression that we fixed this somewhere. Maybe there's another place where it is already fixed, and we can take text from there.

tcNickolas: Looks standard to me, probability of measuring zero when measuring \psi?
But the new expression alone is not a clear definition of distribution, some edit to text might help

DmitryVasilevsky: OK, I found the fixed version. Right above this variant of the function there's another one for @Config(QubitReset). That one is where description has been fixed. We should reuse that description (and probably the entire doc comment.
This also brings an interesting question - different variants of functions for different capabilities may reuse doc comments, but we don't support such reuse.

library/std/src/intrinsic.qs
DmitryVasilevsky: That fixed version I mentioned in a different comment uses bra and ket.

library/std/src/intrinsic.qs
ScottCarda-MS: I've copied over the description from the @config(QubitReset) version, as suggested, with the minor change of renaming that section from "Description" to "Remarks" to fit the format of the other doc strings in the file.

library/std/src/intrinsic.qs
ScottCarda-MS: Yes, the updated description now uses \bra and \ket.

PR #1828, opened 2024-08-09
Allow the parser to add contextual help text to errors; Add contextual help text for parenthesized for loops

sezna:
Given the problems users have faced with non-parenthesized for loop iterators ([like this one](https://stackoverflow.com/questions/78592805/q-sharp-syntax-error-in-quantum-hhl-algorithm)), we wanted to provide some help text for the user in this situation.

Instead of just patching in a fix for precisely this situation, I wanted to use `miette`'s great functionality for composable errors. So, with this change, can now compose any parse error with help text.

I also manually implemented `Debug` for `Error` to omit the help text if it isn't present. This is primarily so this PR doesn't add `None` to every single parser expect test which tests errors. (Happy to revert this and update the tests if desired, but I really don't like spraying the diff with 1k lines of expect-test changes that just say `None`)

### DIFF
--- a/compiler/qsc_parse/src/expr.rs
+++ b/compiler/qsc_parse/src/expr.rs
@@ -170,13 +170,12 @@ fn expr_base(s: &mut ParserContext) -> Result<Box<Expr>> {
         let vars = match pat(s) {
             Ok(o) => o,
             Err(e) => {
-                if let (TokenKind::Open(Delim::Paren), ErrorKind::Token(expected, received, span)) = (peeked.kind,& e.0) {
-                    return Err(Error(ErrorKind::RuleWithHelp(
-                                *expected, *received, *span, "parenthesis are not permitted around for loop iterations"
-                    )));
-                } else {
-                    return Err(e);
+                if let (TokenKind::Open(Delim::Paren), ErrorKind::Token(..)) = (peeked.kind, &e.0) {
+                    return Err(
+                        e.with_help("parenthesis are not permitted around for-loop iterations")
+                    );
                 }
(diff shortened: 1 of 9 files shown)


compiler/qsc_parse/src/expr.rs
```diff
@@ -166,7 +166,18 @@ fn expr_base(s: &mut ParserContext) -> Result<Box<Expr>> {
     } else if token(s, TokenKind::Keyword(Keyword::Fail)).is_ok() {
         Ok(Box::new(ExprKind::Fail(expr(s)?)))
     } else if token(s, TokenKind::Keyword(Keyword::For)).is_ok() {
-        let vars = pat(s)?;
+        let peeked = s.peek();
+        let vars = match pat(s) {
+            Ok(o) => o,
+            Err(e) => {
+                if let (TokenKind::Open(Delim::Paren), ErrorKind::Token(..)) = (peeked.kind, &e.0) {
```
swernli, line 173: Would it be possible to check the error contents for the "in" keyword to make this even more specific?

sezna: Yes, that should work. I'm thinking about if that would miss any cases...but I think that should work? I'll add some tests to exercise it.

sezna: Updated in [db2c617](https://github.com/microsoft/qsharp/pull/1828/commits/db2c617546d3843c3390c321a18f8d19a03ed9a4).

swernli (approved):

ScottCarda-MS (approved):

PR #1829, opened 2024-08-10
Add edit/run for kata examples - issue 591

ggridin:
Add the edit/run for kata examples.
This PR closes https://github.com/microsoft/qsharp/issues/591

Monaco editor model should have a unique URI (with few exceptions). So, either kata lesson ID or kata exercise ID is provided to the model. The Editor.tsx implementation is replacing "kataExercise" parameter with "kataSection" that contains both lesson.id and exercise.id.
The function parameter "key" is not in use and was removed.

### DIFF
--- a/playground/src/editor.tsx
+++ b/playground/src/editor.tsx
@@ -19,6 +19,8 @@ import { Exercise, getExerciseSources } from "qsharp-lang/katas-md";
 import { codeToCompressedBase64, lsRangeToMonacoRange } from "./utils.js";
 import { ActiveTab } from "./main.js";
 
+import type { KataSection } from "qsharp-lang/katas";
+
 type ErrCollection = {
   checkDiags: VSDiagnostic[];
   shotDiags: VSDiagnostic[];
@@ -80,7 +82,7 @@ export function Editor(props: {
   compilerState: CompilerState;
   defaultShots: number;
   evtTarget: QscEventTarget;
-  kataExercise?: Exercise;
+  kataSection?: KataSection;
   onRestartCompiler: () => void;
   shotError?: VSDiagnostic;
   showExpr: boolean;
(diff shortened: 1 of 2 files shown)


ggridin:
@swernli Stefan, thank you for the phantom error root cause and monaco editor logic explanation.
I think I have the right fix now.

Testing done:
1. Successful and failed runs in example editor window.
2. Hover and autocomplete in example editor window.
3. Linked code.
4. Kata exercises and kata samples are working as before.
5. Code reset.

ggridin:
Follow-up questions:
1. Kata exercise and example profile is "unrestricted" only. **Should profile selection be deleted from exercises and examples?** Sample profile selection remains as is.
2. Existing problem found and reproduced in https://microsoft.github.io/qsharp/
If a user selects "Deutsch Algorithm" tutorial, then selects "Deutsch-Jozsa" sample - the sample "Run" button is grayed out.
Existing-issue-sample-run-is-gray

The fix (most likely) should introduce a unique URI for samples and the linked code.
**Should I file a bug for this issue?**

swernli:
@ggridin Thanks, this is looking good! Regarding the questions,

> Kata exercise and example profile is "unrestricted" only. Should profile selection be deleted from exercises and examples? Sample profile selection remains as is.

I think the dropdown for selecting profile on katas, both exercises and examples, is not necessary. They always run in Unrestricted and should remain that way.

> The fix (most likely) should introduce a unique URI for samples and the linked code.
Should I file a bug for this issue?

Good catch! Yes, we should follow up on this and fix it, so an issue would be great. I think your suggestion will likely end up being right fix.

ggridin:
A bug is found: example's syntax error message shows for all examples and exercises. The fix is WIP

swernli:
Sorry for the noise in converting it out of draft, meant to click the button to run the pipeline!

ggridin:
After today's branch sync, the issue mentioned above is gone (not reproducible).

swernli (approved):

DmitryVasilevsky (approved):

PR #1830, opened 2024-08-12
Fix test for multi_qubit_measurements/state_modification

tcNickolas:
Doing Reset on the first qubit before checking the state of the second qubit can lead to empty solution being judged as correct (with low probability). We have to check the state of the second qubit first and only then reset the first one.

### DIFF
--- a/katas/content/multi_qubit_measurements/state_modification/Verification.qs
+++ b/katas/content/multi_qubit_measurements/state_modification/Verification.qs
@@ -37,20 +37,21 @@ namespace Kata.Verification {
                 // Prepare the state to be input to the testImplementation
                 StateInitialize_StateSelection(alpha, qs);
 
-                // operate testImplementation
                 Kata.StateSelection(qs, ind);
-                // reset the first qubit, since its state does not matter
-                Reset(qs[0]);
 
-                // apply adjoint reference operation and check that the result is correct
+                // Apply adjoint of state preparation operation
                 Adjoint StatePrepare_StateSelection(alpha, ind, qs[1]);
 
-                if not CheckAllZero(qs) {
+                // We only care about the state of the second qubit; 
+                // if it's still entangled with the first one or not in zero state, this check will fail.
+                if not CheckZero(qs[1]) {
                     ResetAll(qs);
                     Message("Incorrect.");
                     Message($"The state of the second qubit for {params}, ind = {ind} does not match expectation.");
                     return false;
                 }
+
+                Reset(qs[0]);
             }
         }
         Message("Correct!");


katas/content/multi_qubit_measurements/state_modification/Verification.qs
```diff
@@ -37,20 +37,21 @@ namespace Kata.Verification {
...
-                // operate testImplementation
                 Kata.StateSelection(qs, ind);
-                // reset the first qubit, since its state does not matter
-                Reset(qs[0]);
 
-                // apply adjoint reference operation and check that the result is correct
+                // Apply adjoint of state preparation operation
                 Adjoint StatePrepare_StateSelection(alpha, ind, qs[1]);
 
-                if not CheckAllZero(qs) {
```
DmitryVasilevsky, line 48: @tcNickolas , gates on two different qubits always commute. Even measurements. What is the case that didn't work? The exact reason is probably subtly different?

tcNickolas: Reset is not a gate, it's a measurement, and CheckAllZero is not a gate either. A measurement collapses the state of the system. 

We start in ∣0⟩⊗(a∣0⟩+b∣1⟩) + ∣1⟩⊗(b∣0⟩+a∣1⟩) (not spelling out normalization coefficients here)

Empty solution leaves this state unchanged, so when we apply adjoint of prep to the second qubit, we end up in 
∣0⟩⊗∣0⟩ + ∣1⟩⊗|non-0⟩. CheckAllZero sees the amplitude of the second basis state and returns false, which is correct

If we measure the first qubit before CheckAllZero, the state collapses, we end up either in ∣0⟩⊗∣0⟩ or in ∣1⟩⊗|non-0⟩ with certain probabilities, and if we hit the first scenario, CheckAllZero will return true.

If we hit the first scenario on each loop iteration, the test will pass for the empty solution

DmitryVasilevsky: Thanks for explaining. As far as I understood it, the 'Check' is not a gate (completely true) and therefore putting it after reset makes the whole thing probabilistic. It still verifies solution but is not always correct. While keeping it before reset peaks into state and therefore is non-probabilistic.

DmitryVasilevsky (approved):

katas/content/multi_qubit_measurements/state_modification/Verification.qs
tcNickolas: Yes, doing Reset before check introduces this uncertainty, since it effectively emulates part of the correct solution.

swernli (approved):

PR #1813, opened 2024-08-02
Katas UI review - Multiqubit systems

SoniaLopezBravo:

PR #1832, opened 2024-08-13
Katas UI review - Multiqubit measurements 

SoniaLopezBravo:
Review only files inside *multi_qubit_measurements* folder


tcNickolas:
This looks like it touches a lot of the same files as #1813, so this is going to have conflicts... I would recommend sending this PR after that one is merged

SoniaLopezBravo:
> This looks like it touches a lot of the same files as #1813, so this is going to have conflicts... I would recommend sending this PR after that one is merged

@tcNickolas I merged the branch from #1813 , so there shouldn't be any conflict. You only need to review files inside the multi_qubit_measurements folder.

PR #1833, opened 2024-08-13
Fix missing docs for diagnostic callables

### LINKED ISSUE #1822: Missing API docs for CheckZero, CheckAllZero, DumpMachine
**Describe the bug**

There are no API docs for CheckZero, CheckAllZero, DumpMachine on the website.

**To Reproduce**

* https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.diagnostics/checkzero
* https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.diagnostics/checkallzero
* https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/microsoft.quantum.diagnostics/dumpmachine

swernli:
Fixes #1822

### DIFF
--- a/library/std/src/diagnostics.qs
+++ b/library/std/src/diagnostics.qs
@@ -70,10 +70,10 @@ namespace Microsoft.Quantum.Diagnostics {
     }
 
     /// # Summary
-    /// Checks whether a qubit is in the |0⟩ or ground state, returning true if it is.
+    /// Checks whether a qubit is in the |0⟩ state (the ground state), returning true if it is.
     ///
     /// # Description
-    /// This operation checks whether a qubit is in the |0⟩ or ground state. It will return true only
+    /// This operation checks whether a qubit is in the |0⟩ state (the ground state). It will return true only
     /// if the qubit is deterministically in the |0⟩ state, and will return false otherwise. This operation
     /// does not change the state of the qubit.
     ///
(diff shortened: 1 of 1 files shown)


library/std/src/diagnostics.qs
```diff
@@ -69,11 +69,37 @@ namespace Microsoft.Quantum.Diagnostics {
         body intrinsic;
     }
 
+    /// # Summary
+    /// Checks whether a qubit is in the |0⟩ or ground state, returning true if it is.
```
DmitryVasilevsky, line 73: "|0⟩ or ground state" gives an impression that those are different states. May be use different wording if possible? Like "AKA ground state" or something?

library/std/src/diagnostics.qs
```diff
@@ -69,11 +69,37 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    ///
+    /// # Description
+    /// This operation checks whether a qubit is in the |0⟩ or ground state. It will return true only
+    /// if the qubit is deterministically in the |0⟩ state, and will return false otherwise. This operation
+    /// does not change the state of the qubit.
```
DmitryVasilevsky, line 78: Maybe have a remarks section that such check wouldn't be possible on an actual hardware?

DmitryVasilevsky (approved):

library/std/src/diagnostics.qs
```diff
@@ -69,11 +69,45 @@ namespace Microsoft.Quantum.Diagnostics {
         body intrinsic;
     }
 
+    /// # Summary
+    /// Checks whether a qubit is in the |0⟩ state (the ground state), returning true if it is.
```
tcNickolas, line 73: I don't think we're using "ground state" anywhere else to refer to the |0⟩ state, so I would not bring this in this doc. If somebody's writing Q#, they most likely know what |0⟩ is; and "ground state" is the kind of terminology I'd expect to see in chemistry apps, not in the basic concepts

tcNickolas (approved):

library/std/src/diagnostics.qs
swernli: I'll just remove it then, it's simpler that way.

library/std/src/diagnostics.qs
swernli: Added

library/std/src/diagnostics.qs
swernli: Removed, per Mariia's suggestion

PR #1838, opened 2024-08-13
Port fixed point library to modern QDK

sezna:
This PR ports the classic QDK FixedPoint library to the modern QDK. There is still some work to be done (a README, some tests), but this is a good MVP for the library to go in 1.8.


library/fixed_point/src/Tests.qs
```diff
@@ -0,0 +1,104 @@
...
+    Fact(AbsD(a * a - measured) < 0.001, $"Difference of {AbsD(a * a - measured)} is outside of the expected range. Expected {a * a} and measured result was {measured}. (Inputs were Square({a})");
+    ResetAll(resultRegister);
+    ResetAll(aRegister);
+}
+
+// assume the second register that `op` takes is the result register
+operation TestOperation(a : Double, b : Double, op : (FixedPoint, FixedPoint) => (), reference : (Double, Double) -> Double, name : String) : Unit {
+    Message($"Testing {name}({a}, {b})");
+    use register1 = Qubit[20];
+    let aFxp = new FixedPoint { IntegerBits = 8, Register = register1 };
```
DmitryVasilevsky, line 66: When you do this testing, you are always using numbers of the same structure (20 bits overall, 8 integer qubits). I suggest you vary the structure as well on both arguments. Fixed points with the same number of digits after fractional point are compatible. OTOH, there's no need to verify all 50x50 cases for the number values.

sezna: I added tests testing various widths of FixedPoints in this commit: [8f2e0b7](https://github.com/microsoft/qsharp/pull/1838/commits/8f2e0b70cab72da981a5b94605d47b7009176903)

I discovered some details -- we require at least two integer bits and one fixed point bit for correctness, so I documented that.

library/fixed_point/qsharp.json
```diff
@@ -0,0 +1,35 @@
...
+      }
+    },
+    "Unstable": {
+      "github": {
+        "owner": "Microsoft",
+        "repo": "qsharp",
+        "ref": "d5f0fbcf",
+        "path": "library/unstable"
+      }
+    }
```
swernli, line 20: nitpick: It's not a problem, but does read a little odd that the two libraries this depends on come from the same repo but with different refs.

library/fixed_point/src/Operations.qs
```diff
@@ -0,0 +1,172 @@
...
+/// # Summary
+/// Adds a classical constant to a quantum fixed-point number.
+///
+/// # Input
+/// ## constant
+/// Constant to add to the quantum fixed-point number.
+/// ## fp
+/// Fixed-point number to which the constant will
+/// be added.
+operation AddConstantFxP(constant : Double, fp : FixedPoint) : Unit is Adj + Ctl {
```
swernli, line 20: nitpick: It took me a moment to realize that by "Operations" you meant mathematical operations and not Q# callables that are not functions. Not sure I have a better name, just noting the terminology overloading for a namespace name.

swernli (approved):
Looks good with two minor nitpicks.

library/fixed_point/qsharp.json
sezna: It was because #1888 wasn't in yet. I'll update it now that that's merged.

library/fixed_point/src/Operations.qs
sezna: Hm, what about `MathOps.qs`?

library/fixed_point/qsharp.json
sezna: addressed in [3d6bbcd](https://github.com/microsoft/qsharp/pull/1838/commits/3d6bbcd9c577fbb0b81431fa2f945de0eede01fd)

library/fixed_point/src/Operations.qs
swernli: That seems reasonable to me!

sezna: I realized I've already got an `Operations.qs` in `Signed` -- maybe we keep these the same for now, for consistency, and we can iterate the API in a follow-up PR?

swernli: Ah, ok. Let's keep it consistent then.

library/fixed_point/src/Facts.qs
```diff
@@ -0,0 +1,62 @@
...
+
+/// # Summary
+/// Assert that all fixed-point numbers in the provided array
+/// have identical point positions and qubit numbers.
+///
+/// # Input
+/// ## fixedPoints
+/// Array of quantum fixed-point numbers that will be checked for
+/// compatibility (using assertions).
+function IdenticalFormatFactFxP(fixedPoints : FixedPoint[]) : Unit {
```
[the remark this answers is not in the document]
cesarzc: I also expanded `FxP` to `FixedPoint`, but that is just my preference so that's a nit on top of a nit

library/fixed_point/src/Facts.qs
```diff
@@ -0,0 +1,62 @@
...
+/// Assert that all fixed-point numbers in the provided array
+/// have identical point positions when counting from the least-
+/// significant bit. I.e., number of bits minus point position must
+/// be constant for all fixed-point numbers in the array.
+///
+/// # Input
+/// ## fixedPoints
+/// Array of quantum fixed-point numbers that will be checked for
+/// compatibility (using assertions).
+function IdenticalPointPosFactFxP(fixedPoints : FixedPoint[]) : Unit {
```
cesarzc, line 50: nit: similar to the other comment, a different name for this could be `AssertPointPositionsIdenticalFxP`.

library/fixed_point/src/Facts.qs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// # Input
+/// ## fixedPoints
+/// Array of quantum fixed-point numbers that will be checked for
+/// compatibility (using assertions).
+function IdenticalFormatFactFxP(fixedPoints : FixedPoint[]) : Unit {
+    if IsEmpty(fixedPoints) {
+        return ();
+    }
+    let (position, register) = fixedPoints[0]!;
```
cesarzc, line 31:
```suggestion
    let (position, register) = Head(fixedPoints)!;
```

cesarzc: this is also a nit :P

library/fixed_point/src/Facts.qs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// # Input
+/// ## fixedPoints
+/// Array of quantum fixed-point numbers that will be checked for
+/// compatibility (using assertions).
+function IdenticalPointPosFactFxP(fixedPoints : FixedPoint[]) : Unit {
+    if IsEmpty(fixedPoints) {
+        return ();
+    }
+    let (position, register) = fixedPoints[0]!;
```
cesarzc, line 54:
```suggestion
    let (position, register) = Head(fixedPoints)!;
```

cesarzc: This is also a nit :P

swernli: This one is an interesting one... `Head(arr)` is more readable than `arr[0]` sometimes, but it's more expensive to execute (pays for a function call boundary plus array access, instead of just array access). For efficiency in a library, I would advise avoiding this kind of simple utility function call due to their execution overhead.

library/fixed_point/src/Measurement.qs
```diff
@@ -0,0 +1,18 @@
...
+
+/// # Summary
+/// Measure a fixed-point number, returns its value as Double, and resets
+/// all the register to zero.
+///
+/// # Input
+/// ## fp
+/// Fixed-point number to measure.
+operation MeasureFxP(fp : FixedPoint) : Double {
+    let bits = ForEach(q => MResetZ(q) == One, fp::Register);
```
cesarzc, line 16:
```suggestion
    let measurements = MResetEachZ(fs::Register);
    let bits = ResultArrayAsBoolArray(measurements);
```

cesarzc: This is a suggestion to make the library more friendly to QIR generation. We currently don't support dynamic `Double` types so realistically it won't have any effect immediately, but just something to keep in mind.

cesarzc: We can always do an effor to make it more hardware friendly later, so feel free to ignore this comment.

library/fixed_point/src/Operations.qs
```diff
@@ -0,0 +1,172 @@
...
+/// # Input
+/// ## fp
+/// Fixed-point number to invert.
+///
+/// # Remarks
+/// Numerical inaccuracies may occur depending on the
+/// bit-precision of the fixed-point number.
+///
+/// # See also
+/// - Microsoft.Quantum.Arithmetic.SubtractFxP
```
cesarzc, line 67: Should these "See also" sections be removed? It seems like they are pointing to namespaces that are not part of the Modern QDK.

library/fixed_point/src/Operations.qs
```diff
@@ -0,0 +1,172 @@
...
+/// ## minuend
+/// The minuend of the subtraction - the number from which the other is subtracted.
+///
+/// # Remarks
+/// Computes the difference by inverting `subtrahend` before and after adding
+/// it to `minuend`.  Notice that `minuend`, the first argument is updated.
+///
+/// # See also
+/// - Microsoft.Quantum.Arithmetic.AddFxP
+/// - Microsoft.Quantum.Arithmetic.InvertFxP
```
cesarzc, line 88: Similar comment here, it seems like these "See also" sections probably need to be removed.

library/fixed_point/src/Operations.qs
```diff
@@ -0,0 +1,172 @@
...
+    controlled (controls, ...) {
+        IdenticalFormatFactFxP([fp1, fp2, result]);
+        let n = Length(fp1::Register);
+
+        use tmpResult = Qubit[2 * n];
+        let xsInt = ((fp1::Register));
+        let ysInt = ((fp2::Register));
+        let tmpResultInt = (
+            (tmpResult)
+        );
```
cesarzc, line 126: What is the purpose of tuple wrapping this Qubit register?

library/fixed_point/src/Operations.qs
```diff
@@ -0,0 +1,172 @@
...
+    }
+    controlled (controls, ...) {
+        IdenticalFormatFactFxP([fp, result]);
+        let n = Length(fp::Register);
+
+        use tmpResult = Qubit[2 * n];
+        let xsInt = fp::Register;
+        let tmpResultInt = (
+            (tmpResult)
+        );
```
cesarzc, line 157: Similar comment here, what is the purpose of tuple wrapping this register?

library/fixed_point/src/Types.qs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// # Summary
+/// Represents a register of qubits encoding a fixed-point number. Consists of an integer that is equal to the number of
+/// qubits to the left of the binary point, i.e., qubits of weight greater
+/// than or equal to 1, and a quantum register.
+/// There should always be at leat 2 integer bits in the fixed-point number. The first one represents the sign, the second one is the most significant bit of the int.
```
cesarzc, line 8:
```suggestion
/// There should always be at least 2 integer bits in the fixed-point number. The first one represents the sign, the second one is the most significant bit of the int.
```

cesarzc (approved):

library/fixed_point/src/Operations.qs
sezna: I think this is just an artifact of some refactoring. fixed

PR #1841, opened 2024-08-14
Port signed integer math to modern QDK

sezna:
This PR ports the classic QDK's signed int math functionality to the modern QDK.

It is not intended to provide the perfect API -- it is just a port, that other libraries will build upon, primarily the fixed point library.


library/signed/src/RippleCarry.qs
```diff
@@ -0,0 +1,418 @@
...
+/// # References
+/// - Yasuhiro Takahashi, Seiichiro Tani, Noboru Kunihiro: "Quantum
+///   Addition Circuits and Unbounded Fan-Out", Quantum Information and
+///   Computation, Vol. 10, 2010.
+///   https://arxiv.org/abs/0910.2530
+///
+/// # Remarks
+/// This operation has the same functionality as RippleCarryAdderD and,
+/// RippleCarryAdderCDKM but does not use any ancilla qubits.
+operation RippleCarryAdderTTK(xs : Qubit[], ys : Qubit[], carry : Qubit) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 307: A few functions are already here: https://github.com/microsoft/qsharp/blob/9ae1d2a1ae9c4a885b88d2586302ca53555b5e0f/library/unstable/src/Arithmetic.qs#L159

sezna: Thanks, given the naming differences I wasn't sure if those are actually the same function

sezna: I've removed the port of the old ripple carry and am using `library/unstable` now. Thanks!

library/signed/src/Comparison.qs
```diff
@@ -0,0 +1,104 @@
...
+operation CompareGTSI(xs : Qubit[], ys : Qubit[], result : Qubit) : Unit is Adj + Ctl {
+    use tmp = Qubit();
+    CNOT(Tail(xs), tmp);
+    CNOT(Tail(ys), tmp);
+    X(tmp);
+    Controlled CompareGTI([tmp], (xs, ys, result));
+    X(tmp);
+    CCNOT(tmp, Tail(ys), result);
+    CNOT(Tail(xs), tmp);
+    CNOT(Tail(ys), tmp);
```
cesarzc, line 26:
```suggestion
    within {
        CNOT(Tail(xs), tmp);
        CNOT(Tail(ys), tmp);
    } apply {
        X(tmp);
        Controlled CompareGTI([tmp], (xs, ys, result));
        X(tmp);
        CCNOT(tmp, Tail(ys), result);    
    }
```

library/signed/src/Comparison.qs
```diff
@@ -0,0 +1,104 @@
...
+///     https://arxiv.org/abs/1611.07995
+///
+/// # Remarks
+/// Uses the trick that $x - y = (x'+y)'$, where ' denotes the one's complement.
+operation GreaterThan(xs : Qubit[], ys : Qubit[], result : Qubit) : Unit is Adj + Ctl {
+    body (...) {
+        (Controlled GreaterThan)([], (xs, ys, result));
+    }
+    controlled (controls, ...) {
+        let nQubits = Length(xs);
```
cesarzc, line 78: It seems like the implementation assumes that the length of `xs` and `ys` is the same. If that is the case, then it would be good to add a `Fact` statement.

library/signed/src/Measurement.qs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// ## width
+/// The width of the signed integer to be measured.
+operation MeasureSignedInteger(target : Qubit[], width : Int) : Int {
+    let nBits = Length(target);
+    Fact(nBits <= 64, $"`Length(target)` must be less than or equal to 64, but was {nBits}.");
+    Fact(nBits <= width, $"`Length(target)` must be less than or equal to `width`, but was {nBits}.");
+
+    mutable coefficient = 1;
+    let signBit = MResetZ(target[nBits - 1]);
```
cesarzc, line 31: I think this might be a bit inconsistent with the rest of the library in terms of endianness. Based on the signed bit being represented by the last qubit and on the way the number is calculated looping later on, this looks like big endian instead of little endian. Maybe I'm missing something or not understanding something properly, but I just wanted to make sure we are doing the right thing.

library/signed/src/Measurement.qs
```diff
@@ -0,0 +1,49 @@
...
+/// If the length of the qubit register passed in is greater than the width, this operation
+/// will throw an error.
+///
+/// # Input
+/// ## target
+/// A qubit register representing the signed integer to be measured.
+///
+/// ## width
+/// The width of the signed integer to be measured.
+operation MeasureSignedInteger(target : Qubit[], width : Int) : Int {
```
cesarzc, line 25: I think for this operation the `target` register length needs to be at least 1 so it would be good to add a `Fact` statement for that.

library/signed/src/Measurement.qs
cesarzc: Ok, so I have convinced myself that this is correct 😆 

Sorry for the churn!

library/signed/src/Operations.qs
```diff
@@ -0,0 +1,361 @@
...
+            use aux = Qubit();
+            for (idx, ctl) in Enumerated(xs) {
+                within {
+                    AND(controls[0], ctl, aux);
+                } apply {
+                    Controlled AddI([aux], (xs, (result[idx..idx + n])));
+                }
+            }
+        } else {
+            use helper = Qubit[numControls];
```
cesarzc, line 84: nit: `helper` could be renamed to `aux` to be consistent with the other conditional blocks.

library/signed/src/Operations.qs
```diff
@@ -0,0 +1,361 @@
...
+                        AND(Tail(Most(helper)), ctl, Tail(helper));
+                    } apply {
+                        Controlled AddI([Tail(helper)], (xs, (result[idx..idx + n])));
+                    }
+                }
+            }
+        }
+    }
+    adjoint auto;
+    controlled adjoint auto;
```
cesarzc, line 99: Are these needed if the operation is typed as `is Adj + Ctl`.

library/signed/src/Utils.qs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// Util functions for interal use by the signed integer library.
```
cesarzc, line 4:
```suggestion
// Util functions for internal use by the signed integer library.
```

cesarzc (approved):
Left a few minor comments.

library/signed/src/Operations.qs
sezna: I don't think so, they were just part of the port. Removed.

library/signed/src/Comparison.qs
```diff
@@ -0,0 +1,107 @@
...
+/// Wrapper for integer comparison: `result = x > y`.
+///
+/// # Input
+/// ## xs
+/// First $n$-bit number
+/// ## ys
+/// Second $n$-bit number
+/// ## result
+/// Will be flipped if $x > y$
+operation CompareGTI(xs : Qubit[], ys : Qubit[], result : Qubit) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 42: Do we need this operation?

library/signed/src/Comparison.qs
```diff
@@ -0,0 +1,107 @@
...
+    GreaterThan(xs, ys, result);
+}
+
+/// # Summary
+/// Applies a greater-than comparison between two integers encoded into
```
DmitryVasilevsky, line 47: If this is for signed integers, we should mention this in the comments. If this is for unsigned, we can use 'ApplyIfGreaterLE' from unstable library.

library/signed/src/Comparison.qs
sezna: we might not need to export it as part of the API, but it is used as a helper in `CompareGTSI` and `DivideI`, which is used in  `ComputeReciprocalI`. If we get rid of `ComputeReciprocalI`, we can get rid of `DivideI`, but not `CompareGTSI`.

library/signed/src/Operations.qs
```diff
@@ -0,0 +1,356 @@
...
+/// # Summary
+/// Automatically chooses between addition with
+/// carry and without, depending on the register size of `ys`.
+///
+/// # Input
+/// ## xs
+/// $n$-bit addend.
+/// ## ys
+/// Addend with at least $n$ qubits. Will hold the result.
+operation AddI(xs : Qubit[], ys : Qubit[]) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 232: I suspect you could use RippleCarryTTKIncByLE instead of this entire operation.

sezna: The tests still pass with this change, so I'll do it! Thanks.

library/signed/src/Utils.qs
```diff
@@ -0,0 +1,55 @@
...
+
+operation AndLadder(controls : Qubit[], targets : Qubit[]) : Unit is Adj {
+    let controls1 = [Head(controls)] + Most(targets);
+    let controls2 = Rest(controls);
+    for (a, b, c) in Zipped3(controls1, controls2, targets) {
+        AND(a, b, c);
+    }
+}
+
+function Zipped3<'T1, 'T2, 'T3>(first : 'T1[], second : 'T2[], third : 'T3[]) : ('T1, 'T2, 'T3)[] {
```
DmitryVasilevsky, line 18: We try to avoid using Zipped function, when possible, for faster simulation speed. Both AndLadder and ApplyCCNOTChain can be re-written using a simple loop.

library/signed/src/Comparison.qs
sezna: I don't fully understand -- `ApplyIfGreaterLE` has a pretty different API. How would you use it? It looks like it takes an operation as an input

library/signed/src/Comparison.qs
DmitryVasilevsky: What I mean, it just calls the underlying function without any modifications to arguments. Could we call it directly?

DmitryVasilevsky:
@sezna , Feel free to postpone all my comment for later time if you need. We can always improve code :)

DmitryVasilevsky (approved):

library/signed/src/Comparison.qs
sezna: Oh, I see. Yes, updated. Good catch.

library/signed/src/Comparison.qs
DmitryVasilevsky: Use X for operation.

library/signed/src/Comparison.qs
```diff
@@ -0,0 +1,91 @@
...
+/// Wrapper for signed integer comparison: `result = xs > ys`.
+///
+/// # Input
+/// ## xs
+/// First $n$-bit number
+/// ## ys
+/// Second $n$-bit number
+/// ## result
+/// Will be flipped if $xs > ys$
+operation CompareGTSI(xs : Qubit[], ys : Qubit[], result : Qubit) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 18: Consider changing this to a more generic version that takes an operation that gets applied if condition is met, just like ApplyIfGreaterLE.
Consider making function names consistent with unstable arithmetic and probably addition more comparison functions. I suspect this will eventually reside next to the 'unsigned integer' arithmetic functions from unstable.
(Again, this may be future work)

library/signed/src/Comparison.qs
sezna: Oh, makes sense. Updated. Thanks!

library/signed/src/Comparison.qs
sezna: I filed #1874 to track this -- since this is primarily a port, I'm going to merge this in now and then update the API in a smaller follow-on PR.

PR #1858, opened 2024-08-19
Add samples of testing Q# code with classical return values

tcNickolas:
Part of #1268. 
Introduces a shared README.md for all testing samples and includes small edits to existing testing sample.


samples/testing/classical_values/src/TestCode.qs
```diff
@@ -0,0 +1,27 @@
+import Std.Diagnostics.Fact;
+open Std.Arrays;
+open Std.Random;
```
swernli, line 3: To do the equivalent of `open` with the new import syntax, you'll want to use glob imports:

```suggestion
import Std.Arrays.*;
import Std.Random.*;
```

swernli (approved):
Looking over this made me realize that our test infra doesn't validate stand-alone Python files in our samples. That is out of scope for this PR, so I'll file a follow up. I can see the Rust-based unit test infra automatically picked up the new project as expected, so that means this is consistent with our existing testing strategy.

samples/testing/classical_values/src/TestCode.qs
tcNickolas: That looks more verbose, is there an advantage to doing that instead of open?

cesarzc (approved):

PR #1863, opened 2024-08-20
Include completions for items from the same file; Clean up completions generation for callables

### LINKED ISSUE #1862: Callables from same file don't show up in the completions list

sezna:
This PR closes #1862.

I had previously touched this function in #1820. I did my best to port it and maintain existing functionality, but there was one conditional that I messed up, and it was around callables originating from the same file.

This PR fixes the bug that caused. But also, this PR tries to simplify `callable_decl_to_completion_item` so this is less likely to happen in the future.

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -811,11 +811,11 @@ fn callable_decl_to_completion_item(
     // cases, we do not need to add an import statement.
     // The third condition is if there is no existing imports, and no existing package alias (a
     // form of import), then we need to add an import statement.
-    let additional_text_edit =  
+    let additional_text_edit =
         // first condition
         if current_namespace_name == Some(namespace_as_strs.as_slice()) ||
         // second condition
-        (preexisting_exact_import || preexisting_glob_import) { None } 
+        (preexisting_exact_import || preexisting_glob_import) { None }
         // third condition
         else if preexisting_namespace_alias.is_none() {
             // if there is no place to insert an import, then we can't add an import.
@@ -829,7 +829,7 @@ fn callable_decl_to_completion_item(
                     new_text: format!("import {import_text};{indent}",),
                     range,
                 })
-            } else { None } 
+            } else { None }
         } else {
             None
         };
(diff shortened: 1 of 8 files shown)


language_service/src/completion.rs
```diff
@@ -765,105 +765,89 @@ fn callable_decl_to_completion_item(
...
-            });
+    // calculate the qualification that goes before the import
+    // item
+    // if an exact import already exists, or if that namespace
+    // is glob imported, then there is no qualification
+
+    // If there is no matching import or glob import, then the
+    // qualification is the full namespace name
+
+    // an exact import is an import that matches the namespace
```
swernli, line 785: nitpick: why the spaces between these comments? It seems like the could be all one paragraph with some punctuation between then. Not critical to functionality, of course.

swernli (approved):

language_service/src/completion.rs
sezna: I was just using newlines to separate my sentences. I've updated it to use the more canonical periods-based sentence structure that English provides 😄

language_service/src/completion.rs
```diff
@@ -765,105 +765,89 @@ fn callable_decl_to_completion_item(
     package_alias_from_manifest: &Option<Arc<str>>,
     callable_namespace: &qsc::hir::Idents,
     imports: &[ImportItem],
-    insert_open_at: Option<Range>,
+    insert_import_at: Option<Range>,
     indent: &String,
-) -> Option<(CompletionItem, SortPriority)> {
+) -> (CompletionItem, SortPriority) {
```
billti, line 770: I find this function borderline unreadable, but as it was before-hand, will approve to unblock. We should really focus on code clarity and readability more when authoring though, as it makes it really hard to feel confident that you would have spotted any logic bugs when reviewing.

billti (approved):

language_service/src/completion.rs
sezna: I have been working to incrementally improve this, but you're right, it could use a deeper refactor. I filed #1864 to track this.

PR #1870, opened 2024-08-21
Remove benchmarking comment bot; Run benches weekly and save results as an artifact

sezna:
The benchmark comments have generally been pretty noisy and I think unpopular, especially when iterating on one open PR.

This PR removes the current benchmark commenting functionality and instead runs a once-weekly (Monday morning at midnight) benchmark on `main` and saves the results as JSON. These results can then be easily interpreted by a separate tool that I'm putting together. This will help us catch any performance regressions or trends, but in a once-weekly retrospective report of all merges into main, instead of every single commit. 

The key thing here is that after this PR goes in, we will have a mechanism to retrospect over performance data without having it commented on every commit of every PR.

### DIFF
--- a/.github/workflows/bench-reports.yml
+++ b/.github/workflows/bench-reports.yml
@@ -21,7 +21,7 @@ jobs:
           fetch-depth: 0
           ref: alex/benching
       - name: setup rust
-        uses: actions-rs/toolchain@v3
+        uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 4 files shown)


orpuente-MS:
I agree with removing the current benchmarks from the pipeline, and having a trend of noisy benchmarks over time is more useful than having no data.

swernli (approved):

ScottCarda-MS (approved):

PR #1873, opened 2024-08-22
Add samples of testing Q# code that prepares a quantum state

### LINKED ISSUE #1268: Add samples of testing Q# code

tcNickolas:
Closes #1268.


samples/testing/states/README.md
```diff
@@ -0,0 +1,23 @@
...
+   Use the `check_eq()`  method of `StateDump` class to compare it to the given array of amplitudes, taking into account the possible global phase difference.
+
+
+## Project Structure
+
+This sample project is a multi-file Q# project that showcases both testing methods. The project structure is as follows:
+
+- src
+    - `StatePrep.qs`: Q# file containing the state preparation operations to be tested.
+- `qsharp.json`: Q# project manifest file, instructing compiler to include all files in `src` directory.
```
ScottCarda-MS, line 22:
```suggestion
- `qsharp.json`: Q# project manifest file, instructing the compiler to include all files in `src` directory.
```

ScottCarda-MS (approved):

samples/testing/states/src/StatePrep.qs
```diff
@@ -0,0 +1,18 @@
...
+/// Prepares the state 1/2 (|00⟩ + i|01⟩ - |10⟩ - i|11⟩) = |-⟩ ⊗ |i⟩.
+operation PrepareStateWithComplexPhases(qs : Qubit[]) : Unit {
+    H(qs[0]);
+    Z(qs[0]);
+    H(qs[1]);
+    S(qs[1]);
+}
+
+/// # Summary
+/// Prepares the state 1/2 (-|00⟩ - i|01⟩ + |10⟩ + i|11⟩) = -|-⟩ ⊗ |i⟩.
```
swernli, line 11: Maybe I'm just not used to the notation, but I would've expected the product version of the state to be either `-(|-⟩ ⊗ |i⟩)` or `-|-⟩ ⊗ -|i⟩` to drive home the fact that the -1 phase is global.

samples/testing/states/test_states.py
```diff
@@ -0,0 +1,34 @@
...
+def test_state_exact() -> None:
+    """Test that Q# code prepares the expected state exactly using Python test code."""
+    # Run Q# code that allocates the qubits and prepares the state but doesn't deallocate the qubits.
+    qsharp.eval(f"use qs = Qubit[2]; StatePrep.PrepareStateWithComplexPhases(qs);")
+    # Get the state of the allocated qubits and convert it to a dense vector.
+    state = qsharp.dump_machine().as_dense_state()
+    # Compare two vectors.
+    assert state == pytest.approx(expected_state)
+    # If you use PrepareStateWithGlobalPhase operation in this test instead, the test will fail,
+    # since the prepared state differs from the expected one by a global phase.
```
swernli, line 24: Rather than saying the test will fail with the other method, you could have a second test case that asserts they are not equal. Something like:
```python
def test_state_exact_rejects_global_phase() -> None:
    """Test that Q# code prepares the expected state exactly using Python test code."""
    # Run Q# code that allocates the qubits and prepares the state but doesn't deallocate the qubits.
    qsharp.eval(f"use qs = Qubit[2]; StatePrep.PrepareStateWithGlobalPhase(qs);")
    # Get the state of the allocated qubits and convert it to a dense vector.
    state = qsharp.dump_machine().as_dense_state()
    # Compare two vectors. Here we expect them to fail equality due to global phase.
    assert state != pytest.approx(expected_state)
```

swernli (approved):
Left a couple minor comments, but nothing that should prevent merging. Up to your discretion on those!

samples/testing/states/test_states.py
tcNickolas: Good point! I'm so used to checking that the state _is_ as expected, didn't think about a test to show the other way :-)

samples/testing/states/src/StatePrep.qs
tcNickolas: In the second notation the minuses cancel each other out? 
It's just like regular multiplication, the minus can go in front of the whole thing or of any one of the terms, and here I'm preparing the state as (-|-⟩) ⊗ |i⟩ to get the global phase out of the first term, so it looked more natural. I can add the term `-(|-⟩ ⊗ |i⟩)` between the two terms I have right now to show the line of thought

PR #1883, opened 2024-08-25
Added DoubleAsStringWithPrecision function

devikamehra:
Added DoubleAsStringWithPrecision function in Complex Arithmetics Kata

### DIFF
--- a/katas/content/complex_arithmetic/Common.qs
+++ b/katas/content/complex_arithmetic/Common.qs
@@ -10,19 +10,19 @@ namespace Kata.Verification {
         return Complex(real, imag);
     }
 
-    function ComplexAsString(x : Complex) : String {
+    function ComplexAsString(x : Complex, precision : Int) : String {
         if x::Imag < 0.0 {
-            $"{x::Real} - {AbsD(x::Imag)}i"
+            $"{DoubleAsStringWithPrecision(x::Real,precision)} - {DoubleAsStringWithPrecision(AbsD(x::Imag),precision)}i"
         } else {
-            $"{x::Real} + {x::Imag}i"
+            $"{DoubleAsStringWithPrecision(x::Real,precision)} + {DoubleAsStringWithPrecision(x::Imag,precision)}i"
         }
     }
 
-    function ComplexPolarAsString(x : ComplexPolar) : String {
-        $"{x::Magnitude} * exp({x::Argument}i)"
+    function ComplexPolarAsString(x : ComplexPolar, precision: Int) : String {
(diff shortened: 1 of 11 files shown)


tcNickolas (changes requested):
There is a way to reduce the scope of this change greatly and to make the code more readable. You're passing precision as an argument to the library code, so you have to define it in every task that uses the library. But the precision you use doesn't actually vary between the tasks, and the library not general-purpose but is limited to the needs of this one kata. You can instead hardcode precision within ComplexAsString and ComplexPolarAsString - this way you don't need to change any code outside the library!

Looking at the numbers involved, I don't think we need that much of precision too - none of them are smaller than 1, so 2-3 digits of precision should be plenty.

Thank you!

devikamehra:
> There is a way to reduce the scope of this change greatly and to make the code more readable. You're passing precision as an argument to the library code, so you have to define it in every task that uses the library. But the precision you use doesn't actually vary between the tasks, and the library not general-purpose but is limited to the needs of this one kata. You can instead hardcode precision within ComplexAsString and ComplexPolarAsString - this way you don't need to change any code outside the library!
> 
> Looking at the numbers involved, I don't think we need that much of precision too - none of them are smaller than 1, so 2-3 digits of precision should be plenty.
> 
> Thank you!

Got it @tcNickolas, I have updated the code with changes in only 3 files.

tcNickolas:
Can you please resolve conflicts in Common.qs? CI cannot run with them there

devikamehra:
> Can you please resolve conflicts in Common.qs? CI cannot run with them there

Resolved

katas/content/complex_arithmetic/Common.qs
```diff
@@ -11,15 +11,17 @@ namespace Kata.Verification {
     }
 
     function ComplexAsString(x : Complex) : String {
-        if x.Imag < 0.0 {
-            $"{x.Real} - {AbsD(x.Imag)}i"
+        let precision = 3;
+        if x::Imag < 0.0 {
```
tcNickolas, line 15: I realized what was causing the conflict - an earlier PR that updated the syntax for accessing UDT fields from :: to . (you can see it later in this file, for example, ComplexEqual and ComplexPolarEqual use .)

This change reverts it in part of the file; can you please update your change to the new syntax for consistency?

tcNickolas (approved):
Output looks good, just one suggestion on the resolved conflict.

Thank you!

katas/content/complex_arithmetic/Common.qs
devikamehra: Got it. I have updated the file now.

cesarzc (approved):

PR #1884, opened 2024-08-25
Added DoubleAsStringWithPrecision function

devikamehra:
Added DoubleAsStringWithPrecision function to Single Qubit Gate Kata

### DIFF
--- a/katas/content/single_qubit_gates/amplitude_change/Verification.qs
+++ b/katas/content/single_qubit_gates/amplitude_change/Verification.qs
@@ -14,8 +14,10 @@ namespace Kata.Verification {
             let reference = register => AmplitudeChange(alpha, register[0]);
             let isCorrect = CheckOperationsAreEqualStrict(1, solution, reference);
             if not isCorrect {
+                // In case of an error, this value defines the precision with which complex numbers should be displayed
+                let precision = 6;
                 Message("Incorrect.");
-                Message($"The solution was incorrect for the test case alpha = {alpha}.");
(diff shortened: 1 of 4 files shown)


tcNickolas (changes requested):
Please make sure to use space after commas to make the code more readable.
The numbers we're using here are not usually less than 0.1, so I don't think they warrant such a high precision, 2-3 digits should be plenty - the main goal of this change is to shorten the error messages

Thank you!

katas/content/single_qubit_gates/amplitude_change/Verification.qs
```diff
@@ -14,8 +14,10 @@ namespace Kata.Verification {
             let reference = register => AmplitudeChange(alpha, register[0]);
             let isCorrect = CheckOperationsAreEqualStrict(1, solution, reference);
             if not isCorrect {
+                // In case of an error, this value defines the precision with which complex numbers should be displayed
```
tcNickolas, line 17: Please remove these comments in all files, since they're not accurate and confusing (I definitely spent a moment trying to figure out where the complex numbers are here!) The precision here is not used to display complex numbers like it was used in Complex Arithmetic, all numbers displayed with precision here are real. The code here is so straightforward, I don't think we need comments to explain what precision is - the name of the function that uses them is self-descriptive

tcNickolas (approved):
The output of the updated code looks good! Just one suggestion to remove confusing comments.

Thank you!

katas/content/single_qubit_gates/amplitude_change/Verification.qs
devikamehra: Removed

cesarzc (approved):

Issue #1865, opened 2024-08-20
LaTeX rendering of nice numbers aren't always correct


DmitryVasilevsky:
Useful functions to render some rational, algebraic numbers and exponential form of complex number aren't always producing the expected result when used outside state rendering. Now that we have want to add a few other use cases we should fix that.
Example:
![image](https://github.com/user-attachments/assets/4a2aa1be-dbcb-4b4d-a3b3-e0c75d7a4876)

sezna:
After #1885, these numbers will rarely (never?) show up in output, so we are marking this as low priority.

DmitryVasilevsky:
Code changes in https://github.com/microsoft/qsharp/pull/1885 provide correct handling of standalone numbers (rather than coefficients).  The following can still be improved in general case:
    * i could be used directly in a fraction. Example: $\frac{1}{2}i$ could be rendered as $\frac{i}{2}$.
    * In a simple fraction sign can be moved to numerator: Example: $-\frac{1}{2}$ could be rendered as $\frac{-1}{2}$.
    * Brackets are not needed for standalone rendering. Example: $-\left(1+i\right)$ could be rendered as $-1-i$
    * No need to rationalize the denominator. Example: $\frac{\sqrt{2}}{2}$ could be rendered as $\frac{1}{\sqrt{2}}$

PR #1885, opened 2024-08-25
Add DumpOperation support in Q#

billti:
This enables Q# code to call 'DumpOperation' which will output to console/debugger as text, or Jupyter notebooks as a LaTeX matrix.

e.g. from a notebook cell 

```
%%qsharp

operation Foo(qs: Qubit[]) : Unit {
    H(qs[0]);
    CX(qs[0], qs[1]);
}

operation Main() : Unit {
    use qs = Qubit[2];
    Foo(qs);
    DumpMachine();
    Message("About to dump operation Foo");
    DumpOperation(2, Foo);
    ResetAll(qs);
}

Main()
```

Output is:

image


library/std/src/diagnostics.qs
```diff
@@ -69,6 +69,44 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    /// $\left(\begin{matrix} 0.0 & 0.707 \\\\ 0.707 & 0.0\end{matrix}\right)$:
+    ///
+    /// ```qsharp
+    /// operation DumpH() : Unit {
+    ///     DumpOperation(1, qs => H(qs[0]));
+    /// }
+    /// ```
+    /// Calling this operation has no observable effect from within Q#.
+    /// Note that if `DumpOperation` is called when there are other qubits allocated,
+    /// the matrix displayed may reflect any global phase that has accumulated on the other qubits.
```
swernli, line 94:
```suggestion
    /// the matrix displayed may reflect any global phase that has accumulated from operations
    /// on those other qubits.
```

playground/src/results.tsx
```diff
@@ -53,6 +53,17 @@ function resultIsSame(a: ShotResult, b: ShotResult): boolean {
   return true;
 }
 
+function MatrixLatex(props: { latex: string }) {
+  return (
+    <>
+      <div style="margin: 12px 4px; font-size: 1em; font-style: italic;">
+        DumpOperation
```
swernli, line 60: Does this need the title here? I know this is just for the playground, but I think it would be fine to remove this, as messages and dump machine/register output don't have a title.

wasm/src/tests.rs
```diff
@@ -129,6 +129,32 @@
...
+        operation Bell(q: Qubit[]) : Unit {
+            H(q[0]);
+            CX(q[0], q[1]);
+        }
+    }";
+    let expr = "Test.Main()";
+    let result = run_internal(
+        SourceMap::new([("test.qs".into(), code.into())], Some(expr.into())),
+        |msg| {
+            // TODO: Validate the matrix output
```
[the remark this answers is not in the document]
swernli: Good bot.

pip/qsharp/_qsharp.py
```diff
@@ -169,6 +179,13 @@ def eval(source: str) -> Any:
     ipython_helper()
 
     def callback(output: Output) -> None:
+        if _in_jupyter:
+            try:
+                display(output)
+                return
+            except:
+                # If IPython is not available, fall back to printing the output
+                pass
```
minestarks, line 188: The old behavior of `qsharp.eval` was intentional. May be debatable whether it's the "ideal" behavior but definitely intentional as seen in the [sample notebook](https://github.com/microsoft/qsharp/blob/main/samples/notebooks/sample.ipynb):

image
 


If we feel like we _have_ to change this behavior, I think we should take more care to refactor this code so that it's not duplicated with the code in `_ipython.py` (which I see, also got a new bugfix we'd want to include here): https://github.com/microsoft/qsharp/blob/78ad29efd6b821cf8d1f98e44bbdd50a313dd91f/pip/qsharp/_ipython.py#L19-L29

pip/src/interpreter.rs
```diff
@@ -336,24 +337,25 @@ impl Output {
         self.__repr__()
     }
 
-    fn _repr_html_(&self) -> String {
+    fn _repr_markdown_(&self) -> Option<String> {
```
minestarks, line 340: Why Markdown in favor of HTML? Does it have implications in how the cell output format is selected VS Code? I seem to recall there was some pretty funky UX going on there. Am I correct in assuming that you're trying to simplify the UX here by having just one "rich" output format instead of latex/html etc?

minestarks: Also, do you know if JupyterLab renders these outputs correctly?

compiler/qsc_eval/src/state.rs
```diff
@@ -381,6 +381,106 @@ pub fn get_latex(state: &Vec<(BigUint, Complex64)>, qubit_count: usize) -> Optio
...
 }
 
+fn is_close_enough(val: &Complex64, target: &Complex64) -> bool {
+    (val.re - target.re).abs() < 1e-9 && (val.im - target.im).abs() < 1e-9
+}
+
+// Quick and dirty matching for the most common matrix terms we care about rendering
+// LaTeX for, e.g.  1/sqrt(2), -i/sqrt(2), etc.
+// Anything not in this list gets a floating point form.
+fn get_latex_for_simple_term(val: &Complex64) -> Option<String> {
```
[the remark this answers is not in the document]
billti: I believe I tried that first, and hit https://github.com/microsoft/qsharp/issues/1865 and other issues. I figured it was faster (both implementation and runtime) to hardcode the most common values and their representation, but could add a TODO to look into that again once other issues are fixed.

playground/src/results.tsx
```diff
@@ -240,13 +251,15 @@ export function ResultsTab(props: {
...
-              ) : (
+              ) : evt.type === "DumpMachine" ? (
                 <div>
                   <StateTable
                     dump={evt.state}
                     latexDump={evt.stateLatex}
                   ></StateTable>
                 </div>
+              ) : (
+                <MatrixLatex latex={evt.matrixLatex} />
```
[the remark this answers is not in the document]
billti: No, you're right. I already called this out to Stefan. TypeScript will yell at you in a number of places for not doing exhaustive pattern matching once the new event is in also. We'll need to coordinate with QCOM before they take any drop with this (or maybe help fix it ourselves if easier).

pip/src/interpreter.rs
billti: You can embed LaTeX directly in Markdown and it renders correctly - not so in HTML. Being that the other HTML outputs also render OK as Markdown (as inline HTML), this enabled the 'pretty printing' of the equation as well as the DumpMachine table.

swernli: Yeah, what Bill said. HTML won't render embedded LaTeX, but Markdown renders both HTML and LaTeX, making it the one ring to rule them all. Seemed like a good reason to drop the other representations and align behind Markdown as a single rich output format.

I did get the chance to test all our output variants (messages, state dumps from `DumpMachine` and`DumpRegister`, and matrices from `DumpOperation`) in VS Code notebooks, Jupyter in the browser, and JupyterLab in the browser, and confirmed they are all rendered as expected across the board. Bill identified some quirks of the browser formatting that we worked around to make sure it looked good and didn't do anything odd.

library/std/src/diagnostics.qs
```diff
@@ -69,6 +69,44 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    operation DumpOperation(nQubits : Int, op : Qubit[] => Unit) : Unit {
```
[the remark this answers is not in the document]
swernli: I'd prefer to not make it generic on the return, as operations that return measurement results tend to not be representable as a matrix. A quantum operation is "unitary" when it can be represented by a unitary matrix, which is not the same thing as returning `Unit` but we sometimes conflate. This is because a unitary operation is reversible, so it has a defined (or generated) adjoint, and only operations that return `Unit` can have an adjoint. In fact, if you look at the classic QDK definition of `DumpOperation` [over here](https://github.com/microsoft/QuantumLibraries/blob/f2f5e380225367ee4eb995d922067025dc77f0a0/Standard/src/Diagnostics/Dump.qs#L71) it actually required in the signature that the provided operation was adjointable even though it didn't use the adjoint. Thinking about this in answer to your question, I realized my intent to relax the restriction of adjointable may have worked against its subtle intent here: to ensure that the operation is something that can be represented as a unitary. So I'm starting to think that rather than relaxing further with a generic return, this should actually be as strict as the original and require `Qubit[] => Unit is Adj` in the type.

library/std/src/diagnostics.qs
```diff
@@ -69,6 +69,44 @@ namespace Microsoft.Quantum.Diagnostics {
...
+        use (targets, extra) = (Qubit[nQubits], Qubit[nQubits]);
```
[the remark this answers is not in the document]
swernli: This is a tricky one. My first instinct was to say that this code shouldn't be called if the user is doing QIR codegen, but there isn't anything in here that shows them directly that it should be commented out. But I think you are on to something with using `SimulatableIntrinsic` to avoid this. Let me try a few things and see if the behavior matches expectation.

In general though, `DumpOperation` can have side effects even during simulation that make it undesirable to include in code that is being used for codegen, since it won't always behave the way the user expects.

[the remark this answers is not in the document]
swernli: Ok, your suggestion did just the trick! I made `DumpOperation` and simulatable intrinsic and added it to the list of recognized no-op callables in partial evaluation. That is enough to preserve its behavior in simulation but cause it to be fully skipped (like `DumpMachine` and `DumpRegister`) during QIR codegen.

library/std/src/diagnostics.qs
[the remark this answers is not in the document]
swernli: I did end up making this more restrictive, so now it requires an adjointable operation. I'll need to circle back and make the same change to the Python version for consistency.

compiler/qsc_eval/src/output.rs
```diff
@@ -47,6 +52,15 @@ impl<'a> Receiver for GenericReceiver<'a> {
         Ok(())
     }
 
+    fn matrix(&mut self, matrix: Vec<Vec<Complex64>>) -> Result<(), Error> {
+        writeln!(self.writer, "MATRIX:").map_err(|_| Error)?;
```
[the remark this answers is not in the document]
DmitryVasilevsky: Yes, but we are printing the representation, right? A full name would be something like "operation represented by a matrix in computational basis". We could probably write something like "operation matrix" (similar to "state vector").

compiler/qsc_eval/src/state.rs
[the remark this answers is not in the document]
billti: Dmitry fixed the other impl. so this is used for less values. However it is still needed for some terms that we want to represent differently (esp. over sqrt(2) 'canonical' values)

compiler/qsc_eval/src/output.rs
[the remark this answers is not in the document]
billti: We only use it for DumpOperation now, but I guess theoretically if we ever wanted to output say a density matrix from the noisy simulator or something, we might reuse the same event - which is just basically "show this matrix".

pip/qsharp/_qsharp.py
billti: > The old behavior of qsharp.eval was intentional.

Why was it intentional that they appear differently. What's the rationale for that? I had assumed that was a mistake/oversight.

compiler/qsc_eval/src/state.rs
```diff
@@ -381,13 +385,94 @@ pub fn get_latex(state: &Vec<(BigUint, Complex64)>, qubit_count: usize) -> Optio
...
+        return Some("-\\frac{1}{2} + \\frac{i}{2}".to_string());
+    }
+    if is_close_enough(val, &Complex64::new(0.5, -0.5)) {
+        return Some("\\frac{1}{2} - \\frac{i}{2}".to_string());
+    }
+    None
+}
+
+#[must_use]
+pub fn get_matrix_latex(matrix: &Vec<Vec<Complex64>>) -> String {
```
DmitryVasilevsky, line 430: Consider renaming existing function `get_latex` to `get_state_latex` as there's `get_matrix_latex` now.
Also maybe use naming similar to other functions such as `get_latex_for_,,,` (totally up to you).

DmitryVasilevsky (approved):

pip/qsharp/_qsharp.py
billti: Ping @minestarks . Still not clear on why it was intentional, and if there are benefits to that behavior. (i.e. if should revert changes around it)

library/std/src/Std/Diagnostics.qs
```diff
@@ -64,6 +64,46 @@ function DumpRegister(register : Qubit[]) : Unit {
     body intrinsic;
 }
 
+/// # Summary
+/// Given an operation, dumps the matrix representation of the operation actiong on the given
```
swernli, line 68:
```suggestion
/// Given an operation, dumps the matrix representation of the operation action on the given
```

swernli (approved):
LGTM, though there is still the feedback from Mine to resolve.

minestarks (approved):

pip/qsharp/_qsharp.py
minestarks: I don't think there was a compelling reason for the old behavior, I just wanted it to behave more "console-y" when `qsharp.eval` was used as opposed to the ipython magic `%%qsharp`.

I'm ok changing it but please note my comment on code duplication. The same `display` logic is implemented in `_ipython.py` and I think we should leverage it.

billti: The code in _ipython.py is just for `@register_cell_magic`, in which case if you're in a `%%qsharp` cell we __known___ you're in a notebook context, right? So I don't think that needs the display logic from _qsharp.py, which is specifically to detect if in a text based UI (e.g. IPython in the console) versus a GUI.

Or am I missing something?

billti: Will address later any changes needed on the above question.

PR #1891, opened 2024-08-27
Added DoubleAsStringWithPrecision function - Linear Algebra

### LINKED ISSUE #1788: Use M.Q.Convert.DoubleAsStringWithPrecision to trim Doubles printed in the katas
Follow-up to #1406: now that the function M.Q.Convert.DoubleAsStringWithPrecision is available, we can use it in the katas to shorten the long floating-point numbers.

Here are a few examples of the katas that print long doubles:

* Complex Arithmetic
image
* Single-Qubit Gates kata, task "Prepare Arbitrary State"
* Preparing Quantum States, task "Unequal Superposition"

There are probably others; the typical scenario in which this occurs is printing the Double parameters of the test or returns of the solution for an incorrect solution. We are unlikely to need more than 2-3 digits of precision in these cases.

devikamehra:
Added DoubleAsStringWithPrecision function to Linear Algebra

This is to resolve #1788

### DIFF
--- a/katas/content/linear_algebra/Common.qs
+++ b/katas/content/linear_algebra/Common.qs
@@ -1,7 +1,16 @@
 namespace Kata.Verification {
     open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Convert;
 
-    function ArraysEqualD(actual : Double[][], expected : Double[][]) : Bool {
+    function ComplexAsString(x : Complex, precision: Int) : String {        
+        if x::Imag < 0.0 {
+            $"{DoubleAsStringWithPrecision(x::Real,precision)} - {DoubleAsStringWithPrecision(AbsD(x::Imag),precision)}i"
+        } else {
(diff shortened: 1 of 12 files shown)


devikamehra:
> I'm not sure we need to improve the display of Doubles in this particular kata... The tasks deal with exactly one test case each, and the matrices they work with are carefully hand-picked to have very nice floating-point values in them, with at most one decimal point. I don't think the learner is likely to try and return values with a lot of decimal points either, since most operations involve only simple things like swapping numbers around and addition/subtraction, not a lot of division going on here. This means that enforcing precision 3 would actually make the displayed numbers longer with just extra trailing zeros

So, should I reduce the precision to 1 or remove this altogether?

tcNickolas:
Precision 1 is the default display of doubles with 0 or 1 decimal points, so I don't think we need to change anything in this kata

devikamehra:
Not required

PR #1899, opened 2024-09-03
Basic interop with Qiskit

### LINKED ISSUE #1074: QISKIT support for RE
Azure-quantum service supports "estimate_type": "frontier" but QISKIT samples use the old python-qdk wrappers to call it which does not support "estimate_type".
Consider either support QISKIT in the modern QDK or update the old one to pass the estimate_type.

idavis:
This PR adds the ability to resource estimate, run, and generate QIR from Qiskit `QuantumCircuit`s.

There are three areas which we can break up the review into:
- compiler/qsc_qasm3
- Python user facing API and notebooks. Example usage can be seen in:
  - samples/python_interop/qiskit/interop.ipynb
  - samples/estimation/estimation-qiskit.ipynb
- /compiler changes (excluding qsc_qasm3)

The QASM parser doesn't have usable semantic validation, so we have to do a lot of our own validation as we compile the QASM generated from the `QuantumCircuit`s.

Closes #1074


compiler/qsc_qasm3/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[dependencies]
+bitflags = { workspace = true }
+num-bigint = { workspace = true }
+miette = { workspace = true }
+qsc = { path = "../qsc" }
+rustc-hash = { workspace = true }
+thiserror = { workspace = true }
+
+oq3_source_file = "0.6.0"
```
billti, line 18: Should these be moved to 'workspace' dependencies?

compiler/qsc_qasm3/src/tests/assignment/slices.rs
```diff
@@ -0,0 +1,2 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
orpuente-MS, line 2: This file seems to be empty. Do we need it?

compiler/qsc_qasm3/src/io.rs
```diff
@@ -0,0 +1,89 @@
...
+        P: AsRef<Path>;
+}
+
+/// A source resolver that resolves include files from the file system.
+#[cfg(feature = "fs")]
+pub mod fs {
+    use super::SourceResolver;
+
+    #[derive(Default)]
+    pub struct FsSourceResolver;
```
orpuente-MS, line 49: Did a search for this struct on the repo and it doesn't seem to be used anywhere. Is this dead code?

compiler/qsc_qasm3/src/tests/assignment/slices.rs
idavis: An example of me thinking I need to test something but then couldn't get qiskit to generate it. Deleted.

compiler/qsc_qasm3/src/io.rs
idavis: It is an anachronism. Good catch.

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4364 @@
...
+        let expr = paren_expr.expr()?;
+        let texpr = self.compile_expr(&expr)?;
+        let pexpr = ast_builder::wrap_expr_in_parens(texpr.expr, span);
+        Some(QasmTypedExpr {
+            ty: texpr.ty.clone(),
+            expr: pexpr,
+        })
+    }
+
+    #[allow(clippy::too_many_lines)]
```
orpuente-MS, line 2192: Do we need this `allow` here? This function is about 50 lines long, and the `too_many_lines` lint triggers when you have over 100 lines of code.

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4364 @@
...
+                self.push_unimplemented_error_message(
+                    "negated literal expressions",
+                    literal.syntax(),
+                );
+                None
+            }
+        }
+    }
+
+    #[allow(clippy::too_many_lines)]
```
orpuente-MS, line 2243: Same here, this function is only 55 lines long and the lint triggers when you have more than 100 lines.

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4363 @@
...
+            self.push_semantic_error(kind);
+            None
+        }
+    }
+
+    #[allow(clippy::too_many_lines)]
+    fn compile_gate_decl(&mut self, gate: &oq3_syntax::ast::Gate) -> Option<ast::Stmt> {
+        let name = gate.name()?;
+
+        //let angle_ty = Type::Angle(None, IsConst::True);
```
orpuente-MS, line 2454: Old Code?

```suggestion
```

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4363 @@
...
+                    node,
+                );
+                None
+            }
+            Type::Gate(_, _) => unimplemented!("Gate default values"),
+            Type::Range => unimplemented!("Range default values"),
+            Type::Set => unimplemented!("Set default values"),
+            Type::Void => unimplemented!("Void default values"),
+            Type::ToDo => unimplemented!("ToDo default values"),
+            Type::Undefined => unimplemented!("Undefined default values"),
```
orpuente-MS, line 3089: Are these reachable from Python? Will this crash the program if the user happens to try these from a Python Notebook?

idavis: Those should be bugs and effectively unreachable. We can convert them to semantic errors, with the comment that it is unlikely that the cases can ever be hit.

compiler/qsc_qasm3/src/compile.rs
idavis: It isn't old code, it is a note to future self that the type really should be the commented out line, but `Angle` isn't supported yet. I'll update the comment.

pip/tests-integration/interop_qiskit/test_circuits/random.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import pytest
+from qsharp import TargetProfile
+
+from .. import QISKIT_AVAILABLE
+
+if QISKIT_AVAILABLE:
+    from qiskit.circuit.random import random_circuit
```
minestarks, line 10: Is this really what it sounds like? Generates a random circuit? I'm confused by why this would exist in Qiskit.

More relevant to this PR, I don't think it's a great idea to have nondeterministic tests in the build. It seems that you're already explicitly covering edge cases pretty well?

pip/tests-integration/interop_qiskit/test_circuits/test_circuits.py
```diff
@@ -0,0 +1,683 @@
...
+import numpy as np
+from typing import Tuple, List
+
+from interop_qiskit import QISKIT_AVAILABLE, SKIP_REASON
+
+if QISKIT_AVAILABLE:
+    from qiskit.circuit import QuantumCircuit
+
+
+def get_parameterized_circuit(num_qubits: int) -> "QuantumCircuit":
```
minestarks, line 15: This function doesn't seem to be called explicitly anywhere. Am I missing something magical?

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4386 @@
...
+            (Type::Float(..), Type::Float(..)) => {
+                if let LiteralKind::FloatNumber(value) = literal.kind() {
+                    let value = value.value().expect("FloatNumber must have a value");
+                    let expr = build_lit_double_expr(value, span);
+                    Some(QasmTypedExpr {
+                        ty: ty.clone(),
+                        expr,
+                    })
+                } else {
+                    panic!("Literal must be an FloatNumber")
```
orpuente-MS, line 3487: Typo?

```suggestion
                    panic!("Literal must be a FloatNumber")
```

pip/src/interpreter.rs
```diff
@@ -27,11 +31,33 @@ use qsc::{
...
+fn verify_classes_are_sendable() {
+    fn is_send<T: Send>() {}
+    is_send::<TargetProfile>();
+    is_send::<Result>();
+    is_send::<Pauli>();
+    is_send::<Output>();
+    is_send::<StateDumpData>();
+    is_send::<Circuit>();
+
+    // QSharpError, and QasmError are not Send, but we don't raise
```
minestarks, line 52: But we do raise `QSharpError` in this module. What am I missing?

pip/src/interpreter.rs
```diff
@@ -27,11 +31,33 @@ use qsc::{
     target::Profile,
     LanguageFeatures, PackageType, SourceMap,
 };
+use qsc_qasm3::{OutputSemantics, ProgramType};
 use resource_estimator::{self as re, estimate_expr};
 use std::{cell::RefCell, fmt::Write, path::PathBuf, rc::Rc};
 
+/// If the classes are not Send, the Python interpreter
+/// will not be able to use them in a separate thread.
```
minestarks, line 39: Where did this threading requirement come from?

minestarks (commented):
Publishing the first round of comments to give you something to look at... still reviewing.

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4327 @@
...
+                ..Default::default()
+            },
+            Some(signature),
+        )
+    }
+
+    /// Turns the compiled statements into package of top level nodes
+    fn build_fragments(&mut self) -> Package {
+        let nodes = self
+            .stmts.drain(..)
```
billti, line 199: This line failed clippy for me. (Need to put `.drain(..)` on a new line.

pip/src/interpreter.rs
idavis: We return `QasmError::new_err` or `QSharpError::new_err` which actually returns a `PyErr` that is `Send` and the args passed into the `new_err` call must also impl `Send` because of this.

pip/src/interpreter.rs
idavis: Working with Qiskit backends. For anything that would be run in the interactive under the shared interpreter, everything would be sync.

build.py
```diff
@@ -514,6 +514,8 @@ def run_ci_historic_benchmark():
         "ipykernel",
         "nbconvert",
         "pandas",
+        "qsharp-widgets",
```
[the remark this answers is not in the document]
idavis: I'm on the fence and wrestled with it myself. When running these tests, the widgets may not have been built and thus won't be there to install.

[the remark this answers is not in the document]
minestarks: FWIW, there is a precedence already set by `./build.py --vscode --integration-tests` . This command will run the integration tests but won't build all the prerequisites. You're on the hook for building the product before you run the integration tests. I'd be ok with continuing this pattern for the Python integration tests.

samples/python_interop/qiskit.ipynb
```diff
@@ -0,0 +1,482 @@
...
+    {
+      "cell_type": "markdown",
+      "id": "b06b7857",
+      "metadata": {},
+      "source": [
+        "## Errors\n",
+        "#### Unsupported language features, `QasmError`, and `QSharpError`\n",
+        "The modern QDK's interop with Qiskit is based on Qiskit's OpenQASM 3 support. Qiskit supports a subset of OpenQASM 3 features which may cause issues during conversion. If the Qiskit OpenQASM `Exporter` or OpenQASM parser don't support the feauture yet, a `QasmError` is raised prior to conversion. If there is a QASM parsing failure, this is likely an issue with the Qiskit libraries parsing or export functionality.\n",
+        "\n",
+        "Any failure to transform the OpenQASM into Q#'s internal representation will throw a `QasmError` to distinguish the scenarios. This is most likely due to a semantically invalid OpenQASM program as input.\n",
```
billti, line 293: Is this wording right? It states _"will throw a `QasmError` to distinguish the scenarios"_, but QasmError is the same error thrown in the preceding paragraph/problem, so it's not helping 'distinguish'.

samples/python_interop/qiskit.ipynb
```diff
@@ -0,0 +1,482 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "3170569e",
+      "metadata": {},
+      "source": [
+        "When generating `QIR`, all output registers must be read into before generating QIR. Failure to do so results in a `QSharpError`.\n",
+        "\n",
+        "In this first example, we declare two output bits, but only measure into one. This causes an error because result values can only be a side effect of measurement, and cannot be used like classical variables when compiling for hardware."
```
billti, line 438: > In this first example

There is only one example that follows. More to add here?

pip/tests-integration/interop_qiskit/test_circuits/random.py
idavis: They had value when they were written. Less so now, removed them.

samples/python_interop/qiskit.ipynb
idavis: Looking, I did a lot of reorganization and might have messed up that part.

samples/python_interop/qiskit.ipynb
idavis: Same as above, must have missed it when I was reorganizing and reordering. Thanks.

minestarks (commented):
Reviewed all except `compiler/`.

compiler/qsc_qasm3/src/parse.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+/// Represents a QASM source file that has been parsed.
+#[derive(Clone, Debug)]
+pub struct QasmSource {
+    /// The path to the source file. This is used for error reporting.
+    /// This path is just a name, it does not have to exist on disk.
+    path: PathBuf,
+    /// The source code of the file.
+    source: String,
```
sezna, line 116: Would it be valuable to make this an `Arc<str>` to mimic our own `Source` type? Or even reuse that type?

compiler/qsc_qasm3/src/parse.rs
```diff
@@ -0,0 +1,265 @@
...
+        QasmSource {
+            source: source.as_ref().to_owned(),
+            path: file_path.as_ref().to_owned(),
+            ast,
+            included,
+        }
+    }
+}
+
+impl QasmSourceTrait for QasmSource {
```
sezna, line 167: Based on my cmd+F-ing, this is only implemented for one type. Could this trait just be methods on `QasmSource`?

pip/qsharp/_fs.py
```diff
@@ -1,37 +1,100 @@
...
+        path (str): The relative path.
+
+    Returns:
+        str: The resolved path.
+    """
     return os.path.normpath(join(base, path))
 
 
 def exists(path) -> bool:
+    """
```
sezna, line 77: thanks for the docstrings 👍🏻

pip/src/interop.rs
```diff
@@ -0,0 +1,657 @@
...
+    QasmCompileUnit, QubitSemantics,
+};
+
+use crate::fs::file_system;
+use crate::interpreter::{
+    format_error, format_errors, OptionalCallbackReceiver, QSharpError, QasmError, TargetProfile,
+    ValueWrapper,
+};
+
+use resource_estimator::{self as re};
```
sezna, line 31:
```suggestion
use resource_estimator as re;
```

pip/src/interop.rs
```diff
@@ -0,0 +1,657 @@
...
+    let (package, source_map, signature) = compile_qasm_enriching_errors(
+        source,
+        &name,
+        &resolver,
+        program_type,
+        OutputSemantics::Qiskit,
+        false,
+    )?;
+
+    let package_type = PackageType::Lib;
```
sezna, line 248: am I correct in understanding that all `qasm3` to `qir` compilations will be `PackageType::Lib`, and `LanguageFeatures::default()`? Do we want additional configurability here?

pip/src/interop.rs
```diff
@@ -0,0 +1,657 @@
...
+        source_package_id,
+        capabilities,
+        language_features,
+        &dependencies,
+    )
+}
+
+/// Sanitizes the name to ensure it is a valid identifier according
+/// to the Q# specification. If the name is empty, returns "circuit".
+pub(crate) fn sanitize_name<S: AsRef<str>>(name: S) -> String {
```
sezna, line 561: should we reuse this logic when we transform file names into namespace names for implicit namespaces?

pip/src/interop.rs
idavis: For this code path we don't know the entry expr until after compilation so we can't be exe. When running in interactive, we'll just be adding to the run state, so we will use `interperter.qir(...)` path.

compiler/qsc_qasm3/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+oq3_parser = { workspace = true }
+oq3_lexer = { workspace = true }
+oq3_semantics = { workspace = true }
+
+[dev-dependencies]
+difference = { workspace = true }
+expect-test = { workspace = true }
+indoc = { workspace = true }
+miette = { workspace = true, features = ["fancy"] }
+qsc_qasm3 = { path = ".", features = ["fs"] }
```
swernli, line 28: Is the self-dev dependency here just to make sure the `fs` feature is enabled? This could be worth a comment if so, so that folks know why this shouldn't be removed.

pip/qsharp/_qsharp.py
```diff
@@ -326,9 +363,68 @@ def estimate(
...
+                from qsharp.interop.qiskit.backends import ResourceEstimatorBackend
+
+                backend = ResourceEstimatorBackend(
+                    target_profile=TargetProfile.Unrestricted
+                )
+                job = backend.run(estimate_input, params=params, **options)
+                return job.result()
+            except ImportError as ex:
+                message = "Unsupported input type."
+                raise ValueError(message) from ex
```
swernli, line 424: would it make sense to include something about failing to import qiskit here, so users know they've hit a recognized path that we can't support? It seems a little odd that the general error below includes more information than this one. It's possible the second line here that has `from ex` includes the detailed import error, in which case sorry for the noise!

compiler/qsc/src/codegen.rs
```diff
@@ -4,80 +4,155 @@
...
 
-use qsc_codegen::qir::fir_to_qir;
-use qsc_data_structures::{language_features::LanguageFeatures, target::TargetCapabilityFlags};
-use qsc_frontend::{
-    compile::{Dependencies, PackageStore, SourceMap},
-    error::WithSource,
-};
-use qsc_partial_eval::ProgramEntry;
-use qsc_passes::{PackageType, PassContext};
+pub mod qsharp {
```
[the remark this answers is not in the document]
idavis: This was the pattern set for qir module right below it. So I was following the pattern. We can change, but that is why.

samples/python_interop/qiskit.ipynb
```diff
@@ -0,0 +1,480 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "2b838c23",
+      "metadata": {},
+      "source": [
+        "The modern QDK provides interoperability with Qiskit circuits built upon the core Q# compiler infrastructure.\n",
+        "\n",
+        "This core enable integration and local resource estimation without relying on external tools. Users are able to estimate resources for their Qiskit circuits locally (see the [resource estimation with Qiskit sample notebook](../../estimation/estimation-qiskit.ipynb)), leveraging the Q# compiler's capabilities for analysis, transformation, code generation, and simulation. This also enables the generation of QIR from Qiskit circuits leveraging the [modern QDKs advanced code generation capabilities](https://devblogs.microsoft.com/qsharp/integrated-hybrid-support-in-the-azure-quantum-development-kit/).\n",
```
swernli, line 18: missing "s" in "enables":
```suggestion
        "This core enables integration and local resource estimation without relying on external tools. Users are able to estimate resources for their Qiskit circuits locally (see the [resource estimation with Qiskit sample notebook](../../estimation/estimation-qiskit.ipynb)), leveraging the Q# compiler's capabilities for analysis, transformation, code generation, and simulation. This also enables the generation of QIR from Qiskit circuits leveraging the [modern QDKs advanced code generation capabilities](https://devblogs.microsoft.com/qsharp/integrated-hybrid-support-in-the-azure-quantum-development-kit/).\n",
```

samples/python_interop/qiskit.ipynb
```diff
@@ -0,0 +1,480 @@
...
+        "\n",
+        "circuit.draw(output=\"text\")"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "922f8420",
+      "metadata": {},
+      "source": [
+        "With the circuit created, we can run the circuit with Q#'s backend. By default, it will use the `Unrestricted` profile meaning anything it allowed for simulation."
```
swernli, line 64:
```suggestion
        "With the circuit created, we can run the circuit with Q#'s backend. By default, it will use the `Unrestricted` profile meaning anything is allowed for simulation."
```

compiler/qsc/src/interpret.rs
```diff
@@ -560,14 +580,21 @@ impl Interpreter {
...
+
     /// Runs the given entry expression on the given simulator with a new instance of the environment
     /// but using the current compilation.
     pub fn run_with_sim(
         &mut self,
         sim: &mut impl Backend<ResultType = impl Into<val::Result>>,
         receiver: &mut impl Receiver,
         expr: Option<&str>,
-    ) -> std::result::Result<InterpretResult, Vec<Error>> {
+    ) -> InterpretResult {
```
swernli, line 597: Nice catch! This extra layer of indirection was leftover from several refactors ago, glad you finally eliminated it.

compiler/qsc/src/codegen.rs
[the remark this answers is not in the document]
swernli: I think what Alex means is that, looking at the diff, `get_qir` used to be at the file level module `codegen`, and it appears that both `mod qsharp` and `mod qir` were introduced in this change. `mod qsharp` is all just re-exports, which we do elsewhere with blocks, but `mod qir` includes actual implementation, which we usually split out into a file. I think splitting the difference would be keeping `mod qsharp` in this file for the re-exports and then having a `pub mod qir;` with that content moved to a separate file.

compiler/qsc_qasm3/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Semantics
+
+The two languages have many differences, and insofar as possible, the compilation preserves the source language semantics.
+
+- OpenQASM and Q# have different qubit and initialization semantics.
+  - Q# assumes that qubits are in the |0⟩ state when they are allocated.
+  - OpenQASM does not make this assumption and qubits are in an undefined state.
+- Q# requires that qubits are reset to the |0⟩ at the end of excution.
+- OpenQASM does not require this.
```
swernli, line 25: nitpick: these feel related in the same way the two above are, so it might make sense to put them together. Also, the restriction on qubit reset is at release, not necessarily at end of execution, but in this context where OpenQASM always allocates qubits at the global scope the two happen to be the same, so not sure if that is a distinction worth highlighting here. As a possible suggestion:

```suggestion
- OpenQASM and Q# have different qubit management semantics.
  - Q# assumes that qubits are in the |0⟩ state when they are allocated.
  - OpenQASM does not make this assumption and qubits start in an undefined state.
  - Q# requires that qubits are reset to the |0⟩ state when released.
  - OpenQASM does not require qubits to be in a specific state at the end of execution.
```

compiler/qsc_qasm3/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+QIR specific semantic constraints:
+
+- OpenQASM ouput registers are declared with a fixed size and not all of the indexes may be populated with measurements. In QIR, `Result`s can only ever be acquired through measurement. So if all entries in an output register aren't measured into, a code generation error will occur.
+
+Semantic details
+
+- Gates are implemented as lambda expressions capturing const variables from the global scope.
+  - There is an exception when using `@SimulatableIntrinsic`. Those are defined as full local `operation`s as they are not allowed to capture.
+  - We and change this in the future by copying `const` value decls into the `gate`/`function` scope, but this would require implementing a lot of inlining and partial evaluation which we already do in the compiler.
```
swernli, line 40:
```suggestion
  - We can change this in the future by copying `const` value decls into the `gate`/`function` scope, but this would require implementing a lot of inlining and partial evaluation which we already do in the compiler.
```

compiler/qsc_qasm3/src/lib.rs
```diff
@@ -0,0 +1,595 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#![allow(dead_code)]
```
orpuente-MS, line 4: Why are we allowing `dead_code` in the entire crate? Is this intentional or was it left behind after initial development?

idavis: I had it turned on while initially developing. I also accidentally implemented extra features which aren't turned on and may be detected as dead code, but are valuable to keep as we progress the feature.

idavis: I'll do a pass with it turned on and clean up anything that is really dead

orpuente-MS: Note that you can add a reason to the allow attribute
```rust
#[allow(dead_code, reason="this will support future qiskit features")]
```

samples/python_interop/qiskit.ipynb
```diff
@@ -0,0 +1,480 @@
...
+    {
+      "cell_type": "markdown",
+      "id": "b06b7857",
+      "metadata": {},
+      "source": [
+        "## Errors\n",
+        "#### Unsupported language features, `QasmError`, and `QSharpError`\n",
+        "The modern QDK's interop with Qiskit is based on Qiskit's OpenQASM 3 support. Qiskit supports a subset of OpenQASM 3 features which may cause issues during conversion.\n",
+        "\n",
+        "If the Qiskit OpenQASM `Exporter` or OpenQASM parser don't support the feature yet, a `QasmError` is raised prior to conversion. When an OpanQASM parsing failure occurs, this is likely an issue with the Qiskit libraries parsing and/or export functionality. Additionally, failure to transform the OpenQASM into Q#'s internal representation will throw a `QasmError`. This is most likely due to a semantically invalid OpenQASM program as input or an unsupported language feauture is being used.\n",
```
minestarks, line 293:
```suggestion
        "If the Qiskit OpenQASM `Exporter` or OpenQASM parser don't support the feature yet, a `QasmError` is raised prior to conversion. When an OpenQASM parsing failure occurs, this is likely an issue with the Qiskit libraries parsing and/or export functionality. Additionally, failure to transform the OpenQASM into Q#'s internal representation will throw a `QasmError`. This is most likely due to a semantically invalid OpenQASM program as input or an unsupported language feauture is being used.\n",
```

samples/estimation/estimation-qiskit.ipynb
minestarks, line 1: I'm wondering if you considered putting the `.estimate()` call on `QSharpBackend` instead of creating a separate `ResourceEstimatorBackend` class. It seems to me that the users would have an easier time keeping track of one class that takes care of their needs.

(Or, if they _have_ to be separate backends, then why not separate `QirBackend` and `SimulatorBackend` too? Why does RE get the special treatment?)

Related, but different note: Why does `qsharp.estimate` get an overload that takes `QuantumCircuit` , but `qsharp.run` and `qsharp.compile` don't? Shouldn't we try to keep call patterns consistent between these APIs?

compiler/qsc_qasm3/src/angle.rs
```diff
@@ -0,0 +1,435 @@
...
+}
+
+impl fmt::Display for Angle {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", (*self).try_into().unwrap_or(f64::NAN))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
swernli, line 213: nitpick: while it increases the folder/file count, it would be consistent with the patterns in the rest of the repo to pull these out into a test module in `angle/test.rs`.

compiler/qsc_qasm3/src/compile/tests.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::tests::{parse_all, print_compilation_errors, qasm_to_program_fragments};
+use miette::Report;
+
+#[test]
+fn programs_with_includes_with_includes_can_be_parsed() -> miette::Result<(), Vec<Report>> {
```
swernli, line 8: I think this may be a copy-paste thing, but this test case has the same name as the third one in parse/tests.rs, which might cause minor confusion when looking at test output (or doing a find all). Could this one be renamed to be compile specific rather than use "parsed?"

```suggestion
fn programs_with_includes_with_includes_can_be_compiled() -> miette::Result<(), Vec<Report>> {
```

compiler/qsc_qasm3/src/compile.rs
```diff
@@ -0,0 +1,4283 @@
...
+            Stmt::Cal(..) | Stmt::DefCal(..) | Stmt::DefCalGrammar(..) => {
+                self.compile_calibration_stmt(stmt)
+            }
+            Stmt::AnnotationStatement(annotation) => {
+                let message = "Annotation statements should have been handled in compile_stmts";
+                let span = span_for_syntax_node(annotation.syntax());
+                let kind = SemanticErrorKind::UnexpectedParserError(message.to_string(), span);
+                self.push_semantic_error(kind);
+                None
+            }
```
swernli, line 396: While this is friendlier than a panic, this seems like the kind of place I'd expecte to see an explicit panic. Rather than getting here because of some error in the parser, this actually happens only if the logic in `compile_stmts` changes in a way that breaks this assumption. I think it's worth considering changing this to an explicit panic.

samples/estimation/estimation-qiskit.ipynb
idavis: Resolved with latest changes.

compiler/qsc_qasm3/src/symbols.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn from(value: SymbolId) -> Self {
+        value.0 as usize
+    }
+}
+
+impl From<usize> for SymbolId {
+    fn from(value: usize) -> Self {
+        SymbolId(
+            value.try_into().unwrap_or_else(|_| {
+                panic!("Value, {value}, does not fit into {}", stringify!($id))
```
orpuente-MS, line 52: Here `stringify!($id)` will just expand to the literal "$id" since `$id` is not bound inside a macro context. Is that what you meant? In that case you can just write:
```rust
panic!("Value, {value}, does not fit into $id")
```

orpuente-MS: I guess what you really meant is
```rust
panic!("Value, {value}, does not fit into usize")
```
?

compiler/qsc_qasm3/src/angle.rs
```diff
@@ -0,0 +1,213 @@
...
+use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+
+/// A fixed-point angle type with a specified number of bits.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Angle {
+    value: u64,
+    size: u32,
+}
+
+#[allow(dead_code)]
```
sezna, line 22: what's the need for allowing dead code here?

compiler/qsc_qasm3/src/ast_builder.rs
```diff
@@ -0,0 +1,1562 @@
...
+
+pub(crate) fn build_unmanaged_qubit_alloc<S>(name: S, stmt_span: Span, name_span: Span) -> Stmt
+where
+    S: AsRef<str>,
+{
+    let alloc_ident = Ident {
+        name: Rc::from("__quantum__rt__qubit_allocate"),
+        ..Default::default()
+    };
+    let path_expr = Expr {
```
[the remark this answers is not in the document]
sezna: this looks way better

eslint.config.mjs
```diff
@@ -19,6 +19,7 @@ export default tseslint.config(
       "npm/qsharp/src/*.generated.ts",
       "jupyterlab/lib/",
       "jupyterlab/qsharp-jupyterlab/labextension/",
+      "**/.*/",
```
sezna, line 22: this is _probably_ fine, but would it be better to target specifically `.vscode-test-web`?

compiler/qsc_qasm3/src/angle.rs
orpuente-MS: Note that you can add a reason to the allow attribute,
```rust
#[allow(dead_code, reason = "this will support X future qiskit feature")]
```

sezna (approved):
approving with a few final comments -- I've tested it and read the code numerous times times, and I think it is certainly ready to go in and have any future improvements handled as future PRs.

Note that I did not review the `.py` files outside of documentation and general sensibility, as I'm not deep on modern Python idioms or API standards, so it'd be best if you treat this approval as an approval of everything that isn't `*.py`.

eslint.config.mjs
idavis: We have other 'hidden' folders that we want to ensure get filtered. Will revisit if we find a counter example.

compiler/qsc_qasm3/src/angle.rs
idavis: Lint reasons are experimental

minestarks (approved):
Focused on the Python bits, not as much on the compiler internals. Thanks!

orpuente-MS (approved):
Approved. I focused on the `compiler/` folder.

swernli (approved):
One minor nitpick, not critical before merge but up to you if you want to address it. Excited to see this get in!

PR #1900, opened 2024-09-04
Fix to RCA panic when original tuple binding is dynamic

cesarzc:
This change fixes a panic that happens during Runtime Capabilities Analysis (RCA).

The panic happens when the bound value of a tuple, in which at least one element is an array, is updated and the originally bound value is dynamic.


swernli (approved):
Thanks for fixing!

sezna (approved):

PR #1901, opened 2024-09-05
Use `fancy-no-syscall` only where necessary to preserve advanced miette rendering features

### LINKED ISSUE #1898: Fancy error reporting degraded with latest update.
The miette features need to be updated so that non-wasm targets for shipped components get the `fancy` feature. 

Before the latest dep update, the errors were rendered in the console with colors and arrows, now they are ascii art.

sezna:
Closes #1898

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "backtrace-ext"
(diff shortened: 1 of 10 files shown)


sezna:
Closing until I actually get this to work, its wasting space as a draft PR

PR #1905, opened 2024-09-08
Introduce `Relabel` API

swernli:
This change introduces an API for relabeling qubits without needing to use any quantum SWAP operations. This notion of relabeling is used in some quantum algorithms when all-to-all connectivity is assumed. The change includes support for label swapping in simulation, circuit generation, and QIR codegen. The `Relabel` operation takes qubit arrays that represent a permutation of labels without needing to manually break that permutation down into its equivalent two qubit swaps.


library/std/src/canon.qs
```diff
@@ -246,6 +247,24 @@ namespace Microsoft.Quantum.Canon {
...
+    /// ## q0
+    /// The first qubit to swap.
+    /// ## q1
+    /// The second qubit to swap.
+    ///
+    /// # Remarks
+    /// This operation is useful when you need to swap qubits in a way that does not incur any quantum operations.
+    /// Note that when compiling for execution on hardware with limited qubit connectivity, this operation
+    /// may not result in any changes to qubit adjacency and a SWAP gate may still be required.
+    operation SwapLabels(q0 : Qubit, q1 : Qubit) : Unit {
```
swernli, line 264: Some possible name ideas:

- `Relabel`
- `VirtualSwap`
- `LogicalSwap`
- `NameSwap`
- `NominalSwap`

ivanbasov (approved):
Resource estimation changes (a comment at tests) look good

library/std/src/canon.qs
swernli: Based on feedback, I adjusted the approach to expose `Relabel` as the more generic operation for handling arbitrary permutations.

library/std/src/canon.qs
```diff
@@ -246,6 +247,51 @@ namespace Microsoft.Quantum.Canon {
...
+    /// # Input
+    /// ## q0
+    /// The first qubit to swap.
+    /// ## q1
+    /// The second qubit to swap.
+    ///
+    /// # Remarks
+    /// This operation is useful when you need to swap qubits in a way that does not incur any quantum operations.
+    /// Note that when compiling for execution on hardware with limited qubit connectivity, this operation
+    /// may not result in any changes to qubit adjacency and a SWAP gate may still be required.
```
idavis, line 263:
```suggestion
    /// may not result in any changes to qubit adjacency and a `SWAP` gate may still be required.
```

library/std/src/canon.qs
```diff
@@ -246,6 +247,51 @@ namespace Microsoft.Quantum.Canon {
...
+    /// # Input
+    /// ## current
+    /// Array of qubits to be relabeled.
+    /// ## updated
+    /// Array of qubits with which to relabel the `current` array.
+    ///
+    /// # Remarks
+    /// This operation is useful when you need to relabel qubits in a way that does not incur any quantum operations.
+    /// Note that when compiling for execution on hardware with limited qubit connectivity, this operation
+    /// may not result in any changes to qubit adjacency and one or more SWAP gates may still be required.
```
idavis, line 281:
```suggestion
    /// may not result in any changes to qubit adjacency and one or more `SWAP` gates may still be required.
```

compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -249,6 +250,82 @@ fn two_qubit_rotation(
...
+    let left_set = left.iter().collect::<FxHashSet<_>>();
+    let right_set = right.iter().collect::<FxHashSet<_>>();
+    if left.len() != left_set.len() || right.len() != right_set.len() {
+        return Err(Error::QubitUniqueness(arg_span));
+    }
+    if left_set != right_set {
+        return Err(Error::RelabelingMismatch(arg_span));
+    }
+
+    let mut map = FxHashMap::default();
```
idavis, line 281: Would it make sense to initialize with `2*lhs_set.len()` elements? This would be an upper bound, just not sure which will be better for typical perf to avoid reallocations.

compiler/qsc_circuit/src/builder.rs
```diff
@@ -370,6 +373,92 @@ impl Builder {
...
+/// Note that even though qubit reset & reuse is disallowed,
+/// qubit ids are still reused for new allocations.
+/// Measurements are tracked and deferred.
+#[derive(Default)]
+struct Remapper {
+    next_meas_id: usize,
+    next_qubit_id: usize,
+    next_qubit_hardware_id: HardwareId,
+    qubit_map: IndexMap<usize, HardwareId>,
+    measurements: Vec<(HardwareId, usize)>,
```
idavis, line 392: This isn't a public API, can we create a type for this to improve clarity/intent?

compiler/qsc_circuit/src/builder.rs
```diff
@@ -370,6 +373,92 @@ impl Builder {
...
+
+    #[must_use]
+    fn get_meas_id(&mut self) -> usize {
+        let id = self.next_meas_id;
+        self.next_meas_id += 1;
+        id
+    }
+}
+
+#[derive(Copy, Clone, Default)]
```
idavis, line 459: For type clarity, should we encode `usize`s used for results and Qubits as tuple structs with single usize value? This would give us type safety, intent, and compiles away.

library/std/src/canon.qs
```diff
@@ -246,6 +247,51 @@ namespace Microsoft.Quantum.Canon {
...
+    ///
+    /// # Example
+    /// The following example demonstrates how to relabel qubits in a register:
+    /// ```qsharp
+    /// use qubits = Qubit[3];
+    /// let newOrder = [qubits[2], qubits[0], qubits[1]];
+    /// Relabel(qubits, newOrder);
+    /// ```
+    /// After this operation, any use of `qubits[0]` will refer to the qubit that was originally `qubits[2]`, and so on.
+    operation Relabel(current : Qubit[], updated : Qubit[]) : Unit {
```
idavis, line 291: What about `SwapLabel` and `SwapLabels` for single vs array relabel? I'm worried about cognitive load of having two very different names depending on the args.

compiler/qsc_circuit/src/builder.rs
```diff
@@ -370,6 +373,92 @@ impl Builder {
...
+pub struct HardwareId(pub usize);
```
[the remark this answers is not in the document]
swernli: In this case, because this code is for generating a circuit diagram and the ID corresponds to the wire the instructions will be displayed on. So it really is the hardware ID, or perhaps wire ID.

library/std/src/canon.qs
swernli: The feedback I got from folks (and pattern I saw in some research papers) was to call this permuting of identifiers "relabeling." So for permutations I think `Relabel` is the right term to connect it to the quantum concept. The issue is what to call the single qubit special case, which behaves like a software SWAP instead of a quantum SWAP. One option could be dropping the special case entirely and requiring users to write `Relabel([q0, q1], [q1, q0])` but that doesn't quite feel right to me. I like having a drop in, "free" replacement for `SWAP` I just don't know what to call it to connect to the concept of relabeling given that `Relabel` is taken by the general API.

compiler/qsc_eval/src/intrinsic.rs
swernli: Good idea! I think I can assert that the map will have exactly `left.len()` entries because if `left.len() != left_set.len()` or `left_set != right_set` we return an error above, so by this point I know exactly how many and which keys the map will have, just not whether any of them are getting skipped due to not being permuted (ie: the 1 in [0,1,2]->[2,1,0]). I'll add the reserve.

All that said, I expect the total size of this array to never get that big, since it scales based on the number of qubits being used.

compiler/qsc_circuit/src/builder.rs
swernli: changing to an explicit type made me realize the list of tuples was only ever being used as a count of measurements per qubit (used to be more when this struct was part of QIR codegen, but not it's just for circuits). Simplified to an `IndexMap` of hardware id to count of measurements.

compiler/qsc_circuit/src/builder.rs
swernli: It turns out we don't need it for measurement results, so now it's just `usize` for qubit ids and `HardwareId` for ids that correspond to the circuit wires.

compiler/qsc_circuit/src/builder.rs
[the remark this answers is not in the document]
idavis: I think wire id makes more sense for the circuit.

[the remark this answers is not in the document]
swernli: Sounds good to me, will change it to that!

idavis (approved):

PR #1907, opened 2024-09-09
Add more items to RE system API

msoeken:
This enables to interface with existing implementations for QEC codes and T factories in custom RE projects using the extensibility API.

### DIFF
--- a/resource_estimator/src/system/optimization/tfactory_exhaustive.rs
+++ b/resource_estimator/src/system/optimization/tfactory_exhaustive.rs
@@ -5,6 +5,7 @@ use std::borrow::Cow;
 use std::fmt::Display;
 use std::rc::Rc;
 
+use crate::system::constants::MAX_DISTILLATION_ROUNDS;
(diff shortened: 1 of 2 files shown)


resource_estimator/src/system/optimization/tfactory_exhaustive.rs
```diff
@@ -360,6 +361,19 @@ impl TFactoryBuilder {
     }
 }
 
+impl Default for TFactoryBuilder {
```
ivanbasov, line 364: Should the Default be utilized somewhere in this PR? Or is a follow up PR expected?

ivanbasov (approved):

resource_estimator/src/system/optimization/tfactory_exhaustive.rs
msoeken: There is a follow up PR planned, in which this is used in a user example to construct a default T factory builder.

swernli (approved):

PR #1908, opened 2024-09-09
Playground: Fix error squiggle updates getting dropped

minestarks:
In some cases where the user is typing fast, diagnostic updates (a.k.a. error squiggles) will get dropped. This can happen when the current document version is greater than the version that the diagnostics event pertains to.

I'm removing the version check since it was overkill anyway. As long as the diagnostics updates are received in order, the state should remain consistent.

Repro for future reference:

squiggle-dropped

Interaction between the editor and the language service. Correct behavior:

- user types `Foo` in the editor, document version is updated to 1
- `updateDocument` command with contents `Foo` version=1
- `diagnostics` event with a type error version=1
- user types `(` in the editor, document version is updated to 2
- `updateDocument` command with contents `Foo()`  version=2
- **`diagnostics` event that clears the errors in the file version=2**
- **user types `)` in the editor, document version is updated to 3**
- `updateDocument` command with contents `Foo()` (no change) version=3
- no diagnostics event is raised since there was no change in the document

Now consider the case where the user types so fast that the events above that are in bold get flipped:

- user types `Foo` in the editor, document version is updated to 1
- `updateDocument` command with contents `Foo`   version=1
- `diagnostics` event with a type error version=1
- user types `(` in the editor, document version is updated to 2
- `updateDocument` command with contents `Foo()`  version=2
- **user types `)` in the editor, document version is updated to 3**
- **`diagnostics` event that clears the errors in the file version=2**
  - **this event gets dropped since the version doesn't match the current document version**
- `updateDocument` command with contents `Foo()` (no change) version=3
- no diagnostics event is raised since there was no change in the document

### DIFF
--- a/playground/src/editor.tsx
+++ b/playground/src/editor.tsx
@@ -110,14 +110,9 @@ export function Editor(props: {
   >([]);
   const [hasCheckErrors, setHasCheckErrors] = useState(false);
 
-  function markErrors(version?: number) {
+  function markErrors() {
     const model = editor.current?.getModel();
     if (!model) return;
-
-    if (version != null && version !== model.getVersionId()) {
-      // Diagnostics event received for an outdated model
-      return;
-    }
 
     const errs = [
       ...errMarks.current.checkDiags,
@@ -260,7 +255,11 @@ export function Editor(props: {
     // will be invalid as it captures the *original* props.languageService
     // and not the updated one. Not a problem currently since the language
     // service is never updated, but not correct either.
-    srcModel.onDidChangeContent(async () => {
+    srcModel.onDidChangeContent(async (e) => {
+      e.changes.forEach((change) => {
+        log.debug("changes for version %d are: %o", e.versionId, change);
+      });
+
       // Reset the shot errors whenever the document changes.
       // The markers will be refreshed by the onDiagnostics callback
       // when the language service finishes checking the document.
@@ -304,7 +303,7 @@ export function Editor(props: {
     function onDiagnostics(evt: LanguageServiceEvent) {
       const diagnostics = evt.detail.diagnostics;
       errMarks.current.checkDiags = diagnostics;
-      markErrors(evt.detail.version);
+      markErrors();
       setHasCheckErrors(
         diagnostics.filter((d) => d.severity === "error").length > 0,
       );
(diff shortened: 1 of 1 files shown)


orpuente-MS:
Will this cause lag if the user is typing really fast in a really big document since now all versions will be marked for diagnostics?

billti (commented):
In your steps above, I don't understand while in this sequence (removed some text for clarity)

user types ( in the editor, document version is updated to 2
updateDocument command with contents Foo() version=2
user types ) in the editor, document version is updated to 3
updateDocument command with contents Foo() (no change) version=3

In the second bullet while are the contents "Foo()" when they haven't typed the ")" yet? Also, why do you say no change between version 2 and 3, when seems like the typing of ")" is the change?

minestarks:
> In your steps above, I don't understand while in this sequence (removed some text for clarity)
> 
> user types ( in the editor, document version is updated to 2 updateDocument command with contents Foo() version=2 user types ) in the editor, document version is updated to 3 updateDocument command with contents Foo() (no change) version=3
> 
> In the second bullet while are the contents "Foo()" when they haven't typed the ")" yet? Also, why do you say no change between version 2 and 3, when seems like the typing of ")" is the change?

@billti good catch! I forgot to clarify this. It's because the editor auto-completes the closing brace `)`. If the user _also_ types `)` over the auto-completed brace, the document version is bumped while the editor contents are unchanged.


user types ( in the editor, **editor autocompletes the closing brace** , document version is updated to 2

updateDocument command with contents Foo() version=2

user types ) in the editor, **editor "swallows" this keystroke, as it's typing the autocompleted closing brace**, document version is updated to 3

updateDocument command with contents Foo() (no change) version=3

minestarks:
> Will this cause lag if the user is typing really fast in a really big document since now all versions will be marked for diagnostics? Is it possible not increasing the version of the document if there are no changes?

@orpuente-MS the document version is managed by the editor (Monaco), not us, so this is beyond our control. We _could_ compare against the previous text contents and avoid re-checking, but this happens rarely enough that I don't think it's worth the complexity.


Your perf concern is legitimate - there _are_ perf issues when the user is typing fast. We recompile and re-check for every keystroke. Imagine you're typing `Qubit`. Every single version of the document, such as:
`Q`
`Qu`
`Qub`
will be compiled and checked, sequentially. Eventually, the error squiggles can lag significantly behind the typing (especially in debug builds). Ideally we should be batching these changes, but there are bugs here and we don't always do this successfully. 

This PR is kind of orthogonal to that problem. It doesn't fix it or make it worse.

idavis (approved):

orpuente-MS (approved):

PR #1909, opened 2024-09-09
Completions: existing glob import will prevent an exact import with t…

minestarks:
Completions: existing glob import will prevent an exact import with the same name from being auto-added

When an item and a namespace named `Foo.Bar` both exist, and the _namespace_ is glob-imported, the auto-import code "forgets" to include an auto-import text edit for `Foo.Bar`,  since it thinks `Foo.Bar` already imported.

```qsharp
 namespace Foo {
    operation Bar() : Unit {
    }
}

namespace Foo.Bar {
}

namespace Baz {
    import Foo.Bar.*;
    operation Main(): Unit {
        // selecting `Bar` from the completion list here should add an auto-import for `Foo.Bar` 
    }
}
```

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -739,6 +739,9 @@ fn callable_decl_to_completion_item(
     // An exact import is an import that matches the namespace
     // and item name exactly
     let preexisting_exact_import = imports.iter().any(|import_item| {
+        if import_item.is_glob {
+            return false;
+        }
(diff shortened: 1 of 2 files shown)


ScottCarda-MS (approved):
Nice catch!

ScottCarda-MS:
Does this issue also affect struct fields? i.e. Foo.Bar could be a variable `Foo` with field `Bar`.

sezna (approved):

Issue #1911, opened 2024-09-10
Migrating `Microsoft.Quantum.Core` to the modern Stdlib causes strange errors around resolving primitive types


sezna:
To reproduce this, check out my branch `alex/even-newer-stdlib` and run `cargo test` in `library/`.

For reasons unknown to me (yet), primitive types fail to resolve when the name of the core library is changed, even when the prelude and other hard-coded references to core are updated in the compiler.

sezna:
@swernli is doing a small investigation

sezna:
Figured this one out, fix coming...

sezna:
Closed in #1919

Issue #1910, opened 2024-09-10
Stabilize `Unstable.StatePreparation`


sezna:
@tcNickolas has pointed out that the state preparation library is pretty core to existing Q# code. It hasn't been iterated heavily in a while, and could be a good candidate for stabilization as a part of the standard library.

Related: 
- #1104  (we may want to address this before stabilization)

sezna:
Closed in #2022

PR #1912, opened 2024-09-10
Migrate the standard library to the project system

sezna:
This PR migrates the standard library into a Q# project structure.

Previously, we provided the modern standard library API via re-export, while the "true" stdlib was defined in the old explicit-namespace style. Now, that is inverted: the stdlib's code is housed in a project with implicit namespaces, and the legacy (`Microsoft.Quantum.____`) API is provided via re-exports. 

Why do this? Well, for one, the standard library should look like a regular library and use the latest features we have.
Secondarily, auto-imports and other language dev tools now refer to these items by their `Std.__` name, not their `Microsoft.Quantum.__` name. See the video below for  an example of this in completions:

https://github.com/user-attachments/assets/fa96e5b7-0cca-4a44-9224-7742d50f8370

Note that `Unstable.*` has not been migrated to the modern API, as we want to explore stabilizing state preparation (#1910)

Core is also not included, but for different reasons: #1911. As core is a part of the prelude, it is not likely a user will ever directly import core or use a fully qualified path to refer to refer to a core name, so it is okay if we wait a bit to diagnose that issue.


compiler/qsc_data_structures/src/namespaces/tests.rs
```diff
@@ -25,37 +25,41 @@ fn test_tree_construction() {
...
           children: [
-            ns1(id 11) {
+            Std(id 1) {
               children: [
-                nsc(id 14) {empty node},
-                nsb(id 13) {empty node},
-                nsa(id 12) {empty node},
+                Measurement(id 7) {empty node},
+                Canon(id 2) {empty node},
+                Intrinsic(id 6) {empty node},
```
ScottCarda-MS, line 32: Why did these go from being programmatically named namespaces to explicitly named namespaces?

library/std/src/QIR/Intrinsic.qs
```diff
@@ -0,0 +1,104 @@
...
+
+operation __quantum__qis__reset__body(target : Qubit) : Unit {
+    body intrinsic;
+}
+
+operation __quantum__qis__mresetz__body(target : Qubit) : Result {
+    body intrinsic;
+}
+
+export __quantum__qis__ccx__body, __quantum__qis__cx__body, __quantum__qis__cy__body, __quantum__qis__cz__body, __quantum__qis__rx__body, __quantum__qis__rxx__body, __quantum__qis__ry__body, __quantum__qis__ryy__body, __quantum__qis__rz__body, __quantum__qis__rzz__body, __quantum__qis__h__body, __quantum__qis__s__body, __quantum__qis__s__adj, __quantum__qis__t__body, __quantum__qis__t__adj, __quantum__qis__x__body, __quantum__qis__y__body, __quantum__qis__z__body, __quantum__qis__swap__body, __quantum__qis__m__body, __quantum__qis__reset__body, __quantum__qis__mresetz__body;
```
ScottCarda-MS, line 104: Might be a bit nicer to have these exports use the multiple-line format rather than have them all on one very long line.

compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -513,21 +513,19 @@ fn ambiguous_prelude() {
...
 
-            namespace namespace7 {
+            namespace namespace9 {
                 function item5() : Unit {
-                    A();
+                    item3();
                 }
             }
-
-            // AmbiguousPrelude { name: "A", candidate_a: "Microsoft.Quantum.Canon", candidate_b: "Microsoft.Quantum.Core", span: Span { lo: 181, hi: 182 } }
```
ScottCarda-MS, line 530: If this test no longer produces an Ambiguous Prelude error, then the test is broken, as that was the purpose of the test. If the intended result is for this to not produce Ambiguous Prelude errors anymore, then maybe the test should be removed?

library/std/src/Std/Diagnostics.qs
```diff
@@ -0,0 +1,320 @@
...
+/// last call to `StartCountingQubits`. If the qubits were not being counted, it triggers a runtime failure.
+///
+/// # Output
+/// The number of unique qubits allocated since the last call to `StartCountingQubits`.
+@Config(Unrestricted)
+operation StopCountingQubits() : Int {
+    body intrinsic;
+}
+
+export DumpMachine, DumpRegister, CheckZero, CheckAllZero, Fact, CheckOperationsAreEqual, StartCountingOperation, StopCountingOperation, StartCountingFunction, StopCountingFunction, StartCountingQubits, StopCountingQubits;
```
ScottCarda-MS, line 320: nit: here also I think it would be better to use multiple lines.

library/std/src/Std/ResourceEstimation.qs
```diff
@@ -0,0 +1,169 @@
...
+/// Assumed number of repetitions, factor to multiply the cost with
+operation RepeatEstimates(count : Int) : Unit is Adj {
+    body ... {
+        BeginRepeatEstimates(count);
+    }
+    adjoint ... {
+        EndRepeatEstimates();
+    }
+}
+export SingleVariant, BeginEstimateCaching, EndEstimateCaching, AuxQubitCount, TCount, RotationCount, RotationDepth, CczCount, MeasurementCount, PSSPCLayout, AccountForEstimates, BeginRepeatEstimates, EndRepeatEstimates, RepeatEstimates;
```
ScottCarda-MS, line 169: nit: multiple lines for the export would look better here

ScottCarda-MS (approved):

compiler/qsc_frontend/src/resolve/tests.rs
swernli: I think I see why... this used to be ambiguous prelude because `Microsoft.Quantum.Canon` was in the prelude, but now only `Std.Canon` is in the prelude. To keep the test checking what it is supposed to check, it should have the first namespace updated the same way.

library/src/tests/resources/src/qft_le.qs
```diff
@@ -1,6 +1,6 @@
 namespace Test {
     open Microsoft.Quantum.Diagnostics;
-    open Microsoft.Quantum.Arrays;
+    import Std.Arrays.*;
```
swernli, line 3: Why change this line but not the one above it?

library/src/tests/resources/src/select.qs
```diff
@@ -1,5 +1,5 @@
 namespace Test {
-    open Microsoft.Quantum.Arrays;
+    import Std.Arrays.*;
```
swernli, line 2: Same here... I would expect all of these to change or none of them.

library/src/tests/diagnostics.rs
```diff
@@ -9,7 +9,7 @@ fn check_operations_are_equal() {
     test_expression(
         "{
             open Microsoft.Quantum.Diagnostics;
-            open Microsoft.Quantum.Arrays;
+            import Std.Arrays.*;
```
swernli, line 12: Here too, Diagnostics could be changed to an import.

swernli (approved):
Some questions posted, but overall the change looks good. Looking forward to the follow up that finishes the fixups of the Core namespace!

compiler/qsc_data_structures/src/namespaces/tests.rs
sezna: It was previously the prelude + the programmatically inserted ones, and it still is. It's just the ordering that changed.

compiler/qsc_frontend/src/resolve/tests.rs
sezna: Yep, good catch -- updated!

PR #1914, opened 2024-09-11
Serialize logical post-layout overhead in resource estimation result

msoeken:
This change allows to serialize the logical post-layout overhead in the resource estimation result provided by the resource estimation API. Note that this was already possible in the system API that has a custom result for the VS Code and Python integration.

The change also simplified the type of the resource estimation result, which now is not any longer generic over the post-layout overhead trait.

### DIFF
--- a/resource_estimator/src/system.rs
+++ b/resource_estimator/src/system.rs
@@ -94,7 +94,7 @@ fn estimate_single<L: Overhead + LayoutReportData + PartitioningOverhead + Seria
             distillation_unit_templates,
             job_params.constraints().max_distillation_rounds,
         ),
-        logical_resources,
+        logical_resources.clone(),
         partitioning,
     );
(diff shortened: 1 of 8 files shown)


swernli (approved):

resource_estimator/src/system.rs
```diff
@@ -94,7 +94,7 @@ fn estimate_single<L: Overhead + LayoutReportData + PartitioningOverhead + Seria
             distillation_unit_templates,
             job_params.constraints().max_distillation_rounds,
         ),
-        logical_resources,
+        logical_resources.clone(),
```
ivanbasov, line 97: maybe a comments that the clone happens only once would be helpful

msoeken: Thank you, added a comment.

PR #1915, opened 2024-09-11
Basic samples for RE API

msoeken:
This adds to basic examples that illustrate how to use the RE API for extensibility. These samples only use existing models that are used in the RE integration with the QDK. Future samples will build on top of these and illustrate how to change models for qubits, error correction codes, factories, and post-layout logical overheads.


resource_estimator/examples/basic_qsharp.rs
```diff
@@ -0,0 +1,94 @@
...
+    let result = estimation.estimate().expect("estimation does not fail");
+
+    println!("Number of physical qubits: {}", result.physical_qubits());
+    println!(
+        "Runtime:                   {:.2e} secs",
+        result.runtime() as f64 / 1e9
+    );
+}
+
+fn compute_logical_counts_from_file() -> LogicalResourceCounts {
```
Manvi-Agrawal, line 55: - Given that interpreter and Package store objects might be unfamiliar to some users, it might be helpful to provide a more abstract layer. Thoughts?

- If we decide to abstract this, abstracting it at `qsharp.json` level might also help. Thoughts?

msoeken: Thanks for your feedback. In fact, I don't want to highlight the interpreter mechanism in this sample, but rather provide a sample that can easily be customized. I don't expect that this part of the code will change, except for maybe the source files. But explaining that multiple sources could be loaded can help.

Can you elaborate on the second point please. I don't quite understand it.

Manvi-Agrawal: > Thanks for your feedback. In fact, I don't want to highlight the interpreter mechanism in this sample, but rather provide a sample that can easily be customized. I don't expect that this part of the code will change, except for maybe the source files. But explaining that multiple sources could be loaded can help.

I feel if we have an API along the lines of  `compute_logical_counts_from_file(file, estimate_op)`, it simplifies user experience greatly since many might not familiar with internal workings. It seems the intension is that users will change the file_name and the estimate_operation, which we could simplify by having these things as arguments. Also, if we have this function in this example, the user might be more tempted to dig into this logic. Maybe this functionality can be provided via `resource-estimator` module.

> Can you elaborate on the second point please. I don't quite understand it.

For the second point, the intension is to be able to have an API that takes folder containing path to qsharp.json, or a valid path including `qsharp.json`. Eg: `samples\estimation\df-chemistry\qsharp.json` OR `samples\estimation\df-chemistry\`

msoeken: Thanks for the clarification. Meanwhile, I updated the PR to not contain the Q# sample and to only focus on the Rust functions in the `resource_estimator` crate. Thanks for your feedback!

ivanbasov (approved):

swernli (approved):
One minor comment about a missing header, otherwise changes look ready to merge, thanks!

resource_estimator/examples/basic_logical_counts.rs
```diff
@@ -0,0 +1,67 @@
+#![allow(clippy::cast_precision_loss)]
```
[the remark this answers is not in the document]
msoeken: Added the header.

PR #1919, opened 2024-09-16
Migrate core library to the new Stdlib API/projects system

sezna:
This is a follow-up to #1912, and includes an associated RCA bug fix (#1918 ) that was not exposed until I tried to re-export something from `Core`.

This also makes a subtle change to the compiler where the prelude is no longer created by default in the namespace tree, and is rather created when we compile `core.qs`. This was needed because we no longer allow "clobbering namespaces", that is, re-opening a namespace from an _external_ package.


compiler/qsc_frontend/src/resolve.rs
```diff
@@ -182,6 +182,12 @@ pub(super) enum Error {
         #[label]
         span: Span,
     },
+    #[error("this namespace clobbers (replaces) an existing external namespace of the same name")]
```
swernli, line 185: nitpick: I know "clobbered" is the right and even common term of art for this, but it feels odd in an error message for some reason. Perhaps "overwrites" would be similarly descriptive?

compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -204,11 +204,17 @@ fn compile(
...
+
     let mut globals = super::GlobalTable::new();
     let mut errors = globals.add_local_package(&mut assigner, &package);
     let mut resolver = Resolver::new(globals, dropped_names);
+    // no errors here
     resolver.bind_and_resolve_imports_and_exports(&package);
+    // what about here?
+    println!("i have {} errors", resolver.errors.len());
     resolver.with(&mut assigner).visit_package(&package);
+    println!("now i have {} errors", resolver.errors.len());
```
swernli, line 217: Looks like these might have been leftover from testing.

swernli (approved):
A few minor things to fix, but otherwise this is looking good!

compiler/qsc_frontend/src/resolve/tests.rs
sezna: oops 😓 fixed!

cesarzc (approved):

PR #1921, opened 2024-09-16
Update circuits widget sizing behavior

swernli:
This change fixes a few quirks of the `Circuit` widget, specifically:

 - if the width of the widget is greater than the width of the cell output area, scroll bars are added, and
 - setting a custom zoom level by changing the value in the zoom control disables the resize handling so that VS Code doesn't incorrectly trigger a recalculation of the default zoom and undo the manual change.

Confirmed with manual testing that window resizing in VS Code circuit display is still hadled properly.

### DIFF
--- a/npm/qsharp/ux/circuit.tsx
+++ b/npm/qsharp/ux/circuit.tsx
@@ -39,6 +39,7 @@ function ZoomableCircuit(props: { circuit: qviz.Circuit }) {
   const circuitDiv = useRef<HTMLDivElement>(null);
   const [zoomLevel, setZoomLevel] = useState(100);
   const [rendering, setRendering] = useState(true);
+  const [zoomOnResize, setZoomOnResize] = useState(true);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


billti (approved):

orpuente-MS (approved):

minestarks:
@swernli thanks for digging into this. I pushed my fix over your fix :D  I think it works, but I would appreciate if you tested it in the environment where you saw the bug.

swernli:
> @swernli thanks for digging into this. I pushed my fix over your fix :D I think it works, but I would appreciate if you tested it in the environment where you saw the bug.

The resizing and zoom work great where I tested them, but it looks like the scrolling fix got lost in the shuffle. I pushed that up, and now I think we are good to go.

@billti I think this may fix the issues you were seeing in Python interactive as well.

minestarks:
@swernli oops, sorry, yes, thanks for fixing that! Teamwork! ✊

PR #1930, opened 2024-09-24
Add str value for TargetProfile

idavis:


swernli (approved):

pip/src/interpreter.rs
```diff
@@ -110,6 +110,19 @@ pub(crate) enum TargetProfile {
     Unrestricted,
 }
 
+#[pymethods]
+#[allow(clippy::trivially_copy_pass_by_ref)]
+impl TargetProfile {
+    fn __str__(&self) -> String {
+        match self {
```
minestarks, line 117: Can you use `to_str()` from `Profile`? `Profile::from(self).to_str()` or something like that?

https://github.com/microsoft/qsharp/blob/ffadd7e490f14f3641888ad3c5775a0f754227de/compiler/qsc/src/target.rs#L17-L22

minestarks (approved):
Approved with suggestion to improve code duplication.

PR #1932, opened 2024-09-25
add Q# package registry document

sezna:
A basic document that serves the purpose of a package registry, listing all packages we are aware of.

Most information about the package should be on the package page itself, so here we can just list titles, summaries, and links.

### DIFF
--- a/library/Registry.md
+++ b/library/Registry.md
@@ -1,0 +1,22 @@
+# Q# Package Registry
+
+This document is a list of all known (to us) Q# packages. If you have a package that you would like to add to this list, please open a pull request.
+
+## Unstable
+
+[link](https://github.com/microsoft/qsharp/tree/main/library/unstable)
+
+A general staging ground for some useful APIs that have not yet been stabailized into the standard library.
+
+## Signed
+
+[link](https://github.com/microsoft/qsharp/tree/main/library/signed)
+
+Defines types and functions to work with signed qubit-based integers.
+
+## FixedPoint
+
+[link](https://github.com/microsoft/qsharp/tree/main/library/fixed_point)
+
+Types and functions for fixed-point arithmetic with qubits.
+
--- a/library/Registry.md
+++ b/library/Registry.md
@@ -1,6 +1,6 @@
-# Q# Package Registry
+# Q# Library Registry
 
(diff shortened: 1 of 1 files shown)


library/Registry.md
```diff
@@ -0,0 +1,22 @@
+# Q# Package Registry
+
+This document is a list of all known (to us) Q# packages. If you have a package that you would like to add to this list, please open a pull request.
```
swernli, line 3: I would err on the side of saying "notable" rather than "all known (to us)" in there. Also, did we want to use the work "package" or "project" here? I think we were trying to get away from the implications of "package." Maybe "library projects" is better? Tried it with both changes below.
```suggestion
This document is a list of notable Q# library projects. If you have a package that you would like to add to this list, please open a pull request.
```

library/Registry.md
```diff
@@ -0,0 +1,22 @@
+# Q# Package Registry
+
+This document is a list of notable Q# library projects. If you have a package that you would like to add to this list, please open a pull request.
```
swernli, line 3: Oops, I forgot another instance of the word "package:"
```suggestion
This document is a list of notable Q# library projects. If you have a project that you would like to add to this list, please open a pull request.
```

library/Registry.md
```diff
@@ -0,0 +1,22 @@
+# Q# Package Registry
```
swernli, line 1:
```suggestion
# Q# Project Registry
```

library/Registry.md
```diff
@@ -0,0 +1,22 @@
...
+
+This document is a list of notable Q# library projects. If you have a project that you would like to add to this list, please open a pull request.
+
+## Unstable
+
+[link](https://github.com/microsoft/qsharp/tree/main/library/unstable)
+
+A general staging ground for some useful APIs that have not yet been stabailized into the standard library.
```
cesarzc, line 9:
```suggestion
A general staging ground for some useful APIs that have not yet been stabilized into the standard library.
```

cesarzc (approved):
Approved with one comment on a typo.

library/Registry.md
swernli: Now I'm less certain about this suggestion... settling on a specific terminology is something I think we need some consensus on.

swernli (approved):

Issue #1928, opened 2024-09-24
Request for Durr Hoyer Library


mertall:
**Is your feature request related to a problem? Please describe.**

I haven't seen an implementation of Durr Hoyer that is ready to use. I would like to contribute to a Q# Durr Hoyer library. 
I started this project as a Junior in Undergrad, almost 4 years ago. Past 4 years have been working on my SWE skills and after continued thought and iteration I came up with a quick and dirty implementation of Durr Hoyer Algorithm to find Min and Max. At the very least I think the logic will be educational for many people, I understood MSB comparison and how it relates to quantum information which for me was really cool to see.

**Describe the solution you'd like**

Eventually I would like to take this algorithm and also implement finding range, and median of unsorted database. Also do some actual experiments on quantum computer. 

**Describe alternatives you've considered**

I am considering more optimized methods like [QESA](https://arxiv.org/pdf/1908.07943) which increase accuracy to nearly 100%. 

**Additional context**
I have attached my current Jupyter notebook file, which was ran in Azure quantum Workspace
[DurrHoyerImplementation.ipynb.zip](https://github.com/user-attachments/files/17107621/DurrHoyerImplementation.ipynb.zip)

sezna:
We would love to see this! If you encapsulate the algorithmic logic in a Q# project, we can certainly put together a Q# library for it, and I would be happy to assist you in that!

billti:
FYI if not familiar, here's the docs and creating projects that reference other projects <https://learn.microsoft.com/en-us/azure/quantum/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-qsharp-run#example-external-project>

mertall:
@zesna I will begin work to encapsulate this in a Q# project.
@billti that was my next question. Thank you!

Thank you so much for the opportunity!! This has been a long and rewarding journey. I'm excited to get this across for other users!

mertall:
I created a Github repo with a Q# project. Please advise on next steps to integrate with Q#... should I be cloning Q# and finding a good place for this code? Thank you for your patience. 

https://github.com/mertall/DurrHoyerLibrary/tree/main

Manvi-Agrawal:
Nice Addition to Q# library

> Please advise on next steps to integrate with Q#... should I be cloning Q# and finding a good place for this code? Thank you for your patience.

[...]

PR #1936, opened 2024-09-25
Durr hoyer library min and max

mertall:
Unit Tests worked standalone with 1000 shots. Above 50% accuracy. 💪🏾  (When running locally in my own GitHub Repo)

Can't seem to get DrawRandomInt to work with` python ./build.py`. I can't find any subitutes other than a QuantumRandomNumberGenerator, but that is not feasible when I read the docs on the implementation of the Operation in Q# docs. I was thinking maybe allowing the random int to come from outside of the Algorithm as an input, to be found in a classical computer via a python script run by a hypothetical user. 

Let me know what you all think. @sezna @billti @Manvi-Agrawal 

Issue #1928

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -247,7 +247,7 @@ fn get_indent(compilation: &Compilation, package_offset: u32) -> String {
         .try_into()
         .expect("offset can't be converted to uszie");
     let before_offset = &source.contents[..source_offset];
-    let mut indent = match before_offset.rfind(|c| c == '{' || c == '\n') {
+    let mut indent = match before_offset.rfind(&['{', '\n'][..]) {
         Some(begin) => {
             let indent = &before_offset[begin..];
             indent.strip_prefix('{').unwrap_or(indent)
--- a/library/src/tests/resources/src/durrhoyerlibrary.qs
+++ b/library/src/tests/resources/src/durrhoyerlibrary.qs
@@ -5,7 +5,8 @@ namespace Test {
     import Std.Arrays.*;
     import Std.Convert.*;
     import Std.Random.*
-    import Std.DurrHoyerLibrary.*; function MaxIntArray(arr : Int[]) : Int {
+    import Std.DurrHoyerLibrary.*;
+    function MaxIntArray(arr : Int[]) : Int {
         mutable max = arr[0];
         for i in 1..(Length(arr) - 1) {
             if (arr[i] > max) {
@@ -65,21 +66,6 @@ namespace Test {
 
         // Use the shared logic to run the test with "min" type
         RunDurrHoyerTestWithShots(testLists, expectedMinIndices, shots, "min");
-    }
-
-    // Maximum test case using the shared logic
-    operation RunDurrHoyerMaximumUnitTestWithShots(shots : Int) : Unit {
-        let testLists : Int[][] = [
-            [2, 3, 1, 5, 4],
-            [1, 5, 4, 3, 6],
-            [7, 5, 6, 1, 2]
-        ];
-
-        // Expected results (maximum element index for each list)
-        let expectedMaxIndices : Int[] = [3, 4, 0];
-
-        // Use the shared logic to run the test with "max" type
-        RunDurrHoyerTestWithShots(testLists, expectedMaxIndices, shots, "max");
     }
     operation RunDurrHoyerMaximumUnitTestWithShots(shots : Int) : Unit {
         let testLists : Int[][] = [
--- a/library/src/tests/resources/src/durrhoyerlibrary.qs
+++ b/library/src/tests/resources/src/durrhoyerlibrary.qs
@@ -83,4 +83,30 @@ namespace Test {
         // Use the shared logic to run the test with "max" type
         RunDurrHoyerTestWithShots(testLists, expectedMaxIndices, shots, "max");
     }
+    operation RunDurrHoyerMaximumUnitTestWithShots(shots : Int) : Unit {
+        let testLists : Int[][] = [
+            [2, 3, 1, 5, 4],
+            [1, 5, 4, 3, 6],
+            [7, 5, 6, 1, 2]
+        ];
+
+        // Expected results (maximum element index for each list)
+        let expectedMaxIndices : Int[] = [3, 4, 0];
+
+        // Use the shared logic to run the test with "max" type
+        RunDurrHoyerTestWithShots(testLists, expectedMaxIndices, shots, "max");
+    }
+    operation RunDurrHoyerZeroValuesUnitTestWithShots(shots : Int) : Unit {
+    let testLists = [
+        [0, 3, 1, 2, 4],
+        [6, 0, 4, 3, 1],
+        [7, 5, 6, 0, 2]
+    ];
+
+    // Expected results (minimum element index for each list)
+    let expectedMinIndices = [0, 1, 3];
+
+    // Use the shared logic to run the test with "min" type
+    RunDurrHoyerTestWithShots(testLists, expectedMinIndices, shots, "min");
+    }
 }
--- a/library/src/tests/resources/src/durrhoyerlibrary.qs
+++ b/library/src/tests/resources/src/durrhoyerlibrary.qs
@@ -1,17 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 namespace Test {
-    open Microsoft.Quantum.Intrinsic;
-    open Microsoft.Quantum.Canon;
-    open Microsoft.Quantum.Math;
-    open Microsoft.Quantum.Measurement;
-    open Microsoft.Quantum.Arrays;
-    open Microsoft.Quantum.Convert;
-    open Microsoft.Quantum.Random;
-    open Microsoft.Quantum.Core;
-    open Microsoft.Quantum.Diagnostics;
-    open Microsoft.Quantum.DurrHoyerLibrary;
-
+    import Std.Math.*;
+    import Std.Arrays.*;
+    import Std.Convert.*;
+    import Std.Random.*
+    import Std.DurrHoyerLibrary.*;
     // Function to find the maximum element in an array
     function MaxIntArray(arr : Int[]) : Int {
         mutable max = arr[0];
@@ -77,10 +71,10 @@ namespace Test {
         ];
 
         // Expected results (maximum element index for each list)
-        let expectedMinIndices : Int[] = [2, 4, 3];
+        let expectedMaxIndices : Int[] = [2, 4, 3];
 
         // Iterate over test cases
-        for (list, expectedMinIndex) in Zipped(testLists, expectedMinIndices) {
+        for (list, expectedMaxIndex) in Zipped(testLists, expectedMaxIndices) {
             let lenghtList : Int = Lenght(list)
             let candidateMax : Int = DrawRandomInt(0, lengthList - 1)
             let maxValue = MaxIntArray(list);
@@ -93,10 +87,10 @@ namespace Test {
 
             // Run the Durr-Hoyer algorithm multiple times (shots)
             for _ in 1..shots {
-                let minIndex : Int = DurrHoyerFinding(list, nQubits, "max", candidateMax, lengthList);
+                let maxIndex : Int = DurrHoyerFinding(list, nQubits, "max", candidateMax, lengthList);
 
                 // Check if the found index matches the expected maximum index
-                if (minIndex == expectedMinIndex) {
+                if (maxIndex == expectedMaxIndex) {
                     set correctCount += 1;
                 }
             }
@@ -111,5 +105,4 @@ namespace Test {
             Message($"List: {list}");
             Message($"Probability of finding the maximum is {probability * 100.0}%");
         }
-    }
-}
+    } }
--- a/library/src/tests/resources/src/durrhoyerlibrary.qs
+++ b/library/src/tests/resources/src/durrhoyerlibrary.qs
@@ -66,4 +66,47 @@ namespace Test {
             Message($"Probability of finding the minimum is {probability * 100.0}%");
         }
     }
+    // Function to compute the probability of finding the maximum index
+    operation RunDurrHoyerMaximumUnitTestWithShots(shots : Int) : Unit {
+        // Define test lists for the unit test
+        let testLists : Int[][]= [
+            [5, 3, 1, 2, 4],
+            [6, 5, 4, 3, 1],
+            [7, 5, 6, 1, 2]
+        ];
+
+        // Expected results (maximum element index for each list)
+        let expectedMinIndices : Int[] = [2, 4, 3];
+
+        // Iterate over test cases
+        for (list, expectedMinIndex) in Zipped(testLists, expectedMinIndices) {
+            let maxValue = MaxIntArray(list);
+            let double : Double = IntAsDouble(maxValue + 1);
+            let log : Double = Log(double) / Log(2.0);
+            let nQubits = Ceiling(log);
+
+            // Variable to track how many times we find the correct maximum index
+            mutable correctCount = 0;
+
+            // Run the Durr-Hoyer algorithm multiple times (shots)
+            for _ in 1..shots {
+                let minIndex : Int = DurrHoyerFinding(list, nQubits, "min");
+
+                // Check if the found index matches the expected maximum index
+                if (minIndex == expectedMinIndex) {
+                    set correctCount += 1;
+                }
+            }
+
+            // Calculate the probability of finding the correct maximum
+            let probability = IntAsDouble(correctCount) / IntAsDouble(shots);
+
+            // Assert that the probability is above 50%
+            Fact(probability >= 0.5, $"Probability of finding the maximum for list {list} is less than 50%. Found: {probability * 100.0}%");
+
+            // Optionally print debugging info
+            Message($"List: {list}");
+            Message($"Probability of finding the maximum is {probability * 100.0}%");
+        }
+    }
 }
--- a/library/src/tests/resources/src/durrhoyerlibrary.qs
+++ b/library/src/tests/resources/src/durrhoyerlibrary.qs
@@ -7,10 +7,9 @@ namespace Test {
     import Std.Random.*
     import Std.DurrHoyerLibrary.*;
 
-    // Function to find the maximum element in an array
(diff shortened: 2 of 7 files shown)


mertall:
> @mertall please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
>  > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="{Accenture Federal Services}"]

mertall:
> @mertall the command you issued was incorrect. Please try again.
> 
> Examples are:
> 
> ```
> @microsoft-github-policy-service agree
> ```
> 
> and
> 
> ```
> @microsoft-github-policy-service agree company="your company"
> ```

@microsoft-github-policy-service agree company="Accenture Federal Services"

mertall:
@microsoft-github-policy-service agree company="Accenture Federal Services"

mertall:

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,226 @@
+open Microsoft.Quantum.Intrinsic;
+open Microsoft.Quantum.Canon;
+open Microsoft.Quantum.Math;
+open Microsoft.Quantum.Measurement;
+open Microsoft.Quantum.Arrays;
+open Microsoft.Quantum.Convert;
+open Microsoft.Quantum.Random;
+open Microsoft.Quantum.Core;
+open Microsoft.Quantum.Diagnostics;
```
Manvi-Agrawal, line 9:
```suggestion
import Std.Math;
import Std.Arrays;
import Std.Convert;
import Std.Random;
open Std.Diagnostics;
```
We can remove some of imports here since they are part of prelude(ie  included in Q# by Default). Also, its recommended to import from the Std namespace.

 See [this code](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_data_structures/src/namespaces.rs#L10-L15) for the list of preludes

pub const PRELUDE: &[&[&str]; 4] = &[
    &["Std", "Canon"],
    &["Microsoft", "Quantum", "Core"],
    &["Std", "Intrinsic"],
    &["Std", "Measurement"],
];

Manvi-Agrawal, line 9: Could you please do a audit of remaining imports after removing the preludes? On a brief look, Diagnostics and Random seem to be unused.

mertall: Ah I glossed over that pattern in the other files. thanks for sharing how those preludes come into play. I was confused by that Bec cause I didn't see a Quantum namespace displayed anywhere on the other Std files.

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,226 @@
...
+}
+function ResultAsInt(r : Result) : Int {
+    if (r == One) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
+function BitsToInt(bits : Result[]) : Int {
```
Manvi-Agrawal, line 58: This looks quite similar to `ResultArrayAsInt` function. Possibly remove?
Q# Ref guide: https://learn.microsoft.com/qsharp/api/

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,222 @@
...
+
+// Grover iteration with the oracle and diffusion operator for max
+operation GroverIterationMax(threshold : Int, inputQubits : Qubit[], auxQubit : Qubit) : Unit {
+    OracleMoreThan(threshold, inputQubits, auxQubit);
+    DiffusionOperator(inputQubits);
+}
+
+// Dürr-Høyer for finding min or max algorithm
+operation DurrHoyerAlgorithm(list : Int[], nQubits : Int, type : String) : Int {
+    mutable candidateMin = DrawRandomInt(0, Length(list) - 1);  // Random initial candidate
```
Manvi-Agrawal, line 135: You are calling `DrawRandomInt` here. Is the intension to pass this value to the `DurrHoyerAlgorithm` operation for testing purposes? 

One way I can see that happening is that you provide candidate as an input parameter. That way the testing code from Python/Q# would be able to pass the random integer.

Manvi-Agrawal, line 135: Nit: Possibly rename to candidate since this variable is being used for both max and min.

library/src/tests/resources/src/durrhoyerlibrary.qs
Manvi-Agrawal, line 1: - Some of the comments from previous file regarding imports apply here.
- Also, could you please extract common logic from RunDHMinUT and RunDHMaxUT?

library/src/tests/resources/src/durrhoyerlibrary.qs
```diff
@@ -0,0 +1,112 @@
...
+            // Assert that the probability is above 50%
+            Assert(probability > 0.5, $"Probability of finding the minimum for list {list} is less than 50%. Found: {probability * 100.0}%");
+
+            // Optionally print debugging info
+            Message($"List: {list}");
+            Message($"Probability of finding the minimum is {probability * 100.0}%");
+        }
+    }
+    // Function to compute the probability of finding the maximum index
+    operation RunDurrHoyerMaximumUnitTestWithShots(shots : Int) : Unit {
```
Manvi-Agrawal, line 70: Possibly a copy-paste typo. I dont see any difference at all here except for the function name.

language_service/src/completion.rs
```diff
@@ -247,7 +247,7 @@ fn get_indent(compilation: &Compilation, package_offset: u32) -> String {
         .try_into()
         .expect("offset can't be converted to uszie");
     let before_offset = &source.contents[..source_offset];
-    let mut indent = match before_offset.rfind(|c| c == '{' || c == '\n') {
+    let mut indent = match before_offset.rfind(&['{', '\n'][..]) {
```
Manvi-Agrawal, line 250: Do we really need to change this? IMO this PR should focus only on implementing DH library. 

I saw your commit message saying that build.py doesnt work for you without this change. I will leave this decision to @sezna.

Manvi-Agrawal:
@mertall , Mostly, looks good. Left a couple of nits regarding import and code reuse. Overall you have done a good job writing separating code into functions and using meaningful variable names. Could you please look further into some more places if the code can be reused or a library function can be used. I pointed at a couple of places.

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// Grover iteration with the oracle and diffusion operator for max
+operation GroverIterationMax(threshold : Int, inputQubits : Qubit[], auxQubit : Qubit) : Unit {
+    OracleMoreThan(threshold, inputQubits, auxQubit);
+    DiffusionOperator(inputQubits);
+}
+
+// Dürr-Høyer for finding min or max algorithm
+operation DurrHoyerAlgorithm(list : Int[], nQubits : Int, type : String) : Int {
```
Manvi-Agrawal, line 134: Docs please since this would be a public facing function. You can refer tob existing code on how to write doc comments so that they get featured on Q# website

library/std/src/Std/DurrHoyerLibrary.qs
mertall: I have moved this logic outside of the operation

library/std/src/Std/DurrHoyerLibrary.qs
mertall: I had added comments as required by guide, it looks like it got lost as I was developing. Adding required summary sections now.

language_service/src/completion.rs
mertall: If there is a way to skip over the error when I run the build.py I am happy to change it. I agree it is out of place, and will be hard to log trace down the line while it is included in this PR. maybe we can create a ticket to address this and I can push a PR for that?

library/src/tests/resources/src/durrhoyerlibrary.qs
mertall: thanks for catching this !

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,207 @@
+import Std.Math.*;
+import Std.Arrays.*;
+import Std.Convert.*;
+
+function CountElements(list : Int[], threshold : Int, comparisonType : String) : Int {
```
Manvi-Agrawal, line 5: This function can be simplified by passing in a comparison operator to avoid type=="min"/"max" in code.

Please refer to this [Playground sample](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAACoVRTUvDQBC951c8gocEt9SKCjZuBMVDD70InkoPMW7Mwma2ZDdQifnvstmYbvHgHAbmzZt581F1VFqpCc%252B6I%252FuiRCPImkRJY7HGhuxuz2DrVphaqw8PMZS6ORStNJqwRrIhu8jxpLVKU89AHwFA09niXQmUrjc4rrJoxCvdQngpSMIo5iucySron0y8NCA4M8JObS85VtmcG5wbo8FrtcJ2LXluFg1RVP1uvC0kJW7gN5L%252FTayE9WNy7FYM1wy3DHcM9%252FuAcDoTx02Ifx0EOOJGUuwnXS7PE8VxSqhxrfm4fKJMxXjEcZHj%252BHBS%252BvZIPiOT7lYYU3yK5CJ%252BFQYc%252Fd%252F%252FBn8NP5oObHyQU%252Ba98wNsXdDZev0cDHGaRe6uP3MibiNKAgAA&profile=unrestricted) for more details.

library/std/src/Std/DurrHoyerLibrary.qs
```diff
@@ -0,0 +1,207 @@
...
+                    return candidate;
+                }
+
+                // Calculate the optimal number of Grover iterations
+                let N : Int = Length(list);
+                let optimalIterations : Int = Round((PI() / 4.0) * Sqrt(IntAsDouble(N) / IntAsDouble(M)));
+
+                // Perform Grover iterations for min or max
+                for i in 1..optimalIterations {
+                    if (type == "min") {
```
Manvi-Agrawal, line 153: Re: Similarly here

Possibly rewrite this:

let groverIteration = (type=="min") ? GroverIterationMin | GroverIterationMax;
groverIteration(list[candidate], inputQubits, auxQubit)

library/src/tests/resources/src/durrhoyerlibrary.qs
```diff
@@ -0,0 +1,108 @@
...
+        // Define test lists for the unit test
+        let testLists : Int[][] = [
+            [5, 3, 1, 2, 4],
+            [6, 5, 4, 3, 1],
+            [7, 5, 6, 1, 2]
+        ];
+
+        // Expected results (maximum element index for each list)
+        let expectedMaxIndices : Int[] = [2, 4, 3];
```
Manvi-Agrawal:
```suggestion
        let expectedMaxIndices : Int[] = [0, 0, 0];
```
[Question] Why do we expect max to return [2, 4, 3]? I can understand for min because:
min([5, 3, 1, 2, 4]) = 1, 1 is at idx 2 (0 based idx),
min([6, 5, 4, 3, 1]) = 1, 1 is at idx 4 (0 based idx),
min([7, 5, 6, 1, 2]) = 1, 1 is at idx 3 (0 based idx),


 I can't understand for max because:
max([5, 3, 1, 2, 4]) = 5, 5 is at idx 0 (0 based idx),
max([6, 5, 4, 3, 1]) = 6, 6 is at idx 0 (0 based idx),
max([7, 5, 6, 1, 2]) = 7, 7 is at idx 0 (0 based idx),

mertall: Yes I just pushed fix for that. Saw that when you noticed that names were the same.

library/src/tests/resources/src/durrhoyerlibrary.qs
```diff
@@ -0,0 +1,82 @@
...
+        operationType : String
+    ) : Unit {
+        for (list, expectedIndex) in Zipped(testLists, expectedIndices) {
+            let lengthList : Int = Length(list);
+            let candidateIndex : Int = DrawRandomInt(0, lengthList - 1);
+            let maxValue = MaxIntArray(list);
+            let double : Double = IntAsDouble(maxValue + 1);
+            let log : Double = Log(double) / Log(2.0);
+            let nQubits = Ceiling(log);
+            let testLists : Int[][] = [
```
Manvi-Agrawal, line 35: Some comments on the tests:
- What happens if some/all of the values are negative in a `testList`?
- What happens if 0 is passed as input  in a `testList`?
- Can we please rearrange the values so that expectedMaxIndex is not 0 in all the test cases?

mertall:
moved to fast on requested changes, lost 50% accuracy while trying to abstract...

I get better debugging as a local Q# project. When working within the Q# repo I forked, I do not get the same debugging... Wondering if you had any suggestions to improve dev experience while in forked repo.

for now solution will be to make changes in my local Q# project as debugging is much easier.

Manvi-Agrawal:
> I get better debugging as a local Q# project. When working within the Q# repo I forked, I do not get the same debugging... Wondering if you had any suggestions to improve dev experience while in forked repo. Made an assumption earlier the debugging is same and went flying with changes. Once I moved it over to my Q# project to update that repo, I saw so many squiggly red lines :o ... :') .

If you could elaborate on what is slowing you down it would help us understand this better. I am curious are you running `build.py` everytime to run the tests? That can be a major deal-breaker. I have worked on compiler module a bit to implement the lint rules, and running rust unit tests was very fast.   I guess you are need to ensure that tests in `library/src/tests/durrhoyerlibrary.rs` pass. After a first successful build with `build.py`, you should just run specific Rust unit test(s). First unit test run might be a bit slow, and then it gets better.

You can click on "Run Test" option that comes up in the test file to run a particular test. You should get this option if your repo is properly setup. It also shows command run under the hood `C:\Users\manvi\.cargo\bin\cargo.exe test --package library --lib -- tests::arrays::check_all --exact --show-output` in this case.
image

I was able to run all of the tests in `library\src\tests\arrays.rs` by running the command on powershell that vscode by omitting the test name. I ran `C:\Users\manvi\.cargo\bin\cargo.exe test --package library --lib -- tests::arrays --exact --show-output`  locally in powershell on windows. Based on this observation, I think this command might help you `<cargo.exe_path> test --package library --lib -- tests::durrhoyerlibrary --exact --show-output`.

image

mertall:
<cargo.exe_path> test --package library --lib -- tests::durrhoyerlibrary --exact --show-output

this should help me. I was running the build and made an assumption the tests I created were included during that build. Thank you!

mertall:
my rust test file is improperly created. vs code is not highlighting any code on that file. I am inexperienced in rust so could use help in adjusting that script to meet whatever requirements I am missing. In mean time will make changes on my GitHub repo and copy and paste changes as they work. Thank you of your patience

library/src/tests/resources/src/durrhoyerlibrary.qs
mertall: for negatives in ConvertToBinary:

if (value < 0 or value > maxVal) {
    fail $"Value {value} cannot be represented with {length} bits.";
}

ConvertToBinary will fail if there are negatives

mertall: Added unit test for 0 in input list

mertall:
Testing suite is working, and the test cases are getting hit, but they are not passing due to this error:

error: Frontend(Error(Resolve(GlobImportNamespaceNotFound("DurrHoyerLibrary", Span { lo: 283, hi: 303 })))) })
...
error: Frontend(Error(Resolve(NotFound("DurrHoyerAlgorithm", Span { lo: 1400, hi: 1418 }))))

sezna:
Hey @mertall  -- this looks like awesome work!

I'm not sure this is a good fit for the standard library, though. The standard library is for things that have gone through a process of stabilization, with an API that we're willing to commit to forever. I think this algorithm would be best published as a Q# package (library). If it gets wide adoption, is seen as a "general utility", and we're happy with the API, then we can consider adding it to the standard library. Until then, I would suggest maintaining it as a library that you own.

We have a document here describing how to publish libraries for Q#: [our repo wiki](https://github.com/microsoft/qsharp/wiki/Q%23-External-Dependencies-(Libraries)) and another document [on learn.microsoft.com](https://learn.microsoft.com/en-us/azure/quantum/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-qsharp-run).

After you create your library, you can add it to our registry. Once https://github.com/microsoft/qsharp/pull/1932 goes in, you can put up a PR that adds your library as a list of known libraries.

I hope I'm not taking any wind out of your sails here -- our standard library is _not_ meant to be a good example of a library, as it requires a lot of special casing for testing etc. The other libraries, like `fixed_point`, are true libraries. I'd recommend using those as a template. You'll notice their tests are written in Q# and not Rust.

mertall:
Thank you for the feedback @sezna. To clarify steps moving forward:

1. Maintain project in my own repo here: https://github.com/mertall/DurrHoyerLibrary
2. Start a new PR that follows step listed here to modify this Q# Repo to include the above repo as an external dependancy.

sezna:
> No worries on taking any wind out of my sails. I can't say I was not excited to contribute to Q#! When the time is right it shall happen. Thank you for the support so far!

This is absolutely a contribution to Q#, we are really excited about third party libraries and would love to help you get that set up! And there's plenty of opportunity to contribute here as well 😄

PR #1938, opened 2024-09-26
Optimize RIR reindexing, QIR qubit use

swernli:
This change updates the behavior of the RIR qubit reindexing to assume single block only and then take additional steps to optimize qubit usage with that block. This is consistent with the assumption that qubit reuse is required for adaptive programs rather than a separately configurable capability.

In addition, updates the qasm3 compiling to take advantage of this optimization by having explicitly call `__quantum__qis__m__body`, and updates the corresponding tests.


compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -231,6 +245,21 @@ impl ReindexQubitPass {
     }
 }
 
+fn qubit_used_in_instrs(ids_used: u32, instrs: &[Instruction]) -> bool {
```
idavis, line 248: nit: `ids_used` is describing a single id, but name suggests plural.

compiler/qsc_rir/src/passes/reindex_qubits.rs
```diff
@@ -187,27 +191,37 @@ impl ReindexQubitPass {
...
-                            self.cx_id,
-                            vec![
-                                new_args[0],
-                                Operand::Literal(Literal::Qubit(qubit_map.next_qubit_id)),
-                            ],
-                            None,
-                        ));
-                        self.highest_used_id = self.highest_used_id.max(qubit_map.next_qubit_id);
+                    if *call_id == self.m_id {
+                        if qubit_used_in_instrs(ids_used[0], &instrs[i + 1..]) {
```
idavis, line 195: is the indexing of `[i + 1..]` safe? Will this crash if we're on the last instr?

cesarzc, line 195:
```suggestion
                        let qubit_id = ids_used.first().expect("a measurement call should have one argument"); 
                        if qubit_used_in_instrs(qubit_id, &instrs[i + 1..]) {
```

cesarzc: This is along the same lines as Ian's comment. At first glance, it was not clear what `ids_used[0]` meant and whether it was safe. Explicitly calling `first()` and adding an error message helps to make that clear.

swernli: Switched it to using iterators.

idavis (approved):

cesarzc (approved):

PR #1939, opened 2024-09-26
Fix bug preventing display of circuits where same qubit measured more than once

swernli:
The logic for returning the count of measurements incorrectly reported every qubit as being measured only once no matter how many times it was measured. This was breaking circuit display for non-base profile programs.

### DIFF
--- a/compiler/qsc/src/interpret/circuit_tests.rs
+++ b/compiler/qsc/src/interpret/circuit_tests.rs
@@ -94,7 +94,7 @@ fn measure_same_qubit_twice() {
         .circuit(CircuitEntryPoint::EntryPoint, false)
         .expect("circuit generation should succeed");
 
-    expect![[r#"
+    expect![["
         q_0    ── H ──── M ──── M ──
                          ╘══════╪═══
                                 ╘═══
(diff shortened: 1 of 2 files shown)


compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -72,6 +72,36 @@ fn one_gate() {
...
+            }
+        ",
+        Profile::Unrestricted,
+    );
+
+    let circ = interpreter
+        .circuit(CircuitEntryPoint::EntryPoint, false)
+        .expect("circuit generation should succeed");
+
+    expect![[r#"
```
swernli, line 97:
```suggestion
    expect![["
```

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -72,6 +72,36 @@ fn one_gate() {
...
+        q_0    ── H ──── M ──── M ──
+                         ╘══════╪═══
+                                ╘═══
+    "#]]
```
swernli, line 101:
```suggestion
    "]]
```

minestarks (approved):

orpuente-MS (approved):

PR #1947, opened 2024-10-03
More precise completions and namespace member completions

minestarks:
## Feature overview

This PR enables:

1.  Much more precise keyword completions where we only show the relevant syntax at a particular cursor location. For example:

```qsharp
within { } | 
```

At the cursor location indicated by `|`, we should only get the keyword `apply` .

2. More precise completions for names where we take the syntactic context into account. For example:

```qsharp
operation Foo(x : | )
```

Here, we should only get type names.

3. Namespace member completions. For example:
```qsharp
import Std.Diagnostics.|
```

The completions here should return only items that exist under `Std.Diagnostics`.

### Parser completions module

Introduce the `completion` module in the parser.

- `possible_words_at_offset_in_source` - a new API to return completions at a particular offset . Uses the parser directly as the best source of knowledge about what tokens to expect.
- Introduce the `WordKinds` flags to define the different kinds of word tokens that the parser can expect during parsing. (See `compiler/qsc_parse/src/completion/word_kinds.rs` for longer description)
- See `compiler/qsc_parse/src/completion/collector.rs` for a description of how this all works.
- Sprinkle `expect()`s at all possible locations in the parser where we expect a word.

### Error recovery - AST changes

- Introduce `PathKind::Ok` and `PathKind::Err` .
    - This new approach allows for a valid AST node to exist even when path parsing fails for a partial path. This is good for syntax such as "`open Std.`" where we need:
        - The identifier nodes preceding `.` to be valid, so we can use them to look up possible completions.
        - The `open` statement to also be a valid node, so we can tell that the context requires a namespace (and not, say, a type name).
    - Most existing code will continue to only process valid paths. 
- Remove `Idents` struct and replace it with a trait implemented by various types such as `Path` and `Ident` slices. This should make the convenience methods easier to use without cloning in/out of the different types we need.

### Error recovery - parser changes

- Recovering struct parsing, so that e.g.  `new Foo.`  is still parsed as a valid struct expression node, allowing us to access the `Foo` ident from the AST
- Recovering `open` statement parsing so that `open Foo.` can be recognized as an open item in the AST.

### Language service

In `language_service` is essentially a complete rewrite of `completions.rs`, so I suggest reviewing the new code only and ditching the diff.

### Other parser changes

- Move implicit namespace parsing a bit so that it works better with the completion  mechanism.
    - This also fixes the  spurious "expected item, found eof" error when document only contains one token
- Delete the dead path  `path_field_op` 
- Add some missing keywords to item parsing recovery (`export`, `import`)

### Other notes

- Locals completions exclude `import`s from locals (they conflict with globals already added to completion) (fix in `resolve.rs`)
- `language-configuration.json` : and removed wonky word pattern in playground and VS Code because we don't need it anymore now that we have proper `.` completions! Also update trigger characters.
- Only include the samples snippets in empty documents, because now with the more concise completion list, it gets quite intrusive for them to show up at every word when typing code.
- Update integration test to account for the fact that completions are more accurate now.

### Not in this PR (coming next)

- Struct and UDT field access completions.

### DIFF
--- a/compiler/qsc_ast/src/ast.rs
+++ b/compiler/qsc_ast/src/ast.rs
@@ -1423,10 +1423,14 @@ impl Default for PathKind {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct IncompletePath {
     /// The whole span of the incomplete path,
-    /// including the final `.` and any whitespace that follows it.
+    /// including the final `.` and any whitespace or keyword
+    /// that follows it.
     pub span: Span,
     /// Any segments that were successfully parsed before the final `.`.
     pub segments: Box<[Ident]>,
+    /// Whether a keyword exists after the final `.`.
+    /// This keyword can be presumed to be a partially typed identifier.
+    pub keyword: bool,
 }
 
 impl Display for PathKind {
--- a/compiler/qsc_ast/src/ast.rs
+++ b/compiler/qsc_ast/src/ast.rs
@@ -273,7 +273,7 @@ pub enum ItemKind {
     #[default]
     Err,
     /// An `open` item for a namespace with an optional alias.
-    Open(PathResult, Option<Box<Ident>>),
+    Open(PathKind, Option<Box<Ident>>),
     /// A `newtype` declaration.
     Ty(Box<Ident>, Box<TyDef>),
     /// A `struct` declaration.
@@ -672,7 +672,7 @@ pub enum TyKind {
     /// A type wrapped in parentheses.
     Paren(Box<Ty>),
     /// A named type.
-    Path(PathResult),
+    Path(PathKind),
     /// A type parameter.
     Param(Box<Ident>),
     /// A tuple type.
@@ -884,7 +884,7 @@ pub enum ExprKind {
     /// Parentheses: `(a)`.
     Paren(Box<Expr>),
     /// A path: `a` or `a.b`.
-    Path(PathResult),
+    Path(PathKind),
     /// A range: `start..step..end`, `start..end`, `start...`, `...end`, or `...`.
     Range(Option<Box<Expr>>, Option<Box<Expr>>, Option<Box<Expr>>),
     /// A repeat-until loop with an optional fixup: `repeat { ... } until a fixup { ... }`.
@@ -892,7 +892,7 @@ pub enum ExprKind {
     /// A return: `return a`.
     Return(Box<Expr>),
     /// A struct constructor.
-    Struct(PathResult, Option<Box<Expr>>, Box<[Box<FieldAssign>]>),
+    Struct(PathKind, Option<Box<Expr>>, Box<[Box<FieldAssign>]>),
     /// A ternary operator.
     TernOp(TernOp, Box<Expr>, Box<Expr>, Box<Expr>),
     /// A tuple: `(a, b, c)`.
@@ -1147,7 +1147,7 @@ fn display_repeat(
 
 fn display_struct(
     mut indent: Indented<Formatter>,
-    name: &PathResult,
+    name: &PathKind,
     copy: &Option<Box<Expr>>,
     fields: &[Box<FieldAssign>],
 ) -> fmt::Result {
@@ -1404,7 +1404,7 @@ impl Display for QubitInitKind {
 
 /// A path that may or may not have been successfully parsed.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub enum PathResult {
+pub enum PathKind {
     /// A successfully parsed path.
     Ok(Box<Path>),
 
@@ -1412,9 +1412,9 @@ pub enum PathResult {
     Err(Option<Box<IncompletePath>>),
 }
 
-impl Default for PathResult {
+impl Default for PathKind {
     fn default() -> Self {
-        PathResult::Err(None)
+        PathKind::Err(None)
     }
 }
 
@@ -1429,11 +1429,11 @@ pub struct IncompletePath {
     pub segments: Box<[Ident]>,
 }
 
-impl Display for PathResult {
+impl Display for PathKind {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
-            PathResult::Ok(path) => write!(f, "{path}")?,
-            PathResult::Err(Some(incomplete_path)) => {
+            PathKind::Ok(path) => write!(f, "{path}")?,
+            PathKind::Err(Some(incomplete_path)) => {
                 let mut indent = set_indentation(indented(f), 0);
                 write!(indent, "Err IncompletePath {}:", incomplete_path.span)?;
                 indent = set_indentation(indent, 1);
@@ -1441,7 +1441,7 @@ impl Display for PathResult {
                     write!(indent, "\n{part}")?;
                 }
             }
-            PathResult::Err(None) => write!(f, "Err",)?,
+            PathKind::Err(None) => write!(f, "Err",)?,
         }
         Ok(())
     }
@@ -1905,7 +1905,7 @@ pub struct ImportOrExportItem {
     /// The span of the import path including the glob and alias, if any.
     pub span: Span,
     /// The path to the item being exported.
-    pub path: PathResult,
+    pub path: PathKind,
     /// An optional alias for the item being exported.
     pub alias: Option<Ident>,
     /// Whether this is a glob import/export.
@@ -1942,7 +1942,7 @@ impl ImportOrExportItem {
         match &self.alias {
             Some(_) => self.alias.as_ref(),
             None => {
-                if let PathResult::Ok(path) = &self.path {
+                if let PathKind::Ok(path) = &self.path {
                     Some(&path.name)
                 } else {
                     None
--- a/compiler/qsc_ast/src/ast.rs
+++ b/compiler/qsc_ast/src/ast.rs
@@ -1940,7 +1940,7 @@ impl WithSpan for ImportOrExportItem {
 
 impl ImportOrExportItem {
     /// Returns the alias ident, if any, or the name from the path if no alias is present.
-    /// Returns `None` if this the path has an error.
+    /// Returns `None` if the path has an error.
     #[must_use]
     pub fn name(&self) -> Option<&Ident> {
         match &self.alias {
--- a/compiler/qsc_ast/src/ast.rs
+++ b/compiler/qsc_ast/src/ast.rs
@@ -12,6 +12,7 @@ use std::{
     cmp::Ordering,
     fmt::{self, Display, Formatter, Write},
     hash::{Hash, Hasher},
+    iter::once,
     rc::Rc,
 };
 
@@ -164,7 +165,7 @@ pub struct Namespace {
     /// The documentation.
     pub doc: Rc<str>,
     /// The namespace name.
-    pub name: Idents,
+    pub name: Box<[Ident]>,
     /// The items in the namespace.
     pub items: Box<[Box<Item>]>,
 }
@@ -182,11 +183,21 @@ impl Namespace {
 impl Display for Namespace {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         let mut indent = set_indentation(indented(f), 0);
-        write!(
-            indent,
-            "Namespace {} {} ({}):",
-            self.id, self.span, self.name
-        )?;
+        write!(indent, "Namespace {} {} (", self.id, self.span)?;
+
+        let mut buf = Vec::with_capacity(self.name.len());
+
+        for ident in &self.name {
+            buf.push(format!("{ident}"));
+        }
+        if buf.len() > 1 {
+            // use square brackets only if there are more than one ident
+            write!(indent, "[{}]", buf.join(", "))?;
+        } else {
+            write!(indent, "{}", buf[0])?;
+        }
+
+        write!(indent, "):",)?;
         indent = set_indentation(indent, 1);
 
         if !self.doc.is_empty() {
@@ -262,7 +273,7 @@ pub enum ItemKind {
     #[default]
     Err,
     /// An `open` item for a namespace with an optional alias.
-    Open(Idents, Option<Box<Ident>>),
+    Open(PathResult, Option<Box<Ident>>),
     /// A `newtype` declaration.
     Ty(Box<Ident>, Box<TyDef>),
     /// A `struct` declaration.
@@ -661,7 +672,7 @@ pub enum TyKind {
     /// A type wrapped in parentheses.
     Paren(Box<Ty>),
     /// A named type.
-    Path(Box<Path>),
+    Path(PathResult),
     /// A type parameter.
     Param(Box<Ident>),
     /// A tuple type.
@@ -873,7 +884,7 @@ pub enum ExprKind {
     /// Parentheses: `(a)`.
     Paren(Box<Expr>),
     /// A path: `a` or `a.b`.
-    Path(Box<Path>),
+    Path(PathResult),
     /// A range: `start..step..end`, `start..end`, `start...`, `...end`, or `...`.
     Range(Option<Box<Expr>>, Option<Box<Expr>>, Option<Box<Expr>>),
     /// A repeat-until loop with an optional fixup: `repeat { ... } until a fixup { ... }`.
@@ -881,7 +892,7 @@ pub enum ExprKind {
     /// A return: `return a`.
     Return(Box<Expr>),
     /// A struct constructor.
-    Struct(Box<Path>, Option<Box<Expr>>, Box<[Box<FieldAssign>]>),
+    Struct(PathResult, Option<Box<Expr>>, Box<[Box<FieldAssign>]>),
     /// A ternary operator.
     TernOp(TernOp, Box<Expr>, Box<Expr>, Box<Expr>),
     /// A tuple: `(a, b, c)`.
@@ -1136,7 +1147,7 @@ fn display_repeat(
 
 fn display_struct(
     mut indent: Indented<Formatter>,
-    name: &Path,
+    name: &PathResult,
     copy: &Option<Box<Expr>>,
     fields: &[Box<FieldAssign>],
 ) -> fmt::Result {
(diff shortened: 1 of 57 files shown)


compiler/qsc_ast/src/ast.rs
```diff
@@ -1947,36 +1930,24 @@ impl Display for ImportOrExportItem {
...
-            is_glob: self.is_glob,
-        }
+        Self { span, ..self }
     }
 }
 
 impl ImportOrExportItem {
-    /// Returns the span of the export item. This includes the path and , if any exists, the alias.
+    /// Returns the alias ident, if any, or the name from the path if no alias is present.
+    /// Returns `None` if this the path has an error.
```
swernli, line 1939: nitpick: extra word

```suggestion
    /// Returns `None` if the path has an error.
```

minestarks:
> _Change in memory usage detected by benchmark._
> 
> ## Memory Report for [6fbbec5](https://github.com/microsoft/qsharp/commit/6fbbec5027789cfc5e384b13e4eb61f410f50b52)
> Test	This Branch	On Main	Difference
> compile core + standard lib	17940700 bytes	17932132 bytes	8568 bytes

I ran some tests to narrow it down. This increase is attributable entirely to the changes in the AST. A bunch of `Path` or `Idents` fields have been replaced with `PathKind`, a `Span` field has been added to `ImportOrExportItem`. It adds up to a net increase in memory.

I can't think of a way to collapse the AST any further, we are storing more information in there after all. So I think the 0.04% increase in memory usage is a worthwhile tradeoff for the feature.

language_service/src/completion.rs
```diff
@@ -1,809 +1,334 @@
...
+        CompilationKind::OpenProject {
+            package_graph_sources,
+        } => possible_words_at_offset_in_source(
+            source_contents,
+            Some(source_name_relative),
+            package_graph_sources.root.language_features,
+            cursor_offset,
+        ),
+        CompilationKind::Notebook { project } => possible_words_at_offset_in_fragments(
+            source_contents,
```
swernli, line 111: How do source contents work for notebook cells? I was curious about this and tried playing around with it, ended up seeing some behaviors I didn't expect, so wanted to confirm whether there are known bugs/limitations here or if this is something that bears investigating. I'll follow up with screen shots and more explanation...

swernli: When I try out the language service with these changes, completions in a file work as I'd expect. I get completions for local variables that are in-scope at my current line (the `x` from the line above):
image
or for locally defined callables:
image

But in notebook cells, I don't see these elements if they come from top-level statements (it does work as expected when inside a callable context). So the local variables don't show `x` or `IntOne` from the lines above my cursor:
image
image

Oddly enough, the locals and functions DO show up under two different circumstances. If instead of triggering completions with ctrl+space I just start typing, they show up:
image
image
OR if I put the definitions in a previous cell, then they show up even if I trigger completions before I start typing:
image

I don't think this is a major issue and would probably be fine as a follow up investigation/bug-fix, but it does tell me that there is some difference between the response of completions when triggered manually vs when typing specifically for notebook cells that I don't really understand.

language_service/src/completion/global_items.rs
minestarks, line 1: Hmm, just discovered this file will have to undergo some rewrites. I'd suggest skipping most of it for now. The `pub` structs / methods should stay the same though, so the rest of the PR should be uneffected.

@swernli FYI since I know you've started looking.

After talking to @sezna about namespaces and exports , there's a much simpler way to gather the names under a particular namespace than what I did here. I'll try that approach and post an update here when I'm done.

language_service/src/tests.rs
```diff
@@ -210,18 +210,19 @@ async fn completions_requested_after_document_load() {
...
                 line: 0,
-                column: 76
-            }
+                column: 92,
+            },
         )
         .items
         .len(),
-        13
     );
```
idavis, line 226: Would it aid debugging test failures to compare the lists instead of the number of items in the lists?

playground/src/main.tsx
```diff
@@ -310,7 +304,7 @@ function registerMonacoLanguageServiceProviders(
         }),
       };
     },
-    triggerCharacters: ["@"], // for attribute completions
+    triggerCharacters: ["@", "."],
```
idavis, line 307: Should we add a note that we need to keep this list in sync with the one in `extension.ts`?

language_service/src/test_utils.rs
```diff
@@ -10,12 +10,49 @@ use qsc::{
...
     line_column::{Encoding, Position, Range},
     location::Location,
+    packages::prepare_package_store,
     target::Profile,
     LanguageFeatures, PackageStore, PackageType, SourceMap, Span,
 };
 use qsc_project::{PackageGraphSources, PackageInfo};
 use rustc_hash::FxHashMap;
 
+const FAKE_STDLIB_CONTENTS: &str = r#"
```
idavis, line 20: Having std lib in the name is a little consfusing to me. Is this just some lib, or is it supposed to be an alternative std lib?

compiler/qsc_doc_gen/src/display.rs
```diff
@@ -540,15 +540,16 @@ impl<'a> Display for FunctorExpr<'a> {
...
-impl<'a> Display for AstPath<'a> {
+impl<'a> Display for AstPathKind<'a> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-        match self.path.segments.as_ref() {
-            Some(parts) => write!(f, "{parts}.{}", self.path.name.name),
-            None => write!(f, "{}", self.path.name.name),
+        if let ast::PathKind::Ok(path) = self.path {
+            write!(f, "{}", path.full_name())
+        } else {
+            write!(f, "?")
```
idavis, line 552: Should we delegate this `Display` impl to `IncompletePath` which can render the segments with a `?`

language_service/src/references/tests.rs
```diff
@@ -93,12 +93,12 @@ fn std_callable_ref() {
...
-                            line: 1,
-                            column: 26,
+                            line: 2,
+                            column: 18,
                         },
                         end: Position {
-                            line: 1,
-                            column: 30,
+                            line: 2,
+                            column: 22,
```
idavis, line 101: Why are the lines and columns changing in these tests?

language_service/src/completion/global_items.rs
minestarks: OK, in the interest of unblocking the PR, I'm going to postpone the changes to a new PR. This is good to go :)

Just be aware that **reexports** are known to not be working. i.e. you won't get reexported namespaces/items showing up under the correct namespace for dot completions.

In practice, this only seems to matter when you're trying to do completions under "legacy" namespaces like `Microsoft.Quantum.Diagnostics.|` because those legacy names are actually reexports. Otherwise, I haven't found any Q# code actually using reexports yet.

language_service/src/completion/path_context.rs
```diff
@@ -0,0 +1,251 @@
...
+            visit::walk_path(self, path);
+        }
+    }
+
+    fn visit_path_kind(&mut self, path: &'a qsc::ast::PathKind) {
+        let span = match path {
+            qsc::ast::PathKind::Ok(path) => &path.span,
+            qsc::ast::PathKind::Err(Some(incomplete_path)) => &incomplete_path.span,
+            qsc::ast::PathKind::Err(None) => return,
+        };
```
idavis, line 238: nit: clean up with use to match rest of file

language_service/src/completion/path_context.rs
```diff
@@ -0,0 +1,251 @@
...
+            self.context = Some(PathKind::Struct);
+        }
+    }
+
+    fn visit_path_kind(&mut self, path: &'a qsc::ast::PathKind) {
+        self.qualifier = match path {
+            qsc::ast::PathKind::Ok(path) => path.segments.as_ref().map(AsRef::as_ref),
+            qsc::ast::PathKind::Err(Some(incomplete_path)) => Some(&incomplete_path.segments),
+            qsc::ast::PathKind::Err(None) => None,
+        };
```
idavis, line 71: nit: clean up with use to match rest of file

language_service/src/completion.rs
```diff
@@ -1,809 +1,334 @@
...
-        }
-        None => before_offset,
-    }
-    .to_string();
-    if !indent.starts_with('\n') {
-        indent.insert(0, '\n');
+        .relative_sources()
+        .find(|s| s.offset == source.offset)
+        .expect("source should exist in the user source map")
+        .name;
```
idavis, line 55: Would it make sense to expose something like `as_relative_source(source: Source)` rather than iterating and cloning all of the sources in the map?
```
let source_name_relative = compilation
        .user_unit()
        .sources
        .as_relative_source(source).name
```

playground/src/main.tsx
```diff
@@ -248,12 +248,6 @@ async function loaded() {
 function registerMonacoLanguageServiceProviders(
   languageService: ILanguageService,
 ) {
-  monaco.languages.setLanguageConfiguration("qsharp", {
```
ScottCarda-MS, line 251: I wasn't expecting this to get dropped. Are we not using this anymore?

ScottCarda-MS (commented):
partially reviewed, will continue to review

minestarks:
@idavis found a bug: 

Trying to complete

```qsharp
import QIR.in;
```

causes a panic since `in` is a keyword and that's a parse tree I did not expect to have to handle....

Working on the fix.

language_service/src/test_utils.rs
minestarks: Yes, it's supposed to replace the stdlib in most language service tests, since it's more minimal (faster to build) and I can easily add any edge case I want to cover.

language_service/src/completion/path_context.rs
minestarks: Sad story: there are two different structs named `PathKind` in this file, and the qualifiers here are to disambiguate. Maybe I'll use an alias.

compiler/qsc_ast/src/ast.rs
```diff
@@ -1391,57 +1402,65 @@ impl Display for QubitInitKind {
...
-            id: NodeId::default(),
-            span,
-            segments,
-            name: name.into(),
+impl Display for PathKind {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            PathKind::Ok(path) => write!(f, "{path}")?,
+            PathKind::Err(Some(incomplete_path)) => {
+                let mut indent = set_indentation(indented(f), 0);
```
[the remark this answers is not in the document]
minestarks: Sorry, I don't understand. What's a normal indentation?

minestarks:
> @idavis found a bug:
> 
> Trying to complete
> 
> ```qsharp
> import QIR.in;
> ```
> 
> causes a panic since `in` is a keyword and that's a parse tree I did not expect to have to handle....
> 
> Working on the fix.

Fixed with my latest commit.

playground/src/main.tsx
minestarks: Nope! Good riddance. Called out in the PR description:

> removed wonky word pattern in playground and VS Code because we don't need it anymore now that we have proper `.` completions! Also update trigger characters.

compiler/qsc_ast/src/ast.rs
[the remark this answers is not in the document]
ScottCarda-MS: My bad, I misread what this was doing. This seems like the usual way of doing indentation.

compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -3080,15 +3080,15 @@ fn local_open() {
             namespace B { function Bar() : () {} }
         "},
         "",
-        &expect![[r#"
+        &expect![[r##"
```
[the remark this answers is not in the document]
minestarks: OMG, I took so many passes over this PR trying to revert the sneaky hash changes. I _agree_ with you, but in this file, the warning is suppressed (with `#![allow(clippy::needless_raw_string_hashes)]`) , and the existing test cases are full of `##`. So I thought I should stick to that convention. Now what?! 😆

compiler/qsc_ast/src/mut_visit.rs
```diff
@@ -384,12 +389,26 @@ pub fn walk_path(vis: &mut impl MutVisitor, path: &mut Path) {
...
+    }
+}
+
 pub fn walk_ident(vis: &mut impl MutVisitor, ident: &mut Ident) {
     vis.visit_span(&mut ident.span);
 }
 
-pub fn walk_idents(vis: &mut impl MutVisitor, ident: &mut crate::ast::Idents) {
-    for ref mut ident in &mut *ident.0 {
+pub fn walk_idents(vis: &mut impl MutVisitor, idents: &mut [Ident]) {
```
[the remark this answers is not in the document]
minestarks: Great question, I went back and forth on this for a while. Ultimately, I decided to keep `visit_idents` to avoid updating the current usages and bloating this PR even further. 

This came up in a conversation with Ian as well, and my response was basically:

> In the AST, many of choices came down to "how do I avoid touching a billion lines of code", and this was one of them 

I think I'll sleep on it and see on Monday if I want to open a separate refactoring PR, or push even more changes to this one...

compiler/qsc_parse/src/completion/collector.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub fn expect(&mut self, expected: WordKinds) {
+        match self.state {
+            State::AtCursor => self.collected.extend(expected),
+            State::BeforeCursor | State::End => {}
+        }
+    }
+
+    /// The parser has advanced. Update state.
+    pub fn did_advance(&mut self, next_token: &mut Token, scanner_offset: u32) {
+        match self.state {
```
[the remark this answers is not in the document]
minestarks: This is semantically an "event handler", as in, it's informing the callee, that advancing _did_ occur. 

I got used to this naming convention in LSP / VS Code APIS (e.g. `didDeleteFile` ) but I guess it's not coming across :D 

Any other suggestions? `on_advance()`? `on_advanced()`? `advanced()` ?

language_service/src/completion.rs
minestarks: Yes, I like that idea.

language_service/src/references/tests.rs
minestarks: Because of my changes in `test_utils.rs` - I think I changed some whitespace in the "Fake stdlib" while moving strings around.

language_service/src/completion.rs
minestarks: This turned out to be a preexisting bug. I'll investigate.

minestarks: Completions were a distraction, actually - having a broken expression at the very end of the cell breaks the parsing of the entire cell, there's no AST at all! Notice none of the editor features (hover, go to def etc) work for _any_ code above that expression.

I'll just file a separate bug for this.

minestarks: Tracked in #1960

compiler/qsc_doc_gen/src/display.rs
minestarks: I don't think this would work in this case. The AST structs' `Display` impls all print in that "AST-ish" format, like:

```
Err IncompletePath [157-166]:
    Ident 13 [157-160] "Foo"
    Ident 14 [161-164] "Bar"
```

like what you see in the "AST" tab in the playground. These are not meant to be used when pretty printing to the user. (This line actually had a bug where it was exposing the AST-ish format to the user, which I fixed.

We've had to differentiate these impls from the user visible "pretty printing" impls, which live here. That's why there are a bunch of structs in this file that wrap the AST/HIR structs only to implement `Display`.

language_service/src/completion.rs
swernli: Oh wow, I didn't realize that. Thanks for investigating!

language_service/src/completion/tests.rs
```diff
@@ -477,101 +621,75 @@ fn in_block_with_alias() {
...
+            Alias.↘
         }
-        export Bar;
-    }
-    namespace Other {
-        operation Foo() : Unit {}
-        export Foo;
     }"#},
-        &["Foo"],
+        &["Fake", "Alias.Fake", "Library", "Alias.Library"],
```
ScottCarda-MS, line 633: Might be a bit safer to have it also check against function names that aren't in Alias, like `Foo`, and names that are only in Alias but not elsewhere. In the code as-written, it is a bit ambiguous to me if it is validating the `Fake` or the `Alias.Fake`.

compiler/qsc_frontend/src/typeck/tests.rs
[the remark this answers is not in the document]
idavis: @swernli Can you help out here?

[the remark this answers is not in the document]
ScottCarda-MS: In the past, I've had to do a search-and-replace over "##".

minestarks:
> Fixed with my latest commit.

Okay, I regressed things pretty badly with this fix, as @ScottCarda-MS discovered through testing (thanks!). 

The following piece of code should produce a mostly valid AST (save for the error on line 3). 

```qsharp
namespace Sample {
    operation BellStates() : (Result, Result)[] {
        Std.
        let bellStateTuples = [];

        for (label, prepare) in bellStateTuples {}
        return measurements;
    }

    /// doc comment
    operation PreparePhiPlus() : Unit {
    }
}
```

With my latest commit, there's no AST at all.

cc reviewers @idavis  @swernli . I'm going to have to go back to the drawing board with the "keyword after dot" fix that handles `QIR.in` and similar paths. Expect more parser changes :/

compiler/qsc_frontend/src/typeck/tests.rs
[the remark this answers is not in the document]
minestarks: OK, I just reverted this line to the way it was.

language_service/src/tests.rs
minestarks: Not exactly, since the contents of the list don't matter for the purposes of this test. But your comment was the push I needed to change this test into something a bit smarter instead of this opaque number that changes every time we touch something.

minestarks:
I pushed what I think is my final update. The parser change to fix the last issue ended up being minimal!

Please sign off if no more concerns.

ScottCarda-MS (approved):
Discussed offline that the merge conflict is minor. Everything else looks good, and I tested out the feature manually.

vscode/src/completion.ts
```diff
@@ -82,6 +82,8 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
       });
       return item;
     });
-    return results.concat(this.samples);
+
+    // Include the samples list for empty documents
+    return document.lineCount === 0 ? results.concat(this.samples) : results;
```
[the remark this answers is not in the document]
minestarks: Done.

playground/src/main.tsx
minestarks: Done.

PR #1952, opened 2024-10-08
Add summary lines back into stdlib readmes

sezna:
These were accidentally removed in a previous PR.

### DIFF
--- a/library/std/src/Std/Arrays.qs
+++ b/library/std/src/Std/Arrays.qs
@@ -5,6 +5,7 @@ import Std.Diagnostics.*;
 import Std.Math.*;
 
 /// # Summary
+/// # Summary
 /// Given an array and a predicate that is defined
 /// for the elements of the array, and checks if all elements of the
 /// array satisfy the predicate.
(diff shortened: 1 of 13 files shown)


library/std/src/Std/Arrays.qs
```diff
@@ -4,6 +4,7 @@
 import Std.Diagnostics.*;
 import Std.Math.*;
 
+/// # Summary
```
swernli, line 7: Looks like this one got doubled up.

library/std/src/Std/Logical.qs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 
+/// # Summary
```
swernli, line 5: here too...

library/std/src/Std/Math.qs
```diff
@@ -9,6 +9,7 @@ import Std.Diagnostics.*;
 // Constants PI, E, LogOf2.
 //
 
+/// # Summary
```
swernli, line 12: This may have been an artifact of how the files were created/moved from the originals, in most files the first callable still had summary and it was the rest that got lost.

library/std/src/Std/Math.qs
```diff
@@ -231,68 +249,81 @@ function Min(values : Int[]) : Int {
 // Trigonometric functions
 //
 
+/// # Summary
```
swernli, line 252: Oh, never mind, I guess it's some in the middle too :)

swernli (approved):

minestarks (approved):

PR #1953, opened 2024-10-08
Show no quantum state when debugging code with no qubits

### LINKED ISSUE #1922: When debugging code with no qubits, quantum state in Variables view is reported as |0>
**Describe the bug**

When I debug code that doesn't have qubits allocated, the "Q# Circuit" panel correctly says "No qubits have been allocated", but the Quantum State is shown as |0> which corresponds to one qubit.

**To Reproduce**

Debug any operation that doesn't allocate qubits (for example, calculates classical values before anything quantum). 

**Expected behavior**

No state shown - either no fields under "Quantum State", or a message saying "No qubits have been allocated"

**System information**

- QDK 1.8.0

swernli:
Instead of showing a single bit zero state, this change makes the debugger quantum state view show a message when no qubits are allocated on the system:
image

Fixes #1922

### DIFF
--- a/vscode/src/debugger/session.ts
+++ b/vscode/src/debugger/session.ts
@@ -795,6 +795,22 @@ export class QscDebugSession extends LoggingDebugSession {
             {},
           );
           const state = await this.debugService.captureQuantumState();
+
+          // When there is no quantum state, return a single variable with a message
+          // for the user.
+          if (state.length == 0) {
+            response.body = {
+              variables: [
+                {
+                  name: "None",
+                  value: "No qubits allocated",
+                  variablesReference: 0,
+                },
+              ],
+            };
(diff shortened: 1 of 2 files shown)


minestarks (commented):
Do we need to make the same fix in the `DumpMachine` output in Python and the Playground?

swernli:
> > Do we need to make the same fix in the `DumpMachine` output in Python and the Playground?
> 
> Yeah, I think you are right and it's worth filing follow up issue(s). I'll add those to the backlog!

https://github.com/microsoft/qsharp/issues/1961 for Python and https://github.com/microsoft/qsharp/issues/1962 for the playground.

cesarzc (approved):

PR #1954, opened 2024-10-09
UDT field completions

minestarks:
Builds on #1947 .

Enables:

1. Field access completions, for example:

```qsharp
let x = Complex(1.0, 2.0);
x.| 
```

At the cursor location indicated by `|`, completions should include `Real` since it's a field of `Complex`.

2. Field assignment completions, for example:

```qsharp
struct Foo {
    bar: Int
}

new Foo { | }
```

Here, completions should include `Bar`.

### DIFF
--- a/compiler/qsc_frontend/src/typeck/tests.rs
+++ b/compiler/qsc_frontend/src/typeck/tests.rs
@@ -12,7 +12,7 @@ use expect_test::{expect, Expect};
 use indoc::indoc;
 use qsc_ast::{
     assigner::Assigner as AstAssigner,
-    ast::{Block, Expr, Idents, NodeId, Package, Pat, Path, QubitInit, TopLevelNode},
+    ast::{Block, Expr, Idents, NodeId, Package, Pat, Path, PathKind, QubitInit, TopLevelNode},
     mut_visit::MutVisitor,
     visit::{self, Visitor},
 };
@@ -36,6 +36,16 @@ impl<'a> Visitor<'a> for TyCollector<'a> {
         let ty = self.tys.get(expr.id);
         self.nodes.push((expr.id, expr.span, ty));
         visit::walk_expr(self, expr);
+    }
+
+    fn visit_path_kind(&mut self, path_kind: &'a PathKind) {
+        visit::walk_path_kind(self, path_kind);
+        if let PathKind::Err(Some(incomplete_path)) = path_kind {
+            for part in incomplete_path.segments.iter() {
+                let ty = self.tys.get(part.id);
+                self.nodes.push((part.id, part.span, ty));
+            }
+        }
     }
 
     fn visit_path(&mut self, path: &'a Path) {
@@ -67,7 +77,15 @@ impl<'a> Visitor<'a> for TyCollector<'a> {
 }
 
 fn check(input: &str, entry_expr: &str, expect: &Expect) {
-    let (package, tys, errors) = compile(input, entry_expr);
+    check_with_error_option(input, entry_expr, expect, false);
+}
+
+fn check_allow_parse_errors(input: &str, entry_expr: &str, expect: &Expect) {
+    check_with_error_option(input, entry_expr, expect, true);
+}
+
+fn check_with_error_option(input: &str, entry_expr: &str, expect: &Expect, allow_errors: bool) {
+    let (package, tys, errors) = compile(input, entry_expr, allow_errors);
     let mut collector = TyCollector {
         tys: &tys.terms,
         nodes: Vec::new(),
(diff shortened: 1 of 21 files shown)


compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -4415,3 +4440,236 @@ fn within_block_should_be_unit_error() {
...
+            namespace A {
+                struct B { C : Int }
+                function Foo() : Unit {
+                    let b = new B { C = 5 };
+                    b.C;
+                }
+            }
+        "},
+        "",
+        &expect![[r##"
```
ScottCarda-MS, line 4457: nit: These double hashes are creeping up again 🙁

compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -4415,3 +4440,236 @@ fn within_block_should_be_unit_error() {
...
+}
+
+#[test]
+fn expr_incomplete_field_access() {
+    check_allow_parse_errors(
+        indoc! {"
+            namespace A {
+                struct B { C : Int }
+                function Foo() : Unit {
+                    (new B { C = 5 }).;
```
ScottCarda-MS, line 4529: What is the behavior when you don't end this line in a semicolon? Does it complain that the expected return type of `Unit` doesn't match `?`?

minestarks: Good question, I should add a test case for that since that would emulate real-life typing more closely. What I'm looking for is for the type of the `(new B { C = 5 })` to be inferred correctly, because that will help us gather the field names for completions.

Though these tests don't validate typecheck errors, I can see from the output: Yes, the overall type of this block does become `?` to match the type of the broken expression. Which doesn't match `Unit`.

compiler/qsc_frontend/src/typeck/tests.rs
minestarks: @swernli can you clarify what we should be doing with new tests w/r/t double hashes?

From the `#![allow(clippy::needless_raw_string_hashes)]` suppression at the top of this file, and the `##`s in many other test cases, I'm led to believe we're supposed to _keep_ the double hashes as `expect-test` generates them.

swernli: I don't think we should worry about these. If you manually remove them, the next time the expect-test updates they will get added back. The point of suppressing the warning is not fight that quirk of behavior, so I think we should ignore these differences. The only exception is when it grows each time (two become three, next run makes three become four), but I think that bug in expect-test was fixed so the problem doesn't occur anymore.

compiler/qsc_parse/src/expr.rs
```diff
@@ -710,9 +711,20 @@ fn lambda_op(s: &mut ParserContext, input: Expr, kind: CallableKind) -> Result<B
     Ok(Box::new(ExprKind::Lambda(kind, input, output)))
 }
 
-fn field_op(s: &mut ParserContext, lhs: Box<Expr>) -> Result<Box<ExprKind>> {
+#[allow(clippy::unnecessary_wraps)]
```
sezna, line 714: This is from the `pedantic` profile of clippy, and I myself turn off a lot of pedantic lints too. Is it worth using the normal clippy profile?

Some other lints that have been bothering me:
- disallowing `Default::default()`
- needless raw string hashes

There are a lot of other profiles we can explore in clippy, perhaps now is the time?

sezna: For example, this clippy lint here is actually invalid because you need the function to match a certain signature. These kinds of lints are generating more work, not less.

compiler/qsc_parse/src/expr.rs
```diff
@@ -710,9 +711,20 @@ fn lambda_op(s: &mut ParserContext, input: Expr, kind: CallableKind) -> Result<B
...
+        lhs,
+        match ident(s) {
+            Ok(i) => FieldAccess::Ok(i),
+            Err(e) => {
+                s.push_error(e);
+                FieldAccess::Err
+            }
+        },
+    );
+    Ok(Box::new(expr))
```
sezna, line 727: did we ever figure out why all of our parsers are boxed? or are you just matching conventions?

sezna (approved):
really appreciate the extra documentation and tests

compiler/qsc_parse/src/expr.rs
minestarks: Mostly matching convention.  @swernli and I had a conversation about this a while back. I'll try to give a summary:

- Early on, during some perf optimization work, we discovered that keeping struct sizes small, by boxing, saved a lot of time spent in stack allocation and moving things around.
- The strategy is mentioned in : https://nnethercote.github.io/perf-book/heap-allocations.html#box
- However, today, we hardly (if ever) move the AST around. Most of the work during compilation, as you probably already know, is done using a reference to _the_ AST.
- So we think these early perf decisions may not be relevant anymore, but we have not validated (or invalidated) them with current code. 

For my changes, in my attempt to please everyone, I used the rule of thumb "no structs in the AST larger than 24 bytes", so I boxed things as I hit that limit. The "24" number is purely based on vibes and not hard numbers.

idavis (approved):

ScottCarda-MS (approved):

PR #1963, opened 2024-10-14
Fix to partial evaluation generating branch instructions on constant conditions

cesarzc:
RCA can categorize an expression as dynamic even if it is in fact purely classical. This can happen in cases where a data structure such an array, tuple or UDT contains a mix of static and dynamic values. In such instances, RCA identifies all the contents of the data structure as dynamic even if some values are static. This causes partial evaluation to generate branch instructions on constant conditions under certain circumstances. This change fixes this issue.

### DIFF
--- a/compiler/qsc_partial_eval/src/tests/branching.rs
+++ b/compiler/qsc_partial_eval/src/tests/branching.rs
@@ -1313,17 +1313,17 @@ fn if_else_expression_with_result_literal_fails() {
 }
 
 #[test]
-fn if_comparison_with_mixed_static_dynamic_array() {
+fn if_expression_with_classical_operand_from_hybrid_results_array_comparing_to_literal_zero() {
     let program = get_rir_program(indoc! {r#"
         @EntryPoint()
         operation Main() : Result[] {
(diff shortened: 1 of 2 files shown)


compiler/qsc_partial_eval/src/tests/branching.rs
```diff
@@ -1311,3 +1311,179 @@ fn if_else_expression_with_result_literal_fails() {
         ]],
     );
 }
+
+#[test]
+fn if_expression_with_classical_operand_from_hybrid_results_array_comparing_to_literal_zero() {
```
swernli, line 1316: I might be good to add one more test case that shows that arrays of mixed dynamic and static integers works as expected (I tried and it does, as that code path already handled the mixed content case). Just for completeness.

swernli (approved):

compiler/qsc_partial_eval/src/tests/branching.rs
cesarzc: Good point, added the missing tests for Booleans and integers.

idavis (approved):

PR #1965, opened 2024-10-15
Minor lint fix for Rust 1.81

swernli:
I upgraded locally and clippy was mad about this line, so I applied the fix.


minestarks (approved):
Oh I kept running into this locally but not knowing what's going on with rust versions. Thanks!

orpuente-MS (approved):

PR #1967, opened 2024-10-16
Add custom measurement operations to Q#

orpuente-MS:
This PR adds the ability to declare custom measurements in Q#. Here is an example:
```qsharp
@Measurement()
operation __quantum__qis__mx__body(target : Qubit) : Result {
  body intrinsic;
}
```

The generated QIR will be,
```llvm-ir
declare void @__quantum__qis__mx__body(%Qubit*, %Result*) #1
```

# Design Notes

1. I choose the `@Measurement()` attribute over a new `measurement` keyword to avoid putting extra syntax into the language.
2. A custom measurement must be a `body intrinsic` or a `@SimulatableIntrinsic()`.
3. A measurement must only take `Qubits` as inputs and have `Results` as outputs.

# Notes to reviewers

This PR consists of the following steps:
 1. Adding a `@Measurement()` attribute to the language and piping it down to FIR.
 3. Make the generated QIR have the `#1` attribute indicating an irreversible operation.
 4. Treating a measurement as a dynamic operation in RCA.
 5. Handling custom measurements correctly during partial evaluation: moving the output Results to result registers in the input.
 6. Handling undesired semantics as errors using an HIR pass.
 7. Adding the `@Measurement()` attribute to the completions list.
 8. Unit tests in codegen and in the HIR pass.
 
 # Next steps
 
 After this PR is merged we can add the `@Measurement()` attribute to the measurement operations in the standard library and remove the hard coded logic we have for them in the compiler.


language_service/src/completion.rs
```diff
@@ -289,7 +289,7 @@ impl CompletionListBuilder {
     }
 
     fn push_attributes(&mut self) {
-        static ATTRIBUTES: [&str; 2] = ["@EntryPoint()", "@Config()"];
+        static ATTRIBUTES: [&str; 3] = ["@EntryPoint()", "@Config()", "@Measurement()"];
```
minestarks, line 292: Heads up, this code is changing in my PR. When you hit the merge conflict, I suggest just reverting this and searching again for the attribute list instead of trying to resolve it as a merge conflict...

minestarks:
@orpuente-MS  Do we really need an explicit `@Measurement` attribute if we simply define a "measurement intrinsic" as an "intrinsic that returns a `Result`"?

swernli:
> @orpuente-MS Do we really need an explicit `@Measurement` attribute if we simply define a "measurement intrinsic" as an "intrinsic that returns a `Result`"?

I was having the same thought, and I think it's worth exploring. With the attribute, you must include the attribute for something to get treated as a `CallableKind::Measurement`, and then the operation must adhere to these properties:

- Must only take arguments of type `Qubit`
- Must only return type `Result` or tuple of `Result`
- Must be intrinsic or `SimulatableIntrinsic`

If instead these restrictions became checks during lowering from AST to HIR, then any operation that meets these requirements could be lowered into `CallableKind::Measurement`. Then the user wouldn't have to remember to add the attribute AND only valid callables would have the corresponding kind. It's possible there is a corner case this doesn't cover, but it seems initially like this could work well.

orpuente-MS:
My one concern is that if we do that, the compiler won't help the user with friendly errors if the user doesn't use the combination of inputs & outputs when trying to write a measurement. For example, if the user writes this,

```qsharp
operation __quantum__qis__mx__body(target : Qubit, invalid_arg: Int) : (Int, String) {
  body intrinsic;
}
```

and they intended to write an operation, we have no way to guide them towards the write path by telling them:
```
Hey if you intended to write an operation you should satisfy these conditions...
```

because this could be a non-measurement intrinsic

swernli:
That's a good point, and I think we can focus in on that problem statement: if someone writes something that is valid Q# in general but we know it will fail QIR codegen because it doesn't match the set of intrinsic qualities we support, how do we tell them? I have some thoughts on what that feedback could look like, so perhaps you, @cesarzc and I could brainstorm a little.

Overall, I do think a lot of the infrastructure in this PR could be used to get toward this goal, even if the specific detail of an `@Measurement` attribute itself gets dropped.

orpuente-MS:
Also, I don't like the idea of having a special kind of operation (i.e. measurements) hidden behind a specific set of input & outputs. I think we should be explicit about measurement declarations. Initially I wanted to add a `measurement` keyword so we could have:
```qsharp
measurement Mx(target: Qubit) : Result {
  body intrinsic;
}
```

I like this because it feels like measurements are actually part of the language, which is nice because they are treated differently in literature and result in different codegen.

The reason I decided not to go this route is because it adds a new keyword to the language, and that could break existing user code.

orpuente-MS:
I forgot to mention that @Measurements() can also return `Unit`. The reason is that in this way we can express `Reset` or collapsing operations in general as measurements.

swernli:
Yes, `Reset` is an interesting special case, that I think may need to remain a special case for a while. It should definitely have the "irreversible" QIR attribute and today that is missing. But it also doesn't produce a measurement result anywhere, so it doesn't get the same treatment as others (having its signature changed, getting deferred, etc). Also, when you consider how we do the deferral, there is a difference between measurement and reset (or to put it another way, non-destructive vs destructive measurement) in terms of whether the new, replacement qubit needs to have the corresponding state as the original.

I think you've identified another issue here: how do we (or the user) express that an operation is destructive to the state? As you point out, maybe it's that scenario that requires the attribute because it can't be differentiated based on signature alone.

compiler/qsc_passes/src/measurement.rs
```diff
@@ -0,0 +1,121 @@
...
+};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
+pub enum Error {
+    #[error("a measurement should take at least one Qubit as argument")]
+    #[diagnostic(code("Qsc.Measurement.NoArguments"))]
+    NoArguments(#[label] Span),
+
+    #[error("a measurement should only take Qubits as arguments")]
```
billti, line 22: Is this inherently true? Is it not possible someone could provide a measurement for example that allows you to specify an arbitrary basis as a set of vectors or some other parameterized measurement? (Fine if this is just a restriction that's reasonable for now. Just want to understand).

orpuente-MS:
> Yes, `Reset` is an interesting special case, that I think may need to remain a special case for a while. It should definitely have the "irreversible" QIR attribute and today that is missing. But it also doesn't produce a measurement result anywhere, so it doesn't get the same treatment as others (having its signature changed, getting deferred, etc). Also, when you consider how we do the deferral, there is a difference between measurement and reset (or to put it another way, non-destructive vs destructive measurement) in terms of whether the new, replacement qubit needs to have the corresponding state as the original.

Ohh, I didn't know that Reset was that different, thank you for clarifying. I edited my last message to use `Projection` instead of `Reset` to make my point better.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -274,6 +277,17 @@ impl<'a> Checker<'a> {
         }
     }
 
+    /// Check that there are no custom measurements in Base Profile.
+    fn check_measurement(&mut self, attrs: &[Attr], span: Span) {
```
swernli, line 281: Even though this is only being used for measurement right now, it's a good spot to include future disallow of attributes based on capabilities (like the `@Reset` we anticipate needing), so I think this can be called `check_attrs` to cover the generic case.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -176,6 +178,7 @@ impl<'a> Visitor<'a> for Checker<'a> {
         // We only care about callables.
         if let ItemKind::Callable(callable_decl) = &item.kind {
             self.set_current_callable(item.id);
+            self.check_measurement(&item.attrs, item.span);
```
swernli, line 181: Rather than underline the whole item declaration, I think just underlining the name would be less overwhelming (and less likely to hide other errors). So something like:

```suggestion
            self.check_measurement(&item.attrs, callable_decl.name.span);
```

compiler/qsc_rca/src/errors.rs
```diff
@@ -161,6 +161,11 @@ pub enum Error {
     #[diagnostic(code("Qsc.CapabilitiesCk.MeasurementWithinDynamicScope"))]
     MeasurementWithinDynamicScope(#[label] Span),
 
+    #[error("cannot define a custom measurement")]
+    #[diagnostic(help("cannot define a custom measurement in the configured target profile"))]
+    #[diagnostic(code("Qsc.CapabilitiesCk.CustomMeasurement"))]
```
swernli, line 166: We'll want to add a section on the wiki page for this error with some explanation and update this error to include the link. We can take a look at what that section should say and get that updated and link added here before the PR goes in so that we don't forget to follow up!

compiler/qsc_passes/src/capabilitiesck.rs
orpuente-MS: Yes, makes sense. Updated.

compiler/qsc_passes/src/measurement.rs
```diff
@@ -0,0 +1,127 @@
...
+use qsc_hir::{
+    hir::{Attr, CallableDecl, Item, ItemKind, Package, SpecBody, SpecGen},
+    ty::{Prim, Ty},
+    visit::Visitor,
+};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
+pub enum Error {
+    #[error("a measurement should take at least one Qubit as argument")]
```
swernli, line 18: nitpick: for this and other errors in the file, since `Qubit` and `Result` are types, I would phrase it in a way that leans into that, with patterns like "a custom measurement should take at least one argument of type `Qubit`" or similar. Also, worth using "custom measurement" or perhaps even "a callable with the measurement attribute" to make it clear that we are talking specifically about use of the `@Measurement` feature and not the concept of measurement in general.

compiler/qsc_passes/src/measurement.rs
```diff
@@ -0,0 +1,127 @@
...
+/// A helper structure to find the measurement callables in a Package.
+struct MeasurementFinder<'a> {
+    callables: Vec<(&'a CallableDecl, &'a [Attr])>,
+}
+
+impl<'a> Visitor<'a> for MeasurementFinder<'a> {
+    fn visit_item(&mut self, item: &'a Item) {
+        if let ItemKind::Callable(callable) = &item.kind {
+            if item.attrs.contains(&Attr::Measurement) {
+                self.callables.push((callable, &item.attrs));
```
swernli, line 123: Rather than collecting all of the callables and then checking them, you could call the `validate_measurement_declaration` right in here and collect the errors. That technically scales better here, though I doubt we'll end up in a situation with thousands of these such that the difference would be noticeable :).

compiler/qsc_passes/src/capabilitiesck/tests_adaptive_plus_integers.rs
```diff
@@ -400,6 +400,16 @@ fn measurement_within_dynamic_scope_yields_no_errors() {
     );
 }
 
+#[test]
+fn custom_measurement_yields_errors() {
```
swernli, line 404:
```suggestion
fn custom_measurement_yields_no_errors() {
```

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2260,6 +2264,98 @@ impl<'a> PartialEvaluator<'a> {
...
+                        return Err(Error::EvaluationFailed(
+                            "the arguments of a measurement should be Qubits".to_string(),
+                            args_span,
+                        ));
+                    }
+                }
+            }
+            _ => {
+                return Err(Error::EvaluationFailed(
+                    "the arguments of a measurement should be Qubits".to_string(),
```
swernli, line 2329:
```suggestion
                    "the outputs of a measurement should be Results".to_string(),
```

compiler/qsc_passes/src/measurement.rs
orpuente-MS: Good catch. I changed the code to do that instead.

compiler/qsc_frontend/src/lower.rs
```diff
@@ -921,10 +934,14 @@ impl With<'_> {
...
-        ast::CallableKind::Operation => hir::CallableKind::Operation,
+fn lower_callable_kind(kind: ast::CallableKind, attrs: &[qsc_hir::hir::Attr]) -> hir::CallableKind {
+    if attrs.contains(&qsc_hir::hir::Attr::Measurement) {
+        hir::CallableKind::Measurement
+    } else {
+        match kind {
+            ast::CallableKind::Function => hir::CallableKind::Function,
+            ast::CallableKind::Operation => hir::CallableKind::Operation,
+        }
     }
```
swernli, line 945: It just occurred to me that this logic makes it so that a function can have the `@Measurement` attribute and it will still work, which doesn't really make sense. I think this points to two options:
 - Update this helper function to be able to produce errors (it would take `self` and callers would be updated, so that `self.lowerer.errors.push` can be used as it is elsewhere).
 - Keep just the original two `CallableKind` and update the logic in partial eval and the HIR pass to peek at the attributes to get the info about whether this callable was marked as a measurement.

I think the first option would be fine, just wanted to mention the second as a comparison. Either way, worth adding a unit test to the right spot to confirm that a function with `@Measurement` produces an error.

compiler/qsc/src/codegen/tests.rs
```diff
@@ -525,6 +525,145 @@ mod adaptive_profile {
...
+
+        let (std_id, store) = crate::compile::package_store_with_stdlib(capabilities);
+        let qir = get_qir(
+            sources,
+            language_features,
+            capabilities,
+            store,
+            &[(std_id, None)],
+        )
+        .expect("Failed to generate QIR");
```
swernli, line 554: nitpick: I mentioned this on another PR, but I think it's worth keeping in mind some guidance from the style recommendation in the Rust stdlib (see [Common Message Styles](https://doc.rust-lang.org/std/error/index.html#common-message-styles)) that "expect as precondition" is preferable. It's not critical, but I do think that style makes the readability of the code and the diagnosibility of the tests a little nicer. Food for thought, not worth blocking the PR over.

compiler/qsc/src/codegen/tests.rs
```diff
@@ -525,6 +525,145 @@ mod adaptive_profile {
         "#]].assert_eq(&qir);
     }
```
cesarzc, line 527: Since this is a feature only meant for Adaptive profile, it would be worth adding a negative test where it is asserted that QIR generation fails when a custom measurement is included and you are targeting Base profile.

compiler/qsc/src/codegen/tests.rs
```diff
@@ -744,6 +883,151 @@ mod adaptive_ri_profile {
...
 
+    #[test]
+    fn custom_measurement_generates_correct_qir() {
```
cesarzc, line 887: It seems like there are two tests that have the same name, I'm surprised the compiler doesn't produce errors in this case. Additionally, it looks like the only difference between this test and the other one with the same name is tha capabilities. Is there something specific we want to test regarding behavior that would be affected by other flags that are not `TargetCapabilityFlags::Adaptive`?

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1310,6 +1310,10 @@ impl<'a> PartialEvaluator<'a> {
         args_span: PackageSpan,        // For diagnostic purposes only.
         callee_expr_span: PackageSpan, // For diagnostic puprposes only.
     ) -> Result<Value, Error> {
+        if matches!(callable_decl.kind, qsc_fir::fir::CallableKind::Measurement) {
```
cesarzc, line 1313: Do you plan to update the standard library in a later PR to use `@Measurement` and remove the special cases for measuring qubits?

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2260,6 +2264,98 @@ impl<'a> PartialEvaluator<'a> {
...
+        let measure_callable_id = self.get_or_insert_callable(measurement_callable);
+        let instruction = Instruction::Call(measure_callable_id, operands, None);
+        let current_block = self.get_current_rir_block_mut();
+        current_block.0.push(instruction);
+
+        match results_values.len() {
+            0 => Err(Error::EvaluationFailed(
+                "a measurement should return one or more results".to_string(),
+                args_span,
+            )),
```
cesarzc, line 2353:
```suggestion
            0 => panic!("unexpected unitary measurement"),
```

cesarzc: Since this is checked at the HIR level, this should probably be a panic instead since it is an invalid state that should not happen, and if it happens it means something went wrong.

compiler/qsc_passes/src/measurement/tests.rs
```diff
@@ -0,0 +1,149 @@
...
+        &[],
+        sources,
+        TargetCapabilityFlags::all(),
+        LanguageFeatures::default(),
+    );
+
+    let errors = validate_measurement_declarations(&unit.package);
+    expect.assert_debug_eq(&errors);
+}
+
```
cesarzc, line 25: It would be worth adding some tests that use both the `@Measurement` and `@SimulatableIntrinsic` attributes.

compiler/qsc_frontend/src/lower.rs
orpuente-MS: Good catch. Fixed.

compiler/qsc/src/codegen/tests.rs
orpuente-MS: There were other tests with repeated names in that file. The reason is that the name of the test is its full path:
 ```rust
 adaptive_ri_profile::custom_measurement_generates_correct_qir
 ```
 
 which differentiates it from:
 ```rust
 base::custom_measurement_generates_correct_qir
 ```

compiler/qsc_partial_eval/src/lib.rs
orpuente-MS: Yes, that's the plan.

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -274,6 +277,17 @@ impl<'a> Checker<'a> {
         }
     }
 
+    /// Check that there are no custom measurements in Base Profile.
```
cesarzc, line 280: Even though this approach will surface an error when someone is using a custom measurement in Base profile, it seems like the pattern used is a bit different from the pattern used for all the other features and capabilities.

The RCA pattern and how errors are surfaced is the following:
- A feature is identified and represented by a new flag in `RuntimeFeatureFlags`.
    - In this case, the new feature could be something like `CustomMeasurement`.
- The feature flag maps to a set of capabilities through the `target_capabilities` function.
    - In this case it would map to `TargetCapabilityFlags::Adaptive`.
- The feature is identified as part of RCA, setting the `CustomMeasurement` flag where appropriate.
- The capabilities check pass automatically maps unsupported features based on the capabilities passed, so in theory no changes would be needed here.

I would suggest to use the established pattern instead. If you have any questions, I'll be happy to help!

compiler/qsc_passes/src/capabilitiesck.rs
```diff
@@ -99,6 +100,7 @@ struct Checker<'a> {
     current_callable: Option<LocalItemId>,
     missing_features_map: FxHashMap<Span, RuntimeFeatureFlags>,
     store: &'a qsc_fir::fir::PackageStore,
+    errors: Vec<Error>,
```
cesarzc, line 103: I believe this shouldn't be needed anymore.

compiler/qsc_rca/src/core.rs
```diff
@@ -2289,6 +2292,60 @@ fn derive_instrinsic_operation_application_generator_set(
     }
 }
 
+fn derive_instrinsic_measurement_application_generator_set(
```
cesarzc, line 2295: Since this function and `derive_instrinsic_operation_application_generator_set` are almost identical, it might be worth just having one function.

Since the information whether the callable is a regular operation or a measurement is in the `callable_context` parameter, I would suggest to use that to populate the runtime features for the inherent compute kind.

cesarzc (approved):

compiler/qsc_frontend/src/lower.rs
```diff
@@ -30,8 +30,12 @@ pub(super) enum Error {
     #[error("invalid attribute arguments: expected {0}")]
     #[diagnostic(code("Qsc.LowerAst.InvalidAttrArgs"))]
     InvalidAttrArgs(String, #[label] Span),
+    #[error("invalid use of the Measurement attribute on a function")]
+    #[diagnostic(code("Qsc.LowerAst.InvalidMeasurementAttrOnFunction"))]
+    InvalidMeasurementAttrOnFunction(#[label] Span),
     #[error("missing callable body")]
     #[diagnostic(code("Qsc.LowerAst.MissingBody"))]
+    #[diagnostic(help("try declaring the callable as an operation"))]
```
swernli, line 38: whoops, looks like the help text got added to the wrong error, should get moved up one.

swernli (approved):
One minor fix for help message left, otherwise it looks good!

compiler/qsc_frontend/src/lower.rs
orpuente-MS: Ahh, good catch!

PR #1968, opened 2024-10-17
Usage telemetry in `qsharp` pip package

sezna:
# Pip Package Telemetry

## Summary

This PR introduces a minimal set of telemetry events for the qsharp python package. Users can opt-out by setting the environment variable `NO_TELEMETRY=1`.

We only collect simple event counts so we can understand better which aspects of our library are being used. 

## Notes

A `PythonLoggingProvider` is a struct that knows how to handle `TelemetryEvent`s from our package. `opentelemetry` is one such `PythonLoggingProvider`, and this is the one that will send telemetry to our application insights instance. The other two are `TelemetryDisabled`, used when telemetry is disabled; and `MockLoggingProvider`, used when telemetry output is tested via `pytest`.

Unfortunately, this does mean we are now sending networked telemetry via our python module, and this does cause us to incur a decent dependency penalty. 

This PR also tries to centralize the location of telemetry, similar to how we define [VS Code telemetry events](https://github.com/microsoft/qsharp/blob/main/vscode/src/telemetry.ts#L10), so that any outside user can easily inspect what events we collect and determine if they're okay with it.

## Testing
Existing pytests can be modified to also assert on the expected telemetry to be collected, and I've converted a few already.

## Remaining work
- [ ] Minimize dependencies
- [ ] Get application insights resource provisioned
- [ ] Ideally remove lazy_static for mock tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
 name = "ahash"
 version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


billti:
Over 1,200 lines added to the cargo.lock file!! :-o

Besides all the extra (transitive) dependencies, I'm not a huge fan of calling into Rust code just to log Python telemetry from a Python package. Besides the additional code complexity, I worry about the interaction complexity (are the requests async in Rust? How is that dealt with at the Rust/Python integration level)

It's annoying that the Azure Python OpenTelemetry package is so huge and also has a ton of dependencies (installing it in a clean workspace added 43MB or Python for me, and dozens of packages :-/). A lot of this complexity in the packages comes from wanting to handle traces and logs and auto-instrument frameworks and deal with causality etc.. But all we want to do it log that a thing happened, and maybe how long it took and some extra properties (what OpenTelemetry calls a Metric).

Ultimately, this is just an anonymous HTTP requests with a JSON payload to the Azure Monitor endpoint. I took the liberty of browsing some code and doing a network trace, and just wrote up the Python code to make the request directly. Take a look at the Gist in https://gist.github.com/billti/06cf1c16c36a1f9139aa437ad449b128 , and you have access to the AppInsights workbook to see this data is coming through.

Let's discuss on Monday if this approach might be viable. It's only ~100 lines of Python, but may also have too many limitations long term.

PR #1971, opened 2024-10-22
Added Pauli noise support to sparse simulator

DmitryVasilevsky:
Added support for parametric Pauli noise to sparce simulator. (px, py, pz) can be provided. Sparse simulator will apply X, Y or Z gates after each gate and before each measurement with corresponding probability.
- This allows for simulation with parametric Pauli noise: bit flip, phase flip, depolarizing, etc.
- For two-qubit gates noise is applied to each qubit.
- Allocation and service functions do not apply noise. Idle (Identity) gate is also noiseless.
- Reset applies noise.
- Added tests.
- Released qubits are no longer checked to be in zero state by consumers of the backend via calling qubit_is_zero. Instead, qubit_release returns Boolean. True indicates, that the release was valid - the qubit was in a zero state in a noiseless simulation. It will always be true in noisy simulations when the release of a qubit is always valid.
- The functionality is not yet exposed outside Rust code. Will come in subsequent PRs.

### DIFF
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -214,12 +214,13 @@ impl Backend for SparseSim {
         if res {
             self.sim.x(q);
         }
-        self.apply_noise(q); // Applying noise after reset
+        // After reset qubits start in ground state even with noise.
         res
     }
 
     fn reset(&mut self, q: usize) {
         self.mresetz(q);
+        // After reset qubits start in ground state even with noise.
     }
 
     fn rx(&mut self, theta: f64, q: usize) {
@@ -319,9 +320,8 @@ impl Backend for SparseSim {
     }
 
     fn qubit_allocate(&mut self) -> usize {
-        let q = self.sim.allocate();
-        self.apply_noise(q); // Noise is applied to fresh qubit.
-        q
+        // Fresh qubit start in ground state even with noise.
+        self.sim.allocate()
     }
 
     fn qubit_release(&mut self, q: usize) -> bool {
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -334,7 +334,7 @@ impl Backend for SparseSim {
     }
 
     fn qubit_swap_id(&mut self, q0: usize, q1: usize) {
-        // This is an service function rather than a gate so it doesn't incur noise.
+        // This is a service function rather than a gate so it doesn't incur noise.
         self.sim.swap_qubit_ids(q0, q1);
     }
 
@@ -359,7 +359,7 @@ impl Backend for SparseSim {
     }
 
     fn qubit_is_zero(&mut self, q: usize) -> bool {
-        // This is an service function rather than a measurement so it doesn't incur noise.
+        // This is a service function rather than a measurement so it doesn't incur noise.
         self.sim.qubit_is_zero(q)
     }
 
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -1,12 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use crate::noise::PauliNoise;
+use crate::val::Value;
 use num_bigint::BigUint;
 use num_complex::Complex;
 use quantum_sparse_sim::QuantumSim;
-use rand::RngCore;
+use rand::{rngs::StdRng, Rng, RngCore, SeedableRng};
 
-use crate::val::Value;
+#[cfg(test)]
+mod noise_tests;
 
 /// The trait that must be implemented by a quantum backend, whose functions will be invoked when
 /// quantum intrinsics are called.
@@ -82,7 +85,11 @@ pub trait Backend {
     fn qubit_allocate(&mut self) -> usize {
         unimplemented!("qubit_allocate operation");
     }
-    fn qubit_release(&mut self, _q: usize) {
+    /// `false` indicates that the qubit was in a non-zero state before the release,
+    /// but should have been in the zero state.
+    /// `true` otherwise. This includes the case when the qubit was in
+    /// a non-zero state during a noisy simulation.
+    fn qubit_release(&mut self, _q: usize) -> bool {
         unimplemented!("qubit_release operation");
     }
     fn qubit_swap_id(&mut self, _q0: usize, _q1: usize) {
@@ -102,7 +109,14 @@ pub trait Backend {
 
 /// Default backend used when targeting sparse simulation.
 pub struct SparseSim {
+    /// Noiseless Sparse simulator to be used by this instance.
     pub sim: QuantumSim,
+    /// Pauli noise that is applied after a gate or before a measurement is executed.
+    /// Service functions aren't subject to noise.
+    pub noise: PauliNoise,
+    /// Random number generator to sample Pauli noise.
+    /// Noise is not applied when rng is None.
+    pub rng: Option<StdRng>,
 }
 
 impl Default for SparseSim {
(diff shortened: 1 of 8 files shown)


compiler/qsc_eval/src/backend.rs
```diff
@@ -301,9 +396,16 @@
     }
 
     fn set_seed(&mut self, seed: Option<u64>) {
-        match seed {
-            Some(seed) => self.sim.set_rng_seed(seed),
-            None => self.sim.set_rng_seed(rand::thread_rng().next_u64()),
+        if let Some(seed) = seed {
+            if !self.is_noiseless() {
+                self.rng = Some(StdRng::seed_from_u64(seed));
```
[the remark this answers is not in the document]
DmitryVasilevsky: This is not use for security. Just a simulation of a quantum device.

compiler/qsc_eval/src/backend.rs
```diff
@@ -116,38 +130,86 @@ impl SparseSim {
...
     fn m(&mut self, q: usize) -> Self::ResultType {
+        self.apply_noise(q);
         self.sim.measure(q)
     }
 
     fn mresetz(&mut self, q: usize) -> Self::ResultType {
+        self.apply_noise(q);
         let res = self.sim.measure(q);
         if res {
             self.sim.x(q);
```
swernli, line 215: I hadn't thought of this before, but should `mresetz` (and by extension `reset`) include applying noise on the conditional bit flip? Do want returning to the ground state itself to be subject to probabilistic failure?

compiler/qsc_eval/src/noise.rs
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+}
+
+impl PauliNoise {
+    pub fn from_probabilities(px: f64, py: f64, pz: f64) -> Result<Self, String> {
+        let px_py = px + py;
+        let px_py_pz = px_py + pz;
+        if px < 0.0 || py < 0.0 || pz < 0.0 || px_py_pz > 1.0 {
+            Err("Incorrect Pauli noise probabilities.".to_string())
```
swernli, line 28: nitpick: This error case may need to get more formal depending on how it gets exposed up through the APIs for configuring noise via Python/Jupyter/VS Code. Not an issue in this PR though.

compiler/qsc_eval/src/backend/noise_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+    let _ = PauliNoise::from_probabilities(0.3, 0.4, 0.5)
+        .expect_err("Expected error for probabilities 0.3, 0.4, 0.5.");
+}
+
+#[test]
+fn noisy_simulator() {
+    let sim = SparseSim::new();
+    assert!(sim.is_noiseless(), "Expected noiseless simulator.");
+
+    let noise = PauliNoise::from_probabilities(0.0, 0.0, 0.0).expect("Cannot construct noise.");
```
swernli, line 60: nitpick: these expect messages tend to follow the "expect as error message" style whereas I tend to agree with the leanings of the style recommendation in the Rust stdlib (see [Common Message Styles](https://doc.rust-lang.org/std/error/index.html#common-message-styles)) that "expect as precondition" is preferable. I've gotten used to reading the expect as almost like a sentence: I expect that <msg>. So this would be `.expect("noiseless should be constructable")` with starting with lower case. Since these are tests it's not critical, but especially if a test fails unpredictably the log output from something like CI might be our only clue, so having more natural langauge expectation could be useful. Food for thought, up to you if you fix these particular ones.

compiler/qsc_eval/src/backend/noise_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+    let mut sim = SparseSim::new_with_noise(&noise);
+    let q = sim.qubit_allocate();
+    for _ in 0..100 {
+        sim.x(q);
+        let res1 = sim.m(q);
+        assert!(res1, "Unexpected value. True expected without noise.");
+        sim.x(q);
+        let res2 = sim.m(q);
+        assert!(!res2, "Unexpected value. False expected without noise.");
+    }
```
swernli, line 85: I don't think a loop is needed here, I would just verify that basic bit flipping works to one and then back to zero rather than put this in a loop.

compiler/qsc_eval/src/backend/noise_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+            res1,
+            "Unexpected value. True expected for 100% bitflip noise."
+        );
+        let res2 = sim.m(q); // Always applies X before measuring
+        assert!(
+            !res2,
+            "Unexpected value. False expected for 100% bitflip noise."
+        );
+    }
+    assert!(sim.qubit_release(q), "Unexpected qubit state on release.");
```
swernli, line 107: here too, I don't think the loop is really needed.

compiler/qsc_eval/src/backend/noise_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+#[test]
+fn bitflip_gate() {
+    let noise = PauliNoise::from_probabilities(1.0, 0.0, 0.0).expect("Cannot construct noise.");
+    let mut sim = SparseSim::new_with_noise(&noise);
+    assert!(!sim.is_noiseless(), "Expected noisy simulator.");
+    let q = sim.qubit_allocate();
+    for _ in 0..100 {
```
swernli, line 116: same here, this can be just the simple sequence without the loop.

swernli (approved):
Approved with some minor feedback!

compiler/qsc_eval/src/backend.rs
DmitryVasilevsky: I think it is, and my experimental branch had that. Now I think that allocation of fresh qubits should also be subject to this noise. I changed code to apply noise right after allocation and also after reset.

compiler/qsc_eval/src/backend/noise_tests.rs
DmitryVasilevsky: I keep forgetting what the recommendation is. I reworked the errors.

compiler/qsc_eval/src/backend/noise_tests.rs
DmitryVasilevsky: I changed code and moved 100% bitflip "gate test" to be like other "noisy_via_?" tests without the loop. However, I kept loops in the two other tests. They need loops to test probabilistic nature of noise. First of all, if the implementation in the SparseSim is correct, all iterations should be successful. So, this won't be a flaky or "may fail sometimes" test. It is also very fast - negligible time on my computer. However, if the implementation in the SparseSim is not correct, it may produce an error, and it may do so only sometimes. The more iterations I run, the more chances that the error will reveal itself. So doing this in a loop would make test less valuable.

compiler/qsc_eval/src/backend/noise_tests.rs
DmitryVasilevsky: Same as previous comment.

compiler/qsc_eval/src/backend/noise_tests.rs
DmitryVasilevsky: I removed the loop from this instance and updated code.

compiler/qsc_eval/src/backend.rs
```diff
@@ -82,7 +85,11 @@ pub trait Backend {
     fn qubit_allocate(&mut self) -> usize {
         unimplemented!("qubit_allocate operation");
     }
-    fn qubit_release(&mut self, _q: usize) {
+    /// `false` indicates that the qubit was in a non-zero state before the release,
+    /// but should have been in the zero state.
+    /// `true` otherwise. This includes the case when the qubit was in
+    /// a non-zero state during a noisy simulation.
+    fn qubit_release(&mut self, _q: usize) -> bool {
```
orpuente-MS, line 92: I am a bit confused by the last sentence of the docstring. If the qubit is in a non-zero state due to simulation noise it returns `true`? Shouldn't it return `false` in that case?

orpuente-MS (approved):
Approved with minor comment.

compiler/qsc_eval/src/backend.rs
DmitryVasilevsky: After playing with it for a while I changed my mind. I will now remove noise on freshly allocated qubits and after reset. We will expose apply_noise to qshap. The user will be able to apply noise explicitly if this is needed.

compiler/qsc_eval/src/backend.rs
DmitryVasilevsky: Probably you are missing the second line of the first point - 'but should have been in the zero state'. When simulation is noisy, this assumption is no longer true. It could be in the non-zero state before the release and it's not a problem.
That said, I understand the confusion. I'll probably make this simpler.

Issue #1946, opened 2024-10-03
Deprecate Python 3.8, making 3.9 the oldest target Python version


idavis:
Python 3.8 is EOL this month (October 2024).

Update Python package to target abi39 in Rust and >=3.9 in the Python project.

We can add 3.13 support at the same time.

This will require migrating manylinux build to use a newer image. Mariner2 should work as it uses Python 3.9 and glibc 2.35. This needs to be done in the ADO pipelines.

idavis:
Fixed

PR #1972, opened 2024-10-22
Initial Python telemetry

billti:
Python feature usage telemetry.


pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+from typing import TypedDict, Union
+
+logger = logging.getLogger(__name__)
+
+# TODO: These should be potentially set by the build pipeline. However, they are not
+# secret values, and it might be useful to still get telemetry even from other builds or forks
+AIKEY = os.environ.get("QSHARP_AI_KEY") or "5ef527ea-7943-4e42-866d-e3b291d44e2b"
+AIURL = (
+    os.environ.get("QSHARP_AI_URL")
+    or "https://westus2-2.in.applicationinsights.azure.com//v2.1/track"
```
idavis, line 39: why is there a double `/` in the url?

billti: That was what I sniffed off the network (i.e. that's what the official Azure Monitor Python SDK sends). I'm guessing it's a mistake on their part, but seems benign, so I left it in just for full fidelity/compatibility with the 'official' behavior.

pip/qsharp/_qsharp.py
```diff
@@ -317,6 +322,9 @@ def on_save_events(output: Output) -> None:
         # compilation.
         entry_expr = None
 
+    durationMs = (monotonic() - start_time) * 1000
```
idavis, line 325: Do you want to clamp this to integral ms or leave it as float?

billti: I though float was fine. Any reason to clamp it?

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+            next_post_sec = None
+        else:
+            next_post_sec = time.monotonic() + BATCH_INTERVAL_SEC
+
+
+# When the process is about to exit, notify the telemetry thread to flush, and wait max 3 sec before exiting anyway
+def on_exit():
+    logger.debug("In on_exit handler")
+    telemetry_queue.put("exit")
+    telemetry_thread.join(timeout=3)
```
idavis, line 301: create var `timeout_in_seconds = 3` to avoid magic number. and pass to `join`

billti: The value is used once and passed as an arg prefixed with `timeout=`. I don't think extracting into a separate var is necessary for clarity.

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+
+def _pending_to_payload():
+    """Converts the pending metrics to the JSON payload for Azure Monitor"""
+
+    result_array = []
+    formatted_time = (
+        datetime.now(timezone.utc)
+        .isoformat(timespec="microseconds")
+        .replace("+00:00", "Z")
+    )
```
idavis, line 180: What is this time format? I would have expected something like ISO 8601: `datetime.datetime.now(datetime.timezone.utc).isoformat()`

pip/qsharp/telemetry.py
idavis: Agree to disagree, to me it is like `BATCH_INTERVAL_SEC`.

pip/qsharp/telemetry_events.py
```diff
@@ -0,0 +1,87 @@
...
+def on_compile(profile: str) -> None:
+    log_telemetry("qsharp.compile", 1, properties={**default_props, "profile": profile})
+
+
+def on_compile_end(durationMs: float, profile: str) -> None:
+    log_telemetry(
+        "qsharp.compile.durationMs",
+        durationMs,
+        properties={**default_props, "profile": profile},
+        type="histogram",
```
idavis, line 73: Why is compile's type `histogram`?

pip/qsharp/_qsharp.py
idavis: Do we care about fractional ms for querying? Unless we need float, I would normally use int for faster processing.

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+class PendingMetric(Metric):
+    """Used internally to aggregate metrics before sending"""
+
+    min: float
+    max: float
+
+
+# Maintain a collection of custom metrics to log, stored by metric name with a list entry
+# for each unique set of properties per metric name
+pending_metrics: dict[str, list[PendingMetric]] = {}
```
idavis, line 83: Can't use `dict` and `list` until we implement [#1946](https://github.com/microsoft/qsharp/issues/1946)

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+# The telemetry queue is used to send telemetry from the main thread to the telemetry thread
+# This simplifies any thread-safety concerns, and avoids the need for locks, etc.
+telemetry_queue = SimpleQueue()
+
+
+def log_telemetry(
+    name: str,
+    value: float,
+    count: int = 1,
+    properties: dict = {},
```
idavis, line 94: See above, need to use Dict

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+AI_DEVICE_OS_VERSION = platform.version()
+
+
+class Metric(TypedDict):
+    """Used internally for objects in the telemetry queue"""
+
+    name: str
+    value: float
+    count: int
+    properties: dict
```
idavis, line 70: `Dict[str, Any]`

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+# This simplifies any thread-safety concerns, and avoids the need for locks, etc.
+telemetry_queue = SimpleQueue()
+
+
+def log_telemetry(
+    name: str,
+    value: float,
+    count: int = 1,
+    properties: dict = {},
+    type: str = "counter",
```
idavis, line 95: Should this be a str enum?

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+        "count": count,
+        "properties": properties,
+        "type": type,
+    }
+
+    logger.debug("Queuing telemetry: %s", obj)
+    telemetry_queue.put(obj)
+
+
+def flush_telemetry():
```
idavis, line 130: What is this fuction for? I don't think users should be manually flushing telemetry.

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+        .replace("+00:00", "Z")
+    )
+    for name in pending_metrics:
+        for unique_props in pending_metrics[name]:
+            # The below matches the entry format for Azure Monitor REST API
+            entry = {
+                "ver": 1,
+                "name": "Microsoft.ApplicationInsights.Metric",
+                "time": formatted_time,
+                "sampleRate": 100.0,
```
idavis, line 188: Is `100.0` a special value or just some filler that doesn't apply to us?

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+            next_post_sec = time.monotonic() + BATCH_INTERVAL_SEC
+
+    while True:
+        try:
+            # Block if no timeout, else wait a maximum of time until the next post is due
+            timeout = (
+                None
+                if next_post_sec == None
+                else max(next_post_sec - time.monotonic(), 0)
+            )
```
idavis, line 268:
```suggestion
            timeout = None
            if next_post_sec:
                timeout = max(next_post_sec - time.monotonic(), 0)
```

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+                if not _post_telemetry():
+                    logger.error("Failed to post telemetry on flush")
+            else:
+                on_metric(msg)
+                # Loop until the queue has been drained. This will cause the 'Empty' exception
+                # below once the queue is empty and it's time to post
+                continue
+        except Empty:
+            # No more telemetry within timeout, so write what we have pending
+            _post_telemetry()
```
idavis, line 287: I think this may be needed once the function return types are included for linting.
```suggestion
            _ = _post_telemetry()
```

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
...
+        # If we get here, it's after a post attempt. Pending will still have items if the attempt
+        # failed, so updated the time for the next attempt in that case.
+        if len(pending_metrics) == 0:
+            next_post_sec = None
+        else:
+            next_post_sec = time.monotonic() + BATCH_INTERVAL_SEC
+
+
+# When the process is about to exit, notify the telemetry thread to flush, and wait max 3 sec before exiting anyway
+def on_exit():
```
idavis, line 298: Should this be `_on_exit`? I don't think this is a function users should really see.

pip/qsharp/telemetry.py
billti: Again, taken from the payload the SDK sends. Looks like they use this code, which we could adopt, but seems to do the same thing <https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/util/__init__.py#L24>

pip/qsharp/telemetry_events.py
billti: Basic types in OpenTelemetry are Counter and Histogram. Counter is for simple counting of events (i.e. how many times something happened). Histogram is for measuring quantitative values (e.g. duration, amount, etc.)

Could add a comment that links to <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#histogram> if we want to make it clearer to folks. (The comments on line 103 of telemetry.py do already go over this to some degree).

pip/qsharp/telemetry.py
billti: When the main thread is about it exit is signals to the telemetry thread to flush the queue. Else you'll lose events that still haven't sent on the next interval yet.

pip/qsharp/telemetry_events.py
```diff
@@ -0,0 +1,87 @@
...
+# gets the order of magnitude for the number of qubits
+def get_qubits_bucket(qubits: int):
+    if qubits <= 1:
+        return 1
+    elif qubits >= 1000000:
+        # Limit the buckets upper bound
+        return 1000000
+    else:
+        # Bucket into nearest (rounded up) power of 10, e.g. 75 -> 100, 450 -> 1000, etc.
+        return 10 ** math.ceil(math.log10(qubits))
```
billti, line 41: @sezna , I was going to bucket these differently to shots. Anything less than a 100 shots is trivial and it can legitimately go up to 1 million. For qubits in a simulation, there's a big difference between 1 or 2, 10, or 20 qubits in a simulation (and you get into the realm of not simulatable around 30 or 40). So going up in powers of 10 isn't really granular enough.

pip/qsharp/telemetry.py
idavis: I don't understand, this function isn't called.

pip/qsharp/telemetry.py
idavis: This function isn't called.

pip/qsharp/_qsharp.py
billti: This will happen once after other expensive processing (e.g. a simulation or resource estimation). I don't think the difference between an int or a float for this expression would be observable.

pip/qsharp/telemetry.py
billti: Ah, you're right. I was thinking I'd also flush on the 'init' event just to make sure we got that sent, but didn't end up going that path. (Didn't seem like that approach would add much with the flush on exit also). I'll remove. Thanks for catching!

pip/qsharp/telemetry.py
billti: I did that just to avoid the need for type annotation :-D Changed now.

pip/qsharp/telemetry.py
billti: Added a comment for even more clarity on intent.

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,308 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""
+This module sends telemetry directly to Azure Monitor using a similar mechanism and
```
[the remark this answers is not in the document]
billti: Went through and did a types on all feedback pass. Looks like the two `telemetry*.py` files are about the only ones we have that check clean with Pylance on 'strict' now! :-)

version.py
```diff
@@ -54,6 +54,13 @@ def update_file(file, old_text, new_text):
     r'version = "0.0.0"',
     r'version = "{}"'.format(pip_version),
 )
+
+update_file(
+    os.path.join(root_dir, "pip/qsharp/telemetry.py"),
+    r'QSHARP_VERSION = "0.0.0.dev0"',
+    r'QSHARP_VERSION = "{}"'.format(pip_version),
+)
+
```
idavis, line 63: Thought about unifying version access. I think this will work, then we only update the `__version__` var in `__init__.py` and all the others should be able to dynamically filll the `pyproject.toml` and it is accessible in code for telemetry via `qsharp.__version__` access:

In `__init__.py`
```python
__version__ = "0.0.0dev0"
```

In `get_version.py` (new file, perhaps rename `version.py` to `set_version.py`):
```python
import qsharp  # Replace with your actual module name

def get_version():
    return qsharp.__version__

if __name__ == "__main__":
    print(get_version())
```

Then in each `pyproject.toml`:
```toml
[project]
dynamic = ["version"]

[tool.setuptools.dynamic]
version = {attr = "get_version.get_version"}
```

pip/qsharp/telemetry.py
```diff
@@ -0,0 +1,305 @@
...
+)
+
+# Environment variables take precedence, else disable telemetry for non 'stable' builds
+QSHARP_PYTHON_TELEMETRY = (os.environ.get("QSHARP_PYTHON_TELEMETRY") or "").lower()
+TELEMETRY_ENABLED = (
+    True
+    if QSHARP_PYTHON_TELEMETRY in ["1", "true", "enabled"]
+    else (
+        False
+        if QSHARP_PYTHON_TELEMETRY in ["0", "false", "disabled", "none"]
```
idavis, line 49: Do you want this to be case sensitive?

idavis (approved):

pip/qsharp/telemetry.py
billti: No. That's why I explicitly did it this way with the `lower()` on line 43.

version.py
billti: Reading up on that approach, finding lots of folks having issues getting this to work in certain configurations. I think it's the right thing to do, but maybe as a follow-up PR rather than add risk to this one.

pip/qsharp/telemetry_events.py
```diff
@@ -23,7 +23,10 @@ def get_shots_bucket(shots: int) -> int:
 
 
 # gets the order of magnitude for the number of qubits
-def get_qubits_bucket(qubits: int) -> int:
+def get_qubits_bucket(qubits: str) -> str:
```
billti, line 26: Why did this change from int to str, when most of the time it still returns an int?

sezna: The RE data that gets returned is kind of a nonstandard format, and I wanted to be able to flag if that logicalQubits isn't where I expected it to be. I'll stringify the returns to match the type checking

sezna (approved):

PR #1973, opened 2024-10-23
Copy over quantum-viz.js source code 

minestarks:
This eliminates the dependency on the `@microsoft/quantum-viz.js` package and moves the source over into the `qsharp-lang` package.


ScottCarda-MS (approved):
Happy to see this getting in! Thanks for migrating this!

billti (approved):

PR #1974, opened 2024-10-25
Python clients can now run simulation with Pauli noise

DmitryVasilevsky:
`def run` function now has `pauli_noise: Optional[Tuple[float, float, float]]` parameter to set Pauli noise for simulation. The following classes can be used for typical noise models:
class PauliNoise(Tuple[float, float, float])
class DepolarizingNoise(PauliNoise)
class BitFlipNoise(PauliNoise)
class PhaseFlipNoise(PauliNoise)


pip/src/interpreter.rs
```diff
@@ -333,15 +333,25 @@ impl Interpreter {
...
         let mut receiver = OptionalCallbackReceiver { callback, py };
-        match self.interpreter.run(&mut receiver, entry_expr) {
+
+        let noise = match pauli_noise {
+            None => None,
+            Some((px, py, pz)) => match PauliNoise::from_probabilities(px, py, pz) {
+                Ok(noise_struct) => Some(noise_struct),
+                Err(error_message) => return Err(PyException::new_err(error_message)),
+            },
+        };
```
minestarks, line 352: This looks ideal to me, but since you asked (in the chat), I think this is how you'd write it with `.map` etc:

```rust
let noise = pauli_noise
    .map(|(px, py, pz)| {
        PauliNoise::from_probabilities(px, py, pz).map_err(PyException::new_err)
    })
    .transpose()?;
```

Yours looks a lot cleaner.

pip/tests/test_qsharp.py
```diff
@@ -257,6 +257,21 @@ def test_dump_operation() -> None:
             else:
                 assert res[i][j] == complex(0.0, 0.0)
 
+def test_run_with_noise_produces_noisy_results() -> None:
+    qsharp.init()
+    qsharp.set_quantum_seed(0)
```
minestarks, line 262: Do we need to set classical seed as well here to make it as deterministic as possible? (which seed does noisy simulation use?)

minestarks (approved):

pip/src/interpreter.rs
DmitryVasilevsky: Thank you! I'll keep my code.

pip/tests/test_qsharp.py
DmitryVasilevsky: We can add it for a peace of mind, but we don't need it. Pauli noise uses quantum seed, as it should (noise is quantum). Classical seed is used for things that don't involve quantum, such as DrawRandomInt, etc. And there're no calls to those functions in the test.

DmitryVasilevsky: I looked around and other tests set the seed they need to set (without trying to being "as deterministic as possible"). So, I think I'll keep my code.

pip/qsharp/_native.pyi
```diff
@@ -139,12 +139,19 @@ class Interpreter:
         """
         ...
 
-    def run(self, entry_expr: str, output_fn: Callable[[Output], None]) -> Any:
+    def run(
+        self,
+        entry_expr: str,
+        output_fn: Callable[[Output], None],
+        pauli_noise: Optional[Tuple[float, float, float]],
```
swernli, line 146: nitpick: this parameter is called `noise` both above this in _qsharp.py and below this in the Rust code, so why not just call it `noise` here too?

swernli (approved):

pip/qsharp/_native.pyi
DmitryVasilevsky: I updated.

PR #1975, opened 2024-10-25
Expose pauli noise settings in VS Code

billti:
Added VS Code settings for the noise model to use when running histograms.

### DIFF
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -20,8 +20,8 @@ use qsc::{
         CircuitEntryPoint,
     },
     target::Profile,
-    LanguageFeatures, PackageStore, PackageType, SourceContents, SourceMap, SourceName, SparseSim,
-    TargetCapabilityFlags,
+    LanguageFeatures, PackageStore, PackageType, PauliNoise, SourceContents, SourceMap, SourceName,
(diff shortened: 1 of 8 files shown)


idavis (approved):

wasm/src/lib.rs
```diff
@@ -397,6 +412,32 @@ pub fn run(
...
+            pauliArray
+                .get(1)
+                .as_f64()
+                .expect("Probabilities should be floats"),
+            pauliArray
+                .get(2)
+                .as_f64()
+                .expect("Probabilities should be floats"),
+        )
+        .expect("Unable to create Pauli noise from the array provided")
```
DmitryVasilevsky, line 436: Will this expect result in panic? We probably want to do the return 'Err(JsError::new( ... ))' and return the error string from underlying Err case. These numbers come from a user input and shouldn't result in Panic.
Even if you are checking all conditions for Pauli noise in your code...

DmitryVasilevsky (approved):

wasm/src/lib.rs
billti: It should return an error. (Also, only VS Code calls this currently and has already done a bunch of validation, so should never happen).

minestarks (approved):

PR #1978, opened 2024-10-28
Improve the built-in API docs UX

billti:
This adds structure, navigation, and search to the API Docs page.

### DIFF
--- a/compiler/qsc_doc_gen/src/generate_docs.rs
+++ b/compiler/qsc_doc_gen/src/generate_docs.rs
@@ -350,6 +350,18 @@ enum PackageKind {
     Core,
 }
 
+impl Display for PackageKind {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        let s = match &self {
+            PackageKind::UserCode => "__Main__",
+            PackageKind::AliasedPackage(alias) => alias,
+            PackageKind::StandardLibrary => "__Std__",
+            PackageKind::Core => "__Core__",
+        };
+        write!(f, "{s}")
+    }
+}
+
 impl Display for Metadata {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         let kind = match &self.kind {
@@ -366,11 +378,19 @@ title: {}
 ms.date: {{TIMESTAMP}}
 ms.topic: {}
 qsharp.kind: {}
+qsharp.package: {}
 qsharp.namespace: {}
 qsharp.name: {}
 qsharp.summary: \"{}\"
 ---",
-            self.uid, self.title, self.topic, kind, self.namespace, self.name, self.summary
+            self.uid,
+            self.title,
+            self.topic,
+            kind,
+            self.package,
+            self.namespace,
+            self.name,
+            self.summary
         )
     }
 }
--- a/vscode/src/documentation.ts
+++ b/vscode/src/documentation.ts
@@ -38,9 +38,13 @@ export async function showDocumentationCommand(extensionUri: Uri) {
     }
   }
 
+  let projectName = program.programConfig.projectName ?? "App";
+  if (projectName.endsWith(".qs")) projectName = projectName.slice(0, -3);
+
   const message = {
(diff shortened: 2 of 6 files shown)


compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -366,11 +378,19 @@ title: {}
 ms.date: {{TIMESTAMP}}
 ms.topic: {}
 qsharp.kind: {}
+qsharp.package: {}
```
DmitryVasilevsky, line 381: This is compatible with documentation consumption for MS learn site, right?

vscode/src/webview/docview.tsx
```diff
@@ -2,13 +2,433 @@
...
-  const contentToRender = props.fragmentsToRender.join("<br/>\n\n---\n\n");
+export interface IDocFile {
+  filename: string;
+  metadata: string;
+  contents: string;
+}
+
+interface ItemDocs {
+  pkg: string;
+  module: string;
```
DmitryVasilevsky, line 15: Here and everywhere. Are we Ok using 'module' terminology?

vscode/src/webview/docview.tsx
```diff
@@ -2,13 +2,433 @@
...
+  // currPath is of the format: "<pkg>/<module>/<member>", e.g.
+  // "Std/Canon/CCNOT" or "Std/Microsoft.Quantum.Diagnostics/AssertMeasurementEqual" or
+  // "Unsigned/Main/GetInt" or "Main/Particle/Particle". When at the top level, currPath is "".
+
+  const [currPath, setPath] = useState("");
+  const [searchText, setSearchText] = useState("");
+
+  useEffect(() => {
+    // If the member is navigated to, scroll to it after rendering
+    const member = currPath.split("/")[2];
```
DmitryVasilevsky, line 172: what's going to happen at the top level?
> _When at the top level, currPath is ""_

vscode/src/webview/docview.tsx
```diff
@@ -2,13 +2,433 @@
...
+      ? GetPageContents(currPath, props.fragmentsToRender, setPath)
+      : {};
+
+  const searchResults =
+    searchText === ""
+      ? []
+      : getSearchResults(searchText, props.fragmentsToRender);
+
+  // Used to bold the text links when overing
+  function overLi(e: MouseEvent) {
```
DmitryVasilevsky, line 188: Maybe CSS style instead of events if possible?

DmitryVasilevsky (approved):

vscode/src/webview/docview.tsx
billti: member will be an empty string

compiler/qsc_doc_gen/src/generate_docs.rs
billti: (Note: Confirming with the docs team just to be safe. I'll get the Ok before I merge).

vscode/src/webview/docview.tsx
billti: Changed to namespace

vscode/src/webview/docview.tsx
DmitryVasilevsky: So indexing [2] will not go out of bounds?

vscode/src/documentation.ts
```diff
@@ -27,20 +27,24 @@ export async function showDocumentationCommand(extensionUri: Uri) {
...
     // For example, table of content is a separate file in a special format
     // We check presence of qsharp.name in metadata to make sure we take
     // only files that contain documentation from some qsharp object.
     if (file.metadata.indexOf("qsharp.name:") >= 0) {
-      documentation.push(file.contents);
+      documentation.push(file);
     }
   }
 
+  let projectName = program.programConfig.projectName ?? "App";
```
minestarks, line 41: This shouldn't need a fallback since it's a `string` (not null).

minestarks (approved):

vscode/src/documentation.ts
billti: Done

PR #1980, opened 2024-10-29
Added sample notebook with noise

DmitryVasilevsky:
Demonstrates use of Pauli noise in Python and display histogram widget in the notebook.

### DIFF
--- a/samples/notebooks/noise.ipynb
+++ b/samples/notebooks/noise.ipynb
@@ -1,0 +1,209 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using QDK to simulate Pauli noise\n",
+    "This notebook shows how to run simulations with Pauli noise, such as bit-flip or depolarizing noise.\n",
+    "\n",
+    "First, make sure prerequisites are available. Packages `qsharp` and `qsharp_widgets` must be already installed."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import qsharp\n",
+    "import qsharp_widgets\n",
+    "%pip install pandas"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Simulation with noise\n",
+    "\n",
+    "Define a simple program that creates a Bell state on two qubits and measures both qubits."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 238,
+   "metadata": {
+    "vscode": {
+     "languageId": "qsharp"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%%qsharp\n",
+    "\n",
+    "operation BellPair() : Result[] {\n",
+    "    use q = Qubit[2];\n",
+    "    H(q[0]);\n",
+    "    CNOT(q[0], q[1]);\n",
+    "    MResetEachZ(q)\n",
+    "}"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


samples/notebooks/noise.ipynb
```diff
@@ -0,0 +1,209 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import qsharp\n",
+    "import qsharp_widgets\n",
+    "%pip install pandas"
```
billti, line 21: Why do you `pip install pandas`? I don't see it used anywhere

samples/notebooks/noise.ipynb
```diff
@@ -0,0 +1,209 @@
...
+    "We can see that with 1% noise, cat state can still be clearly seen, but when noise approaches 25%, the cat state is indistinguishable from noise."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Arbitrary Pauli noise\n",
+    "\n",
+    "Standard bit-flip, phase-flip, and [depolarizing](https://en.wikipedia.org/wiki/Quantum_depolarizing_channel) noise are available, but arbitrary Pauli noise is also possible. The following example runs the same Cat5 program. First it applies noise with 20% probability, bit-flip half the time and phase-flip half the time. In a second experiment it applies Pauli-Y noise with 10% probability."
```
billti, line 172: Maybe put parens around `bit-flip half the time and phase-flip half the time` so it reads better. (Or precede with a semicolon rather than a comma).

samples/notebooks/noise.ipynb
```diff
@@ -0,0 +1,209 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using QDK to simulate Pauli noise\n",
```
billti, line 7: Add 'the' (i.e. `Using the QDK`), or go with just "Simulating Pauli noise'.

samples/notebooks/noise.ipynb
DmitryVasilevsky: Added parenthesis.

samples/notebooks/noise.ipynb
```diff
@@ -0,0 +1,209 @@
...
+   "source": [
+    "results = qsharp.run(\"BellPair()\", 20)\n",
+    "results"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that measurements always agree within a shot es expected. Now run 20 shots of the same program with 10% [depolarizing noise](https://en.wikipedia.org/wiki/Quantum_depolarizing_channel). Depolarizing noise is applied to each gate and each measurement."
```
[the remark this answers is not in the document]
DmitryVasilevsky: Fixed typo.
I think 20 shots is better here because all results are shown as a list. 100 or 1000 will be overwhelming. This is a simple example where user can look at the actual results rather than some aggregation.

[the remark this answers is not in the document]
billti: How do you decide when to show and list and when to show a histogram? Is there a reason not to just stick with one output format?

billti (approved):

samples/notebooks/noise.ipynb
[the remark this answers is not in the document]
DmitryVasilevsky: I want to start with an example that is as simple as possible. This example only runs simulation of a very simple program with a simple noise - nothing else. As a result, user sees pure output of the program. I believe this helps user understand what's going on. Once user sees how to run simulations with noise and how noise affects output, a bit more complex programs and processing gets introduced.

PR #1981, opened 2024-10-29
Add support for custom resets using the `@Reset()` attribute

orpuente-MS:
This PR adds support for custom resets using the `@Reset` attribute. It consists of the following changes:
 - Adding a `@Reset` attribute to the language.
 - Piping it down to FIR.
 - Using the attribute to generate the correct QIR, including the `"irreversible"` attribute.
 - Semantic validation passes in `qsc_passes` based on the `@Reset` attribute.
 - The right handling of reset operations during partial evaluation.
 - Unit tests.
 - Update previous tests to include the `#1` irreversible attribute in QIR when calling the Reset operation.

### DIFF
--- a/compiler/qsc_codegen/src/qir.rs
+++ b/compiler/qsc_codegen/src/qir.rs
@@ -508,7 +508,7 @@ impl ToQir<String> for rir::Callable {
                     self.call_type,
                     rir::CallableType::Measurement | rir::CallableType::Reset
                 ) {
-                    // Measurement callables are a special case that needs the irreversable attribute.
+                    // These callables are a special case that need the irreversable attribute.
                     " #1"
                 } else {
                     ""
--- a/compiler/qsc_codegen/src/qir.rs
+++ b/compiler/qsc_codegen/src/qir.rs
@@ -504,7 +504,10 @@ impl ToQir<String> for rir::Callable {
             return format!(
                 "declare {output_type} @{}({input_type}){}",
                 self.name,
-                if self.call_type == rir::CallableType::Measurement {
+                if matches!(
+                    self.call_type,
+                    rir::CallableType::Measurement | rir::CallableType::Reset
+                ) {
                     // Measurement callables are a special case that needs the irreversable attribute.
                     " #1"
                 } else {
--- a/compiler/qsc_frontend/src/lower.rs
+++ b/compiler/qsc_frontend/src/lower.rs
@@ -30,10 +30,10 @@ pub(super) enum Error {
     #[error("invalid attribute arguments: expected {0}")]
     #[diagnostic(code("Qsc.LowerAst.InvalidAttrArgs"))]
     InvalidAttrArgs(String, #[label] Span),
-    #[error("invalid use of the Measurement attribute on a function")]
+    #[error("invalid use of the {0} attribute on a function")]
     #[diagnostic(help("try declaring the callable as an operation"))]
-    #[diagnostic(code("Qsc.LowerAst.InvalidMeasurementAttrOnFunction"))]
-    InvalidMeasurementAttrOnFunction(#[label] Span),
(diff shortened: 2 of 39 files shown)


compiler/qsc_codegen/src/qir.rs
```diff
@@ -504,7 +504,10 @@ impl ToQir<String> for rir::Callable {
             return format!(
                 "declare {output_type} @{}({input_type}){}",
                 self.name,
-                if self.call_type == rir::CallableType::Measurement {
+                if matches!(
+                    self.call_type,
+                    rir::CallableType::Measurement | rir::CallableType::Reset
+                ) {
                     // Measurement callables are a special case that needs the irreversable attribute.
```
swernli, line 511: nitpick: this comment could be made more general now to match the fixed code.

```suggestion
                    // These callables are a special case that need the irreversable attribute.
```

compiler/qsc_frontend/src/lower.rs
```diff
@@ -429,29 +438,40 @@ impl With<'_> {
...
-                    self.lowerer
-                        .errors
-                        .push(Error::InvalidMeasurementAttrOnFunction(span));
-                    hir::CallableKind::Function
-                }
+        match kind {
+            ast::CallableKind::Function => {
+                self.check_invalid_attrs_on_function(attrs, span);
+                hir::CallableKind::Function
             }
```
orpuente-MS, line 468: Should this be checked during lowering or should it be part of the `qsc_passes` checking the semantics of Measurements and Resets?

swernli: I think it's ok for this one to be checked during lowering, though I suppose it could go either way. I don't feel strongly about it.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2372,6 +2375,60 @@ impl<'a> PartialEvaluator<'a> {
         result_value
     }
 
+    fn reset_qubits(
```
swernli, line 2378: I wonder if it's better to have this logic here or to instead update `eval_expr_call_to_intrinsic_qis` to take a parameter for the callable type and then just call through to that. On the one hand, this gives us a place to put other special handling if need it. On the other hand, this is basically the same as `eval_expr_call_to_intrinsic_qis` with the only difference seeming to be callable type. Unless I'm missing some other difference?

swernli (approved):
Signed off with some minor comments, nothing that I think prevents merging (especially since we may follow up with unifying Measurement and Reset handling using this infrastructure after the release).

compiler/qsc_frontend/src/lower.rs
orpuente-MS: I will have a better answer for this once Measurements and Resets are unified.

compiler/qsc_partial_eval/src/lib.rs
orpuente-MS: I will have a better answer for this once Measurements and Resets are unified. I would like to keep the logic that handles attributes together. Let's wait until the unification PR to decide on this.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2372,6 +2375,60 @@ impl<'a> PartialEvaluator<'a> {
...
+
+        let measurement_callable = Callable {
+            name: callable_decl.name.name.to_string(),
+            input_type,
+            output_type,
+            body: None,
+            call_type: CallableType::Reset,
+        };
+
+        // Resove the call arguments, create the call instruction and insert it to the current block.
```
cesarzc, line 2404:
```suggestion
        // Resolve the call arguments, create the call instruction and insert it to the current block.
```

cesarzc (approved):
Signing off, looking forward to the unification PR!

PR #1982, opened 2024-10-30
Telementry event for noisy simulation

DmitryVasilevsky:
This adds telemetry even when noisy simulation is used in vscode plugin.


billti (approved):
Thanks!

minestarks (approved):

PR #1984, opened 2024-10-30
`DumpMachine` output in Python and console should be empty with no qubits allocated

### LINKED ISSUE #1961: `DumpMachine` output in Python and console should be empty with no qubits allocated
Following up on #1922, we should change the output of `DumpMachine` in Python/Jupyter as well as console to show a message when no qubits are allocated rather than treating it the same as a single qubit in the zero state.

swernli:
This change updates the display of quantum state when no qubits are allocated to show a message rather than the same output as one qubit in the zero state. This matches the updated debugger behavior. Note that I've tried to ensure this is NOT a breaking change for consumers of the WASM/npm package by extending the data type rather than changing the behavior or contents of existing fields, and verifyied this by leaving the dev playground unchanged and confirming it preserved the old behavior there (playground will be updated in a follow up PR and is tracked by a separate issue).

Fixes #1961

image

### DIFF
--- a/npm/qsharp/src/compiler/events.ts
+++ b/npm/qsharp/src/compiler/events.ts
@@ -8,7 +8,10 @@ import { IServiceEventTarget } from "../workers/common.js";
 // Create strongly typed compiler events
 export type QscEventData =
   | { type: "Message"; detail: string }
-  | { type: "DumpMachine"; detail: { state: Dump; stateLatex: string | null } }
+  | {
+      type: "DumpMachine";
+      detail: { state: Dump; stateLatex: string | null; qubitCount: number };
+    }
   | { type: "Matrix"; detail: { matrix: number[][][]; matrixLatex: string } }
   | { type: "Result"; detail: Result };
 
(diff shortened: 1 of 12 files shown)


minestarks (approved):

npm/qsharp/src/compiler/events.ts
```diff
@@ -8,7 +8,10 @@ import { IServiceEventTarget } from "../workers/common.js";
 // Create strongly typed compiler events
 export type QscEventData =
   | { type: "Message"; detail: string }
-  | { type: "DumpMachine"; detail: { state: Dump; stateLatex: string | null } }
+  | {
+      type: "DumpMachine";
+      detail: { state: Dump; stateLatex: string | null; qubitCount: number };
+    }
```
billti, line 14: This _might_ be a breaking change for QCOM. Either way, let's get 1.10 out and then merge this.

billti (approved):

billti (approved):

PR #1985, opened 2024-10-30
Unify implementations of custom measurements and custom resets

orpuente-MS:
This PR unifies the implementations of custom measurements and custom resets. It does two things:
 - Removes the `CallableKind::Measurement` variant from HIR and FIR.
 - Uses the `Attr::Measurement` to identify that a callable is a measurement.

### DIFF
--- a/compiler/qsc_doc_gen/src/display.rs
+++ b/compiler/qsc_doc_gen/src/display.rs
@@ -158,7 +158,7 @@ impl HirCallableDecl<'_> {
     fn get_param_offset(&self) -> u32 {
         let offset = match self.decl.kind {
             hir::CallableKind::Function => "function".len(),
-            hir::CallableKind::Operation | hir::CallableKind::Measurement => "operation".len(),
+            hir::CallableKind::Operation => "operation".len(),
         } + 1 // this is for the space between keyword and name
         + self.decl.name.name.len()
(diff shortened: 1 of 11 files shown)


minestarks (approved):
I don't know the motivation behind the change, so I can't judge it based on that, but the code changes look good to me.

orpuente-MS:
> I don't know the motivation behind the change, so I can't judge it based on that, but the code changes look good to me.

Just to leave some context documented here. When I implemented custom measurements a couple of PRs ago, I encoded the fact that an operation was a measurement as the new (now removed) `CallableKind::Measurement` variant. It looked like a good idea at the moment because measurements are treated different in theory and literature.

In the next PR when adding reset operations, an edge case came afloat, the `MResetZ` operation, which is a measurement and an operation at the same time. So, taking @swernli suggestion, instead of adding more variants to the `CallableKind` enum, I just piped down the attributes of a callable (`@Measurement`, `@Reset`, etc.) and used that information to determine if an operation was a reset, which happened to be simpler and more scalable when we add new attributes.

In this PR I just undid the addition of the `CallableKind::Measurement` variant and used the infrastructure already added for resets (the piped-down attributes) to determine if an operation is a measurement.

swernli (approved):

PR #1986, opened 2024-10-30
Disable completions in attribute arguments

minestarks:
Disable completions inside attribute arguments, despite what the parser says, since we support either hardcoded values or nothing at all.

examples:

```qsharp
@Config( | ) 
operation Foo() : Unit {}
```

```qsharp
@EntryPoint( | ) 
operation Foo() : Unit {}
```

no completions at this locations.

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+mod ast_context;
 mod fields;
 mod global_items;
 mod locals;
-mod path_context;
 #[cfg(test)]
 mod tests;
 mod text_edits;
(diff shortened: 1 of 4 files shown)


language_service/src/completion.rs
```diff
@@ -131,6 +139,11 @@ fn collect_hardcoded_words(expected: WordKinds) -> Vec<Completion> {
                 completions.extend([
                     Completion::new("EntryPoint".to_string(), CompletionItemKind::Interface),
                     Completion::new("Config".to_string(), CompletionItemKind::Interface),
+                    Completion::new("Unimplemented".to_string(), CompletionItemKind::Interface),
```
swernli, line 142: Do want people to know about `Unimplemented`? Given the uses for it, I wonder if it would be better to leave it out. We introduced it in case we want to have more specific errors for library functions we intentionally haven't implemented, but we never actually used it anywhere.

swernli (approved):

orpuente-MS (approved):

PR #1987, opened 2024-10-30
More parser error recovery, unlocking completions in more locations

### LINKED ISSUE #1960: Broken expression node at the end of a cell breaks the AST
In VS Code, in a Jupyter notebook cell, type some code that ends with an incomplete expression, like:

```qsharp
%%qsharp

function Foo() : Unit {}
let abc = Foo();
let x = 
```

Try using editor features like hover, go-to-def for any of the valid code above the last line.

**Expected**: Code above the last broken line works

**Actual**: The entire cell seems to be broken.

minestarks:
1. Callable signatures tolerate a lot more broken syntax now. This allows us to complete type parameters correctly in broken syntax that we commonly see when the user is in the middle of typing:

```qsharp
operation Test<'T>() : () => |
```
```qsharp
operation Test<'T>(x: |)
```
```qsharp
operation Test<'T>() : |
```
```qsharp
operation Test<'T>(x: Std.|) : Unit {}
```

2. Error recovery in notebook cells if a statement can't be parsed. Fixes #1960 and allows completions to work in scenarios like this:
```qsharp
%%qsharp

function Foo() : Unit {}
let x = |  // expect Foo in the list
```

### DIFF
--- a/compiler/qsc_frontend/src/typeck/tests.rs
+++ b/compiler/qsc_frontend/src/typeck/tests.rs
@@ -4555,7 +4555,7 @@ fn expr_incomplete_field_access_no_semi() {
         &expect![[r##"
             #14 55-57 "()" : Unit
             #18 65-99 "{\n        (new B { C = 5 }).\n    }" : ?
-            #20 75-93 "(new B { C = 5 })." : ?
+            #20 75-98 "(new B { C = 5 }).\n    " : ?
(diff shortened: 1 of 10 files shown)


swernli (approved):
This is a really nice change, thanks! Of note for those who think about perf, I did run comparisons against main and saw no impact on parser performance.

orpuente-MS (approved):
The code and logic seem fine. Approved. These are amazing improvements. Thanks!

PR #1990, opened 2024-10-31
Add telemetry events for interop

idavis:


sezna (approved):

pip/qsharp/interop/qiskit/backends/backend_base.py
```diff
@@ -277,7 +277,9 @@ def run_job(
         output["status"] = "COMPLETED"
         output["backend_name"] = self.name
         output["backend_version"] = self.backend_version
-        output["time_taken"] = str(time.time() - start)
+
+        duration = datetime.timedelta(seconds=(monotonic() - start))
```
billti, line 281: I assume you verified the resulting string is that same as was previously done on line 280 (or that it's not breaking if not)

pip/qsharp/interop/__init__.py
```diff
@@ -1,2 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
+
+from .. import telemetry_events
+
+
+telemetry_events.on_qiskit_import()
```
billti, line 7: Do we care about them just importing the 'interop' module?

pip/qsharp/interop/qiskit/jobs/qsjob.py
```diff
@@ -15,7 +16,7 @@
 from qiskit.providers import JobV1, JobStatus, JobError
 
 from ..execution import DetaultExecutor
-
+from ... import telemetry_events
```
billti, line 19: Is 3 dots enough here? Seems like that would get you to the ./interop dir.

pip/qsharp/interop/qiskit/jobs/qsjobset.py
```diff
@@ -17,6 +18,7 @@
 
 from .qsjob import QsSimJob, RunInputCallable
 from ..execution import DetaultExecutor
+from ... import telemetry_events
```
billti, line 21: Weird if it's working. Maybe I just don't understand the syntax well enough. Seems like one dot is current (jobs) dir, two would be 'qiskit' dir, and three would only get you to 'interop' dir. Not sure why it doesn't need four to get to the root from here.

pip/qsharp/telemetry_events.py
```diff
@@ -85,3 +85,42 @@ def on_estimate_end(durationMs: float, qubits: Union[str, int]) -> None:
...
+
+def on_qiskit_import() -> None:
+    log_telemetry("qiskit.import", 1)
+
+
+def on_qiskit_run(shots: int, num_circuits: int) -> None:
+    log_telemetry(
+        "qiskit.run",
+        1,
+        properties={"shots": get_shots_bucket(shots), "num_circuits": num_circuits},
```
billti, line 101: What is the range for 'num_circuits'? Properties on metrics should be a limited set of value (which is why we bucket shots and qubits). Could this end up being dozens (or more) of distinct values?

pip/qsharp/interop/qiskit/jobs/qsjobset.py
idavis: Two or more leading dots indicate a relative import to the parent(s) of the current package, one level per dot after the first. The parent of the current package is `qiskit`.  Two levels up (due to three dots) gets us to the root package.

pip/qsharp/interop/qiskit/backends/backend_base.py
idavis: It was on my list to double check. You were right, the old value was float in secs. Could have sworn `time()` returned a struct for the `-` to give formatting.

pip/qsharp/interop/qiskit/jobs/qsjobset.py
billti: Pretty sure 3 dots isn't getting you to the root level. It's getting you to qsharp/interop. However, in qsharp/interop/__init__.py you have `from .. import telemetry_events` so it's finding that re-export.

idavis: You are correct

billti (approved):

PR #1992, opened 2024-10-31
Bump version to 1.10

billti:
And make it type check clean

### DIFF
--- a/version.py
+++ b/version.py
@@ -4,14 +4,15 @@
 # Licensed under the MIT License.
 
 import os
-from datetime import datetime, timezone
 
 # To be updated every time we start a new major.minor version.
-major_minor = "1.9"
+major_minor = "1.10"
 
 # Default to 'dev' builds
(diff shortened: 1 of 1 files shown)


swernli (approved):

minestarks (approved):

version.py
```diff
@@ -4,14 +4,15 @@
...
 
 # To be updated every time we start a new major.minor version.
-major_minor = "1.9"
+major_minor = "1.10"
 
 # Default to 'dev' builds
 BUILD_TYPE = os.environ.get("BUILD_TYPE") or "dev"
 BUILD_NUMBER = os.environ.get("BUILD_NUMBER")
+if not BUILD_NUMBER:
+    raise Exception("BUILD_NUMBER environment variable must be set")
```
DmitryVasilevsky, line 15: Maybe change the error message below ("BUILD_NUMBER environment variable must be set to a valid integer") so that it doesn't look almost exactly like this one to avoid confusion when analyzing logs?

DmitryVasilevsky (approved):

PR #1995, opened 2024-10-31
Add custom operations sample

orpuente-MS:
This PR adds a new language sample showing how to use custom measurements and a more comprehensive sample showing how to use custom operations (intrinsics and measurements).


samples/algorithms/MajoranaQubits.qs
```diff
@@ -0,0 +1,102 @@
+/// # Sample
+/// Majorana Qubits
+///
+/// # Description
+/// In hardware providing majorana qubits, common quantum operations
+/// are implemented using measurements and Pauli corrections. This
+/// sample shows a hypotetical hardware provider exposing some custom
+/// gates to Q# and a small library built on top of it.
+
+namespace Main {
```
swernli, line 10: Rather than have a sample that brings back multiple namespaces in a file, I think it might make more sense to have this as a multifile sample in a folder.

samples/algorithms/MajoranaQubits.qs
```diff
@@ -0,0 +1,102 @@
...
+        let x = HardwareProvider.Mxx(q1, q2);
+        (x, z)
+    }
+}
+
+/// A set of custom measurements exposed from a hardware
+/// provider using Majorana Qubits.
+namespace HardwareProvider {
+    @SimulatableIntrinsic()
+    operation X(q : Qubit) : Unit {
```
swernli, line 69: I'm torn between making this friendly/readable and following convention... while this will work fine as-is, it will produce QIR that has the function signature `void @X(%Qubit*)` which does not conform the naming conventions in the QIR spec. If we want to encourage someone to look at the QIR and compare it to what comes out of the compiler normally (and in theory show them how it would look if they were doing this with the intent of submitting it to a spec-compliant provider), I think these should be named things like `__quantum__qis__x__body` and `__quantum__qis__mx__body` etc.

orpuente-MS: Yes, I agree with educating the community on what QIR compliance looks like. Will update to reflect that mindset.

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,62 @@
...
+operation CNOT(control : Qubit, target : Qubit) : Unit {
+    use ancilla = Qubit();
+    let a = Mzz(control, target);
+    let b = Mxx(target, ancilla);
+    let c = Mx(target);
+    if b == One {
+        Z(control);
+    }
+    if [a, c] == [One, One] {
+        X(ancilla);
```
billti, line 19: Not sure if we need some explanatory comments here (or it we expect folks have read the paper), but what's the point of doing an X on 'ancilla' here? It's about to go out of scope and nothing further is done with it in this operation.

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+/// Prepare a Bell Pair.
+/// Source: [1] Figure 18a.
+operation BellPair(q1 : Qubit, q2 : Qubit) : Unit {
+    // Bring the qubits to their ground state.
```
billti, line 26: But you don't flip them if the Mz measures 'One', so how is this ensuring they are in ground state? (The simulatable instrinsic also doesn't ensure they are Zero)

samples_test/src/tests/language.rs
```diff
@@ -144,6 +144,8 @@ pub const COPYANDUPDATEOPERATOR_EXPECT_DEBUG: Expect = expect![[r#"
     Updated array: [10, 11, 100, 13]
     Updated array: [10, 100, 12, 200]
     ()"#]];
+pub const CUSTOMMEASUREMENTS_EXPECT: Expect = expect!["Zero"];
+pub const CUSTOMMEASUREMENTS_EXPECT_DEBUG: Expect = expect!["Zero"];
```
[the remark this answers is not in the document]
orpuente-MS: Because this PR is also adding a small sample to the languages folder, and we have some automated code that generates tests for those.

samples/algorithms/MajoranaQubits/src/GateSet.qs
orpuente-MS: My bad. The ancilla qubit was the middle one in the diagram. I read it wrong.

samples/algorithms/MajoranaQubits/src/GateSet.qs
orpuente-MS: 'Ground state' was the wrong thing to say here. Is more like it's collapsing the Z axis, projecting them to the X basis, where the next Mxx is performed.

orpuente-MS: I should definitely understand how this works and include comments explaining it.

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,65 @@
...
+/// Source:
+///  [1] Surface code compilation via edge-disjoint paths
+///      https://arxiv.org/pdf/2110.11493
+
+/// Apply a CNOT gate to the given qubits.
+/// Source: [1] Figure 3.
+operation CNOT(control : Qubit, target : Qubit) : Unit {
+    // Prepare an ancilla qubit in the |+> state.
+    use ancilla = Qubit();
+    H(ancilla);
```
swernli, line 13: In keeping with the measurement based compuation theme, this can be Mx with a conditional Z:

```suggestion
    if Mx(ancilla) == One {
        Z(ancilla);
    }
```

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,65 @@
...
+
+    let a = Mzz(control, ancilla);
+    let b = Mxx(ancilla, target);
+    let c = Mx(ancilla);
+    if b == One {
+        Z(control);
+    }
+    if [a, c] == [One, One] {
+        X(target);
+    }
```
swernli, line 23: I think this needs adjusting... unless the resource you are using defines it differently, I've always seen it used this way:

```suggestion
    let c = Mz(ancilla);
    if b == One {
        Z(control);
    }
    if a != c {
        X(target);
    }
```

swernli: with that change, you'll probably need to add a reset of the ancilla before it goes out of scope.

samples/algorithms/MajoranaQubits/src/Main.qs
```diff
@@ -0,0 +1,17 @@
...
+/// sample shows a hypotetical hardware provider exposing some custom
+/// gates to Q# and a small library built on top of it.
+
+/// Sample program using gates for custom hardware provider.
+operation Main() : (Result, Result) {
+    use qs = Qubit[2];
+    GateSet.BellPair(qs[0], qs[1]);
+    let res = GateSet.BellMeasurement(qs[0], qs[1]);
+    ResetAll(qs);
+    res
```
swernli, line 16: At the risk of complicating things, I think it could be helpful to show how the Bell measurement correctly distinguishes between each of the four bell states. You could do this by using `BellPair` and `BellMeasurement` to generate the first value, which is always (Zero, Zero), then show that doing an `X` in between to one qubit takes you to a distinct measured Bell state, then a `Z`, then both `X` and `Z`, completing the set of four states.

samples/algorithms/MajoranaQubits/src/GateSet.qs
orpuente-MS: I went with your suggestion, but wrapped that logic in a `PreparePlus` operation.

samples/algorithms/MajoranaQubits/qsharp.json
```diff
@@ -0,0 +1,12 @@
...
+    "lints": [
+        {
+            "lint": "deprecatedWithOperator",
+            "level": "error"
+        },
+        {
+            "lint": "deprecatedDoubleColonOperator",
+            "level": "error"
+        }
+    ]
```
swernli, line 11: I'm torn on this one... while it's good to show this can be done, since this sample doesn't use any of the new struct syntax that replaces this syntax, these lints are fairly unrelated.

orpuente-MS: Removed. I had copied it from the MultiFileProject sample in the language folder.

samples/algorithms/MajoranaQubits/src/Main.qs
orpuente-MS: I just tried this, and it worked beautifully. Definitely worth adding, I'm just trying to come up with the right way to include it in the sample. Maybe as commented lines that user can uncomment before running the code again?

swernli: If this were just a sample for simulation, I would suggest including all of them or even a loop or something, and then returning a list of tuples. But given that this is a sample where we want to encourage the user to generate QIR and see how the custom measurements show up, I think keeping the `Main` smaller is best, so I like your idea of having the commented out lines and some explanation there!

orpuente-MS: I did that for now, but I'm open to other suggestions

orpuente-MS: I did that. Open to suggestions to make those commented lines more appealing to the eye 😄.

swernli: I think the comments look great!

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,66 @@
...
+    if a != c {
+        X(target);
+    }
+}
+
+/// Prepare a Bell Pair.
+/// Source: [1] Figure 18a.
+operation BellPair(q1 : Qubit, q2 : Qubit) : Unit {
+    // Collapse the qubits to the X basis.
```
swernli, line 40: nitpick: I think this might be collapsing to the Z basis, but there might be different conventions about how to describe this. Based on my understanding, I'd phrase this as:
```suggestion
    // Project each of the qubits onto the Pauli-Z basis.
```

swernli (approved):

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,76 @@
...
+/// Below is a map showing how the Bell states map to the Result pairs:
+///   |𝚽⁺⟩ -> (Zero, Zero)
+///   |𝚿⁺⟩ -> (Zero, One)
+///   |𝚽⁻⟩ -> (One, Zero)
+///   |𝚿⁻⟩ -> (One, One)
+operation BellMeasurement(q1 : Qubit, q2 : Qubit) : (Result, Result) {
+    let z = Mzz(q1, q2);
+    let x = Mxx(q1, q2);
+    (x, z)
+}
```
orpuente-MS, line 61: I added how Bell states are mapped to Result pairs to the docstring. Not sure if this is needed.

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,76 @@
...
+        Z(control);
+    }
+
+    if a != c {
+        X(target);
+    }
+}
+
+
+/// Prepare a qubit in the |+⟩ state.
```
msoeken, line 30: Should we add a PrepereZero for symmetry? How about naming them PrepareX and PrepareZ instead.

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,76 @@
...
+/// Apply a CNOT gate to the given qubits.
+/// Source: [1] Figure 3.
+operation CNOT(control : Qubit, target : Qubit) : Unit {
+    // Prepare an ancilla qubit in the |+⟩ state.
+    use ancilla = Qubit();
+    PreparePlus(ancilla);
+
+    let a = Mzz(control, ancilla);
+    let b = Mxx(ancilla, target);
+    let c = M(ancilla);
```
msoeken, line 17: I think for symmetry it would be nice to have a custom gate called `Mz`, even though it's being implemented using `M`.

msoeken (approved):

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,82 @@
...
+    (x, z)
+}
+
+/// User friendly wrapper around the Mx hardware gate.
+operation Mx(q : Qubit) : Result {
+    HardwareIntrinsics.__quantum__qis__mx__body(q)
+}
+
+/// User friendly wrapper around the Mz hardware gate.
+operation Mz(q : Qubit) : Result {
```
swernli, line 70:
```suggestion
    HardwareIntrinsics.__quantum__qis__mz__body(q)
```

samples/algorithms/MajoranaQubits/src/GateSet.qs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    if a != c {
+        X(target);
+    }
+}
+
+
+/// Prepare a qubit in the |+⟩ state.
+/// `q` must be in the |0⟩ state.
```
swernli, line 31: This isn't true anymore, actually. Because this uses Mx, it is guaranteed to project to the X basis regardless of where the qubit started.

samples/algorithms/MajoranaQubits/src/GateSet.qs
orpuente-MS: I think the GitHub suggestion got confused here. You meant to edit line 71, right? I will edit it manually.

PR #1997, opened 2024-11-01
Noise settings available in Q#

DmitryVasilevsky:
Added ConfigurePauliNoise and ApplyIdleNoise to Diagnostric namespace to control noise from Q# program.

### DIFF
--- a/compiler/qsc/src/interpret/circuit_tests.rs
+++ b/compiler/qsc/src/interpret/circuit_tests.rs
@@ -514,6 +514,34 @@ fn custom_intrinsic_mixed_args() {
 }
 
 #[test]
+fn custom_intrinsic_apply_idle_noise() {
+    let mut interpreter = interpreter(
+        r"
+    namespace Test {
+        import Std.Diagnostics.*;
+        @EntryPoint()
+        operation Main() : Unit {
+            ConfigurePauliNoise(BitFlipNoise(1.0));
+            use q = Qubit();
+            ApplyIdleNoise(q);
+        }
+    }",
+        Profile::Unrestricted,
+    );
+
+    let circ = interpreter
+        .circuit(CircuitEntryPoint::EntryPoint, false)
+        .expect("circuit generation should succeed");
+
+    // ConfigurePauliNoise has no qubit qrguments so it shouldn't show up.
+    // ApplyIdleNoise is a quantum operation so it shows up.
+    expect![[r#"
+        q_0     ApplyIdleNoise
+    "#]]
+    .assert_eq(&circ.to_string());
+}
+
+#[test]
 fn operation_with_qubits() {
     let mut interpreter = interpreter(
         r"
--- a/compiler/qsc/src/interpret/circuit_tests.rs
+++ b/compiler/qsc/src/interpret/circuit_tests.rs
@@ -533,7 +533,7 @@ fn custom_intrinsic_apply_idle_noise() {
         .circuit(CircuitEntryPoint::EntryPoint, false)
         .expect("circuit generation should succeed");
 
-    // ConfigurePauliNoise has no qubit qrguments so it shouldn't show up.
+    // ConfigurePauliNoise has no qubit arguments so it shouldn't show up.
     // ApplyIdleNoise is a quantum operation so it shows up.
     expect![[r#"
         q_0     ApplyIdleNoise
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -137,14 +137,17 @@ impl SparseSim {
 
     #[must_use]
     pub fn new_with_noise(noise: &PauliNoise) -> Self {
-        Self {
-            sim: QuantumSim::new(None),
-            noise: *noise,
-            rng: if noise.is_noiseless() {
-                None
-            } else {
-                Some(StdRng::from_entropy())
-            },
+        let mut sim = SparseSim::new();
+        sim.set_noise(noise);
+        sim
+    }
+
+    fn set_noise(&mut self, noise: &PauliNoise) {
+        self.noise = *noise;
+        if noise.is_noiseless() {
+            self.rng = None;
+        } else {
+            self.rng = Some(StdRng::from_entropy());
         }
     }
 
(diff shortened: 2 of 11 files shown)


library/std/src/Std/Diagnostics.qs
```diff
@@ -368,6 +368,46 @@ operation StopCountingQubits() : Int {
...
+/// Use 0.0 for all parameters to simulate without noise.
+///
+/// # Input
+/// ## px
+/// Probability of applying X gate.
+/// ## py
+/// Probability of applying Y gate.
+/// ## pz
+/// Probability of applying Z gate.
+@Config(Unrestricted)
```
minestarks, line 388: AFAIK, these shouldn't be limited to `Unrestricted`, because we can still run the simulator in `Base` or `Adaptive` configuration, and applying noise would still be a valid scenario.

compiler/qsc_eval/src/backend.rs
```diff
@@ -393,6 +396,26 @@ impl Backend for SparseSim {
             | "AccountForEstimatesInternal"
             | "BeginRepeatEstimatesInternal"
             | "EndRepeatEstimatesInternal" => Some(Ok(Value::unit())),
+            "ConfigurePauliNoise" => {
```
minestarks, line 399: IIRC we need to handle custom intrinsics in other `Backend` implementations as well, perhaps by no-opping them. To check, you can try generating a Circuit. How does it handle this custom call?

minestarks (commented):
I like the naming.

Shall we consider raising an event (similar to other diagnostic events like `Message`, `DumpMachine`, etc) so that the host (i.e. VS Code, Python etc) can inform the user that noise has been applied, if it chooses to?

This could be used to populate accurate info in the histogram view (#1975) or simply just output a message during execution like "noise was applied".

compiler/qsc_eval/src/backend.rs
swernli: Yes, I agree, these new noise intrinsics should be recognized and treated as no-op in circuit generation, partial evaluation, and resource estimation.

library/std/src/Std/Diagnostics.qs
swernli: Yeah, since these do not return a value other than `Unit` they can be safely treated as no-op in QIR codegen (they will never affect control flow) and as such can have the config restriction removed. This is similar to how `DumpMachine()` and `DumpRegister()` are treated.

compiler/qsc_eval/src/backend.rs
DmitryVasilevsky: I believe I updated them all.
Ideally, we shouldn't update other backends when we add a custom intrinsic to one. But we aren't ready for this yet with our current architecture. We discussed this with Mine.

library/std/src/Std/Diagnostics.qs
DmitryVasilevsky: Removed Unrestricted restriction :)
Original thought was that this only valid for simulator, so we would want to have this in unrestricted only. But we do simulate other profile.

compiler/qsc_partial_eval/src/management.rs
```diff
@@ -133,7 +133,7 @@ impl Backend for QuantumIntrinsicsChecker {
     ) -> Option<std::result::Result<Value, String>> {
         match name {
             "BeginEstimateCaching" => Some(Ok(Value::Bool(true))),
-            "EndEstimateCaching" | "GlobalPhase" => Some(Ok(Value::unit())),
+            "EndEstimateCaching" | "GlobalPhase" | "ConfigurePauliNoise" => Some(Ok(Value::unit())),
```
swernli, line 136: looks like `ApplyIdleNoise` is missing from this one.

swernli (approved):

compiler/qsc_partial_eval/src/management.rs
swernli: I think the comment mentioning "functions" is a bit misleading, because `GlobalPhase` is also an operation and like `ApplyIdleNoise` is a no-op for codegen:
https://github.com/microsoft/qsharp/blob/5ac52cd3e105dbdecc26fe93966683f878974b57/library/std/src/Std/InternalHelpers.qs#L67-L69
They both fall into a special category of quantum operations that are simulator specific and don't need to be generated into QIR, so I think they should both end up here.

DmitryVasilevsky: Yes, this makes sense! I added it here AND in lib.rs to be ignored for partial evaluation. That's what was missing.

compiler/qsc_eval/src/val.rs
```diff
@@ -263,6 +263,14 @@ impl Value {
         v
     }
 
+    #[must_use]
+    pub fn get_double(&self) -> f64 {
```
minestarks, line 267: Suggest renaming to `unwrap_double` to stick with the naming convention here.

compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -513,6 +513,34 @@ fn custom_intrinsic_mixed_args() {
...
+        }
+    }",
+        Profile::Unrestricted,
+    );
+
+    let circ = interpreter
+        .circuit(CircuitEntryPoint::EntryPoint, false)
+        .expect("circuit generation should succeed");
+
+    // ConfigurePauliNoise has no qubit qrguments so it shouldn't show up.
```
minestarks, line 536:
```suggestion
    // ConfigurePauliNoise has no qubit arguments so it shouldn't show up.
```

compiler/qsc_eval/src/val.rs
DmitryVasilevsky: We discussed this with Stefan. There's already unwrap_double, but it consumes the argument (so requires copy/move of an argument in my case). In my case borrowing (instead of consuming) works just fine and that's why I need this get function.

compiler/qsc_eval/src/backend.rs
```diff
@@ -393,6 +399,26 @@ impl Backend for SparseSim {
...
+                let [xv, yv, zv] = &*arg.unwrap_tuple() else {
+                    panic!("tuple arity for SetPauliNoise intrinsic should be 3");
```
minestarks, line 404: nit

```suggestion
                    panic!("tuple arity for ConfigurePauliNoise intrinsic should be 3");
```

minestarks (approved):
Thanks!

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1344,6 +1344,7 @@ impl<'a> PartialEvaluator<'a> {
             | "AccountForEstimatesInternal"
             | "BeginRepeatEstimatesInternal"
             | "EndRepeatEstimatesInternal"
+            | "ApplyIdleNoise"
```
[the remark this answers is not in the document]
DmitryVasilevsky: AFAIK, this is used in QIR generation. ConfigurePauliNoise is a classical function and doesn't appear in the generated QIR anyway. ApplyIdleNoise was appearing in QIR until I added it here.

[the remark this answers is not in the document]
DmitryVasilevsky: I verified this by stepping through the code. I also added test to make sure it's true.

PR #1999, opened 2024-11-01
No completions in comments

minestarks:
Suppress completions inside comments like

```qsharp
import Foo;
// Hello there |
import Bar;
```

### DIFF
--- a/language_service/src/completion.rs
+++ b/language_service/src/completion.rs
@@ -102,6 +102,16 @@ fn expected_word_kinds(
     source_contents: &str,
     cursor_offset: u32,
 ) -> WordKinds {
+    // We should not retun any completions in comments.
+    // This compensates for a bug in [`possible_words_at_offset_in_source`] .
+    // Ideally, that function would be aware of the comment context and not
+    // return any completions, however this is difficult to do today because
+    // of the parser's unawareness of comment tokens.
+    // So we do a simple check here where we have access to the full source contents.
+    if in_comment(source_contents, cursor_offset) {
+        return WordKinds::empty();
+    }
+
     match &compilation.kind {
         CompilationKind::OpenProject {
             package_graph_sources,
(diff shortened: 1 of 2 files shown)


idavis (approved):

language_service/src/completion.rs
```diff
@@ -121,6 +131,16 @@ fn expected_word_kinds(
     }
 }
 
+fn in_comment(source_contents: &str, cursor_offset: u32) -> bool {
+    // find the last newline before the cursor
+    let last_line_start = source_contents[..cursor_offset as usize]
+        .rfind('\n')
+        .unwrap_or(0);
+    // find the last comment start before the cursor
+    let last_comment_start = source_contents[last_line_start..cursor_offset as usize].rfind("//");
```
swernli, line 140: Is it possible this logic could get confused by string literals, as in:
```qsharp
let x = "// this is in a string" + StrUtils.|
```

DmitryVasilevsky, line 140: Would this be OK if this is in a literal string?

billti, line 140: Probably not worth worrying about, but I assume `//` anywhere in the line will make it think it is in a comment (i.e. even if those chars are in a string literal - which seems like a low enough probability edge case not to woryy about).

orpuente-MS: Well, that line doesn't start with `//` (it might've been a GitHub antic). I think you meant something like,

```qsharp
let x = "foo
// this is in a string" + StrUtils.|
```

orpuente-MS: Yes, it shouldn't be that common. I think it is worth it to commit to this now and file an issue for an improvement.

orpuente-MS (approved):
Approving. I think this solution is good enough for now, but I guess merging depends on general consensus.

language_service/src/completion.rs
minestarks: Per side discussion - this is a valid concern, but requires a bit larger fix. Instead of patching up this case, I'd rather spend a bit more time to look at how we fix the root cause (in the lexer itself). I'll open an issue for it.

minestarks: Per side discussion - this is a valid concern, but requires a bit larger fix. Instead of patching up this case, I'd rather spend a bit more time to look at how we fix the root cause (in the lexer itself). I'll open an issue for it.

minestarks: Per side discussion - this is a valid concern, but requires a bit larger fix. Instead of patching up this case, I'd rather spend a bit more time to look at how we fix the root cause (in the lexer itself). I'll open an issue for it.

minestarks:
Created #2000 for context

PR #2001, opened 2024-11-01
Doc Gen Tool Generates Unrestricted Items

### LINKED ISSUE #1998: Docs Generation Does Not Include Unrestricted Items
The docs generation tool does not generate docs for library items listed as `@Config(Unrestricted)`. These items should be included in the docs generation and the table of contents.

ScottCarda-MS:
This fixes #1998. This also updates the table of contents to put Unstable namespaces at the end.


minestarks:
@ScottCarda-MS  what about `@Config(Base)` / `Config(Adaptive)` items? Are we intentionally excluding them now? 

(Acknowledging that exposing _all_ configurations sounds difficult, given the current design of conditional compilation. So a compromise may be needed here).

ScottCarda-MS:
> @ScottCarda-MS what about `@Config(Base)` / `Config(Adaptive)` items? Are we intentionally excluding them now?
> 
> (Acknowledging that exposing _all_ configurations sounds difficult, given the current design of conditional compilation. So a compromise may be needed here).

This should include all items, including Base and Adaptive. I added a test for finding the Adaptive `Std.Intrinsic.AND` operation. I couldn't find any Base items in our library to test against.

minestarks:
@ScottCarda-MS 
But `AND` has a `@Config(not Adaptive)` implementation as well, are you sure that's not the one you're getting?

```qsharp
@Config(not Adaptive)
operation AND(control1 : Qubit, control2 : Qubit, target : Qubit) : Unit is Adj {
    PhaseCCX(control1, control2, target);
}
```

Would doc generation work for items that have a `Adaptive`/`Base` implementation and nothing else?

DragosDanielBoia:

DragosDanielBoia:

ScottCarda-MS:
> Would doc generation work for items that have a `Adaptive`/`Base` implementation and nothing else?

I'm not able to tell at this point as there is no such content in the existing library code. Trying to update the libraries to include such code is running into issues. I'm going to save this for a future PR if/when we add such content to the libraries.

Since the existing test doesn't determine what I thought it did (as you pointed out, AND has both an Adaptive and not Adaptive definition), I'm removing the test.

ScottCarda-MS:
Closing in favor of #2012

PR #2004, opened 2024-11-05
Add eval and cell events

idavis:


pip/qsharp/_qsharp.py
```diff
@@ -243,7 +243,15 @@ def callback(output: Output) -> None:
                 pass
         print(output, flush=True)
 
-    return get_interpreter().interpret(source, callback)
+    telemetry_events.on_eval()
+    start_time = monotonic()
+
+    results = get_interpreter().interpret(source, callback)
```
minestarks, line 249: I forget, does this throw if the Q# code fails? Do we want to send the end event then?

minestarks: I'll leave it up to the folks who'll be analyzing the telemetry.

minestarks (approved):

swernli (approved):

PR #2005, opened 2024-11-05
Add class constraints for built-in classes

sezna:
This PR exposes some relevant built-in classes via class constraints.

# Changes
- Add `TypeParameter`, `ClassConstraint`, and `ConstraintParameter` AST nodes to represent the notion of a generic type that might have class constraints, and constraints that may have parameters (e.g. `Exp[Int]`).
  -  update visitors accordingly
- `Ty` can now contain arbitrary constraints
- Add notion of primitive classes to the AST
- Update constraint checking algorithm to check type parameters for primitive constraints at callable decl time
- Update constraint generation algorithm to add additional user-specified constraints to arguments that are passed in to callables (`fn constrained_ty`)
- Add completions for primitive classes to the language service
- Adds a sample describing class constraints

tiny changes:
- Add docstrings discretionally
- Remove some unnecessary derives
- Tried to clarify and unify jargon around classes, constraints, parameters, and arguments in general
- Some expect_tests were updated in minor ways as their underlying types changed (`Ty` and `TyParam`, for example)
- Improve an error message for `MissingTy` (it mentioned something about global types not being inferred when 99% of the time it does not apply to a global type, it applies to a callable parameter with a missing type)

# TODO for this PR
- [ ] Track down why duplicate constraints are getting generated
I did not want to block reviewing this PR on this bug, since it isn't an experience breaker, but I would like to get it resolved before this goes in. Sometimes, multiple errors are rendered, and I've narrowed it down to duplicate constraints getting generated. I could store constraints in some sort of `Set`, so they're deduplicated, but I'd rather figure out what we are repeatedly visiting to generate those constraints instead of deduplicating afterwards. I'll be tracking this down while the PR is reviewed.


Cargo.toml
```diff
@@ -80,6 +80,7 @@ pyo3 = "0.22"
 quantum-sparse-sim = { git = "https://github.com/qir-alliance/qir-runner", tag = "v0.7.4" }
 async-trait = "0.1"
 tokio = { version = "1.35", features = ["macros", "rt"] }
+itertools = { version = "0.10" }
```
swernli, line 83: Looks like this got added but isn't used anywhere?

sezna:
I opened #2007 to replace this, since there are too many commits and it is hiding comments.

PR #2007, opened 2024-11-05
Add class constraints for built-in classes

sezna:
This PR exposes some relevant built-in classes via class constraints.

# Changes
- Add `TypeParameter`, `ClassConstraint`, and `ConstraintParameter` AST nodes to represent the notion of a generic type that might have class constraints, and constraints that may have parameters (e.g. `Exp[Int]`).
  -  update visitors accordingly
- `Ty` can now contain arbitrary constraints
- Add notion of primitive classes to the AST
- Update constraint checking algorithm to check type parameters for primitive constraints at callable decl time
- Update constraint generation algorithm to add additional user-specified constraints to arguments that are passed in to callables (`fn constrained_ty`)
- Add completions for primitive classes to the language service
- Adds a sample describing class constraints

tiny changes:
- Add docstrings discretionally
- Remove some unnecessary derives
- Tried to clarify and unify jargon around classes, constraints, parameters, and arguments in general
- Some expect_tests were updated in minor ways as their underlying types changed (`Ty` and `TyParam`, for example)
- Improve an error message for `MissingTy` (it mentioned something about global types not being inferred when 99% of the time it does not apply to a global type, it applies to a callable parameter with a missing type)


compiler/qsc_frontend/src/typeck/convert.rs
```diff
@@ -189,18 +227,48 @@ pub(super) fn ast_ty_def(names: &Names, def: &TyDef) -> (UdtDef, Vec<MissingTyEr
...
+    names: &Names,
+    generics: &[ast::TypeParameter],
+) -> (Vec<qsc_hir::ty::GenericParam>, Vec<TyConversionError>) {
+    let mut errors = Vec::new();
+    let mut generics_buf = Vec::with_capacity(generics.len());
+    for param in generics {
+        let (bounds, new_errors) =
+            class_constraints_from_ast(names, &param.constraints, &mut Default::default());
+        errors.extend(new_errors);
+        generics_buf.push(GenericParam::Ty {
```
sezna, line 242: @sezna rename these

compiler/qsc_frontend/src/typeck/infer.rs
```diff
@@ -383,6 +411,20 @@ impl Inferrer {
         self.constraints.push_back(Constraint::Class(class, span));
     }
 
+    /// Returns a unique type variable with specified constraints.
+    fn constrained_ty(
+        &mut self,
+        meta: TySource,
+        with_constraints: impl Fn(Ty) -> Vec<Constraint>,
```
sezna, line 418: consider `.into_boxed_slice()`

compiler/qsc_frontend/src/typeck/infer.rs
```diff
@@ -891,38 +998,80 @@ fn check_eq(ty: Ty, span: Span) -> (Vec<Constraint>, Vec<Error>) {
...
+        }
         _ => (
             Vec::new(),
             vec![Error(ErrorKind::MissingClassExp(base.display(), span))],
         ),
     }
 }
 
+// i'm using the wildcard below to enforce that Ty::Param is always matched in the err branch, as
+// it shouldn't be constrained by HasField as long as we don't support structural typing
```
sezna, line 1073: @sezna

compiler/qsc_lowerer/src/lib.rs
```diff
@@ -878,10 +878,57 @@ impl Lowerer {
             name_span: field.name_span,
         }
     }
-}
 
-fn lower_generics(generics: &[qsc_hir::ty::GenericParam]) -> Vec<qsc_fir::ty::GenericParam> {
-    generics.iter().map(lower_generic_param).collect()
+    fn lower_generics(
```
sezna, line 882: This had to become a method, because a type parameter can now contain class constraint parameters which could be concrete types. Those concrete types need to be lowered. `.lower_ty()` is a method, so we therefore have to make this a method to have access to `self`.

compiler/qsc_ast/src/ast.rs
```diff
@@ -1431,6 +1435,28 @@ impl Default for PathKind {
...
+    #[must_use]
+    pub fn name(&self) -> Option<&Ident> {
+        match self {
+            PathKind::Ok(path) => Some(&path.name),
+            PathKind::Err(Some(incomplete_path)) => incomplete_path.segments.last(),
+            PathKind::Err(None) => None,
+        }
+    }
+}
+
```
minestarks, line 1459: I had reservations about adding these methods back when I first defined `PathKind` because I couldn't think of a good reason you'd want to access the segments or name part of the path, without knowing if it's an `Err` or `Ok` first. Besides, there's always the `Idents` impl if you want to straightforwardly iterate over the contained `Ident`s.

On a more practical note, these are unused anyway, so maybe just delete?

compiler/qsc_ast/src/ast.rs
```diff
@@ -1968,3 +1994,129 @@ impl ImportOrExportItem {
...
+/// A list of class constraints, used when constraining a type parameter.
+#[derive(Default, PartialEq, Eq, Clone, Hash)]
+pub struct ClassConstraints(pub Box<[ClassConstraint]>);
+
+/// An individual class constraint, used when constraining a type parameter.
+/// To understand this concept, think of parameters in a function signature -- the potential arguments that can
+/// be passed to them are constrained by what type is specified. Type-level parameters are no different, and
+/// the type variables that are passed to a type parameter must satisfy the constraints specified in the type parameter.
+#[derive(Default, PartialEq, Eq, Clone, Hash, Debug)]
```
minestarks, line 2057: I've been trying to squash `Default` impls for the AST nodes because I think they also lead to bugs. I haven't been successful in eliminating all of them, but it'd be nice if we can stop adding more. It's too easy to end up with `""` idents in the AST, and then have the resolver spew nonsense errors like "`""` is not a name` etc. I think the compiler code would just be easier to work with if we stuck to invariants such as "no empty string identifiers". 

This one seems to be unused, so I think you can just delete it.

language_service/src/completion/tests.rs
minestarks, line 1: Commenting in this random file, but this is about the language service in general:

Do you want to update the user-friendly function signature display to include the class bounds?

image

Seems like the hover text here would be better to show `function F<'x: Exp[Int] + Num>(x: 'x) : Unit`

samples/language/ClassConstraints.qs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function Main() : Unit {
+    Message($"{AllEqual([1, 1, 1])}");
+    Message($"{AllEqual([1, 2, 3])}");
+
+    // Because we wrote this function generically, we are able to pass in different types, as
+    // long as they can be compared via the class `Eq`.
+    Message($"{AllEqual([true, true, false])}");
+    Message($"{AllEqual(["a", "b"])}");
```
swernli, line 40: nitpick: some of these are hard to parse because of the way string interpolation and syntax highlighting play together, so I think it would be clearer to pull the function invocations up out of the string, bind the result and just print that.

language_service/src/name_locator.rs
minestarks, line 1: Not sure where exactly the bug is, but there's a bug in the inline rename feature that's likely originating from this file somewhere.

```qsharp
function F<'x: Exp[Int] + Num>(x: 'x) : Unit {
}
```

Place the cursor over `'x` and invoke the "rename symbol" command . Give it a new name. Note how the class constraints disappear . This is likely due to the rename span returned to the editor being too large (it should just contain the type variable but it contains the bounds too)

compiler/qsc_ast/src/ast.rs
```diff
@@ -1968,3 +1994,129 @@ impl ImportOrExportItem {
...
+            self.ty,
+            if self.constraints.0.is_empty() {
+                ""
+            } else {
+                ": "
+            },
+            self.constraints
+        )
+    }
+}
```
minestarks, line 2047: This and the `Debug` impl for `ClassConstraint` are the first custom `Debug` impls in this file. The norm is to define `Display` for AST nodes, and use that in tests / AST dumps . We can debate whether that's the "right" way but it's a pretty entrenched pattern. Can you just make these `Display` impls maybe?

compiler/qsc_parse/src/ty.rs
```diff
@@ -72,9 +73,53 @@ pub(super) fn array_or_arrow(s: &mut ParserContext<'_>, mut lhs: Ty, lo: u32) ->
...
+/// `T: Eq + Iterator[Bool] + Class3`
+///     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ bounds
+fn class_constraints(s: &mut ParserContext) -> Result<ClassConstraints> {
+    let mut bounds: Vec<ClassConstraint> = Vec::new();
+    loop {
+        s.expect(WordKinds::PrimitiveClass);
+        let bound_name = ident(s)?;
+        // if there's a less-than sign, or "open angle bracket", try to parse type parameters for
+        // the class
+        // e.g. `Iterator[Bool]`
```
swernli, line 104: code is correct, but comment is out of date, should be about angle brackets:

```suggestion
        // if there's an "open bracket", try to parse type parameters for the class
        // e.g. `Iterator[Bool]`
```

compiler/qsc_ast/src/ast.rs
```diff
@@ -1968,3 +1994,129 @@ impl ImportOrExportItem {
...
+impl std::fmt::Debug for ClassConstraints {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        // A + B + C + D
+        write!(
+            f,
+            "{}",
+            self.0
+                .iter()
+                .map(|x| x.name.to_string())
```
minestarks, line 2117: This misses the constraint parameters in the debug dump.

compiler/qsc_ast/src/ast.rs
```diff
@@ -1968,3 +1994,129 @@ impl ImportOrExportItem {
...
+        }
+    }
+}
+
+impl ConstraintParameter {
+    /// Getter for the `ty` field.
+    #[must_use]
+    pub fn ty(&self) -> &Ty {
+        &self.ty
+    }
```
minestarks, line 2088: Why these trivial getters for public fields? Again I'd argue for consistency across the AST nodes - the existing convention seems to be accessing fields directly.

compiler/qsc_doc_gen/src/display.rs
```diff
@@ -615,12 +615,12 @@ where
...
 }
 
-fn display_type_params(generics: &[GenericParam]) -> String {
+fn display_type_params(generics: &[HirTypeParameter]) -> String {
     let type_params = generics
         .iter()
         .filter_map(|generic| match generic {
-            GenericParam::Ty(name) => Some(name.name.clone()),
-            GenericParam::Functor(_) => None,
+            HirTypeParameter::Ty { name, .. } => Some(name.clone()),
```
swernli, line 622: I think it's worth filing a follow up to make sure docs generation includes type bounds. Right now it's not critical since the stdlib doesn't have anything with bounds, but the generated docs for your new library (other PR) will be wrong in VS Code until this is updated.

compiler/qsc_frontend/src/typeck/tests/bounded_polymorphism.rs
```diff
@@ -0,0 +1,822 @@
...
+            #42 167-172 "a: 'E" : Param<"'E": 0>
+            #46 174-180 "b: Int" : Int
+            #53 187-224 "{\n                a ^ b\n            }" : Param<"'E": 0>
+            #55 205-210 "a ^ b" : Param<"'E": 0>
+            #56 205-206 "a" : Param<"'E": 0>
+            #59 209-210 "b" : Int
+            Error(Type(Error(IncorrectNumberOfConstraintParameters { expected: 1, found: 2, span: Span { lo: 56, hi: 59 } })))
+            Error(Type(Error(IncorrectNumberOfConstraintParameters { expected: 1, found: 2, span: Span { lo: 56, hi: 59 } })))
+            Error(Type(Error(IncorrectNumberOfConstraintParameters { expected: 1, found: 0, span: Span { lo: 162, hi: 165 } })))
+            Error(Type(Error(IncorrectNumberOfConstraintParameters { expected: 1, found: 0, span: Span { lo: 162, hi: 165 } })))
```
swernli, line 109: Wasn't this the duplicate error bug you said you fixed? Or was there more than one and one of them was benign?

compiler/qsc_frontend/src/typeck.rs
```diff
@@ -119,4 +124,60 @@ enum ErrorKind {
     #[diagnostic(help("provide a type annotation"))]
     #[diagnostic(code("Qsc.TypeCk.AmbiguousTy"))]
     AmbiguousTy(#[label] Span),
+    #[error("unsupported parametric class bound")]
+    #[diagnostic(code("Qsc.TypeCk.UnsupportedParametricClassBound"))]
+    UnsupportedParametricClassBound(#[label] Span),
```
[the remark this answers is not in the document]
swernli: Wait, never mind. I got confused with the code between the lowerer and the resolver.

compiler/qsc_ast/src/ast.rs
sezna: That makes sense, I've gone ahead and removed them. I didn't need them ultimately, but I thought they might be useful...but dead code is dead code!

compiler/qsc_ast/src/ast.rs
sezna: Makes sense!

compiler/qsc_ast/src/ast.rs
sezna: That makes sense, I've gone ahead and removed them. I didn't need them ultimately, but I thought they might be useful...but dead code is dead code!

compiler/qsc_doc_gen/src/display.rs
sezna: Went ahead and added a printout of bounds here using `ClassConstraints`' `Display` impl.

compiler/qsc_frontend/src/typeck/tests/bounded_polymorphism.rs
sezna: I think this one is going to be a fast-follow, I'd rather get the feature in for the release and track this one down shortly after. That being said, I'll track this down as soon as I'm done with the other things on this branch...so maybe I'll finish before it goes in. Tl;dr I'd rather this not block the PR, and I'll fix it ASAP, whether that be before or after it goes in.

compiler/qsc_frontend/src/typeck.rs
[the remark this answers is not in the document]
sezna: Using `UnrecognizedClass` works for me! Updated.

compiler/qsc_frontend/src/typeck/tests/bounded_polymorphism.rs
swernli: Sounds good!

swernli (approved):
Signing off on the compiler changes, thought I see there is some good feedback from Mine still pending. Assuming that goes well, I think this is ready to merge.

language_service/src/completion/tests.rs
sezna: image

compiler/qsc_fir/src/ty.rs
```diff
@@ -178,20 +178,77 @@ fn instantiate_arrow_ty<'a>(
...
+    Iterable {
+        item: Ty,
+    },
+
+    /// A class that is not built-in to the compiler.
+    NonNativeClass(Rc<str>),
+    Num,
+    Integral,
+    Show,
+}
```
orpuente-MS, line 227: Can we add docstrings to the other variants besides `NonNativeClass`?

compiler/qsc_fir/src/ty.rs
```diff
@@ -178,20 +178,77 @@ fn instantiate_arrow_ty<'a>(
...
+                .collect::<Vec<_>>()
+                .join(", ");
+            write!(f, "{bounds}")
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub enum ClassConstraint {
+    #[default]
```
orpuente-MS, line 212: Is it correct to say that the default of a `ClassConstraint` is the `Eq` constraint? Where do we use this?

compiler/qsc_fir/src/ty.rs
```diff
@@ -178,20 +178,77 @@ fn instantiate_arrow_ty<'a>(
...
+    Exp {
+        power: Ty,
+    },
+    Iterable {
+        item: Ty,
+    },
+
+    /// A class that is not built-in to the compiler.
+    NonNativeClass(Rc<str>),
+    Num,
```
orpuente-MS, line 224: As we discussed, it is better if we split the `Num` constraint into `Mul`, `Div`, `Sub`, etc. to stay consistent with `Add` and `Exp` being their own contraints.

wasm/src/language_service.rs
```diff
@@ -164,6 +164,7 @@ impl LanguageService {
                         qsls::protocol::CompletionItemKind::Variable => "variable",
                         qsls::protocol::CompletionItemKind::TypeParameter => "typeParameter",
                         qsls::protocol::CompletionItemKind::Field => "field",
+                        qsls::protocol::CompletionItemKind::Class => "class",
```
minestarks, line 167: Oh I almost forgot. You need to handle this enum case in both the vscode and playground and map them to the corresponding enum values in the VS Code and Monaco APIs. (There should be two changes, under `vscode/` and `playground/`). That way the proper icon will show up in the UI.

minestarks (approved):
Looking forward to this going live, thanks!

language_service/src/name_locator.rs
sezna: https://github.com/user-attachments/assets/6eb1cd71-8d45-4a6c-b05a-eaf1b31c0c1e

fixed!

compiler/qsc_fir/src/ty.rs
sezna: This is a bad copy-paste on my part -- we have to have a default in the parser for parser recovery (a pattern I'm not actually very happy with) and I copy/pasted it here.

It hasn't been used anywhere so I removed it. Good catch, thank you.

PR #2009, opened 2024-11-06
Include samples in completions when the document is empty

minestarks:
Fixing the behavior that regressed in #1947 when I was attempting to limit samples to empty documents only.

### DIFF
--- a/vscode/src/completion.ts
+++ b/vscode/src/completion.ts
@@ -87,13 +87,14 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
     });
 
     // Include the samples in contexts that are syntactically appropriate.
-    // The presence of the "namespace" keyword in the completion list is a
-    // hint that the cursor is at the top level.
+    // The presence of the "operation" keyword in the completion list is a
+    // hint that the cursor is at a point we can insert the sample code.
+    
     const shouldIncludeSamples =
       results.findIndex(
         (i) =>
           i.kind === vscode.CompletionItemKind.Keyword &&
-          i.label === "namespace",
+          i.label === "operation",
       ) !== -1;
 
     return !shouldIncludeSamples ? results : results.concat(this.samples);
--- a/vscode/src/completion.ts
+++ b/vscode/src/completion.ts
@@ -86,7 +86,15 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
       return item;
     });
 
+    let isEmptyOrWhitespace = true;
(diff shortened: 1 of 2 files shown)


vscode/src/completion.ts
```diff
@@ -86,7 +86,15 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
...
+    for (let i = 0; i < document.lineCount; i++) {
+      if (!document.lineAt(i).isEmptyOrWhitespace) {
+        isEmptyOrWhitespace = false;
+        break;
+      }
+    }
+
     // Include the samples list for empty documents
-    return document.lineCount > 0 ? results : results.concat(this.samples);
+    return !isEmptyOrWhitespace ? results : results.concat(this.samples);
```
billti, line 98: I'm not sure this is the right approach. What if I start a file with some TODO comments (which I often do) or a license header (which is common)? Or I import some items knowing I'll be using them as I expand on the sample I'm starting from?

I think it's reasonable to only show when the user is on a blank line though.

billti: Actually, in the spirit of the 'correctness', being that the samples are all 'top level' code, could we just show them in situations when at the top level (i.e. the places where we currently show 'namespace').

I'd rather limit the 'magic' of when samples do or don't appear in a completion list that may not be intuitive to users. Have them appear only at top level seems to fit in the with correctness goal of only having  completions show where they are valid.

minestarks: > i.e. the places where we currently show 'namespace' 

Oh, now I think I see what you were getting at. I'll just check for the presence of `namespace` in the list, should be trivial and fairly accurate :)

minestarks: Per offline discussions. Will change "namespace" to "operation" in anticipation of #2023

billti (approved):

vscode/src/completion.ts
```diff
@@ -86,7 +86,17 @@ class QSharpCompletionItemProvider implements vscode.CompletionItemProvider {
       return item;
     });
 
-    // Include the samples list for empty documents
-    return document.lineCount > 0 ? results : results.concat(this.samples);
+    // Include the samples in contexts that are syntactically appropriate.
+    // The presence of the "operation" keyword in the completion list is a
+    // hint that the cursor is at a point we can insert the sample code.
+
+    const shouldIncludeSamples =
```
sezna, line 93: interesting heuristic...i'm going to trust y'all on this one

sezna (approved):

PR #2011, opened 2024-11-06
Fix to RCA panic when mapping a tuple input pattern to a non-tuple expression

cesarzc:
This change fixes an RCA panic that occurs when a tuple input pattern maps to a non-tuple expression.

### DIFF
--- a/compiler/qsc_rca/src/core.rs
+++ b/compiler/qsc_rca/src/core.rs
@@ -2515,6 +2515,7 @@ fn map_input_pattern_to_input_expressions(
     match &pat.kind {
         PatKind::Bind(_) | PatKind::Discard => vec![expr_id.expr],
         PatKind::Tuple(pats) => {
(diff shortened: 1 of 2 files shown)


swernli (approved):

compiler/qsc_rca/src/core.rs
```diff
@@ -2533,8 +2534,10 @@ fn map_input_pattern_to_input_expressions(
                 }
                 input_param_exprs
             } else {
-                assert!(pats.len() == 1);
-                vec![expr_id.expr]
+                // All elements in the pattern map to the same expression.
+                // This is one of the boundaries where we can loose specific information since we are "unpacking" the
```
minestarks, line 2538:
```suggestion
                // This is one of the boundaries where we can lose specific information since we are "unpacking" the
```

minestarks (approved):

DmitryVasilevsky (approved):

PR #2012, opened 2024-11-08
Docs Gen Tool Updates

### LINKED ISSUE #1866: Generated documentation should include 'Std' way of referring items in the standard library
Currently documentation includes 'fully qualified name' which it uses old namespaces, Microsoft.Quantum.Arrays.All for example. We also want the new way (Std.Arrays.All) to be included as well.

### LINKED ISSUE #1998: Docs Generation Does Not Include Unrestricted Items

ScottCarda-MS:
Many updates to the docs gen tool:
- Generates docs for `@Config(Unrestricted)` items.
- Moves Unstable namespaces to the bottom of the Table of Contents.
- Generates docs for reexported Core items.
- Generates index.md files for namespaces.
- Generates a top-level index.md file.
- Trims datetime to just be the date.
- Adds description metadata field.

fixes #1998
fixes #1866


npm/qsharp/generate_docs.js
```diff
@@ -57,10 +57,18 @@ docs.forEach((doc) => {
     default:
       throw new Error(`Invalid file path: ${doc.filename}`);
   }
-  var contents =
-    doc.metadata.replace("ms.date: {TIMESTAMP}", `ms.date: ${today_str}`) +
-    "\n\n" +
-    doc.contents;
+  var contents = "";
+  if (doc.filename === "index.md") {
```
DmitryVasilevsky, line 61: Looks like this file is part of the npm package. I'm wondering if it is better to move this file closer to the library and return the content of it along with other files?

wasm/src/tests.rs
```diff
@@ -504,17 +504,20 @@ fn test_doc_gen() {
     let docs = qsc_doc_gen::generate_docs::generate_docs(None, None, None);
     assert!(docs.len() > 100);
     for (name, metadata, contents) in docs {
-        // filename will be something like "Microsoft.Quantum.Canon/ApplyToEachC.md"
```
DmitryVasilevsky, line 507: You may want to update this:
https://github.com/microsoft/qsharp/blob/6f376c48d82466ea8a50439a30432ba77d1bf5c3/playground/src/docs.tsx#L23

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -22,6 +22,135 @@ use std::sync::Arc;
...
+        buf.join(".")
+    }
+
+    fn display_for_toc(&self) -> String {
+        format!(
+            "---
+uid: {}
+title: {}
+description: {}
+author: bradben
```
DmitryVasilevsky, line 58: I suggest we move this to some pipeline or deployment script.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -189,101 +320,103 @@ pub fn generate_docs(
...
+    // items with a namespace should be also sorted alphabetically with the index file appearing first.
     // Also, items without any metadata (table of content) should come last.
     files.sort_by_key(|file| {
-        (
-            file.1.package.clone(),
-            file.1.namespace.clone(),
-            file.1.name.clone(),
-        )
+        let prefix = if file.0.ends_with("index.md") {
+            "0"
```
sezna, line 351: can this be an int, so the sorting order is more clear? (not sure if stringified numbers always sort in the order we'd expect)

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -189,101 +320,103 @@ pub fn generate_docs(
...
-    // Create toc line
-    let line = format!("  - {{name: {}, uid: {}}}", metadata.name, metadata.uid);
-
     let met: Arc<Metadata> = Arc::from(metadata);
-    files.push((file_name, met, file_content));
+    files.push((file_name, met.clone(), file_content));
 
-    // Return (ns, line)
-    Some((ns.clone(), line))
+    // Return (ns, metadata)
```
sezna, line 415: not sure if this comment is particularly practical

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -316,102 +449,213 @@ Fully qualified name: {fqn}
...
+    let name = true_metadata.name;
 
-#[derive(PartialOrd, Ord, Eq, PartialEq, Clone)]
-enum PackageKind {
-    UserCode,
-    AliasedPackage(String),
-    StandardLibrary,
-    Core,
+    let summary = format!(
+        "This is an exported item. The actual definition is found here: [{name}](xref:Qdk.{true_fqn})"
```
sezna, line 473: looks awesome!

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -316,102 +449,213 @@ Fully qualified name: {fqn}
...
+                        None // We ignore "QIR" namespaces
+                    } else {
+                        Some(name.name())
+                    }
+                }
+                _ => None,
+            }
+        }
+        None => None,
+    }
```
sezna, line 626: I wrote this in the github comment box so it may not be correct, but I think this code can be simplified a bit:

```suggestion
    let local_id = item.parent?;
    let parent = package
        .items
        .get(local_id)
        .expect("Could not resolve parent item id");
    let ItemKind::Namespace(name, _) = &parent.kind else { return None };
    if name.starts_with("QIR") {
        None // We ignore "QIR" namespaces
    } else {
        Some(name.name())
    }

```

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -316,102 +449,213 @@ Fully qualified name: {fqn}
...
+// Returns the package and item of the root definition for an item.
+// If given the root definition, it will return the same item.
+fn resolve_export<'a>(
+    default_package_id: PackageId,
+    package: &'a Package,
+    include_internals: bool,
+    item: &'a Item,
+    display: &'a CodeDisplay,
+) -> Option<(&'a Package, &'a Item)> {
+    // Filter items
```
sezna, line 639: perhaps specify what exactly is being filtered out -- the code kind of says it, but since this comment is already here..might as well make it a full sentence 😸

npm/qsharp/index.md
```diff
@@ -0,0 +1,32 @@
...
+| [`Std.Random`](xref:Qdk.Std.Random-toc)                                                                   | Items for creating random values.                            |
+| [`Std.Range`](xref:Qdk.Std.Range-toc)                                                                     | Items for working with ranges.                               |
+| [`Std.ResourceEstimation`](xref:Qdk.Std.ResourceEstimation-toc)                                           | Items for working with the Azure Quantum Resource Estimator. |
+| [`Microsoft.Quantum.Unstable.Arithmetic`](xref:Qdk.Microsoft.Quantum.Unstable.Arithmetic-toc)             | Items for working with arithmetic comparisons.               |
+| [`Microsoft.Quantum.Unstable.StatePreparation`](xref:Qdk.Microsoft.Quantum.Unstable.StatePreparation-toc) | Items for preparing a quantum state.                         |
+| [`Microsoft.Quantum.Unstable.TableLookup`](xref:Qdk.Microsoft.Quantum.Unstable.TableLookup-toc)           | Items for performing table lookups.                          |
```
swernli, line 32: Here too the key factor is its table lookup using qubits, so worth calling out. 
```suggestion
| [`Microsoft.Quantum.Unstable.TableLookup`](xref:Qdk.Microsoft.Quantum.Unstable.TableLookup-toc)           | Items for performing quantum table lookups.                          |
```

swernli (commented):
Some feedback on the namespace descriptions. Also, unrelated to these changes, I noticed a mistake in a doc comment for DumpOperation, so if we are going to regenerate docs content I'd love to get that in too (edit: opened a separate PR, #2016).

compiler/qsc_doc_gen/src/generate_docs.rs
ScottCarda-MS: expanded the comment to be more descriptive

compiler/qsc_doc_gen/src/generate_docs.rs
ScottCarda-MS: removed

compiler/qsc_doc_gen/src/generate_docs.rs
ScottCarda-MS: All of these have been replaced with `{AUTHOR}` and `{MS_AUTHOR}` tags to be replaced during publishing.

npm/qsharp/generate_docs.js
ScottCarda-MS: I've moved the contents of the file to a rust file in the doc gen tool folder. It doesn't work well to have the markdown file there and have the rust code read from that because it would need a relative path which would be relative to where ever the user executes the code from.

swernli (approved):
Signing off on the namespace descriptions! I gave the other code a high-level review and it looks fine, but others may want to weigh in more on the details of how this interacts with the VS Code documentation feature.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -22,6 +23,136 @@ use std::sync::Arc;
 type Files = Vec<(Arc<str>, Arc<str>, Arc<str>)>;
 type FilesWithMetadata = Vec<(Arc<str>, Arc<Metadata>, Arc<str>)>;
 
+// Namespace -> metadata for items
+type ToC = FxHashMap<Rc<str>, Vec<Arc<Metadata>>>;
```
minestarks, line 27: Why this mix of `Rc` and `Arc` in this file? Do you need `Arc` at all, since this is all single threaded?

ScottCarda-MS: Strange, I definitely needed them at one point, because I remember having to go out of my way to use them, but I can't remember why. Turning them back to Rc's works fine now, so we'll go with that. I've turned them back into Rc's.

compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -316,102 +445,227 @@ Fully qualified name: {fqn}
...
+    let metadata = Metadata {
+        uid: "Microsoft.Quantum.apiref-toc".to_string(),
+        title: "Q# standard libraries for the Microsoft Quantum Development Kit".to_string(),
+        topic: "landing-page".to_string(),
+        kind: MetadataKind::TableOfContents,
+        package: PackageKind::StandardLibrary,
+        namespace: empty_ns.clone(),
+        name: "Overview".into(),
+        summary:
+            "Table of contents for the Q# standard libraries for Microsoft Quantum Development Kit"
```
minestarks, line 562: I think it's called "Azure Quantum Development Kit"

minestarks (approved):

compiler/qsc_doc_gen/src/generate_docs.rs
ScottCarda-MS: updated

compiler/qsc_doc_gen/src/table_of_contents.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// Returns the string of the contents of the generated top-level index.md file.
```
DmitryVasilevsky, line 4: I suggest we add a comment about where this is used. So that when we change the TOC and the test fails, we may assess the impact when updating the test.

DmitryVasilevsky (approved):

playground/src/docs.tsx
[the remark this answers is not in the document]
ScottCarda-MS: Looks like this feature has been broken on the playground in main:
image

At the least with the changes in this PR, the Std docs are available. I might look into showing 2 levels in as you suggest.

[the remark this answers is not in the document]
ScottCarda-MS: Ah found it. I added another capture group to the regex in that logic, and there was code later on that referenced things by capture group index. I incremented that index and can see the full list of namespaces again.

Issue #1934, opened 2024-09-25
Publish more of our written samples to quantum.microsoft.com


Manvi-Agrawal:
- It might be worth including more samples in `samples.mjs`, so that they get featured on  [samples website](https://quantum.microsoft.com/en-us/tools/quantum-coding) for greater discovery. Seems this hasnt gotten updated in a long time. Just has 6 examples(Basic, BV, DJ, Hidden-Shift, Random Number Generator).

- Can we move samples not in samples.mjs list to a new repo? I liked classic QDK samples because they were in a separate repo. @tcNickolas has told me previously that monolith helps to simplify the dependencies. I highly doubt that somebody would likely visit this massive codebase to see these curated samples. Currently, they serve little purpose for both for Q# compiler team and people onboarding to Q#. Maybe add new samples repo as a submodule here if folks think it might be useful, although I know git submodules are a bit annoying.

**Additional context**
Creating separate issue from https://github.com/microsoft/qsharp/issues/1709 for better triaging.

sezna:
Yeah, it would probably be good to add more of our written samples to the list of published ones. We are not inclined to use a new repository, though, we moved to a monorepo with the modern QDK (whereas the classic QDK had many repos). But, we don't expect samples to be discovered via browsing this repo. We link straight to the correct directory from our documentation and quantum.microsoft.com.

Manvi-Agrawal:
> Yeah, it would probably be good to add more of our written samples to the list of published ones.

Sure, that makes sense. Just to confirm, published samples are primarily sourced from the `samples.mjs` list, right?

> We prefer not to introduce a new repository. We transitioned to a monorepo with the modern QDK (unlike the classic QDK with its many repos). However, discovering samples through browsing this repo isn't our main goal. We link directly to the relevant directory from our documentation and `quantum.microsoft.com`.

Thanks, @sezna, for the additional context. As far as I'm aware, `github.com` doesn't currently offer the ability to download a specific folder. Additionally, my attempt to download the `samples` and `samples/algorithms` folders from `github.dev` resulted in an error stating "github.dev can't open folder because it contains system files." While tools might exist for downloading specific folders, I haven't personally used any.

[...]

PR #2013, opened 2024-11-11
Add `Qtest` library that uses class constraints

sezna:
This PR introduces Qtest, a library for running Q# unit tests. It includes two namespaces: `Functions` for testing functions and `Operations` for testing operations. `Operations` also allows for qubit state preparation before tests.

To demonstrate, I converted `MeasureSignedIntTests()` in `signed/src/Tests.qs` to use Qtest. [The result is fewer lines of code](https://github.com/microsoft/qsharp/pull/2013/files#diff-fa6f0fda29e55d52d8254ab813dcda39094b02928370e841ea741c526e3d5338L16) and clearer test intentions. Additionally, all tests run before reporting any failures, unlike the current behavior which stops at the first failure.

I haven't converted all existing Q# tests yet, as that's beyond this PR's scope. This PR aims to validate the API's utility before proceeding with further conversions.

### DIFF
--- a/library/qtest/src/Functions.qs
+++ b/library/qtest/src/Functions.qs
@@ -1,112 +1,60 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import Util.TestCaseResult, Util.OutputMessage;
+import Std.Arrays.Mapped, Std.Arrays.All;
+
 /// # Summary
-/// Runs a series of named test functions with expected results.
+/// Runs a number of test cases and returns true if all tests passed, false otherwise.
+/// Prints a report of what passed and what failed as output.
 ///
-/// # Description
-/// Given an array of test names, test functions, and expected results (of the same type), returns `true`
-/// if all tests passed, and `false` if any one test failed.
-/// Prints the expected and received result for any failed tests.
+/// For a more flexible test running function, see `RunAllTestCases` which returns
+/// test results instead of printing out to output.
 ///
 /// # Input
-/// ## test_cases
-/// An array of three-arity tuples of the form `(test_name, callable_to_test, expected_result)`.
-/// `callable_to_test` will be called and its result will be compared to `expected_result`.
+/// Takes a list of test cases. A test case is a tuple of `(String, () -> T, 'T)`, where
+/// the first String is the name of the test, the function is the test case itself, and the
+/// final element of the tuple is the expected return value from the test case.
 ///
 /// # Example
 /// ```qsharp
-/// function Main() : Unit {
-///     TestCases([
-///         ("Should return 42", TestCaseOne, 42),
-///         ("Should add one", () -> AddOne(5), 6)
-///     ]);
-/// }
-///
-/// function TestCaseOne() : Int {
-///     42
-/// }
-///
-/// function AddOne(x: Int) : Int {
-///     x + 1
-/// }
+/// CheckAllTestCases([("Should return 42", () -> 42, 42)]);
 /// ```
-function TestCases<'Result : Eq + Show > (test_cases : (String, () -> 'Result, 'Result)[]) : Bool {
-    let failed_test_buf = TestCasesSilent(test_cases);
+function CheckAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[]) : Bool {
+    let test_results = RunAllTestCases(test_cases);
 
-    if Length(failed_test_buf) == 0 {
-        Message($"{Length(test_cases)} test(s) passed.");
-        true
-    } else {
-        Message($"{Length(failed_test_buf)} tests failed.");
-        for item in failed_test_buf {
-            Message($"{item}")
-        }
-        false
-    }
+    OutputMessage(test_results);
+    All(test_case -> test_case.did_pass, test_results)
 }
 
 /// # Summary
-/// Similar to `Qtest.Functions.TestCases`, but returns test failure info in an array of strings instead of directly messaging to
-/// output. Useful if you want to handle your own messaging when writing tests, for CI or similar.
-
-/// See `Qtest.Functions.TestCases` for more details.
-///
-/// # Description
-/// Given an array of test names, test functions, and expected results (of the same type), returns an array
-/// of strings representing all failed test cases (if any). Strings are of the form "test_name: expected {}, got {}"
-///
+/// Runs all given test cases and returns a `TestCaseResult` for each test, representing whether or not it passed
+/// and what the failure message, if any.
+/// This is a good alternative to `CheckAllTestCases` when you want custom output based on the results of your tests,
+/// or more control over how test results are rendered.
 /// # Input
-/// ## test_cases
-/// An array of three-arity tuples of the form `(test_name, callable_to_test, expected_result)`.
-/// `callable_to_test` will be called and its result will be compared to `expected_result`.
+/// Takes a list of test cases. A test case is a tuple of `(String, () -> T, 'T)`, where
+/// the first String is the name of the test, the function is the test case itself, and the
+/// final element of the tuple is the expected return value from the test case.
 ///
 /// # Example
 /// ```qsharp
-/// function Main() : Unit {
-///     let failure_messages = TestCasesSilent([
-///         ("Should return 42", TestCaseOne, 42),
-///         ("Should add one", () -> AddOne(5), 6)
-///     ]);
-///     Std.Diagnostics.Fact(Length(failure_messages) == 0, "No tests should fail.")
-/// }
-///
-/// function TestCaseOne() : Int {
-///     42
-/// }
-///
-/// function AddOne(x: Int) : Int {
-///     x + 1
-/// }
+/// RunAllTestCases([("Should return 42", () -> 42, 42)]);
 /// ```
-function TestCasesSilent<'Result : Eq + Show > (test_cases : (String, () -> 'Result, 'Result)[]) : String[] {
+function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[]) : TestCaseResult[] {
     let num_tests = Length(test_cases);
-    mutable failed_test_buf = [];
 
-    for (name, case, result) in test_cases {
-        let (did_pass, message) = TestCase(case, result)!;
-        if not did_pass {
-            set failed_test_buf = failed_test_buf + [$"{name}: {message}"];
-        }
-    }
-
-    failed_test_buf
+    Mapped((name, case, result) -> TestCase(name, case, result), test_cases)
 }
 
-struct TestCaseResult {
-    did_pass : Bool,
-    message : String,
-}
-
-function TestCase<'Result : Eq + Show > (test_case : () -> 'Result, expected : 'Result) : TestCaseResult {
+/// Internal (non-exported) helper function. Runs a test case and produces a `TestCaseResult`
+function TestCase<'T : Eq + Show>(name : String, test_case : () -> 'T, expected : 'T) : TestCaseResult {
     let result = test_case();
     if result == expected {
         new TestCaseResult { did_pass = true, message = "" }
     } else {
-        new TestCaseResult { did_pass = false, message = $"expected: {expected}, got: {result}" }
+        new TestCaseResult { did_pass = false, message = $"{name}: expected: {expected}, got: {result}" }
     }
 }
 
-
-
-export TestCases; 
+export CheckAllTestCases, RunAllTestCases; 
--- a/library/qtest/src/Functions.qs
+++ b/library/qtest/src/Functions.qs
@@ -1,0 +1,112 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// # Summary
+/// Runs a series of named test functions with expected results.
+///
+/// # Description
+/// Given an array of test names, test functions, and expected results (of the same type), returns `true`
+/// if all tests passed, and `false` if any one test failed.
+/// Prints the expected and received result for any failed tests.
+///
+/// # Input
+/// ## test_cases
+/// An array of three-arity tuples of the form `(test_name, callable_to_test, expected_result)`.
+/// `callable_to_test` will be called and its result will be compared to `expected_result`.
+///
+/// # Example
+/// ```qsharp
+/// function Main() : Unit {
+///     TestCases([
+///         ("Should return 42", TestCaseOne, 42),
+///         ("Should add one", () -> AddOne(5), 6)
+///     ]);
+/// }
+///
+/// function TestCaseOne() : Int {
+///     42
+/// }
+///
+/// function AddOne(x: Int) : Int {
+///     x + 1
+/// }
+/// ```
+function TestCases<'Result : Eq + Show > (test_cases : (String, () -> 'Result, 'Result)[]) : Bool {
+    let failed_test_buf = TestCasesSilent(test_cases);
+
+    if Length(failed_test_buf) == 0 {
+        Message($"{Length(test_cases)} test(s) passed.");
+        true
+    } else {
+        Message($"{Length(failed_test_buf)} tests failed.");
+        for item in failed_test_buf {
+            Message($"{item}")
+        }
+        false
+    }
+}
+
+/// # Summary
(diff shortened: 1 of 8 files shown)


library/qtest/src/Tests.qs
```diff
@@ -0,0 +1,26 @@
...
+        ]),
+        "Test harness failed to return true for all passing tests."
+    );
+    Std.Diagnostics.Fact(
+        Length(Functions.TestCasesSilent([
+            ("Should return 42", TestCaseOne, 43),
+            ("Should add one", () -> AddOne(5), 42)
+        ])) == 2,
+        "Test harness failed to return messages for failing tests."
+    );
```
ScottCarda-MS, line 17: Could be good to assert that the returned messages are as expected, in addition to asserting there are the correct number of them. Would also be good to have a test case that mixes in some passing test cases with some failing test cases.

ScottCarda-MS (approved):
Just one suggestion about adding some test cases. Other than that, looks good!

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// function TestCaseOne() : Int {
+///     42
+/// }
+///
+/// function AddOne(x: Int) : Int {
+///     x + 1
+/// }
+/// ```
+operation TestCasesSilent<'Result : Eq + Show > (test_cases : (String, Int, (Qubit[]) => (), (Qubit[]) => 'Result, 'Result)[]) : String[] {
```
Manvi-Agrawal, line 84: How about naming this `TestCasesVerbose` and returning an array of (Bool, String). This would allow people to gather test execution feedback and do some sort of post processing.

Manvi-Agrawal:
@sezna I am curious why are we having a QTest library in the new QDK instead of `@Test` attribute which was there in the classic QDK? Are we assuming that people are used to new QDK and classic QDK doesnt matter anymore?

library/signed/src/Tests.qs
```diff
@@ -8,47 +8,24 @@ import Measurement.MeasureSignedInteger;
...
-    let res = MeasureSignedInteger(a, 5);
-    Fact(res == -2, $"Expected -2, received {res}");
-
-    // 0b11000 == -8
-    X(a[3]);
-    X(a[4]);
-    let res = MeasureSignedInteger(a, 5);
-    Fact(res == -8, $"Expected -8, received {res}");
-
+function MeasureSignedIntTests() : (String, Int, (Qubit[]) => (), (Qubit[]) => Int, Int)[] {
```
Manvi-Agrawal, line 16: Minor comment: I know we have example in the docstrings. Possibly in the follow up PR, we could add some examples in `samples/testing` folder demonstrating `QTest` library in detail for ease of adoption.

library/qtest/src/Functions.qs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// function TestCaseOne() : Int {
+///     42
+/// }
+///
+/// function AddOne(x: Int) : Int {
+///     x + 1
+/// }
+/// ```
+function TestCases<'Result : Eq + Show > (test_cases : (String, () -> 'Result, 'Result)[]) : Bool {
```
swernli, line 34: nitpick: the type variable `'Result` is too close to the built-in type `Result` so could you use something like `'Output` instead?

library/qtest/src/Operations.qs
Manvi-Agrawal: A good follow up might be to provide collective statistics: number of tests passed/failed along with which test case passed/failed. This might be really helpful in CI/CD. Here are some ambitious ideas:
- **VSCode:** How about a VSCode Cmd Pallete that has a GUI feel. For example: Python tests, Rust tests, etc.
- **CI/CD:** Possibly having an artifact for test case results that some one can open in VSCode This might require Rust interfacing since Q# doesnt support file I/O. And Q# extension could then show results in GUI form from test artifact
- **Python:** Having a `qsharp_test` library that does some convenient wraparound to eliminate wrappers . `samples/testing` has examples of Python interop.

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// function AddOne(x: Int) : Int {
+///     x + 1
+/// }
+/// ```
+operation TestCases<'Result : Eq + Show > (test_cases : (String, Int, (Qubit[]) => (), (Qubit[]) => 'Result, 'Result)[]) : Bool {
+    let failed_test_buf = TestCasesSilent(test_cases);
+    
+    if Length(failed_test_buf) == 0 {
+        Message($"{Length(test_cases)} test(s) passed.");
```
Manvi-Agrawal, line 39: We could simplify this by saying `Message($"{Length(test_cases)}-Length(failed_test_buf)}/{Length(test_cases)} test(s) passed."); outside of if condition.

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,116 @@
...
+        true
+    } else {
+        Message($"{Length(failed_test_buf)} tests failed.");
```
Manvi-Agrawal, line 42: nitpick: `Message($"{Length(failed_test_buf)} test(s) failed.");` since `failed_test_buf` can be 1 in this else branch

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,116 @@
...
+/// if all tests passed, and `false` if any one test failed.
+/// Prints the expected and received result for any failed tests.
+///
+/// # Input
+/// ## test_cases
+/// An array of five-arity tuples of the form `(test_name, num_qubits, qubit_prep_callable, callable_to_test, expected_result)`.
+/// `num_qubits` will be allocated in a qubit array and passed to `qubit_prep_callable` to prepare the state before executing 
+/// `callable_to_test`. Afterwards, `callable_to_test` will be called and its result will be compared to `expected_result`.
+///
+/// # Example
```
Manvi-Agrawal, line 18: Nit: Seems this example and one below needs to be changed since TestCases function of operations takes 5 arguments. It seems a copy paste typo from previous one.

library/qtest/src/Functions.qs
```diff
@@ -0,0 +1,112 @@
...
+/// ## test_cases
+/// An array of three-arity tuples of the form `(test_name, callable_to_test, expected_result)`.
+/// `callable_to_test` will be called and its result will be compared to `expected_result`.
+///
+/// # Example
+/// ```qsharp
+/// function Main() : Unit {
+///     let failure_messages = TestCasesSilent([
+///         ("Should return 42", TestCaseOne, 42),
+///         ("Should add one", () -> AddOne(5), 6)
```
Manvi-Agrawal, line 69: Question: If someone wants to test AddOne function for multiple arguments, will they have multiple entries in the TestCases array? I was wondering if it could be simplified by having input array and expected output array for a function? This is motivated by pytest.paramererize that allows listiing inputs and expected outputs

library/qtest/src/Operations.qs
sezna: These are great ideas, @Manvi-Agrawal  -- thank you! We were just looking at integrating with VS Code first -- something using this perhaps https://code.visualstudio.com/api/extension-guides/testing

For CI/CD, we could support a standardized output format from Q# and parse that in Python, then dump out JSON for analysis. Something like that could be automated. What do you think about that?

library/qtest/src/Tests.qs
sezna: Added some more tests here including the ones you outlined :)

library/signed/src/Tests.qs
sezna: Given that this is just one library, I'm hesitant to enshrine it immediately. I wonder if others might want to contribute other test formats. That being said, I'm going to update more libraries that I've authored to use this testing pattern, and that'll hopefully be a good source of example code as well.

library/qtest/src/Operations.qs
sezna: Updated the input to provide some more verbose output.

library/qtest/src/Operations.qs
sezna: Updated this to use a generic message-printing function that just figures out the correct grammar. 😄

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// function TestCaseOne() : Int {
+///     42
+/// }
+///
+/// function AddOne(x: Int) : Int {
+///     x + 1
+/// }
+/// ```
+operation TestCases<'Result : Eq + Show > (test_cases : (String, Int, (Qubit[]) => (), (Qubit[]) => 'Result, 'Result)[]) : Bool {
```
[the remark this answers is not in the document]
sezna: Did a renaming pass to `RunAllTestCases` for the one that returns data, and `CheckAllTestCases` for the one that prints out the message and asserts.

library/qtest/src/Functions.qs
sezna: I think that could be another really good bit of functionality, perhaps as a follow-on addition to this library? I like how that looks, but I don't think it blocks this PR -- what do you think? I'd be happy to take on the task or review your code if you wanted to take a swing at it.

sezna: #2007 is in now, so on the latest build these class constraints work. You'll have to build from main until we do a release, though (which should be soon).

library/qtest/src/Operations.qs
sezna: I'm resolving this so the PR is unblocked, but feel free to continue chatting in this thread, or file an issue and tag me to discuss further.

library/signed/src/Tests.qs
```diff
@@ -8,47 +8,24 @@ import Measurement.MeasureSignedInteger;
 /// This entrypoint runs tests for the signed integer library.
 operation Main() : Unit {
     UnsignedOpTests();
-    MeasureSignedIntTests();
+    Fact(Qtest.Operations.CheckAllTestCases(MeasureSignedIntTests()), "SignedInt tests failed");
     SignedOpTests();
```
swernli, line 12: Why did only this middle case get updated but not the other two around it?

swernli (approved):
The QTest library looks good, just a question about the place it's used in one of the existing library tests. I'm fine if more usage of QTest is waiting for a follow up PR.

library/signed/src/Tests.qs
sezna: per the description, I wanted to convert at least one test case to make sure we are happy with the API and have an example of it being used. However, I think upgrading all tests is out of scope for the PR, and it would be unwise to convert everything before the API on this library is finalized (i.e. the PR is merged)

library/qtest/src/Operations.qs
Manvi-Agrawal: Nit: I cant see updated example here on Github, and I am unable to unresolve conversation, so commenting again to make sure it doesnt get lost

Manvi-Agrawal: PS: I realize that the function signature has been changed, it would be still good to have another toy example here to highlight that operation is being tested. Otherwise, someone might get confused by `Qtest.Functions.CheckAllTestCases` and `Qtest.Operations.CheckAllTestCases`. Also, it seems we need to update `() -> 42` to `() => 42`

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,80 @@
...
+
+/// # Summary
+/// Runs a number of test cases and returns true if all tests passed, false otherwise.
+/// Prints a report of what passed and what failed as output.
+///
+/// For a more flexible test running function, see `RunAllTestCases` which returns
+/// test results instead of printing out to output.
+///
+/// # Input
+/// Takes a list of test cases. A test case is a tuple of `(String, () -> T, 'T)`, where
```
Manvi-Agrawal, line 15: Nit: ``(String, () => T, 'T)``

library/qtest/src/Operations.qs
```diff
@@ -0,0 +1,80 @@
...
+
+}
+
+/// # Summary
+/// Runs all given test cases and returns a `TestCaseResult` for each test, representing whether or not it passed
+/// and what the failure message, if any.
+/// This is a good alternative to `CheckAllTestCases` when you want custom output based on the results of your tests,
+/// or more control over how test results are rendered.
+/// # Input
+/// Takes a list of test cases. A test case is a tuple of `(String, () -> T, 'T)`, where
```
Manvi-Agrawal, line 38: Nit: Similarly here. `(String, () => T, 'T)`

library/qtest/src/Tests.qs
```diff
@@ -0,0 +1,41 @@
...
+function Main() : Unit {
+    let sample_tests = [
+        ("Should return 42", TestCaseOne, 43),
+        ("Should add one", () -> AddOne(5), 42),
+        ("Should add one", () -> AddOne(5), 6)
+    ];
+
+    Fact(
+        not Functions.CheckAllTestCases(sample_tests),
+        "Test harness failed to return false for a failing tests."
```
Manvi-Agrawal, line 15: Nit: for a failing test/for the failing test(s).

library/signed/src/Tests.qs
Manvi-Agrawal: > That being said, I'm going to update more libraries that I've authored to use this testing pattern, and that'll hopefully be a good source of example code as well.

I dont understand this quite well. I feel this is very similar to saying that if wants to see `pytest` usage, please see the code for XYZ python library/libraries which use `pytest` as a testing library.

I think `samples/testing` takes an outside view of Q#, giving users an insight to various ways of testing in Q#. I feel having `QTest` examples there should help to boost the adoption of `QTest` library.  Based on my prior experience with Q#, I sometimes find digging into Q# library, Q# internal rust/python tests, helpful. However, I feel that it shouldnt be an expectation.

However, since Q# is a new language and currently samples reside in the same github repo as the QDK repo which is another story(see #1934 ), probably converting existing library tests to use `QTest` library is not a bad option. I would like to have more thoughts on this. @swernli, @tcNickolas, @billti thoughts?

library/qtest/src/Functions.qs
Manvi-Agrawal: Sure sounds good. Created#2026 for the same.

library/qtest/src/Operations.qs
Manvi-Agrawal: PS2: Seems PR was merged before I commented.

library/signed/src/Tests.qs
```diff
@@ -8,47 +8,24 @@ import Measurement.MeasureSignedInteger;
...
-
-    // 0b11000 == -8
-    X(a[3]);
-    X(a[4]);
-    let res = MeasureSignedInteger(a, 5);
-    Fact(res == -8, $"Expected -8, received {res}");
-
+function MeasureSignedIntTests() : (String, Int, (Qubit[]) => (), (Qubit[]) => Int, Int)[] {
+    [
+        ("0b0001 == 1", 4, (qs) => X(qs[0]), (qs) => MeasureSignedInteger(qs, 4), 1),
```
Manvi-Agrawal, line 18: @sezna , it seems these tests should not pass since the signature of `QTest.Operations.TestCases` was changed. Do the current library tests get checked in CI or am I missing something here.

library/qtest/src/Operations.qs
Manvi-Agrawal: Sounds good @sezna . Created #2029 to track. Feel free to modify issue description.

library/signed/src/Tests.qs
sezna: Hey Manvi, I'm not seeing an issue here. The function call here [does indeed use the latest api](https://github.com/microsoft/qsharp/pull/2013/files/85beeeaba7734d37a48e2f750c0f5e0ea3c84d4c#diff-fa6f0fda29e55d52d8254ab813dcda39094b02928370e841ea741c526e3d5338R11).

Manvi-Agrawal: @sezna, I realized it now, that this still accepts a tuple containing 5 values. I think I got confused by 2 factors:
- **Incorrect [doc string](https://github.com/microsoft/qsharp/blob/main/library/qtest/src/Operations.qs#L14-L22)** : this seems to be copy paste of the functions namespace. Could you please fix it?
- **Unnecessary braces around (Qubit[]) in method signature**: Its written as `(Qubit[])=> ()` here and in `Operations.qs`. This can be written as `Qubit[] => ()`.

PR #2018, opened 2024-11-13
Indentation fix in doc comment formatting

minestarks:
Noticed a minor issue in the formatter where we weren't treating doc comments as comments.


ScottCarda-MS (approved):
Nice catch! Thanks for finding and fixing this.

swernli (approved):

PR #2020, opened 2024-11-13
Track qubit live-ness during simulation

swernli:
This change adds some basic qubit live-ness tracking to the simulator, achieved by changing the `Value::Qubit` variant to a weak pointer to the qubit and having the execution environment own and track the `Rc` that owns the qubit. This provides a few benefits:

- Fixes a panic when a qubit is passed to the simulator after it has been released.
- Adds graceful error handling to treat use-after-release as a runtime error.
- Adds a specific error for "double release" of a qubit if it is being manually managed.
- Updates the display of the qubit type to show the qubit ID when it is live and `Qubit<released>` when it is not, enabling easier debugging of qubit lifetimes.

Of note, local perf testing showing no impact on performance (thanks, Rust zero-cost abstractions!), though it stands to reason this change could theoretically have some impact on programs that use many thousands of qubits.

### DIFF
--- a/compiler/qsc_eval/src/intrinsic.rs
+++ b/compiler/qsc_eval/src/intrinsic.rs
@@ -65,7 +65,7 @@ pub(crate) fn call(
             let qubits_len = qubits.len();
             let qubits = qubits
                 .iter()
-                .filter_map(|q| q.clone().unwrap_qubit().upgrade().map(|q| q.0))
+                .filter_map(|q| q.clone().unwrap_qubit().try_deref().map(|q| q.0))
                 .collect::<Vec<_>>();
             if qubits.len() != qubits_len {
                 return Err(Error::QubitUsedAfterRelease(arg_span));
@@ -86,7 +86,7 @@ pub(crate) fn call(
             let qubits_len = qubits.len();
             let qubits = qubits
                 .iter()
-                .filter_map(|q| q.clone().unwrap_qubit().upgrade().map(|q| q.0))
+                .filter_map(|q| q.clone().unwrap_qubit().try_deref().map(|q| q.0))
                 .collect::<Vec<_>>();
             if qubits.len() != qubits_len {
                 return Err(Error::QubitUsedAfterRelease(arg_span));
@@ -111,7 +111,7 @@ pub(crate) fn call(
         "CheckZero" => Ok(Value::Bool(
             sim.qubit_is_zero(
                 arg.unwrap_qubit()
-                    .upgrade()
+                    .try_deref()
                     .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                     .0,
             ),
@@ -194,7 +194,7 @@ pub(crate) fn call(
         "__quantum__qis__m__body" => Ok(Value::Result(
             sim.m(arg
                 .unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0)
                 .into(),
@@ -202,7 +202,7 @@ pub(crate) fn call(
         "__quantum__qis__mresetz__body" => Ok(Value::Result(
             sim.mresetz(
                 arg.unwrap_qubit()
-                    .upgrade()
+                    .try_deref()
                     .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                     .0,
             )
@@ -213,7 +213,7 @@ pub(crate) fn call(
             let qubits_len = qubits.len();
             let qubits = qubits
                 .iter()
-                .filter_map(|q| q.upgrade().map(|q| q.0))
+                .filter_map(|q| q.try_deref().map(|q| q.0))
                 .collect::<Vec<_>>();
             if qubits.len() != qubits_len {
                 return Err(Error::QubitUsedAfterRelease(arg_span));
@@ -237,7 +237,7 @@ fn one_qubit_gate(
 ) -> Result<Value, Error> {
     gate(
         arg.unwrap_qubit()
-            .upgrade()
+            .try_deref()
             .ok_or(Error::QubitUsedAfterRelease(arg_span))?
             .0,
     );
@@ -255,11 +255,11 @@ fn two_qubit_gate(
     } else {
         gate(
             x.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
             y.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
         );
@@ -280,7 +280,7 @@ fn one_qubit_rotation(
         gate(
             angle,
             y.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
         );
@@ -299,15 +299,15 @@ fn three_qubit_gate(
     } else {
         gate(
             x.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
             y.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
             z.unwrap_qubit()
-                .upgrade()
+                .try_deref()
                 .ok_or(Error::QubitUsedAfterRelease(arg_span))?
                 .0,
         );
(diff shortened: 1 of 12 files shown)


compiler/qsc_eval/src/lib.rs
```diff
@@ -1086,6 +1093,67 @@ impl State {
...
+        arg_span: PackageSpan,
+        out: &mut impl Receiver,
+    ) -> Result<(), Error> {
+        self.push_frame(Vec::new().into(), callee_id, functor);
+        self.increment_call_count(callee_id, functor);
+        let name = &callee.name.name;
+        let val = match name.as_ref() {
+            "__quantum__rt__qubit_allocate" => {
+                let q = Rc::new(Qubit(sim.qubit_allocate()));
+                env.track_qubit(Rc::clone(&q));
```
minestarks, line 1115: I'm seeing this in a couple of places in this file - why not just call the method the "normal" way?

```suggestion
                env.track_qubit(q.clone());
```

swernli: This one is a Rust convention, though one we admittedly don't follow uniformly. Explicitly using the function syntax makes it easy to distinguish between shallow copy with a reference increment and deep copy of the underlying data. This is described in the Rust Book chapter on reference counting (see [here](https://doc.rust-lang.org/book/ch15-04-rc.html#using-rct-to-share-data), about halfway down the page):

> We could have called `a.clone()` rather than `Rc::clone(&a)`, but Rust’s convention is to use `Rc::clone` in this case. The implementation of `Rc::clone` doesn’t make a deep copy of all the data like most types’ implementations of clone do. The call to `Rc::clone` only increments the reference count, which doesn’t take much time. Deep copies of data can take a lot of time. By using `Rc::clone` for reference counting, we can visually distinguish between the deep-copy kinds of clones and the kinds of clones that increase the reference count. When looking for performance problems in the code, we only need to consider the deep-copy clones and can disregard calls to `Rc::clone`.

[There's even a clippy lint for it](https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_ref_ptr), though it's allow by default. I suppose if we really wanted to be consistent about this and make future investigations into deep copies easier to do we would turn that on in our repo settings. Turns out we use `Rc::clone` in 35 places (including the new ones in this PR) but turning on the lint shows we have 169 places where we call `.clone()` on ref-counted types (including three spots I introduced in this PR, whoops). I'll at least switch those three spots to `Weak::clone` for consistency in this PR, but I think it would be up for discussion if we want to be stricter about this pattern across the code base.

minestarks (approved):
> live-ness 

I feel like the word you're looking for is "lifetime"?

compiler/qsc_eval/src/lib.rs
minestarks: Thanks, TIL! This makes sense to me.

compiler/qsc_eval/src/lib.rs
```diff
@@ -426,7 +447,7 @@ impl Env {
     #[must_use]
     pub fn get_variables_in_frame(&self, frame_id: usize) -> Vec<VariableInfo> {
         let candidate_scopes: Vec<_> = self
-            .0
+            .scopes
```
sezna, line 450: i always prefer named fields to tuples, thanks for doing this

compiler/qsc_eval/src/val.rs
```diff
@@ -308,7 +338,7 @@ impl Value {
     /// # Panics
     /// This will panic if the [Value] is not a [`Value::Qubit`].
     #[must_use]
-    pub fn unwrap_qubit(self) -> Qubit {
+    pub fn unwrap_qubit(self) -> Weak<Qubit> {
```
sezna, line 341: there is _some_ code golf opportunity in defining a second function, `unwrap_upgraded_qubit()` where you centralize the expect/upgrade calls. It's not unidiomatic because this function name already has `unwrap` in it, denoting that it could panic. So the call to `expect()` would be fine.

compiler/qsc_partial_eval/src/management.rs
```diff
@@ -9,20 +11,23 @@ use qsc_eval::{
...
     next_callable: CallableId,
     next_block: BlockId,
     next_result_register: usize,
     next_var: usize,
 }
 
 impl ResourceManager {
-    pub fn map_qubit(&self, q: Qubit) -> usize {
+    pub fn map_qubit(&self, q: &Weak<Qubit>) -> usize {
+        let q = q.upgrade().expect("qubit should still be alive");
```
sezna, line 30: what's the difference between "qubit should still be alive", "qubit should be checked for liveness before call into backend", "qubit used after release", and the various other `expect()` messages? Are they trying to convey nuanced, different things, or is it just different ways of saying the same thing?

I ask because this is a very subtle/nuanced change -- these messages refer to some implicit functionality in this part of our code and make an analogy to an advanced Rust concept. 

A docstring that describes how we are using `Weak` and how that correlates to our qubit liveness model would be very helpful in the future I think

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2972,30 +2991,21 @@ fn eval_bin_op_with_integer_literals(
...
-                span,
-            )),
+            (_, 0) => Err(EvalError::DivZero(span).into()),
             (lhs, rhs) => Ok(Value::Int(lhs % rhs)),
         }
     }
 
     fn eval_integer_exp(lhs_int: i64, rhs_int: i64, span: PackageSpan) -> Result<Value, Error> {
         let Ok(rhs_int_as_u32) = u32::try_from(rhs_int) else {
-            return Err(Error::EvaluationFailed(
```
sezna, line 2995: is this just a drive-by update?

compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -216,7 +253,16 @@ fn two_qubit_gate(
     if x == y {
         Err(Error::QubitUniqueness(arg_span))
     } else {
-        gate(x.unwrap_qubit().0, y.unwrap_qubit().0);
+        gate(
+            x.unwrap_qubit()
```
sezna, line 257: You asked about code golf before -- I think some abstraction here would definitely help. Not just for code golf, but for clarity of what is going on. A wrapper struct would be easiest in my opinion. It would also provide a centralized place to dump some docstrings and overall notes on the usage of `Weak` to model qubit liveness here

compiler/qsc_partial_eval/src/lib.rs
swernli: Ugh... the pull request extension ate all my comment responses again...

Yes, this is just some unrelated clean up. We had added the ability to convert directly from normal evaluator errors into partial eval errors late in the process, but then missed a few places that still used hard-coded strings rather then the matching evaluator errors. This cleans up the last of these.

compiler/qsc_eval/src/intrinsic.rs
swernli: I had been thinking about our conversation on this point and added the `QubitRef` struct to wrap the notion of a weak reference to a qubit. Hopefully that, along with the docstrings it brings, helps resolve the concern here. It doesn't make the call sites much shorter, but it does make intent clearer and lets us find places that panic more easily since they are a different function call.

compiler/qsc_partial_eval/src/management.rs
swernli: I hope the `QubitRef` struct introduced helps with this. Now there is one spot that panics, and the callers of that function can be found with easy search (or from stack traces in case this happens in practice).

compiler/qsc_eval/src/val.rs
swernli: I pulled this up into the `QubitRef` struct rather than put it directly onto the `Value` impl.

compiler/qsc_eval/src/intrinsic.rs
sezna: Great, I am more concerned about clarity than concision here, because I know that if I was assigned a bug in these files I'd need a lot of help to know what is going on

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1311,6 +1311,23 @@ impl<'a> PartialEvaluator<'a> {
         args_span: PackageSpan,        // For diagnostic purposes only.
         callee_expr_span: PackageSpan, // For diagnostic puprposes only.
     ) -> Result<Value, Error> {
+        // check qubits here!
```
sezna, line 1314:
```suggestion
        // check qubits for liveness here!
```

would this be correct?

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1311,6 +1311,23 @@ impl<'a> PartialEvaluator<'a> {
         args_span: PackageSpan,        // For diagnostic purposes only.
         callee_expr_span: PackageSpan, // For diagnostic puprposes only.
```
sezna, line 1312: not your fault but typo in comment here if we are taking drive by suggestions

sezna (approved):
Big fan of `QubitRef`

compiler/qsc_partial_eval/src/lib.rs
swernli: huh, this comment got left behind from notes to myself, but now I might as well clean it up and leave it in place for posterity.

compiler/qsc_partial_eval/src/lib.rs
swernli: Fixed!

PR #2022, opened 2024-11-13
Moved formerly unstable to Std library

DmitryVasilevsky:
We had Arithmetic, Arithmetic Utils (internal), Table Lookup and State Preparation in the unstable library. Now they are moved to the Std library.

Taking Arithmetic as an example, we now place Q# code in the Std.Arithmetic and re-export it in Microsoft.Quantum.Unstable.Arithmetic for compatibility. Separate _unstable_ library is removed.

We probably should look at using AND gate in subsequent PR.


ScottCarda-MS:
When do you see us being ready to drop the unstable reexports of these libraries?

ScottCarda-MS (approved):

sezna (approved):
It doesn't have to happen in this PR, but it would be good to update the libraries that rely on `Unstable` to use `Std` before release ([signed](https://github.com/microsoft/qsharp/blob/main/library/signed/qsharp.json#L7) and [fixed point](https://github.com/microsoft/qsharp/blob/main/library/fixed_point/qsharp.json#L13))

PR #2023, opened 2024-11-13
Standard Q# Sample Entry Points and Implicit Namespaces

ScottCarda-MS:
- Changes the samples with non-standard entry point names to use `Main()` as their entry point, removing the need for the `@EntryPoint()` attribute.
- Removes explicit namespaces in the samples.


swernli (approved):

samples/python_interop/teleportation_project/src/TeleportOperations/Teleport.qs
```diff
@@ -1,16 +1,13 @@
-namespace TeleportLib {
+import TeleportOperations.PrepareState.PrepareState.*;     // references the PrepareState.qs file
```
minestarks, line 1: This is preexisting so don't worry about it if you don't want to touch it in this PR - but I feel like the `PrepareState` subfolder can be removed from this project, it only has 1 file in it and adds an unnecessary level to the namespace hierarchy.

vscode/src/memfs.ts
```diff
@@ -41,8 +41,9 @@ function populateSamples(vfs: MemFS) {
   vfs.createDirectory(playgroundRootUri.with({ path: "/samples" }));
 
   samples.forEach((sample) => {
+    const sanitized_file_name = sample.title.replace(/\W/g, "");
```
minestarks, line 44: I can't tell if special characters like `'` or `(` or handled elsewhere or missed (some of the sample titles have them)

samples/python_interop/teleportation_project/src/TeleportOperations/Teleport.qs
ScottCarda-MS: Since it is a sample about project structure, it makes sense to me that the sample demos sub-folders in projects.

vscode/src/memfs.ts
ScottCarda-MS: The regex handles those characters. \W is anything that is not alphanumeric + underscore.

ScottCarda-MS: `\W` is equivalent to `[^a-zA-Z0-9_]`

minestarks (approved):
Great change, and no small feat with all these ad hoc sample notebooks. Thanks!

vscode/src/memfs.ts
minestarks: Oh, I always think "w" stands for "whitespace" 🤦  . Maybe a code comment to help out regex noobs like me.

samples/python_interop/teleportation_project/src/TeleportOperations/Teleport.qs
minestarks: Oh, I missed that. I don't think it's particularly valuable to demo subfolders since the "point" of the sample is Python interop, but you're right the README does mention it so might as well keep it consistent 🤷‍♀️

vscode/src/memfs.ts
ScottCarda-MS: added

samples/notebooks/azure_submission.ipynb
```diff
@@ -1,240 +1,168 @@
...
-      },
-      "outputs": [
-        {
-          "ename": "QSharpCellError",
-          "evalue": "\u001b[31mQsc.CapabilitiesCk.UseOfDynamicBool\u001b[0m\n\n  \u001b[31m×\u001b[0m cannot use a dynamic bool value\n   ╭─[\u001b[36;1;4mline_0\u001b[0m:4:1]\n \u001b[2m4\u001b[0m │     let result = M(q);\n \u001b[2m5\u001b[0m │     if (result == Zero) {\n   · \u001b[35;1m        ──────────────\u001b[0m\n \u001b[2m6\u001b[0m │         Message(\"The result is Zero\");\n   ╰────\n\u001b[36m  help: \u001b[0musing a bool value that depends on a measurement result is not\n        supported by the current target\n",
-          "output_type": "error",
-          "traceback": [
-            "\u001b[31mQsc.CapabilitiesCk.UseOfDynamicBool\u001b[0m",
-            "",
-            "  \u001b[31m×\u001b[0m cannot use a dynamic bool value",
```
billti, line 83: Looks like some of these checked-in outputs were removed, but per the input cell is supposed to show the error.

billti: @minestarks mentioned this was intentional, so ignore this comment. Thanks!

billti (approved):

sezna (approved):
much better samples

PR #2037, opened 2024-11-18
Add `TestMatrix` functionality to qtest

### LINKED ISSUE #2026: [QTest] Support for test case data in `CheckAllTestCases` and `RunAllTestCases`
Currently `CheckAllTestCases` and `RunAllTestCases` function take a tuple of the form: `(name : String, test_case : () -> 'T, expected : 'T)`. So, Iif someone wants to test a function for multiple inputs/outputs it leads to code duplication.

```qsharp
CheckAllTestCases([
("Should add 1 to 5", () -> AddOne(5), 6),
("Should add 1 to 5", () -> AddOne(6), 7),
]);

function AddOne(x: Int) : Int {
     return x+1;
}
```

Allow `CheckAllTestCases` and `RunAllTestCases` function by having input array and expected output array for a function. This is motivated by [pytest.mark.parameterize](https://docs.pytest.org/en/stable/how-to/parametrize.html) that allows listing inputs and expected outputs

```qsharp
CheckAllTestCases([
("Should add 1 to 5", (x) -> AddOne(x), [5,6], [6, 7]),
]);

function AddOne(x: Int) : Int {
     return x+1;
}
```

Have duplicate entries in testcases array in `CheckAllTestCases` and `RunAllTestCases` function.

**Additional context**
See https://github.com/microsoft/qsharp/pull/2013#discussion_r1840886028

sezna:
I believe this should address the ideas @Manvi-Agrawal proposed in #2026. What do you think Manvi?

Closes #2026

### DIFF
--- a/library/qtest/src/Functions.qs
+++ b/library/qtest/src/Functions.qs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import Util.TestCaseResult, Util.OutputMessage;
-import Std.Arrays.Mapped, Std.Arrays.All;
+import Std.Arrays.Mapped, Std.Arrays.All, Std.Arrays.Enumerated;
 
 /// # Summary
 /// Runs a number of test cases and returns true if all tests passed, false otherwise.
@@ -47,6 +47,26 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[]) 
     Mapped((name, case, result) -> TestCase(name, case, result), test_cases)
 }
 
+/// # Summary
+/// Given a function to test and an array of test cases of the form (input, expected_output), and a test mode, runs the test cases and returns the result of the test mode.
+///
+/// # Inputs
+/// - `test_suite_name` : A string representing the name of the test suite.
+/// - `func` : The function to test.
+/// - `test_cases` : An array of tuples of the form (input, expected_output).
+/// - `mode` : A function that takes an array of tuples of the form (test_name, test_case, expected_output) and returns a value of type 'U.
+///            Intended to be either `Qtest.Functions.CheckAllTestCases` or `Qtest.Functions.RunAllTestCases`.
+///
+/// # Example
+/// ```qsharp
+/// TestMatrix("Add One", x -> x + 1 [(2, 3), (3, 4)], CheckAllTestCases);
+/// ```
+
+function TestMatrix<'T, 'O : Show + Eq, 'U>(test_suite_name : String, func : 'T -> 'O, test_cases : ('T, 'O)[], mode : ((String, () -> 'O, 'O)[]) -> 'U) : 'U {
+    let test_cases_qs = Mapped((ix, (input, expected)) -> (test_suite_name + $" {ix + 1}", () -> func(input), expected), Enumerated(test_cases));
+    mode(test_cases_qs)
+}
+
 /// Internal (non-exported) helper function. Runs a test case and produces a `TestCaseResult`
 function TestCase<'T : Eq + Show>(name : String, test_case : () -> 'T, expected : 'T) : TestCaseResult {
     let result = test_case();
@@ -57,4 +77,4 @@ function TestCase<'T : Eq + Show>(name : String, test_case : () -> 'T, expected 
     }
 }
 
-export CheckAllTestCases, RunAllTestCases; 
+export CheckAllTestCases, RunAllTestCases, TestMatrix; 
--- a/library/qtest/src/Functions.qs
+++ b/library/qtest/src/Functions.qs
@@ -59,7 +59,7 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[]) 
 ///
 /// # Example
 /// ```qsharp
-/// TestMatrix("Add One", x -> x + 1 [(2, 3), (3, 4)], CheckAllTestCases);
+/// TestMatrix("Add One", x -> x + 1, [(2, 3), (3, 4)], CheckAllTestCases);
 /// ```
 
 function TestMatrix<'T, 'O : Show + Eq, 'U>(test_suite_name : String, func : 'T -> 'O, test_cases : ('T, 'O)[], mode : ((String, () -> 'O, 'O)[]) -> 'U) : 'U {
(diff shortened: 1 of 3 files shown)


library/qtest/src/Functions.qs
```diff
@@ -47,6 +47,26 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[])
...
+/// # Inputs
+/// - `test_suite_name` : A string representing the name of the test suite.
+/// - `func` : The function to test.
+/// - `test_cases` : An array of tuples of the form (input, expected_output).
+/// - `mode` : A function that takes an array of tuples of the form (test_name, test_case, expected_output) and returns a value of type 'U.
+///            Intended to be either `Qtest.Functions.CheckAllTestCases` or `Qtest.Functions.RunAllTestCases`.
+///
+/// # Example
+/// ```qsharp
+/// TestMatrix("Add One", x -> x + 1 [(2, 3), (3, 4)], CheckAllTestCases);
```
Manvi-Agrawal, line 62:
```suggestion
/// TestMatrix("Add One", x -> x + 1, [(2, 3), (3, 4)], CheckAllTestCases);
```

library/qtest/src/Functions.qs
```diff
@@ -47,6 +47,26 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[])
...
+/// ```
+
+function TestMatrix<'T, 'O : Show + Eq, 'U>(test_suite_name : String, func : 'T -> 'O, test_cases : ('T, 'O)[], mode : ((String, () -> 'O, 'O)[]) -> 'U) : 'U {
```
Manvi-Agrawal, line 65: Thanks @sezna for taking this up. Nice addition of `mode` function for further flexibility :-)

**Minor nitpick:** RunTestMatrix, thoughts?

library/qtest/src/Tests.qs
```diff
@@ -2,11 +2,45 @@
...
+    Fact(Length(results) == 2, "test matrix did not return results for all test cases");
+    Fact(All(result -> result.did_pass, results), "test matrix did not pass all test cases");
+}
+
+function BasicTests() : Unit {
     let sample_tests = [
         ("Should return 42", TestCaseOne, 43),
-        ("Should add one", () -> AddOne(5), 42),
+        ("Should add one", () -> AddOne(5), 6),
         ("Should add one", () -> AddOne(5), 6)
```
Manvi-Agrawal, line 44: Is it intentional to have duplicate tests?

library/qtest/src/Functions.qs
Manvi-Agrawal: I like the mode function. We can leave it as it is if you think renaming the function causes confusion. I suggested RunTestMatrix since it gives the feeling that we are doing something. TestMatrix gives the data vibe. I thought mode=CheckAllTestCases/RunAllTestCases in RunTestMatrix is good enough.

PS: sent from mobile.not responsible for autocorrect typos.

sezna: I kind of like just having `TestMatrix`, due to the preexisting Check/Run dichotomy. Do you feel strongly about this?

library/qtest/src/Functions.qs
```diff
@@ -47,6 +47,26 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[])
...
+/// - `test_cases` : An array of tuples of the form (input, expected_output).
+/// - `mode` : A function that takes an array of tuples of the form (test_name, test_case, expected_output) and returns a value of type 'U.
+///            Intended to be either `Qtest.Functions.CheckAllTestCases` or `Qtest.Functions.RunAllTestCases`.
+///
+/// # Example
+/// ```qsharp
+/// TestMatrix("Add One", x -> x + 1, [(2, 3), (3, 4)], CheckAllTestCases);
+/// ```
+
+function TestMatrix<'T, 'O : Show + Eq, 'U>(test_suite_name : String, func : 'T -> 'O, test_cases : ('T, 'O)[], mode : ((String, () -> 'O, 'O)[]) -> 'U) : 'U {
```
minestarks, line 65: This is a bit hard to follow, can we format the argument list vertically?

```qsharp
function TestMatrix<'T, 'O : Show + Eq, 'U>(
    test_suite_name : String,
    func : 'T -> 'O,
    test_cases : ('T, 'O)[],
    mode : ((String, () -> 'O, 'O)[]) -> 'U
) : 'U {
```

(Just type a newline after the argument list opening brace `(` and run the formatter, it should do the right thing)

minestarks, line 65: I'm trying to understand the reasoning behind making `mode` such a flexible type. Is the user ever expected provide their own implementation here? If it can only ever be one of two well-known functions, maybe we can make it an "enum" ....... and by "enum" I mean hardcoded strings 😬  Okay maybe that's not much better either.

So I guess I'm asking, is `mode` typed this way because it's the best we can do with the current Q# (lacking enums, constants etc) or is there another reason?

minestarks: What if we split this into `RunTestMatrix` and `CheckTestMatrix`?

sezna: It is typed this way because it is easy to pass in "CheckAll.." vs "RunAll.." and is also pluggable (users can provide their own implementation if desired). What if I maintained this flexibility but also exported some wrappers that just pass the `mode` in for the user, and called those wrappers `RunTestMatrix` and `CheckTestMatrix`? like this: [bf16c1e](https://github.com/microsoft/qsharp/pull/2037/commits/bf16c1efc6157436008bc3770f673b13c1664bb8)

minestarks: Yes I like that compromise. The simple use case should be simple, and I think the specific exports you added are good for that.

swernli (approved):

library/qtest/src/Functions.qs
```diff
@@ -47,6 +47,47 @@ function RunAllTestCases<'T : Eq + Show>(test_cases : (String, () -> 'T, 'T)[])
     Mapped((name, case, result) -> TestCase(name, case, result), test_cases)
 }
 
+/// # Summary
+/// Given a function to test and an array of test cases of the form (input, expected_output), and a test mode, runs the test cases and returns the result of the test mode.
```
orpuente-MS, line 51:
```suggestion
/// Given a function to test, an array of test cases of the form (input, expected_output), and a test mode, runs the test cases and returns the result of the test mode.
```

orpuente-MS (approved):

PR #2040, opened 2024-11-19
Library for rotation operations

msoeken:
This is a Q# library for rotation operations. So far it includes an implementation of the Hamming weight phasing operation. I intend to include other rotation operations at a later point.

### DIFF
--- a/library/rotations/src/HammingWeightPhasing.qs
+++ b/library/rotations/src/HammingWeightPhasing.qs
@@ -1,0 +1,76 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import Std.Arrays.Enumerated, Std.Arrays.Most, Std.Arrays.Partitioned, Std.Arrays.Tail;
+import Std.Convert.IntAsDouble;
+import Std.Diagnostics.Fact;
+import Std.Math.Floor, Std.Math.Lg;
+
+operation HammingWeightPhasing(angle: Double, qs : Qubit[]) : Unit {
+    WithHammingWeight(qs, (sum) => {
+        for (i, sumQubit) in Enumerated(sum) {
+            Rz(IntAsDouble(2^i) * angle, sumQubit);
+        }
+    });
+}
+
+internal operation WithHammingWeight(qs : Qubit[], action : Qubit[] => Unit) : Unit {
+    let n = Length(qs);
+
+    if n <= 1 {
+        action(qs);
+    } elif n == 2 {
+        use sum = Qubit();
+
+        within {
+            AND(qs[0], qs[1], sum);
+            CNOT(qs[0], qs[1]);
+        } apply {
+            action([qs[1], sum]);
+        }
+    } elif n == 3 {
+        WithSum(qs[0], qs[1..1], qs[2..2], action);
+    } else {
+        let power = Floor(Lg(IntAsDouble(n - 1)));
+        let split = Partitioned([n - 2^power, 2^power - 1], qs);
+        Fact(Length(split) == 3 and Length(split[2]) == 1, $"Unexpected split for n = {n}");
+
+        WithHammingWeight(split[0], (leftHW) => {
+            WithHammingWeight(split[1], (rightHW) => {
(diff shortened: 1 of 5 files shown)


library/rotations/src/Tests.qs
```diff
@@ -0,0 +1,50 @@
+import HammingWeightPhasing.HammingWeightPhasing;
+import Std.Diagnostics.CheckOperationsAreEqual;
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ScottCarda-MS, line 3: Looks like the imports got a strange ordering here in relations to the copyright comment. I know this can happen when the imports are added automatically with auto-complete.

library/rotations/src/HammingWeightPhasing.qs
```diff
@@ -0,0 +1,99 @@
...
+///   Mathias Soeken, Earl Campbell, Martin Roetteler, Giovanni De Micheli.
+operation HammingWeightPhasing(angle : Double, qs : Qubit[]) : Unit {
+    WithHammingWeight(qs, (sum) => {
+        for (i, sumQubit) in Enumerated(sum) {
+            Rz(IntAsDouble(2^i) * angle, sumQubit);
+        }
+    });
+}
+
+internal operation WithHammingWeight(qs : Qubit[], action : Qubit[] => Unit) : Unit {
```
swernli, line 36: nitpick: the `internal` keyword does not actually do anything anymore, as only the items that are explicitly exported are publicly accessible. That said, it does convey some intent here, so I think it's fine to leave it for stylistic reasons if that's what you prefer.

swernli (approved):

library/rotations/src/HammingWeightPhasing.qs
msoeken: Yes, exactly. In a way to highlight in one file which of the operations are exported and which one are used for the implementation.

library/rotations/src/Tests.qs
```diff
@@ -0,0 +1,48 @@
...
+        WithHammingWeight(qs, sum => {
+            Fact(MeasureInteger(sum) == HammingWeightI(number), $"wrong Hamming weight computed for number = {number}");
+        });
+        ResetAll(qs);
+    }
+}
+
+operation TestPhasing() : Unit {
+    for numQubits in 1..6 {
+        Fact(CheckOperationsAreEqual(numQubits, qs => HammingWeightPhasing(2.0, qs), qs => ApplyToEachA(Rz(2.0, _), qs)), "Operations are not equal");
```
DmitryVasilevsky, line 46: Maybe also loop over several angles from different quadrants.

library/rotations/src/HammingWeightPhasing.qs
```diff
@@ -0,0 +1,99 @@
...
+import Std.Convert.IntAsDouble;
+import Std.Diagnostics.Fact;
+import Std.Math.Floor, Std.Math.Lg, Std.Math.MaxI, Std.Math.MinI;
+
+/// # Summary
+/// Applies a Z-rotation (`Rz`) with given angle to each qubit in qs.
+///
+/// # Description
+/// This implementation is based on Hamming-weight phasing to reduce the number
+/// of rotation gates.  The technique was first presented in [1], and further
```
DmitryVasilevsky, line 14: Suggest mentioning that this reduction comes at the cost of Clifford gates. Which would probably make it less efficient for simulation.

DmitryVasilevsky (approved):

library/rotations/src/HammingWeightPhasing.qs
```diff
@@ -0,0 +1,99 @@
...
+            CNOT(qs[0], qs[1]);
+        } apply {
+            action([qs[1], sum]);
+        }
+    } elif n == 3 {
+        WithSum(qs[0], qs[1..1], qs[2..2], action);
+    } else {
+        let power = Floor(Lg(IntAsDouble(n - 1)));
+        let (leftLen, rightLen) = (n - 2^power, 2^power - 1);
+        // handle corner case if n is power of 2
```
DmitryVasilevsky, line 55: Probably elaborate. We don't want first partition to be longer than the second?

PR #2042, opened 2024-11-20
Make it easier to access dumps and matrices from Python

swernli:
This adds some convenience accessors for getting dumps, messages, and matrices as separate entries from an invocation of `qsharp.run` and `qsharp.eval`. Previously, the only way to get a state dump from a run was the awkward:

```python
state = qsharp.StateDump(qsharp.run("DumpMachine()", shots=1, save_events=True)[0]["events"][0].state_dump())
```

This change preserves the existings "events" entry in the saved output, which has everything intermingled in the order from each shot, but also introduces dumps, messages, and matrices that will keep just the ordered output of that type. This makes the above pattern slightly better (and more discoverable):

```python
state = qsharp.run("DumpMachine()", shots=1, save_events=True)[0]["dumps"][0]
```

This adds similar functionality to `qsharp.eval` which now supports `save_events=True` to capture output, so for single shot execution you can use:

```python
state = qsharp.eval("DumpMachine()", save_events=True)["dumps"][0]
```


pip/qsharp/_qsharp.py
```diff
@@ -219,18 +219,129 @@ def get_interpreter() -> Interpreter:
...
+        as the check ignores global phase.
+
+        :param state: The state to check against, provided either as a dictionary of state indices to complex amplitudes,
+            or as a list of real amplitudes.
+        :param tolerance: The tolerance for the check. Defaults to 1e-10.
+        """
+        phase = None
+        # Convert a dense list of real amplitudes to a dictionary of state indices to complex amplitudes
+        if isinstance(state, list):
+            state = {i: state[i] for i in range(len(state))}
```
orpuente-MS, line 272: Nitpick. Just as in rust, in Python is more idiomatic (and faster) to use the enumerate iterator
`{ i : val for i, val in enumerate(state) }`

because you are skipping the bound checks each time you do state[i]

orpuente-MS (approved):

pip/qsharp/_qsharp.py
swernli: Ooo, good call. Today I learned :)

pip/qsharp/_qsharp.py
```diff
@@ -219,18 +219,129 @@ def get_interpreter() -> Interpreter:
...
+class ShotResult(TypedDict):
+    """
+    A single result of a shot.
+    """
+
+    events: List[Output]
+    result: Any
+    messages: List[str]
+    matrices: List[Output]
+    dumps: List[StateDump]
```
minestarks, line 307: A dict with hardcoded attributes is nothing but a class though :) We should declare this type as a class. A `@dataclass` might be even better (Python's concept for a POD struct).

Using a dict here feels like JavaScript leaked into Python a little bit. 

(If you have to keep it this way for backcompat, ignore)

pip/qsharp/_qsharp.py
```diff
@@ -219,18 +219,129 @@ def get_interpreter() -> Interpreter:
...
     """
     ipython_helper()
 
+    results: ShotResult = {
+        "events": [],
+        "result": None,
+        "messages": [],
+        "matrices": [],
+        "dumps": [],
+    }
```
minestarks, line 333: I like the overall solution.

1. I don't love how we have 3 arrays with no ordering between them, and then we have another array with an order but no types. Can't I (as the user) have it all? I want my event list to be ordered AND typed. 

2. I also feel like we're fighting having to use actual Python features here 😆  . The following just feel more "native" and take advantage of Python's own dynamic typing rather than the dynamically-exposing-Rust-enums we're doing here.


```python
output for output in result.events if isinstance(output, Matrix)
```

```python
for output in result.events:
  if isinstance(output, Matrix):
    print(str(output))
```

Now granted that's a lot more code to write. OK, so let's forgo using actual Python classes. How about something like the below? We do already have `output.state_dump()`, so adding `matrix()` and `message()` feel like a logical next step.


```python
output for output in result.events if output .matrix() is not None
```

```python
for output in result.events:
  if output.matrix() is not None:
    print(str(output))
```

And if that's too wordy, maybe throw in a method for filtering on the fly

```python
result.events.matrices()   # A convenience method that filters on the events list, not a separately stored list
```

pip/qsharp/_qsharp.py
swernli: I think you make a great point, but it would be a breaking change for anyone using the result since they'd have to change from `output['result']` to `output.result`. Given that, I'll just extend the typed dictionary instead.

pip/qsharp/_qsharp.py
swernli: Hmm... if `ShotResult` were a class I could definitely make this all work more nicely. I guess the reason I was gravitating toward keeping them in separate lists is that if I have code like this:
```qsharp
{
    Message("Foo");
    DumpMachine();
    Message("Bar");
    Message("Baz");
    DumpMachine();
}
```
I'm far more likely to think to myself, "I want the second dump," than, "I want the fifth output." Yes, I could do it with filtering and sublcassing and all that, but that requires that I know the right subclasses and that I write the filter, as you mention. So I'd want to put this all into a member function, but I can't do that on `TypedDict`. At least with this approach, if someone has their `results` and they try to use `results[0][...]` with autocomplete on the "..." they'll get to see the options of whats in there without having to know the different ways they could filter the "events" entry that has anything.

Frustratingly, the real solution here is a class... which is a breaking change :(

minestarks: I think compatibility with `qsharp_widgets` limits our options quite a bit here, so I'm happy with whatever you can come up with given the constraints.

minestarks (approved):

pip/qsharp/_qsharp.py
swernli: I ended up landing at a middle ground of keeping the type specific lists while also making the overall "events" list have type converted entries.

PR #2046, opened 2024-12-02
Avoid qiskit-1.3

swernli:
Looks like something changed in Qiskit v1.3 that breaks our interop. Pinning to 1.2 while we investigate.


minestarks:
@swernli think you could file an issue so we don't get distracted and forget?

PR #2050, opened 2024-12-03
Fix Qiskit interop with v1.3

### LINKED ISSUE #2047: Qiksit interop breaks with v1.3
There was a breaking change in Qiskit v1.3 that caused interop tests to fail. We pinned to 1.2 in #2046 and should figure out why it's failing and handle both before and after 1.3 somehow.

See [this pipeline](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=81605&view=results) for the failure info, or run the integration tests for Python locally with the latest Qiskit installed.

swernli:
Qiskit v1.3 changed the behavior of `num_qubits` on the `Target` class in way that broke our `QirTarget` usage, since we expect to default to `None` but the new logic ignores the `num_qubits` parameter to `__init__` and always defaults to zero qubits, failing later calls to transpilation. By shadowing the parent class property in `QirTarget` we can continue returning `None` and trigger the right behavior in transpilation while preserving a setter for use with v1.2.

Note: I tested locally with both v1.2.4 and v1.3.0 integration tests to confirm the behavior is as expected.

Fixes #2047

### DIFF
--- a/build.py
+++ b/build.py
@@ -514,7 +514,7 @@ if build_pip and build_widgets and args.integration_tests:
         "ipykernel",
         "nbconvert",
         "pandas",
-        "qiskit>=1.2.2,<2.0.0",
+        "qiskit>=1.3.0,<2.0.0",
     ]
     subprocess.run(pip_install_args, check=True, text=True, cwd=root_dir, env=pip_env)
 
--- a/build.py
+++ b/build.py
@@ -69,10 +69,10 @@ parser.add_argument(
 )
 
 parser.add_argument(
-        "--ci-bench",
-        action=argparse.BooleanOptionalAction,
-        default=False,
-        help="Run the benchmarking script that is run in CI (default is --no-ci-bench)",
+    "--ci-bench",
+    action=argparse.BooleanOptionalAction,
+    default=False,
+    help="Run the benchmarking script that is run in CI (default is --no-ci-bench)",
 )
 
 args = parser.parse_args()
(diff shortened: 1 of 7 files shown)


pip/tests-integration/test_requirements.txt
```diff
@@ -1,5 +1,5 @@
 pytest==8.2.2
-qiskit>=1.2.2,<1.3.0
+qiskit>=1.3.0,<2.0.0
```
minestarks, line 2: Now that we have some compat code in there, I wonder if there's an easy way to set up these tests to be run against 1.2 as well. But maybe that's a bigger change for another PR....

pip/qsharp/interop/qiskit/backends/backend_base.py
```diff
@@ -319,19 +320,26 @@ def _transpile(self, circuit: QuantumCircuit, **options) -> QuantumCircuit:
 
         circuit = self.run_qiskit_passes(circuit, options)
 
-        orig = self.target.num_qubits
-        try:
```
minestarks, line 323: Are you confident we don't need the `try` block anymore?

minestarks (approved):

pip/tests-integration/test_requirements.txt
swernli: Yeah, running tests against multiple downlevel versions of Qiskit could be interesting, but I think I'd prefer to treat that as best effort and only have CI care about latest. The recent versions seem to be changing frequently enough that I don't know we'd want to commit to keeping all the compat stuff going for that long.

pip/qsharp/interop/qiskit/backends/backend_base.py
swernli: In this case, yes, because all it was doing was making sure the `finally` set the `num_qubits` on the target back to the original value. Now we are instead respecting the target configured value throughout which means a user's custom target either needs to explicitly return `None` (as we now do) or pick a specific value that will cause transpilation to reject a circuit that is too big (which seems to be the newer, desired behavior for Qiskit targets).

build.py
```diff
@@ -514,7 +514,7 @@ def run_ci_historic_benchmark():
         "ipykernel",
         "nbconvert",
         "pandas",
-        "qiskit>=1.2.2,<1.3.0",
+        "qiskit>=1.2.2,<2.0.0",
```
orpuente-MS, line 517: The requirements.txt file says `qiskit>=1.3.0` but this says `qiskit>=1.2.2`. Is this correct?

orpuente-MS (approved):

build.py
swernli: Good catch! I had originally left this because I thought this determines what we try and download when the Python module is installed, but this is just the build script. This should be updated to 1.3.

swernli:
```suggestion
        "qiskit>=1.3.0,<2.0.0",
```

PR #2051, opened 2024-12-04
Support `ApplyUnitary` in simulation

swernli:
This adds a new library function `ApplyUnitary` that given a matrix of complex numbers will apply that matrix to the given qubits. This requires a new version of the simulator that exposes the functionality publicly.

Starting as draft to wait for https://github.com/qir-alliance/qir-runner/pull/208

### DIFF
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -429,6 +429,11 @@ impl Backend for SparseSim {
                     .collect::<Vec<_>>();
                 let matrix = unwrap_matrix_as_array2(matrix, &qubits);
 
+                let adj = matrix.t().map(Complex::<f64>::conj);
+                if matrix.dot(&adj) != Array2::eye(1 << qubits.len()) {
+                    return Some(Err("matrix is not unitary".to_string()));
+                }
+
                 self.sim.apply(&matrix, &qubits, None);
 
                 Some(Ok(Value::unit()))
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use crate::noise::PauliNoise;
 use crate::val::Value;
+use crate::{noise::PauliNoise, val::unwrap_tuple};
+use ndarray::Array2;
 use num_bigint::BigUint;
 use num_complex::Complex;
 use quantum_sparse_sim::QuantumSim;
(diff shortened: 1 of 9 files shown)


library/std/src/Std/Intrinsic.qs
```diff
@@ -1140,6 +1140,58 @@ operation Z(qubit : Qubit) : Unit is Adj + Ctl {
     adjoint self;
 }
 
+/// # Summary
+/// Applies the given unitary matrix to the given qubits. The matrix must be square, with dimensions dictated by the number of qubits.
```
DmitryVasilevsky, line 1144: Unitary matrix is always square. I suggest leaving only the first sentence in the summary and consolidating second sentence and remarks in description. Then removing remarks. As stated in readme, remarks are for "General knowledge that may be useful for understanding of the functionality provided."

compiler/qsc_eval/src/backend.rs
```diff
@@ -419,6 +420,36 @@ impl Backend for SparseSim {
                 self.apply_noise(q);
                 Some(Ok(Value::unit()))
             }
+            "Apply" => {
```
DmitryVasilevsky, line 423: Did you check if Apply needs to be handled in other backends? I had to add handling for other intrinsics in a few places.

compiler/qsc_eval/src/backend.rs
```diff
@@ -419,6 +420,36 @@ impl Backend for SparseSim {
...
+                let [matrix, qubits] = unwrap_tuple(arg);
+                let qubits = qubits.unwrap_array();
+                let qubits = qubits
```
DmitryVasilevsky, line 426: I know other legs of this match are also rather lengthy, but maybe there's a way to move matrix transformation into a separate function.

library/std/src/Std/Intrinsic.qs
```diff
@@ -1140,6 +1140,58 @@ operation Z(qubit : Qubit) : Unit is Adj + Ctl {
...
+    let num_qubits = Length(qubits);
+    if num_rows != 1 <<< num_qubits {
+        fail "matrix passed to ApplyUnitary must have dimensions 2^Length(qubits).";
+    }
+
+    Apply(matrix, qubits);
+}
+
+@Config(Unrestricted)
+operation Apply(matrix : Complex[][], qubits : Qubit[]) : Unit {
```
DmitryVasilevsky, line 1191: I like "Apply" as a name, but on the other hand "Apply" is a very generic name. I'm afraid it could be easily confused or conflict with something else... Could we make it more specific? If not, it's fine.

DmitryVasilevsky (approved):

library/std/src/Std/Intrinsic.qs
swernli: I guess I was assuming not everyone knows a priori that unitary requires square so was trying to explain that, but given this is already a fairly advanced utility it's probably safe to let the terminology do the heavy lifting. I'll make the changes you suggest.

compiler/qsc_eval/src/backend.rs
swernli: I did. In this case, circuits already do the right thing in that they treat any custom intrinsic as a box with the params displayed:
image
For the other backends/scenarios they all will fail with an error about an unrecognized intrinsic which is actually the desired behavior. Neither QIR generation nor resource estimation know how to handle an arbitrary unitary, so should fail if invoked on code that includes this call.

library/std/src/Std/Intrinsic.qs
swernli: This one is not exposed to the user, only `ApplyUnitary` is exported. So this is effectively an internal utility. At first I had it called `ApplyUnitaryInternal` but then noticed that circuits will display this name, since it's the intrinsic, and not the wrapper, and it looked weird to have `ApplyUnitaryInternal` in the circuit diagram. So I settled at something shorter that looks better for circuits. The other bit of reasoning that I had was that every intrinsic needs to globally unique in its name so it seemed better to pick something non-specific as that felt less likely to overlap with a custom intrinsic someone might try to use later.

compiler/qsc_eval/src/backend.rs
swernli: Good idea, I'll see what I can refactor here!

orpuente-MS (approved):

compiler/qsc_eval/src/backend.rs
```diff
@@ -419,6 +420,19 @@ impl Backend for SparseSim {
...
                 Some(Ok(Value::unit()))
             }
+            "Apply" => {
+                let [matrix, qubits] = unwrap_tuple(arg);
+                let qubits = qubits
+                    .unwrap_array()
+                    .iter()
+                    .filter_map(|q| q.clone().unwrap_qubit().try_deref().map(|q| q.0))
+                    .collect::<Vec<_>>();
+                let matrix = unwrap_matrix_as_array2(matrix, &qubits);
```
swernli, line 430: I need to add another check here to make sure the matrix is unitary in it's numerical contents. If someone passes in a matrix that causes the state vector to be empty, like the all zero matrix, then it will panic the simulator.

swernli: Added a check here that uses the `ndarray` functionality to verify the constraints.

PR #2052, opened 2024-12-05
Add Hover Descriptions to Attributes

### LINKED ISSUE #1994: Add documentation to attributes in Q#
We have added a couple more attributes since the last release and we plan to add a few more going forward. It would be helpful to the users to have some documentation show up when you hover over an attribute.

The documentation strings can be hard coded into the language service, since there is no way to define custom attributes in Q#.

A screenshot showing the lack of documentation when you hover over an attribute.
Image

ScottCarda-MS:
Adds hover descriptions to built-in attributes in the Q# language server.

Closes #1994

### DIFF
--- a/compiler/qsc_hir/src/hir.rs
+++ b/compiler/qsc_hir/src/hir.rs
@@ -1371,11 +1371,11 @@ impl Attr {
 Valid arguments are `Base`, `Adaptive`, `IntegerComputations`, `FloatingPointComputations`, `BackwardsBranching`, `HigherLevelConstructs`, `QubitReset`, and `Unrestricted`.
 
 The `not` operator is also supported to negate the attribute, e.g. `not Adaptive`.",
-            Attr::EntryPoint => "Indicates that a callable is an entry point to a program.",
+            Attr::EntryPoint => "Indicates that the callable is the entry point to a program.",
             Attr::Unimplemented => "Indicates that an item is not yet implemented.",
-            Attr::SimulatableIntrinsic => "Indicates that an item should be treated as an intrinsic callable for QIR code generation and any implementation should be ignored.",
-            Attr::Measurement => "Indicates that a callable is a measurement. This means that the operation will be marked as \"irreversible\" in the generated QIR, and output Result types will be moved to the arguments.",
-            Attr::Reset => "Indicates that a callable is a reset. This means that the operation will be marked as \"irreversible\" in the generated QIR.",
+            Attr::SimulatableIntrinsic => "Indicates that an item should be treated as an intrinsic callable for QIR code generation and any implementation should only be used during simulation.",
+            Attr::Measurement => "Indicates that an intrinsic callable is a measurement. This means that the operation will be marked as \"irreversible\" in the generated QIR, and output Result types will be moved to the arguments.",
+            Attr::Reset => "Indicates that an intrinsic callable is a reset. This means that the operation will be marked as \"irreversible\" in the generated QIR.",
         }
     }
 }
--- a/compiler/qsc_hir/src/hir.rs
+++ b/compiler/qsc_hir/src/hir.rs
@@ -1344,7 +1344,7 @@ impl Display for Ident {
 /// An attribute.
 #[derive(Clone, Debug, PartialEq)]
 pub enum Attr {
-    /// Provide pre-processing information about when an item should be included in compilation.
+    /// Provides pre-processing information about when an item should be included in compilation.
     Config,
     /// Indicates that a callable is an entry point to a program.
     EntryPoint,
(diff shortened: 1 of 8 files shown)


language_service/src/hover.rs
```diff
@@ -51,6 +51,27 @@ struct HoverGenerator<'a> {
 }
 
 impl<'a> Handler<'a> for HoverGenerator<'a> {
+    fn at_attr_ref(&mut self, name: &'a ast::Ident) {
+        let description = match name.name.as_ref() {
+                "Config" => "Provides pre-processing information about when an item should be included in compilation.
```
DmitryVasilevsky, line 56: There's Attr enum in qsharp\compiler\qsc_hir\src\hir.rs with these choices and almost the same text in comments. I would expect this information to be next to that enum. Does language_service share anything with the compiler? If so, consider moving it close to that enum.

ScottCarda-MS: Good thinking. That way if something needs to change with the attributes, all the information is in one place (the hir).

compiler/qsc_hir/src/hir.rs
```diff
@@ -1361,6 +1361,25 @@ pub enum Attr {
...
+    /// Gets the string description of the attribute.
+    #[must_use]
+    pub fn description(&self) -> &'static str {
+        match self {
+            Attr::Config => "Provides pre-processing information about when an item should be included in compilation.
+
+Valid arguments are `Base`, `Adaptive`, `IntegerComputations`, `FloatingPointComputations`, `BackwardsBranching`, `HigherLevelConstructs`, `QubitReset`, and `Unrestricted`.
+
+The `not` operator is also supported to negate the attribute, e.g. `not Adaptive`.",
+            Attr::EntryPoint => "Indicates that a callable is an entry point to a program.",
```
swernli, line 1374: nitpick: since we only allow a single entry point, this should be "the"
```suggestion
            Attr::EntryPoint => "Indicates that the callable is the entry point to a program.",
```

compiler/qsc_hir/src/hir.rs
```diff
@@ -1361,6 +1361,25 @@ pub enum Attr {
...
+            Attr::Unimplemented => "Indicates that an item is not yet implemented.",
+            Attr::SimulatableIntrinsic => "Indicates that an item should be treated as an intrinsic callable for QIR code generation and any implementation should be ignored.",
```
swernli, line 1376:
```suggestion
            Attr::SimulatableIntrinsic => "Indicates that an item should be treated as an intrinsic callable for QIR code generation and any implementation should only be used during simulation.",
```

compiler/qsc_hir/src/hir.rs
```diff
@@ -1361,6 +1361,25 @@ pub enum Attr {
...
+            Attr::Measurement => "Indicates that a callable is a measurement. This means that the operation will be marked as \"irreversible\" in the generated QIR, and output Result types will be moved to the arguments.",
```
swernli, line 1377:
```suggestion
            Attr::Measurement => "Indicates that an intrinsic callable is a measurement. This means that the operation will be marked as \"irreversible\" in the generated QIR, and output Result types will be moved to the arguments.",
```

compiler/qsc_hir/src/hir.rs
```diff
@@ -1361,6 +1361,25 @@ pub enum Attr {
...
+            Attr::Reset => "Indicates that a callable is a reset. This means that the operation will be marked as \"irreversible\" in the generated QIR.",
```
swernli, line 1378:
```suggestion
            Attr::Reset => "Indicates that an intrinsic callable is a reset. This means that the operation will be marked as \"irreversible\" in the generated QIR.",
```

swernli (approved):

language_service/src/name_locator.rs
```diff
@@ -5,19 +5,24 @@ use std::mem::replace;
...
 use crate::compilation::Compilation;
-use qsc::ast::visit::{walk_expr, walk_namespace, walk_pat, walk_ty, walk_ty_def, Visitor};
+use qsc::ast::visit::{
+    walk_attr, walk_expr, walk_namespace, walk_pat, walk_ty, walk_ty_def, Visitor,
+};
 use qsc::ast::{FieldAccess, Idents, PathKind};
 use qsc::display::Lookup;
 use qsc::{ast, hir, resolve};
 
 #[allow(unused_variables)]
```
orpuente-MS, line 15: Could we remove the `#[allow(unused_variables)]` attribute here? There doesn't seem to be any unused variables.

orpuente-MS (approved):

language_service/src/name_locator.rs
ScottCarda-MS: done

PR #2054, opened 2024-12-06
Expose user-defined callables into Python

swernli:
This change adds new interop capabilities for the qsharp Python module. Any user-defined global callables with types supported by interop will automatically be surfaced into the `qsharp.code` module and available for invocation or import directly in Python.


compiler/qsc_frontend/src/error.rs
```diff
@@ -70,15 +72,16 @@ impl<E: Diagnostic + Send + Sync> WithSource<E> {
     /// Takes a span that uses `SourceMap` offsets, and returns
     /// a span that is relative to the `Source` that the span falls into,
     /// along with a reference to the `Source`.
-    pub fn resolve_span(&self, span: &SourceSpan) -> (&Source, SourceSpan) {
+    pub fn resolve_span(&self, span: &SourceSpan) -> Result<(&Source, SourceSpan), MietteError> {
```
minestarks, line 75: I'm worried about this. Is there really a legitimate case where the span can get out of sync with the sources contained in the error?

swernli: it's not that the span gets out sync, but rather that there is no source code for the line the call originates from. But with some more testing, it appears that this is specific to invoking stdlib intrinsic callables directly from Python. Since these changes don't support that, I can take this and the related code out and it's one less oddity to worry about in this PR. That said, if we later need to support Python calls into intrinsics from the stdlib, we might need to revisit this. Longer explanation below...

When an intrinsic callable fails internally it can't point to those internals so it instead uses the span of the call expression in the error. But if the line that actually invoked that callable was in Python, then there is no source code for that call expression and thus no span. I ran into this when I tried manually calling `__quantum__rt__qubit_release` on a qubit in the non-zero state, because it uses the span from the arguments of the call expression:
https://github.com/microsoft/qsharp/blob/f02677479532caa2c252ab58b6d4e15a7247fc0a/compiler/qsc_eval/src/lib.rs#L1130
Since the original call was from a line of Python like, `qsharp.env.QIR.Runtime.__quantum__rt__qubit_release(q)`, there is no Q# source for that argument span. This helper would panic while trying to create the error to report.

Again, all that is moot now, so I'll remove this part of the change. But we should keep this in the back of our minds if we ever open up the ability to directly invoke standard library intrinsics.

minestarks: That's good to know thanks for explaining! When the time comes we can discuss how to handle the case. In the meantime I'll stop worrying :)

compiler/qsc_frontend/src/error.rs
```diff
@@ -72,6 +72,7 @@ impl<E: Diagnostic + Send + Sync> WithSource<E> {
     /// along with a reference to the `Source`.
     pub fn resolve_span(&self, span: &SourceSpan) -> (&Source, SourceSpan) {
         let offset = u32::try_from(span.offset()).expect("expected the offset to fit into u32");
+        // Since some sources may be missing, treat any failure to find the source as "out of bounds" error.
```
minestarks:
```suggestion
```

This got left behind!

minestarks: oops you got it already.

compiler/qsc_frontend/src/error.rs
swernli: Good call out, though! Code is removed now.

pip/tests/test_qsharp.py
```diff
@@ -373,3 +375,77 @@ def test_target_profile_from_str_match_enum_values() -> None:
     assert qsharp.TargetProfile.from_str(str_value) == target_profile
     with pytest.raises(ValueError):
         qsharp.TargetProfile.from_str("Invalid")
+
+
+def test_callables_exposed_into_env() -> None:
```
minestarks, line 380: Could you break this out into a few different test cases?

pip/tests/test_qsharp.py
```diff
@@ -373,3 +375,77 @@ def test_target_profile_from_str_match_enum_values() -> None:
...
+    qsharp.eval("function Unsupported(q : Qubit[]) : Unit { }")
+    with pytest.raises(AttributeError):
+        qsharp.env.Unsupported
+    qsharp.eval('function Unsupported() : Qubit { fail "won\'t be called" }')
+    with pytest.raises(AttributeError):
+        qsharp.env.Unsupported
+    qsharp.eval("function Unsupported(a : Std.Math.Complex) : Unit { }")
+    with pytest.raises(AttributeError):
+        qsharp.env.Unsupported
+    qsharp.eval('function Unsupported() : Std.Math.Complex { fail "won\'t be called" }')
```
minestarks, line 438: Part of me wants to try an project UDTs as Python classes, but... yeah, maybe it's best to banish the thought 😜

swernli: Ha, yeah. Sometimes we have to walk before we can run.

pip/tests/test_qsharp.py
```diff
@@ -373,3 +375,77 @@ def test_target_profile_from_str_match_enum_values() -> None:
...
+    # imported callables should fail gracefully
+    with pytest.raises(qsharp.QSharpError):
+        Four()
+
+
+def test_callables_with_unsupported_types_not_exposed_into_env() -> None:
+    qsharp.init()
+    qsharp.eval("function Unsupported(q : Qubit) : Unit { }")
+    with pytest.raises(AttributeError):
+        qsharp.env.Unsupported
```
minestarks, line 428: Where's the `AttributeError` coming from? Is that what happens when the name is just undefined?

Just as a cute idea, could we generate "stub" Python functions for these that raise a helpful error, like "this Q# function can't be used from Python because it uses types that cannot be passed between Q# and Python". I feel like that would eliminate some "why isn't this working?!?!" frustrations. I'll leave it up to you to decide whether that's overkill.

minestarks: > "why isn't this working?!?!" frustrations

On second thought, I'd be more concerned that a user would try (unwittingly) with a Q# function that uses disallowed types, and then see that it doesn't work and just give up on the feature. A specific error message would encourage them to keep trying.

pip/src/interpreter.rs
```diff
@@ -494,6 +574,89 @@ impl Interpreter {
...
+            ))),
+            Prim::Pauli => Ok(Value::Pauli(match obj.extract::<Pauli>(py)? {
+                Pauli::I => fir::Pauli::I,
+                Pauli::X => fir::Pauli::X,
+                Pauli::Y => fir::Pauli::Y,
+                Pauli::Z => fir::Pauli::Z,
+            })),
+            Prim::Qubit | Prim::Range | Prim::RangeTo | Prim::RangeFrom | Prim::RangeFull => Err(
+                PyException::new_err(format!("unhandled primitive input type: {prim_ty:?}")),
+            ),
```
minestarks, line 626: Can we just avoid projecting Q# callables that take unmarshallable types as input? (Correct me if we're already doing that - I see that we're doing it for return types, but I don't see it being done for input types).

And once we do that, this should be a panic.

pip/tests/test_qsharp.py
swernli: Will do!

pip/src/interpreter.rs
swernli: The change already includes checking of both input and output (see the `item_supports_interop` function, which checks both types). These were nice errors for when the other code would try converting anything it found, but you are right in that these paths should never get it now. So when it's just for future us in developing newer features a panic is good enough, and panics are actually reported really nicely in Python via pyO3 so that makes investigation easy. I'll switch these!

pip/tests/test_qsharp.py
swernli: Hmm... that's an interesting idea... Part of me feels weird about populating the Python environment with objects that will never work, but you make a good point about giving the user some feedback. And in a sense, these are objects the user asked for, so better for them to see them consistently even if they can't be called. I'll try a few versions of this idea and see what works.

swernli: I think I found a good way to surface the behavior you suggested, and it even tells you the first unsupported type in the signature that it found:
image
Now I need to update the tests...

pip/qsharp/_qsharp.py
```diff
@@ -252,6 +275,33 @@ def callback(output: Output) -> None:
...
+
+        def callback(output: Output) -> None:
+            if _in_jupyter:
+                try:
+                    display(output)
+                    return
+                except:
+                    # If IPython is not available, fall back to printing the output
+                    pass
+            print(output, flush=True)
```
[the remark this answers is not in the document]
swernli: It's analogous to lines 257-265; the `return` after display prevents this `print` from being executed if `display` runs as expected.

pip/qsharp/_qsharp.py
```diff
@@ -190,6 +192,25 @@ def init(
...
+    # Loop through the environment module and remove any dynamically added attributes that represent
+    # Q# callables. This is necessary to avoid conflicts with the new interpreter instance.
+    keys_to_remove = []
+    for key in env.__dict__:
+        if hasattr(env.__dict__[key], "__qs_gen") or isinstance(
+            env.__dict__[key], types.ModuleType
+        ):
+            keys_to_remove.append(key)
+    for key in keys_to_remove:
+        env.__delattr__(key)
```
idavis, line 204: Is there something that we're trying go avoid clearing? Naively it would be simpler to just clear everything.

pip/qsharp/_qsharp.py
```diff
@@ -190,6 +192,25 @@ def init(
...
+
+    # Also remove any namespace modules dynamically added to the system.
+    keys_to_remove = []
+    for key in sys.modules:
+        if key.startswith("qsharp.env."):
+            keys_to_remove.append(key)
+    for key in keys_to_remove:
+        sys.modules.__delitem__(key)
```
idavis, line 212: Where are these added to the module?

pip/tests/test_qsharp.py
```diff
@@ -373,3 +375,99 @@ def test_target_profile_from_str_match_enum_values() -> None:
     assert qsharp.TargetProfile.from_str(str_value) == target_profile
     with pytest.raises(ValueError):
         qsharp.TargetProfile.from_str("Invalid")
+
+
+def test_callables_exposed_into_env() -> None:
+    qsharp.init()
+    qsharp.eval("function Four() : Int { 4 }")
+    assert qsharp.env.Four() == 4
```
idavis, line 383: Is this syntax used to avoid confusion in determining if the call is Q# or Python native? My first thought was to declare the callable in current scope, but I can see that it might cause confusion.

pip/qsharp/_qsharp.py
swernli: It's over in the native code: https://github.com/microsoft/qsharp/blob/swernli/python-interop/pip/src/interpreter.rs#L1002-L1057

pip/tests/test_qsharp.py
swernli: putting it in a well-known module would help avoid conflicts/shadowing of things in the Python environment. @minestarks and I are still debating on the module name itself and looking for ideas... `qsharp.env` was the first one but we are wondering if something like `qsharp.code` or `qsharp.my` or `qsharp.current` or something. Any thoughts on that?

pip/qsharp/_qsharp.py
swernli: Yeah, a module registered from the file system comes with a bunch of Python-generated properties that I don't want to remove, for example:
image

pip/src/interpreter.rs
```diff
@@ -828,3 +991,67 @@ where
...
+    name: &str,
+    item_id: StoreItemId,
+) -> PyResult<()> {
+    if namespace.is_empty() && name == "<lambda>" {
+        // We don't want to bind auto-generated lambda callables.
+        return Ok(());
+    }
+
+    // env is expected to be a module, any other type will raise an error.
+    let mut module: Bound<PyModule> = env.extract(py)?;
```
idavis, line 1016: Why are we not passing the env as a pymodule from the beginning? Feels odd that it is a pyobject in the pyi and calling methods.

pip/qsharp/_qsharp.py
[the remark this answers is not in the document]
idavis: I missed the return.

pip/src/interpreter.rs
swernli: Doing it that way failed compilation on the Rust side, but it turns out @minestarks has a better approach that she is going to push up. With that the module no longer needs to be passed at all, and a bunch of the module manipulation code in Rust goes away.

pip/qsharp/env/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Environment module that receives any user-defined Q# callables as Python functions.
```
minestarks, line 4: Would a module level docstring do any good here?

pip/qsharp/__init__.py
```diff
@@ -56,5 +56,6 @@
     "PauliNoise",
     "DepolarizingNoise",
     "BitFlipNoise",
-    "PhaseFlipNoise"
+    "PhaseFlipNoise",
+    "GlobalCallable",
```
minestarks, line 60: This essentially declares the "public" API for our package. Exposing `GlobalCallable` feels unnecessary though? You should be able to access it just fine for internal use (via `qsharp._native`) without exposing it through here.

compiler/qsc_hir/src/global.rs
```diff
@@ -157,6 +157,17 @@ impl PackageIter<'_> {
                     intrinsic: decl.body.body == SpecBody::Gen(SpecGen::Intrinsic),
                 }),
             }),
+            (ItemKind::Callable(decl), None) => Some(Global {
```
minestarks, line 160: If I understand this correctly, these include items declared in local scopes. These aren't really "globals" though, are they? Bit concerned about what assumptions they're breaking elsewhere. If nothing else, I think it's misleading to call them globals.

compiler/qsc/src/interpret.rs
```diff
@@ -315,6 +318,70 @@ impl Interpreter {
         })
     }
 
+    /// Given a package ID, returns all the global items in the package.
+    /// Note this does not currently include re-exports.
+    pub fn get_global_items(
```
minestarks, line 323: Per rust naming conventions I'd suggest removing the `get_` prefix from these methods

minestarks (commented):
Posting another batch of comments, but I need to finish going through the `compiler` folder. Python side looks good

compiler/qsc_hir/src/global.rs
swernli: It actually wont. It includes callables whose direct parents are namespaces or callables that have no direct parent (defined in the global scope). Rather cleverly, our compiler generates callables lifted from within other callables (including lambdas) as having the parent be the callable item in which they were defined, so those get filtered out by this logic:
image

pip/qsharp/__init__.py
swernli: That's a good point. In some version of the changes, I think I had wanted this to be public so that users could do `isinstance(variable, qsharp.GlobalCallable)` but since that gets wrapped up a Python generated callable it should never come up. This is definitely an internal type now and doesn't need to be exposed.

compiler/qsc/src/interpret.rs
swernli: Huh... as it happens, we use `get_` all over the place in our codebase... 15 in this file alone and 355 across the whole of the Rust code. I'm torn between updating these to follow the convention and leaving them as-is to match our existing patterns.

pip/qsharp/env/__init__.py
swernli: Added as part of changing the name from `env` to `code`

pip/src/interpreter.rs
```diff
@@ -225,7 +232,7 @@ thread_local! { static PACKAGE_CACHE: Rc<RefCell<PackageCache>> = Rc::default();
 impl Interpreter {
     #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
swernli: We could, sure, but I'd rather leave that for a different PR. Any structs we add here would need to be supported on the constructor invoked from Python, so would need to be exposed via pyO3 as native classes. That seems like just enough of a change that I'd want to separate it out from this work.

pip/qsharp/_qsharp.py
```diff
@@ -252,6 +274,56 @@ def callback(output: Output) -> None:
     return results
 
 
+# Helper function that knows how to create a function that invokes a callable. This will be
+# used by the underlying native code to create functions for callables on the fly that know
+# how to get the currently intitialized global interpreter instance.
+def _make_callable(callable, namespace, callable_name):
```
idavis, line 280: missing types

pip/src/interpreter.rs
```diff
@@ -494,6 +576,83 @@ impl Interpreter {
     }
 }
 
+/// Finds any Q# type recursively that does not support interop with Python, meaning our code cannot convert it back and forth
+/// across the interop boundary.
+fn first_unsupport_interop_ty(ty: &Ty) -> Option<&Ty> {
```
idavis, line 581: is this a typo? `first_unsupport_interop_ty`

swernli: Whoops, yes, will fix.

compiler/qsc_hir/src/global.rs
minestarks: > callables that have no direct parent (defined in the global scope)

I guess this is breaking my brain. Callables can't be defined in the global scope, can they?

Oh, based on the screenshot I think you're referring to the "persistent local scope"? I guess, I mean... is it okay to call them globals? I can convince myself it's harmless, but it _is_ introducing a concept we didn't use to have (global items at the top-level). I still think it's muddying the concepts a bit in the compiler internals.

pip/src/interpreter.rs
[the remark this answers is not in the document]
minestarks: Agreed. I actually took a stab at this and it turned out to be nontrivial so nvm

compiler/qsc/src/interpret.rs
minestarks: Huh. I err on the side of following the pattern especially if it's in the same file. But ideally we'd just update the existing violations too :)

compiler/qsc/src/interpret.rs
```diff
@@ -315,6 +318,70 @@ impl Interpreter {
...
+    /// Get the global items defined in the user source passed into initialization of the interpreter.
+    pub fn get_source_package_global_items(
+        &self,
+    ) -> Vec<(Vec<Rc<str>>, Rc<str>, fir::StoreItemId)> {
+        self.get_global_items(self.source_package)
+    }
+
+    /// Get the global items defined in the open package being interpreted, which will include any items
+    /// defined by calls to `eval_fragments` and the like.
+    pub fn get_open_package_global_items(&self) -> Vec<(Vec<Rc<str>>, Rc<str>, fir::StoreItemId)> {
```
minestarks, line 355: Elsewhere in this module we call this the `current` package or just `package` (unfortunate choice but here we are). I'd stick with that naming to avoid introducing another term into the same file. (I realize "open package" means something for the incremental compiler, but that's not the same layer of abstraction as this file, so the names change slightly).

The naming of the various packages at play have been a persistent pain point for me so I'll be picky about consistency here, sorry!

compiler/qsc_hir/src/global.rs
swernli: I don't quite agree... yes this is a change in that the global iterator now picks up items that are defined in the top level of a notebook cell where previously it didn't. I would say this is actually fixing a subtle bug we didn't realize we had but that never made a difference before. The global iterator would only show callables that were members of a namespace, which means when you iterated the open package for a notebook you wouldn't see any callables that were defined as top-level statements, only those defined in namespaces (which are supported in cells but discouraged because it has weird duplicate definition behaviors). The reason this didn't end up mattering is that we only used `global::iter_package` in a few places that occur before notebook compilation comes into the picture:
 - When constructing a new package store from an existing compile unit (compiler/qsc_frontend/src/compile.rs line 227)
 - When adding an external package dependency to a compilation (compiler/qsc_frontend/src/resolve.rs line 1621)
 - When running passes on the core library (compiler/qsc_passes/src/lib.rs line 159)

It just so happens that all these never need to run on items defined in a notebook cell without a namespace, so we never noticed they were missing from the iteration. Adding the new case to the iteration doesn't affect any of these locations because none of those run on packages that items defined outside of a namespace.

compiler/qsc/src/interpret.rs
swernli: Let's leave that for another change; while I could update everything in this file it would trickle into callers and then the rest of the code would be half converted.

compiler/qsc/src/interpret.rs
swernli: Good call, I'll update this to be consistent with the patterns in the file.

compiler/qsc/src/interpret.rs
```diff
@@ -482,6 +549,32 @@ impl Interpreter {
         self.run_with_sim(&mut sim, receiver, expr)
     }
 
+    /// Invokes the given callable with the given arguments using the current environment, simlator, and compilation.
```
minestarks, line 554: Have you given any thought into making this behave like `run`, in that it runs on a brand new simulator instance?

After all it's more or less intended to be a more ergonomic way of calling `run`... right?

minestarks (approved):
Ok, I left a lot of comments on the design, approved presuming we'll reach a resolution on those. I think it came together really cleverly - thank you!

compiler/qsc/src/interpret.rs
swernli: Right now that could work, yes, but if we eventually want to support returning more complicated types across the boundary into Python (like callables, lambdas with captures, or even qubits) they have to correlate back to the environment and simulator that is being kept alive. So in that sense, this is more like `eval_fragments` than it is like `run`. I'll move the function up to be near those to support that association and make it easier on our future selves if/when we need that.

compiler/qsc/src/interpret.rs
```diff
@@ -315,6 +318,72 @@ impl Interpreter {
         })
     }
 
+    /// Given a package ID, returns all the global items in the package.
+    /// Note this does not currently include re-exports.
+    pub fn get_global_items(
+        &self,
+        package_id: PackageId,
+    ) -> Vec<(Vec<Rc<str>>, Rc<str>, fir::StoreItemId)> {
```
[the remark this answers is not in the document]
swernli: I think you are right, that will probably be cleaner. I'll update the API and keep `fir::StoreItemId` as an internal thing in favor of `Value` that's already an interop concept. I also like your naming suggestion you made in the other chat, so I'll use `user_globals()` and `source_globals()` is very clean!

compiler/qsc_hir/src/global.rs
minestarks: OK... I've actually changed my mind, you're right. Your example from the side chat is a pretty convincing example demonstrating how these are actual globals:

image

compiler/qsc/src/interpret.rs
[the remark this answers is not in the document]
swernli: I made the change, and it works but honestly feels a little awkward. Because I can't return a `Value` directly across into Python and `ValueWrapper` doesn't support round-tripping `GlobalCallable` I end up having to pull out the `fir::StoreItemId` and `FunctorApp` in a conversion method. While I could try to put this conversion logic into `ValueWrapper` so that `Value::Global` gets converted into `GlobalCallable` that just moves the conversion logic and involves another class without really simplifying anything. I pushed up the change, because from the perspective of the interpret.rs file everything looks clean, but now the Rust code that does Python interop handles all of the weird conversions itself and still touches `fir::StoreItemId` so other than APIs in one file I'm not sure this gained us much.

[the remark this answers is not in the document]
minestarks: I'm surprised the interop ended up being ugly, but I still think it's a worthy tradeoff for keeping the native interface clean. That one gets used by more consumers so it deserves the extra care IMO

idavis (approved):

PR #2055, opened 2024-12-06
Move VS Code language service modules into their own folder

minestarks:
With the new feature I'm planning to add to VS Code extension, I needed to organize this folder a bit before I proceed. It was getting out of hand.

This PR moves everything language-service related into its own folder under `vscode/src`.

The only slight, expected, behavior change here is in the initialization order of diagnostics (which had to happen as a result of me separating out the different kinds of diagnostics.) I don't think it should have any observable effect.


swernli (approved):

billti (approved):
Thanks!

PR #2059, opened 2024-12-10
Test Explorer  in VS Code Extension and `@Test()` Attribute

sezna:
This PR is for enabling VS Code's Test Explorer functionality in our extension, and it introduces the `@Test()` attribute on callables. It also switches our libraries over to use `@Test()` instead of the old `Main`-based pattern.

Any callable that takes no arguments can be annotated as a `@Test()`. If it does not crash (via `Fact` or otherwise), it is considered a pass. If it does crash, then it is considered a failure.

After this PR is in, when you open a .qs file, you'll see this icon: 
image

Clicking this icon will take you to the test explorer:
image

The test explorer shows all discovered tests in their namespace hierarchy. Clicking the "run" play button on a namespace runs all child tests contained in that namespace. A test turns green if its last run passed, or red if it didn't.

### "Run Test" button next to the test itself 
image

### "Failed test" icon on failed tests
image

# Notes:
1. This PR also fixes a bug where the Q# event handler would print `[object Object]` as the output of a failed run.
2. This PR adds an event emitter to the vscode extension. This event emitter is used to emit events any time `updateDocument` is triggered. The test explorer listens for this event to refresh tests for a specific URI. The test discovery code has to scan for tests across an entire package, since tests could be added to a namespace from any file due to explicit namespaces. But only the updates from the new URI are actually populated to the test explorer.

# Potential future work:
* Support test cancellation tokens
* Support "continuous mode" where we auto-rerun tests based on watching for file changes
* Test coverage tracking (VS Code supports tracking lines of code coverage through the test explorer)
* Debug/Coverage profile modes (currently, only the "Run" profile is supported)
* A "run test" codelens on callables annotated with `@Test()` (this was deferred for now as I want that experience to integrate with the Debugger service and that requires a bit more work).


compiler/qsc_frontend/src/lower.rs
```diff
@@ -443,6 +443,7 @@ impl With<'_> {
                     None
                 }
             },
+            Ok(hir::Attr::Test) => Some(hir::Attr::Test),
```
swernli, line 446: This case is missing the validation that no arguments are provided to the attribute by checking for an empty tuple.

wasm/src/test_explorer.rs
```diff
@@ -0,0 +1,71 @@
...
+        unit.package
+    });
+
+    let items_with_test_attribute = package
+        .items
+        .iter()
+        .filter(|(_, item)| item.attrs.iter().any(|attr| *attr == Attr::Test));
+
+    let callables = items_with_test_attribute
+        .filter(|(_, item)| matches!(item.kind, qsc::hir::ItemKind::Callable(_)));
```
swernli, line 36: I wonder if this could be a place to take advantage of the `qsc_hir::globals::iter_package` that we have. I only know it's there because I recently used it for another change. Not sure if it adds much or not, but it might save you the work below of finding the parent namespace as it already has the logic within the iterator to collect the namespace as an array of strings.

swernli: oh, hmm... but then you'd end up having to do extra work to look up the items again to see if they have the test attribute or modify the `Global` to include something that has the information you need.

wasm/src/test_explorer.rs
minestarks, line 1: Can we please please please push this logic somewhere down in the `compiler/` folder? It's too much business logic for the interop layer.

minestarks: More concretely, what I mean is, `collect_test_callable` essentially implements the spec of "how Q# tests are discovered".

Since the `Test` attribute is something that's hardcoded into the compiler at this point, it stands to reason that _how_ that attribute is used it also something that's hardcoded into the compiler. Or at least _close_ to the compiler, which to me means the `qsc` module.

There's nothing about the test discovery behavior that should be specific to the WASM module or JavaScript.

Pushing it down to the `qsc` module and putting it behind clear API should also make further changes easier should we decide to expose test discovery through Python and/or the command line. But even if we never go down that path I think it still makes sense for the above reasons.

Manvi-Agrawal:
@sezna Quick question: After this work, is there a plan of reworking kata infrastructure based on @Test attribute. That significantly reduce migration effort of Katas. 

If that happens, I envision adding support for @Test attribute and new migration to take advantage of that. Then, the existing Katas in this repo could be migrated to @Test attribute. Thoughts?

PS: sent from mobile. Not responsible for autocorrect errors

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+import {
+  getCompilerWorker,
+  ICompilerWorker,
+  IProjectConfig,
+  log,
+  ProgramConfig,
+  QscEventTarget,
+} from "qsharp-lang";
+import { getActiveQSharpDocumentUri } from "./programConfig";
+import { IProgramConfig } from "../../npm/qsharp/lib/web/qsc_wasm";
```
billti, line 17: You should be importing from `qsharp-lang`, not walking up and down the directory tree

billti: (i.e. this should be in the list imported on line 8)

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Constructs the handler to pass to the `TestController` that refreshes the discovered tests.
+ * if `shouldDeleteOldTests` is `true`, then clear out previously discovered tests. If `false`, add new tests but don't dissolve old ones.
+ *
+ */
+function mkRefreshHandler(
+  ctrl: vscode.TestController,
+  context: vscode.ExtensionContext,
+  shouldDeleteOldTests: boolean = true,
```
billti, line 64: This function is only called twice, once without `shouldDeleteOldTests` (so it defaults to `true`), and once passed `true` explicitly, so is there any value in having this parameter?

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  ctrl.resolveHandler = async (item) => {
+    if (!item) {
+      context.subscriptions.push(...startWatchingWorkspace(ctrl, context));
+      return;
+    }
+  };
+
+  function updateNodeForDocument(e: vscode.TextDocument) {
+    if (e.uri.scheme !== "file") {
```
billti, line 153: I believe this means Test Explorer won't work for the playground (which has its own virtual file system), perhaps GitHub (which has it's own scheme) and maybe others. Any reason to restrict this to 'file' URIs?

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+    });
+    watcher.onDidChange(async () => {
+      await refresher();
+    });
+
+    watcher.onDidDelete(async () => {
+      await refresher();
+    });
+
+    // findInitialFiles(controller, pattern);
```
billti, line 212: Delete

vscode/src/testExplorer.ts
swernli: I think the virtual file system and the URI scheme are separate; things in a vfs are still scheme `file`. It works as expected in the browser playground for me:
image

swernli: By restricting to `file` scheme it makes sure tests are detected from things like diffs or files open from PRs (the fixes in #1754 and #1867 respectively).

swernli: Actually, looking at those old fixes, I wonder if it would be best to use `isQsharpDocument` from common for this, since we already rely on that logic for reporting errors and other things.

billti: (in fact in the url in your screen shot you can see the first param for the folder to open starts with `qsharp-vfs://` (url encoded). I'll figure it out and report back. Might be worth a wiki page on how alternate schemes like playground and github are handled for future reference)

billti: > Actually, looking at those old fixes, I wonder if it would be best to use isQsharpDocument from common for this

Good call. And there we explicitly excluded schemes we know to be problematic, which allows us to be flexible for other schemes we _should_ be able to handle.

vscode/src/testExplorer.ts
sezna: This is a remnant of refactoring. Removed.

vscode/src/testExplorer.ts
sezna: I can switch to that. I was following the patterns used in the sample from the VS Code docs. Good call on using `!isQsharpDocument` -- updated.

vscode/src/testExplorer.ts
sezna: Ah yeah, LSP auto-import hit me here. Updated.

vscode/src/language-service/codeLens.ts
```diff
@@ -7,7 +7,7 @@ import {
   qsharpLibraryUriScheme,
 } from "qsharp-lang";
 import * as vscode from "vscode";
-import { toVscodeRange } from "../common";
+import { toVsCodeRange } from "../common";
```
minestarks, line 10: Okay, but now the capitalization doesn't match the other `toVscode*` functions from that module. (Also I totally think of vscode as one word...)

npm/qsharp/src/main.ts
```diff
@@ -91,4 +91,9 @@ export function getLanguageServiceWorker(): ILanguageServiceWorker {
   );
 }
 
+export function collectTestCallables(config: IProgramConfig): string[] {
+  ensureWasm();
+  return wasm!.collect_test_callables(config);
+}
+
```
minestarks, line 98: You shouldn't need these changes in this file  - `Compiler`/`ICompiler` are already reexported through here (which, for context, is the Node.js specific module entrypoint, and is only used in tests)

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -77,6 +78,8 @@ export interface ICompiler {
     exerciseSources: string[],
     eventHandler: IQscEventTarget,
   ): Promise<boolean>;
+
+  collectTestCallables(program: IProgramConfig): Promise<string[]>;
```
minestarks, line 82: Totally a nitpick but I feel like `getTestCallables` is a simpler and nicer name.

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (!projectConfig) {
+    log.info("No project detected; not starting test explorer");
+    return null;
+  }
+
+  return {
+    profile: getTarget(),
+    ...projectConfig,
+  };
+}
```
minestarks, line 54: Just use `getActiveProgram` from `programConfig.ts`, it's intended for this exact purpose and handles edge cases.

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+
+function localGetCompilerWorker(
+  context: vscode.ExtensionContext,
+): ICompilerWorker {
+  const compilerWorkerScriptPath = vscode.Uri.joinPath(
+    context.extensionUri,
+    "./out/compilerWorker.js",
+  ).toString();
+  const worker = getCompilerWorker(compilerWorkerScriptPath);
+  return worker;
```
minestarks, line 29: Bah, these two lines are duplicated in so many places (not your fault). If you're feeling boy scout-y, I'd love to have it refactored into a common utility.

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,267 @@
...
+    watcher.onDidCreate(async () => {
+      await refresher();
+    });
+    watcher.onDidChange(async () => {
+      await refresher();
+    });
+
+    watcher.onDidDelete(async () => {
+      await refresher();
+    });
```
minestarks, line 200: IIRC the `onDid*` methods always return objects you need to push to `context.subscriptions` so they get unsubscribed when the extension is unloaded/reloaded (best practice to avoid event handler leaks). Basically, the same pattern as the `workspace.onDid*` calls above

vscode/src/language-service/codeLens.ts
sezna: Ah, I was trying to match `toVsCodeDiagnostic`, and also vs code's branding which states that it is "VS" and then "Code" as two words. I can revert this if you feel strongly

minestarks:
Code looks great but I think there are significant changes here that are all missing tests. Please add: 
- tests for `collect_test_callables` (and other relevant native code)
- error handling from JavaScript (maybe in `npm/qsharp/test`) - I see that `collect_test_callables` is fallible, how do errors propagate up?
- at least one integration test for the VS Code extension code.
- Q# library tests .... using the test framework you just added :) We can run that in CI right?

vscode/src/language-service/codeLens.ts
minestarks: Dammit I didn't notice `toVsCodeDiagnostic` :( I just wanted them all to match and was trying to save you some unnecessary changes. I don't really mind which spelling we choose as long as we're converging.

wasm/src/test_explorer.rs
sezna: @swernli  and I had a discussion about this and the end result was to add a pass that detected test attribute issues and returned more user-friendly errors. I've added this pass in 5a28ef104b5d9c0e23ec2d2c76c9c66d797f42a7

wasm/src/test_explorer.rs
sezna: I've moved this logic into the compiler itself in c22d14dac1c98c1abf6662fa3fe440994f4f400f

vscode/src/language-service/codeLens.ts
sezna: Would you rather me revert this, or perhaps do a quick rename-symbol on the rest?

npm/qsharp/src/compiler/compiler.ts
sezna: I chose `collect` because it is actually performing iteration and collection logic, while `get` implies getter-style functionality (just reading a field, little to no computation). I'm happy to switch to `get` if there's a compelling reason though

vscode/src/testExplorer.ts
sezna: Oh good call, I missed that. Updated.

sezna: 16a2aeda

vscode/src/testExplorer.ts
sezna: I'm afraid I don't fully understand -- quick clarifying question incoming. The watcher itself is returned and added to `context.subscriptions` ([ref](https://github.com/microsoft/qsharp/pull/2059/files/7d2aadc70a74f93516ab8423c919a0cedcc4340b#diff-a3660ba9ec05210e62ad06609feb55c16cf3678ecce0b036ed0c3232a8559aaeR147)) -- are you saying that the watcher itself should _not_  get returned and instead I should return the result of the `onDid*` calls and let the watcher get dropped?

sezna: For your reference I wrote this code based on [this example from VS Code](https://github.com/microsoft/vscode-extension-samples/blob/main/test-provider-sample/src/extension.ts#L204).

compiler/qsc_passes/src/test_attribute.rs
```diff
@@ -0,0 +1,45 @@
...
+use qsc_data_structures::span::Span;
+use qsc_hir::{hir::Attr, visit::Visitor};
+use thiserror::Error;
+
+#[cfg(test)]
+mod tests;
+
+#[derive(Clone, Debug, Diagnostic, Error)]
+pub enum TestAttributeError {
+    #[error("This callable has parameters. Tests cannot have parameters.")]
```
swernli, line 14: nitpick: I think the phrasing could be updated to match the style of our other errors, so be one sentence declaring what the problem is, sort of like how the measurement attribute errors look. So something like, "a callable with the test attribute cannot have parameters"

compiler/qsc_hir/src/hir.rs
```diff
@@ -279,6 +279,53 @@ impl Display for Package {
...
+
+        let callable_names = callables
+        .filter_map(|(_, item)| -> Option<std::result::Result<String, String>>{
+            if let ItemKind::Callable(callable) = &item.kind {
+                if !callable.generics.is_empty() {
+                    return Some(Err(format!("Callable {} has generic type parameters. Test callables cannot have generic type parameters.", callable.name.name)));
+                }
+                if callable.input.kind != PatKind::Tuple(vec![]) {
+                    return Some(Err(format!("Callable {} has input parameters. Test callables cannot have input parameters.", callable.name.name)));
+                }
```
swernli, line 301: now that we have the pass to validate the errors, maybe this function can just silently filter them out? Then this function can return the ones it finds and not need to error out if one is invalid.

sezna: Elsewhere in the compiler we tend to keep invalid things around for the sake of recovery. If other passes are added after this, in my opinion, we'd probably want them to have access to these callables for better user feedback.

swernli (approved):

billti (approved):

vscode/src/testExplorer.ts
minestarks: I _was_ saying both `watcher` and these return values should get pushed the `subscriptions`... but now that I've looked at the sample code you referenced, clearly either I'm wrong or it's not important to VS Code. 

Exactly what you should do with these return values is undocumented and I've never been sure, just made my own inferences. Copying sample code is probably the safest. Feel free to resolve.

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,234 @@
...
+    return;
+  }
+  const run = ctrl.createTestRun(request);
+  const evtTarget = new QscEventTarget(false);
+  evtTarget.addEventListener("Message", (msg) => {
+    run.appendOutput(`Test ${testCase.id}: ${msg.detail}\r\n`);
+  });
+
+  evtTarget.addEventListener("Result", (msg) => {
+    if (msg.detail.success) {
```
minestarks, line 211: 1. I think listeners need to be removed with `removeEventListener` after the test case is run? (If you copied the debugger code, I think the same applies there too)

2. Just checking, do you want to handle `"DumpMachine"` and `"DumpMatrix"` too, or is it intentionally filtered to messages?

3. (BTW, there is a more performant `uiresultsrefresh` event, which is an alternative to handling these individual events, that batches message/dumpmachine/results outputs so you don't have a million callbacks into JS if you're, for example, writing `Message()` in a loop from Q#. If performance becomes a concern that's worth looking into. I don't necessarily think it has to be done right now though).

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,234 @@
...
+        },
+      };
+      run.failed(testCase, message);
+    }
+    run.end();
+  });
+
+  const callableExpr = `${testCase.id}()`;
+  try {
+    await worker.run(program, callableExpr, 1, evtTarget);
```
minestarks, line 228: remember to destroy the worker

wasm/src/test_explorer.rs
minestarks, line 1: nittiest of nits: "Test Explorer" is a VS Code feature but this is the WASM layer (which is IDE agnostic). I'd name it something a tad more generic like "test" or "test discovery".

npm/qsharp/src/compiler/compiler.ts
minestarks: Only because `getHir`, `getAst` , `getEstimates` etc exist (and they're all doing more than "get"ting too). My only ask is that you take a quick look at the overall public API in this file and decide for yourself what name would fit best.

vscode/src/language-service/codeLens.ts
minestarks: I don't mind if you rename-symbol the rest, since that would make them all match

npm/qsharp/src/compiler/compiler.ts
sezna: Alright, that convinces me, I like consistency.

vscode/src/testExplorer.ts
sezna: This ended up getting removed anyway since we trigger test refreshes on `updateDocument` now

compiler/qsc_hir/src/hir.rs
```diff
@@ -279,6 +279,52 @@ impl Display for Package {
     }
 }
 
+impl Package {
+    /// Returns a collection of the fully qualified names of any callables annotated with `@Test()`
+    pub fn collect_test_callables(&self) -> std::result::Result<Vec<String>, String> {
```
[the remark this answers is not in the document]
sezna: good catch! fixed

vscode/src/testExplorer.ts
sezna: I decided to remove these manual listeners and reuse the debugger event target, which has all of the dumpmachine rendering logic built-in. While doing this, I found a bug in that renderer for `Result` types and fixed that as well.

vscode/src/testExplorer.ts
sezna: reusing the worker now

minestarks (approved):

compiler/qsc_hir/src/hir.rs
```diff
@@ -279,6 +279,57 @@ impl Display for Package {
...
+        let callables = items_with_test_attribute
+            .filter(|(_, item)| matches!(item.kind, ItemKind::Callable(_)));
+
+        let callable_names = callables
+            .filter_map(|(_, item)| -> Option<_> {
+                if let ItemKind::Callable(callable) = &item.kind {
+                    if !callable.generics.is_empty()
+                        || callable.input.kind != PatKind::Tuple(vec![])
+                    {
+                        return None;
```
billti, line 302: So does this just silently drop a callable with the Test attribute if it happens to be generic? No warning or error?

compiler/qsc_passes/src/test_attribute/tests.rs
```diff
@@ -0,0 +1,79 @@
...
+        sources,
+        TargetCapabilityFlags::all(),
+        LanguageFeatures::default(),
+    );
+    assert!(unit.errors.is_empty(), "{:?}", unit.errors);
+
+    let errors = validate_test_attributes(&mut unit.package);
+    Validator::default().visit_package(&unit.package);
+    if errors.is_empty() {
+        expect.assert_eq(&unit.package.to_string());
```
billti, line 27: Is this line ever used? (i.e. it seems all the tests here expect some type of error). Maybe remove the dead code, or add a test which tests the success case.

compiler/qsc_hir/src/hir.rs
sezna: The error is detected in `qsc_passes`. Returning an error here doesn't show a pretty error to the user, so instead we detect it in passes and produce a spanned error. This pattern was suggested by @swernli and I think it is cleaner than returning an error here, which just prints an unspanned error in a VS Code dialog box.

vscode/src/testExplorer.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const programResult = await getActiveProgram();
+    if (!programResult.success) {
+      throw new Error(programResult.errorMsg);
+    }
+
+    const program = programResult.programConfig;
+
+    for (const testCase of request.include || []) {
+      await runTestCase(ctrl, testCase, request, worker, program);
```
billti, line 124: I'm not sure I follow the use of the "common worker" here. I thought we discussed the language service could return the list of tests (as this has already parsed/compiled the project on every update and is thus close to zero cost), but the tests should run in an isolated worker for reliability.

Here is seems we're using a common worker to find all the tests (thus on every 'update document' we're compiling the project twice, once for the language service and once for the test discovery - even though most projects will have no tests), and are also running all the tests in that same common worker - not isolated.

Or am I reading this wrong?

compiler/qsc_hir/src/hir.rs
billti: Yeah I saw those checks there. Just wondering the value of even testing for it here also, as the project is already in error from the other pass that detects these, and thus I assume shouldn't run anyway.

sezna: It just prevents the test from showing up in the list. The pass doesn't filter out the callables from the HIR, because we prefer to preserve as much of the AST as possible for error recovery in passes. So even in an error state, the compilation _does_ return some HIR for a good user experience. We can't let those invalid callables show up in the test explorer right?

vscode/src/testExplorer.ts
sezna: I see what you're saying now, thanks for typing it out. After our in-person discussion, I had it in my mind that I needed to a) mitigate the worker leak and b) reuse the worker from the language service. I had those two goals in mind when I wrote that code. 

I like the approach of exposing a specific "get test callables" function in the compiler API, because it allows for other scenarios to consume it and write testing infrastructure. Integrating that functionality into the language service to be a language-service-exclusive feature sounds like signing ourselves up for more work in the future.

I think a good approach here would be to modify `compile` such that the returned asset (a `CompileUnit`) contains the list of test callables, and remove `getTestCallables()` elsewhere. Thoughts?

PR #2060, opened 2024-12-11
Added Ising model sample

DmitryVasilevsky:
This adds Ising model sample with explanations.


msoeken:
Thanks @DmitryVasilevsky , this would be the third Ising model simulation in the repository alongside https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-ising-2D.ipynb and https://github.com/microsoft/qsharp/blob/main/samples/estimation/Dynamics.qs. Maybe we could remove Dynamics.qs in favor of this file and load this new file into estimation-ising-2D.ipynb and only do the explanation and RE in there.

Second, it would be nice, if the the simulation was wrapped into a quantum phase estimation to determine the ground state energy.

aarthims:
I skimmed through the code and here is one thought I had. @msoeken there is a difference in the [estimation-ising-2D](https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-ising-2D.ipynb) and this Ising model simulation in that the former is doing 4th Order Trotterization and this code does 1st Order Trotterization. 

As a demo, both do work. However, for utility scale we would want to use 4th order formulas as they minimize errors in the final output. 

I do agree that the [Dynamics](https://github.com/microsoft/qsharp/blob/main/samples/estimation/Dynamics.qs) sample and the [estimation-ising-2D](https://github.com/microsoft/qsharp/blob/main/samples/estimation/estimation-ising-2D.ipynb) can be rolled into one. 

@DmitryVasilevsky if you can wrap this code with phase estimation (i.e., the simulation will serve as the oracle for phase estimation) as @msoeken suggested, that will be a valuable addition as none of our samples show that at this time.

DmitryVasilevsky:
> Thanks @DmitryVasilevsky , this would be the third Ising model simulation in the repository alongside... 

We discussed this and we would like to have a simple standalone sample of Ising model (a more advanced algorithm) that is suitable for base profile. Changing existing samples and adding phase estimation we may consider later.

DmitryVasilevsky:
> As a demo, both do work. However, for utility scale we would want to use 4th order formulas as they minimize errors in the final output.

Thank you for reviewing! We would like to have a standalone simple sample of a more advanced algorithm. Hence this sample. It is suitable for base profile. Changes to other samples and phase estimation for eigenvalue we may consider later; I will open an issue for this.

swernli (approved):

billti (approved):

DmitryVasilevsky:
Added issue https://github.com/microsoft/qsharp/issues/2067

PR #2061, opened 2024-12-12
Move to Python 3.9 as min ver and Ubuntu 22.04 min compat ver

### LINKED ISSUE #1946: Deprecate Python 3.8, making 3.9 the oldest target Python version

idavis:
Fixes #1946

- Cross build image is now set to CBL Mariner with glibc 2.35 and Python 3.9
- Manylinux tag is now `manylinux_2_35_x86_64`
- A docker image is no longer built, we directly use the Mariner Python base image.


swernli (approved):

ScottCarda-MS (approved):

minestarks (approved):
Nice, thank you! Have you ran a release pipeline to test?

PR #2062, opened 2024-12-12
Deprecate `set` Keyword

### LINKED ISSUE #1496: Make `set` keyword optional and include lint to suggest avoiding it.
The `set` and `let` keywords are very similar and easy to confuse, leading to accidental shadowing sometimes. In addition, `let` is already sufficient to disambiguate declaration/binding from assignment, so `set` doesn't provide any actual value syntactically. We can leave the keyword in but make it optional in the parser, and then later follow up with a lint rule about it.

ScottCarda-MS:
Makes the `set` keyword optional for assignment expressions. Adds a lint and code action to deprecate the use of the `set` keyword. The lint level is set to "Allow".

Closes #1496

### DIFF
--- a/compiler/qsc_linter/src/linter/ast.rs
+++ b/compiler/qsc_linter/src/linter/ast.rs
@@ -50,65 +50,71 @@ pub fn run_ast_lints(
 /// The trait provides default empty implementations for the rest of the methods,
 /// which will be optimized to a no-op by the rust compiler.
 pub(crate) trait AstLintPass {
-    fn check_attr(&self, _attr: &Attr, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
-    fn check_block(&self, _block: &Block, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
+    fn check_attr(&mut self, _attr: &Attr, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
+    fn check_block(&mut self, _block: &Block, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
     fn check_callable_decl(
-        &self,
+        &mut self,
         _callable_decl: &CallableDecl,
         _buffer: &mut Vec<Lint>,
         _compilation: Compilation,
     ) {
     }
-    fn check_expr(&self, _expr: &Expr, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
+    fn check_expr(&mut self, _expr: &Expr, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
     fn check_functor_expr(
-        &self,
+        &mut self,
         _functor_expr: &FunctorExpr,
         _buffer: &mut Vec<Lint>,
         _compilation: Compilation,
     ) {
     }
-    fn check_ident(&self, _ident: &Ident, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
-    fn check_item(&self, _item: &Item, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
+    fn check_ident(&mut self, _ident: &Ident, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
+    fn check_item(&mut self, _item: &Item, _buffer: &mut Vec<Lint>, _compilation: Compilation) {}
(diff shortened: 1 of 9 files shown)


compiler/qsc_linter/src/linter/ast.rs
```diff
@@ -106,25 +152,20 @@ macro_rules! declare_ast_lints {
...
-            }
-        }
-
         impl $lint_name {
             const DEFAULT_LEVEL: LintLevel = $default_level;
 
+            fn new() -> Self {
+                #[allow(clippy::needless_update)]
+                Self { level: Self::DEFAULT_LEVEL, ..Default::default() }
+            }
```
swernli, line 167: It looks like this and some of the other changes to this file may be left over from a different design... the lint is warning that the only element of the struct has already been set, which is true. So the existing `default` implementation was correct and can be restored, this `new` can be removed and places where it was called returned to calling `default`.

compiler/qsc_linter/src/tests.rs
```diff
@@ -14,6 +14,32 @@ use qsc_frontend::compile::{self, CompileUnit, PackageStore, SourceMap};
...
+fn set_keyword_lint() {
+    check(
+        &wrap_in_callable("set x = 3;", CallableKind::Function),
+        &expect![[r#"
+            [
+                SrcLint {
+                    source: "set",
+                    level: Allow,
+                    message: "deprecated use of `set` keyword",
+                    help: "`set` keywords are deprecated for assignments, they can be removed",
```
swernli, line 27: nitpick: I think I prefer this phrasing
```suggestion
                    help: "`set` keywords are deprecated for assignments and can be removed",
```

swernli (approved):
A couple minor comments on the lint portion, but otherwise quite happy with how this turned out!

compiler/qsc_linter/src/linter/ast.rs
ScottCarda-MS: The design for the `default` doesn't work for when we want to have lints that have state. 

It is better, and less confusing, imho that the design for HIR lints and AST lints match so that devs don't need to memorize two different ways of making lints. Having feature parity earlier would have prevented certain lints from being biased towards being implemented as HIR lints because of the added flexibility instead of what would probably have been more appropriately AST lints.

ScottCarda-MS: As such, in this PR, instead of making use of the more flexibly design of HIR lints, I am adding these features to the AST lint design and making AST lints to make the AST and HIR lints have feature parity.

swernli: Oh, I see what you mean. With your latest changes now there is an AST lint that has more state; in the earlier version this change wasn't needed but now it's necessary. Thanks for clarifying!

swernli (approved):
Looked at the new lint, phrasing is consistent and should be helpful to folks!

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -28,10 +28,17 @@ declare_ast_lints! {
     (NeedlessParens, LintLevel::Allow, "unnecessary parentheses", "remove the extra parentheses for clarity"),
     (RedundantSemicolons, LintLevel::Warn, "redundant semicolons", "remove the redundant semicolons"),
     (DeprecatedNewtype, LintLevel::Allow, "deprecated `newtype` declarations", "`newtype` declarations are deprecated, use `struct` instead"),
+    (DeprecatedSet, LintLevel::Allow, "deprecated use of `set` keyword", "the `set` keyword is deprecated for assignments and can be removed"),
+    (DiscourageDaisyChain, LintLevel::Warn, "discouraged use of chain assignment", "assignment expressions always return `Unit`, so chaining them may not be useful"),
```
orpuente-MS, line 32: Should we call the lint `DiscourageChainAssignment` instead? I have never seen the term daisy chain and we refer to it as "chain assignment" in the lint and help text.

orpuente-MS (approved):

compiler/qsc_linter/src/lints/ast.rs
ScottCarda-MS: done

PR #2064, opened 2024-12-12
use common extension ID instead of hardcoded one

sezna:
This was a drive-by update from my work in #2059. Just wanted it in a separate PR to keep the noise down.

It seems to have triggered some `prettier` fixes too.


billti:
Was there a motivating reason for this change? (i.e. do we ever change it during development, testing, or publishing?)

sezna:
I was registering a new code action and noticed that we do this inconsistently across other files. If we have a source of truth defined somewhere we might as well use it consistently or not at all.

swernli (approved):

minestarks (approved):
Personally I preferred the hardcoded versions, since when I'm trying to find a specific command's implementation, usually I try to find the command's user-friendly description in `package.json` and copy-paste the command name (e.g. `"qsharp-vscode.createProject"`) and search for that in the source.

I don't mind the change though since I'm a fan of consistency.

Issue #1384, opened 2024-04-14
Sample for modern QDK project with Python 


Manvi-Agrawal:
**Is your feature request related to a problem? Please describe.**
Its a bit cumbersome to copy-paste code from Sample for modern QDK project with Python as described [here](https://learn.microsoft.com/en-us/azure/quantum/user-guide/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-python-run#steps-for-creating-a-q-project). 

**Describe the solution you'd like**
Discussed with @tcNickolas who recommends creating a sample for this in the qsharp repository itself, since it's the main source of truth for Q# samples.

billti:
Like this one? https://github.com/microsoft/qsharp/blob/main/samples/notebooks/project.ipynb

Manvi-Agrawal:
This notebook is nice. It would be good if documentation could point to this notebook 📒

tcNickolas:
The notebook sample is great for showing how to work with Q# only from a notebook, without Python. 

* It needs to be updated (or a new sample created) to show how to pass arguments from Python to Q#, and get the results from Q# into Python for post-processing.
* We also need to cover the scenario of Q# + Python non-notebook project. Python files don't have Q# magic to rely on, so they need to use qsharp.eval, the syntax is not trivial to figure out.

billti:
We do have the double factorized chemistry sample for resource estimation which is straight Python and Q# (https://github.com/microsoft/qsharp/blob/main/samples/estimation/df-chemistry/chemistry.py#L472), but agree we could do with something more accessible.

tcNickolas:
I would suggest creating a new folder under https://github.com/microsoft/qsharp/blob/main/samples/ called "python_interop", and add two projects to it, each in its own folder:

1. [Teleportation project from our documentation](https://learn.microsoft.com/en-us/azure/quantum/user-guide/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-python-run#steps-for-creating-a-q-project), covering the basic no-inputs-no-outputs scenario, just calling Q# from Python, so that we can point to that project from that docs page.
2. A slightly more complicated project that would call Q# with some input parameter and get some return value from it. 
[...]

PR #2066, opened 2024-12-13
Add Python interop sample notebook

swernli:
This adds a Jupyter notebook sample for invoking Q# callables from Python. I confirmed the new notebook runs in integration tests as expected.


samples/python_interop/calling_qsharp.ipynb
Manvi-Agrawal, line 1: This looks quite similar to `samples/notebooks/sample.ipynb` on a high level. See earlier discussion in https://github.com/microsoft/qsharp/issues/1384 regarding python interop.  Probably we can update that notebook. Thoughts?

Although, I can completely understand that one might think that python interop should have a python notebook that explains about calling python functions. :-)

swernli: That's a good point. Rather than introduce an entirely new notebook, this pattern could just be incorporated into the existing samples notebooks. I'll look through the others and see if that end up making more sense, then update the PR based on what I find.

swernli: I went ahead and merged most of this content in sample.ipynb to reduce the number of samples and make the existing one more comprehensive. As a potential follow up, we could revisit the structure of this sample.ipynb and turn into more of a general guide.

Manvi-Agrawal: Thanks @swernli . Completely agree with you. One of the other things could be to move some/all notebooks from "notebooks" folder to python interop folder under "jupyter notebooks"/"notebooks" folder for greater discovery. This is because it took me a while to figure this existing one, and you were kinda unaware of this one, so I was wondering new users will end up getting confused as well. Tagging @tcNickolas, @minestarks, @cesarzc , @billti for additional thoughts? 

PS: sent from mobile. Not responsible for autocorrect errors

minestarks (approved):

samples/python_interop/calling_qsharp.ipynb
minestarks: I don't think these notebooks fit with the theme of what currently exists in the "python interop" folder.

What we call "Python interop" right now is actually "Qiskit Interop" . The name we chose for the folder is overly generic.

I'd probably want to go the other way and move `python_interop` under `notebooks` while renaming it to `qiskit`:

```
--notebooks
------sample.ipynb
------circuits.ipynb
------qiskit.ipynb
-----------qiskit
---------------generating_n_random_bits
---------------teleportation_project
```

But that's probably out of scope for this PR.

swernli: The python_interop folder is more general than you think. The two folders have nothing to do with Qiskit, they just show generic use of interop with Python and Q#. Only the qiskit.ipynb notebook uses Qiskit. As a result, I've gone ahead and updated one of these samples to use the new mechanism for running Q# callables.

minestarks: Oh my bad, I somehow read those as qiskit samples. Then I think they'd be better as notebooks rather than a `.py` file along with a `README.md` ... and they can all live together under `notebooks`?

minestarks: Or maybe just a top level folder called `python`. They don't have to all be notebooks. I can't quite make up my mind about it tbh.

Manvi-Agrawal: I think having a top level folder as Python and notebooks being a part of it is a good idea. One of the reasons I like having some simple examples for python projects to run is that it allows beginners to get comfortable with Python+Q# style projects( #1384 has some discussion around this). They add value addition in terms of code usability and scripting. On the other hand, I tend to think notebooks as kind of doing something exploratory work.

samples/notebooks/sample.ipynb
```diff
@@ -414,6 +426,106 @@
...
+    "except TypeError as e:\n",
+    "    print(f\"TypeError: {e}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0f0795e9",
+   "metadata": {},
+   "source": [
+    "If you define a Q# callables a namespace (or when loading from a Q# project), they will be exposed with a matching hiearchy of modules in Python:"
```
billti, line 466: The opening 6 or 7 words here doesn't read quite right.

billti: Typo in 'hierarchy' also

billti (approved):

PR #2073, opened 2024-12-19
Added simple VQE sample

DmitryVasilevsky:
This PR adds simple self-contained example of a Variational Quantum Eigensolver (VQE).

This example includes:

1. Simple classical optimization to find minimum of a multi-variable function in order to find the approximation to the minimum eigenvalue of a Hamiltonian.
2. Finding Hamiltonian expectation value as a weighted sum of terms.
3. Finding one term expectation value by performing multiple shots of preparing ansatz state and measuring it.
4. Ansatz state preparation similar to the circuit in the published paper "Ground-state energy estimation of the water molecule on a trapped ion quantum computer" (https://arxiv.org/abs/1902.10171)

To keep this sample simple and self-contained Hamiltonian is artificial (doesn't correspond to any actual chemistry problem). Also, this sample doesn't include any application of VQE, for example, finding a geometry of a molecule.

### DIFF
--- a/samples/algorithms/SimpleVQE.qs
+++ b/samples/algorithms/SimpleVQE.qs
@@ -34,26 +34,27 @@ operation Main() : Double {
         // Start from these angles for ansatz state preparation
         [1.0, 1.0],
         // Use initial step pi/8 to find minimum
-        PI()/8.0,
+        PI() / 8.0,
         // Stop optimization if step is 0
         0.0,
         // Stop optimization after 100 attempts to improve
-        100)
+        100
+    )
 }
 
 /// # Summary
 /// Find expectation value of a Hamiltonian given parameters for the
 /// ansatz state and number of shots to evaluate each term.
-operation FindHamiltonianExpectationValue(thetas: Double[], shots: Int): Double {
+operation FindHamiltonianExpectationValue(thetas : Double[], shots : Int) : Double {
     let terms = [
-        ([PauliZ,PauliI,PauliI,PauliI], 0.16),
-        ([PauliI,PauliI,PauliZ,PauliI], 0.25),
-        ([PauliZ,PauliZ,PauliI,PauliI], 0.17),
-        ([PauliI,PauliI,PauliZ,PauliZ], 0.45),
-        ([PauliX,PauliX,PauliX,PauliX], 0.2),
-        ([PauliY,PauliY,PauliY,PauliY], 0.1),
-        ([PauliY,PauliX,PauliX,PauliY], 0.02),
-        ([PauliX,PauliY,PauliY,PauliX], 0.22),
+        ([PauliZ, PauliI, PauliI, PauliI], 0.16),
+        ([PauliI, PauliI, PauliZ, PauliI], 0.25),
+        ([PauliZ, PauliZ, PauliI, PauliI], 0.17),
+        ([PauliI, PauliI, PauliZ, PauliZ], 0.45),
+        ([PauliX, PauliX, PauliX, PauliX], 0.2),
+        ([PauliY, PauliY, PauliY, PauliY], 0.1),
+        ([PauliY, PauliX, PauliX, PauliY], 0.02),
+        ([PauliX, PauliY, PauliY, PauliX], 0.22),
     ];
     mutable value = 0.0;
     for (basis, coefficient) in terms {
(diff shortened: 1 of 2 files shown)


samples_test/src/tests/algorithms.rs
```diff
@@ -258,6 +258,26 @@ pub const SIMPLEISING_EXPECT: Expect =
...
+    Beginning descent from value 0.9795.
+    Value improved to 0.9511000000000001.
+    Value improved to 0.8876000000000001.
+    Value improved to 0.8764000000000001.
+    Value improved to 0.8702000000000001.
+    Value improved to 0.8641000000000002.
+    Value improved to 0.8549000000000001.
+    Descent done. Attempts: 100, Step: 0.00000037450702829239286, Arguments: [1.5886651273511148, 1.3926990816987241], Value: 0.8549000000000001.
+    0.8549000000000001"#]];
+pub const SIMPLEVQE_EXPECT_DEBUG: Expect = expect![[r#"
```
billti, line 271: What's the difference between this and the non-debug version above?

DmitryVasilevsky: There's no difference. Just like in most other sample tests (if not all of them).
When we autogenerate Rust test cases [here](https://github.com/microsoft/qsharp/blob/740c4923bc047adbbb68aa81171dad5bee6b1235/samples_test/build.rs#L63), we include both 'run' and 'debug' version of the test. So both strings are needed with our current approach to sample testing.

samples/algorithms/SimpleVQE.qs
```diff
@@ -0,0 +1,200 @@
...
+/// Prepare the ansatz state for given parameters on a qubit register
+operation PrepareAnsatzState(qs : Qubit[], thetas : Double[]) : Unit {
+    BosonicExitationTerm(thetas[0], qs[0], qs[2]);
+    CNOT(qs[0], qs[1]);
+    NonBosonicExitataionTerm(thetas[1], qs[0], qs[1], qs[2], qs[3]);
+}
+
+/// # Summary
+/// Bosonic exitation circuit
+operation BosonicExitationTerm(
```
msoeken, line 97: It would be great to describe how this circuit was found or provide a reference that explains more.

samples/algorithms/SimpleVQE.qs
```diff
@@ -0,0 +1,200 @@
...
+    Rxx(theta, moX, moY);
+    S(moX);
+    Adjoint S(moY);
+    Rxx(-theta, moX, moY);
+    S(moY);
+}
+
+/// # Summary
+/// Non-bosonic exitation circuit
+operation NonBosonicExitataionTerm(
```
msoeken, line 113: It would be great to describe how this circuit was found or provide a reference that explains more.

Further, it seems that more of the operations can be merged into the `within` block, shortening the code.

msoeken (approved):
Looks good to me, left some comments.

samples/algorithms/SimpleVQE.qs
DmitryVasilevsky: I updated the comment to mention referenced paper, which is referenced in the beginning of the sample.

samples/algorithms/SimpleVQE.qs
DmitryVasilevsky: Moved more operations into within block.
Updated comment to mention referenced paper, which is referenced in the beginning of the file.

samples/algorithms/SimpleVQE.qs
```diff
@@ -0,0 +1,200 @@
+/// # Sample
+/// Simplified Sample of a Variational Quantum Eigensolver
+///
+/// # Description
+/// This is an example of a Variational Quantum Eigensolver (VQE).
+/// This example includes:
+///   1. Simple classical optimization to find minimum of a multi-variable function
+///      in order to find an approximation to the minimum eigenvalue of a hamiltonian
```
swernli, line 8: nitpick: "Hamiltonian" should be capitalized here, I think
```suggestion
///      in order to find an approximation to the minimum eigenvalue of a Hamiltonian
```

swernli: I could be wrong on this, but if not, it might be worth fixing in the comments throughout.

swernli (commented):
Looks good; I think it would be worth adding it to the samples.mjs that controls which samples show up in the editor!

samples/algorithms/SimpleVQE.qs
DmitryVasilevsky: You are right. I'll capitalize this in comments.

DmitryVasilevsky: Updated.

DmitryVasilevsky:
> ... adding it to the samples.mjs that controls which samples show up in the editor!
Not sure if this is a good sample to have as a quick template. Not all samples are in that list, and this sample is a more specialized one.

swernli:
> > ... adding it to the samples.mjs that controls which samples show up in the editor!
> 
> Not sure if this is a good sample to have as a quick template. Not all samples are in that list, and this sample is a more specialized one.

Ok, we can leave part for a later PR if we decide we want to include it.

swernli (approved):

PR #2075, opened 2024-12-19
Update sample notebook

billti:
This is the sample notebook you can create from the command palette (which is very handy to avoid having to download/copy one to get started).

It was a little out of date. This has some minor improvements.

### DIFF
--- a/vscode/src/notebookTemplate.ts
+++ b/vscode/src/notebookTemplate.ts
@@ -99,9 +99,14 @@ import matplotlib.pyplot as plt
 import numpy as np
 from collections import Counter
 
-results = qsharp.run("RandomBit()", shots=1000)
+# Use save_events to suppress the automatic printing of events for every shot
+results = qsharp.run("RandomBit()", shots=1000, save_events=True)
+
 # Sort the results so that the histogram labels appear in the correct order
+# We only care about the result values, not the event data, so extract just the result
+results = [entry["result"] for entry in results]
 results.sort()
+
 # Count the number of times each result appears
 counts = Counter(results)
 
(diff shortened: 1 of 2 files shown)


billti:
I have no idea why yarn.lock is showing up again in this diff. It's the same as what just merged to main.

minestarks (approved):

vscode/src/notebookTemplate.ts
```diff
@@ -99,9 +99,14 @@ import matplotlib.pyplot as plt
 import numpy as np
 from collections import Counter
 
-results = qsharp.run("RandomBit()", shots=1000)
+# Use save_events to suppress the automatic printing of events for every shot
+results = qsharp.run("RandomBit()", shots=1000, save_events=True)
```
billti, line 103: I think this might be related to a change you made @swernli , but it was printing 1000 DumpMachines now when running to just get the results, so had to change here to suppress that.

swernli: The sample always did that, didn't it? There is an explicit `DumpMachine()` in the `RandomBit` operation that's being called. If it wasn't producing output before, that was a bug.

vscode/src/notebookTemplate.ts
```diff
@@ -111,7 +116,35 @@ plt.title("RandomBit() Results")
...
+      value: `from qsharp_widgets import EstimatesPanel
+
+estimate = qsharp.estimate("RandomBit()", [
+    {"errorBudget": 0.333, "qubitParams": {"name": "qubit_gate_ns_e3"}},
+    {"errorBudget": 0.333, "qubitParams": {"name": "qubit_gate_us_e4"}},
+    {"errorBudget": 0.333, "qubitParams": {"name": "qubit_maj_ns_e6"}, "qecScheme": {"name": "floquet_code"}}
+])
+EstimatesPanel(estimate)
+`,
     },
```
swernli, line 148: Does it really make sense to have a resource estimation sample that returns almost no usage? `RandomBit` doesn't have any T gates, CCNOT gates, or rotations, so the counts won't really show anything interesting.

billti: I didn't want to clutter it by adding a more complex algorithm. The notebook is just intended to show API usage. But yeah, could add something more interesting later. (Any ideas for something that fits in a few lines in a cell?)

swernli: No, for bare bones, "here is how you get started with this," I think it's fine then.

swernli (approved):
Noting a side conversation here: maybe this should be renamed "notebook sample" instead of "template" just because of the goal of showing users all that could do rather than a specific step-by-step of what they must do, which is what I tend to think of templates as.

PR #2078, opened 2024-12-26
Adaptive profile floating-point computations extension

idavis:
Adds dynamic float support implementing the floating-point computations for the AdaptiveRIF profile. This also implements `fcmp` missing from the spec which was [filed as a spec bug](https://github.com/qir-alliance/qir-spec/issues/41).

The `fcmp` instructions chosen were the ordered set. The spec says operations like dividing by `0` are undefined behavior and diving by `0` usually gives a `NAN` value. For any `fcmp` calls we need to choose ordered or unordered. I'm not sure that it matters which we decide.

This PR does not change the default profiles for the defined hardware targets.


compiler/qsc/src/interpret/tests.rs
```diff
@@ -968,19 +968,13 @@ mod given_interpreter {
...
                 !1 = !{i32 7, !"qir_minor_version", i32 0}
                 !2 = !{i32 1, !"dynamic_qubit_management", i1 false}
                 !3 = !{i32 1, !"dynamic_result_management", i1 false}
-                !4 = !{i32 1, !"classical_ints", i1 true}
-                !5 = !{i32 1, !"qubit_resetting", i1 true}
-                !6 = !{i32 1, !"classical_floats", i1 false}
-                !7 = !{i32 1, !"backwards_branching", i1 false}
-                !8 = !{i32 1, !"classical_fixed_points", i1 false}
-                !9 = !{i32 1, !"user_functions", i1 false}
-                !10 = !{i32 1, !"multiple_target_branching", i1 false}
```
billti, line 983: Did we used to just emit the attributes saying these were all supported?

compiler/qsc/src/target.rs
```diff
@@ -39,6 +47,7 @@ impl FromStr for Profile {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
```
billti, line 48: Maybe `s.to_lowercase().as_str()` and just keep the latter patterns for matching (then any mix of casing is handled).

compiler/qsc_codegen/src/qir.rs
```diff
@@ -136,6 +136,29 @@ impl ToQir<String> for rir::Operand {
...
+            rir::FcmpConditionCode::OrderedAndNotEqual => "one".to_string(),
+            rir::FcmpConditionCode::Ordered => "ord".to_string(),
+            rir::FcmpConditionCode::UnorderedOrEqual => "ueq".to_string(),
+            rir::FcmpConditionCode::UnorderedOrGreaterThan => "ugt".to_string(),
+            rir::FcmpConditionCode::UnorderedOrGreaterThanOrEqual => "uge".to_string(),
+            rir::FcmpConditionCode::UnorderedOrLessThan => "ult".to_string(),
+            rir::FcmpConditionCode::UnorderedOrLessThanOrEqual => "ule".to_string(),
+            rir::FcmpConditionCode::UnorderedOrNotEqual => "une".to_string(),
+            rir::FcmpConditionCode::Unordered => "uno".to_string(),
+            rir::FcmpConditionCode::True => "true".to_string(),
```
billti, line 157: Very minor nit: But why is `True` last? Seems it should be adjacent to `False`, or alphabetical (i.e. before the `Unordered`s).

compiler/qsc_codegen/src/qir.rs
```diff
@@ -314,6 +352,31 @@ fn call_to_qir(
...
+    rhs: &rir::Operand,
+    variable: rir::Variable,
+    program: &rir::Program,
+) -> String {
+    let lhs_ty = get_value_ty(lhs);
+    let rhs_ty = get_value_ty(rhs);
+    let var_ty = get_variable_ty(variable);
+    assert_eq!(
+        lhs_ty, rhs_ty,
+        "mismatched input types ({lhs_ty}, {rhs_ty}) for icmp {op}"
```
billti, line 367: Maybe I'm misreading this, but why does the message here (and line 370) refer to `icmp` if this is inside the `fcmp` method?

compiler/qsc_codegen/src/qir.rs
```diff
@@ -455,7 +546,7 @@ fn get_value_ty(lhs: &rir::Operand) -> &str {
         rir::Operand::Literal(lit) => match lit {
             rir::Literal::Integer(_) => "i64",
             rir::Literal::Bool(_) => "i1",
-            rir::Literal::Double(_) => "f64",
+            rir::Literal::Double(_) => get_f64_ty(),
```
billti, line 549: What's the value in breaking the pattern of the other types here that all use an inline literal and calling a helper, if that helper just always returns the literal "double"?

compiler/qsc/src/interpret/tests.rs
billti: (Never mind - I see the 'false' thing now. Probably better to just not list them, as you're now doing).

compiler/qsc_codegen/src/qir.rs
idavis: That is the order in which they are defined in the spec, so I kept it in that order.

compiler/qsc_codegen/src/qir.rs
idavis: I broke the pattern to provide documentation for others picking up the code so that they can hover and see why.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2977,6 +3199,39 @@ fn eval_bin_op_with_bool_literals(
...
+        panic!("at least one literal is not an double: {lhs_literal}, {rhs_literal}");
+    };
+
+    match bin_op {
+        BinOp::Eq => Ok(Value::Bool((lhs - rhs).abs() < f64::EPSILON)),
+        BinOp::Neq => Ok(Value::Bool((lhs - rhs).abs() > f64::EPSILON)),
+        BinOp::Gt => Ok(Value::Bool(lhs > rhs)),
+        BinOp::Gte => Ok(Value::Bool(lhs >= rhs)),
+        BinOp::Lt => Ok(Value::Bool(lhs < rhs)),
+        BinOp::Lte => Ok(Value::Bool(lhs <= rhs)),
```
billti, line 3226: Why does 'eq' and 'neq' allow for an 'epsilon' tolerance in comparisons, but other comparisons such a 'gte' don't? That would seem to lead to some unfortunate asymmetries (e.g. if 'a' is greater than 'b' to within epsilon, then "a == b" could evaluate to true, yet "a <= b" on the same values would evaluate to false).

pip/qsharp/_native.pyi
```diff
@@ -90,14 +90,22 @@ class TargetProfile(Enum):
...
     reset capabilities, as defined by the QIR specification.
     """
 
+    Adaptive_RIF: TargetProfile
+    """
+    Target supports the Adaptive profile with integer & floating point computation and qubit reset capabilities.
+
+    This profile includes all of the required Adaptive Profile and Adaptive_RI
+    capabilities, as well as the optional floating point computation and qubit
+    reset capabilities, as defined by the QIR specification.
```
billti, line 106: You already said it includes the 'Adaptive_RI' capabilities, which includes reset capabilities.

pip/src/interpreter.rs
```diff
@@ -106,6 +106,12 @@ pub(crate) enum TargetProfile {
     /// capabilities, as well as the optional integer computation and qubit
     /// reset capabilities, as defined by the QIR specification.
     Adaptive_RI,
+    /// Target supports the Adaptive profile with integer & floating point computation and qubit reset capabilities.
+    ///
+    /// This profile includes all of the required Adaptive Profile and Adaptive_RI
+    /// capabilities, as well as the optional floating point computation and qubit
+    /// reset capabilities, as defined by the QIR specification.
```
billti, line 113: As prior comment, including `Adaptive_RI` already includes reset capabilities.

pip/tests-integration/resources/output/Adaptive_RI/ArithmeticOps.ll
```diff
@@ -104,16 +104,10 @@ attributes #1 = { "irreversible" }
...
 !2 = !{i32 1, !"dynamic_qubit_management", i1 false}
 !3 = !{i32 1, !"dynamic_result_management", i1 false}
-!4 = !{i32 1, !"classical_ints", i1 true}
-!5 = !{i32 1, !"qubit_resetting", i1 true}
-!6 = !{i32 1, !"classical_floats", i1 false}
-!7 = !{i32 1, !"backwards_branching", i1 false}
-!8 = !{i32 1, !"classical_fixed_points", i1 false}
-!9 = !{i32 1, !"user_functions", i1 false}
-!10 = !{i32 1, !"multiple_target_branching", i1 false}
+!4 = !{i32 1, !"int_computations", !"i64"}
```
billti, line 113: Will this change in module flags metadata name from `classical_ints` to `int_computations` be a breaking change anywhere in our existing flow (including any service validation)? Hopefully Quantinuum can already handle either, as they helped write the spec, but maybe worth a check before changing.

pip/tests-integration/resources/output/Adaptive_RI/WithinApply.ll
```diff
@@ -38,16 +38,10 @@ attributes #1 = { "irreversible" }
...
 !2 = !{i32 1, !"dynamic_qubit_management", i1 false}
 !3 = !{i32 1, !"dynamic_result_management", i1 false}
-!4 = !{i32 1, !"classical_ints", i1 true}
-!5 = !{i32 1, !"qubit_resetting", i1 true}
-!6 = !{i32 1, !"classical_floats", i1 false}
-!7 = !{i32 1, !"backwards_branching", i1 false}
-!8 = !{i32 1, !"classical_fixed_points", i1 false}
-!9 = !{i32 1, !"user_functions", i1 false}
-!10 = !{i32 1, !"multiple_target_branching", i1 false}
+!4 = !{i32 1, !"int_computations", !"i64"}
```
billti, line 47: Should we have a test case in this folder that includes an `.ll` file using the new capability? (Which maybe means a new sample)

pip/tests-integration/resources/output/Adaptive_RI/ArithmeticOps.ll
idavis: To my knowledge this won't affect anything and PyQIR has no idea whether either exist as the adaptive profile work hasn't been started. @swernli @cesarzc care to weigh in?

pip/qsharp/_native.pyi
idavis: Thanks, will review.

pip/src/interpreter.rs
idavis: Reworded them. The adaptive profile requires qubit reset now, so I updated the adaptive RI docs as well.

npm/qsharp/src/compiler/compiler.ts
minestarks, line 1: I don't think we need a default profile in this file. The default of `adaptive_ri` is a bit distracting - it got me wondering why use such an odd default until I discovered it's dead code anyway. 

`getAst` and `getHir` are only used internally in our Q# playground, where `profile` is already a required parameter in the caller. `getHir` is also used in a unit test in `basics.js` but that can easily be updated. `ICompiler.getAst` and `ICompiler.getHir` should be updated to make `profile` a required parameter. `profile ?? get_default_profile()` should be removed and `profile` should be passed directly instead. This won't be a breaking change for any existing users.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1905,6 +1975,129 @@ impl<'a> PartialEvaluator<'a> {
...
+        bin_op_expr_span: PackageSpan, // For diagnostic purposes only.
+    ) -> Result<rir::Variable, Error> {
+        let rir_variable = match bin_op {
+            BinOp::Add => {
+                let bin_op_variable_id = self.resource_manager.next_var();
+                let bin_op_rir_variable = rir::Variable::new_double(bin_op_variable_id);
+                let bin_op_rir_ins =
+                    Instruction::Fadd(lhs_operand, rhs_operand, bin_op_rir_variable);
+                self.get_current_rir_block_mut().0.push(bin_op_rir_ins);
+                bin_op_rir_variable
```
minestarks, line 1993: Agree with clippy that this function is too long. A low hanging fruit is that 3/5 lines in all of these arms (except division) are identical and can live outside the `match`. (For division, I don't think it matters if you do the divide-by-zero check after you've already allocated the next variable, so I think that could stay inside the `match`?)

minestarks (approved):

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2977,6 +3150,39 @@ fn eval_bin_op_with_bool_literals(
...
+    let (Literal::Double(lhs), Literal::Double(rhs)) = (lhs_literal, rhs_literal) else {
+        panic!("at least one literal is not an double: {lhs_literal}, {rhs_literal}");
+    };
+
+    match bin_op {
+        BinOp::Eq => Ok(Value::Bool(f64_approx_eq(lhs, rhs))),
+        BinOp::Neq => Ok(Value::Bool(f64_approx_neq(lhs, rhs))),
+        BinOp::Gt => Ok(Value::Bool(lhs > rhs)),
+        BinOp::Gte => Ok(Value::Bool(f64_approx_gte(lhs, rhs))),
+        BinOp::Lt => Ok(Value::Bool(lhs < rhs)),
```
billti, line 3176: It occurred to me last night that this is still problematic. For example, if rhs is greater than lhs but within epsilon, then both "rhs > lhs" and "rhs <= lhs" will be true when they should be mutually exclusive (indeed, "<=" if often defined in terms of "not >").

I have some thoughts on what a 'correct' impl would look like, but do you know if there is any prior art we should follow here in other languages re handling floats to within an epsilon?

pip/tests-integration/resources/adaptive_rif/input/Doubles.qs
```diff
@@ -0,0 +1,28 @@
...
+        mutable count = 0.0;
+        use q = Qubit();
+        for _ in 1..10 {
+            X(q);
+            if M(q) == One {
+                X(q);
+                set count += 1.0;
+                // These are all valid operations on doubles.
+                // they will be uncommented and tested in the future.
+                // once we fix the issue with the QIR generation.
```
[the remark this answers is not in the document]
idavis: I am waiting for #2087 before I merge so this will be fleshed out before merge.

pip/tests-integration/resources/output/Adaptive_RI/ArithmeticOps.ll
swernli: Yup, you're right, this won't affect any consumers as no one uses these flags yet. Now is the time to make the fix so they are consistent with the spec.

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -140,26 +140,18 @@ export class Compiler implements ICompiler {
 
   async getAst(
     code: string,
-    languageFeatures?: string[],
-    profile?: TargetProfile,
+    languageFeatures: string[],
+    profile: TargetProfile,
```
swernli, line 144: Does making these non-optional have an affect on QCOM? I was under the impression that we did this to avoid breaking changes to the API surface here, but maybe I'm remembering that incorrectly (@minestarks, do you remember?).

compiler/qsc_partial_eval/src/lib.rs
idavis: It is all removed.

compiler/qsc_rir/src/rir.rs
```diff
@@ -238,14 +251,45 @@ impl Display for ConditionCode {
...
+    Ordered,
+    UnorderedOrEqual,
+    UnorderedOrGreaterThan,
+    UnorderedOrGreaterThanOrEqual,
+    UnorderedOrLessThan,
+    UnorderedOrLessThanOrEqual,
+    UnorderedOrNotEqual,
+    Unordered,
+    True,
+}
```
swernli, line 272: nitpick: looks like we only ever use the ordered variants here, are the others just included for completeness? Just wondering if it's worth scoping down to just the ones we produce from codegen. Not really important, there's no harm in having them here, mostly curious as to the motivation.

swernli (approved):
Looks good, signing off with a few minor comments.

compiler/qsc_rir/src/rir.rs
idavis: Completeness is the reason. It didn't feel right to map half of an LLVM enum.

npm/qsharp/src/compiler/compiler.ts
minestarks: This is ok. This would have been breaking if `getAst` or `getHir` were used in QCOM, but they're not. I like this version better since it avoids adding a meaningless default.

vscode/src/qirGeneration.ts
```diff
@@ -22,49 +22,52 @@ export class QirGenerationError extends Error {
...
+  const isLocalQirGeneration = targetSupportsAdaptive === undefined;
   const config = program.programConfig;
   const targetProfile = config.profile;
-  const is_unrestricted = targetProfile === "unrestricted";
-  const is_unsupported_adaptive = targetProfile === "adaptive_rif";
-  const is_base = targetProfile === "base";
+  const isUnrestricted = targetProfile === "unrestricted";
+  const isUnsupportedAdaptiveSubmissionProfile =
+    targetProfile === "adaptive_rif";
+  const targetProfileIsBase = targetProfile === "base";
```
swernli, line 38: nitpick: this variable is the one boolean that doesn't start with "is" now.

swernli: actually, I guess that's not true... `targetSupportsAdaptive` also doesn't start with "is" so maybe this doesn't matter.

vscode/src/qirGeneration.ts
```diff
@@ -22,44 +22,52 @@ export class QirGenerationError extends Error {
...
     error_msg +=
       "using the Adaptive_RI profile is not supported for targets that can only accept Base profile QIR.";
   }
 
   // Check that the current target is base or adaptive_ri profile, and current doc has no errors.
-  if (is_unrestricted || (!is_base && supports_adaptive === false)) {
+  if (
+    isUnrestricted ||
+    isSubmittingAdaptiveToBaseAzureTarget ||
+    isUnsupportedAdaptiveSubmissionProfile
```
swernli, line 59: this should blocks generating QIR for adaptive RIF even outside of submission, which I don't think was the goal. You want something with one more part of the condition:
```suggestion
    (isUnsupportedAdaptiveSubmissionProfile && !isLocalQirGeneration)
```

vscode/src/qirGeneration.ts
```diff
@@ -22,44 +22,52 @@ export class QirGenerationError extends Error {
...
-      ? "Generating QIR "
-      : "Submitting to Azure ";
-  if (is_unrestricted) {
+  let error_msg = isLocalQirGeneration
+    ? "Generating QIR "
+    : "Submitting to Azure ";
+  if (isUnrestricted) {
     error_msg += "is not supported when using the unrestricted profile.";
-  } else if (!is_base && supports_adaptive === false) {
+  } else if (isSubmittingAdaptiveToBaseAzureTarget) {
```
swernli, line 50: I think you need another case here to add to the error message for when the user is trying to generate Adaptive RIF during QIR submission. Otherwise the box is half empty:
image

PR #2079, opened 2024-12-27
"Import External Q# Package" VS Code command palette command

sezna:
This PR formalizes the "registry" md document into a JSON file, and then adds a command palette command which consumes it. This will hopefully increase discoverability of the project system, and formalizes our listing of known libraries a bit more.

https://github.com/user-attachments/assets/20f8dd7b-5b53-4b03-8990-aef41d6b3da7


vscode/src/createProject.ts
```diff
@@ -265,93 +254,175 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+    projectFiles.forEach((file) => {
+      const dirName = file.path.slice(0, -"/qsharp.json".length);
+      const relPath = getRelativeDirPath(qsharpJsonDir.path, dirName);
+      projectChoices.push({
+        name: dirName.slice(dirName.lastIndexOf("/") + 1),
+        ref: {
+          path: relPath,
+        },
+      });
+    });
```
minestarks, line 307: Would look more natural to write this as a `.map()` in my opinion.

vscode/src/createProject.ts
```diff
@@ -265,93 +254,175 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+
+    // Apply the edits to the qsharp.json
+    const edit = new vscode.WorkspaceEdit();
+    edit.replace(
+      qsharpJsonUri,
+      new vscode.Range(0, 0, qsharpJsonDoc.lineCount, 0),
+      JSON.stringify(manifestObj, null, 2),
+    );
+    if (!(await vscode.workspace.applyEdit(edit))) {
+      vscode.window.showErrorMessage(
```
minestarks, line 419:
```suggestion
      await vscode.window.showErrorMessage(
```

vscode/src/createProject.ts
```diff
@@ -265,93 +254,175 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+      await vscode.workspace.findFiles("**/qsharp.json")
+    ).filter((file) => file.toString() !== qsharpJsonUri.toString());
+
+    const projectChoices: Array<{ name: string; ref: LocalProjectRef }> = [];
+
+    projectFiles.forEach((file) => {
+      const dirName = file.path.slice(0, -"/qsharp.json".length);
+      const relPath = getRelativeDirPath(qsharpJsonDir.path, dirName);
+      projectChoices.push({
+        name: dirName.slice(dirName.lastIndexOf("/") + 1),
```
minestarks, line 302: I can't find any bugs from just reading the code, but the `/` separator has me wondering if this has been tested on Windows (desktop and web)? LMK if you want me to give it a spin on my laptop.

-- in my experience, Windows browser, Windows desktop, and linux/mac browser/desktop all have unique path handling concerns. And even just in the browser: the playground, the test FS, and the local FS all show up as different virtual filesystems with different quirks. So I've become very distrustful of filesystem code... forgive me.

sezna: FWIW, this is previously-existing code from Bill's implementation, I just moved it elsewhere. It's been live for a while now, and I believe all paths in the `vscode` APIs are normalized  -- I can't find the exact docs but this hints at it https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L1466

billti: I think Mine is right. In doing some testing on my Windows ARM VM I noticed the relative paths it generated go all the way to the root of the drive then back down, and this spot may well be the cause. Good catch!

vscode/src/registry.json
```diff
@@ -0,0 +1,64 @@
+{
+  "knownPackages": [
+    {
+      "name": "Signed",
+      "description": "Defines types and functions to work with signed qubit-based integers.",
+      "dependency": {
+        "github": {
+          "owner": "Microsoft",
```
billti, line 8: Being these are used to build GitHub URLs, "Microsoft" should probably be all lowercase.

vscode/src/createProject.ts
```diff
@@ -265,93 +254,175 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
-        // Convert any spaces, dashes, dots, tildes, or quotes in project names
-        // to underscores. (Leave more 'exotic' non-identifier patterns to the user to fix)
-        //
-        // Note: At some point we may want to detect/avoid duplicate names, e.g. if the user already
-        // references a project via 'foo', and they add a reference to a 'foo' on GitHub or in another dir.
-        projectChoices.forEach(
-          (val, idx, arr) =>
-            (arr[idx].name = val.name.replace(/[- "'.~]/g, "_")),
+        // ask the user if they want to import from github, or locally on their computer
+        const importChoice = await vscode.window.showQuickPick(
```
billti, line 258: I'm not sure we need to bifurate based on project location here and add an extra UI step/click for the user. In the vast majority of cases I expect users will have zero to maybe 1 or 2 local projects in their workspace, so the current behavior of listing local projects first followed by GitHub to choose from seems fine. (The 'local directory' wording could also be misleading, as we don't allow you to choose any local directory, only those projects that are already open in the current workspace).

sezna: Alright, the only reason I did it this way is because they are different flows -- the github flow asks one more question than the local project flow (the `ref`), so conflating them made it a bit messy. But I'll merge them and keep it as clean as possible

vscode/src/createProject.ts
sezna: again, this is just relocated code, but i've rewritten it as a map

vscode/src/createProject.ts
sezna: I've updated this to use the vscode URI API which normalizes paths, however I don't have a readily available windows machine to test. I've read the [source](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/uri.ts) for this URI implementation, and I believe it does do the normalization we need.

minestarks (approved):

vscode/src/createProject.ts
```diff
@@ -265,93 +255,168 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
...
+  ) {
+    if (!manifestObj["dependencies"]) manifestObj["dependencies"] = {};
+    manifestObj["dependencies"][label] = ref;
+
+    // Apply the edits to the qsharp.json
+    const edit = new vscode.WorkspaceEdit();
+    edit.replace(
+      qsharpJsonUri,
+      new vscode.Range(0, 0, qsharpJsonDoc.lineCount, 0),
+      JSON.stringify(manifestObj, null, 2),
```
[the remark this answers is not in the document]
billti: I thought VS Code used 2 and prettier has a default of 2 (https://prettier.io/docs/en/options#tab-width). I could be wrong.

vscode/src/registry.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    {
+      "name": "FixedPoint",
+      "description": "Types and functions for fixed-point arithmetic with qubits.",
+      "dependency": {
+        "github": {
+          "owner": "microsoft",
+          "repo": "qsharp",
+          "refs": [
+            { "ref": "546b053", "notes": "latest stable" },
```
billti, line 26: We should bump this after #2121 going in, as the library doesn't work without it.

vscode/src/createProject.ts
billti: I though you said you addressed this? I still see the quick pick to choose between GitHub or local project. Or was there a misunderstanding?

billti: Actually. Don't worry about it. It's fine for now.

vscode/src/createProject.ts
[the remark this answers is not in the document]
billti: ChatGPT also says "If there’s no existing standard in your project or organization, 2 spaces is a widely accepted and practical choice for JSON formatting" :-) . I think we can leave this as-is.

billti (approved):

vscode/src/createProject.ts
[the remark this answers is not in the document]
minestarks: I mean I don't mind 2, but VS Code's default is definitely 4 and so that's what almost everyone will see 🤷‍♀️

vscode/src/registry.json
```diff
@@ -0,0 +1,64 @@
...
+  "knownPackages": [
+    {
+      "name": "Signed",
+      "description": "Defines types and functions to work with signed qubit-based integers.",
+      "dependency": {
+        "github": {
+          "owner": "microsoft",
+          "repo": "qsharp",
+          "refs": [
+            { "ref": "546b053", "notes": "latest stable" },
```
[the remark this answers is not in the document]
billti: Let's discuss as a separate issue. Both have their pros and cons, but happy to just get this shipped for now, and I don't think it's a blocker.

PR #2082, opened 2024-12-29
Fix `Relabel` for odd size arrays

### LINKED ISSUE #2077: Relabel applies wrong permutation for >=5 qubits
**Describe the bug**

[Relabel](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/std.canon/relabel) operation doesn't work correctly.

**To Reproduce**

Run the following code:
```
operation Main(): Unit {
    use q = Qubit[5];  // little-endian 5-bit integer.
    X(q[1]);  // q:=2.
    Relabel(q, q[2..4]+q[0..1]);  // Cyclic shift right by 3, q:=q<<3, must get q=16.
    Message($"ans={MeasureInteger(q)}");  // prints 4, expected 16.
}
```

**Expected behavior**

State of qubits before relabel is `[0,1,0,0,0]`. We would expect that after relabel state is `[0,0,0,0,1]`. In particular, any use of `q[4]` should refer to what was `q[1]`. However, the state we get after relabel is `[0,0,1,0,0]`.

**Screenshots**

N/A

**System information**

* qsharp version 1.11
* QDK version 1.12
* Note: I can also reproduce this in https://microsoft.github.io/qsharp/

**Additional context**

* I ran into this issue while trying to use Relabel to implement cyclic shift of an integer [like this](https://github.com/fedimser/quant-arith-re/blob/bc6921b762b86cbcf5b2f07539192dad3115a7ac/src/QuantumArithmetic/NZLS2023.qs#L30-L36). When I ran tests for it, they failed for some inputs for n>=5 qubits.
* I found a workaround - write my own code to decompose permutation into swaps (aka transpositions), and then use Relabel to apply each swap individually ([like this](https://github.com/fedimser/quant-arith-re/commit/bc23ad4f6a94ab3e2fd4fb776fb361f9129a7ec1)).

swernli:
When relabeling qubits arrays with odd lengths 5 or greater, the final case in the match statement triggered with a bug that caused the mappings tracked for swaps to be updated incorrectly, resulting in the wrong qubit labels being swapped. This fixes the bug, updates some variable names and comments for clarity, and adds new test cases to verify the expected behavior. It also fixes a minor test bug in a related test that was verifying the wrong state.

Fixes #2077

### DIFF
--- a/compiler/qsc_eval/src/intrinsic.rs
+++ b/compiler/qsc_eval/src/intrinsic.rs
@@ -392,14 +392,14 @@ pub fn qubit_relabel(
             }
             (false, true) => {
                 // The right qubit has been relabeled, so we need to swap the left qubit with the
-                // qubit that the right qubit was relabeled to.
-                let mapped = *map
+                // new label for the right qubit.
+                let label = *map
                     .keys()
                     .find(|k| map[*k] == r)
                     .expect("mapped qubit should be present as both key and value");
-                swap(l, mapped);
+                swap(l, label);
                 map.insert(l, r);
-                map.insert(mapped, l);
+                map.insert(label, l);
             }
             (true, false) => {
                 // The left qubit has been relabeled, so we swap the qubits as normal but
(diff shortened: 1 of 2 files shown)


sezna (approved):

DmitryVasilevsky:
Please consider an alternative version.
https://github.com/microsoft/qsharp/pull/2083
I believe it's easier to understand.

compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -392,14 +392,14 @@ pub fn qubit_relabel(
             }
             (false, true) => {
                 // The right qubit has been relabeled, so we need to swap the left qubit with the
-                // qubit that the right qubit was relabeled to.
```
DmitryVasilevsky, line 395: Please consider an alternative version.
https://github.com/microsoft/qsharp/pull/2083
I believe it's easier to understand.

DmitryVasilevsky: I'll actually do a separate PR. This PR should be good as is.

DmitryVasilevsky (approved):

PR #2085, opened 2024-12-31
Simplify `Relabel` logic

swernli:
This simplifies the relabeling logic to collapse the four match cases down into one for easier readibility. It avoids trying to optimize special cases for qubits that may not have been mapped and treats all labels as having an intial mapping to themselves.

Based on feedback from @DmitryVasilevsky in https://github.com/microsoft/qsharp/pull/2082#discussion_r1899442453

### DIFF
--- a/compiler/qsc_eval/src/intrinsic.rs
+++ b/compiler/qsc_eval/src/intrinsic.rs
@@ -377,56 +377,30 @@ pub fn qubit_relabel(
         return Err(Error::RelabelingMismatch(arg_span));
     }
 
-    let mut map = FxHashMap::default();
-    map.reserve(left.len());
+    // Start with a mapping of each qubit to itself.
+    let mut mappings: FxHashMap<usize, usize> =
+        left.iter().copied().zip(left.iter().copied()).collect();
     for (l, r) in left.into_iter().zip(right.into_iter()) {
+        // Trivial case where the qubit is already mapped to itself in the relabel, which can be short circuited.
         if l == r {
             continue;
         }
-        match (map.contains_key(&l), map.contains_key(&r)) {
-            (false, false) => {
-                // Neither qubit has been relabeled yet.
-                swap(l, r);
-                map.insert(l, r);
-                map.insert(r, l);
-            }
-            (false, true) => {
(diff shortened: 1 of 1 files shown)


compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -377,56 +377,30 @@ pub fn qubit_relabel(
...
+        let mapped_l = *mappings.get(&l).expect("mapped qubit should be present");
+        let mapped_r = *mappings.get(&r).expect("mapped qubit should be present");
+
+        // We only need to swap if the label is not pointing to the correct qubit.
+        if mapped_l != r && mapped_r != l {
+            // Do a reverse lookup to find which label is currently mapped to the desired right qubit.
+            // This tells us which label to use in the swap, which we will use in the update of the mappings too.
+            let label_r = *mappings
+                .keys()
+                .find(|k| mappings[*k] == r)
```
DmitryVasilevsky, line 399: This step is linear, which makes the whole thing quadratic. Granted, our qubit count won't be large, but let's figure out linear algorithm (in average case). Should be doable with mutable array and mutable hash map.

swernli: The previous implementation avoided doing this reverse lookup every time by only performing it when necessary. You are right, I wouldn't expect `Relabel` to be called with very large lists of qubits, and even if it were I don't think this would take lots of time unless relabels were happening very frequently.

If we do want to optimize this further, we could maintain two maps, one from label to actual and one from actual to label. Or, if we are comfortable with some higher memory costs, we could use an `IndexMap` for fast lookup (but then any gaps in qubit ids would be empty entries in the underlying vector). Overall, I'm don't really think it's worth optimizing this further until we have reason to believe that large lists of qubits will be relabeled frequently, and then we can perf test to find the right optimizations.

DmitryVasilevsky: I think we should eventually move this to simulator. And we could improve it at that time. For now, I suggest we add a comment that this step may take O(n) time, but we are only expecting small number of qubits.

DmitryVasilevsky (approved):

idavis (approved):

PR #2086, opened 2025-01-02
Fix `Exp` on qubit arrays larger than 2 with single `PauliI`

swernli:
This fixes a bug in the decomposition for `Exp` that would cause a runtime failure if an array of more than two qubits was used and a single `PauliI` was passed in the Pauli array. The decomposition ignores `PauliI` by calling the `RemovePauliI` helper function, but then incorrectly uses the original arrays when rather than the filtered arrays in the rest of the operation. This adds a test to confirm the fix (without the fix the test triggers a runtime failure in `MapPauli`).

### DIFF
--- a/library/src/tests/intrinsic.rs
+++ b/library/src/tests/intrinsic.rs
@@ -3010,6 +3010,39 @@ fn test_exp() {
 }
 
 #[test]
+fn test_exp_mixed_paulis() {
+    let dump = test_expression(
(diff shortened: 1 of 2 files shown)


DmitryVasilevsky (approved):

DmitryVasilevsky:
Wow! Thanks!

minestarks (approved):

Issue #1950, opened 2024-10-04
Syntax highlighting could use improvement


minestarks:
The TextMate grammar used for Q# syntax highlighting in VS Code is missing some patterns that would help the syntax highlighting look nicer.

Compare:

![image](https://github.com/user-attachments/assets/00922582-e758-4783-9ba8-48484f617f3f)

The Q# looks pretty boring in comparison to other languages with good support.

I think we can start by defining patterns for function calls and numeric literals, which are seen in these screenshots. There may be others that are interesting as well.

Morcifer:
Hi @minestarks, @sezna, can I pick this one up?

I managed to get the function definitions working, at least:
![Image](https://github.com/user-attachments/assets/a24c4266-d09f-4393-a91e-66f8889218aa)

Getting function calls to work seems to interfere with strings, though, so this fix will likely have small changes to existing patterns as well...

PR #2088, opened 2025-01-04
Syntax highlighting for functions, variables and numbers

Morcifer:
Supports syntax highlighting for:

- function definitions
- function calls
- numbers
- variables

See issue https://github.com/microsoft/qsharp/issues/1950.

The change for function calls also applies to operation definitions and to attributes (see bigger example below).
The change for variables also applies to attributes and to imports.

I might be nice to make attributes and imports get their own designated rules. I can do them as part of this PR, but I think it might be best to do as a followup issue - because I believe this PR is actually already an improvement for them.

Simple example for the change (from the original issue description):

image

And a bigger example:

image

### DIFF
--- a/vscode/syntaxes/qsharp.tmLanguage.json
+++ b/vscode/syntaxes/qsharp.tmLanguage.json
@@ -94,7 +94,7 @@
         {
           "comment": "function calls",
           "name": "meta.function.call.qsharp",
-          "begin": "([A-Za-z0-9_]+)(\\()",
+          "begin": "\\b(?<!@)([A-Za-z0-9_]+)(\\()",
           "beginCaptures": {
             "1": {
               "name": "entity.name.function.qsharp"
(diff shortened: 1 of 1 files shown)


Morcifer:

swernli:
Thanks for the contribution, @Morcifer! One idea I had after looking at your screenshots: would it be possible to use the presence of the `@` symbol to avoid highlighting attributes like `@Config()` and `@EntryPoint()` as if they were function calls?

Morcifer:
> would it be possible to use the presence of the `@` symbol to avoid highlighting attributes like `@Config()` and `@EntryPoint()` as if they were function calls?

It's definitely possible, though it still treats the attributes inside the parentheses of the Config as variables... It would be simpler for me to just make a Config-specific rule than to figure out how to disable that one, I think...

image

swernli:
I think the items inside the parens of `@Config` look fine in the screenshot you posted. This is indeed a nice improvement!

minestarks (approved):
Great improvement, thank you!

swernli (approved):

PR #2091, opened 2025-01-06
Support further Python Interop for callables in `.estimate()`, `.compile()`, `.run()` and `.circuit()`

swernli:
This follows up on #2054 by adding the enhanced support for Q# callables exposed into Python and Python argument arrays in resource estimation, QIR compilation, running multiple shots, and circuit generation.


compiler/qsc_codegen/src/qir.rs
```diff
@@ -8,58 +8,48 @@ mod instruction_tests;
...
-}
-
 pub fn fir_to_qir(
     fir_store: &qsc_fir::fir::PackageStore,
     capabilities: TargetCapabilityFlags,
     compute_properties: Option<PackageStoreComputeProperties>,
     entry: &ProgramEntry,
 ) -> Result<String, qsc_partial_eval::Error> {
-    let mut program = get_rir_from_compilation(fir_store, compute_properties, entry, capabilities)?;
+    let compute_properties = compute_properties;
```
swernli, line 26: Whoops, this looks like it got left behind from an earlier experiment.

pip/src/interpreter.rs
minestarks, line 1: What about the third sibling, `.run()` ? I think it would make sense to allow that to take a callable, as well.

And of course, there's `.circuit()`, which can already take a callable, but uses a different pattern... which is unfortunate, because this new pattern is better. I wouldn't be opposed to updating `.circuit()` to allow the new pattern as well, but I understand if you don't want to touch it in this PR.

compiler/qsc/src/interpret.rs
```diff
@@ -639,6 +639,43 @@ impl Interpreter {
         })
     }
 
+    pub fn qirgen_from_callable(
```
idavis, line 642: Please add a doc comment

pip/src/interpreter.rs
swernli: I think it's worth updating both of those. I'll get that added and see how it looks.

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -178,7 +194,44 @@ impl<'a> PartialEvaluator<'a> {
...
+    }
+
+    fn new_from_package_id(
+        package_store: &'a PackageStore,
+        compute_properties: &'a PackageStoreComputeProperties,
+        package_id: PackageId,
+        capabilities: TargetCapabilityFlags,
+    ) -> Self {
+        // Create the entry-point callable.
+        let mut resource_manager = ResourceManager::default();
```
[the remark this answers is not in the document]
swernli: Done!

pip/src/interpreter.rs
swernli: I've added the support for `.run()` and `.circuit()`, let me know what you think!

minestarks (approved):
Matches what we already discussed, and all the changes seem pretty mechanical and unite tested... so I don't have a lot to pick on 😄

pip/qsharp/_native.pyi
```diff
@@ -218,16 +231,28 @@ class Interpreter:
...
+
         :raises QSharpError: If there is an error synthesizing the circuit.
         """
         ...
 
-    def estimate(self, entry_expr: str, params: str) -> str:
+    def estimate(
+        self,
+        params: str,
+        entry_expr: Optional[str],
```
billti, line 245: So maybe entry_expr should have been optional above (prior comment)? Do you provide one or the other? What happens if you provide both?

I think better help docs on these APIs that now take both an entry expression and a callable would be helpful.

pip/qsharp/_qsharp.py
```diff
@@ -441,16 +441,16 @@ def callback(output: Output) -> None:
...
-    _callable.__qs_gen = True
+    _callable.__global_callable = callable
 
     # Add the callable to the module.
     module.__setattr__(callable_name, _callable)
 
 
 def run(
-    entry_expr: str,
+    entry_expr: Union[str, Callable],
```
billti, line 451: Maybe this makes more sense. Or is it the 'back compat' concern which is causing us to just expand params on the other APIs rather than union them?

pip/qsharp/_native.pyi
```diff
@@ -170,6 +172,8 @@ class Interpreter:
         :param output_fn: A callback function that will be called with each output.
         :param noise: A tuple with probabilities of Pauli-X, Pauli-Y, and Pauli-Z errors
             to use in simulation as a parametric Pauli noise.
+        :param callable: The callable to run.
```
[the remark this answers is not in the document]
swernli: These APIs in __native.pyi are just the interfaces for the Rust code and are not part of the public API surface for our qsharp module. They just need to be updated to reflect what the Rust code uses.

Per your other comment, you are right, both `entry_expr` and `output_fn` are `Option` types in Rust but missing their `Optional` type annotation here. I'll fix that!

pip/qsharp/_native.pyi
swernli: These help docs aren't part of our public API surface. I'll update them anyway, but this is just for our benefit, not anything for users to see.

pip/qsharp/_qsharp.py
swernli: This is where I'm using more Pythonic patterns to expand what types the user-facing API can accept without breaking existing code or extending the number of parameters. I can't do that in __native.pyi the same way because those are interop wrappers around the native Rust code, so if I wanted to have the equivalent of a `Union` I'd need to introduce a Rust struct and then reflect that via pyO3 as a class into the Python wrapper, which seemed like overkill for what is essentially an internal API.

compiler/qsc/src/interpret.rs
```diff
@@ -639,6 +672,45 @@ impl Interpreter {
...
 
+    /// Performs QIR codegen using the given callable with the given arguments on a new instance of the environment
+    /// and simulator but using the current compilation.
+    pub fn qirgen_from_callable(
+        &mut self,
+        callable: &Value,
+        args: Value,
+    ) -> std::result::Result<String, Vec<Error>> {
+        if self.capabilities == TargetCapabilityFlags::all() {
+            return Err(vec![Error::UnsupportedRuntimeCapabilities]);
```
sezna, line 683: Does QIR gen only work for Base profile still? If so, is checking for all target capabilities correct, or should we check for `!= Base`?

compiler/qsc/src/interpret.rs
```diff
@@ -639,6 +672,45 @@ impl Interpreter {
...
+        }
+
+        let Value::Global(store_item_id, _) = callable else {
+            panic!("value is not a global callable");
```
sezna, line 687: we return a `Result` here, so this function is fallible -- any reason to use `panic!()` over an error?

compiler/qsc/src/interpret.rs
```diff
@@ -639,6 +672,45 @@ impl Interpreter {
...
+        .map_err(|e| {
+            let hir_package_id = match e.span() {
+                Some(span) => span.package,
+                None => map_fir_package_to_hir(self.package),
+            };
+            let source_package = self
+                .compiler
+                .package_store()
+                .get(hir_package_id)
+                .expect("package should exist in the package store");
```
sezna, line 706: Pretty sure I've done this in a bunch of places too, but perhaps these panics could be wrapped up in a new `Error::Internal(String)` or `Error::InvariantFailed(String)` for more robust code?

compiler/qsc/src/interpret.rs
```diff
@@ -656,29 +728,46 @@ impl Interpreter {
...
+            CircuitEntryPoint::EntryPoint => (None, None),
         };
 
         let circuit = if simulate {
             let mut sim = sim_circuit_backend();
 
-            self.run_with_sim_no_output(entry_expr, &mut sim)?;
+            match invoke_params {
+                Some((callable, args)) => {
+                    let mut sink = std::io::sink();
```
sezna, line 748: how does this `std::io` sink work here when running in, e.g., wasm? I'm sure it works, I'm just used to us running all output through our logging abstractions

compiler/qsc_codegen/src/qir.rs
```diff
@@ -60,6 +61,24 @@ pub fn fir_to_qir(
     Ok(ToQir::<String>::to_qir(&program, &program))
 }
```
sezna, line 63: this is a documentation-light file, but perhaps it'd be good to document at least newly-added `pub` fns

compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -70,6 +70,22 @@ pub fn partially_evaluate(
     partial_evaluator.eval()
 }
```
sezna, line 72: ditto re: new `pub fn`s without docs

compiler/qsc/src/interpret.rs
swernli: No, this case is correct because we only reject if all capabilities are asked for. We can support Base and two flavors of Adaptive now.

compiler/qsc/src/interpret.rs
swernli: I would say for this one it's ok, because getting all the way here and then not finding something in a package store is a terminal failure, not really something we'd want to recover from.

compiler/qsc/src/interpret.rs
swernli: I think it just pipes stuff to null in an efficient way... or as [the docs](https://doc.rust-lang.org/std/io/struct.Sink.html) put it:

> A writer which will move data into the void.

We already used this pattern for a couple of other places, notably the resource estimation path, so I think it's considered safe.

compiler/qsc/src/interpret.rs
swernli: Good question... I was thinking about how this is called by the Python layer specifically, but this could do a more normal failure. This one doesn't really need to be a panic.

billti (approved):

PR #2093, opened 2025-01-07
Multiple Webview Tabs

ScottCarda-MS:
Instead of always consuming the last Q# webview tab, open a new webview tab when activated with a new operation. Activation with the same operation will still consume/update the last webview for that operation. Works on circuit and histogram panels.

### DIFF
--- a/vscode/src/circuit.ts
+++ b/vscode/src/circuit.ts
@@ -332,6 +332,7 @@ export function updateCircuitPanel(
     calculating?: boolean;
   },
 ) {
+  const panelId = params?.operation?.operation || projectName;
   const title = params?.operation
     ? `${params.operation.operation} with ${params.operation.totalNumQubits} input qubits`
     : projectName;
@@ -352,7 +353,7 @@ export function updateCircuitPanel(
     command: "circuit",
     props,
   };
-  sendMessageToPanel("circuit", reveal, message);
+  sendMessageToPanel({ panelType: "circuit", id: panelId }, reveal, message);
 }
 
 /**
--- a/vscode/src/circuit.ts
+++ b/vscode/src/circuit.ts
@@ -350,7 +350,6 @@ export function updateCircuitPanel(
   };
 
   const message = {
-    command: "circuit",
     props,
   };
   sendMessageToPanel({ panelType: "circuit", id: panelId }, reveal, message);
--- a/vscode/src/documentation.ts
+++ b/vscode/src/documentation.ts
@@ -14,7 +14,7 @@ export async function showDocumentationCommand(extensionUri: Uri) {
 
   // Reveal panel and show 'Loading...' for immediate feedback.
   sendMessageToPanel(
-    { panelType: "documentation", id: "" }, // This is needed to route the message to the proper panel
+    { panelType: "documentation" }, // This is needed to route the message to the proper panel
     true, // Reveal panel
     null, // With no message
   );
@@ -48,7 +48,7 @@ export async function showDocumentationCommand(extensionUri: Uri) {
   };
 
   sendMessageToPanel(
-    { panelType: "documentation", id: "" }, // This is needed to route the message to the proper panel
+    { panelType: "documentation" }, // This is needed to route the message to the proper panel
     true, // Reveal panel
     message, // And ask it to display documentation
   );
--- a/vscode/src/documentation.ts
+++ b/vscode/src/documentation.ts
@@ -14,7 +14,7 @@ export async function showDocumentationCommand(extensionUri: Uri) {
 
   // Reveal panel and show 'Loading...' for immediate feedback.
   sendMessageToPanel(
-    "documentation", // This is needed to route the message to the proper panel
+    { panelType: "documentation", id: "" }, // This is needed to route the message to the proper panel
     true, // Reveal panel
     null, // With no message
   );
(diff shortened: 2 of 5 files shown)


vscode/src/documentation.ts
```diff
@@ -14,7 +14,7 @@ export async function showDocumentationCommand(extensionUri: Uri) {
 
   // Reveal panel and show 'Loading...' for immediate feedback.
   sendMessageToPanel(
-    "documentation", // This is needed to route the message to the proper panel
+    { panelType: "documentation", id: "" }, // This is needed to route the message to the proper panel
```
minestarks, line 17: Though `""` technically works, I dislike the use of empty string to mean "none". Making `id` an optional field and simply not passing it in here would make the semantics a bit clearer I think.

vscode/src/webviewPanel.ts
```diff
@@ -390,38 +395,65 @@ export function registerWebViewCommands(context: ExtensionContext) {
...
+    return panels[type][id];
+  } else {
+    const title =
+      type == "help"
+        ? "Q# Help"
+        : type == "documentation"
+          ? "Q# Documentation"
+          : type == "estimates"
+            ? "Q# Estimates"
+            : `${id} ${panelTypeToTitle[type]}`;
```
minestarks, line 453: I think a title like `Q# Histogram - NameOfFile.qs` would look more consistent with the other tabs. 

Screenshot 2025-01-24 134235

vscode/src/webview/webview.tsx
```diff
@@ -47,24 +49,26 @@ type EstimatesState = {
 
 type CircuitState = {
   viewType: "circuit";
+  panelId: string;
   props: CircuitProps;
 };
 
 type DocumentationState = {
   viewType: "documentation";
+  panelId: string;
```
minestarks, line 58: Do you really need to pass in `panelId` for Documentation and Estimates, since they don't actually have distinct panel IDs?

minestarks (approved):
I like this change.

(Next time we touch webviewPanel.ts we really should refactor out the different panel types.)

vscode/src/circuit.ts
```diff
@@ -332,6 +332,7 @@ export function updateCircuitPanel(
     calculating?: boolean;
   },
 ) {
+  const panelId = params?.operation?.operation || projectName;
```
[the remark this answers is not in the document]
ScottCarda-MS: Yah bit of an edge case, but you are right. The panels go off of the title, and that is what a user would see as the identity of the panel anyways. Seems like an acceptable edge case behavior.

vscode/src/webviewPanel.ts
ScottCarda-MS: done

vscode/src/documentation.ts
ScottCarda-MS: I've made the id field optional, though it gets filled in as `""` when we actually access the underlying map. Maps have to have strings (or numbers) as their keys and can't have `undefined`.

vscode/src/webview/webview.tsx
ScottCarda-MS: removed

sezna:
Should we get _another_ tab if I run the same thing twice? Like, two histogram tabs if I run histogram twice?  I know the intent of the change is not to do that, but I'm asking more if anybody feels like that's desirable. I'm undecided...but it is what I expected when I checked out the branch.

Other than that, I'm unable to break it. There's a flicker when the tab opens on my screen, I assume from the HTML and CSS rendering in not-perfect-harmony, so that's probably unavoidable.

sezna (approved):

ScottCarda-MS:
> Should we get _another_ tab if I run the same thing twice? Like, two histogram tabs if I run histogram twice? I know the intent of the change is not to do that, but I'm asking more if anybody feels like that's desirable. I'm undecided...but it is what I expected when I checked out the branch.
> 
> Other than that, I'm unable to break it. There's a flicker when the tab opens on my screen, I assume from the HTML and CSS rendering in not-perfect-harmony, so that's probably unavoidable.

I think, rather than having multiple histograms for the same program, what I would want as a user would be to have one histogram that I can _add onto_ with further histogram calls. This is similar to how the resource estimator view works. Though, I think either the user or we would have to be careful about representing the results alongside results from the a modified version of the code.

I.e. I generated a histogram from 100 shots, make a change to the code, and click histogram again. If we naïvely group the results together in an updated histogram, then it is awkwardly grouping together results from two different versions of the code.

ScottCarda-MS:
> > Should we get _another_ tab if I run the same thing twice? Like, two histogram tabs if I run histogram twice? I know the intent of the change is not to do that, but I'm asking more if anybody feels like that's desirable. I'm undecided...but it is what I expected when I checked out the branch.
> > Other than that, I'm unable to break it. There's a flicker when the tab opens on my screen, I assume from the HTML and CSS rendering in not-perfect-harmony, so that's probably unavoidable.
> 
> I think, rather than having multiple histograms for the same program, what I would want as a user would be to have one histogram that I can _add onto_ with further histogram calls. This is similar to how the resource estimator view works. Though, I think either the user or we would have to be careful about representing the results alongside results from the a modified version of the code.
> 
> i.e. I generated a histogram from 100 shots, make a change to the code, and click histogram again with another 100 shots. If we naïvely group the results together in an updated histogram from the combined 200 shots, then it is awkwardly grouping together results from two different versions of the code.

Merging this PR in, as this could be a future feature.

PR #2095, opened 2025-01-07
Update test callable discovery to work within the language service

sezna:
This PR rewrites #2059  to trigger test discovery from within the language server, as opposed to externally in an entirely separate context. Please read the description in that PR for more detail.

It also adds tests for LS state in both the Rust-based LS tests and the JS-based `basics.js` npm API tests.


wasm/src/test_discovery.rs
```diff
@@ -0,0 +1,60 @@
...
+    },
+    r#"export interface ITestDescriptor {
+        callableName: string; 
+        location: ILocation;
+    }"#,
+    ITestDescriptor
+}
+
+#[wasm_bindgen]
+pub fn get_test_callables(config: ProgramConfig) -> Result<Vec<ITestDescriptor>, String> {
```
minestarks, line 28: Dead code?

wasm/src/language_service.rs
```diff
@@ -587,3 +613,9 @@ extern "C" {
     )]
     pub type DiagnosticsCallback;
 }
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(typescript_type = "(callables: [string, ILocation][]) => void")]
```
minestarks, line 619: Use the `ITestCallable` type you already defined in the other wasm file, instead of the `[string, ILocation]` tuple. Tuples are clunky in TS.

npm/qsharp/src/language-service/language-service.ts
```diff
@@ -26,8 +26,7 @@ import {
 } from "../workers/common.js";
 type QscWasm = typeof import("../../lib/web/qsc_wasm.js");
 
-// Only one event type for now
-export type LanguageServiceEvent = {
+export type LanguageServiceDiagnosticEvent = {
```
minestarks, line 29: Wow I had totally forgotten that diagnostics were `Event`s at this layer. I'm so happy/relieved to rediscover this. Since we're not plumbing callback arguments through seven layers of JavaScript, it's not going to be a QCOM breaking change :)

npm/qsharp/src/language-service/language-service.ts
```diff
@@ -274,6 +285,18 @@ export class QSharpLanguageService implements ILanguageService {
       log.error("Error in onDiagnostics", e);
     }
   }
+
+  async onTestCallables(callables: [string, ILocation][]) {
```
minestarks, line 289: As mentioned in the WASM layer, I think this type can be `ITestCallables` instead of a tuple, the reasoning being that objects are just easier to use/read in JavaScript

npm/qsharp/src/compiler/compiler.ts
minestarks, line 1: I think the changes in this file are dead code now.

wasm/src/language_service.rs
```diff
@@ -58,7 +59,32 @@ impl LanguageService {
...
+
+        let test_callables_callback = test_callables_callback
+            .dyn_ref::<js_sys::Function>()
+            .expect("expected a valid JS function")
+            .clone();
+
+        let test_callables_callback = move |update: TestCallables| {
+            let callables = update
+                .callables
+                .iter()
```
minestarks, line 71: You can `.into_iter()` this and avoid cloning the contents.

minestarks (commented):
Not done looking at all the files, posting all my comments for today.

library/signed/src/Tests.qs
```diff
@@ -5,17 +5,10 @@ import Std.Diagnostics.Fact;
...
-
-}
-
-function MeasureSignedIntTests() : (String, Int, (Qubit[]) => (), (Qubit[]) => Int, Int)[] {
-    [
-        ("0b0001 == 1", 4, (qs) => X(qs[0]), (qs) => MeasureSignedInteger(qs, 4), 1),
+@Test()
+operation MeasureSignedIntTests() : Unit {
+    let testCases = [
+        ("0b0001 == 1", 4, (qs) => X(qs[0]), (qs) => MeasureSignedInteger(qs, 6), 11),
```
minestarks, line 11: This doesn't actually pass for me .

vscode/src/language-service/testExplorer.ts
```diff
@@ -0,0 +1,198 @@
...
+  testController.createRunProfile(
+    "Interpreter",
+    vscode.TestRunProfileKind.Run,
+    runHandler,
+    true,
+    undefined,
+    false,
+  );
+
+  const testMetadata = new WeakMap<vscode.TestItem, number>();
```
minestarks, line 132: Interesting, I don't think I've ever seen a `WeakMap` used in practice! Is the idea that the test controller owns these items, and not this map?

library/signed/src/Tests.qs
minestarks, line 1: In addition to the tests here I'd love it if you added a samples under `samples/` showing how to use this new attribute.

language_service/src/protocol.rs
```diff
@@ -33,6 +34,12 @@ pub struct DiagnosticUpdate {
     pub errors: Vec<ErrorKind>,
 }
 
+#[derive(Debug)]
+pub struct TestCallables {
+    pub callables: Vec<(String, Location)>,
+    pub version: Option<u32>,
```
minestarks, line 40: Unused field (and rightly so!)

minestarks (changes requested):
Finished looking. I really like the overall approach. Also I assume integration tests are coming up in a different PR.

minestarks:
@sezna  hitting this odd bug where you can't navigate to the test case after it fails.

test-explorer

vscode/src/language-service/testExplorer.ts
billti: That's from the docs. Search on `weakmap` on https://code.visualstudio.com/api/extension-guides/testing

sezna: This idea came from this page: https://code.visualstudio.com/api/extension-guides/testing (search for weakmap) -- to be honest, I'm not 100% sure why they recommend a `WeakMap` versus a regular key/value store here.

billti: > to be honest, I'm not 100% sure why they recommend a WeakMap versus a regular key/value store here

Because a regular key/value store would hold a strong reference to the TestItem that is the key you put in it, so it would never get garbage collected (unless you manually scrubbed the map and removed items which no longer exist in the controller, which you don't in this code currently. You delete no longer needed items in line 182, but you don't remove them from the map, so they will remain held in memory by that reference if not using a WeakMap).

From the MDN docs: `an object's presence as a key in a WeakMap does not prevent the object from being garbage collected. Once an object used as a key has been collected, its corresponding values in any WeakMap become candidates for garbage collection as well `

swernli:
I ran into an issue where if I open a notebook that includes Q# cells the test explorer loses its content and can't get it back until a full window reload:

https://github.com/user-attachments/assets/9d3925be-7e01-4bb9-bbcb-a7c11745767d

library/signed/src/Tests.qs
sezna: oops, typo'd `1` as `11`. jittery hands.

sezna:
Just FYI @minestarks  @swernli  that test-disappearing bug is fixed and this is all green again.

npm/qsharp/src/language-service/language-service.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import { ITestDescriptor } from "../../lib/node/qsc_wasm.cjs";
```
billti, line 4: This should be imported as part of the following import list.

npm/qsharp/src/language-service/language-service.ts
```diff
@@ -274,6 +286,18 @@ export class QSharpLanguageService implements ILanguageService {
       log.error("Error in onDiagnostics", e);
     }
   }
+
+  async onTestCallables(callables: ITestDescriptor[]) {
+    try {
+      const event = new Event("testCallables") as LanguageServiceEvent & Event;
```
billti, line 292: Should this and onDiagnostics (line 278 above) now be the specific event type? (e.g. LanguageServiceDiagnosticEvent above and the TestCallables version here)

playground/src/editor.tsx
```diff
@@ -8,12 +8,12 @@ import {
   CompilerState,
   ICompilerWorker,
   ILanguageServiceWorker,
-  LanguageServiceEvent,
   QscEventTarget,
   VSDiagnostic,
   log,
   ProgramConfig,
   TargetProfile,
+  LanguageServiceDiagnosticEvent,
```
billti, line 16: With this rename it'll likely be a breaking change for QCOM. We'll need to ensure we communicate this when they take a new drop.

wasm/src/language_service.rs
```diff
@@ -58,7 +60,35 @@ impl LanguageService {
                 )
                 .expect("callback should succeed");
         };
-        let mut worker = self.0.create_update_worker(diagnostics_callback, host);
+
+        let test_callables_callback = test_callables_callback
```
billti, line 64: Is the initial call done on line 40 for the diagnostics callback unnecessary then? (Mine thinks it might be)

billti:
When a test case fails and I click on it to go to the test source, I get the below error.

Interestingly, when the test passes, it seems to work fine!

image

billti:
The 'Run Tests' button is not running any tests. When I click Run on the items in the tree view it works, but this button doesn't.

image

playground/src/editor.tsx
sezna: It is wrapped in a type of the same name as before, this is just one variant of the sum type -- @minestarks [mentioned that it isn't breaking for QCOM](https://github.com/microsoft/qsharp/pull/2095#discussion_r1906211203) here. Are you sure this is breaking? Happy to communicate it either way.

swernli (approved):
Signing off based on experience with local testing and using this. Can't wait to get it in!

language_service/src/compilation.rs
```diff
@@ -46,6 +47,8 @@ pub(crate) enum CompilationKind {
     /// one or more sources, and a target profile.
     OpenProject {
         package_graph_sources: PackageGraphSources,
+        /// a human-readable name for the package (not a unique URI -- meant to be read by humans)
+        human_readable_name: Arc<str>,
```
minestarks, line 51: Just cause we refer to this as the "friendly name" in `project.rs`, how about `friendly_name`? Since you also used `pretty_name` below, I guess this went through a few renames... I think they're all good options but maybe we should stick to one term.

language_service/src/protocol.rs
```diff
@@ -33,6 +34,21 @@ pub struct DiagnosticUpdate {
...
+pub type CompilationUri = String;
+
+/// A human readable name that represents the compilation.
+pub type HumanReadableName = String;
+
+#[derive(Debug)]
+pub struct TestCallables {
+    pub callables: Vec<(CompilationUri, CallableName, Location, HumanReadableName)>,
+}
+
```
minestarks, line 51: At which point do you admit that just declaring a struct with named fields will be easier? :P

language_service/src/state/tests.rs
```diff
@@ -1590,34 +2041,52 @@ impl Display for DiagnosticUpdate {
...
+    };
+
     CompilationStateUpdater::new(
         Rc::new(RefCell::new(CompilationState::default())),
         diagnostic_receiver,
+        test_callable_receiver,
         TestProjectHost {
             fs: TEST_FS.with(Clone::clone),
         },
+        Encoding::Utf16,
```
minestarks, line 2065: Gotcha again! Because unit tests are run all in Rust, we use utf8 here. (utf16 is used only when sending locations to JavaScript)

(it really doesn't matter for your test cases now, but this can trip us up in the future if we add a test case in the future that uses unicode characters or something)

```suggestion
        Encoding::Utf16,
```

language_service/src/state/tests.rs
```diff
@@ -1590,34 +2041,52 @@ impl Display for DiagnosticUpdate {
...
+        let mut v = received_test_cases.borrow_mut();
+        v.push(update);
+    };
+
     CompilationStateUpdater::new(
         Rc::new(RefCell::new(CompilationState::default())),
         diagnostic_receiver,
+        test_callable_receiver,
         TestProjectHost { fs: fs.clone() },
+        Encoding::Utf16,
```
minestarks, line 2089:
```suggestion
        Encoding::Utf8,
```

playground/src/editor.tsx
minestarks: Oh crap I'm sorry. I missed that this might trigger a TypeScript error, even though it's not a breaking change at runtime. Should be straightforward to update QCOM though.

vscode/src/language-service/testExplorer.ts
```diff
@@ -0,0 +1,232 @@
...
+                  uri: vscode.Uri.parse(failureLocation),
+                },
+        };
+        run.failed(testCase, message);
+      }
+      run.end();
+    });
+
+    const callableExpr = `${testCase.id}()`;
+    const uri = testMetadata.get(testCase)?.compilationUri;
```
minestarks, line 122: You don't really need to store `compilationUri` anywhere as you can use `testCase.uri` directly here, no?

vscode/src/projectSystem.ts
```diff
@@ -24,6 +24,14 @@ async function findManifestDocument(
...
   const currentDocumentUri = URI.parse(currentDocumentUriString);
 
+  // if this document is itself a manifest file, then we've found it
+  if (currentDocumentUri.path.endsWith("qsharp.json")) {
+    return {
+      directory: Utils.dirname(currentDocumentUri),
+      manifest: currentDocumentUri,
+    };
+  }
+
```
minestarks, line 34: I assume this change was because you look up the program using `compilationUri` in `testExplorer.ts` ?

minestarks (approved):

vscode/src/language-service/testExplorer.ts
sezna: `testCase.uri` is the uri where the test case is located, but `compilationUri` is the uri of the compilation, i.e. manifest file...technically both would work but i thought using the compilation uri was cleaner

vscode/src/projectSystem.ts
sezna: yep

vscode/src/language-service/testExplorer.ts
sezna: I went ahead and switched to finding the compilation based on testCase.uri here

vscode/src/language-service/testExplorer.ts
```diff
@@ -0,0 +1,220 @@
...
+    const callableExpr = `${testCase.id}()`;
+    const uri = testCase.uri;
+    if (!uri) {
+      log.error(`No compilation URI for test ${testCase.id}`);
+      run.appendOutput(`No compilation URI for test ${testCase.id}\r\n`);
+      return;
+    }
+    const programResult = await getProgramForDocument(uri);
+
+    log.info(JSON.stringify(programResult, null, 2));
```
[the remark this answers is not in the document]
sezna: good catch!

vscode/src/language-service/testExplorer.ts
```diff
@@ -0,0 +1,220 @@
...
+  const testVersions = new WeakMap<vscode.TestItem, number>();
+  async function onTestCallables(evt: {
+    detail: {
+      callables: ITestDescriptor[];
+    };
+  }) {
+    let currentVersion = 0;
+    for (const [, testItem] of testController.items) {
+      currentVersion = (testVersions.get(testItem) || 0) + 1;
+      break;
```
billti, line 160: So this immediately breaks after the first item? Why use a for loop at all rather than just get the first item (or 1 if there isn't one)?

billti: Oh right, it's a WeakMap and there is no ordering... don't mind me.

sezna: I did this because there's no index signature in TS for testcontroller.items:

image

and `.get` wants a string key:
image

but it _is_ an iterator. So this was how I figured I could get the first element while appeasing the API. Is there a better way?

vscode/src/language-service/testExplorer.ts
```diff
@@ -0,0 +1,219 @@
...
+      const vscLocation = toVsCodeLocation(location);
+      const parts = [friendlyName, ...callableName.split(".")];
+
+      let rover = testController.items;
+      for (let i = 0; i < parts.length; i++) {
+        const part = parts[i];
+        const id =
+          i === parts.length - 1
+            ? callableName
+            : parts.slice(0, i + 1).join(".");
```
billti, line 173: I'm struggling to follow the logic here a bit. Can you add some explanatory comments? For example, parts is made up of the callable name split by ".", and here you are iterative over those parts, yet if on the last part the 'id' is the whole callable name, not including the friendly name and not split by ".", but if on an earlier part it is including the friendlyName and upto and including that part?

I'm not sure I follow. Some comments would be a lot easier than having to reverse engineer to try and understand.

sezna: I've documented this in 92fbb1a -- let me know if it still isn't clear

PR #2100, opened 2025-01-09
Added comments to sample tests

DmitryVasilevsky:
1. Added clarification comment why we need two expected strings in sample testing.
2. Made general comment a normal comment rather than doc comment - otherwise it was bound to the first constant in the file.

### DIFF
--- a/samples_test/src/tests.rs
+++ b/samples_test/src/tests.rs
@@ -20,6 +20,10 @@ use qsc::{
 };
 use qsc_project::{FileSystem, StdFs};
 
+// Two tests are needed to check interpreter working in debug and non-debug mode.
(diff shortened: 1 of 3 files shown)


samples_test/src/tests.rs
```diff
@@ -20,6 +20,10 @@ use qsc::{
 };
 use qsc_project::{FileSystem, StdFs};
 
+// Two tests are needed to check interpreter working in debug and non-debug mode.
+// This results in two exepected stings defined. Although two strings are typically the same,
```
orpuente-MS, line 24:
```suggestion
// This results in two expected strings defined. Although two strings are typically the same,
```

orpuente-MS (approved):

swernli (approved):

PR #2103, opened 2025-01-14
Add lint groups to Q#

orpuente-MS:
This PR adds the necessary infrastructure to define lint groups. It also adds the `deprecation` lint group, which allows the user to enable all deprecation lints at the same time, which are set to `allow` by default.

A lint group can be added to the `qsharp.json` manifest as follows:

```json
{
   "lints": [
      { "group": "deprecations", "level": "warn" }
   ]
}
```


compiler/qsc_linter/src/lint_groups.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// The lint group.
+    pub lint_group: LintGroup,
+    /// The group level.
+    pub level: LintLevel,
+}
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub enum LintGroup {
+    Pedantic,
```
minestarks, line 18: Why "pedantic"?

It has a slightly negative connotation, like "so picky as to be annoying". I do know Clippy uses it, but I think the name makes some sense there: they have many other groups you would want to enable _before_ pedantic. So this is the one that's like [a bunch of opinionated lints we didn't know where else to put](https://doc.rust-lang.org/clippy/usage.html#clippypedantic). Even their documentation hedges it by saying "[if you enable this group, expect to also use `#[allow]` attributes generously](https://doc.rust-lang.org/clippy/lints.html#pedantic)" 

If this is our _only_ lint group, I don't think "pedantic" makes sense.

My other problem with the name is it doesn't describe a meaningful category or purpose. As a user, how can I tell what I'm getting out of these lints by opting into them?


Do we need groups? If we do, what about more functional categories like "correctness", "style", "deprecations" etc?

sezna: I agree with you here, some extra color to the conversation: I think the pedantic lints in Clippy tend to be much more subjective and not as universally-accepted. I don't think the name _pedantic_ conveys that, but maybe we can come up with a better word. Stylistic? Strict? Opinionated? I like opinionated as a group name.

orpuente-MS: I removed pedantic and added the `deprecation`, which as the name suggests lets you enable all deprecation related lints at the same time, which are allowed by defautl.

minestarks:
Should we update our qsharp.json schema? ( vscode\qsharp.schema.json )

compiler/qsc_linter/src/lint_groups.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// The lint group.
+    pub lint_group: LintGroup,
+    /// The group level.
+    pub level: LintLevel,
+}
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub enum LintGroup {
+    Deprecations,
```
ScottCarda-MS, line 18: So it is just this deprecation group for now? Do you have any thoughts on what some other lint groups might be in the future?

vscode/qsharp.schema.json
```diff
@@ -26,28 +26,50 @@
...
-              "redundantSemicolons",
-              "deprecatedNewtype",
-              "deprecatedSet",
-              "discourageChainAssignment",
-              "needlessOperation",
-              "deprecatedFunctionConstructor",
-              "deprecatedWithOperator",
-              "deprecatedDoubleColonOperator"
-            ]
+        "oneOf": [
```
ScottCarda-MS, line 31: Do you have public-facing documentation somewhere that tells users how to specify lint settings in their Q# project files? If so please remember to update that documentation as well. If not, please find an appropriate place to add it.

compiler/qsc_linter/src/lint_groups.rs
orpuente-MS: Maybe one for OpenQASM. But for now that's it.

compiler/qsc_linter/src/linter.rs
```diff
@@ -211,11 +254,21 @@ impl Display for LintLevel {
     }
 }
 
-/// End-user configuration for each lint level.
+/// End-user configuration for a specific lint or a lint group.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum LintOrGroupConfig {
```
minestarks, line 260: Since `level` is a common field, I probably would have organized the types like this - fewer structs. It doesn't make a meaningful difference at runtime, and it might make deserialization awkward I don't know. So take it or leave it - I just wanted to make sure you considered it

```rs
struct LintConfig {
    kind: LintConfigKind,
    level: LintLevel,
}

enum LintConfigKind {
    Lint(LintKind),
    Group(LintGroup)
}
```

minestarks: Hmm, I just saw `unfold_group` above - my suggestion would probably make that function more awkward. So I can see why you did it this way

compiler/qsc_linter/src/lint_groups.rs
minestarks, line 1: Please add copyright header

minestarks (approved):

ScottCarda-MS (approved):

orpuente-MS:

PR #2104, opened 2025-01-14
Add lint for double (in)equality

orpuente-MS:
This PR adds a lint for double equality and inequality.

It also updates the DF Chemistry sample to perform a correct `NAN` check.

### DIFF
--- a/compiler/qsc_linter/src/lints/hir.rs
+++ b/compiler/qsc_linter/src/lints/hir.rs
@@ -6,10 +6,10 @@ use std::rc::Rc;
 use qsc_data_structures::span::Span;
 use qsc_hir::{
     hir::{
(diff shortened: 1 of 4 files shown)


swernli (approved):

compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -35,12 +35,42 @@ use super::lint;
...
 }
 
+#[derive(Default)]
+struct DoubleEquality {
+    level: LintLevel,
+}
+
+impl HirLintPass for DoubleEquality {
+    fn check_expr(&mut self, expr: &Expr, buffer: &mut Vec<Lint>, _compilation: Compilation) {
+        // Ignore the case where the identifier is being compared for equlity with itself,
```
sezna, line 52:
```suggestion
        // Ignore the case where the identifier is being compared for equality with itself,
```

sezna (approved):

PR #2105, opened 2025-01-16
Chemistry library, SPSA sample, and notebook

swernli:
Converting the old chemistry library and Variational Quantum Algorithms sample to modern QDK. As part of this, I tried where possible to update to our newer Q# programming patterns. I've tried to make sure we export on the reasonable part of the APIs as public facing, but if there is something that is not exported and should be or something that is exported and shouldn't be we can adjust.

For the original library code, see https://github.com/microsoft/QuantumLibraries/tree/main/Chemistry/src/Runtime.
For the original sample notebook, see https://github.com/microsoft/Quantum/blob/main/samples/algorithms/variational-algorithms/Variational%20Quantum%20Algorithms.ipynb


msoeken (approved):

library/chemistry/src/JordanWigner/Convenience.qs
```diff
@@ -0,0 +1,112 @@
...
+/// ## trotterStepSize
+/// Step size of Trotter integrator.
+/// ## trotterOrder
+/// Order of Trotter integrator.
+///
+/// # Output
+/// A tuple where: `Int` is the number of qubits allocated,
+/// `Double` is `1.0/trotterStepSize`, and the operation
+/// is the Trotter step.
+function TrotterStepOracle(qSharpData : JordanWignerEncodingData, trotterStepSize : Double, trotterOrder : Int) : (Int, (Double, (Qubit[] => Unit is Adj + Ctl))) {
```
DmitryVasilevsky, line 42: Here and in other functions. qSharpData identifier doesn't describe this parameter well. Maybe jwData? Also probably better to use individual fields rather than splitting into fields with !. statePrepData and energyShift aren't used here.
This applies to other functions. Also, this is just a small issue, we can discuss if we need to change this. I can push things like this into your branch.

DmitryVasilevsky: I prepared a PR.

library/chemistry/src/JordanWigner/StatePreparation.qs
```diff
@@ -0,0 +1,779 @@
...
+/// ```qsharp
+/// use qubits = Qubit();
+/// let qubitsLE = LittleEndian([qubits]);
+/// PrepareArbitraryStateCP([ComplexPolar(1.0/Sqrt(2.0),0.0),ComplexPolar(1.0/Sqrt(2.0),PI()/2.0)],qubitsLE); // = |i>
+/// ```
+///
+/// # References
+/// - [Synthesis of Quantum Logic Circuits](https://arxiv.org/abs/quant-ph/0406176)
+///   Vivek V. Shende, Stephen S. Bullock, Igor L. Markov
+operation PrepareArbitraryStateCP(coefficients : ComplexPolar[], qubits : Qubit[]) : Unit is Adj + Ctl {
```
DmitryVasilevsky, line 188: I'm removing some of these and replacing them with our Std library implementations.

library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
```diff
@@ -0,0 +1,1184 @@
...
+    // Create uniform superposition over index and alt coeff register.
+    PrepareUniformSuperposition(nCoeffs, indexRegister);
+    ApplyToEachCA(H, uniformKeepCoeffRegister);
+
+    // Write bitstrings to altIndex and keepCoeff register.
+    let unitaryGenerator = (nCoeffs, idx -> WriteQuantumROMBitString(idx, keepCoeff, altIndex, data, _, _, _, _));
+    MultiplexOperationsFromGenerator(unitaryGenerator, indexRegister, (keepCoeffRegister, altIndexRegister, dataRegister, altDataRegister));
+
+    // Perform comparison
+    CompareUsingRippleCarry(uniformKeepCoeffRegister, keepCoeffRegister, flagQubit);
```
DmitryVasilevsky, line 626: We have this in Arithmetic.

DmitryVasilevsky (approved):

PR #2110, opened 2025-01-20
Remove scipy

billti:
Scipy does not work on Windows ARM64 currently (see https://github.com/scipy/scipy/issues/21562) which is a platform we support.

This is the only sample that uses scipy, so removing that usage for the equivalent numpy usage.

CC @msoeken and @aarthims

### DIFF
--- a/samples/estimation/df-chemistry/chemistry.py
+++ b/samples/estimation/df-chemistry/chemistry.py
@@ -3,11 +3,11 @@
 import json
 import math
 import numpy as np
+import numpy.linalg as LA
 import numpy.typing as npt
 import qsharp
 from argparse import ArgumentParser
 from dataclasses import dataclass
-from scipy import linalg as LA
 from urllib.parse import urlparse
 from urllib.request import urlretrieve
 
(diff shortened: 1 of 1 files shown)


orpuente-MS (approved):

samples/estimation/df-chemistry/chemistry.py
```diff
@@ -294,7 +294,7 @@ def eigen_decomp(
         matrix[np.tril_indices(dimension, 0)] = vector
 
         # Compute the eigen decomposition of the lower triangular matrix
-        evals, evecs = LA.eigh(matrix, lower=True)
+        evals, evecs = LA.eigh(matrix, UPLO="L")
```
msoeken, line 297:
```suggestion
        evals, evecs = np.linalg.eigh(matrix, UPLO="L")
```

to be consistent with the following two lines. In that case the new import can also be removed.

msoeken (approved):
Looks good to me. Thank you

minestarks (approved):

samples/estimation/df-chemistry/chemistry.py
billti: I fixed them all to use LA consistently, as otherwise the numpy.linalg submodule isn't actually imported, which in Python it is supposed to  be before it is used.

minestarks:

PR #2112, opened 2025-01-21
Fix global phase for controlled-T, R1

### LINKED ISSUE #2106: DumpOperation has unexpected global phase for double-controlled-R1 gate
**Describe the bug**

`Std.Diagnostics.DumpOperation`, when applied to R1 gate with 2 controls, produces incorrect matrix.

**To Reproduce**

First, run this code:
```
operation Op(qs: Qubit[]) : Unit is Adj {
    Controlled Z([qs[0], qs[1]], (qs[2]));
}
operation Main() : Unit {
    Std.Diagnostics.DumpOperation(3, Op);
}
```

You will get this correct output:
```
 1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  1.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖, −1.0000+0.0000𝑖
```

Then, run this code:
```
operation Op(qs: Qubit[]) : Unit is Adj {
    Controlled R1([qs[0], qs[1]], (Std.Math.PI(), qs[2]));
}
operation Main() : Unit {
    Std.Diagnostics.DumpOperation(3, Op);
}
```
You will get:
```
 0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.9239−0.3827𝑖,  0.0000+0.0000𝑖
 0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖,  0.0000+0.0000𝑖, −0.9239+0.3827𝑖
```
This is the correct matrix, multiplied by the global phase `0.9239−0.3827𝑖`.

The second snippet should return the same matrix as the first, because by [definition](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/std.intrinsic/r1),  `R1(π)=diag(1, e^iπ)=diag(1,-1)=Z`.

**Expected behavior**

Matrices from both snippets should match.

**Screenshots**

N/A

**System information**

qsharp 1.12.1

**Additional context**

* [Documentation for DumpOperation](https://learn.microsoft.com/en-us/qsharp/api/qsharp-lang/std.diagnostics/dumpoperation) does say that global phase is expected when there are other qubits allocated. However, in this example there are no other qubits besides the three qubits on which this operation acts. If this is intended behaviour, please update the documentation.
* I discovered this bug when migrating tests for my [quantum_decomp](https://github.com/fedimser/quantum_decomp) library to use modern Q# in tests (and use DumpUnitary instead of calling DumpMachine many times). It worked for 2x2 and 4x4 matrices, but failed for 8x8 matrices. [Temporary fix](https://github.com/fedimser/quantum_decomp/commit/31bb37d26014ece69d3e2676b0699f198cdcac70) is to compare matrices up to global phase.

swernli:
The decomposition for controlled-`T` in the stdlib uses `Rz` gates that are correct up to a global phase, which is fine for hardware but distracting in simulation. Correcting for that global phase in `T` fixes the global phase in `S`, which in turn allows removing a patch that made `R1` use a global phase consistent with `S`.

Fixes #2106


library/src/tests/intrinsic.rs
```diff
@@ -1832,38 +1832,38 @@ fn test_base_mctadj_4_control() {
...
+        |1011010110⟩: 0.1768+0.0000𝑖
+        |1011110111⟩: 0.1768+0.0000𝑖
+        |1100011000⟩: 0.1768+0.0000𝑖
+        |1100111001⟩: 0.1768+0.0000𝑖
+        |1101011010⟩: 0.1768+0.0000𝑖
+        |1101111011⟩: 0.1768+0.0000𝑖
+        |1110011100⟩: 0.1768+0.0000𝑖
+        |1110111101⟩: 0.1768+0.0000𝑖
+        |1111011110⟩: 0.1768+0.0000𝑖
+        |1111111111⟩: 0.1250−0.1250𝑖
```
billti, line 1866: That's a LOT of expect tests that were previously 'wrong'. How did we not catch this before? Or did we just not worry about it in these cases before for some reason?

swernli: Yeah, it's an embarrassing oversight on my part. Because we have other tests on smaller states for S that looked correct, I assumed these were correct and didn't dig into the state. But even still, controlled-T being off should have been easier to catch.

One thing we can consider, now that we have DumpOperation support, is converting these test to be based on the matrix rather than the state, which is easier to verify both for the original gate and for controls of any size (since it's just diagonal ones for the additional qubits). If you think it's worth the churn we could do that conversion here, if not we can save it for another test focused PR.

billti: No problem. Thanks for the details. I'll leave it up to you as to if/when the changes to validation make sense.

library/std/src/Std/InternalHelpers.qs
```diff
@@ -90,6 +83,9 @@ internal operation CT(control : Qubit, target : Qubit) : Unit is Adj {
     CNOT(control, target);
     Adjoint Rz(angle, target);
     CNOT(control, target);
+    // This decomposition for controlled-T introduces a global phase (due to the unmatched call to Rz from above).
+    // We correct for this global phase in simulation with R(PauliI), which is a no-op on hardware.
+    Adjoint R(PauliI, angle, target);
```
DmitryVasilevsky, line 88: If the purpose of this is just to apply global phase (rather than provide some canonical decomposition form) maybe just call ApplyGlobalPhase?
ApplyGlobalPhase(angle / 2.0)

swernli: That's a good point, this is an internal helper so there is no need to worry about expressing it in terms of `R`. I'll make that change.

orpuente-MS (approved):

PR #2113, opened 2025-01-21
put a limit on number of operations in circuit

### LINKED ISSUE #2099: Running a program that involves more than `isize::MAX` gates triggers a panic in circuit builder
Because in compiler/qsc/src/interpret.rs we always create a chained simulator that includes a circuit builder, every intrinsic gate call gets added to a running tally of gates performed. Rust has a limit on the capacity of `Vec` to `isize::MAX` so if more than that many gates are used the push of the gate after the max triggers a panic. We can work around this temporarily by having the circuit builder stop tracking when hits a hard coded limit, but we should also consider updates to the interpreter behavior (and the corresponding `dump_circuit()` exposed in Python) to avoid tracing gates for a circuit if not in a debug mode, which should have additional performance benefits.

[Here's a playground repro (screenshot below).](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAEy2MPQvCQBBE%252B%252FsVA2mSJifaKXaClUX8ACtlk6xxwdzFuxUL8b%252FLxUwxxWPmWYsMB%252BqHBxtrsWVVcR2iUlBujbWJZthwbIIMKt4lcLxLhEQQenHS0wNVhiH4LlAPvZOiIYea8YrcQv3fh3eQ0V5laHzLpTF%252B4EDJih2JywsscXKi%252BBgA6Y0n1qhetWherEZ48wFXiMOsLOeXxXzappzz5zT6jr3nyDqyr%252FkBVSuKmekAAAA%253D&profile=unrestricted)

Image

minestarks:
Fixes #2099

### DIFF
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -676,6 +676,7 @@ impl Interpreter {
         } else {
             let mut sim = CircuitBuilder::new(CircuitConfig {
                 base_profile: self.capabilities.is_empty(),
+                max_operations: CircuitConfig::DEFAULT_MAX_OPERATIONS,
             });
 
             self.run_with_sim_no_output(entry_expr, &mut sim)?;
@@ -890,6 +891,7 @@ fn sim_circuit_backend() -> BackendChain<SparseSim, CircuitBuilder> {
             // will still respect the selected profile. This also
             // matches the behavior of the simulator.
             base_profile: false,
+            max_operations: CircuitConfig::DEFAULT_MAX_OPERATIONS,
         }),
     )
 }
--- a/compiler/qsc/src/interpret/circuit_tests.rs
+++ b/compiler/qsc/src/interpret/circuit_tests.rs
@@ -776,7 +776,6 @@ fn controlled_operation() {
 }
 
 #[test]
-#[allow(clippy::too_many_lines)]
 fn internal_operation() {
     let mut interpreter = interpreter(
         r"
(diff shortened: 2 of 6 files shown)


compiler/qsc/src/interpret/circuit_tests.rs
minestarks, line 1: Unrelated change. Just fixing the test output formatting here to match the rest of the file.

swernli (approved):

orpuente-MS (approved):

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -81,6 +81,18 @@ pub struct Qubit {
 pub struct Config {
     /// Perform Base Profile decompositions
     pub base_profile: bool,
+    /// Maximum number of operations the builder will add to the circuit
+    pub max_operations: usize,
+}
+
+impl Config {
+    /// Set to the current UI limit + 1 so that it still triggers
+    /// the "this circuit has too many gates" warning in the UI.
```
billti, line 90: Nice! Thanks.

billti (approved):

compiler/qsc/src/interpret.rs
```diff
@@ -676,6 +676,7 @@ impl Interpreter {
         } else {
             let mut sim = CircuitBuilder::new(CircuitConfig {
                 base_profile: self.capabilities.is_empty(),
+                max_operations: CircuitConfig::DEFAULT_MAX_OPERATIONS,
```
billti, line 679: Does this impact resource estimation at all? Obviously when running the resource estimation it wouldn't be that useful if it had a low-ish cap on the number of operations.

PR #2120, opened 2025-01-22
Fix widgets sometimes rendering in light theme when VS Code is in a dark theme

billti:
(See related issues https://github.com/microsoft/vscode-jupyter/issues/7161 and https://github.com/microsoft/vscode-jupyter/issues/9403 for more background)

Our CSS override to address the widget theming issue wasn't working somtimes, which was very frustrating, as in Dark Mode our widgets would often appear as they should in light mode.

The reason is that the order in which the style sheets get injected is non-deterministic, and as the 'problematic' CSS rule and the override to fix it have the same specificity, it would be 'last one wins'. Sometimes our style-sheet is last and it works...

image

And sometimes the IPyWidgets 'force it to white' rule is last so the issue persists...

image

The unfortunate best _hack_ I could come up with is to programmatically ensure the override CSS rule is always last in the `style`s in the header.

Not pretty, but it works.

### DIFF
--- a/widgets/js/index.tsx
+++ b/widgets/js/index.tsx
@@ -40,6 +40,22 @@ type RenderArgs = {
 export function render({ model, el }: RenderArgs) {
   const componentType = model.get("comp");
 
+  // There is an existing issue where in VS Code it always shows the widget background as white.
+  // (See https://github.com/microsoft/vscode-jupyter/issues/7161)
+  // We tried to fix this in CSS by overridding the style, but there is a race condition whereby
+  // depending on which style gets injected first (ours or ipywidgets), it may or may not work.
+
+  // The solution here is to force our own override to be last in the style list if not already.
+  // It's a bit of a hack, but it works, and I couldn't find something better that wouldn't be fragile.
+
+  if (
+    !el.ownerDocument.head.lastChild?.textContent?.includes("widget-css-fix")
+  ) {
+    const forceStyle = el.ownerDocument.createElement("style");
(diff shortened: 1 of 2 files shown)


ScottCarda-MS (approved):

widgets/js/index.tsx
```diff
@@ -40,6 +40,22 @@ type RenderArgs = {
...
+  // We tried to fix this in CSS by overridding the style, but there is a race condition whereby
+  // depending on which style gets injected first (ours or ipywidgets), it may or may not work.
+
+  // The solution here is to force our own override to be last in the style list if not already.
+  // It's a bit of a hack, but it works, and I couldn't find something better that wouldn't be fragile.
+
+  if (
+    !el.ownerDocument.head.lastChild?.textContent?.includes("widget-css-fix")
+  ) {
+    const forceStyle = el.ownerDocument.createElement("style");
```
minestarks, line 54: I don't have a great understanding of ipywidgets, so I'm not sure. Is `ownerDocument` not something we have control over? Can we not embed a `<style>` tag in the HTML directly?

minestarks (approved):

widgets/js/index.tsx
billti: It's not a ipywidgets thing, it's a DOM thing. The element isn't attached to the DOM yet, but every element is associated with an owning document. (You have to call createElement on a document). So this gets the owning document to modify the styles on, and creates a 'style' element that will be associated with that document.

PR #2131, opened 2025-01-27
Deduplicate lints to avoid having repeated lints in operations that have multiple specializations

orpuente-MS:
This PR deduplicates lints in the `run_lints` function, which is the entry point to the linter, to avoid having repeated lints in operations that have multiple specializations.

### DIFF
--- a/compiler/qsc_linter/src/linter.rs
+++ b/compiler/qsc_linter/src/linter.rs
@@ -32,7 +32,13 @@ pub fn run_lints(
     let mut lints = Vec::new();
     lints.append(&mut ast_lints);
     lints.append(&mut hir_lints);
+    remove_duplicates(&mut lints);
     lints
+}
+
+pub(crate) fn remove_duplicates<T: Eq + std::hash::Hash + Clone>(vec: &mut Vec<T>) {
+    let mut seen = rustc_hash::FxHashSet::default();
+    vec.retain(|x| seen.insert(x.clone()));
 }
 
 #[derive(Clone, Copy)]
@@ -117,6 +123,21 @@ pub struct Lint {
     pub code_action_edits: Vec<(String, Span)>,
 }
 
+impl std::hash::Hash for Lint {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.span.hash(state);
+        self.kind.hash(state);
+    }
+}
+
+impl std::cmp::PartialEq for Lint {
+    fn eq(&self, other: &Self) -> bool {
+        self.span == other.span && self.kind == other.kind
+    }
+}
+
+impl std::cmp::Eq for Lint {}
+
 impl std::fmt::Display for Lint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.message)
@@ -188,7 +209,7 @@ pub struct LintConfig {
 }
 
 /// Represents a lint name.
-#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, Hash)]
 #[serde(untagged)]
 pub enum LintKind {
     /// AST lint name.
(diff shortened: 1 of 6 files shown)


compiler/qsc_linter/src/linter.rs
```diff
@@ -32,9 +32,15 @@ pub fn run_lints(
     let mut lints = Vec::new();
     lints.append(&mut ast_lints);
     lints.append(&mut hir_lints);
+    remove_duplicates(&mut lints);
```
DmitryVasilevsky, line 35: Did you consider not to collecting duplicate lints rather than removing them later?

orpuente-MS: I guess I can just collect them in a HashSet from the start, in that way I don't have duplicates in the first place.

I am only afraid that sets don't promise you that the elements are in a specific order, so testing might fail. Any opinions on this?

swernli: Sets don't preserve order, so personally I'd prefer this approach.

swernli (approved):
Looks like a good fix, thanks!

compiler/qsc_linter/src/linter.rs
```diff
@@ -32,9 +32,15 @@ pub fn run_lints(
...
     lints
 }
 
+pub(crate) fn remove_duplicates<T: Eq + std::hash::Hash + Clone>(vec: &mut Vec<T>) {
+    let mut seen = rustc_hash::FxHashSet::default();
```
sezna, line 40: This doesn't actually only remove repeated lints for operations with multiple specs as the title says, it removes all duplicate lints. Are you worried about this obfuscating if we ever introduce other duplicate lints in the future? Or are we ok with just deduplicating? It feels a bit like fixing a symptom instead of a cause, and if we are worried primarily about spec decls, could we deduplicate just for spec decls?

swernli: It's a fair question, but given that this does the hash by `kind` and `span` (same when doing equality check), it means we only drop things if they are the same lint for the exact same span, so I think it's safe to be generally enabled.

orpuente-MS: Just to clarify, what was duplicated here was the expression itself in the HIR. To be more precise, the same expression referring to the same span in the source code, was being referenced in multiple places in the HIR. That's why the linter found it multiple times.

That being said. There are two potential ways in which we could introduce a duplicated lint in the future:
 1. If the expression being analyzed appears multiple times in the HIR, in which case the solution is the current one.
 2. If a contributor to the linter pushes the exact same lint to the `lints` vector multiple times. And even if *somehow* code like that passes our code review process, it shouldn't introduce any undesired behavior or be a performance issue, because pushing a duplicated lint and then deduplicating it is much cheaper than the whole process of traversing the AST & HIR and pattern matching against it. So, I am not really worried about this.

> could we deduplicate just for spec decls?
Not really. Two ways of doing this come to my mind:
 1. The root cause here is that multiple spec decls reference the exact same source code. So, you could use a `HashSet` to check if you have pushed the same lint (referring to the same issue at the exact same span in the source code) to the `lints` vector before. This is pretty much the current solution.
 2. Keeping track of if you are inside a spec decl and not pushing lints in that case, but that wouldn't work well. You can manually define your own specialization declarations in a different place in the source code, and in that case signaling lints is totally valid. So not only you need to keep track if you are in a specialization declaration, but you also need to figure out if it was written manually instead of autogenerated. And that's not it, you also need to the same check for all the lints. Not to mention that there could be lints of interest that apply inside specialization declarations, even if they are autogenerated.

orpuente-MS: I just made our test bench run the unit tests *without* deduplication to protect us against point 2. in the previous message.

sezna: I think that's a reasonable compromise.

I know the spec decls resulted in duplicate HIR, which then duplicated the lints -- I was just concerned that deduplicating _all_ lints to handle that might prevent us from catching other bugs -- for example, if we run a lint pass twice by accident.

Given that the tests will catch this now, I'm approving. Thanks!

sezna (approved):

PR #2134, opened 2025-01-28
Change dependency graph for qsc and qsc_qasm3

idavis:
This change reorganizes the deps for qsc, qsc_qasm3, and the pip package. The `qsc` crate will now depend on `qsc_qasm3` and reexport the appropriate API surface. The pip package now consumes that exported API from `qsc`.


minestarks (approved):
Works for me, but what was the motivation for this?

idavis:
Enabling future work for consuming the crates more generally.

orpuente-MS (approved):

PR #2141, opened 2025-01-31
Update node.js version to the active LTS version 22.13.1

DmitryVasilevsky:
You'll need at least 22.13 to build now. Previous version 18.17.1 was nearing end of support.


minestarks (commented):
Thanks! There are two more spots in the root `package.json` where 18.17 is hardcoded (one under `"engines"` and another one in the `"@types/node"` dependency. Could you please update those to match? (After that, you'll want to run a build and check in the updated `package-lock.json` as well.)

DmitryVasilevsky:
> Thanks! There are two more spots...

Updated.

minestarks (approved):

package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=18.17.1"
+    "node": ">=22.13.1"
```
billti, line 10: Note that if you're adding this line, then the package will give errors for anyone using a lower version of Node.js. Is this version a minimum requirement now? (i.e. to we want to require anyone using this project to be on node 22.13.1 or later?)

DmitryVasilevsky: As per discussion with @minestarks last Friday, yes.
> If you mean v22 (the default one that's offered on the download page) then yes, I believe that's what we should use.

billti: Tagging @idavis on this one. I believe our ADO pipeline uses the Node.js that comes with the image. Our last ADO run points to this image <https://github.com/actions/runner-images/blob/ubuntu22/20250120.2/images/ubuntu/Ubuntu2204-Readme.md> which lists Node 18.20.5 (and I don't believe we install a later version).

I don't think our project should require a later version than our pipelines run with.

billti (approved):

PR #2109, opened 2025-01-18
Fix language service panic when file isn't listed in the `files` field of `qsharp.json`

minestarks:
Fixes #2090

Notably, this also changes how we treat the `files` field in `qsharp.json` for local projects their local dependencies.

Previously, if the `qsharp.json` explicitly listed any `files` , that list would take priority and we would skip crawling the directory.

With this change: we always crawl the `src/` directory and include all found `*.qs` files in the project. But we raise an explicit error to the user if any of those files are missing from the `files` list.

PR #2147, opened 2025-02-03
fix `qsharp.init(project_root='.')`

minestarks:
Fixing regression described in https://github.com/microsoft/qsharp/pull/2109#issuecomment-2629169244


idavis (approved):

swernli (approved):
Tested locally with some scenarios, confirmed it works as expected!

PR #2149, opened 2025-02-04
Reimplement Parser

idavis:
## An intro to the PR with some useful links for the relevant sections:
 
This PR has two major components. A custom QASM3 parser, which replaces the IBM parser. And a refactored version of the compiler.
 
The custom QASM3 parser has three components:
1. The raw lexer.
2. The cooked lexer.
3. Parser.
 
The refactored compiler has two components:
4. Lowerer.
5. Compiler.
 
Below is more detailed overview of the purpose of each component:
 
The OpenQASM 3.0 grammar will be useful while reviewing sections 1-2
https://openqasm.com/grammar/index.html
 
1. The raw lexer:
Takes the source code as input and returns a stream of "raw tokens". You can think of these mostly as characters like '(' and type literals.
 
2. The cooked lexer:
Takes the raw token stream and returns a stream of cooked tokens. Cooked tokens have more knowledge about QASM3 syntax, and they make parsing easier.
 
3. Parser:
Takes the stream of cooked tokens as input and returns an AST. We call it the syntax AST, to differentiate it from the semantic AST, which will be introduced later.

The grammar and specification of QASM3 don't fully agree. We had to take some decissions while implementing the parser. They are documented in the rust code. If you want to double check anything, the language spec will come in handy:
https://openqasm.com/language/index.html
 
4. Lowerer:
Takes the AST returned by the parser and performs semantic analysis on it. This is where we report any QASM3 related errors, like "functions can only be defined on the global scope." The lowerer returns a semantic AST. You will want to use the language spec to verify that the errors that we are checking for are actually on the spec, or that we are not missing any checks.
https://openqasm.com/language/index.html
 
5. Compiler:
Takes the semantic AST and compiles it to a Q# AST. This is where we report any Q# related errors, like unsupported QASM3 features because they don't make sense in Q#. The compiler is a very straighforward mapping from the semantic AST to the Q# AST, since most of the heavy lifting is done during lowering.

## Recomended Sections for reviewers:
  qsc_qasm3/src/lib.rs (Entry point and general structure)
    Mine, Stefan
 
  qsc_qasm3/src/parser/completion.rs (Completion)
  qsc_qasm3/src/io.rs (IO)
    Mine
 
  qsc_qasm3/src/lex.rs (Raw and Cooked lexer)
    Scott
 
  qsc_qasm3/src/parser.rs (Parser)
  qsc_qasm3/src/parser/ast.rs (QASM3 AST, double check against grammar & Spec)
    Scott
 
  qsc_qasm3/src/oqasm_helpers.rs
  qsc_qasm3/src/semantic/types.rs (QASM3 types)
  qsc_qasm3/src/semantic.rs (Lowerer entry point)
  qsc_qasm3/src/lowerer.rs
    This requires QASM3 specific knowledge. We can use as many eyes as possible.
    Dmitry, Scott
 
  qsc_qasm3/src/stdlib.rs
  qsc_qasm3/src/stdlib/angle.rs
  qsc_qasm3/src/stdlib/QasmStdrs
    Dmitry
 
  qsc_qasm3/src/types.rs (Types)
  qsc_qasm3/src/ast_builder.rs
  qsc_qasm3/src/compiler.rs (Compiler)
  qsc_qasm3/src/runtime.rs (Runtime Features)
    Stefan
 
  fuzz/ (Testing)
    Mine, Stefan
 
  pip/src/ (Interop)
  qsc/src/
  qsc_codegen/src/
    Mine, Stefan
 
  .github/fuzz (GitHub pipeline)
    Ian, Stefan

# Tracking items

## Lexing

- [x] Basic raw tokens
- [x] Basic cooked tokens
- [x] Cook pragma and annotation

## Parsing

- [x] Version #2148
- [x] Include #2148
- [x] pragma #2148
- [x] annotation #2148
- [x] Statements -
  - [x] alias #2191
  - [x] array concatenation #2191
  - [x] assignment #2191
  - [x] barrier #2200
  - [x] box #2200
  - [x] break #2191
  - [x] cal #2200
  - [x] calibration grammar #2200
  - [x] classical decl
    - [x] Bit, int, uint, float, angle, bool, duration, stretch #2165
    - [x] Arrays #2180
    - [x] Measurement result #2180
  - [x] const decls #2165
  - [x] continue #2191
  - [x] def #2176
  - [x] defcal #2200
  - [x] delay #2200
  - [x] end #2191
  - [x] expression statement #2191
  - [x] extern #2176
  - [x] for loops #2191
  - [x] gate call #2200
  - [x] gate statement #2176
  - [x] gphase #2200
  - [x] if statement #2191
  - [x] IO decls #2165 
  - [x] measure arrow #2200
  - [x] old style decl #2176
  - [x] quantum decl #2160
  - [x] reset #2200
  - [x] return #2176
  - [x] switch statements #2178
  - [x] while loops #2191
  - [x] gate def #2176
- [x] Precedence #2166 
- [x] Expressions -
  - [x] ident #2148
  - [x] unary #2166 
  - [x] binary #2166 
  - [x] literal #2160
  - [x] timing literal #2200
  - [x] function call #2166 
  - [x] cast #2166 
  - [x] assignment #2216
  - [x] assignment op #2216
- [x] Check for cyclic includes #2278
- [x] Document parsing functions with grammar defs and spec notes  #2222
- [x] Remove measurement exprs from const decl parsing. Update notes on the grammar docs as this is a spec bug. #2272

## Lowering

- [x] Constant Evaluation for array sizes and type widths #2246
- [x] Statements -
  - [x] alias #2221 
  - [x] assignment 
    - [x] simple #2221 
    - [x] indexed  #2221 
  - [x] barrier #2246
  - [x] box (unimplemented)
  - [x] break #2272
  - [x] cal (unimplemented)
  - [x] calibration grammar (unimplemented)
  - [x] classical decls with default and literal initializers
    - [x] Bit, int, uint, float, angle, bool, duration, stretch #2221 
    - [x] creg #2246
    - [x] qreg #2246
    - [x] Bitarrays #2246
    - [x] Arrays (unimplemented)
    - [x] Measurement result #2246
  - [x] classical decls with all expressions and casting #2214
    - [x] Bit, int, uint, float, angle, bool, duration, stretch #2221 
    - [x] creg #2246
      - [x]  tests see bit[] tests for cases #2246
    - [x] qreg #2246
      - [x] tests - see qubit[] tests for cases #2246
    - [x] Bitarrays #2246
    - [x] Arrays (unimplemented)
    - [x] Measurement result #2246
  - [x] continue #2272
  - [x] def #2254
  - [x] defcal (unimplemented)
  - [x] delay (unimplemented)
  - [x] end #2271
  - [x] expression statement #2221 
  - [x] extern #2274
  - [x] for loops #2232
  - [x] gate call #2246
    - [x] broadcast (unimplemented)
  - [x] gphase #2254
  - [x] if statement #2232
  - [x] IO decls
  - [x] measure arrow #2246
  - [x] old style decl #2246
  - [x] quantum decl #2246
  - [x] reset #2246
  - [x] return #2254
    - [x] add cast of return value to containing function return ty #2272
  - [x] switch statements #2232
  - [x] while loops #2232
  - [x] gate def #2254
  - [x] Return semantic error on anonymous blocks and switch statements if version is set to 3.0 as they were introduced in 3.1 #2232 
- [x] Expressions
  - [x] binary #2239
  - [x] unary #2239
  - [x] function call #2254
  - [x] literals
    - [x] bit, bool, float, int, uint #2221
    - [x] angle #2267
    - [x] complex #2268
    - [x] duration #2272
    - [x] cast (unimplemented)
- [x] Review ignored tests from old parser #2272
- [x] SimulatableIntrinsic support #2271 
  - [x] Error when not applied to a gate or def #2271

## Compiling

- [x] Only set `EntryPoint()` attr on operation when compiling in file mode
- [x] Statements -
  - [x] alias (unsupported) #2239
    - [x] array concatenation
  - [x] assignment #2239
  - [x] barrier #2246
  - [x] box (unimplemented)
    - [x] pragma impls (unimplemented)
  - [x] break (unsupported) #2272
  - [x] cal (unimplemented)
  - [x] calibration grammar (unimplemented)
  - [x] classical decls #2239
    - [x] Bit, int, float, angle, bool, duration, stretch #2239
    - [x] creg #2246
    - [x] qreg #2246
    - [x] Arrays (unimplemented)
    - [x] Measurement result #2246
  - [x] continue (unsupported) #2272
  - [x] def #2254
    - [x] simulatable intrinsic #2271 
  - [x] end #2271
  - [x] expression statement #2239
  - [x] extern (unimplemented)
  - [x] for loops #2254
  - [x] gate call #2246
  - [x] gphase #2254
  - [x] if statement #2254
  - [x] IO decls #2239
  - [x] measure arrow #2246
  - [x] old style decl #2246
  - [x] quantum decl #2246
  - [x] reset #2246
  - [x] return #2254
  - [x] switch statements #2254
  - [x] while loops #2239
  - [x] gate def #2254
    - [x] simulatable intrinsic #2272
    - [x] gate scopes capture const variables in the outside scopes. We compiled gates to lambda operations to get this behavior. But this breaks simulatable intrinsics. Now that we have const evaluation, we can overcome this. #2270 
  - [x] delay (unimplemented)
  - [x] barrier #2246
- [x] Expressions
  - [x] binary #2239
  - [x] unary #2239
  - [x] function call #2254
  - [x] literals
    - [x] bit, bool, float, int, uint #2221
    - [x] angle #2267
    - [x] complex #2268
    - [x] duration (unsupported) #2239
    - [x] cast (unimplemented)
- [x] Move runtime functions in new dependency package so that they are parsed as a unit and imported by user code. This enables spans and a file name for runtime calls that get generated.

## Fit and Finish

- [x] Move Q# related semantic errors from Lowerer to Compiler
- [x] Add better error messages to const evaluator. We can pass `&mut Lowerer` as an argument so that it can push semantic errors. #2277
- [x] Clean up TODO items #2285
- [x] Review compiler error messages #2286
- [x] Review parser error messages #2286
- [x] Review lexer error messages #2286
- [x] Delete old compiler internals that are no longer used and remove stale refs #2284
- [x] Benchmark parser #2288
- [ ] Remove dead_code

## Postponed

- [ ] Update ast_builder calls with module names instead of old ns names 
- [ ] `bit x = 0; bit y = ~x;` is valid QASM3, but it gets compiled to `let x = Zero; let y = ~~~x;` which is invalid Q# code, since the `Result` type in Q# doesn't support unary bitwise negation. This is the same for all other operations that bit should support, Q# Result only supports equality.
- [ ] We are currently not enforcing that qasm3 `uint` types must be positive, since they are compiled to Q# `Int` which are signed. We might need a `UInt` type similar to the `Angle` type introduced in #2267 to be able to enforce this constraint.
- [ ] uint 63 bitness with 64 bigint. Separate ops for uint?
- [ ] Lower `Ident` as `Rc<Symbol>` and not as `SymbolId` to avoid an unnecessary `SymbolTable` lookup? In that way we don't even need to pass the symbol table to the compiler.
- [ ] Use the formal parameters' `SymbolId`s when creating the function type, so that we can give the user a better error message when they pass an argument that fails implicit casting to a function. There is a catch: we need to insert the function symbol into the symbol table before pushing the scope where the formal parameters live, but we need the `SymbolId`s of the formal parameters to construct the function symbol.
- [ ] Create qubit cleanup calls which can be conditionally added to end of program during compile, not done for fragments #2281
- [ ] Every time we allow an arithmetic lint in `compiler/qsc_qasm3/src/semantic/ast/const_eval.rs` we need to issue the same lint in Q#.
- [ ] Double check cast from `int` to `uint` and cast from `uint` to `int` in const evaluator. Both types are represented as `i64` so there is nothing to do, which is confusing.
- [ ] Profile the usage of `Box<[Box<T>]>` with iai-callgrind in a large OpenQASM3 sample to verify that is actually faster than using Vec<T>. Even though Box<T> uses less stack space, it reduces cache locality, because now you need to be jumping around in memory to read contiguous elements of a list. I suspect that what we really want to use if `Box<[T]>`.
- [ ] Input decls are pushed to the symbol table, but should not be in the stmts list. This may be an issue for tooling as there isn't a way to have a forward declared variable in Q#. `compiler/qsc_qasm3/src/compiler.rs::QasmCompiler::compile_output_decl_stmt`.
- [ ] See the comment on `compiler/qsc_qasm3/src/compiler.rs::QasmCompiler::create_entry_item` saying "This can create a collision on multiple compiles when interactive. We also have issues with the new entry point inference logic."
- [ ] Minimize number of errors reported in the compiler. There are many situations where we report duplicated or unnecessary extra errors that could make harder for the user to fix the actual problem. For example, see the test named `fuzzer_issue_2294`. The one valuable error there is the one saying "undefined symbol _`.


compiler/qsc_qasm3/src/io.rs
```diff
@@ -1,37 +1,46 @@
...
         let path = std::fs::canonicalize(path).map_err(|e| {
-            crate::Error(crate::ErrorKind::IO(format!(
+            Error(ErrorKind::IO(format!(
                 "Could not resolve include file path: {e}"
             )))
         })?;
+
+        self.ctx().check_include_errors(&path)?;
+
         match std::fs::read_to_string(&path) {
```
billti, line 35: We've generally avoided std::fs operations as they don't compile to wasm and run on the web well. Is the intent this should ultimately work in the VS Code extension and/or vscode.dev sessions also?

billti: (I guess this is Trait default impl that can be overridden, and I believe the many of the std::fs impls for Wasm only throw if called, but still odd to see a default trait impl using APIs we know will fail in VS Code and the web)

idavis: See `#[cfg(feature = "fs")]` at the top of the fn def. There are other impls for non-fs features for when we don't have access

compiler/qsc/src/compile.rs
```diff
@@ -37,6 +37,9 @@ pub enum ErrorKind {
     #[error("Cycle in dependency graph")]
     /// `DependencyCycle` occurs when there is a cycle in the dependency graph.
     DependencyCycle,
+
+    #[diagnostic(transparent)]
+    OpenQasm(#[from] crate::qasm::error::Error),
```
minestarks, line 42: Is this used? All tests pass if I comment it out.

compiler/qsc_codegen/src/qsharp.rs
```diff
@@ -61,6 +61,22 @@ pub fn write_package_string(package: &Package) -> String {
     format_str(&s)
 }
 
+#[must_use]
+pub fn write_item_string(item: &Item) -> String {
```
minestarks, line 65: Where is this used?

compiler/qsc_qasm3/src/parser/completion.rs
```diff
@@ -0,0 +1,37 @@
...
+/// in the given notebook cell (using the fragments parser).
+///
+/// This is useful for providing completions in an editor.
+#[must_use]
+pub fn possible_words_at_offset_in_fragments(input: &str, at_offset: u32) -> WordKinds {
+    let mut collector = ValidWordCollector::new(at_offset);
+    let mut scanner = ParserContext::with_word_collector(input, &mut collector);
+    let _ = prgm::parse(&mut scanner);
+    collector.into_words()
+}
```
minestarks, line 37: Exposing this seems unnecessary, unless we plan to use a separate parser to parse fragments (i.e. notebook cells) than we do for .qasm files.

compiler/qsc_qasm3/src/tests/declaration/integer.rs
```diff
@@ -331,61 +295,18 @@ fn const_explicit_bitness_int_decl() -> miette::Result<(), Vec<Report>> {
...
     ";
 
-    let Err(errors) = compile_qasm_stmt_to_qsharp(source) else {
-        panic!("Expected error");
-    };
+    let qsharp = compile_qasm_stmt_to_qsharp(source)?;
     expect![
-        r#"Cannot assign a value of Float(None, True) type to a classical variable of Int(None, False) type."#
+        r#"
+        mutable x = Microsoft.Quantum.Math.Truncate(-42.);
```
billti, line 307: Not `Std.Math`?

compiler/qsc_qasm3/src/parser/completion/collector.rs
```diff
@@ -0,0 +1,186 @@
...
+/// EOF:
+/// `def Foo(|int[64] x , int[64] y) : {}|`
+///  - at `EOF`
+///
+fn cursor_at_token(cursor_offset: u32, next_token: Token, scanner_offset: u32) -> bool {
+    match next_token.kind {
+        // Order matters here as the cases overlap.
+        TokenKind::Identifier
+        | TokenKind::Keyword(_)
+        | TokenKind::ClosedBinOp(ClosedBinOp::AmpAmp | ClosedBinOp::BarBar)
```
minestarks, line 166: This case doesn't make sense for qasm. The reason it was special cased for Q# is that `and` and `or` are words, but they happened to be categorized under the `ClosedBinOp` enum. They're not words in qasm, they're punctuation

minestarks: The intent of this match is to match all _words_ in the language. Are there are other word tokens that aren't covered by `TokenKind::Identifier` and `TokenKind::Keyword`? (What about `TokenKind::GPhase` etc? Also, why are those not keywords anyway?)

compiler/qsc_codegen/src/qsharp.rs
idavis: it was used in tests, looks like no longer used

compiler/qsc_qasm3/src/tests/declaration/integer.rs
idavis: First bullet under Postponed heading.

compiler/qsc_qasm3/src/tests/output.rs
```diff
@@ -270,64 +258,64 @@ c2[2] = measure q[4];
...
-  call void @__quantum__rt__result_record_output(%Result* inttoptr (i64 3 to %Result*), i8* null)
-  call void @__quantum__rt__result_record_output(%Result* inttoptr (i64 2 to %Result*), i8* null)
-  call void @__quantum__rt__array_record_output(i64 2, i8* null)
-  call void @__quantum__rt__result_record_output(%Result* inttoptr (i64 1 to %Result*), i8* null)
-  call void @__quantum__rt__result_record_output(%Result* inttoptr (i64 0 to %Result*), i8* null)
-  ret void
-}
+        define void @ENTRYPOINT__main() #0 {
+        block_0:
+          call void @__quantum__qis__rz__body(double 0.4999999999999997, %Qubit* inttoptr (i64 0 to %Qubit*))
```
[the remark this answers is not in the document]
idavis: Side effect of how qasm angles work. It is all based on fragment of 0-2*pi

compiler/qsc_qasm3/src/lex/cooked.rs
```diff
@@ -0,0 +1,774 @@
...
+
+    // Builtin identifiers and operations
+    GPhase,
+    Inv,
+    Pow,
+    Ctrl,
+    NegCtrl,
+    Dim,
+    DurationOf,
+    Measure,
```
minestarks, line 103: So are these keywords then? Or builtin identifiers? If I try to use these words elsewhere in the code (e.g. as a variable name), should that be a parse error?

I'm a bit confused overall by the organization of tokens here. IMO tokens categories should be pretty basic (keywords, identifiers etc) , and then builtin identifiers should be separated out during parsing. 

Some keywords are organized under `TokenKind::Keyword` and some (these, `annotation`, `pragma`) are sitting out in this enum by themselves. Why? What's the distinction?

compiler/qsc_qasm3/src/lex/cooked.rs
```diff
@@ -0,0 +1,774 @@
...
+            | Error::Unknown(_, s) => s,
+        }
+    }
+}
+
+/// A token kind.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Sequence)]
+pub enum TokenKind {
+    Annotation,
+    Pragma,
```
minestarks, line 91: Doc comments for these would be good showing what they look like in source code (just as other `TokenKind`s have)

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[repr(transparent)]
+    #[derive(Default, PartialEq, Debug, Clone, Copy)]
+    pub struct WordKinds: u128 {
+
+        //
+        // Begin names.
+        //
+
+        /// A path in an expression. Callables, UDT constructors, local variables.
+        const PathExpr = 1 << 0;
```
minestarks, line 36: The doc comment doesn't sound convincing - do we have really have dotted names of callables, udt constructors etc? Or should this be customized for qasm?

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        /// A path segment that follows a `.`
+        /// A more specific name kind can be inferred from a recovered AST.
+        const PathSegment = 1 << 1;
+        /// A primitive class.
+        const PrimitiveClass = 1 << 2;
```
minestarks, line 42: This seems unused/irrelevant in qasm.

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
```diff
@@ -0,0 +1,175 @@
...
+
+/// A path (see: [`Predictions`])
+///
+/// Maps to a subset of values in [`Predictions`], but an enum
+/// for friendly consumption.
+#[derive(Debug, Clone, Copy)]
+pub enum PathKind {
+    /// A path in an expression. Callables, UDT constructors, local variables.
+    Expr,
+}
```
minestarks, line 175: Some of this doesn't make sense to me for QASM, but I think I'm just not going to worry about it until I see their usage from the language service. It all comes down to _how_ this stuff actually gets used in the language service implementation. Once we get there, we may find that these complex types are just not necessary.

compiler/qsc_qasm3/src/lib.rs
```diff
@@ -54,319 +53,29 @@ impl Error {
 /// - Semantic errors
 /// - IO errors
 #[derive(Clone, Debug, Diagnostic, Eq, Error, PartialEq)]
+#[error(transparent)]
 pub enum ErrorKind {
```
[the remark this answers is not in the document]
orpuente-MS: We had an `#[allow(dead_code)]` attribute left. Nice catch. However, it still doesn't catch the ones you are referring to. I wonder why.

compiler/qsc_codegen/src/qsharp.rs
orpuente-MS: Thanks. Removed in the latest PR

compiler/qsc_qasm3/src/parser/completion.rs
orpuente-MS: Removed

compiler/qsc_qasm3/src/io.rs
minestarks: Why not use the generic implementation that already exists in `pip/src/interop.rs` ? An implementation for `FileSystem` already exists over `std::fs` (see `StdFs` in `compiler/qsc_project/src/fs.rs`) . You don't need two impls for `SourceResolver`.

compiler/qsc_qasm3/src/io.rs
```diff
@@ -1,37 +1,46 @@
...
-            Ok(source) => Ok((path, source)),
-            Err(_) => Err(crate::Error(crate::ErrorKind::NotFound(format!(
+            Ok(source) => {
+                self.ctx().add_path_to_include_graph(path.clone());
+                Ok((path, source))
+            }
+            Err(_) => Err(Error(ErrorKind::NotFound(format!(
```
minestarks, line 40: Should this be "not found"? Looks like you'd get here in the case of a read error. In any case, should we be wrapping/propagating the original error message from `read_to_string` here?

compiler/qsc_qasm3/src/io.rs
minestarks: I deleted the default impl for this method and the `cfg` checks, and the tests all pass. Perhaps it was unused? The unit tests seem to use an in-memory implementation, not the real filesystem one. (Also note the signatures didn't quite match - one `resolve` uses `&mut self` and a custom `Error` type, the other one doesn't)

pip/src/interop.rs
```diff
@@ -55,12 +55,20 @@ impl<T> SourceResolver for ImportResolver<T>
...
-    fn resolve<P>(&self, path: P) -> miette::Result<(PathBuf, String)>
+    fn ctx(&mut self) -> &mut SourceResolverContext {
+        &mut self.ctx
+    }
+
+    fn resolve<P>(&mut self, path: P) -> miette::Result<(PathBuf, String), qsc::qasm::io::Error>
     where
         P: AsRef<Path>,
     {
         let path = self.path.join(path);
```
minestarks, line 66: Please use `self.fs.join` to avoid dumb issues path separators, and to resolve away `.` and `..`s in paths.

pip/src/interop.rs
```diff
@@ -55,12 +55,20 @@ impl<T> SourceResolver for ImportResolver<T>
 where
     T: FileSystem,
 {
-    fn resolve<P>(&self, path: P) -> miette::Result<(PathBuf, String)>
+    fn ctx(&mut self) -> &mut SourceResolverContext {
+        &mut self.ctx
+    }
+
+    fn resolve<P>(&mut self, path: P) -> miette::Result<(PathBuf, String), qsc::qasm::io::Error>
```
minestarks, line 62: How much do you care about symlinks? Since we don't canonicalize paths in this impl, we won't detect cycles involving symlinks.

FWIW, not "handling" symlinks is the correct behavior to me. There are an infinite number of dumb things you can do with paths so let's not try chasing them. However, we may want to have some kind of depth limit in the paths to prevent the user from shooting themselves in the foot.

pip/tests-integration/interop_qiskit/test_re.py
minestarks, line 1: Why would these change?

minestarks (approved):
I only reviewed the files in my recommended list - lmk if you want me to look at anything else

compiler/qsc_qasm3/src/lex/cooked.rs
orpuente-MS: I was just trying to follow the organization they have in the official grammar. They have a section with these lines. But I agree with what you say. And it doesn't make much sense for barrier, measure, and reset to be treated as "builtin identifiers" when return, break, and continue are keywords.

```
/* Builtin identifiers and operations */

GPHASE: 'gphase';
INV: 'inv';
POW: 'pow';
CTRL: 'ctrl';
NEGCTRL: 'negctrl';

DIM: '#dim';

DURATIONOF: 'durationof';

DELAY: 'delay';
RESET: 'reset';
MEASURE: 'measure';
BARRIER: 'barrier';

BooleanLiteral: 'true' | 'false';
```

orpuente-MS: I will convert `Inv`, `Pow`, `Ctrl`, `NegCtrl`, `Dim`, and `Measure` into keywords

`GPhase` and `DurationOf` will remain as builtin identifiers, because they are a gate and a function.

compiler/qsc_qasm3/src/parser/completion/collector.rs
orpuente-MS: `GPhase` and `DurationOf` are not keywords because they are a gate and a function, and I want them to get the same treatment of gates and functions from the text editor instead of being highlighted as keywords. But I'm open to suggestions if this doesn't sound like a good idea.

pip/tests-integration/interop_qiskit/test_re.py
idavis: Angle domain constrictions

pip/src/interop.rs
idavis: https://github.com/microsoft/qsharp/pull/2303

compiler/qsc_qasm3/src/io.rs
idavis: https://github.com/microsoft/qsharp/pull/2303

compiler/qsc_qasm3/src/lib.rs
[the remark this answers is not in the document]
orpuente-MS: Removed the dead code.

#2301

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
orpuente-MS: We only have namespaced annotations. I guess we will want support completion for those.
Eg.:
```
@Microsoft.SimulatableIntrinsic
gate my_gate { ... }
```

compiler/qsc_qasm3/src/parser/completion/collector.rs
orpuente-MS: Fixed in #2305

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
orpuente-MS: Improved the comment in #2305

compiler/qsc_qasm3/src/parser/completion/word_kinds.rs
orpuente-MS: Removed in #2305

compiler/qsc/src/compile.rs
idavis: It is used in an incoming branch.

orpuente-MS (approved):

compiler/qsc_qasm3/src/io.rs
orpuente-MS: Nice catch. Fixed this in #2310

PR #2151, opened 2025-02-04
Copilot chat panel for VS Code

minestarks:
This adds a panel to VS Code that provides Copilot chat functionality. 

We define a set of client-side tools to be invoked by the model. These tools interface with the user's Azure Quantum workspace, providing a way to submit jobs, get information about the user's jobs and available targets etc.

The panel isn't yet enabled by default. It requires a special configuration to be defined in VS Code settings.


vscode/src/copilot/azqTools.ts
```diff
@@ -0,0 +1,455 @@
...
+  result: {
+    recentJobs: MinimizedJob[];
+    lastNJobs: number;
+    lastNDays: number;
+  };
+}> {
+  const workspace = await getConversationWorkspace(conversationState);
+
+  const recentJobs = (await getRecentJobs(workspace)).map((job) => {
+    return {
```
billti, line 167: Is this different filtering of properties that 'responseJobToJob' in 'workspaceActions.ts'?

vscode/src/copilot/copilot.ts
```diff
@@ -0,0 +1,251 @@
...
+  requestChatCompletion(
+    messages: QuantumChatMessage[],
+    onDelta: (delta: string) => void,
+  ): Promise<{
+    content?: string;
+    toolCalls?: ToolCall[];
+  }>;
+}
+
+function cleanContent(content: string | undefined) {
```
billti, line 229: Looks like this TODO is taken care of (line 233), no?

vscode/src/copilot/webview/copilot.tsx
```diff
@@ -0,0 +1,382 @@
...
+function loaded() {
+  console.log("Loaded");
+  restartChat([], undefined);
+}
+
+/**
+ * Handles updates to the copilot state.
+ */
+function onMessage(event: MessageEvent<CopilotUpdate>) {
+  console.log("Received message: ", event.data);
```
billti, line 354: Do we want to console.log all messages received by default?

vscode/src/copilot/webview/hlsj-qsharp.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Currently Highlight.js for Q# points to https://github.com/fedonman/highlightjs-qsharp/blob/main/src/languages/qsharp.js
+
+export default function (hljs: any) {
+  return {
+    name: "Q#",
+    case_insensitive: true,
+    keywords: {
+      keyword:
+        "namespace open operation function body adjoint controlled self auto if elif else return fail for in new use borrow repeat until fixup",
```
billti, line 12: This looks a bit out of date: struct, export, import...?

vscode/src/copilot/webviewViewProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+      CopilotWebviewViewProvider.viewType,
+      provider,
+      {
+        webviewOptions: { retainContextWhenHidden: true },
+      },
+    ),
+  );
+}
+
+class CopilotWebviewViewProvider implements WebviewViewProvider {
```
billti, line 29: Should this also have a WebviewPanelSerializer? Else do you lose all chat history when you restart VS Code or reload the window? (See WebviewPanel.ts for the existing webviews)

vscode/src/programConfig.ts
```diff
@@ -66,6 +66,24 @@ export function getActiveQSharpDocumentUri(): vscode.Uri | undefined {
...
+      success: false,
+      errorMsg: "Unable to find a visible Q# document",
+    };
+  }
+
+  return await getProgramForDocument(docUri);
+}
+
+export function getVisibleQSharpDocumentUri(): vscode.Uri | undefined {
+  return vscode.window.visibleTextEditors.find((editor) =>
```
billti, line 82: If there are several visible, will this return the one that has/had focus? Or just the first one it finds?

vscode/src/copilot/azqTools.ts
minestarks: That's right. I'll double check this code, but my intention is that `GetJobs` returns less details than `GetJob`  , most importantly by dropping `errorData` . `errorData` can get very large and fills up input tokens very quickly if we include it in `GetJobs` for every job. But we do want to include it when we get details for a single job using `GetJob`.

Meanwhile `responseJobToJob` drops properties we'll never read such as `containerUri` straight off the bat. They just confuse the copilot.

vscode/src/copilot/webview/hlsj-qsharp.ts
minestarks: I think I took from the original branch without looking 😬  I think you have more context here than I do, I'll ask you later about where we got this from and how exactly we want to consume highlight.js

vscode/src/copilot/webviewViewProvider.ts
minestarks: You do lose chat history... I meant that to be a "feature" but we can revisit. I figured most often conversations are pretty short, and intended to do a specific task related to _current_ state of your workspace. So I figured that having a conversation history that's from days ago, including stale details about your workspace, will do more harm than help.

vscode/src/programConfig.ts
minestarks: Pretty sure it's the first one it finds. Since this is invoked when the user is interacting with the copilot panel, it's actually guaranteed that no Q# editors will have focus. I'll play ground with the multiple-window scenario and see if the behavior makes sense.

The other thought I had is to just give the user an indicator of what the active editor window is, in the copilot panel itself. Like just say "BellState.qs" somewhere in small text. Similar to how the GH Copilot in VS Code shows you which file(s) are in context.

vscode/src/copilot/webview/copilot.tsx
minestarks: Nope I should remove this.

vscode/src/copilot/webview/hlsj-qsharp.ts
minestarks: I ended up rewriting the higlight.js language definition. Pretty limited, but it'll do.

vscode/src/azure/workspaceActions.ts
```diff
@@ -499,45 +524,53 @@
...
-    prompt: "Job name",
-    value: new Date().toISOString(),
-  });
-  if (!jobName) return; // TODO: Log a telemetry event for this?
+  if (!jobName) {
+    jobName = await vscode.window.showInputBox({
+      prompt: "Job name",
+      value: new Date().toISOString(),
+    });
+    if (!jobName) return; // TODO: Log a telemetry event for this?
```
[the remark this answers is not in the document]
minestarks: Not new in this PR

vscode/src/copilot/azqChatService.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  async requestChatCompletion(
+    messages: QuantumChatMessage[],
+    handleResponseDelta: (delta: string) => void,
+  ): Promise<{
+    content?: string;
+    toolCalls?: ToolCall[];
+  }> {
+    const token = this.msaAuth ? await this.getMsaChatSession() : "XXXX";
```
DmitryVasilevsky, line 53: Is it needed?

vscode/src/azure/treeView.ts
```diff
@@ -131,6 +131,8 @@ export type Job = {
     | "Finishing"
     | "Cancelled";
   outputDataUri?: string;
+  count: number;
+  shots: number;
```
billti, line 135: Are both of these always present, or mutually exclusive? If not both always present, one or both should be marked optional.

vscode/src/telemetry.ts
```diff
@@ -257,6 +259,26 @@ type EventTypes = {
...
+    };
+    measurements: {
+      conversationMessages: number; // includes the one just being submitted
+    };
+  };
+  [EventType.ChatTurnEnd]: {
+    properties: {
+      associationId: string;
+      flowStatus: UserFlowStatus;
+      toolCalls: string; // JSON number array
```
billti, line 276: I'm not sure what to make of this comment. What is this string going to contain?

vscode/src/copilot/webview/copilot.css
```diff
@@ -0,0 +1,177 @@
+/* Copyright (c) Microsoft Corporation.
+   Licensed under the MIT License. */
+
+body {
+  padding: 12px;
+  height: 100%;
+  /*  match the native vscode chat style */
+  font-family:
+    Segoe WPC,
+    Segoe UI,
```
billti, line 10: I believe these font names should be in double quotes if they include spaces. (Or at least that's common practice even if it isn't required).

vscode/src/copilot/webview/copilot.css
```diff
@@ -0,0 +1,177 @@
...
+
+.left-message-row {
+  justify-content: flex-start;
+}
+
+.user-message,
+.assistant-message {
+  padding: 0 16px;
+  margin-bottom: 16px;
+  border: 1px solid #ccc;
```
billti, line 98: Some colors (like 'gray') are theme agnostic, but is this very light color correct in both dark and light themes?

vscode/src/copilot/webview/copilot.tsx
```diff
@@ -0,0 +1,444 @@
...
+  ChatElement,
+  CopilotCommand,
+  CopilotUpdate,
+  HistogramData,
+  Status,
+} from "../shared";
+import "./copilot.css";
+// Set up the Markdown renderer with KaTeX support
+import mk from "@vscode/markdown-it-katex";
+import hljs from "highlight.js";
```
billti, line 19: I see the `./out/copilot/webview/copilot.js` file is now one of the largest in the extension. I quick browse of it would seem to indicate that it's due to the syntax highlighting on `highlight.js` for all the languages we don't really care about. Do you know if we can limit the languages in pulls in the syntax for (e.g. limit to just Q#, Python,... maybe Rust and C/C++)? That might save a couple of MB.

vscode/src/copilot/webviewViewProvider.ts
billti: Hmmmm, even if I just drag the Copilot to the other panel it loses all history. When implementing our other webviews I saw a number actions which effectively must close/reopen the view and thus lose all context without serialization, so I think we might need this.

vscode/src/telemetry.ts
minestarks: Yes, wtf was I doing here? I think this was meant to be a list of the _names_ of the tools that were invoked, but maybe I tried to do something clever instead.

`string[]` is not a type we can send, so it's going to be a JSON array that we can easily decode in the query - e.g. `["getJobs", "getWorkspace"]` etc

billti (approved):

vscode/src/azure/workspaceActions.ts
```diff
@@ -442,6 +442,31 @@ export async function queryWorkspace(workspace: WorkspaceConnection) {
...
+// Drop properties we don't care to track and clean up the types.
+function responseJobToJob(job: ResponseTypes.Job): Job {
+  return {
+    id: job.id,
+    name: job.name,
+    target: job.target,
+    status: job.status,
+    outputDataUri: job.outputDataUri,
+    count:
+      typeof job.inputParams.count === "string"
```
idavis, line 454: what happens if `parseInt` fails?

minestarks: It returns `NaN`. Is that handled further down the line? I'm not sure. I'm also not sure what happens if the type is not a string nor a number. I'll go ahead and make the sanitization a bit stricter here.

vscode/src/copilot/azqTools.ts
ScottCarda-MS: Could be good to list as an explicit comment that `errorData` is specifically something that is being avoided for handling multiple jobs here.

vscode/src/copilot/openAiChatService.ts
```diff
@@ -0,0 +1,353 @@
...
+          handleResponseDelta(contentDelta);
+          content += contentDelta;
+        }
+
+        if (choice.delta.tool_calls && choice.delta.tool_calls.length > 0) {
+          for (const toolCall of choice.delta.tool_calls) {
+            // This handling is a bit weird because the tool calls are also streamed in chunks.
+            // A single tool call can be split across multiple messages.
+            // As far as we can tell, the `arguments` string is sent in chunks but the
+            // other fields are not.
```
ScottCarda-MS, line 94: Is this a behavior seen working against the OpenAI API, or specific to our backend?

vscode/src/copilot/webviewViewProvider.ts
ScottCarda-MS: As a counter-example, when I use a chatbot in my hobby projects, it can be days in between when I have time to work on them, and I appreciate that the history of the conversation is preserved so I can pick up where I left off.

ScottCarda-MS (approved):

vscode/src/copilot/openAiChatService.ts
minestarks: This is specific to the OpenAI API! Our own Azure Quantum backend won't stream tool calls like this - it will return them all at once at the end.

vscode/src/programConfig.ts
minestarks: I want to clarify this UX a bit, but it's going to be a follow up PR.

vscode/src/copilot/azqChatService.ts
minestarks: Removed the XXXX header when auth is not needed.

vscode/src/azure/treeView.ts
minestarks: In practice both were always present but I think it's because I only ever test where we've submitted jobs from VS Code where we filled in both fields. I don't think we should trust these to be there for jobs other clients may have submitted.

I changed these to be optional and just overall improved the handling to better sanitize the data coming from the service.

vscode/src/copilot/webview/copilot.css
minestarks: This looked fine to my eyes in both modes, but I'm making a note to go back and refine the colors after I merge the PR

vscode/src/telemetry.ts
minestarks: Oh it is a string array, the comment was just wrong. Fixed the comment

PR #2159, opened 2025-02-05
Update devcontainer to Ubuntu noble base image

swernli:
This updates the repo's devcontainer to use a base image with a newer Ubuntu version making it more compatible out of the box with our build tooling requirements and installable published packages.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,8 +1,8 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/cpp
+// README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
 {
 	"name": "qsharp",
-	"image": "mcr.microsoft.com/devcontainers/cpp",
+	"image": "mcr.microsoft.com/devcontainers/base:noble",
 	"features": {
 		"ghcr.io/devcontainers/features/python:1": {},
 		"ghcr.io/devcontainers/features/node:1": {


.devcontainer/devcontainer.json
```diff
@@ -1,8 +1,8 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/cpp
+// README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
 {
 	"name": "qsharp",
-	"image": "mcr.microsoft.com/devcontainers/cpp",
+	"image": "mcr.microsoft.com/devcontainers/base:noble",
```
swernli, line 5: Of note, this image doesn't come with playwright installed, which prevents running VS Code integration tests (it's needed to launch a headless browser instance). We could add with a post-create command `npx playwright install-deps` but it takes a significant amount of time to do the installation so would slow down container creation. Is it worth making that part of the default or leaving it as an optional dependency if a dev wants to run the integration tests? @billti, @minestarks, @idavis any preferences?

For what it's worth, the error message when the integration tests fails clearly states which command to run to install the missing component, so it's relatively easy for a dev to know what to do.

minestarks: Yeah best to leave it out of container creation. Takes too long.

I think we could add [`--with-deps`](https://playwright.dev/docs/browsers#install-system-dependencies) to the command we already run before running the tests ([here](https://github.com/microsoft/qsharp/blob/6f655eb4e6b914ce5e7a23b738c2b95ddfa06735/vscode/package.json#L517)) . But maybe we can add that another time. It may have side effects I'm not aware of, so let's not complicate things in this PR. I use devcontainers as my main environment so I'll keep an eye out for issues.

minestarks:
@swernli I don't know if it's a fluke, but the VS Code extensions under `customizations` didn't actually get installed when I created a codespace from this devcontainer. They do get installed when I create a codespace from `main`.

minestarks:
> @swernli I don't know if it's a fluke, but the VS Code extensions under `customizations` didn't actually get installed when I created a codespace from this devcontainer. They do get installed when I create a codespace from `main`.

I tried again and it worked this time. Works for me 🤷‍♀️

minestarks (approved):

idavis (approved):

PR #2174, opened 2025-02-11
Support code lenses on more callables

swernli:
This change adds support for code lenses on any callable that qualifies for them instead of just the entry point. In practice, this means any function or operation that doesn't take arguments can be used as the starting place for a run, debug, estimate, histogram, or (in the case of operations) circuit. This is paticularly helpful for unit tests identifed by the `@Test()` attribute, as it lets them be run or debug outside of the test infra with a single click.

### DIFF
--- a/compiler/qsc_circuit/src/operations.rs
+++ b/compiler/qsc_circuit/src/operations.rs
@@ -47,6 +47,11 @@ pub enum Error {
 #[must_use]
 pub fn qubit_param_info(item: &Item) -> Option<(Vec<u32>, u32)> {
     if let ItemKind::Callable(decl) = &item.kind {
+        if decl.input.ty == Ty::UNIT {
+            // Support no parameters by allocating 0 qubits.
+            return Some((vec![], 0));
+        }
+
         let (qubit_param_dimensions, total_num_qubits) = get_qubit_param_info(&decl.input.ty);
 
         if !qubit_param_dimensions.is_empty() {
--- a/compiler/qsc_circuit/src/operations.rs
+++ b/compiler/qsc_circuit/src/operations.rs
@@ -33,7 +33,7 @@ pub enum Error {
 /// If the item is not a callable, returns `None`.
 /// If the callable takes any non-qubit parameters, returns `None`.
 ///
-/// If the callable only takes qubit parameters or no parameters, (including qubit arrays):
+/// If the callable only takes qubit parameters (including qubit arrays) or no parameters:
 ///
 /// The first element of the return tuple is a vector,
 /// where each element corresponds to a parameter, and the
--- a/compiler/qsc_circuit/src/operations.rs
+++ b/compiler/qsc_circuit/src/operations.rs
@@ -33,7 +33,7 @@ pub enum Error {
 /// If the item is not a callable, returns `None`.
 /// If the callable takes any non-qubit parameters, returns `None`.
 ///
-/// If the callable only takes qubit parameters, (including qubit arrays):
+/// If the callable only takes qubit parameters or no parameters, (including qubit arrays):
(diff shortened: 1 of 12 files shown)


vscode/src/webviewPanel.ts
```diff
@@ -44,229 +44,233 @@ export function registerWebViewCommands(context: ExtensionContext) {
...
-        { associationId },
-        {},
-      );
-      const program = await getActiveProgram();
-      if (!program.success) {
-        throw new Error(program.errorMsg);
-      }
+    commands.registerCommand(
+      `${qsharpExtensionId}.showRe`,
+      async (expr?: string) => {
```
DmitryVasilevsky, line 49: Is this mostly formatting changes because you need to pass expr?

compiler/qsc_circuit/src/operations.rs
```diff
@@ -47,6 +47,11 @@ pub enum Error {
 #[must_use]
 pub fn qubit_param_info(item: &Item) -> Option<(Vec<u32>, u32)> {
     if let ItemKind::Callable(decl) = &item.kind {
+        if decl.input.ty == Ty::UNIT {
+            // Support no parameters by allocating 0 qubits.
+            return Some((vec![], 0));
```
DmitryVasilevsky, line 52: Return type is an Option. What's the difference between returning None and Some((vec![], 0))?

vscode/src/webviewPanel.ts
swernli: yeah, the only real change is receiving `expr` and then passing it down to the next layer (on line 216 in the new file). It's unfortunate that the formatting changed so drastically, because it makes the diff look like lot's of things changed when they really didn't.

compiler/qsc_circuit/src/operations.rs
swernli: If the callable takes parameters other than qubits, it will return `None` to indicate that we can't generate a circuit.

DmitryVasilevsky: Can you add a comment. This isn't very intuitive.

DmitryVasilevsky (approved):

compiler/qsc_circuit/src/operations.rs
```diff
@@ -33,7 +33,7 @@ pub enum Error {
 /// If the item is not a callable, returns `None`.
 /// If the callable takes any non-qubit parameters, returns `None`.
 ///
-/// If the callable only takes qubit parameters, (including qubit arrays):
+/// If the callable only takes qubit parameters or no parameters, (including qubit arrays):
```
minestarks, line 36:
```suggestion
/// If the callable only takes qubit parameters (including qubit arrays) or no parameters:
```

vscode/src/debugger/activate.ts
```diff
@@ -70,10 +70,32 @@ function registerCommands(context: vscode.ExtensionContext) {
...
+    vscode.commands.registerCommand(
+      `${qsharpExtensionId}.debugExpression`,
+      (expr: string) =>
+        startDebugging(undefined, {
+          name: "Debug Expression",
+          stopOnEntry: true,
+          entry: expr,
+        }),
+    ),
   );
```
minestarks, line 95: I think I would have just updated the existing `debugEditorContents` and `runEditorContentsWithCircuit` commands with this new parameter. You won't have to update the existing call sites if you make the second parameter optional. Just to avoid proliferating command definitions. These are too similar to the old ones. 

I'm also surprised because I thought commands had to be defined in `package.json` but I don't see that change.

minestarks (approved):

minestarks (approved):

PR #2183, opened 2025-02-14
npm/qsharp - Make `languageFeatures` an optional parameter again

minestarks:
Revert breaking change for QCOM. I don't recall the specific reason we had to make this required - I think it was an accident that snuck in when we made the `languageFeatures` a required method in OTHER method signatures in this file (#2078)

### DIFF
--- a/npm/qsharp/src/compiler/compiler.ts
+++ b/npm/qsharp/src/compiler/compiler.ts
@@ -94,7 +94,7 @@ export type ProgramConfig = (
       /** An array of source objects, each containing a name and contents. */
       sources: [string, string][];
       /** An array of language features to be opted in to in this compilation. */
-      languageFeatures: string[];
+      languageFeatures?: string[];
     }
   | {
       /** Sources from all resolved dependencies, along with their languageFeatures configuration */


billti:
The quoted PR where this broke _seems_ like it was a deliberate change... but I don't see where this revert would break anything. If you've tested and looks good, then that's good with me.

billti (approved):

minestarks:
Just going over the details for my own sake: There were a few other arguments also called `languageFeatures`, that were changed to be required arguments, and those were deliberate, but this one must have got lumped in by accident.  @idavis  and I had discussed and decided it was ok to make `languageFeatures` optional on the `getAst` and `getHir` methods ([discussed in this comment](https://github.com/microsoft/qsharp/pull/2078/files#r1905933993)) This was not breaking since QCOM doesn't use those methods.  What I missed in the final revision is that there was _another_ `languageFeatures` property on `ProgramConfig` that was made required, and _that_ was the breaking change. 😭 

It ended up being nbd (QCOM has long been fixed) - but I wanted to get this PR out for hygiene. It's just a better API to leave that property optional.

Issue #2184, opened 2025-02-15
Multi-target gates are rendered unclearly in ASCII art circuits


Morcifer:
This issue is a followup of issue https://github.com/microsoft/qsharp/issues/1476 and [this comment](https://github.com/microsoft/qsharp/pull/2126#discussion_r1931095791) in the PR that solved it.

Simple gates in the case of measurements use vertical lines to show which qubits are involved:
```
q_0    ── H ─────────── ● ──── M ──
                        │      ╘═══
q_1    ── H ──── X ──── X ─────────
```

But when rendering circuits where there is one gate that operates on multiple targets and there are no measurements, the rendering is somewhat unclear.

For example the circuit in the test `custom_intrinsic_mixed_args` in `compiler/qsc/src/interpret/circuit_tests.rs` renders as 
```
q_0    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_1    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_2    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_3    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_4    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_5    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_6    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_7    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_8    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
q_9    ─ AccountForEstimatesInternal([(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6)], 1) ──
```

while circuits such as:
```
use q0 = Qubit();
[...]

PR #2185, opened 2025-02-15
Multi-target gate circuit art improvement

Morcifer:
This PR has a proposed fix for issue https://github.com/microsoft/qsharp/issues/2184, but it mainly includes refactoring of the circuit-art code based on comments from PR https://github.com/microsoft/qsharp/pull/2126.

The only real functionality change in this code is that multi-target gates get extra lines that behave in the same way as measurement lines, in order to better convey the qubits involved.

For example, a circuit like this one:
```
use q0 = Qubit();
use q1 = Qubit();
Rxx(1.0, q0, q1);

use q2 = Qubit();
use q3 = Qubit();
Rxx(1.0, q2, q3);
```
instead of rendering as
```
q_0    ─ rxx(1.0000) ─
q_1    ─ rxx(1.0000) ─
q_2    ─ rxx(1.0000) ─
q_3    ─ rxx(1.0000) ─
```
will instead render as 
```
q_0    ─ rxx(1.0000) ─
              ┆
q_1    ─ rxx(1.0000) ─
q_2    ─ rxx(1.0000) ─
              ┆
q_3    ─ rxx(1.0000) ─
```

This is very similar to what we get if we measure the first qubit of each of these two-qubit gate, i.e.
```
q_0    ─ rxx(1.0000) ─── M ──
              ┆          ╘═══
q_1    ─ rxx(1.0000) ────────
q_2    ─ rxx(1.0000) ─── M ──
              ┆          ╘═══
q_3    ─ rxx(1.0000) ────────
```

On top of this change in rendering, the first few commits in this PR are actually just code refactoring. As such, it's probably best to review this PR commit by commit, for two reason - a. it's going to be much easier to follow, and b. any of these commits can be reverted if you think they make the code more complicated instead of making it easier to read.

The commits are as follows:
1. [a400655](https://github.com/microsoft/qsharp/pull/2185/commits/a4006557b4b0fcd0bacd14c5c1142eeecefbaa76) A rework of PR https://github.com/microsoft/qsharp/pull/2126, which, to be honest, is how I think I should have implemented it in the first place.
2. [c141799](https://github.com/microsoft/qsharp/pull/2185/commits/c14179907ed45ebd02e07199db85a401edaf61b1) A cleanup based on [this comment](https://github.com/microsoft/qsharp/pull/2126#pullrequestreview-2579485021). It makes the code significantly better, in my opinion.
3. [4012ee8](https://github.com/microsoft/qsharp/pull/2185/commits/4012ee8378f0080cf79b2220dead041ccc964edf) A small simplification to have the ColumnRenderer have a `new` method which is in charge of making sure that the column width is an odd number, and also a `Default` trait implementation for the case where we don't need a specific width.
4. [cfa7382](https://github.com/microsoft/qsharp/pull/2185/commits/cfa7382630b83ee0aefd64ee414ca6b5b600f2d2) This is the actual enhancement - it makes sure there is a classical line (same as what you get from a measurement) if there are two consecutive qubits for the same operation. If you think there is a better solution to make the rendering clearer, let me know. But of the ones I could think of, I found this one to be the least disruptive one (to both the existing code, and the existing circuits).

### DIFF
--- a/compiler/qsc_circuit/src/circuit.rs
+++ b/compiler/qsc_circuit/src/circuit.rs
@@ -110,6 +110,8 @@ enum Wire {
 }
 
 enum CircuitObject {
+    Blank,
+    Wire,
     WireCross,
     WireStart,
     DashedCross,
@@ -192,24 +194,22 @@ impl Row {
                 for column in 1..end_column {
                     let val = self.objects.get(&column);
                     let column_width = *column_widths.get(&column).unwrap_or(&MIN_COLUMN_WIDTH);
-                    if let Some(v) = val {
-                        s.write_str(&fmt_qubit_circuit_object(v, column_width))?;
-                    } else {
-                        s.write_str(&get_qubit_wire(column_width))?;
-                    }
+                    let object = val.unwrap_or(&CircuitObject::Wire);
+
+                    s.write_str(&fmt_qubit_circuit_object(object, column_width))?;
                 }
             }
             Wire::Classical { start_column } => {
                 for column in 0..end_column {
                     let val = self.objects.get(&column);
                     let column_width = *column_widths.get(&column).unwrap_or(&MIN_COLUMN_WIDTH);
-                    if let Some(v) = val {
-                        s.write_str(&fmt_classical_circuit_object(v, column_width))?;
-                    } else if start_column.map_or(false, |s| column > s) {
-                        s.write_str(&get_classical_wire(column_width))?;
-                    } else {
-                        s.write_str(&get_blank(column_width))?;
-                    }
+                    let object = match (val, start_column) {
+                        (Some(v), _) => v,
+                        (None, Some(s)) if column > *s => &CircuitObject::Wire,
+                        _ => &CircuitObject::Blank,
+                    };
+
+                    s.write_str(&fmt_classical_circuit_object(object, column_width))?;
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


compiler/qsc_circuit/src/circuit.rs
```diff
@@ -325,6 +330,26 @@ impl Display for Circuit {
...
+                let qubit = target.q_id;
+
+                if consequtive_qubits.contains(&qubit) {
+                    continue;
+                }
+
+                let next_qubit = qubit + 1;
+
+                // Check if the next qubit is also in this operation.
+                if operation.targets.iter().any(|t| t.q_id == next_qubit) {
```
Morcifer, line 347: Technically, in terms of pure complexity, it would make more sense to make a hashsets out of these ids, inside the loop of line 336.

However, given the overhead of creating hashsets, and moreover, the relatively slow performance hashsets in Rust (even FxHashSet, from my experience), I'm almost sure that this ends up faster for a small number of target qubits.

But if it's actually common for these gates to have a large number of targets, I can easily replace this bit of code and use a FxHashSet here, just let me know.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -181,35 +183,37 @@ impl Row {
     fn fmt(
         &self,
         f: &mut std::fmt::Formatter<'_>,
-        column_widths: &ColumnWidthsByColumn,
+        column_renderers: &ColumnWidthsByColumn,
         end_column: usize,
     ) -> std::fmt::Result {
         // Temporary string so we can trim whitespace at the end
+        let default_renderer = ColumnRenderer::default();
```
minestarks, line 190: The comment above this line got pushed up - it's supposed to refer to this line below this.

However I think you can just use `unwrap_or_default` below instead of declaring a default here. It's not going to incur an extra allocation or anything like that (since the struct is just an int anyway)

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -96,7 +96,7 @@ impl Config {
 }
 
 type ObjectsByColumn = FxHashMap<usize, CircuitObject>;
-type ColumnWidthsByColumn = FxHashMap<usize, usize>;
+type ColumnWidthsByColumn = FxHashMap<usize, ColumnRenderer>;
```
minestarks, line 99: I like how you organized this. I just think we can simplify the names.

```suggestion
type FullColumns = FxHashMap<usize, Column>;
```

That's what the new type is... it's just the column! No need to overthink it imo.

And now we can think of the hashmap as containing only the columns that have objects in them, so I suggested a name that reflects that.

compiler/qsc_circuit/src/circuit.rs
minestarks: Agree with the reasoning. It would have to be a custom intrinsic in Q# that takes a very large number of qubits, and I just don't expect to see that in practice.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -335,7 +360,15 @@ impl Display for Circuit {
 
             register_to_row.insert((q.id, None), rows.len() - 1);
 
-            for i in 0..q.num_children {
+            // If this qubit has no children, but it is in a multi-qubit operation with
+            // the next qubit, we add one classical row for clarity.
```
minestarks, line 364:
```suggestion
            // the next qubit, we add an empty row to make room for the vertical connector.
            // We can just use a classical wire type for this row since the wire won't actually be rendered.
```

I now see how you're exploiting the "classical" wire type to create an empty row. I think that's fine (no need to overengineer it and add a 3rd wire type) but it's subtle enough that I think it warrants an explanatory comment.

minestarks (changes requested):
Thank you for the fix and additional test cases. Just a few requests for readability.

minestarks:
Heads up to @ScottCarda-MS who I think has been working near this code.

compiler/qsc_circuit/src/circuit.rs
Morcifer: I'm not 100% sure I agree with this:

My thinking is that the "ByColumn" postfix is consistent with `ObjectsByColumn` in the row above, because both are a hashmap with the key being the column index. 

And the `ColumnRenderer` itself is a struct that takes circuit objects and turns them into ASCII art, so calling it a `Column` sounds a bit confusing to me. 

Not to mention how, if you agree with my point about `ByColumn` but don't agree with this point, then we'll call this hashmap `ColumnByColumn`, and that's just silly. :)

Having said that, I will of course rename both of these if you still think your suggested names are better.

compiler/qsc_circuit/src/circuit.rs
Morcifer: `unwrap_or_default` won't work, because `column_renderers.get(&column).unwrap()` is a &ColumnRenderer, while the default part of `unwrap_or_default` returns an actual ColumnRenderer. 

And doing something like `.unwrap_or(&ColumnRenderer::default())` also doesn't work because of `temporary value dropped while borrowed`.

Rust is fighting against me here. Can you think of another way to do it that won't end up moving (or needing to clone) the renderer we get from `column_renderers`?

minestarks: Oh, interesting. I would just add `#[derive(Clone, Copy)]` on the struct to stop "fighting" rust. It's really not a big deal cloning this object since it's simply a `usize`.

compiler/qsc_circuit/src/circuit.rs
Morcifer: Actually, I realize now that I can resolve most of this comment and the other open one pretty easily, it's an oversight on my side - this alias isn't necessary, because we don't actually need a hashmap for this. Let me try it out... 

You'll still have to decide between `Column` and `ColumnRenderer`, though.

minestarks: Okay, short answer: Totally agree with you on the bit where you say it would be inconsistent with `ObjectsByColumn` map right above this - I hadn't noticed that. Consistency trumps all, so let's stick with your version. _Maybe_ consider `ColumnWidth` instead of `ColumnRenderer`.

---

Long answer: humor me for a second while I try to explain what I originally had in mind. Naming discussions tend to get pretty philosophical.

- "column index": it's just a number and that's what the original hashmap index `usize` refers to.
- "column": that's the struct that describes the "physical" column in the diagram, so to speak. It contains attributes of the column, such as its width. 

Personally I'm biased against struct names like "Something-Doer" because I prefer structs to have names that describe what the thing is, not what it does.

I guess then the natural conclusion would have been to rename these to `ObjectsByColumnIndex` and `ColumnsByColumnIndex` ... and then rename all the other variables that refer to `column`s below to `column_index` ... and I mean... that _is_ getting into silly territory.

Morcifer: Yeah, that works - please look at the latest commit. Basically, since the column renderers are made by iterating over `0..end_column`, they can just be put in a vector. That means there's no need for this hashmap at all.

I can still rename the `ColumnRenderer` to `Column`, though then the enumeration in line 192 and 200 will be over `(column_index, column)` instead of the current `(column, renderer)`. Not sure what's clearer in this case.

compiler/qsc_circuit/src/circuit.rs
Morcifer: The solution to the `ColumnWidthsByColumn` naming in the latest commit ended up eliminating this problem as well as simplifying this code quite a lot (especially with clippy's suggestion of using `enumerate` rather than keep the for loop in this method).

Let me know what you think.

compiler/qsc_circuit/src/circuit.rs
Morcifer: I see I managed to miss the short-answer-long-answer comment you made a couple of days ago (that is - when last I loaded this page, I had 3 comments in a row, which made for a slightly more coherent conversation with myself. 😄).

You convinced me that `ColumnRenderer` isn't ideal. `ColumnWidth` bothers me a bit, though, because I find code like `column_width.fmt_on_..._wire` to be confusing. So, given that there is already a `Row` struct in this file, I tried `Column` to see what happens.

I find that it makes the meaning of `column` a bit ambiguous - sometimes it's this struct, and sometimes it's still the index (e.g. in `Wire::Classical { start_column }`). However, it's very clear from the context what meaning is used when, and mistakes are easily caught by the compiler because of the types. So I can definitely live with it, and I renamed it in the latest commit.

minestarks (approved):

ScottCarda-MS (approved):

minestarks:
Thanks @Morcifer !

PR #2187, opened 2025-02-18
Basic Circuit Editor for .qsc files

ScottCarda-MS:
This PR adds a custom editor view in the VS Code extension for files with the .qsc extension. The editor makes use of the same library code that is used to draw circuit diagrams elsewhere in the VS Code extension. The editor provides basic drag-and-drop editing of circuits with a simple gate set, including the ability to have controlled gates, and gates with arguments.
The circuit files are included in Q# projects and can be called into from Q# as if they were Q# operations. This is done by generating Q# files for circuit files when compiling.


npm/qsharp/ux/circuit-vis/panel.ts
```diff
@@ -0,0 +1,332 @@
...
+    gate: "Measure",
+    isMeasurement: true,
+    controls: [{ qId: 0, type: RegisterType.Qubit }],
+    targets: [{ qId: 0, type: RegisterType.Classical, cId: 0 }],
+  },
+  Reset: {
+    gate: "|0〉",
+    targets: [{ qId: 0, type: RegisterType.Qubit }],
+  },
+  ResetX: {
```
billti, line 325: I don't think `|1>` is ResetX. 0 and 1 are in the Z basis. I would have thought ResetX maps to `|+>`?

ScottCarda-MS: Oh good point. `ResetX` does mean something else. I've renamed this to `ResetOne`.

vscode/package.json
```diff
@@ -524,7 +524,19 @@
...
       }
-    }
+    },
+    "customEditors": [
+      {
+        "viewType": "qsharp-webview.circuit",
+        "displayName": "Quantum Circuit",
+        "selector": [
+          {
+            "filenamePattern": "*.qviz"
```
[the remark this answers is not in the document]
ScottCarda-MS: Changed to `.qsc` extension as per discussion. I set the icon to the same as the `.qs` files. I think we should come up with a unique icon for the circuits in the near future though.

npm/qsharp/src/shared/circuit.ts
```diff
@@ -3,15 +3,34 @@
...
+
 /**
  * Circuit to be visualized.
  */
 export interface Circuit {
+  version: string;
   /** Array of qubit resources. */
   qubits: Qubit[];
-  operations: Operation[];
+  operations: Operation[][];
```
minestarks, line 18: In general, I dislike array-of-array as a type... for one, multi-dimensional arrays are just annoying to code against. But also, there's usually an intermediate type in there you're hiding, that deserves its own type for clarity.

What does an `Operation[]` represent? A column, I assume? You can wrap that it its own object:

```ts
interface Column { 
    operations: Operation[];
}
```

This also makes the type more future-proof in case you ever want to add a new property to the column itself (like, what if columns could have annotations? I don't know, just making stuff up)

And then this property here can become
```ts
    columns: Column[];
```

npm/qsharp/src/shared/circuit.ts
```diff
@@ -52,15 +71,17 @@ export interface Operation {
   /** Formatted gate arguments to be displayed. */
   displayArgs?: string;
   /** Nested operations within this operation. */
-  children?: Operation[];
+  children?: Operation[][];
+  /** Number of columns to span. */
```
minestarks, line 75: Consider saying in the comment what the default is if this property is omitted. 1?

npm/qsharp/src/shared/circuit.ts
```diff
@@ -52,15 +71,17 @@ export interface Operation {
...
+  isConditional?: boolean;
   /** Whether gate is a controlled operation. */
-  isControlled: boolean;
+  isControlled?: boolean;
   /** Whether gate is an adjoint operation. */
-  isAdjoint: boolean;
+  isAdjoint?: boolean;
   /** Control registers the gate acts on. */
   controls?: Register[];
   /** Target registers the gate acts on. */
```
minestarks, line 87: For readability, can you put `targets` at the top below the other required properties? This is the most important property buried beneath all the optional ones.

npm/qsharp/ux/circuit-vis/panel.ts
billti: Actually, I realized last night I was wrong here. If a qubit starts in the `|0>` state, then applying an X does indeed put it into the `|1>` state. Applying an H puts it into what is commonly called the `|+>` state which is along the X basis of measurement, which I was getting confused with. Sorry!

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: done

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: done

PR #2195, opened 2025-02-20
Uniform superposition preparation moved to std

DmitryVasilevsky:
Moved uniform superposition preparation into Std.StatePreparaion (simplified version based on df chemistry)
Improved checks and updated doc comments
Added tests


library/src/tests/state_preparation.rs
```diff
@@ -254,3 +254,33 @@ fn check_preparation_doc_sample() {
...
+
+#[test]
+fn check_uniform_superposition_preparation_exhaustive() {
+    let out = test_expression_with_lib(
+        "Test.TestPrepareUniformSuperpositionExhaustive()",
+        STATE_PREPARATION_TEST_LIB,
+        &Value::Tuple(vec![].into()),
+    );
+
+    expect![[""]].assert_eq(&out);
```
orpuente-MS, line 285: I don't think we need this `expect![[""]]`, since we are not asserting any output from rust. I noticed you are already checking the state in Q# with `Fact(CheckAllZero(qs), "All qubits must be in zero state.");`, that should be enough.

library/src/tests/resources/src/state_preparation.qs
```diff
@@ -106,4 +106,23 @@ namespace Test {
...
+
+    operation TestPrepareUniformSuperpositionExhaustive() : Unit {
+        let NBits = 4;
+        use qs = Qubit[NBits];
+        for NStates in 1..2^NBits-1 {
+            PrepareUniformSuperposition(NStates, qs);
+            let t = Std.Math.Sqrt(1.0 / IntAsDouble(NStates));
+            Adjoint Std.StatePreparation.PreparePureStateD(Repeated(t, NStates), Reversed(qs));
+            Fact(CheckAllZero(qs), "All qubits must be in zero state.");
+            ResetAll(qs);
```
swernli, line 124: nitpick: since you have a `Fact` above that the qubits are all zero there is no need to reset them here.
```suggestion
```

library/src/tests/resources/src/state_preparation.qs
```diff
@@ -106,4 +106,23 @@ namespace Test {
...
+    operation TestPrepareUniformSuperposition(nStates : Int) : Unit {
+        use qs = Qubit[10];
+        PrepareUniformSuperposition(nStates, qs);
+        DumpMachine();
+        ResetAll(qs);
+    }
+
+    operation TestPrepareUniformSuperpositionExhaustive() : Unit {
+        let NBits = 4;
+        use qs = Qubit[NBits];
```
swernli, line 118: Since the intent of each iteration of the loop is start from qubits in the zero state and assert they return to the zero state, you could move the allocation into the loop and let the qubits go out of scope between each iteration. That will also have the minor benefit of preventing any accumulation of global phase across iterations which can help with later investigations.

swernli (approved):
I really like how much code this change gets out of the samples and libraries and into a more compact stdlib operation, thanks!

library/src/tests/state_preparation.rs
DmitryVasilevsky: Extra checks do not hurt, but you are right, there's no much to check here. I removed it.

library/src/tests/resources/src/state_preparation.qs
DmitryVasilevsky: Removed.

library/src/tests/resources/src/state_preparation.qs
DmitryVasilevsky: Makes sense. Some part of me tries to save on allocation/release of qubits, but that may be wrong intuition. More allocations and release of qubits may pave a way to qubit reuse optimizations.

PR #2201, opened 2025-02-25
Add Azure credits notification message.

swernli:
This adds a message when loaded saved workspaces on first extension activation alerting users the expiring Azure Quantum Credits program.

### DIFF
--- a/vscode/src/azure/commands.ts
+++ b/vscode/src/azure/commands.ts
@@ -41,6 +41,11 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
   // Add any previously saved workspaces
   const savedWorkspaces = await context.secrets.get(workspacesSecret);
   if (savedWorkspaces) {
+    // Temporary reminder message that Azure Quantum Credits will be deprecated.
+    vscode.window.showInformationMessage(
+      `Azure Quantum Credits will no longer be available after June 1st, 2025.`,
+    );
+
     log.debug("Loading workspaces: ", savedWorkspaces);
     const workspaces: WorkspaceConnection[] = JSON.parse(savedWorkspaces);
     for (const workspace of workspaces) {


vscode/src/azure/commands.ts
```diff
@@ -41,6 +41,11 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
   // Add any previously saved workspaces
   const savedWorkspaces = await context.secrets.get(workspacesSecret);
   if (savedWorkspaces) {
+    // Temporary reminder message that Azure Quantum Credits will be deprecated.
```
minestarks, line 44: Shouldn't we persist a setting to remember that the user has dismissed this once?

swernli: We could... though there are other extensions that don't bother with that for certain notifications. Given that we will remove this soon, I didn't think it was worth the settings work, but I could be convinced :).

billti: Given that it's our most valuable users that would be most annoyed by repeated popups (i.e. those that use our extension regularly _and_ connect to an Azure workspace), I'm inclined to think it's worth the effort.

Copilot generated the below snippet as a "common pattern", and looking at the docs for those APIs it seems legit and is hopefully that simple.

```ts
    const hasSeenMessage = context.globalState.get<boolean>('hasSeenMessage', false);

    if (!hasSeenMessage) {
        vscode.window.showInformationMessage("This is a message with a checkbox.", { modal: true }, "Don't show again")
            .then(selection => {
                if (selection === "Don't show again") {
                    context.globalState.update('hasSeenMessage', true);
                }
            });
    }
```

billti: (Not sure we'd need it to be modal though)

swernli: Added option to suppress message.

billti (approved):

PR #2202, opened 2025-02-26
Chemistry lib: readablility, lints, syntax, tests

DmitryVasilevsky:
This change updates chemistry library:
- Removes old sysntax such as a! and [x, size=5]
- Reduced use of w/= where possible
- Makes long argument lists more readable
- Fixes error in comments
- Updates pairs of double representing a complex number to use complex type
- Adds a few unit tests

No renames yet.

### DIFF
--- a/library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
+++ b/library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
@@ -53,17 +53,12 @@ struct OptimizedBEGeneratorSystem {
     SelectTerm : (Int -> OptimizedBETermIndex)
 }
 
-// Get OptimizedBETermIndex coefficients
-function GetOptimizedBETermIndexCoeff(term : OptimizedBETermIndex) : Double {
-    term.Coefficient
-}
-
 
 // Get OptimizedBEGeneratorSystem coefficients
 function OptimizedBEGeneratorSystemCoeff(optimizedBEGeneratorSystem : OptimizedBEGeneratorSystem) : Double[] {
     mutable coefficients = [];
     for idx in 0..optimizedBEGeneratorSystem.NumTerms - 1 {
-        coefficients += [GetOptimizedBETermIndexCoeff(optimizedBEGeneratorSystem.SelectTerm(idx))];
+        coefficients += [optimizedBEGeneratorSystem.SelectTerm(idx).Coefficient];
     }
     return coefficients;
 }
@@ -312,7 +307,7 @@ function OptimizedBlockEncodingGeneratorSystem(data : JWOptimizedHTerms) : Optim
     mutable oneNorm = 0.0;
 
     for idx in 0..finalIdx - 1 {
-        oneNorm = oneNorm + AbsD(GetOptimizedBETermIndexCoeff(majIdxes[idx]));
+        oneNorm = oneNorm + AbsD(majIdxes[idx].Coefficient);
     }
 
     let majIdxes = majIdxes[0..finalIdx - 1];
--- a/library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
+++ b/library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
@@ -54,20 +54,16 @@ struct OptimizedBEGeneratorSystem {
 
 // Get OptimizedBETermIndex coefficients
 function GetOptimizedBETermIndexCoeff(term : OptimizedBETermIndex) : Double {
-    let (a, b, c, d, e, f) = term!;
-    return a;
+    term.Coefficient
 }
 
 
 // Get OptimizedBEGeneratorSystem coefficients
 function OptimizedBEGeneratorSystemCoeff(optimizedBEGeneratorSystem : OptimizedBEGeneratorSystem) : Double[] {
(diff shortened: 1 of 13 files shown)


library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
```diff
@@ -54,20 +55,16 @@ struct OptimizedBEGeneratorSystem {
 
 // Get OptimizedBETermIndex coefficients
 function GetOptimizedBETermIndexCoeff(term : OptimizedBETermIndex) : Double {
-    let (a, b, c, d, e, f) = term!;
-    return a;
+    term.Coefficient
 }
```
swernli, line 59: For the two places this is used you could just instead add `.Coefficient` to whatever argument was being passed, and then eliminate this function.

library/chemistry/src/Tests.qs
```diff
@@ -365,3 +404,8 @@ function NearEqualityFactD(actual : Double, expected : Double) : Unit {
         fail $"Values were not equal within tolerance\nActual: {actual}, Expected: {expected}, Tolerance: {tolerance}";
     }
 }
+
+// For convenience of constructing complex constants
+function NewComplex(r : Double, i : Double) : Complex {
```
swernli, line 409: This feels slightly redundant given the existence of the function call constructor, but that's also what we are trying to avoid... looking at the places where this is used, all except two are part of constructing a `JordanWignerInputState` so maybe a way to make this helper seem more helpful is to have it be a convenience method for constructing those. Then the remaining two spots can use the normal `new Complex` style.

library/chemistry/src/Tests.qs
```diff
@@ -316,46 +321,80 @@ operation JordanWignerUCCSTermTest() : Unit {
...
 @Test()
 operation JordanWignerUCCDTermPRSQTest() : Unit {
-    let term3 = [new JordanWignerInputState { Amplitude = (2.0, 0.0), FermionIndices = [0, 4, 2, 3] }];
+    let term3 = [new JordanWignerInputState { Amplitude = NewComplex(2.0, 0.0), FermionIndices = [0, 4, 2, 3] }];
     let state3 = [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.909297, 0., 0., 0., 0.,-0.416147, 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.];
     JordanWignerUCCTermTestHelper(5, [0, 4], term3, state3);
 }
 
+import JordanWigner.JordanWignerClusterOperatorEvolutionSet.ComputeJordanWignerBitString;
+
```
swernli, line 365: Should this get moved up with the other imports or is there a motivation for having it near the tests that use it?

library/chemistry/src/Tests.qs
```diff
@@ -1,14 +1,10 @@
...
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import JordanWigner.JordanWignerClusterOperatorEvolutionSet.JordanWignerClusterOperatorPQRSTermSigns;
-import JordanWigner.OptimizedBEOperator.OptimizedBEXY;
-import JordanWigner.OptimizedBEOperator.SelectZ;
-import JordanWigner.StatePreparation.PrepareSparseMultiConfigurationalState;
-import JordanWigner.StatePreparation.PrepareUnitaryCoupledClusterState;
-import JordanWigner.Utils.JordanWignerInputState;
+
```
swernli, line 4: It looks like maybe an extra line here.

swernli (approved):
The clean ups are great, thanks for working on this!

library/chemistry/src/JordanWigner/JordanWignerOptimizedBlockEncoding.qs
DmitryVasilevsky: Updated.

library/chemistry/src/Tests.qs
DmitryVasilevsky: We had a short discussion here in the office about this way of contructing struct and the lint we have about it. When it is a big struct, and is not constructed very often, this field-by field assignments are perfect. But when it is a simple struct and you need to construct a bunch of literals, spelling it all out is kind-of annoying (I tried!). And I'm talking specifically about constants. It is rather natural and anambiguous to write something like Complex(1.0, 0.0).
Could we extend Q# to support imginary unit literal? I want to write 1.0+0.0i :)

msoeken (approved):

library/chemistry/src/Tests.qs
DmitryVasilevsky: I moved construction to NewJordanWignerInputState instead, but it feels like pushing the same problem to a different place. We'll discuss later.

library/chemistry/src/Tests.qs
DmitryVasilevsky: I was thinking to put import close to the functions that test that particular import, but it didn't clarify much. I updated to use the usual approach.

library/chemistry/src/Tests.qs
DmitryVasilevsky: Not sure what you mean, but I verified consistent spacing in this file.
I did put an extra line between std lib imports and imports from this chemistry library, I think it makes it easier to understand - separation of stuff imported from elsewhere and stuff imported from neighboring files in the library.

Issue #1849, opened 2024-08-15
Katas playground: switching to tutorial and back to samples, disables "Run" button in the sample.


ggridin:
**Describe the bug**

Katas: switching to tutorial and back to samples, disables "Run" button in the sample.

**To Reproduce**

1. Open [Q# playground](https://microsoft.github.io/qsharp/)
2. Click "Teleportation" tutorial
3. Click "Entanglement" sample
4. See the error
5. Workaround: reload the page

**Expected behavior**

Loading tutorial pages must not affect samples 

**Screenshots**

![Sample run disabled](https://github.com/user-attachments/assets/390f47ce-859d-41c2-b83e-d6b5ad7d444c)

**System information**

Windows 10, 11Chrome and MS Edge browser

PR #2206, opened 2025-03-04
Fix run button disable. Switching to tutorial and back to samples (#1849)

thecoder93:
Hi, This should fix the problem #1849 
I think that the first time the model is created and then the get is done. For `samples` kataSection is `undefined` and I don't know why the EntryPoint is missing. What do you think? Probably it's only a workaround.


thecoder93:
I saw that this workaround is not good, because there are same kata into different editor. I will work on it.

thecoder93:
I understood. From Monaco doc: _Each model is identified by a URI. This is why it's not possible for two models to have the same URI._ 

Currently, we don't have the IDs for `samples`. Do you think to add it? I see also missing the structured type for Sample under `npm/qsharp` and the relative `getAll` such as `Katas`.
I think that with this we could have the independent ids for katas and Samples, thus when we do getModel from Monaco we get properly id.

thecoder93:
After several tests, I noticed a side effect when using updateDocument. Then the file is not closed with closeDocument. This should solve the problem. I think independent uri are therefore not necessary, or they would be an NTH.
What do you think? 

@billti @minestarks

minestarks:
Sorry @thecoder93 , you caught us at a particularly busy time - just now getting to look at this fix. The code change looks reasonable but I don't immediately see the connection to #1849. Testing.

thecoder93:
No problem @minestarks
I think that if you follow steps described inside the issue, after this fix, you shouldn't have the problem.

minestarks (approved):

minestarks:
Indeed I can repro the bug and this seems to fix it. Good catch @thecoder93 and thank you!

ScottCarda-MS (approved):

PR #2210, opened 2025-03-05
Add reference to QuantumArithmetic library

fedimser:
* This is Q# library implementing various integer arithmetic algorithms with quantum integers. It includes algorithms for addition, subtraction, comparison, multiplication, division, modular exponentiation, square root and greatest common divisor. 
* The library is located here: https://github.com/fedimser/quant-arith-re
* Adding it as requested in https://github.com/fedimser/quant-arith-re/issues/1

### DIFF
--- a/vscode/src/registry.json
+++ b/vscode/src/registry.json
@@ -82,9 +82,7 @@
         "github": {
           "owner": "fedimser",
           "repo": "quant-arith-re",
-          "refs": [
-            { "ref": "main", "notes": "nightly, unstable" }
-          ],
+          "refs": [{ "ref": "main", "notes": "nightly, unstable" }],
           "path": "lib"
         }
       }
--- a/vscode/src/registry.json
+++ b/vscode/src/registry.json
@@ -82,7 +82,10 @@
         "github": {
           "owner": "fedimser",
           "repo": "quant-arith-re",
-          "path": ""
+          "refs": [
+            { "ref": "main", "notes": "nightly, unstable" }
+          ],
+          "path": "lib"
         }
       }
     }
--- a/vscode/src/registry.json
+++ b/vscode/src/registry.json
@@ -74,6 +74,17 @@
           "path": "library/qtest"
         }
       }
+    },
+    {
+      "name": "QuantumArithmetic",
+      "description": "Quantum arithmetic algorithms",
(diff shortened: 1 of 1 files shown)


fedimser:
I am not sure if I specified github refspec correctly, please check this. My "qsharp.json" is in the root of repository, this is why "path" is empty.

vscode/src/registry.json
```diff
@@ -74,6 +74,17 @@
...
       }
+    },
+    {
+      "name": "QuantumArithmetic",
+      "description": "Quantum arithmetic algorithms",
+      "dependency": {
+        "github": {
+          "owner": "fedimser",
+          "repo": "quant-arith-re",
+          "path": ""
```
billti, line 85: If the Q# project is in the root of the repo (which yours is), you don't need to specify the path property, so you can just delete this.

vscode/src/registry.json
```diff
@@ -74,6 +74,17 @@
...
         }
       }
+    },
+    {
+      "name": "QuantumArithmetic",
+      "description": "Quantum arithmetic algorithms",
+      "dependency": {
+        "github": {
+          "owner": "fedimser",
+          "repo": "quant-arith-re",
```
billti, line 84: Can you add the `refs` property also as per the other examples in this file please. If you don't current have any branches or tags to list explicitly, just have the `{ "ref": "main", "notes": "nightly, unstable" }` entry. Thanks!

vscode/src/registry.json
fedimser: I moved the library so qsharp.josn is now in this folder: https://github.com/fedimser/quant-arith-re/tree/main/lib . So now,  as I understand, path is "lib".

vscode/src/registry.json
fedimser: Done.

DmitryVasilevsky (approved):

billti (approved):

fedimser:
Thank you for the fix and accepting the PR! 
If you (and you colleagues) like the library, I would appreciate a star on GitHub! 
Also, if there are other, potentially useful, quantum algorithms for integer arithmetic, I would be interested in implementing them in my library, or accepting a contribution in form of a pull request.

PR #2212, opened 2025-03-05
Distance coefficient power parameter in QEC scheme

msoeken:
This PR extends the formula template for QEC logical error rate formulas from:

$$ a \left(\frac{p}{p_{\mathrm{th}}}\right)^{\frac{d+1}{2}} $$

to

$$ a d^k \left(\frac{p}{p_{\mathrm{th}}}\right)^{\frac{d+1}{2}} $$

where $k$ is the new distance coefficient power discussed in [arXiv:2411.10406, Section III.B, page 25](https://arxiv.org/pdf/2411.10406#page=25).

For $k = 0$, the default, the formula remains unchanged compared to the current one. However, changing $k$ to other values allows to model a different error correction behavior, as motivated by the paper.

The parameter is called `distance_coefficient_power` or `distanceCoefficientPower` in the JSON parameter. It is also added to the Python RE parameter utility structure (together with `max_code_distance` which was absent so far).

### DIFF
--- a/resource_estimator/src/estimates/error_correction/code_with_threshold_and_distance.rs
+++ b/resource_estimator/src/estimates/error_correction/code_with_threshold_and_distance.rs
@@ -12,6 +12,7 @@ pub struct CodeWithThresholdAndDistance<Evaluator> {
     evaluator: Evaluator,
     crossing_prefactor: f64,
     error_correction_threshold: f64,
+    distance_coefficient_power: u32,
     max_code_distance: Option<u64>,
 }
 
@@ -20,11 +21,13 @@ impl<Evaluator> CodeWithThresholdAndDistance<Evaluator> {
         evaluator: Evaluator,
         crossing_prefactor: f64,
         error_correction_threshold: f64,
+        distance_coefficient_power: u32,
     ) -> Self {
         Self {
             evaluator,
             crossing_prefactor,
             error_correction_threshold,
+            distance_coefficient_power,
             max_code_distance: None,
         }
     }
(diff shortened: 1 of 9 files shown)


resource_estimator/src/estimates/error_correction/code_with_threshold_and_distance.rs
```diff
@@ -115,19 +129,25 @@ impl<Evaluator: CodeWithThresholdAndDistanceEvaluator> ErrorCorrection
...
-        let numerator = 2.0 * (self.crossing_prefactor / required_logical_qubit_error_rate).ln();
-        let denominator = (self.error_correction_threshold / physical_error_rate).ln();
-
-        let code_distance = (((numerator / denominator) - 1.0).ceil() as u64) | 0x1;
-
-        if let Some(max_distance) = self.max_code_distance {
-            if max_distance < code_distance {
-                return Err(format!("The computed code distance {code_distance} is too high; maximum allowed code distance is {max_distance}; try increasing the total logical error budget"));
+        // if distance coefficient power is 0, the function can be inverted and we can avoid looping over all code parameters
+        if self.distance_coefficient_power == 0 {
```
DmitryVasilevsky, line 133: Suggest moving the function code itself (above) and the inverse function code to the separate Rust functions to make it easier to understand.

DmitryVasilevsky (approved):

resource_estimator/src/estimates/error_correction/code_with_threshold_and_distance.rs
msoeken: Thanks, resolved in recent commit.

billti (approved):

PR #2213, opened 2025-03-05
Added SimpleBellPair sample

DmitryVasilevsky:
We are now missing a **simple** Bell pair sample. This pull request adds one.
As this will likely be one of the first samples people encounter, please make sure that it is simple enough while reviewing.

### DIFF
--- a/samples/algorithms/SimpleBellPair.qs
+++ b/samples/algorithms/SimpleBellPair.qs
@@ -27,7 +27,7 @@ operation Main() : (Result, Result) {
 /// Prepare Bell pair |Φ+⟩ = (|00⟩+|11⟩)/√2 on two qubits.
 /// Qubits are assumed to be in |00⟩ state.
 operation PrepareBellPair(q1 : Qubit, q2 : Qubit) : Unit {
-    // Set qubit `q1` in superposition of |0⟩ and |1⟩ by applying the Hadamard gate.
+    // Set qubit `q1` in superposition of |0⟩ and |1⟩ by applying a Hadamard gate.
     H(q1);
 
     // Entangle the two qubits `q1` and `q2` using the `CNOT` gate.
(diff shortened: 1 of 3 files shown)


samples_test/src/tests/algorithms.rs
```diff
@@ -254,6 +254,16 @@ pub const SHOR_EXPECT_DEBUG: Expect = expect![[r#"
     Found factor=17
     Found factorization 187 = 17 * 11
     (17, 11)"#]];
+pub const SIMPLEBELLPAIR_EXPECT: Expect = expect![[r#"
```
ScottCarda-MS, line 257: Looking at these test cases, I'm not sure I understand why we need a regular and a debug case for each test. The two cases seem like they are the same for every test. Can we get rid of the redundancy?

ScottCarda-MS (approved):

Manvi-Agrawal:
We have simple bell pair generation in the algorithms/entanglement.qs. How about renaming that to entanglement-simple-bell-pair.qs? That way we improve searchability and avoid redundancy. Thoughts?

samples_test/src/tests/algorithms.rs
DmitryVasilevsky: Good question. I had the same question some time ago. I believe there's a comment in the file that explains it. There're two modes of Q# interpreter and we need to test both. If we use the same expect for both Rust analyzer gets messed up.

samples/algorithms/SimpleBellPair.qs
Manvi-Agrawal, line 1: We have simple bell pair generation in the algorithms/entanglement.qs. How about renaming that to entanglement-simple-bell-pair.qs? That way we improve searchability and avoid redundancy. Thoughts?

samples_test/src/tests/algorithms.rs
DmitryVasilevsky: https://github.com/microsoft/qsharp/blob/5c7522c18ef26d3893e2f4866be7719d662cd132/samples_test/src/tests.rs#L23 - this is the place where the situation is described.

samples/algorithms/SimpleBellPair.qs
DmitryVasilevsky: Thanks, good observation. I still think that we need a separate sample for discoverability and AI training. But we may need to adjust entanglement sample. We'll discuss.

DmitryVasilevsky: I've reworked all simple samples.

DmitryVasilevsky:
See https://github.com/microsoft/qsharp/pull/2240

PR #2217, opened 2025-03-10
Fix name resolution from a project's Main.qs

### LINKED ISSUE #2207: Name resolution from a project's Main.qs in Python or Circuits does not work
If you follow the Python or Jupyter instructions from the documentation [Running a Program](https://learn.microsoft.com/en-us/azure/quantum/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-notebook-run#running-the-programs), the code will actually fail to run. For some reason, callables from Main.qs do not end up accessible either by their short names or their fully qualified names in `%%qsharp` magic cells or in `qsharp.eval`. Note that they DO show up and work as expected in the newer style Python interop, where they can be run like `qsharp.code.Main.Main()`.

My best guess at starting place for investigation: we do some extra tricks to make items from Main.qs available without an additional qualifying namespace when loading those projects as dependencies. But we also do some special tricks to make the loaded project from a `qsharp.init` call show up not as a dependency of the notebook/Python but as a code defined in the current context. I think somehow these two special cases are working against each other and causing thins from Main.qs to not show up at all. The other Python interop mechanism, exposing namespaces and callables through `qsharp.code`, uses an entirely different execution mechanism, so that could explain why it is unaffected.

UPDATE: Turns out this affects circuits as well... if you are in a project that has a Main.qs, generating circuits for any operation in that Main.qs fail with the same name resolution error.

swernli:
This fixes an issue where items from the Main.qs in a project would not be visible with the explicit `Main` namespace. When the dependency has no alias we should skip the special handling that elides the `Main` so that items are accessible as expected (and as documented in [Running the Programs](https://learn.microsoft.com/en-us/azure/quantum/how-to-work-with-qsharp-projects?tabs=tabid-qsharp%2Ctabid-notebook-run#running-the-programs))

Fixes #2207

### DIFF
--- a/compiler/qsc/src/interpret/tests.rs
+++ b/compiler/qsc/src/interpret/tests.rs
@@ -1447,12 +1447,11 @@ mod given_interpreter {
     }
 
     fn is_unit_with_output(result: &InterpretResult, output: &str, expected_output: &str) {
-        assert_eq!(expected_output, output);
-
         match result {
             Ok(value) => assert_eq!(Value::unit(), *value),
             Err(e) => panic!("Expected unit value, got {e:?}"),
         }
+        assert_eq!(expected_output, output);
     }
 
     fn is_only_error<E>(result: &Result<Value, Vec<E>>, output: &str, expected_errors: &Expect)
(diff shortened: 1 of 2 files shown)


compiler/qsc/src/interpret/tests.rs
```diff
@@ -2089,5 +2089,30 @@ mod given_interpreter {
                 entry: None,
             }
         }
+
+        #[test]
+        fn name_resolution_from_source_named_main_should_succeed() {
+            let sources = SourceMap::new(
+                [(
+                    "Main".into(),
```
minestarks, line 2097: Shouldn't this be `Main.qs` for the implicit namespace to be correctly parsed? (I guess I'm wrong because clearly this code compiles?)

compiler/qsc/src/interpret/tests.rs
```diff
@@ -2089,5 +2089,30 @@ mod given_interpreter {
...
+                sources,
+                PackageType::Lib,
+                TargetCapabilityFlags::all(),
+                LanguageFeatures::default(),
+                store,
+                &[(std_id, None)],
+            )
+            .expect("interpreter should be created");
+
+            let (result, output) = line(&mut interpreter, "Foo()");
```
minestarks, line 2114: I was expecting `Main.Foo()` to be the correctly qualified name here, and I wouldn't expect `Foo()` to resolve at all. After all that's how it works when you reference `Main` items from a different Q# file.

swernli: We dug into this, and you are right that this case should be an error. With the much nicer fix we landed at, I've pulled this out into a separate negative test.

compiler/qsc/src/interpret/tests.rs
swernli: The source map is expected to have already dropped the ".qs" extension, so that's why it's missing here.

minestarks (approved):

DmitryVasilevsky:
> The name resolution behavior table of doom:

Do we have this in the wiki or in the code. Could be useful...

orpuente-MS (approved):

PR #2218, opened 2025-03-10
Sccarda/separate measurements

ScottCarda-MS:


npm/qsharp/src/shared/circuit.ts
```diff
@@ -55,18 +55,32 @@ export interface Qubit {
   numResults?: number;
 }
 
+export interface Measurement {
```
minestarks, line 58: You typically avoid declaring common properties twice by doing something like this:

```ts
interface SomeCommonType {
    common_prop?: string
}

export type Measurement = SomeCommonType & {
    kind: "measurement",
    measurement_prop: string
}

export type Operation = SomeCommonType & {
    kind: "operation",
    operation_prop: string
}

export type Component = Measurement | Operation
```

npm/qsharp/ux/circuit-vis/circuit.ts
```diff
@@ -10,6 +10,7 @@ export {
   type Component,
   type DataAttributes,
   type Operation,
+  type Measurement,
```
minestarks, line 13: I haven't read the rest of the PR so maybe you use this later - but you shouldn't need to export these individual types if you just export `Component` - I don't expect you'll need to refer to those types by name

npm/qsharp/src/shared/circuit.ts
minestarks: And I haven't read the rest of the PR, but often, you don't even need to export or even name the `Measurement` and `Operation` subtypes - you can just write out the properties in `Component`'s definition.

npm/qsharp/ux/circuit-vis/circuit.ts
ScottCarda-MS: There are some functions throughout that only make sense when operating on Operations or Measurements. Adding controls to gates, for example, can't be done on Measurements.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -397,19 +398,23 @@ class CircuitEvents {
...
+          if (targets.some((reg) => reg.qubit == this.qubits.length - 1)) {
             return true;
           }
+          const controls =
+            comp.type === "Measurement" ? comp.qubits : comp.controls;
           if (
-            op.controls &&
-            op.controls.some((reg) => reg.qubit == this.qubits.length - 1)
+            controls &&
+            controls.some((reg) => reg.qubit == this.qubits.length - 1)
```
minestarks, line 411: So the issue here is that `controls` and `targets` got renamed for measurement? I don't think this looks bad - it may even be more readable since I don't know what a "control" and "target" is for a measurement.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -450,9 +455,12 @@ class CircuitEvents {
...
         dropzone.addEventListener("click", () => {
-          if (this.selectedOperation != null) {
-            const successful = addControl(this.selectedOperation, wireIndex);
-            this.selectedOperation = null;
+          if (
+            this.selectedComponent != null &&
+            this.selectedComponent.type === "Operation"
+          ) {
+            const successful = addControl(this.selectedComponent, wireIndex);
+            this.selectedComponent = null;
```
minestarks, line 463: This looks like an improvement to me, because you can't add controls to a measurement?

npm/qsharp/ux/circuit-vis/panel.ts
```diff
@@ -268,60 +270,71 @@ const _gate = (
...
  */
 interface GateDictionary {
-  [index: string]: Operation;
+  [index: string]: Component;
 }
 
 /**
  * Object for default gate dictionary
  */
 const defaultGateDictionary: GateDictionary = {
```
minestarks, line 279: This whole declaration was kind of redundant to begin with (all the `targets` fields are the same) so I don't find the addition of another field that offensive

npm/qsharp/ux/circuit-vis/contextMenu.ts
```diff
@@ -47,32 +47,36 @@ const addContextMenuToHostElem = (
     });
 
     const adjointOption = _createContextMenuItem("Toggle Adjoint", () => {
-      selectedOperation.isAdjoint = !selectedOperation.isAdjoint;
+      if (selectedComponent.type !== "Operation") return;
+      selectedComponent.isAdjoint = !selectedComponent.isAdjoint;
```
minestarks, line 51: If you're going for minimal changes, I feel like it's less wordy to write:

```suggestion
        selectedOperation.isAdjoint = (selectedComponent.type !== "Operation" || !selectedOperation.isAdjoint);
```

If you're going for _overall_ conciseness, this function probably can be refactored a bit to be shorter so you don't have `addControlOption` etc even defined for a measurement.

ScottCarda-MS: I've tried reordering the function a bit. Let me know if this makes things clearer for this function.

npm/qsharp/ux/circuit-vis/panel.ts
ScottCarda-MS: I've written a helper function to make that dictionary definition a little more readable.

npm/qsharp/ux/circuit-vis/events.ts
ScottCarda-MS: Yep, those were confusing terms for measurements, especially given that they worked the other way around from how you'd expect them to from non-measurements.

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: I've added a BaseOperation type to collect the shared fields, and made the Operation type a union type between the Measurement type and the Unitary type (which was the old Operation type). Let me know if this is a cleaner way to do this.

npm/qsharp/src/shared/circuit.ts
```diff
@@ -56,32 +56,52 @@ export interface Qubit {
...
   /** Specify conditions on when to render operation. */
   conditionalRender?: ConditionalRender;
-  /** Custom data attributes to attach to gate element.
-  Note that this is never written to file, so it is not part of the circuit schema */
-  dataAttributes?: DataAttributes;
 }
 
+export interface Measurement extends BaseOperation {
+  /** Discriminator for the Operation type */
+  kind: "Measurement";
```
minestarks, line 81: I'm used to seeing camelcased names in these kinds of values. Not sure if there's an "official" convention though

ScottCarda-MS: done

PR #2221, opened 2025-03-11
Implement base declarations and binary expression lowering

idavis:


orpuente-MS (approved):
Approved with minor comment.

PR #2223, opened 2025-03-12
Added support for light/dark theme toggle in the Playground editor

filipw:
The playground editor theme is now in based on the current system theme.
There is also a new button to toggle between dark/light themes.

out

### DIFF
--- a/npm/qsharp/ux/qsharp-ux.css
+++ b/npm/qsharp/ux/qsharp-ux.css
@@ -48,6 +48,7 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for 
   --error-background-color: #ffe3e3;
   --warning-background-color: #fff6d7;
   --bar-selected-outline: #587ddd;
+  --file-background: #fff;
 }
 
 /* TODO: Make the below playground specific classes */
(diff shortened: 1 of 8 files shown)


billti:
Thanks! It would definitely be nice to have, though I think something more wholistic might be preferable. It seems a little odd when the editor changes to dark theme but the rest of the page is still in a light theme.

I did support theming on the katas preview page if you want to see how it's done there. For example, you can open <https://microsoft.github.io/qsharp/katas.html> and change the system theme to see the whole site change (or you can explicitly open in [light mode](https://microsoft.github.io/qsharp/katas.html?theme=light) or [dark mode](https://microsoft.github.io/qsharp/katas.html?theme=dark). Would something like that be possible?

filipw:
Thanks, I wanted to keep the change small, as toggling just Monaco is relatively non-invasive - I didn't know I had the mandate to change the whole thing 😀

I've now pushed a change where the theme management is moved up to the nav component and propagates from there to the editors and the body. I also (obviously) had to introduce a bunch of new CSS.

I might have missed something because I am not sure I know all the nooks and crannies of the playground and what might be shown when, but it seems to work. 

out

minestarks:
Hi @filipw , I appreciate the work you put into this, but I'm going to close the PR for the same reason I closed #2228 - since the Q# Playground is supposed to be a minimal testing environment for the developers of this repo, we try to keep it as simple and maintenance-free as possible. So we decided to consider most UX features out of scope. 

Thank you for your contributions regardless!

PR #2233, opened 2025-03-17
Removed usage of deprecated set keyword from the samples

filipw:
The language samples still stated:

> // Mutable variables can be mutated with the `set` keyword

Technically this is true, but since `set` is now [deprecated](https://github.com/microsoft/qsharp/pull/2062) I removed it from all the samples.

### DIFF
--- a/samples/language/Variables.qs
+++ b/samples/language/Variables.qs
@@ -13,7 +13,7 @@ function Main() : Unit {
     mutable mutableInt = 43;
     Message($"Mutable Int: {mutableInt}");
 
-    // Mutable variables can be mutated or reassigned
+    // Mutable variables can be mutated or reassigned.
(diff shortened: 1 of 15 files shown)


swernli (approved):

samples/language/Variables.qs
```diff
@@ -13,10 +13,13 @@ function Main() : Unit {
     mutable mutableInt = 43;
     Message($"Mutable Int: {mutableInt}");
 
-    // Mutable variables can be mutated with the `set` keyword:
-    set mutableInt -= 1;
+    // Mutable variables can be mutated or reassigned
```
minestarks, line 16:
```suggestion
    // Mutable variables can be mutated or reassigned.
```

minestarks (approved):

PR #2236, opened 2025-03-18
Add an icon to clear the copilot chat

billti:
This add a "Clear chat" icon (see below on the right) to the Quantum Copilot chat panel, and wires up the pre-existing clear chat functioality to it

image

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -240,6 +240,11 @@
           "command": "qsharp-vscode.workspacesAdd",
           "when": "view == quantum-workspaces",
           "group": "navigation"
+        },
+        {
+          "command": "qsharp-vscode.copilotClear",
+          "when": "view == quantum-copilot",
+          "group": "navigation"
         }
       ],
       "view/item/context": [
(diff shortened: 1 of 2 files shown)


minestarks:
The `restartChat` method is the right one. Only problem is that if there is a request currently in flight, and you clear the chat, the assistant response isn't cancelled. So when the assistant response comes back, the response will pop up in the new, blank, conversation. I think this corrupts the conversation history :/

We could try
- disabling "clear chat" if there's a request in progress
- associate requests with the current conversation id, and drop responses that don't match that conversation id. (restartChat refreshes the conversation id.)
- lazy fix: drop responses if the UI isn't in the expected state (ending with a user message). This will almost certainly create race conditions though, as you can easily get in a state where you receive the "old" assistant response to a new user message that was submitted after the chat was cleared.

billti:
OK, not the most elegant (I'd rather the icon just be grayed out than disappear when disabled), but might be good enough for now with a backlog item to make more 'elegant'.

minestarks (approved):

minestarks:
Just noting that if we want to match the native chat experience (as I tend to), they use a "+" icon for this and call it "New Chat" . 

image

Not that I think theirs is the most intuitive, but at least it's a paradigm that's already there.

PR #2238, opened 2025-03-18
Circuit Editor

ScottCarda-MS:
This PR adds a custom editor view in the VS Code extension for files with the .qsc extension. The editor makes use of the same library code that is used to draw circuit diagrams elsewhere in the VS Code extension. The editor provides basic drag-and-drop editing of circuits with a simple gate set, including the ability to have controlled gates, and gates with arguments.
The circuit files are included in Q# projects and can be called into from Q# as if they were Q# operations. This is done by generating Q# files for circuit files when compiling.


compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    // Check if all operations are Unitary
+    let is_ctl_adj = circuit.component_grid.iter().all(|col| {
+        col.components.iter().all(|op| {
+            if let Operation::Unitary(unitary) = op {
+                unitary.gate != "|0〉" && unitary.gate != "|1〉"
+            } else {
+                false
+            }
+        })
```
swernli, line 70: This could be an `.any` that checks for any instances of resets or measurements and stops rather than verifying that every operation isn't reset or measurement.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,294 @@
...
+        for (index, circuit) in circuits.iter().enumerate() {
+            let circuit_name = format!("{file_name}{index}");
+            let circuit_str = build_operation_def(circuit_name, circuit);
+            qsharp_str.push_str(&circuit_str);
+        }
+        qsharp_str
+    }
+}
+
+pub fn build_operation_def(circuit_name: String, circuit: &Circuit) -> String {
```
swernli, line 36: This function got pretty long, it's probably worth refactoring it down into smaller logical parts to make it easier to reason about.

compiler/qsc_circuit/Cargo.toml
```diff
@@ -19,6 +19,7 @@ qsc_data_structures = { path = "../qsc_data_structures" }
 qsc_frontend = { path = "../qsc_frontend" }
```
swernli, line 19: I can't comment down below, so just putting this here. Looks like this Cargo.toml was missing the lint settings:
```
[lints]
workspace = true
```
And there are some warnings that need to be resolved. Search shows it's also missing from the qsc_formatter crate, but that deserves a separate issue.

swernli (commented):
Some initial review of the Rust code.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,294 @@
...
+        build_operation_def(file_name, &circuits[0])
+    } else {
+        let mut qsharp_str = String::new();
+        for (index, circuit) in circuits.iter().enumerate() {
+            let circuit_name = format!("{file_name}{index}");
+            let circuit_str = build_operation_def(circuit_name, circuit);
+            qsharp_str.push_str(&circuit_str);
+        }
+        qsharp_str
+    }
```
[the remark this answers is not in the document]
ScottCarda-MS: This is future-proofing. I think, especially in the context of generated Q# files, we will want the ability to have multiple defined circuits (which will get turned into operations) per file, but this is not implemented yet.

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -477,3 +599,246 @@ impl Display for Circuit {
...
+        let q_reg_idx_list: Vec<_> = q_regs.iter().map(|reg| reg.qubit).collect();
+        let cls_controls: Vec<_> = ctrls.iter().filter(|reg| reg.is_classical()).collect();
+        let is_classically_controlled = !cls_controls.is_empty();
+
+        if !is_classically_controlled && q_regs.is_empty() {
+            continue;
+        }
+
+        let min_reg_idx = if is_classically_controlled {
+            0
```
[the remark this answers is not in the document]
ScottCarda-MS: This code is a rust-ified version of preexisting typescript code that is present in the qviz code. I didn't really want to touch the classically-controlled logic as the circuit editor doesn't interact with it at all. The original code can be found [here](https://github.com/microsoft/qsharp/blob/main/npm/qsharp/ux/circuit-vis/process.ts#L148).

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -5,76 +5,188 @@
...
+}
+
+impl Operation {
+    /// Returns the kind of the operation.
+    pub fn kind(&self) -> String {
+        match self {
+            Operation::Measurement(_) => "measurement".to_string(),
+            Operation::Unitary(_) => "unitary".to_string(),
+        }
+    }
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

compiler/qsc_circuit/src/circuit.rs
```diff
@@ -477,3 +599,246 @@ impl Display for Circuit {
...
+///
+/// # Returns
+///
+/// A 2D array of operations.
+pub fn operation_list_to_grid(
+    mut operations: Vec<Operation>,
+    max_q_id: usize,
+) -> Vec<Vec<Operation>> {
+    for op in &mut operations {
+        if op.children().len() == 1 {
```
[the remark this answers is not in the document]
ScottCarda-MS: I added a comment above to try to explain this better.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: done

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: refactored to have smaller functions

compiler/qsc_circuit/Cargo.toml
ScottCarda-MS: I've added this in and went and fixed the lints it found. Thanks for catching this!

npm/qsharp/src/shared/circuit.ts
```diff
@@ -3,23 +3,130 @@
...
 import { Register } from "./register.js";
 
+/**
+ * Current format version.
+ */
+export const CURRENT_VERSION = 1;
+
+export interface CircuitGroup {
+  circuits: Circuit[];
+  version?: number;
```
minestarks, line 13: Can this be required?

npm/qsharp/src/shared/circuit.ts
```diff
@@ -3,23 +3,130 @@
...
+  componentGrid: ComponentGrid;
 }
 
+export type ComponentGrid = Column[];
+
+export interface Column {
+  components: Component[];
+}
+
+export type Component = Operation;
```
minestarks, line 31: Consider adding a comment describing the distinction between these two types

npm/qsharp/src/shared/circuit.ts
```diff
@@ -3,23 +3,130 @@
...
+
+/**
+ * Update circuit group to current format version.
+ */
+export const updateToCurrentVersion = (
+  circuits: CircuitGroup,
+): CircuitGroup => {
```
minestarks, line 38: Could you please prefer the "regular" function declaration syntax?

```ts
export function updateToCurrentVersion(circuits: CircuitGroup): CircuitGroup {}
```

wasm/src/lib.rs
```diff
@@ -141,7 +141,17 @@ pub fn get_circuit(
...
 
-    serde_wasm_bindgen::to_value(&circuit).map_err(|e| e.to_string())
+    let circuitData = qsc::circuit::CircuitGroup {
+        circuits: vec![circuit],
+        version: Some(qsc::circuit::CURRENT_VERSION),
+    };
+
+    serde_wasm_bindgen::to_value(&circuitData).map_err(|e| e.to_string())
+}
+
```
minestarks, line 151: It feels to me like this should continue to return a bare `Circuit` and the wrapping of it in a `CircuitGroup` can be handled on a higher layer.

npm/qsharp/ux/circuit.tsx
```diff
@@ -13,29 +15,211 @@ import { Spinner } from "./spinner.js";
...
+            components: op.children.map((child: any) => toOperation(child)),
+          },
+        ];
+      }
+      return convertedOp;
+    }
+  }
+}
+
+function toCircuitGroup(circuit: any): qviz.CircuitGroup {
```
minestarks, line 110: Consider the case where you haven't even opened the circuit widget, but you're running a project that contains a legacy .qsc file. How does that file get migrated to the new version? I think this forces us to move this code deeper into the native bits.

minestarks (commented):
Posting first round of comments

minestarks (commented):
Not done, but leaving another batch of comments to unblock you

vscode/src/createProject.ts
```diff
@@ -132,15 +132,18 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
             },
           );
           for (const [name, type] of dirFiles) {
-            if (type === vscode.FileType.File && name.endsWith(".qs")) {
+            if (
+              type === vscode.FileType.File &&
+              (name.endsWith(".qs") || name.endsWith(".qsc"))
```
minestarks, line 137: This is a preexisting bug, but this catches files named `.qs`  and `.qsc`. Maybe explicitly exclude those?

vscode/src/createProject.ts
```diff
@@ -123,7 +123,7 @@ export async function initProjectCreator(context: vscode.ExtensionContext) {
           return;
         }
 
-        async function getQsFilesInDir(dir: vscode.Uri) {
+        async function getQsAndQscFilesInDir(dir: vscode.Uri) {
```
minestarks, line 126: Suggestion: call this `getSourceFilesInDir` to avoid the tiresome "qs and qsc" repetition

vscode/src/common.ts
```diff
@@ -25,6 +25,16 @@ export function isQsharpDocument(document: TextDocument): boolean {
...
 
+// Returns true for all circuit documents
+export function isCircuitDocument(document: TextDocument): boolean {
+  return (
+    document.languageId === "qsc" &&
+    (Utils.extname(document.uri) === ".qsc" || document.isUntitled) &&
+    document.uri.scheme !== "git" &&
+    document.uri.scheme !== "pr"
+  );
+}
```
minestarks, line 36: I actually think you could go ahead and fold this into `isQsharpDocument`. i.e., just count `.qsc` files as Q# documents as well. I'm pretty sure all the callers of `isQsharpDocument` are also cases where we want to treat `.qsc` files as Q# files as well.

vscode/src/language-service/activate.ts
minestarks, line 1: You missed `onDidCloseTextDocument` (but you won't need to update it, if you follow my advice about just checking for .qsc files directly inside `isQsharpDocument`). 

Also please add another `QsharpDocumentType` on lines 199-229 where we send telemetry, so that the telemetry event gets fired with a unique `QsharpDocumentType` for .qsc files. That way you're also getting "free" usage telemetry for your feature.

minestarks (commented):
More comments....

compiler/qsc_circuit/src/builder.rs
```diff
@@ -252,17 +255,24 @@ impl Builder {
...
-            circuit.qubits.push(crate::circuit::Qubit {
+            qubits.push(crate::circuit::Qubit {
                 id: i,
-                num_children: num_measurements,
+                num_results: num_measurements,
             });
         }
 
-        circuit
+        let max_q_id = qubits.iter().map(|qubit| qubit.id).max().unwrap_or(0);
```
minestarks, line 270: Isn't this just `qubits.len()`? (I think this code used to be more complicated before Stefan's base profile changes, so maybe this is a holdover from that?)

compiler/qsc_circuit/src/builder.rs
```diff
@@ -252,17 +255,24 @@ impl Builder {
...
+
+        Circuit {
+            component_grid: op_grid_to_comp_grid(operation_list_to_grid(operations, max_q_id)),
```
minestarks, line 273: `op_grid_to_comp_grid(operation_list_to_grid(...))` are always called together like this. Maybe just expose one function, `operation_list_to_component_grid`?

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,379 @@
...
+            let args = op.args();
+            if !should_add_pi && !args.is_empty() {
+                should_add_pi = args.iter().any(|arg| arg.contains("π"));
+            }
+        }
+    }
+
+    if should_add_pi {
+        // Add the π constant
+        qsharp_str.push_str(&format!("{indent}let π = Std.Math.PI();\n"));
```
swernli, line 127: latest clippy warns about this pattern, so use this instead:

```suggestion
        writeln!(qsharp_str, "{indent}let π = Std.Math.PI();");
```

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,379 @@
...
+        let op_x = Unitary {
+            gate: "X".to_string(),
+            is_adjoint: false,
+            controls: vec![],
+            targets: ket.targets.clone(),
+            args: vec![],
+            children: vec![],
+        };
+        let operation_str = operation_call(&op_x, qubits);
+        call_str.push_str(&format!("{indent}{operation_str};\n"));
```
swernli, line 245: here too:

```suggestion
        writeln!(call_str, "{indent}{operation_str};").expect("writing to string should succeed");
```

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,379 @@
...
+                controls: unitary.controls.clone(),
+                is_adjoint: !unitary.is_adjoint, // Adj R
+            },
+            qubits,
+        );
+        if unitary.is_adjoint {
+            format!("{indent}{general_r_str};\n{indent}{rx_str};\n")
+        } else {
+            format!("{indent}{rx_str};\n{indent}{general_r_str};\n")
+        }
```
swernli, line 224: I had to think about this one quite a bit... technically it is correct to reverse the operation order when doing the adjoint, but in this case the phase applied by `R(PauliI, _, _)` and the rotation performed by `Rx` commute with each other. So you could simplify this code to just always emit the `Rx` followed by `R` instead of changing the order based on `unitary.is_adjoint`.

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+      circuit.trim().replace(/\r\n/g, "\n") ===
+      document.getText().trim().replace(/\r\n/g, "\n")
+    ) {
+      return;
+    }
+
+    const edit = new vscode.WorkspaceEdit();
+
+    // Just replace the entire document every time for this example extension.
+    // A more complete extension should compute minimal edits instead.
```
minestarks, line 160: "Example extension"? Is this left over from sample code?

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+    const text = document.getText();
+    if (text.trim().length === 0) {
+      return {};
+    }
+
+    try {
+      return JSON.parse(text);
+    } catch {
+      throw new Error(
+        "Could not get document as json. Content is not valid json",
```
minestarks, line 140: How is this error surfaced to the user? Does someone catch this? 

This is what I see in the UI if the JSON is corrupt:

image

vscode/src/webview/editor.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+  switch (message.command) {
+    case "circuit":
+      {
+        // Check if the received circuit is different from the current state
+        if (
+          state.viewType === "circuit" &&
+          JSON.stringify(state.props.circuit) ===
+            JSON.stringify(message.props.circuit)
+        ) {
```
orpuente-MS, line 107: How often does this `onMessage` function triggers? Comparing the full JSONs each time seems like an expensive operation for the UI, specially as the circuits get bigger.

npm/qsharp/ux/circuit-vis/sqore.ts
```diff
@@ -336,41 +449,102 @@ export class Sqore {
...
+
+  public usePanel(options?: PanelOptions): Sqore {
+    this.extensions = [...this.extensions, extensionPanel(options)];
+    return this;
+  }
+
+  public useEvents(): Sqore {
+    this.extensions = [...this.extensions, extensionEvents];
+    return this;
+  }
```
minestarks, line 513: There are several things I'm not really following with this design. 

- Why do these attributes all need to be exposed as separate methods? Isn't a circuit just editable, or not editable? 

- What's the deal with `extensions`?

- I know fluent interfaces have their fans, but I'm just general confused by them. Why go with a fluent interface here? These could return void... or they could just be flags passed into the constructor.

swernli (approved):
Signing off on the Rust bits (modulo outstanding comments from others).

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: I've changed this to the more straightforward decomp we talked about with HSH, which also does not need reversing.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: fixed

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: fixed

vscode/src/circuitEditor.ts
ScottCarda-MS: removed

vscode/package.json
```diff
@@ -472,6 +472,20 @@
           "dark": "./resources/file-icon-dark.svg"
         },
         "configuration": "./language-configuration.json"
+      },
+      {
+        "id": "qsc",
```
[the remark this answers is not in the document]
ScottCarda-MS: Yes I like qsharpcircuit. I like the more specific id here. Done

samples/CircuitIntegration/qsharp.json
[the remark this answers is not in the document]
ScottCarda-MS: done

samples/CircuitIntegration/src/Main.qs
```diff
@@ -0,0 +1,20 @@
+/// # Sample
+/// Circuit Integration
+///
+/// # Description
```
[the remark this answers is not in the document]
ScottCarda-MS: Good suggestion

wasm/src/lib.rs
```diff
@@ -141,7 +141,17 @@ pub fn get_circuit(
         .circuit(entry_point, simulate)
         .map_err(interpret_errors_into_qsharp_errors_json)?;
 
-    serde_wasm_bindgen::to_value(&circuit).map_err(|e| e.to_string())
+    let circuitData = qsc::circuit::CircuitGroup {
```
orpuente-MS, line 144: Should this be in snake_case? `circuitData` -> `circuit_data`.

npm/qsharp/ux/circuit-vis/sqore.ts
```diff
@@ -185,35 +263,49 @@ export class Sqore {
     // Add styles
     const css = document.createElement("style");
     css.innerHTML = style(this.style);
+    css.className = "qviz-style";
     svg.appendChild(css);
 
+    // ToDo: idk why we need to add this style to the document head.
```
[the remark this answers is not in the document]
ScottCarda-MS: It is a bit silly, as you said above, but if we are adding a style node, then we'll need to do this logic to make it consistent across initial load and after an interaction. The solution is to do what you suggest above and not add this style element in the first place. The ToDo wording has been removed in the meantime.

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  public async resolveCustomTextEditor(
+    document: vscode.TextDocument,
+    webviewPanel: vscode.WebviewPanel,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _token: vscode.CancellationToken,
+  ): Promise<void> {
+    log.info("Resolving CircuitEditorProvider");
+    console.log("Resolving CircuitEditorProvider");
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm going to keep these in here while debugging the sync logic, but I will be sure to remove them when that is resolved. I'll keep this comment open as a reminder.

vscode/src/webview/editor.tsx
```diff
@@ -0,0 +1,153 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore - there are no types for this
+import mk from "@vscode/markdown-it-katex";
+import markdownIt from "markdown-it";
+const md = markdownIt("commonmark");
+md.use(mk, {
+  enableMathBlockInHtml: true,
+  enableMathInlineInHtml: true,
+});
+setRenderer((input: string) => md.render(input));
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

vscode/src/circuit.ts
```diff
@@ -247,7 +247,11 @@ async function getCircuitOrError(
       simulate,
       params.operation,
     );
-    return { result: "success", simulated: simulate, circuit };
+    return {
+      result: "success",
+      simulated: simulate,
+      circuit: circuit,
+    };
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  constructor(private readonly context: vscode.ExtensionContext) {
+    log.info("Constructing CircuitEditorProvider");
+  }
+
+  public async resolveCustomTextEditor(
+    document: vscode.TextDocument,
+    webviewPanel: vscode.WebviewPanel,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _token: vscode.CancellationToken,
```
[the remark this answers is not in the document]
ScottCarda-MS: done

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+          return;
+      }
+    });
+
+    const updateWebview = () => {
+      const circuit = this.getDocumentAsJson(document);
+      const filename = document.fileName.split(/\\|\//).pop()!.split(".")[0];
+
+      const props = {
+        title: `${filename} Circuit`,
```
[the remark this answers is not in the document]
ScottCarda-MS: done

wasm/src/lib.rs
ScottCarda-MS: good catch! I'm surprised clippy didn't catch this!

vscode/src/common.ts
```diff
@@ -14,6 +14,7 @@ import {
 import * as vscode from "vscode";
 
 export const qsharpLanguageId = "qsharp";
+export const qsharpCircuitLanduageId = "qsharpcircuit";
```
minestarks, line 17: `qsharpCircuitLanguageId` typo

minestarks (commented):
Finished the first review pass. TBH, the UI stuff in circuit-vis left me feeling pretty out of my depth, so I gave it a pretty basic pass. This didn't feel great, so I thought long and hard about what could make it more digestible. I feel strongly now the answer is "unit tests" . If I saw some examples of `Sqore` (and other functions) used in tests as they're intended to be used, I'd feel SO much more comfortable navigating the code. May not be possible in this PR, but let's prioritize getting those in.

compiler/qsc_circuit/src/builder.rs
ScottCarda-MS: Ah yes, this used to be a useful safeguard against an edge case, but I can see that the code has been simplified since then and now it is always just the number of qubits, as you suggest.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,342 @@
...
+    qsharp_str.push_str(&generate_qubit_validation(
+        circuit_name,
+        &qubits,
+        indentation_level,
+    ));
+
+    let mut body_str = String::new();
+    let mut should_add_pi = false;
+
+    // ToDo: Add support for children operations
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

compiler/qsc_circuit/src/builder.rs
ScottCarda-MS: I've moved the functionality of `op_grid_to_comp_grid` into `operation_list_to_grid` so there's only one function now, except in the test file where `op_grid_to_comp_grid` was still useful.

compiler/qsc_project/src/project.rs
```diff
@@ -243,8 +249,11 @@ pub trait FileSystemAsync {
             })?;
         let mut files = vec![];
         for item in filter_hidden_files(listing.into_iter()) {
+            let extension = item.entry_extension();
             match item.entry_type() {
-                Ok(EntryType::File) if item.entry_extension() == "qs" => files.push(item.path()),
+                Ok(EntryType::File) if extension == "qs" || extension == "qsc" => {
```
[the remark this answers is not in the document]
ScottCarda-MS: updated

compiler/qsc_project/src/project.rs
```diff
@@ -419,22 +428,40 @@ pub trait FileSystemAsync {
         // all the found files in the package sources. This way compilation
```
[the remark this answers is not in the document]
ScottCarda-MS: done

vscode/src/createProject.ts
ScottCarda-MS: done

vscode/src/createProject.ts
ScottCarda-MS: done

vscode/src/webview/editor.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  // Use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
+  // to detect changes to the theme attribute.
+  const callback = (mutations: MutationRecord[]) => {
+    for (const mutation of mutations) {
+      if (mutation.attributeName === themeAttribute) {
+        updateGitHubTheme();
+      }
+    }
+  };
```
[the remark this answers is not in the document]
ScottCarda-MS: I've removed them from here and imported the `setThemeStylesheet` from the webview file.

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: I've removed this function as it wasn't doing anything useful.

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: done

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: done

vscode/src/webview/editor.tsx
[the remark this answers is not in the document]
ScottCarda-MS: Something about fixing this broke the editor, so I reverted the change and am reopening this comment.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+    this._addQubitLineControlEvents();
+    this._addDocumentEvents();
+  }
+
+  /**
+   * Dispose the CircuitEvents instance and remove event listeners
+   */
+  dispose() {
+    this._removeToolboxElementsEvents();
+    this._removeDocumentEvents();
```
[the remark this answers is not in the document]
ScottCarda-MS: These events are special as they aren't attached to something in the circuit itself. Because the circuit gets redrawn throughout the editing process, events attached to other elements are disposed by virtue of that their elements are gone. The toolbox and document events are different because every time the circuit gets redrawn, a new instance of their events gets attached. This is why it is important to explicitly remove them when the circuit is disposed.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+   ***************************/
+
+  documentKeydownHandler = (ev: KeyboardEvent) => {
+    const selectedLocation = this.selectedOperation
+      ? getGateLocationString(this.selectedOperation)
+      : null;
+    if (ev.ctrlKey && selectedLocation) {
+      this.container.classList.remove("moving");
+      this.container.classList.add("copying");
+    } else if (ev.key == "Delete" && selectedLocation) {
```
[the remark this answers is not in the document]
ScottCarda-MS: Ah, this should be removed.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+
+  /***************************
+   * Events Adding Functions *
+   ***************************/
+
+  documentKeydownHandler = (ev: KeyboardEvent) => {
+    const selectedLocation = this.selectedOperation
+      ? getGateLocationString(this.selectedOperation)
+      : null;
+    if (ev.ctrlKey && selectedLocation) {
```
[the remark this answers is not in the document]
ScottCarda-MS: I think we will definitely need to add some kind of "Help" or "About" or welcome tutorial panel/popup at some point to help with discoverability, but this is out of scope for this PR.

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+  removeAllWireDropzones,
+} from "./draggable";
+
+let events: CircuitEvents | null = null;
+
+const extensionEvents = (
+  container: HTMLElement,
+  sqore: Sqore,
+  useRefresh: () => void,
+): void => {
```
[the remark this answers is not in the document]
ScottCarda-MS: In some of these cases, like here and especially with some of the event handlers, I *do* think it makes more sense to declare them like data rather than like functions because they are used like data in how they are referenced. They are never called directly and are handlers that are passed to APIs.

npm/qsharp/ux/circuit-vis/circuitManipulation.ts
```diff
@@ -0,0 +1,550 @@
...
+      registers = [
+        ...sourceOperation.targets,
+        ...(sourceOperation.controls || []),
+      ];
+      break;
+    case "measurement":
+      registers = [...sourceOperation.qubits, ...sourceOperation.results];
+      break;
+    case "ket":
+      registers = [...sourceOperation.targets]; // ToDo: is it necessary to do this ... syntax here?
```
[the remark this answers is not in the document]
ScottCarda-MS: This is change to just set registers to sourceOperation.targets, as registers is just used to check if the targetWire is targeting an existing register.

npm/qsharp/ux/circuit-vis/metadata.ts
```diff
@@ -50,7 +52,7 @@ export interface Metadata {
   /** Gate width. */
   width: number;
   /** Children operations as part of group. */
-  children?: (Metadata | Metadata[])[];
+  children?: Metadata[][] | Metadata[][][];
```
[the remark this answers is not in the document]
ScottCarda-MS: This will be a good cleanup thing to do, but given the size of this PR I'm going to wait for a future PR to do this.

vscode/src/webview/editor.tsx
ScottCarda-MS: This gets called when the circuit is first loaded and whenever a change is made directly to the text document and the text document needs to update the GUI. Currently that second case is disabled, but when enabled that will trigger when the user uses VS Code's undo/redo functionality.

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,342 @@
...
+    } else if unitary.is_adjoint {
+        "Adjoint "
+    } else {
+        ""
+    };
+
+    let mut args = vec![];
+
+    // Create the regex for matching numbers (both integers and doubles)
+    let number_regex = Regex::new(r"((\d+(\.\d*)?)|(\.\d+))").expect("Regex should compile");
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm having trouble seeing the suggestion. How would the expr parser let me identify and replace all integers for their double equivalent inside the string? And if so, that seems like it would be a much more complex approach to a relatively simple goal. Is having the regex_lite used here really so bad?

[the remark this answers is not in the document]
ScottCarda-MS: We do make use of regex_lite elsewhere in the repo, by the way. It is not the full-blow regex create.

vscode/src/common.ts
ScottCarda-MS: fixed

npm/qsharp/ux/circuit-vis/events.ts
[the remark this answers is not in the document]
ScottCarda-MS: removed

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+    webviewPanel.webview.options = {
+      enableScripts: true,
+    };
+    webviewPanel.webview.html = this.getHtmlForWebview(webviewPanel.webview);
+
+    webviewPanel.webview.onDidReceiveMessage((e) => {
+      switch (e.command) {
+        case "alert":
+          vscode.window.showErrorMessage(e.text);
+          return;
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+          }
+        }
+      });
+      this.circuitSvg.appendChild(dropzone);
+    });
+  }
+}
+
+/**
+ * Create a custom confirm dialog box
```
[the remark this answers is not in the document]
ScottCarda-MS: removed wording of "custom"

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+  private selectedWire: number | null;
+  private movingControl: boolean;
+  private mouseUpOnCircuit: boolean;
+  private dragging: boolean;
+  private disableLeftAutoScroll: boolean;
+
+  constructor(container: HTMLElement, sqore: Sqore, useRefresh: () => void) {
+    this.renderFn = useRefresh;
+
+    this.container = container;
```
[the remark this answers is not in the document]
ScottCarda-MS: done

wasm/src/debug_service.rs
```diff
@@ -56,7 +56,11 @@ impl DebugService {
 
     pub fn get_circuit(&self) -> Result<JsValue, String> {
         let circuit = self.debugger().circuit();
-        serde_wasm_bindgen::to_value(&circuit).map_err(|e| e.to_string())
+        let circuitData = qsc::circuit::CircuitGroup {
```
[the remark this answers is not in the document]
ScottCarda-MS: moved into the npm

vscode/src/common.ts
ScottCarda-MS: done

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -0,0 +1,800 @@
...
+  dispose() {
+    this._removeToolboxElementsEvents();
+    this._removeDocumentEvents();
+  }
+
+  /***************************
+   * Events Adding Functions *
+   ***************************/
+
+  documentKeydownHandler = (ev: KeyboardEvent) => {
```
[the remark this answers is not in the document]
ScottCarda-MS: See https://github.com/microsoft/qsharp/pull/2238#discussion_r2038246933

vscode/src/language-service/activate.ts
ScottCarda-MS: I've added the QsharpDocumentType as you suggest to get telemetry for circuit files.

npm/qsharp/ux/circuit.tsx
```diff
@@ -128,9 +321,21 @@ function ZoomableCircuit(props: { circuit: qviz.Circuit }) {
...
+      if (editCallback) {
+        circuitPanel = circuitPanel.useOnCircuitChange(editCallback);
+      }
+      circuitPanel.useEvents().draw(container);
+    } else {
+      qviz.create(circuitGroup).draw(container);
+    }
     // circuit-vis hardcodes the styles in the SVG.
     // Remove the style elements -- we'll define the styles in our own CSS.
     const styleElements = container.querySelectorAll("style");
```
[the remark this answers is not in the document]
ScottCarda-MS: removed the injected style node

npm/qsharp/ux/circuit-vis/styles.ts
```diff
@@ -233,4 +241,75 @@ const _expandCollapse = `
         visibility: visible;
         opacity: 1;
         transition: opacity 1s;
-    }`;
+    }
+`;
+
+const _panel = `
```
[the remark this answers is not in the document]
ScottCarda-MS: moved the relevant styles to the CSS file

vscode/src/webview/editor.tsx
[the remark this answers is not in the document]
ScottCarda-MS: fixed

compiler/qsc_project/src/tests/harness.rs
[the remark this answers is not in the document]
ScottCarda-MS: I've added a couple of unit tests for projects with circuits in them.

npm/qsharp/ux/circuit-vis/sqore.ts
ScottCarda-MS: I've removed the extensions mechanism for extending the Sqore and just call all the components deliberately in the render function.

npm/qsharp/ux/qsharp-circuit.css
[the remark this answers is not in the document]
ScottCarda-MS: Done. That looks nice!

npm/qsharp/ux/circuit.tsx
ScottCarda-MS: I added the json_to_circuit.rs file to qsc_circuit in the rust layer to handle this conversion for compilation.

vscode/src/circuitEditor.ts
```diff
@@ -0,0 +1,169 @@
...
+      };
+
+      const message = {
+        command: "circuit",
+        props,
+      };
+      webviewPanel.webview.postMessage(message);
+    };
+
+    // Update the webview when the text document changes
```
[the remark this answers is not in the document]
ScottCarda-MS: We've made improvements to the sync logic and it seems to be running stable now. Thanks for you help on this!

vscode/src/circuitEditor.ts
ScottCarda-MS: Okay I've changed the logic a bit so that the error stops the circuit from rendering and displays the error message instead:
image

compiler/qsc_project/src/project.rs
```diff
@@ -419,22 +428,40 @@ pub trait FileSystemAsync {
...
+                        .to_string();
+                    contents = match circuits_to_qsharp(&name, contents.as_ref()) {
+                        Ok(c) => Ok(Arc::from(c)),
+                        Err(e) => Err(Error::Circuit {
+                            path: path.to_string_lossy().to_string(),
+                            error: e.to_string(),
+                        }),
+                    }?;
+                }
+            }
```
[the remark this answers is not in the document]
ScottCarda-MS: the conversion from circuit to Q# was moved to the package logic, as per your suggestion

wasm/src/lib.rs
```diff
@@ -141,7 +141,17 @@ pub fn get_circuit(
...
+        circuits: vec![circuit],
+        version: Some(qsc::circuit::CURRENT_VERSION),
+    };
+
+    serde_wasm_bindgen::to_value(&circuitData).map_err(|e| e.to_string())
+}
+
+#[wasm_bindgen]
+pub fn circuits_to_qsharp(file_name: &str, circuits_json: &str) -> Result<String, String> {
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

vscode/src/language-service/activate.ts
```diff
@@ -277,13 +279,28 @@ function registerDocumentUpdateHandlers(
...
+      isCircuitDocument(document) ||
+      (isQsharpDocument(document) && !isQsharpNotebookCell(document))
+    ) {
+      let content = document.getText();
+
+      // Check if the document is a .qsc file and convert it to Q# if needed
+      if (document.fileName.endsWith(".qsc")) {
+        try {
+          const name = document.fileName.split(/\\|\//).pop()!.split(".")[0];
+          content = await circuitsToQSharp(name, content);
```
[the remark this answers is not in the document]
ScottCarda-MS: done

vscode/src/language-service/activate.ts
```diff
@@ -277,13 +279,28 @@ function registerDocumentUpdateHandlers(
...
-      // Regular (not notebook) Q# document.
+  async function updateIfQsharpDocument(document: vscode.TextDocument) {
+    if (
+      isCircuitDocument(document) ||
+      (isQsharpDocument(document) && !isQsharpNotebookCell(document))
+    ) {
+      let content = document.getText();
+
+      // Check if the document is a .qsc file and convert it to Q# if needed
+      if (document.fileName.endsWith(".qsc")) {
```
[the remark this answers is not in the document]
ScottCarda-MS: the conversion from circuit to Q# was moved to the package logic, as per your suggestion

vscode/src/webview/utils.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const themeAttribute = "data-vscode-theme-kind";
```
orpuente-MS, line 4: This file is called `utils.ts` but it only contains a theme related utility. Could we rename it to `theme.ts`?

ScottCarda-MS: done

minestarks (approved):
LGTM. Thanks!

compiler/qsc_circuit/src/circuit_to_qsharp.rs
```diff
@@ -0,0 +1,376 @@
...
+        "{indent}if Length(qs) != {} {{\n\
+        {inner_indent}fail \"Invalid number of qubits. Operation {} expects a qubit register of size {}.\";\n\
+        {indent}}}\n",
+        qubits.len(),
+        circuit_name,
+        qubits.len()
+    )
+}
+
+fn handle_measurement(
```
orpuente-MS, line 154: Can we rename the functions called `handle_*` to something more explicit, like `generate_measurement_call`?

orpuente-MS (approved):

compiler/qsc_circuit/src/circuit_to_qsharp.rs
ScottCarda-MS: done

PR #2240, opened 2025-03-18
Rework of simple samples

DmitryVasilevsky:
- Moved simple samples from algorithms to getting_started subfolder.
- Simplified code
- Updated comments
- Renamed files, changed content to show different concepts and features of the Q#
- Added SimpleTeleportation (still not suitable for base profile).

Getting_started sample (that we call Minimal) is left in the languge subfolder.


msoeken (approved):

samples/getting_started/CatStatesAndGHZ.qs
```diff
@@ -0,0 +1,63 @@
...
+/// The Greenberger–Horne–Zeilinger state, or GHZ state, is a state with 3
+/// qubits defined as: |GHZ〉 = (|000〉 + |111〉) / √2.
+///
+/// The GHZ state is said to be a maximally entangled state, a multi-qubit
+/// state where the state of any one qubit is not separable from the state
+/// of any of the other qubits.
+///
+/// The generalized form of the GHZ state across any number of qubits is
+/// called a Cat state, and the GHZ state is a special case of the Cat
+/// state where the number of qubits is 3.
```
swernli, line 20: This isn't really true. The Bell states are specifically among 2 qubits, and the GHZ state is among 3 or more states. So "cat state" isn't really a generalized form of GHZ, just another name for it. As long as the register has more than two qubits in a superposition of exactly two states it is accurate to call it a cat state or a GHZ state.

samples/getting_started/BellStates.qs
```diff
@@ -1,35 +1,37 @@
...
-    }
-    return measurements;
+/// # Summary
+/// Allocates a pair of qubits, prepares them using `preparation` operation,
+/// Then measures and resets them. Returns a pair of results.
+operation PrepareAndMeasurePair(
+    preparation : Qubit[] => Unit,
+    name : String
+) : (Result, Result) {
+    use pair = Qubit[2]; // Allocate a pair of qubits
```
swernli, line 30: in most of our existing samples we follow the pattern of having comments above the line they apply to rather than to the right. I think we should stick to this rather than having a mix of both styles.

samples/getting_started/CatStatesAndGHZ.qs
```diff
@@ -0,0 +1,63 @@
...
+
+    use ghz = Qubit[3]; // Allocate 3 qubits for GHZ state.
+    PrepareGHZState(ghz); // Prepare a GHZ state in the register.
+    DumpRegister(ghz); // Show the GHZ state in the register.
+    let ghzResults = MResetEachZ(ghz); // Measure and reset qubits.
+
+    use cat5 = Qubit[5]; // Allocate 5 qubits for GHZ state.
+    PrepareCatState(cat5); // Prepare a Cat₅ state in the register.
+    DumpRegister(cat5); // Show the Cat₅ state in the register.
+    let catResults = MResetEachZ(cat5); // Measure and reset qubits.
```
swernli, line 34: I'm not convinced there is value in doing this twice. `PrepareGHZState` should not have the assert on size 3 since GHZ states can be established on any size array, and then there is no difference between the two. I would simplify this sample to just show one of the cases, either 3 or 5, and simplify the file name and operation names here to use either GHZ or Cat. Then it can just be one line in the descriptive comment above main that clarifies that the terms are equivalent.

samples/getting_started/JointMeasurement.qs
```diff
@@ -1,11 +1,12 @@
...
-/// Joint Measurement
+import Std.Diagnostics.*;
+
+/// # Summary
+/// Joint Measurement sample
 ///
 /// # Description
-/// Joint measurements, also known as Pauli measurements, are a generalization
+/// This Q# program demonstrates how to use Joint measurements.
+/// Joint measurement, also known asPauli measurements are a generalization
```
swernli, line 8: Some spacing got messed up here.

```suggestion
/// Joint measurement, also known as Pauli measurements, are a generalization
```

samples/getting_started/Measurement.qs
```diff
@@ -1,28 +1,33 @@
...
 /// generally changes the quantum state that describes that system.
-///
-/// In Q#, the result of a measurement is a value of the type `Result`, that is,
-/// `One` or `Zero`.
-///
-/// This Q# program exemplifies different types of measurements.
-import Std.Measurement.*;
-
 operation Main() : (Result, Result[]) {
+    use q = Qubit(); // Allocate a qubit. Qubit is in |0〉 state after allocation.
```
swernli, line 12: Here too (and in the other files) I think we should stick to the pattern of having comments above the lines they refer to.

samples/getting_started/Measurement.qs
```diff
@@ -1,28 +1,33 @@
...
+    X(q); // Flip the state. Qubit is in |1〉 state now.
+
     // The `M` operation performs a measurement of a single qubit in the
     // computational basis, also known as the Pauli Z basis.
-    use q = Qubit();
     let result = M(q);
+
+    // Reset qubit back to |0〉 state.
+    // An alternative way to measure and then reset a qubit is to use
+    // `MResetZ` operation.
```
swernli, line 21: I'm not sure we should be leaving this to a comment. I would rather explicitly show the `MResetZ` and `MResetEachZ` operations so that it can more easily be picked up both by users and AI training that this is the efficient alternative to `M` followed by `Reset`.

samples/getting_started/QuantumHelloWorld.qs
```diff
@@ -0,0 +1,19 @@
+/// # Summary
+/// Quantum Hello World sample
+///
+/// # Description
+/// This is one of the simplest Q# programs that contains quantum part.
+/// This code prints the message then allocates a qubit and immediately measures it.
+/// Since the qubit starts in |0〉 state such measurement will always yield `Zero`.
+operation Main() : Result {
+    // Print the message. Note that some quantum computers may skip this step.
```
swernli, line 9: I'm not sure what to make of the statement "some quantum computers may skip this step." At this point, we know that all of them will, so we should either make a stronger statement about how the `Message` function produces stdout and is ignored when compiling for quantum systems (perhaps beyond the scope of a hello world sample) or just remove this part of the comment and focus on this being an example of producing stdout from Q#.

samples/getting_started/RandomBit.qs
```diff
@@ -0,0 +1,20 @@
...
+    use qubit = Qubit();
+
+    // Set the qubit in superposition by applying a Hadamard transformation.
+    H(qubit);
+
+    // Measure then reset the qubit.
+    // There is a 50% probability of measuring either `Zero` or `One`.
+    let result = MResetZ(qubit);
+
+    result
```
swernli, line 19: nitpick: I'm not sure there is value in binding the output just to do an implicit return here. I think for this kind of sample I'd prefer either an implicit return with just the call, as in `MResetZ(qubit)`, or an explicit return with just the call, as in `return MResetZ(qubit);`

samples/getting_started/Superposition.qs
```diff
@@ -1,22 +1,22 @@
...
-    use qubit = Qubit();
 
     // Set the qubit in superposition by applying a Hadamard transformation.
     H(qubit);
 
-    // Measure the qubit. There is a 50% probability of measuring either
-    // `Zero` or `One`.
-    let result = M(qubit);
+    // Show the quantum state (when running on a simulator).
+    Std.Diagnostics.DumpMachine();
```
swernli, line 17: It looks like the only difference between this sample and the "RandomBit.qs" sample is the presence of this `DumpMachine`. Is there a reasonable way we can combine those? My first instinct is to drop RandomBit.qs since now there is a RandomBits sample that produces 5 bits and covers the same concepts, and leave this sample as a basic one for the concept of superposition.

samples/getting_started/RandomBitsNISQ.qs
```diff
@@ -0,0 +1,31 @@
+/// # Summary
+/// Simple Quantum Random Number Generator sample
+///
+/// # Description
+/// This program implements a quantum random number generator by setting qubits
+/// in superposition and then using the measurement results as random bits.
+/// This is equivalent to generating a random number in the range of 0..2ᴺ-1.
+operation Main() : Result[] {
```
swernli, line 8: I don't think this file needs the "NISQ" suffix. We generally only use that to differentiate between two versions of the same sample/algorithm when one can run in Base profile and the other cannot. There isn't a non-NISQ RandomBits.qs to differentiate this one from, so this one can use that name.

samples/getting_started/SimpleTeleportation.qs
```diff
@@ -0,0 +1,51 @@
...
+    if b1 {
+        Z(qBob);
+    }
+    if b2 {
+        X(qBob);
+    }
+
+    // Make sure that the obtained result is the same
+    X(qBob);
+    let correct = Std.Diagnostics.CheckZero(qBob);
```
swernli, line 40: Hmm... using `CheckZero` here makes the verification deterministically correct, but I worry it might be more confusing than just using measurement. Is there a reason to not use measurement here without the `X` and check that the expected `One` is returned?

samples/getting_started/CatStatesAndGHZ.qs
ScottCarda-MS: I'm not really sure I understand how the GHZ and the Bell states are related. The GHZ is a superposition between the "All" or "Nothing" basis states, but only one of the 4 Bell states is like that. Bell states, to me, look a lot more like the collection of states in superposition between states of the same parity, where parity is the "even" or "odd" count of |1⟩'s. Thinking of them like that, Bell states _could_ be generalizable to any number of qubits, but that might be called something else, and the GHZ would not qualify under that as it is a superposition between an "even" state and ad "odd" state for all odd numbers of qubits.

samples/getting_started/QuantumHelloWorld.qs
ScottCarda-MS: I'm in favor of the latter, for removal.

samples/getting_started/BellStates.qs
DmitryVasilevsky: Although I like to place short comments at the end of a statement, I agree that most of our comments are placed before the corresponding line. I changed most comments to precede the corresponding line. Only left a few that shows the state after statement.

samples/getting_started/Superposition.qs
```diff
@@ -1,22 +1,22 @@
...
 
-    // Reset the qubit so it can be safely released.
-    Reset(qubit);
-    return result;
+    // Measure the qubit. Reset the qubit. Return the result.
```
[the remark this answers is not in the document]
DmitryVasilevsky: I would prefer to keep 'return the result' in a separate sentence. While the first two actions are part of the MResetZ, the last action is achieved by the virtue of placing this statement at the end of the function.
If you are not OK with the updated version, please reactivate.

samples/getting_started/RandomBitsNISQ.qs
DmitryVasilevsky: Agreed. This was a result of renaming QRNGNISQ (yuck!) and now it doesn't make sense to have NISQ any longer. Renamed.

samples/getting_started/Superposition.qs
DmitryVasilevsky: Probably a good idea to drop RandomBit.qs altogether. I liked having a separate stand-alone sample for a random bit, but after I changed RandomBits to include a dedicated function GenerateRandomBit there's not much need for a separate Random bit...
I removed it.

samples/getting_started/QuantumHelloWorld.qs
DmitryVasilevsky: I don't fully agree that we should drop it entirely. This is one of the first samples (if not the very first sample). If a person submits it to hardware, there will be no message, and it could be a surprise. I'll change this to 'when running on a simulator'. This is true (and logically doesn't preclude it from appearing when running elsewhere).

samples/getting_started/Measurement.qs
DmitryVasilevsky: OK, let's stir users towards using MResetZ and MResetEachZ rather than separate M and Reset. I updated the sample. Check the new comment.
Originally it had just measurements presumably because M is simpler to understand, and some users may be familiar with measurements from theory...

samples/getting_started/SimpleTeleportation.qs
DmitryVasilevsky: The reason is the determinism you mentioned. Just one execution is enough to make sure that the resulting state is correct. Also, if I use measurements, I need to change the message. I can no longer assert that the **state** is correct, I can only say that the **measurement result** is correct (and state may still be incorrect).
Also, I'm not impressed by teleporting |1> state. Feels like communicating two classical bits defeats the trick in this case. A more complex state would be more interesting to teleport...

samples/getting_started/CatStatesAndGHZ.qs
DmitryVasilevsky: Updated with new comments. Check it out one more time.

samples/getting_started/CatStatesAndGHZ.qs
DmitryVasilevsky: Left only one sample.

swernli (approved):

PR #2241, opened 2025-03-19
Added copy button to code blocks in copilot

billti:
This adds a button to copy to clipboard to the top right of code blocks, e.g.

image

### DIFF
--- a/vscode/src/copilot/webview/copilot.tsx
+++ b/vscode/src/copilot/webview/copilot.tsx
@@ -17,6 +17,8 @@ import "./copilot.css";
 // Set up the Markdown renderer with KaTeX support
 import mk from "@vscode/markdown-it-katex";
 import hljs from "highlight.js/lib/core";
+import CopyButtonPlugin from "highlightjs-copy";
+import "highlightjs-copy/styles/highlightjs-copy.css";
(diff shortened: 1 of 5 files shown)


vscode/src/copilot/webview/copilot.tsx
```diff
@@ -248,10 +240,21 @@ function UserMessage(props: { content: string }) {
  * An assistant message in the chat history.
  */
 function AssistantMessage(props: { content: string }) {
+  // Highlight.js needs to be called on the code blocks after they are present in the
```
minestarks, line 243: Random side note: highlight.js and Q# highlighting should probably be part of the `Markdown` element because it would be nice to have in other places we use Markdown, like API documentation.

Not asking for changes, just thought I'd note before I forget.

minestarks (approved):

vscode/src/copilot/webview/copilot.tsx
billti: I thought about that, but you can't just have a `ref` on the `<Markdown>` element itself, as it's not a DOM node, and I didn't want to go messing with the internals of the Markdown control and have to make a bigger change (in other places it's used) just for this.

PR #2243, opened 2025-03-19
Fix code highlighting in copilot

billti:
This improves the syntax highlighting for Q# (and Python by way of improved styling) in the Copilot pane. It now closely matches the syntax highlighting in VS Code in the 'Modern' (default) themes.

Example in dark mode

To entangle 3 qubits, you can create a Greenberger-Horne-Zeilinger (GHZ) state

Same in light mode

To entangle 3 qubits, you can create a Greenberger-Horne-Zeilinger (GHZ) state


vscode/src/copilot/webview/hlsj-qsharp.ts
```diff
@@ -1,78 +1,89 @@
 // Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
billti: I didn't create the file, so it was already incorrect. But yes, you're right. I can rename it also.

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+      "\\b(not|and|or)\\b|\\b(w/)|(=)|(!)|(<)|(>)|(\\+)|(-)|(\\*)|(\\/)|(\\^)|(%)|(\\|)|(\\&\\&\\&)|(\\~\\~\\~)|(\\.\\.\\.)|(\\.\\.)|(\\?)",
+  };
+
+  const QSHARP_NUMBER_MODE: Mode = {
+    scope: "number",
+    begin: "\\b[\\d_]*\\.?[\\d_]\\b",
+  };
+
+  const QSHARP_FN_CALL: Mode = {
+    scope: "title",
```
minestarks, line 101: Suggest `title.function` (see https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html)

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+    begin: /\$"/,
+    end: '"',
+    contains: [
+      {
+        scope: "subst",
+        begin: /\{/,
+        end: /\}/,
+        keywords: QSHARP_KEYWORDS,
+        contains: [
+          /* add later */
```
minestarks, line 122: Looks like this is missing the list of modes (should be largely a copy of the list below, except with `QSHARP_INTERP_STRING_MODE` replaced by `"self"`)

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return {
+    name: "qsharp",
+    case_insensitive: false,
+    keywords: QSHARP_KEYWORDS,
+    contains: [
+      hljs.C_LINE_COMMENT_MODE,
+      hljs.QUOTE_STRING_MODE,
+      QSHARP_INTERP_STRING_MODE,
+      QSHARP_OPERATOR_MODE,
```
minestarks, line 137: Why did the list order change? I think `QSHARP_INTERP_STRING_MODE` needs to be above `hljs.QUOTE_STRING_MODE` so it gets matched first.

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+      QSHARP_NUMBER_MODE,
+      QSHARP_FN_CALL,
+      QSHARP_PUNCTUATION,
```
minestarks, line 140: I prefer the suffix `_MODE` to stick to the convention hljs seems to have established

```suggestion
      QSHARP_FUNCTION_TITLE_MODE,
      QSHARP_PUNCTUATION_MODE
```

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  const QSHARP_FN_CALL: Mode = {
+    scope: "title",
+    match: /\b[a-zA-Z_][a-zA-Z0-9_]*\s*(?=\()/,
+  };
+
+  const QSHARP_PUNCTUATION = {
+    match: /[;:(){},]/,
+    className: "punctuation",
+    relevance: 0,
```
minestarks, line 108: `relevance` shouldn't be.... relevant in syntax highlighting. It's used for language detection, which we never use this library for.

vscode/src/copilot/webview/hljs-qsharp.ts
minestarks: Does this match the closing brace too?

vscode/src/copilot/webview/hljs-qsharp.ts
billti: Cut/paste from another language definition. I can remove it. Does it make a difference if it's there though?

vscode/src/copilot/webview/hljs-qsharp.ts
billti: I was using their codebase as a reference. They do not have that convention (e.g. see https://github.com/highlightjs/highlight.js/blob/main/src/languages/javascript.js#L469 )

vscode/src/copilot/webview/hljs-qsharp.ts
billti: No, it also doesn't match the opening brace (it's a lookahead). Again, matching their codebase for JavaScript here (which I assume is one of their most used and best tested languages). See https://github.com/highlightjs/highlight.js/blob/main/src/languages/javascript.js#L392

vscode/src/copilot/webview/hljs-qsharp.ts
billti: I tested. The order doesn't matter. (If it's trying to match the next token/chars that start with `$"`, then QUOTE_STRING isn't going to match even if first in the list as it doesn't start with a $)

vscode/src/copilot/webview/hljs-qsharp.ts
minestarks: Oops, I meant to ask "opening brace", and you answered the question I meant to ask, thanks!

vscode/src/copilot/webview/hljs-qsharp.ts
minestarks: Exactly, that's why I decided against going down the rabbit hole. Once you decide support certain things, you have to support expressions overall.

I felt that it interpolated strings are important since the samples contain quite a few. They never contain anything more than a single identifier though. Can't we just live with it?

vscode/src/copilot/webview/hljs-qsharp.ts
minestarks: Makes sense, I'd misunderstood the order thing.

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+  "Bool",
+  "Qubit",
+  "Pauli",
+  "Result",
+  "Range",
+  "String",
+  "Unit",
+  "Ctl",
+  "Adj",
+  "is",
```
minestarks, line 67: These categories look fairly loose, but I'd probably put `is` under "decl" rather than "type".

(I'm ok if you disagree and leave it - if we were to start nitpicking, we'd probably end up defining like twelve separate categories. Let's not.)

vscode/src/copilot/webview/hljs-qsharp.ts
```diff
@@ -0,0 +1,143 @@
...
+  "newtype",
+  "struct",
+  "operation",
+  "function",
+  "new",
+  "body",
+  "adjoint",
+  "Adjoint",
+  "controlled",
+  "Controlled",
```
minestarks, line 46: I'd suggest putting capital `Adjoint` and `Controlled` under a different category, maybe "general". For one, they're not declaration keywords, they're operators. For two, I find it too easy to get mixed up between the lowercase/uppercase versions, and if they have different coloring that may at least be another visual hint that they're different.

minestarks (approved):

vscode/src/copilot/webview/hljs-qsharp.ts
billti: Think the latest change fixes it:

image

vscode/src/copilot/webview/hljs-qsharp.ts
billti: Done

vscode/src/copilot/webview/hljs-qsharp.ts
billti: Copied from our VS Code syntax. Would rather keep consistent for now <https://github.com/microsoft/qsharp/blob/main/vscode/syntaxes/qsharp.tmLanguage.json#L149>

vscode/src/copilot/webview/hljs-qsharp.ts
billti: Copied from our VS Code syntax <https://github.com/microsoft/qsharp/blob/main/vscode/syntaxes/qsharp.tmLanguage.json#L165>. I'd rather keep consistent for now.

PR #2245, opened 2025-03-20
Add spinner and fix histogram spacing

billti:
Used the VS Code loading spinner when busy

https://github.com/user-attachments/assets/6c5c2dc4-7dbc-4eb4-8419-37df56878034

Also fixed the spacing at the bottom of this histogram:

Before

image

After

image

Also added the 'bash' language in code highlighting, as sometimes commands or plain text try to use this language (and the console was showing warnings if it wasn't registered).

### DIFF
--- a/vscode/src/copilot/webview/copilot.tsx
+++ b/vscode/src/copilot/webview/copilot.tsx
@@ -35,7 +35,7 @@ import { WebviewApi } from "vscode-webview";
 
 const vscodeApi: WebviewApi<ChatElement[]> = acquireVsCodeApi();
 
-// Only include the Python and Q# languages so as not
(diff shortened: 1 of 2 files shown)


vscode/src/copilot/webview/copilot.tsx
```diff
@@ -37,6 +38,7 @@ const vscodeApi: WebviewApi<ChatElement[]> = acquireVsCodeApi();
 // Only include the Python and Q# languages so as not
```
minestarks, line 38:
```suggestion
// Only include a small set of languages so as not
```

minestarks (approved):

PR #2254, opened 2025-03-26
Compile gphase, gatedef, and control flow statements

orpuente-MS:
This PR completes the lowering and compilation for:
 - [x] gate definitions
 - [x] gphase
 - [x] for stmt
 - [x] if stmt
 - [x] switch stmt
 - [x] return stmt
 - [x] classical function calls
 - [x] classical function decls

It also does some partial work to make the compiler infallible, by returning `ExprKind::Err` instead of None when compiling expressions.

### DIFF
--- a/compiler/qsc_qasm3/src/compiler.rs
+++ b/compiler/qsc_qasm3/src/compiler.rs
@@ -10,21 +10,22 @@ use qsc_frontend::{compile::SourceMap, error::WithSource};
 use crate::{
     ast_builder::{
         build_arg_pat, build_array_reverse_expr, build_assignment_statement, build_barrier_call,
-        build_binary_expr, build_cast_call, build_cast_call_two_params, build_classical_decl,
-        build_complex_from_expr, build_convert_call_expr, build_expr_array_expr, build_for_stmt,
-        build_gate_call_param_expr, build_gate_call_with_params_and_callee, build_gate_decl_lambda,
-        build_if_expr_then_block, build_if_expr_then_block_else_block,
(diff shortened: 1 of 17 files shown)


compiler/qsc_qasm3/src/compiler.rs
```diff
@@ -387,8 +398,7 @@ impl QasmCompiler {
             return None;
         }
```
idavis, line 399: delete

compiler/qsc_qasm3/src/compiler.rs
```diff
@@ -752,12 +881,15 @@ impl QasmCompiler {
...
     }
 
-    fn compile_expr(&mut self, expr: &semast::Expr) -> Option<qsast::Expr> {
+    fn compile_expr(&mut self, expr: &semast::Expr) -> qsast::Expr {
         match expr.kind.as_ref() {
             semast::ExprKind::Err => {
                 // todo: determine if we should push an error here
```
idavis, line 887: we can remove this comment as we now know we are trying to compile

compiler/qsc_qasm3/src/compiler.rs
```diff
@@ -1136,22 +1287,22 @@ impl QasmCompiler {
             &Type::Angle(..) => {
                 let msg = "Cast bit to angle";
                 self.push_unimplemented_error_message(msg, expr.span);
```
idavis, line 1289: Given that the calling code now pushes a semantic error when ::ExprKind::Err is returned, we are pushing both unimplemented and unsupported errors for the same cases.

PR #2255, opened 2025-03-26
Use fresh env for `qsharp.run` with Python interop functions wrapping Q# operations

### LINKED ISSUE #2253: `qsharp.run` fails with Python interop functions wrapping Q# operations
When using `qsharp.run` to execute multiple shots of an operation, I should be able to pass the Python `Callable` wrapper around the operation and have it work. Instead, this triggers a "qubit used after release" error incorrectly:

Image

swernli:
This matches the behavior used for other places when a fresh simulator instance is paired with a fresh environment for isolated execution.

Fixes #2253

### DIFF
--- a/compiler/qsc/src/interpret.rs
+++ b/compiler/qsc/src/interpret.rs
@@ -567,8 +567,8 @@ impl Interpreter {
         })
     }
 
-    // Invokes the given callable with the given arguments using the current environment and compilation but with a fresh
-    // simulator configured with the given noise, if any.
+    // Invokes the given callable with the given arguments using the current compilation but with a fresh
+    // environment and simulator configured with the given noise, if any.
     pub fn invoke_with_noise(
         &mut self,
         receiver: &mut impl Receiver,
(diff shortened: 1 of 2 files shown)


pip/tests/test_qsharp.py
```diff
@@ -431,7 +431,21 @@ def test_run_with_result(capsys) -> None:
...
+    assert results == [qsharp.Result.Zero, qsharp.Result.Zero, qsharp.Result.Zero]
+    stdout = capsys.readouterr().out
+    assert stdout == "Hello, world!\nHello, world!\nHello, world!\n"
+
+
+def test_run_with_result_from_callable_while_global_qubits_allocated(capsys) -> None:
+    qsharp.init()
+    qsharp.eval("use q = Qubit();")
+    qsharp.eval(
+        'operation Foo() : Result { Message("Hello, world!"); use q = Qubit(); M(q) }'
```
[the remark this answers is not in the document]
swernli: the test verifies a bug fix. Without the fix, this test case fails with an error that a qubit is used after release (it was a reference counting problem). The linked issue has the failure details.

compiler/qsc/src/interpret.rs
```diff
@@ -567,8 +567,8 @@ impl Interpreter {
         })
     }
 
-    // Invokes the given callable with the given arguments using the current environment and compilation but with a fresh
-    // simulator configured with the given noise, if any.
+    // Invokes the given callable with the given arguments using the current compilation but with a fresh
+    // environment and simulator configured with the given noise, if any.
     pub fn invoke_with_noise(
```
minestarks, line 572: Does this change the behavior of invoking `qsharp.code.*` at all? Or is this only used for `.run()`?

swernli: This path is only used by `qsharp.run` when a callable is passed in. So now we (correctly) only use `self.env` in two places: `eval_increment` which is what cells and `qsharp.eval` sit on top of, and `invoke` which is what actually calling those python wrapper functions uses. Everything else (circuits, running shots, resource estimation, partial eval for QIR) uses fresh environments.

minestarks (approved):

ScottCarda-MS (approved):

PR #2272, opened 2025-04-01
Unignore unit tests; lower and compile missing items

orpuente-MS:
This PR:
- [x] Unignores 32/52 unit tests that are now supported.
- [x] Passes a couple of spans to the ast_builder that were previously set to default.
- [x] Makes constant measurements like in `const bit res = measure q;` a parser error.
- [x] Only generates gate functors when `@SimulatableIntrinsic` is not present.
- [x] break stmt
- [x] continue stmt
- [x] lower duration literals
- [x] fixes bug when handling `void` return type
- [x] return stmt casts to the return type of the function


compiler/qsc_qasm3/src/semantic/tests/statements/continue_stmt.rs
```diff
@@ -0,0 +1,146 @@
...
+            Program:
+                version: <none>
+                statements:
+                    Stmt [0-9]:
+                        annotations: <empty>
+                        kind: Err
+
+            [Qsc.Qasm3.Compile.InvalidScope
+
+              x continue can only appear in loop scopes.
```
idavis, line 84: this message is onconsistent with braek error of the same time. Do they have different legal locations?

compiler/qsc_qasm3/src/semantic/lowerer.rs
```diff
@@ -1152,7 +1174,7 @@ impl Lowerer {
         let set_declaration = self.lower_enumerable_set(&stmt.set_declaration);
 
         // Push scope where the loop variable lives.
-        self.symbols.push_scope(ScopeKind::Block);
+        self.symbols.push_scope(ScopeKind::Loop);
```
idavis, line 1177: there is no loop scope: https://openqasm.com/language/scope.html#scoping-of-variables

PR #2278, opened 2025-04-07
Cyclic and multiple include errors

orpuente-MS:
This PR adds cyclic and multiple include errors to qasm


compiler/qsc_qasm3/src/parser.rs
```diff
@@ -230,13 +230,22 @@ impl QasmSource {
...
 {
     match resolver.resolve(&path) {
-        Ok((path, source)) => parse_qasm_source(source, path, resolver),
+        Ok((path, source)) => {
+            let parse_result = parse_qasm_source(source, path, resolver);
+
+            // Once we finish parsing the source, we pop the file from the
+            // resolver. This is needed to keep track of multiple includes
+            // and cyclic includes.
+            resolver.ctx().pop_current_file();
```
idavis, line 245: should this be done in `parse_qasm_source` instead?

PR #2284, opened 2025-04-09
Remove external parser. Fix Qiskit's failure to emit gates.

idavis:


pip/qsharp/interop/qiskit/backends/backend_base.py
```diff
@@ -183,7 +183,7 @@ def __init__(
 
         # selt the default options for the exporter
```
orpuente-MS, line 184: selt -> set

pip/tests-integration/interop_qiskit/test_gateset_qasm.py
```diff
@@ -24,9 +24,10 @@ def run_transpile_test(
     # remove the first four lines, which are the header
```
orpuente-MS, line 24: four -> five
```suggestion
    # remove the first five lines, which are the header
```

Issue #944, opened 2023-12-14
Multi-file: Runtime error spans are incorrectly mapped to source files


minestarks:
The bug exists in `run()` in the wasm layer. When the runtime error contains a span, and the span is coming from a file in the program that's _not_ the first one, the wrong span is reported.

I don't think this span is used in the UX anywhere but since it's such a general function it may bite us in the future if we decide to surface these spans.

sezna:
I need to see if this is still happening after all of the multi-file changes. Assigning myself.

PR #2291, opened 2025-04-11
Fix Test Explorer issues

billti:
This PR fixes two issues.

Firstly, Q# code used in chat contexts was showing up in the Test Explorer as additional documents (and possibly other locations). It seems to use a few schemes all starting with 'chat', so excluding those.

Secondly, the location given for failing tests was often wrong, and falling back to the first 'related' location span when we're also using that for the document uri anyway resolves that issue. (And seems like the correct thing to do anyway).

### DIFF
--- a/vscode/src/common.ts
+++ b/vscode/src/common.ts
@@ -22,7 +22,7 @@ export function isQsharpDocument(document: TextDocument): boolean {
     (Utils.extname(document.uri) === ".qs" || document.isUntitled) &&
     document.uri.scheme !== "git" &&
     document.uri.scheme !== "pr" &&
-    // The Copilot Chat window also creates documents with various schemes that start 
+    // The Copilot Chat window also creates documents with various schemes that start
     // with "chat", such as "chat-editing-text-model" and others.
     !document.uri.scheme.startsWith("chat")
   );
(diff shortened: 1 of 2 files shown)


vscode/src/common.ts
```diff
@@ -21,7 +21,8 @@ export function isQsharpDocument(document: TextDocument): boolean {
     document.languageId === qsharpLanguageId &&
     (Utils.extname(document.uri) === ".qs" || document.isUntitled) &&
     document.uri.scheme !== "git" &&
-    document.uri.scheme !== "pr"
+    document.uri.scheme !== "pr" &&
+    !document.uri.scheme.startsWith("chat")
```
DmitryVasilevsky, line 25: Schemes may vary in this case? Maybe add a comment.

minestarks (approved):

idavis (approved):

billti:
Note: related to https://github.com/microsoft/qsharp/issues/944

vscode/src/common.ts
billti, line 25:
```suggestion
    // The Copilot Chat window also creates documents with various schemes that start 
    // with "chat", such as "chat-editing-text-model" and others.
    !document.uri.scheme.startsWith("chat")
```

PR #2311, opened 2025-04-17
Swap Control and Target

ScottCarda-MS:
In circuit editing, when moving a control onto a target or vise-versa, the control and target should swap positions.

Swap control target

This also fixes an issue with copy-moving controls. The behavior is now that it will create a new control for the same operation and possibly move the operation horizontally in the circuit.

Copy-move control

### DIFF
--- a/npm/qsharp/ux/circuit-vis/circuitManipulation.ts
+++ b/npm/qsharp/ux/circuit-vis/circuitManipulation.ts
@@ -129,25 +129,47 @@ const _moveY = (
   movingControl: boolean,
 ): void => {
   // Check if the source operation already has a target or control on the target wire
-  let registers: Register[];
+  let targets: Register[];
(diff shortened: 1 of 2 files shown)


minestarks (approved):
Nice!

DmitryVasilevsky:
I'm not sure that moving control to a different line should add another control (your second example). This seems a bit counterintuitive. We should probably discuss it, but the change can go in if needed.

DmitryVasilevsky (approved):

ScottCarda-MS:
> I'm not sure that moving control to a different line should add another control (your second example). This seems a bit counterintuitive. We should probably discuss it, but the change can go in if needed.

This is specifically the copy-move with holding down the ctrl key. This copy-moves other gates as well. Regular moving a control will not copy the control.

PR #2315, opened 2025-04-18
Retain circuit editor when hidden

billti:
This avoid the delay when switching away from the circuit editor and then back while it reloads the UX.


ScottCarda-MS (approved):
Tried it out an works well. Thanks!

orpuente-MS (approved):

PR #2321, opened 2025-04-21
Added RoundHalfAwayFromZero to standard library

DmitryVasilevsky:
* Added RoundHalfAwayFromZero to standard library.
* Simplified Round (which is actually rounding half towards zero)
* Updated comments and added reference to Wikipedia
* Added tests
* Changed QasmStd to use this new function instead of separate implementation

### DIFF
--- a/library/std/src/Std/Math.qs
+++ b/library/std/src/Std/Math.qs
@@ -396,16 +396,27 @@ function Floor(value : Double) : Int {
 }
 
 /// # Summary
-/// Returns the nearest integer to the specified number.
-/// For example: Round(3.7) = 4; Round(-3.7) = -4
+/// Returns the nearest integer to the specified number. Half is rounded towards zero.
+/// For example: Round(3.7) = 4; Round(-3.7) = -4; Round(3.5) = 3;
+///
+/// # References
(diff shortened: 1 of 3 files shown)


idavis (approved):

library/std/src/Std/Math.qs
```diff
@@ -396,16 +396,27 @@ function Floor(value : Double) : Int {
 }
 
 /// # Summary
-/// Returns the nearest integer to the specified number.
-/// For example: Round(3.7) = 4; Round(-3.7) = -4
+/// Returns the nearest integer to the specified number. Half is rounded towards zero.
+/// For example: Round(3.7) = 4; Round(-3.7) = -4; Round(3.5) = 3;
+///
+/// # References
+/// [Wikipedia article - Rounding](https://en.wikipedia.org/wiki/Rounding#Rounding_half_toward_zero)
```
ScottCarda-MS, line 403: Just curious what influenced the choice for the default behavior for the Round function to be to round half toward zero?

swernli (approved):

library/std/src/Std/Math.qs
ScottCarda-MS: discussed offline

DmitryVasilevsky: Old QDK behaved like that.

PR #2324, opened 2025-04-22
Added ApplyOperationPower to Std.Canon

DmitryVasilevsky:
* Added ApplyOperationPower that applies operation $U^n$, where $n$ is an integer. If $n$ is negative, the adjoint of $U$ is used. If $n$ is 0, the operation is not applied.
* Used this operation in StdQasm.
* Seems to be useful in general.

### DIFF
--- a/library/std/src/Std/Canon.qs
+++ b/library/std/src/Std/Canon.qs
@@ -583,6 +583,25 @@ operation ApplyXorInPlaceL(value : BigInt, target : Qubit[]) : Unit is Adj + Ctl
     }
     adjoint self;
 }
+
+/// # Summary
+/// Applies operation `u` `power` times.
+/// If `power` is negative, the adjoint of `u` is used.
+/// If `power` is 0, operation `u` is not applied.
+///
+/// # Description
+/// Applies $U^n$ where $n$ = `power`.
+operation ApplyOperationPower(power: Int, u: Unit => Unit is Adj) : Unit is Adj {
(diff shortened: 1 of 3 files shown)


library/std/src/Std/Canon.qs
```diff
@@ -583,6 +583,22 @@ operation ApplyXorInPlaceL(value : BigInt, target : Qubit[]) : Unit is Adj + Ctl
     }
     adjoint self;
 }
+
+/// # Summary
+/// Applies operation `op` to the `target` `power` times.
+/// If `power` is negative, the adjoint of `op` is used.
+/// If `power` is 0, the operation `op` is not applied.
+operation ApplyOperationPowerA<'T>(
```
idavis, line 591: Why do we need the `A` suffix?

DmitryVasilevsky: It's a convention that we use in certain cases. For example, compare ApplyToEachA. We pass an operation `op` that is adjointable and this operation is also adjointable. **It probably doesn't matter now since we only have one operation**, but I anticipate that we will also add a CA version.
Assuming we have two versions A and CA, one won't be able to use ApplyOperationPowerCA version if a function 'op' is just A. But in a context that would require CA, only ApplyOperationPowerCA could be used.
This is something we discuss on occasions, but so far it's a Q# peculiarity.

swernli (approved):

orpuente-MS (approved):

PR #2327, opened 2025-04-22
Use LaTeX fonts for math symbols in circuit diagrams

billti:
Circuit rendering now appears as per the below in VS Code (and have tested in JupyterLabs in both Safari and Edge browsers, and in light and dark themes).

image


orpuente-MS (approved):

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -359,15 +387,15 @@ const _x = (metadata: Metadata, _: number): SVGElement => {
 const _ket = (label: string, metadata: Metadata): SVGElement => {
   const { x, targetsY, width } = metadata;
   const gate = _unitary(
-    `|${label}〉`,
+    `|${label}⟩`,
```
ScottCarda-MS, line 390: It is hard to tell by eye-balling it. Are these the characters we want to use for the bar and angle-bracket? I see you've changed the bars in the comment above but not here, and you changed the bracket here but not above.

npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
```diff
@@ -71,38 +71,15 @@ const formatInputs = (
...
   const el: SVGElement = text("", leftPadding, y, 16);
 
-  // Create the main text node
-  const mainText = document.createElementNS(
-    "http://www.w3.org/2000/svg",
-    "tspan",
-  );
-  mainText.textContent = "|𝜓";
+  const subtext = subscript
+    ? `<tspan baseline-shift="sub" font-size="65%">${subscript}</tspan>`
```
ScottCarda-MS, line 75: Was this done to avoid using the `document.createElementNS` calls? The function calls seem more "official" than hardcoding the string, but is there a reason to prefer the hardcoded string?

npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
```diff
@@ -71,38 +71,15 @@ const formatInputs = (
...
-  const closingText = document.createElementNS(
-    "http://www.w3.org/2000/svg",
-    "tspan",
-  );
-  closingText.textContent = "⟩";
-
-  // Append all parts to the main SVG text element
-  el.appendChild(mainText);
-  el.appendChild(closingText);
+  el.innerHTML = `|<tspan class="qs-mathtext">ψ</tspan>${subtext}⟩</tspan>`;
```
ScottCarda-MS, line 78: Could be fine, but this is another place to make sure we are using the specific characters we want for the bar and bracket.

npm/qsharp/ux/circuit-vis/metadata.ts
```diff
@@ -15,7 +15,7 @@ export enum GateType {
   Swap,
   /** X gate. */
   X,
-  /** |0〉 or |1〉 gate. */
+  /** |0⟩or |1⟩ gate. */
```
ScottCarda-MS, line 18: Another place to check characters.

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
billti: See from line 330 in util.ts. I checked all the files and made them consistent.

npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
billti: See from line 330 in util.ts. I checked all the files and made them consistent.

npm/qsharp/ux/circuit-vis/metadata.ts
billti: See from line 330 in util.ts. I checked all the files and made them consistent.

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
ScottCarda-MS: Actually, I see you have a mathChars table in utils.ts now. We could use that instead to enforce consistency.

npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
ScottCarda-MS: We could use the mathChars table from utils.ts for the characters, including the psi.

npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
billti: It's simpler/shorter/easier to read in this case. Either works.

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
billti: I thought it might just be a little unreadable/busy to use some property of an import interpolated everywhere we have a special char, but could try and see how it looks.

ScottCarda-MS (approved):
Just left a couple of questions. It'd be nice to make use of that mathChars table you added to ensure consistency with the characters.

ScottCarda-MS:
Also, there is a regex [here](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_circuit/src/json_to_circuit.rs#L230) that checks against a variety of possible bar and bracket character use. It looks like you added a new bar character that wasn't captured before. Could you add it to the regex?

npm/qsharp/ux/circuit-vis/utils.ts
```diff
@@ -327,6 +327,20 @@ const getGateElems = (container: HTMLElement): SVGGraphicsElement[] => {
...
+// Non-ASCII chars are fraught with danger. Copy/paste these when possible.
+// Use the following regex in VS Code to find invalid unicode chars
+// [^\x20-\x7e\u{03b8}-\u{03c8}\u{2020}\u{2223}\u{27e8}\u{27e9}]
+
+const mathChars = {
+  theta: "θ", // \u{03b8}
+  pi: "π", // \u{03c0}
+  psi: "ψ", // \u{03c8}
+  dagger: "†", // \u{2020}
+  divides: "∣", // \u{2223}
```
ScottCarda-MS, line 339: Actually, it *does* seem like we are using the "vertical line" | everywhere except here and in the comment you edited in gateFormatter.ts. It'd be confusing a probably lead to inconsistencies to keep this in here as it would indicate that this is the character we use for bra-ket notation when it isn't. Alternatively, we could change everywhere where we're using the vertical line, but that'd be a bigger change.

billti: I'm going to take this out. KaTeX uses that special bar char shown here, but when I used it in Jupyter Notebooks (which use a different font) it looked terrible in Safari, so I went back to just the regular ASCII bar char, which looks OK everywhere. I'll remove this.

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
ScottCarda-MS: I think it does indeed make it hard to read when using those references, so I'm fine with them not being used like this.

npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -225,6 +225,31 @@ const _measure = (x: number, y: number): SVGElement => {
...
+  // Style the entire Greek + Coptic block (https://unicodeplus.com/block/0370)
+  // Note this deliberately leaves ASCII digits [0-9] non-italic
+  const italicChars = /[a-zA-Z\u{0370}-\u{03ff}]+/gu;
+
+  label = label.replace(italicChars, `<tspan class='qs-mathtext'>$&</tspan>`);
+
+  // Replace a trailing ' with the proper unicode dagger symbol
+  label = label.replace(
+    /'$/,
+    `<tspan dx="2" dy="-3" style="font-size: 0.8em;">†</tspan>`,
```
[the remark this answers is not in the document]
billti: I've added all the new handling as easy to disable via one var, so will leave the "'" in the original flow and replace along with the other LaTeX handling when enabled.

PR #2328, opened 2025-04-23
Capitalize the Rotation gate names in trace circuits

ScottCarda-MS:
The rotation gates created by the circuit builder were getting lower-case names: `rx`, `rxx` for example. These are inconsistent with the other gate names like `H` and `X`. It is also inconsistent with the Q# equivalent operations `Rx(theta, q)`.

This PR changes the output of the circuit builder to have upper-case names for the rotation gates, making them consistent with the other gate names and the Q# operations.

### DIFF
--- a/compiler/qsc_circuit/src/builder.rs
+++ b/compiler/qsc_circuit/src/builder.rs
@@ -88,35 +88,35 @@ impl Backend for Builder {
 
     fn rx(&mut self, theta: f64, q: usize) {
         let q = self.map(q);
-        self.push_gate(rotation_gate("rx", theta, [q]));
+        self.push_gate(rotation_gate("Rx", theta, [q]));
     }
 
     fn rxx(&mut self, theta: f64, q0: usize, q1: usize) {
         let q0 = self.map(q0);
         let q1 = self.map(q1);
-        self.push_gate(rotation_gate("rxx", theta, [q0, q1]));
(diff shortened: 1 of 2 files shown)


minestarks (approved):
Thanks!

compiler/qsc_circuit/src/builder.rs
```diff
@@ -88,35 +88,35 @@ impl Backend for Builder {
 
     fn rx(&mut self, theta: f64, q: usize) {
         let q = self.map(q);
-        self.push_gate(rotation_gate("rx", theta, [q]));
+        self.push_gate(rotation_gate("Rx", theta, [q]));
```
DmitryVasilevsky, line 91: Do we have a central place for gate names for all of our Rust code? Should we?

DmitryVasilevsky (approved):

compiler/qsc_circuit/src/builder.rs
ScottCarda-MS: It doesn't seem like we do, and I'm not really in favor of adding one. Normally the gate names are a simple "X" or "S", and it would be pretty tedious to have to make a reference to some centralized location in code. I can see the argument that it would make gate name references safer, as demonstrated by the need for fix in this PR, but in general I think it would cause headache and be more trouble than its worth. That's just my take on it though, and I can see that it might just come down to a matter of preference.

PR #2329, opened 2025-04-23
Circuit Code Improvements

ScottCarda-MS:
- Remove lodash dependency
- Rename Metadata in npm to a GateRenderData
- Encapsulate the circuit-editor CSS classes under the circuit-panel class
- Use num_qubits instead of max_q_id

### DIFF
--- a/compiler/qsc_circuit/src/circuit.rs
+++ b/compiler/qsc_circuit/src/circuit.rs
@@ -827,6 +827,11 @@ fn transform_to_col_row(aligned_ops: Vec<Vec<Option<usize>>>) -> Vec<Vec<Option<
 fn group_operations(operations: &[Operation], num_qubits: usize) -> Vec<Vec<usize>> {
     let mut grouped_ops = vec![vec![]; num_qubits];
 
+    let max_q_id = match num_qubits {
+        0 => 0,
+        _ => num_qubits - 1,
+    };
+
     for (instr_idx, op) in operations.iter().enumerate() {
         let ctrls = match op {
             Operation::Measurement(m) => &m.qubits,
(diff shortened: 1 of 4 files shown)


compiler/qsc_circuit/src/circuit.rs
```diff
@@ -856,6 +859,10 @@ fn group_operations(operations: &[Operation], max_q_id: usize) -> Vec<Vec<usize>
                 None => 0,
             }
         };
+        let max_q_id = match num_qubits {
+            0 => 0,
+            _ => num_qubits - 1,
+        };
```
minestarks, line 865: Hm, are you sure about this? Your original implementation didn't special case `0` and it seemed to work just fine. (Or if it didn't work before, could you add a test case to show it?)

Also, you're adding ` to `max_reg_idx` down below, so maybe there's a more straightforward solution, where you don't subtract 1 here only to add it back later?

npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -420,7 +419,9 @@ class CircuitEvents {
     dropzoneElems.forEach((dropzoneElem) => {
       dropzoneElem.addEventListener("mouseup", async (ev: MouseEvent) => {
         const copying = ev.ctrlKey;
-        const originalGrid = cloneDeep(this.componentGrid);
+        const originalGrid = JSON.parse(
+          JSON.stringify(this.componentGrid),
+        ) as ComponentGrid;
```
minestarks, line 424: ~What's the goal of stringifying then parsing?~   Ohh - is a deep clone? Neat trick but I've never seen it before - a quick comment would help confused people like me

npm/qsharp/ux/circuit-vis/utils.ts
```diff
@@ -11,6 +11,38 @@ import {
 import { ComponentGrid, Operation } from "./circuit";
 import { Register } from "./register";
 
+/**
+ * Performs a deep equality check between two objects or arrays.
+ * @param obj1 - The first object or array to compare.
+ * @param obj2 - The second object or array to compare.
+ * @returns True if the objects are deeply equal, false otherwise.
+ */
+export const deepEqual = (obj1: any, obj2: any): boolean => {
```
minestarks, line 20:
```suggestion
export const deepEqual = (obj1: unknown, obj2: unknown): boolean => {
```

This would be a good use of `unknown` over `any`

minestarks (approved):

compiler/qsc_circuit/src/circuit.rs
ScottCarda-MS: This logic is clunky because of the is_classically_controlled case. We awkwardly need several different but related concepts:
- The number of qubits in the circuit
- The maximum id of the qubits in the circuit (num_qubits - 1)
- The minimum register index for the operation
- The maximum register index for the operation

Before we were passing into this function the maximum id of the qubits in the circuit, but now it is taking the number of qubits, which I argue is just a simpler and more straightforward concept. This also allows us to side-step the issue of subtracting 1 from possibly zero qubits causing a runtime underflow error.

npm/qsharp/ux/circuit-vis/events.ts
ScottCarda-MS: Oh this is actually used in a number of places in this code and from before we inherited it. I thought this was a standard, but still good to add a clarifying comment. Done!

compiler/qsc_circuit/src/circuit.rs
ScottCarda-MS: Okay I've tried to clean this up a bit to make it a bit more readable. Hopefully that helps!

PR #2330, opened 2025-04-23
Support Enter and Escape Keys with Prompts

ScottCarda-MS:
- Enables the use of the "Enter" key to affirm input and the "Escape" key to cancel input in argument prompt.
- Enables the use of the "Enter" key to confirm and the "Escape" key to cancel on remove-qubit confirmation prompt.
- Automatically replaces "pi" with the symbol "π" in user input for arguments for better readability.

### DIFF
--- a/npm/qsharp/ux/circuit-vis/contextMenu.ts
+++ b/npm/qsharp/ux/circuit-vis/contextMenu.ts
@@ -238,7 +238,7 @@ const validateExpression = (input: string): boolean => {
     // Validate the remaining expression (without parentheses)
     const sign = "[+-]?";
     const number = "((\\d+(\\.\\d*)?))"; // Matches integers and decimals
-    const pi = "(?i:π|pi)"; // Matches π or pi
+    const pi = "(?i:π)"; // Matches π
     const value = `${sign}(${number}|${pi})`; // Matches a signed number or π
     const operator = "[+\\-*/]"; // Matches arithmetic operators
     const expressionRegex = new RegExp(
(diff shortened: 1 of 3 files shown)


npm/qsharp/ux/circuit-vis/contextMenu.ts
```diff
@@ -325,33 +358,24 @@ const _createInputPrompt = (
...
-  const validateAndToggleOkButton = () => {
-    const isValid = validateInput(inputElem.value.trim());
-    okButton.disabled = !isValid;
+  // Handle Escape key globally while prompt is open
+  const handleGlobalKeyDown = (event: KeyboardEvent) => {
+    if (event.key === "Escape") {
+      event.preventDefault();
+      cancelButton.click();
+    }
   };
```
swernli, line 377: Does this need to be global here? There is already a global handler defined in events.ts, and the previous one in this file is for key events on the input field. Not sure if this is needed or redundant.

npm/qsharp/ux/circuit-vis/contextMenu.ts
```diff
@@ -311,6 +311,39 @@ const _createInputPrompt = (
   const buttonsContainer = document.createElement("div");
   buttonsContainer.classList.add("prompt-buttons");
 
+  // Create the OK button
+  const okButton = document.createElement("button");
+  okButton.classList.add("prompt-button");
+  okButton.textContent = "OK";
+
+  // Function to replace "pi" with "π" (case-insensitive)
+  const replacePiWithSymbol = (input: string) => input.replace(/pi/gi, "π");
```
swernli, line 320: what does the `/gi` mean here? Is that regex handling for case insensitivity or is the string "gi" actually important somehow?

ScottCarda-MS: The `g` means "global" which means that it will match all instances of the pattern in the string, not just the first one. The `i` does as you suggest, case insensitivity.

npm/qsharp/ux/circuit-vis/contextMenu.ts
ScottCarda-MS: This is needed because the other one defined in this file, for the input prompt, as well as this one are both temporary handlers. They are specific to the prompt that creates them and will be removed globally when the prompt is removed. Even if we didn't remove them, we'd still need to create new ones each time because their logic is specific to the prompt instance that creates them and removing them just makes sure to not "leak" orphaned event handlers on the document.

minestarks (approved):

PR #2338, opened 2025-04-25
Support intrinsic `SX` gate

swernli:
This adds full support for an intrinsic `SX` gate by including it in the "canonical" gate set used by the QDK. This includes support across QIR generation, circuits, resource estimation, and Qiskit interop.

### DIFF
--- a/compiler/qsc_eval/src/backend.rs
+++ b/compiler/qsc_eval/src/backend.rs
@@ -65,6 +65,9 @@ pub trait Backend {
     fn s(&mut self, _q: usize) {
         unimplemented!("s gate");
     }
+    fn sx(&mut self, _q: usize) {
+        unimplemented!("sx gate");
+    }
     fn swap(&mut self, _q0: usize, _q1: usize) {
         unimplemented!("swap gate");
     }
@@ -292,6 +295,13 @@ impl Backend for SparseSim {
 
     fn s(&mut self, q: usize) {
         self.sim.s(q);
+        self.apply_noise(q);
+    }
+
+    fn sx(&mut self, q: usize) {
+        self.sim.h(q);
+        self.sim.s(q);
+        self.sim.h(q);
         self.apply_noise(q);
     }
 
@@ -591,6 +601,11 @@ where
     fn s(&mut self, q: usize) {
         self.chained.s(q);
         self.main.s(q);
+    }
+
+    fn sx(&mut self, q: usize) {
+        self.chained.sx(q);
+        self.main.sx(q);
     }
 
     fn swap(&mut self, q0: usize, q1: usize) {
--- a/library/std/src/Std/Intrinsic.qs
+++ b/library/std/src/Std/Intrinsic.qs
@@ -828,6 +828,93 @@ operation S(qubit : Qubit) : Unit is Adj + Ctl {
 }
 
 /// # Summary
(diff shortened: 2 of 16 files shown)


pip/tests-integration/test_base_qir.py
```diff
@@ -109,18 +110,19 @@ def check_call(i: int, name: str, num_args: int) -> None:
...
     check_call(12, "__quantum__qis__s__adj", 1)
-    check_call(13, "__quantum__qis__t__body", 1)
-    check_call(14, "__quantum__qis__t__adj", 1)
-    check_call(15, "__quantum__qis__x__body", 1)
-    check_call(16, "__quantum__qis__y__body", 1)
-    check_call(17, "__quantum__qis__z__body", 1)
-    check_call(18, "__quantum__qis__swap__body", 2)
-    check_call(19, "__quantum__qis__cx__body", 2)
-    check_call(20, "__quantum__qis__m__body", 2)
+    check_call(13, "__quantum__qis__sx__body", 1)
```
DmitryVasilevsky, line 113: Does this mean that whatever executes this QIR should now handle SX?

swernli: Technically, QIR does not proscribe a particular gate list, but this does add it to the "canonical" set of gates used by the QDK as mentioned in the PR description. This works with the service and is handled by other tools like qir-runner: https://github.com/qir-alliance/qir-runner/pull/221.

compiler/qsc_eval/src/backend.rs
```diff
@@ -295,6 +298,13 @@ impl Backend for SparseSim {
         self.apply_noise(q);
     }
 
+    fn sx(&mut self, q: usize) {
+        self.sim.h(q);
```
idavis, line 302: having sx on sim would be preferred.

swernli: That `sim` variable is the `QuantumSim` struct from [qir-runner](https://github.com/qir-alliance/qir-runner/blob/main/sparsesim/src/lib.rs#L35), which does not support SX natively. It could be extended to do so, but that would require changing how gate caching works because of the way SX and H would interact. That's why in https://github.com/qir-alliance/qir-runner/pull/221 the same technique was employed: expose the intrinsic but emulate it via H,S,H on the simulator.

library/std/src/Std/Intrinsic.qs
```diff
@@ -827,6 +827,93 @@ operation S(qubit : Qubit) : Unit is Adj + Ctl {
...
+///     \end{bmatrix}.
+/// \end{align}
+/// $$
+operation SX(qubit : Qubit) : Unit is Adj + Ctl {
+    body ... {
+        __quantum__qis__sx__body(qubit);
+    }
+    adjoint ... {
+        __quantum__qis__x__body(qubit);
+        __quantum__qis__sx__body(qubit);
```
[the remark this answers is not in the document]
swernli: Yes, in this case because I'm using the regular gates rather than `Rx` means it will have all the right phase. You can see this with `DumpOperation` and the matrices it produces:
image

ScottCarda-MS (approved):

swernli:
> Looks good, a few minor comments!

somehow this comment for another PR ended up here... Meant for sign off elsewhere!

idavis (approved):

Issue #2276, opened 2025-04-04
`qsc_formatter` missing lint section in Cargo.toml


swernli:
The Cargo.toml for `qsc_formatter` is missing this section:
```
[lints]
workspace = true
```
When that is added, there are bunch of warnings that need to be fixed in the crate.

PR #2339, opened 2025-04-27
Adding linting to qsc_formatter and fixing the warnings

Morcifer:
This PR resolves issue #2276.

There were several straightforward lints that made perfect sense to fix (e.g. [adding the must_use](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate) and the [backticks](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown)). 

But there were a few that I wasn't sure about:

* too_many_lines: For these, I preferred to allow it rather than fix it. For the tests, I think that's the right approach. For the `apply_rules`, it's because I think the method is more readable as a very long function, especially given that it's the match statement that's taking up most of the lines.
* match_same_arms: This one also relates to `apply_rules`. My opinion is that completely fixing the branches ends up going from many *straightforward* conditions to less (but still many) *complicated* conditions. So I fixed some of the arms but then disabled the lint for the rest. If you want, I can fix the rest of the arms instead.
* trivially_copy_pass_by_ref: I'm not 100% sure I agree with removing the reference and using a copy instead, but clippy probably knows much better than me, so I listened to it and changed the function parameters from `&TokenKind` and `&Keyword` to `TokenKind` or `Keyword`.


compiler/qsc_parse/src/lex/concrete.rs
```diff
@@ -11,13 +11,14 @@ use super::{cooked, raw};
...
 /// here because they are needed for formatting.
+#[derive(Clone, Copy)]
 pub struct ConcreteToken {
     pub kind: ConcreteTokenKind,
     pub span: Span,
 }
 
 /// This enum extends the cooked token kind to include whitespace and comment token kinds.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq)]
```
Morcifer, line 21: This was needed for the trivially_copy_pass_by_ref fix.

compiler/qsc_parse/src/lex/concrete.rs
```diff
@@ -11,13 +11,14 @@ use super::{cooked, raw};
 /// Whitespace and comment tokens were removed during the creation of cooked tokens
 /// because they are generally not useful for compilation, but they are reintroduced
 /// here because they are needed for formatting.
+#[derive(Clone, Copy)]
```
Morcifer, line 14: This was needed for the trivially_copy_pass_by_ref fix.

compiler/qsc_formatter/src/formatter/tests.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(clippy::unicode_not_nfc)]
```
Morcifer, line 4: This is also what's done in compiler\qsc\src\interpret\circuit_tests.rs, to allow the `〉` in `|00〉`.

swernli (approved):
Thanks for the contribution!

ScottCarda-MS (approved):
Thanks for fixing this!

PR #2343, opened 2025-04-28
Add "Update Copilot instructions" command

minestarks:
This adds a new command "Q#: Update copilot instructions" to update the `.github/copilot-instructions.md` file to help make Copilot work better with Q#.

This flow also gets invoked whenever the user creates a Q# project.

image


vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as vscode from "vscode";
+import { log } from "../../npm/qsharp/dist/log";
```
billti, line 5: This should just be from the `qsharp` package.

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
...
+
+const codingInstructionsTitle = "# Q# coding instructions (updated April 2025)";
+
+const codingInstructionsContent = `${codingInstructionsTitle}
+
+  Follow these instructions when generating Q# code in .qs files,
+  and Q# project folders tha include a \`qsharp.json\` file.
```
billti, line 12: Typo on `that`

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  Modern Q# does not use \`namespace\` blocks to enclose code.
+  When Q# code is in different files in a project, each function or operation is in a namespace
+  which is the name of the containing file. For example, if \`Main.qs\` has an operation \`Foo\`,
+  then \`Tests.qs\` could reference the operation as \`Main.Foo\`, or bring \`Foo\` into scope by
+  adding \`import Main.Foo;\` in the file.
+
+  ## Libraries
+
+  A Q# project can reference a library from GitHub but updating the \`dependencies\` entry of
```
billti, line 48: My typo there, should be `by updating`

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
...
+): Promise<boolean> {
+  try {
+    const githubDir = vscode.Uri.joinPath(workspaceFolder, ".github");
+    const instructionsFile = vscode.Uri.joinPath(
+      githubDir,
+      "copilot-instructions.md",
+    );
+
+    // Check if file exists
+    await vscode.workspace.fs.stat(instructionsFile);
```
billti, line 115: Should it do something with the result of this? Seems like the below should be wrapped in an `if` block conditional on this check.

billti: (Or does it just throw, hence the catch below?)

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
...
+    const hasInstructions = await hasQSharpCopilotInstructions(folder.uri);
+    if (hasInstructions) {
+      vscode.window.showInformationMessage(
+        "copilot-instructions.md already contains Q# instructions",
+      );
+      return;
+    }
+  }
+
+  const workspaceFolder = workspaceFolders[0].uri;
```
billti, line 151: Add a TODO here to determine which workspace folder to add the instructions to. Just picking the first open folder if they have multiple isn't correct.

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,226 @@
...
+    "We're about to update your `copilot-instructions.md` file.\n\n" +
+      "This file helps GitHub Copilot understand and work better with Q# files and features provided by the Quantum Development Kit extension.\n\n" +
+      "Would you like to proceed with updating `copilot-instructions.md`?",
+    { modal: true },
+    "Yes",
+    "No",
+  );
+
+  if (response !== "Yes") {
+    return; // User canceled or dismissed the dialog
```
billti, line 168: Can be a follow up, but might be an interesting telemetry point to see how many folks bail out here.

vscode/src/ghCopilot.ts
minestarks: Hmm, are you thinking of a definite way we could check, or more like heuristics? We could prefer workspaces that have Q# files (or the one that has the Q# file that's currently open), and fall back to the first one if we can't find a preferred workspace. Something like that? I'll add a todo.

billti: the package.json only makes this command available if the resource is a .qs or qsharp.json file, so I assume the user is focused on file in a project/workspace for this to be available? (Likewise if invoked via creating a new project - that's done in the context of a specific folder/workspace)

billti: The second param for registerCommand is an optional Uri for the resource it was invoked for (see the `async (folderUri: vscode.Uri | undefined) => {`) code where the `create project` command is registered).

vscode/src/ghCopilot.ts
minestarks: I went ahead and added the data point

vscode/test/suites/workspace-commands/index.ts
minestarks, line 1: These got accidentally added - will revert. Tests coming in a different PR

vscode/src/ghCopilot.ts
```diff
@@ -0,0 +1,260 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { log } from "qsharp-lang";
+import * as vscode from "vscode";
+import { EventType, sendTelemetryEvent, UserFlowStatus } from "./telemetry";
+
+const codingInstructionsTitle = "# Q# coding instructions (updated April 2025)";
+
+const codingInstructionsFooter = `---End: Q# coding instructions---\n\n`;
```
minestarks, line 10: Make this a comment block using markdown (HTML) syntax

vscode/src/ghCopilot.ts
minestarks: VS Code doesn't pass this when invoked from the command palette, currently. Investigating. I can just as well use the "get currently active Q# program" function we use everywhere else for our commands (QIR generation etc)

minestarks: done.

orpuente-MS (approved):

billti (approved):

PR #2344, opened 2025-04-29
Avoid changing layout if the panel is already visible

billti:
As it was, it would constantly rearrange the panels when you re-run a histogram etc. even if it was already visible and you'd laid out the panels to your liking.

### DIFF
--- a/vscode/src/webviewPanel.ts
+++ b/vscode/src/webviewPanel.ts
@@ -529,7 +529,13 @@ export class QSharpWebViewPanel {
   }
 
   reveal(column: ViewColumn) {
-    this.panel.reveal(column, true);
+    // If it's already visible, don't do anything else it messes up the existing layout
+    // This isn't perfect, as if the tab is just in the background it will still call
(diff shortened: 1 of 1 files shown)


orpuente-MS (approved):

vscode/src/webviewPanel.ts
```diff
@@ -529,7 +529,13 @@ export class QSharpWebViewPanel {
   }
 
   reveal(column: ViewColumn) {
-    this.panel.reveal(column, true);
+    // If it's already visible, don't do anything else it messes up the existing layout
+    // This isn't perfect, as if the tab is just in the background it will still call
+    // reveal to bring to the foreground and change the layout, but this is the best we
+    // can do with the current API.
```
ScottCarda-MS, line 535: Played around with it a bit, and couldn't get it to mess up the layout. Is there a repro for it still changing the layout in a way that a user wouldn't expect?

ScottCarda-MS (approved):

minestarks (approved):

PR #2345, opened 2025-04-29
Add Python evaluation API

idavis:


pip/tests/test_qasm.py
```diff
@@ -0,0 +1,808 @@
...
+from qsharp import (
+    init,
+    TargetProfile,
+    set_quantum_seed,
+    BitFlipNoise,
+    QSharpError,
+    Result,
+)
+from qsharp.estimator import EstimatorParams, QubitParams, QECScheme, LogicalCounts
+from qsharp.interop.qasm import (
```
billti, line 16: It seems a bit long-winded having `interop` in the module path, and makes QASM sound a bit like an afterthought bolted on with a shim layer. Would just `qsharp.qasm` be feasible? Simpler, shorter, and makes QASM sound more like it has "first class" support in the package.

idavis: @billti Do you want that for Qiskit as well?

billti: Would that be a breaking change? It would look cleaner naming-wise, though Qiskit does still feel like an 'interop' scenario, whereas with the recent work QASM is more directly supported in the package. I could go either way...

idavis: I can use a redirction shim which would report a warning for any user importing it while automaticatically re-exporting everything from the new location. Then later we could delete the redirection shim.

pip/src/interop.rs
```diff
@@ -79,18 +80,46 @@ where
...
+///     QasmError: If there is an error generating, parsing, or analyzing the OpenQASM source.
+///     QSharpError: If there is an error interpreting the input.
 #[pyfunction]
 #[allow(clippy::too_many_arguments)]
 #[pyo3(
-    signature = (source, callback=None, read_file=None, list_directory=None, resolve_path=None, fetch_github=None, **kwargs)
+    signature = (source, callback=None, noise=None, read_file=None, list_directory=None, resolve_path=None, fetch_github=None, **kwargs)
 )]
-pub fn run_qasm(
+pub fn run_qasm_program(
```
swernli, line 118: nitpick: does this need to be `pub`? The others are `pub(crate)` which seems to work fine with pyO3 and how it's used within this crate.

```suggestion
pub(crate) fn run_qasm_program(
```

pip/qsharp/interop/qasm/_run.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+            results[-1]["result"] = run_results
+
+            if on_result:
+                on_result(results[-1])
+            # For every shot after the first, treat the entry expression as None to trigger
+            # a rerun of the last executed expression without paying the cost for any additional
+            # compilation.
+            source = None
+        if not save_events:
```
swernli, line 133: Technically this isn't needed... we only go into this branch if `callable` is defined, and `callable` is only defined in the branch above where `source` is already set to `None`.

```suggestion
        if not save_events:
```

pip/qsharp/interop/qasm/_ipython.py
```diff
@@ -0,0 +1,24 @@
...
+
+def display_or_print(output: Output) -> None:
+    if _in_jupyter:
+        try:
+            display(output)
+            return
+        except:
+            # If IPython is not available, fall back to printing the output
+            pass
+    print(output, flush=True)
```
swernli, line 24: Defining this is one place is nice... I wonder why we didn't do it for the code in pip/qsharp/_qsharp.py? If the helper logic in this file were moved to a more central place then everything could take advantage of it since this same pattern is used in three more places outside of your changes in existing code. Not critical for this change though.

swernli (approved):

billti (approved):

pip/qsharp/interop/qasm/_compile.py
```diff
@@ -0,0 +1,102 @@
...
+    Args:
+        source (str): An OpenQASM program. Alternatively, a callable can be provided,
+            which must be an already imported global callable.
+        *args: The arguments to pass to the callable, if one is provided.
+        **kwargs: Additional keyword arguments to pass to the compilation when source program is provided.
+          - name (str): The name of the circuit. This is used as the entry point for the program.
+          - target_profile (TargetProfile): The target profile to use for code generation.
+          - search_path (Optional[str]): The optional search path for resolving file references.
+          - output_semantics (OutputSemantics, optional): The output semantics for the compilation.
+          - program_ty (ProgramType, optional): The type of program compilation to perform.
```
billti, line 36: Why abbreviate this to `program_ty` as opposed to `program_type`? That's a very Rust-y/compiler developer abbreviation and one I'm not sure our quantum audience would intuit.

PR #2347, opened 2025-04-29
Fix paths to be relative and add single file sample

### LINKED ISSUE #2266: Update Python samples to use project relative to script
In the Python samples that load a Q# project, we often set the project_root as "." However this is relative to the process working directory, which is likely the workspace in VS Code, which means the script fails to run if not at the root of the workspace.

We should probably change the Python samples to use something like,

```python
from pathlib import Path
this_dir = Path(__file__).resolve().parent
qsharp.init(project_root=this_dir)
```

billti:
Fixes #2266 

For the Python files, make loading a project relative to the current script, rather than current dir, else they fail to run when you open the root of the qsharp repo.

Also added a single file 'eval' sample under "samples/singlefile" as we didn't have a basic example of that previously.

VS Code auto-format made some additional changes on save also.


swernli (approved):

samples/singlefile/run.py
minestarks, line 1: I don't understand what this sample is for or who it's meant to benefit.

But if we keep it, can we rename the directory to be `single_file` to match the convention in this directory?

image

billti: Refactored sample structure and naming. See if you like this better.

samples/python_interop/eval_single_file/README.md
```diff
@@ -0,0 +1,7 @@
+# An example of loading and running code from a Q# file
+
+This example shows how the `qsharp` Python package can load and execute a file containing
+Q# code.
+
+Most other samples show using a `qsharp.init` call to load a Q# project, or using Q# cells
+in Jupyter Notebooks. This example simple reads a Q# file from disk and evaluates it.
```
billti, line 7:
```suggestion
in Jupyter Notebooks. This example simply reads a Q# file from disk and evaluates it.
```

minestarks (approved):

PR #2349, opened 2025-04-29
Add GitHub Copilot tools for Azure Quantum

minestarks:
Integrates Azure Quantum into the GitHub Copilot chat panel.

Example:

image


vscode/src/copilot/azqTools.ts
```diff
@@ -411,11 +440,13 @@ async function submitToTarget(
 
   const quantumUris = new QuantumUris(workspace.endpointUri, workspace.id);
 
-  const confirmed = await CopilotView.getConfirmation(
-    `Submit job "${jobName}" to ${target.id} for ${numberOfShots} shots?`,
-  );
-  if (!confirmed) {
-    return { result: "Job submission was cancelled by the user" };
+  if (confirmation) {
```
billti, line 443: Did I merge this PR right? Seems like this confirmation is still set to the old Copilot In Microsoft Quantum chat window. (Unless I'm misreading it)

vscode/src/copilot/azqTools.ts
```diff
@@ -313,6 +325,22 @@ async function downloadJobResults(
   }
 }
 
+/**
+ * This is the same as `downloadJobResults`, except it hides the histogram
+ * values from the tool output, and instead returns a simple user-facing message.
+ * The client UI is expected to be able to render the actual histogram in `widgetData`.
+ */
+async function renderHistogramForJobResults(
```
billti, line 333: Why is this called 'renderHistogram' when it seems from the comment above like it's job is the opposite (i.e. it hides the histogram)?

vscode/src/copilot/azqTools.ts
billti: OK. Got it now. That whole indirection through tool, and invokeTool, etc. made it a little hard to follow at first. Thx.

vscode/src/copilot/azqTools.ts
```diff
@@ -304,7 +316,7 @@ async function downloadJobResults(
         shotCount,
       };
       return {
-        result: "Results were successfully rendered.",
+        result: JSON.stringify({ histogram }),
```
billti, line 319: Does the branch between 306 - 312 work fine if taken in the copilot flow? I'm not sure what the Agent would make of that return message and not being aware the document was opened.

billti (approved):
A couple comments to verify, else looks good. Thanks!

vscode/src/copilot/azqTools.ts
minestarks: My phrasing may have been unclear. It "hides the histogram _values_". The raw data. So it won't spit out `{ [1,0]: 0.5 }` etc in the _tool response_ as this confuses the Azure Quanum LLM. Instead, the tool result, as the LLM sees it, is simply "results were successfully rendered". Separately, we "send" the results directly to the UI via the widget mechanism.*

*I hate this

billti: > we "send" the results directly to the UI via the widget mechanism.*

I've not seen it open the widget in the UI. Have you tested that on this branch lately? I only see it describe the histogram in the chat response. I'll dig into it if that is what is expected

minestarks: This function is called by the old Quantum Copilot, not GH Copilot. As of my last change the CH copilot didn't pop up the Histogram UI (I never intended it to). As of your latest commit it does.

vscode/src/copilot/azqTools.ts
minestarks: This worked well in the GH copilot flow since we did not pop up the Histogram window in my original change - GH copilot will just respond to the user with the histogram results. 

Now that we pop up the window too (with your commit) , I'll double check the user experience. It's probably going to continue responding as before, and not know anything about the UI

vscode/src/gh-copilot/tools.ts
```diff
@@ -0,0 +1,129 @@
...
+
+async function submitToTarget(input: {
+  job_name: string;
+  target_id: string;
+  number_of_shots: number;
+}): Promise<any> {
+  return (await azqTools.submitToTarget(workspaceState, input, false)).result;
+}
+
+function submitToTargetConfirm(input: {
```
ScottCarda-MS, line 61: Is this actually one of the tools? It seems different from the others in that it never calls azqTools and isn't async. If it isn't one of the tools, could we put it at the bottom of the file with the other Misc. functions?

vscode/src/gh-copilot/tools.ts
```diff
@@ -0,0 +1,129 @@
...
+
+async function getProviders(): Promise<any> {
+  return (await azqTools.getProviders(workspaceState)).result;
+}
+
+async function getTarget(input: { target_id: string }): Promise<any> {
+  return (await azqTools.getTarget(workspaceState, input)).result;
+}
+
+export function registerLanguageModelTools(context: vscode.ExtensionContext) {
```
ScottCarda-MS, line 92: Same question as above. Is this really a tool?

vscode/src/gh-copilot/tools.ts
minestarks: No, this is not a tool, but rather a function that generates the confirmation prompt for the SubmitToTarget tool.  It's a UI feature that the GitHub Copilot chat window provides

image

vscode/src/gh-copilot/instructions.ts
```diff
@@ -15,6 +15,48 @@ const codingInstructionsContent = `${codingInstructionsTitle}
...
+
+\`\`\`
+project_root
+|--qsharp.json
+|--src
+|--|--Main.qs
+|--|--Tests.qs
+\`\`\`
+
+A typical \`qsharp.json\` will be an empty JSON file.
```
ScottCarda-MS, line 49:
```suggestion
A typical \`qsharp.json\` will be a JSON file with an empty JSON object in it.
```
I worry that it will get confused and think to make a file that is actually empty, rather than a file that contains `{}`. I see the example below clarifies this, but it'd be good to be as consistent and clear as can be.

vscode/src/gh-copilot/tools.ts
minestarks: Yeah, no it's not a tool either... I'll remove that /// MISC comment since i think it just ended up in an awkward spot after merged/moved things around in this file. Otherwise I think I'm happy with the code organization

vscode/src/gh-copilot/instructions.ts
minestarks: I like that, thanks.

ScottCarda-MS (approved):

vscode/src/copilot/azqTools.ts
minestarks: I played with it, I think it's fine. _Maybe_ we could have the tool "tell" Copilot about what just happened in the UI so it sounds more "aware", but I don't think it's worth the effort at this moment.

image

PR #2351, opened 2025-04-30
Improved Drag and Drop

ScottCarda-MS:
Allows gates to be dragged to spots that would otherwise be invalidated by their own control lines.

Personal Dropzones

Allows qubit lines to be dragged and dropped like gates.

Move Qubit Lines

Qubits can be deleted by dragging them off the circuit, similar to gates.


minestarks (approved):
Thanks for walking me through the architecture - feature looks great. Still thinking about how we may want to componentize and test these later on

orpuente-MS (approved):

PR #2355, opened 2025-05-01
Setting up initial language service configuration.

idavis:


vscode/src/common.ts
```diff
@@ -51,6 +52,18 @@ export function isQsharpNotebookCell(document: TextDocument): boolean {
...
 
+// Returns true for all OpenQASM documents, including unsaved files, notebook cells, etc.
+export function isOpenQasmDocument(document: TextDocument): boolean {
+  return (
+    document.languageId === openqasmLanguageId &&
+    (Utils.extname(document.uri) === ".qasm" ||
+      Utils.extname(document.uri) === ".inc" ||
+      document.isUntitled) &&
+    document.uri.scheme !== "git" &&
+    document.uri.scheme !== "pr"
```
billti, line 63: You might want to copy the 'chat' exclusion from line 43 above also, else any QASM in a Copilot chat can confuse things.

vscode/src/debugger/activate.ts
```diff
@@ -88,14 +104,15 @@ function registerCommands(context: vscode.ExtensionContext) {
...
+  function startQSharpDebugging(
     resource: vscode.Uri | undefined,
     config: { name: string; [key: string]: any },
     options?: vscode.DebugSessionOptions,
   ) {
     clearCommandDiagnostics();
 
-    if (vscode.debug.activeDebugSession?.type === "qsharp") {
+    const ty = vscode.debug.activeDebugSession?.type;
+    if (ty === "qsharp") {
```
billti, line 115: Why split this out into multiple statements?

idavis: leftover from changes that are no longer in the code

compiler/qsc_project/src/project.rs
```diff
@@ -36,6 +43,7 @@ pub struct Project {
     pub lints: Vec<LintOrGroupConfig>,
     /// Any errors encountered while loading the project.
     pub errors: Vec<Error>,
+    pub project_type: ProjectType,
```
minestarks, line 46: Can you add a doc comment? And maybe move this to under `path` to make the order more logical

compiler/qsc_qasm/src/io.rs
minestarks, line 1: Should I be reviewing this? I think you're still working on includes, so this is temporary code?

minestarks (commented):
Random set of comments for what I've looked at so far

playground/src/editor.tsx
```diff
@@ -206,6 +208,7 @@ export function Editor(props: {
       sources: [["code", code]],
       languageFeatures: [],
       profile: profile,
+      projectType: "qsharp",
```
billti, line 211: Is this a breaking change for QCOM? (i.e. will they need to pass this extra param now also). If not (i.e it's optional/unnecessary), why not just leave this file untouched?

language_service/src/completion.rs
```diff
@@ -432,3 +459,95 @@ impl Completion {
         }
     }
 }
+
+fn get_qasm_completions(
+    compilation: &Compilation,
+    _source_name: &str,
+    _position: Position,
+    _source_name_relative: &str,
```
minestarks, line 467: Can we just remove the unused args?

language_service/src/completion.rs
minestarks, line 1: Please add completion unit test cases . I might start with one for a keyword, and then one for an identifer.

language_service/src/completion.rs
```diff
@@ -129,6 +155,7 @@ fn expected_word_kinds(
             }),
             cursor_offset,
         ),
+        CompilationKind::OpenQASM { .. } => unreachable!(""),
```
minestarks, line 158: please add a string

language_service/src/completion.rs
```diff
@@ -432,3 +459,95 @@ impl Completion {
         }
     }
 }
+
+fn get_qasm_completions(
```
minestarks, line 463: Would be nice to pull the qasm-related stuff into a separate module like `qasm_completion.rs` - this is getting chunky

language_service/src/completion.rs
```diff
@@ -432,3 +459,95 @@ impl Completion {
...
+
+    // The tricky bit: locals, names we need to gather from the compilation.
+    let name_completions = collect_names_qasm(expected_words_at_cursor, cursor_offset, compilation);
+
+    // We have all the data, put everything into a completion list.
+    into_completion_list(once(hardcoded_completions).chain(name_completions))
+}
+
+#[allow(clippy::items_after_statements)]
+fn collect_hardcoded_words_qasm(
```
minestarks, line 488: Has this been updated for QASM? still seems very Q# specific.

language_service/src/qasm.rs
```diff
@@ -0,0 +1,101 @@
...
+    Project {
+        package_graph_sources: PackageGraphSources {
+            root: PackageInfo {
+                sources,
+                language_features: LanguageFeatures::default(),
+                dependencies: FxHashMap::default(),
+                package_type: None,
+            },
+            packages: FxHashMap::default(),
+        },
```
minestarks, line 84: I don't think reusing the `PackageGraphSources` type is adding much value to QASM "projects" . Almost all of the fields here are Q#-specific.

I suggest making `package_graph_sources` a Q#-specific field (stick it under the `ProjectType::QSharp` enum) . It's QASM equivalent would be just `sources` (which would be a field of `ProjectType::OpenQASM`)

compiler/qsc_project/src/project.rs
```diff
@@ -20,6 +20,15 @@ use thiserror::Error;
 
 pub const GITHUB_SCHEME: &str = "qsharp-github-source";
 
+#[derive(Debug, Clone, Default)]
+pub enum ProjectType {
+    #[default]
+    /// A Q# project. Described by a `qsharp.json` manifest or a single Q# file
+    QSharp,
+    /// A QASM project. Described by an `OpenQASM` source file and its includes
+    OpenQASM,
```
minestarks, line 29: I already commented about this at `load_openqasm_project`: I think you can use this enum to separate fields that are Q#-specific from QASM. Specifically , `package_graph_sources`. I don't think that type is adding much value for QASM and it looks like you're having to fight it .

`lints` is also a good candidate to make Q#-specific.

language_service/src/completion/text_edits.rs
```diff
@@ -27,8 +27,10 @@ impl TextEditRange {
         };
         finder.visit_package(&compilation.user_unit().ast.package);
 
-        let insert_open_at = match compilation.kind {
-            CompilationKind::OpenProject { .. } => finder.start_of_namespace,
+        let insert_open_at = match &compilation.kind {
+            CompilationKind::OpenProject { .. } | CompilationKind::OpenQASM { .. } => {
```
minestarks, line 31: What happens if we make the `OpenQASM` arm unreachable? This whole file is _so_ inapplicable to QASM, I'll be sad if we're calling this somehow in QASM cases.

language_service/src/compilation.rs
```diff
@@ -239,11 +246,61 @@ impl Compilation {
         }
     }
 
+    pub(crate) fn new_qasm(
+        package_type: PackageType,
+        target_profile: Profile,
+        package_graph_sources: PackageGraphSources,
```
minestarks, line 252: I already commented about this in a couple places now - maybe let's just drop the use of `PackageGraphSources` for QASM, and use just `sources` (i.e. the `(str,str)[]` one) instead. The "package graph" is a very Q# specific concept and you're dragging the type around only to have to fight it in QASM-land :)

minestarks (commented):
Finished reviewing the `language_service` rust crate - dropping this batch of comments before moving on

language_service/src/state.rs
```diff
@@ -582,6 +600,10 @@ impl<'a> CompilationStateUpdater<'a> {
     }
 }
 
+fn is_openqasm_file(doc_uri: &Arc<str>) -> bool {
+    doc_uri.ends_with("qasm") || doc_uri.ends_with("inc")
```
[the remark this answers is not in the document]
idavis: Makes sense, this will be...intrusive...

language_service/src/qasm.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::{path::Path, sync::Arc};
+
+use qsc::{qasm::parser::ast::StmtKind, LanguageFeatures};
+use qsc_project::{FileSystemAsync, PackageGraphSources, PackageInfo, Project};
+use rustc_hash::FxHashMap;
+
+pub async fn load_openqasm_project<T>(project_host: &T, doc_uri: &Arc<str>) -> Project
```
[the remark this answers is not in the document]
idavis: Putting it in qsc_project will require a ref to qsc_qasm in its Cargo.toml. From Python we have a loader that already knows how to resolve files given a search path and the Python FS, but we don't need projects in Python.

language_service/src/completion.rs
idavis: What part seems Q# specific in hardcoded words?

language_service/src/qasm.rs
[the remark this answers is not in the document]
minestarks: Unifying project load between Language service & Python would be good. Python still does "projects" in that it will resolve includes and construct a program based on multiple files, right? 

> Putting it in qsc_project will require a ref to qsc_qasm in its Cargo.toml

This doesn't seem like a bad thing to me, as `qsc_project` is already aware of QASM projects.  And it's an improvement over having the project system stack (in WASM & JS) take a dependency on the language service

minestarks (commented):
finished language_service crate, wasm and npm layers - big thing for me to look at next is vs code

npm/qsharp/src/compiler/compiler.ts
```diff
@@ -106,6 +107,8 @@ export type ProgramConfig = (
 ) & {
   /** Target compilation profile. */
   profile?: TargetProfile;
+  /** The type of project. This is used to determine how to load the project. */
+  projectType?: ProjectType;
```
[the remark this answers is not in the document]
minestarks: On the other hand, a Q# project can now contain .qsc files, which are a different language, but still within the same project type... so.... 🤯  maybe this is ok.

language_service/src/compilation.rs
```diff
@@ -239,11 +249,55 @@ impl Compilation {
...
+                source.clone(),
+                path.clone(),
+                Some(&mut resolver),
+                package_type,
+                capabilities,
+            );
+
+        let dependencies = dependencies
+            .into_iter()
+            .chain(once((source_package_id, None)))
```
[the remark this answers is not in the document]
idavis: I think I must have taken it from the notebook compilation above.

vscode/package.json
```diff
@@ -511,17 +538,18 @@
     ],
     "debuggers": [
       {
-        "type": "qsharp",
+        "type": "qdk",
```
minestarks, line 541: Since the debug type may have been persisted to a launch.json, I'm not sure if it's a good idea to change its name now.

vscode/src/config.ts
```diff
@@ -32,16 +32,16 @@ export async function setTarget(target: TargetProfile) {
...
     case "adaptive_rif":
-      return "Q#: QIR Adaptive RIF";
+      return "QIR: Adaptive RIF";
     case "unrestricted":
-      return "Q#: unrestricted";
+      return "QIR: unrestricted";
     default:
       log.error("invalid target profile found");
-      return "Q#: invalid";
+      return "QIR: invalid";
```
minestarks, line 44: Wouldn't it make more sense to just remove the Q#: prefix and leave it as "QIR base" etc, instead of putting a colon after QIR?

vscode/src/circuit.ts
```diff
@@ -337,8 +337,8 @@ export function updateCircuitPanel(
     ? `${params.operation.operation} with ${params.operation.totalNumQubits} input qubits`
     : projectName;
 
-  // Trim the Q#: prefix from the target profile name - that's meant for the ui text in the status bar
-  const target = `Target profile: ${getTargetFriendlyName(targetProfile).replace("Q#: ", "")} `;
+  // Trim the QIR: prefix from the target profile name - that's meant for the ui text in the status bar
+  const target = `Target profile: ${getTargetFriendlyName(targetProfile).replace("QIR: ", "")} `;
```
minestarks, line 341: Since you're essentially removing the extra prefix, I think we can just remove the `.replace` call completely

vscode/src/statusbar.ts
```diff
@@ -73,40 +73,42 @@ export function activateTargetProfileStatusBarItem(): vscode.Disposable[] {
...
 
 const targetProfiles = [
-  { configName: "base", uiText: "Q#: QIR base" },
-  { configName: "adaptive_ri", uiText: "Q#: QIR Adaptive RI" },
-  { configName: "adaptive_rif", uiText: "Q#: QIR Adaptive RIF" },
-  { configName: "unrestricted", uiText: "Q#: unrestricted" },
+  { configName: "base", uiText: "QIR: base" },
+  { configName: "adaptive_ri", uiText: "QIR: Adaptive RI" },
+  { configName: "adaptive_rif", uiText: "QIR: Adaptive RIF" },
+  { configName: "unrestricted", uiText: "QIR: unrestricted" },
```
minestarks, line 100: Same comment as in other places - I thought the names were "QIR base", "QIR Adaptive RI/RIF", and "unrestricted" - the "Q#:" part was just a prefix (which I'm fine with removing)

vscode/src/utils.ts
```diff
@@ -29,3 +32,29 @@ export function getRandomGuid(): string {
     hex.substring(20, 32)
   );
 }
+
+/**
+ * This is temporary until we're able to report proper stdlib and project URIs from
+ * the wasm layer. See https://github.com/microsoft/qsharp/blob/f8d344b32a1f1f918f3c91edf58c975db10f4370/wasm/src/diagnostic.rs
```
minestarks, line 38: I don't see this commit. I don't really understand what's going on here.

minestarks: Oh, this is just moved from elsewhere? I wrote this?! It's probably fine then.

minestarks (approved):

vscode/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "qsharp-lang-vscode-dev",
   "displayName": "[DEV BUILD] Azure Quantum Development Kit (QDK)",
-  "description": "Q# Language Support",
+  "description": "Quantum Language Support",
```
billti, line 4: I'd lean towards _adding_ OpenQASM, not removing Q#. When folks go to the marketplace and search on either, we want our extension to show up.

PR #2361, opened 2025-05-02
show copilot instructions prompt at startup

minestarks:
Directly invoked by the user via "create Q# project" or "update Copilot instructions for Q#":

image

One-time prompt at startup, or indirectly invoked whenever our copilot tools get called:

image


vscode/src/gh-copilot/instructions.ts
```diff
@@ -206,24 +209,40 @@ export async function updateCopilotInstructions(resource: vscode.Uri) {
...
+  sendTelemetryEvent(
+    EventType.UpdateCopilotInstructionsStart,
+    {
+      trigger: userInvoked ? "user" : "startup",
+    },
+    {},
+  );
+
+  const buttons = [{ title: "Yes" }, { title: "No", isCloseAffordance: true }];
+  if (!userInvoked) {
```
billti, line 221: If it's user invoked, should we show a confirmation at all, or just update it? It's not like we're doing anything destructive, we only add content. If they explicitly invoked the command, prompting again might be excessive.

vscode/src/gh-copilot/instructions.ts
```diff
@@ -206,24 +209,40 @@ export async function updateCopilotInstructions(resource: vscode.Uri) {
...
-  // Show a yes/no prompt to the user
   const response = await vscode.window.showInformationMessage(
-    "We're about to update your `copilot-instructions.md` file.\n\n" +
-      "This file helps GitHub Copilot understand and work better with Q# files and features provided by the Quantum Development Kit extension.\n\n" +
-      "Would you like to proceed with updating `copilot-instructions.md`?",
-    { modal: true },
-    { title: "Yes" },
-    { title: "No", isCloseAffordance: true },
+    "Add Q# guidance to copilot-instructions.md?\n\n" +
+      "Updating this file will help GitHub Copilot understand and work better with Q# files and other Quantum Development Kit features.",
```
billti, line 227: I wonder if we should add a short-link here for more info. (e.g. `aka.ms/qdk.copilot`). We can create that to point to the copilot wiki for now, and the official docs later.

billti (dismissed):

vscode/src/gh-copilot/instructions.ts
minestarks: Maybe. My goal in showing the prompt was to help the user understand what they just clicked on. They may have been just poking around in the command palette - without this message, there's nothing telling them what actually happened, and I doubt most users would just know about `copilot-instructions.md`.

idavis (dismissed):

minestarks:
I'm not merging this yet because I want to refine this a little to work better with VS Code Next's support for instructions file. For one, we should be able to drop our own `.github/instructions/qsharp.instructions.md` in the workspace and avoid having to append to the global `copilot-instructions.md` . There are other things you can do with settings/user profiles but I can't make up my mind on where exactly we can hook in.

https://github.com/microsoft/vscode-docs/blob/vnext/docs/copilot/copilot-customization.md#specify-custom-instructions-in-settings

vscode/src/gh-copilot/instructions.ts
```diff
@@ -292,7 +318,24 @@ export function registerGhCopilotInstructionsCommand(
...
     vscode.commands.registerCommand(
       "qsharp-vscode.updateCopilotInstructions",
-      updateGhCopilotInstructionsCommand,
+      () => updateGhCopilotInstructionsCommand(true),
     ),
   );
+
+  // Also do a one-time prompt at startup
+  if (
+    context.globalState.get<boolean>(
```
minestarks, line 327: might want to make this `workspaceState` since the instructions files are workspace specific

vscode/src/gh-copilot/instructions.ts
```diff
@@ -4,295 +4,361 @@
...
+/**
+ * Command to update or create the Copilot instructions file for Q#.
+ * Shows a prompt to the user and updates the file if confirmed.
+ */
+export async function updateCopilotInstructions(
+  trigger: "Command" | "Project" | "Activation" | "ChatToolCall",
+  context: vscode.ExtensionContext,
+): Promise<string | undefined> {
+  const globalStateUri = context.globalStorageUri;
+  const userInvoked = trigger === "Command";
```
minestarks, line 17:
```suggestion
  const userInvoked = trigger === "Command" || trigger === "Project";
```

vscode/src/gh-copilot/instructions.ts
minestarks: N/A since my last change

vscode/src/gh-copilot/instructions.ts
minestarks: I decided to leave the prompt in , after using it a bit, this feels right.

vscode/src/gh-copilot/instructions.ts
minestarks: Added the learn more button/link

minestarks:
This PR is ready with a new implementation, please review

vscode/src/gh-copilot/instructions.ts
```diff
@@ -4,295 +4,361 @@
...
-    );
-
-    // Create .github directory if doesn't exist
-    await vscode.workspace.fs.createDirectory(githubDir);
+async function copyInstructionsFileToGlobalStorage(
+  context: vscode.ExtensionContext,
+) {
+  const source = vscode.Uri.joinPath(
+    context.extensionUri,
+    "chat-instructions",
```
billti, line 307: Should the `chat-instructions` dir go inside the `resources` dir to keep the top-level `vscode` dir a little cleaner, and consolidate where the extension resources go?

vscode/src/gh-copilot/tools.ts
```diff
@@ -116,15 +117,14 @@ function tool(
 }
 
 async function invokeTool(
+  context: vscode.ExtensionContext,
   input: any,
   toolFn: (input: any) => Promise<any>,
 ): Promise<vscode.LanguageModelToolResult> {
-  log.info("Invoking tool");
+  updateCopilotInstructions("ChatToolCall", context);
```
billti, line 124: Am I reading this right that was call this on every tool invocation now, and if not configured already we always prompt the user?

vscode/src/gh-copilot/instructions.ts
```diff
@@ -4,295 +4,361 @@
...
+  const response = await showConfirmationPrompt(userInvoked);
 
-\`\`\`
-project_root
-|--qsharp.json
-|--src
-|--|--Main.qs
-|--|--Tests.qs
-\`\`\`
+  if (response !== "Yes") {
```
billti, line 40: Where do you handle the "Don't show again" response if invoked from a tool call flow?

vscode/src/gh-copilot/instructions.ts
```diff
@@ -4,295 +4,361 @@
...
+ * Shows a prompt to the user and updates the file if confirmed.
+ */
+export async function updateCopilotInstructions(
+  trigger: "Command" | "Project" | "Activation" | "ChatToolCall",
+  context: vscode.ExtensionContext,
+): Promise<string | undefined> {
+  const globalStateUri = context.globalStorageUri;
+  const userInvoked = trigger === "Command" || trigger === "Project";
+
+  if (isExtensionInstructionsConfigured(globalStateUri)) {
```
billti, line 19: If I'm reading correctly (and some of this code is a little hard to follow), then this is just checking that the instructions directory is listed in the user settings, and it doesn't check that the instructions are actually in that directory - is that correct? Is the assumption that if the directory is configured, then the instructions are always going to already be in there? Is that a safe assumption?

vscode/src/gh-copilot/tools.ts
minestarks: Correct

vscode/src/gh-copilot/instructions.ts
minestarks: Oof, I missed that - thank you!

vscode/src/gh-copilot/instructions.ts
```diff
@@ -4,295 +4,361 @@
...
       "qsharp-vscode.updateCopilotInstructions",
-      updateGhCopilotInstructionsCommand,
+      () => updateCopilotInstructions("Command", context),
     ),
   );
+
+  // Copy the instructions file to the global storage location
+  // The global storage location is stable across versions,
+  // but our instructions content may change from version to version.
+  await copyInstructionsFileToGlobalStorage(context);
```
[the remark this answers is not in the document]
minestarks: That's right, we copy every single time to keep it up to date. I briefly thought about "optimizing" this but it's a miniscule amount of work compared to whatever else goes on at extension load. 

(The code is so hard to follow for me too, even after many refactors. I'm not sure why I had so much trouble with it.)

vscode/src/gh-copilot/tools.ts
minestarks: Fixed the "don't show again" handling. 

Will try working this _into_ the chat now as we discussed, but separately from this PR

vscode/src/gh-copilot/instructions.ts
minestarks: Done

billti (approved):

PR #2372, opened 2025-05-07
Add/Remove Qubit Lines through Drag-and-Drop

ScottCarda-MS:
This removes the Add/Remove Qubit button controls at the top of the circuit editor in favor of more intuitive drag-n-drop adding and removal of qubits. When dragging a gate, a "ghost" qubit line will appear at the bottom of the circuit where users can drop their gates to create a new qubit line for the gate.

<insert gif>

Qubit lines can be removed by dragging the qubit line off the circuit editor, as before, but now they are also removed when there are trailing unused qubit lines in the circuit.

<insert gif>

Finally, empty qubits will have the "ghost" qubit line visible by default as well as a small instruction message to get users started.

image


compiler/qsc_circuit/src/circuit_to_qsharp/tests.rs
minestarks, line 1: I can't tell why we made the length more permissive. I would have preferred the strict length check. Was it necessary for the UI change, or is it an unrelated change?

ScottCarda-MS: It is because you now can't have trailing unused qubits. This necessarily limits what circuits users can write. Even though unused qubit lines are a bit silly, they might be necessary for a given architecture. Having the more permissive length check means that those unreachable circuit states map into the reachable circuit states.

orpuente-MS (approved):

Issue #2144, opened 2025-01-31
API Documentation doesn't show up if there's an error in a dependency


minestarks:
Today, the "Show API documentation" works even when there are errors in the current project. This is good.

However, if there are errors in a _dependency_, the "Show API documentation" shows a blank page.

**REPRO**

**`MyProj/src/Main.qs`**

```qsharp
operation Main() : Unit {
    WillCauseCompilerError; // This error doesn't block documentation generation ✅
}
```

**`MyProj/src/qsharp.json`**

```json
{
    "dependencies": {
        "MyDep": {
            "path": "../MyDep"
        }
    }
}
```

**`MyDep/src/Main.qs`**

```qsharp
operation DependencyFunction() : Unit {
    WillCauseCompilerError; // This blocks documentation generation! ❌
}

export DependencyFunction;
```

**`MyDep/src/qsharp.json`**

```json
{}
```

Invoke "Show API Documentation" command in VS Code.

**Expected**

The documentation should still be generated for whatever did compile.

Nice to have: To avoid confusion, the documentation panel should also show any compiler errors that were in the project.

**Actual**

Blank documentation page.

Seen in 3b32013a6e1ff26f5d12c12116b1946eeda4bb0d

ScottCarda-MS:
Fixed with #2374

PR #2374, opened 2025-05-07
Ignore Dependency Errors when Getting Documentation

ScottCarda-MS:
Addresses #2144 by ignoring dependency errors when getting documentation.


minestarks (approved):

wasm/src/project_system.rs
```diff
@@ -399,6 +399,7 @@ impl From<PackageInfo> for qsc_project::PackageInfo {
 pub(crate) fn into_qsc_args(
     program: ProgramConfig,
     entry: Option<String>,
+    ignore_dependency_errors: bool,
```
DmitryVasilevsky, line 402: ignore_dependency_errors is kind of a negation built into variable name. Consider fail_on_dependency_errors or something similar instead.

DmitryVasilevsky (approved):

wasm/src/project_system.rs
ScottCarda-MS: I was thinking about this in terms of the "default" value being `false`, even though Rust doesn't actually have default values, so that you'd get the usual behavior passing `false` in, and would get the updated behavior by passing `true`. In this case the updated behavior is a suppression of errors. I can see what you mean by it being a bit of a negation in that you pass in `true` to make it do less.

PR #2377, opened 2025-05-07
Add explicit cast support

orpuente-MS:
Add explicit cast support and fixes a few issues with implicit casting as a side effect.

### DIFF
--- a/compiler/qsc_qasm/src/semantic/lowerer.rs
+++ b/compiler/qsc_qasm/src/semantic/lowerer.rs
@@ -668,14 +668,16 @@ impl Lowerer {
     }
 
     fn lower_cast_expr(&mut self, cast: &syntax::Cast) -> semantic::Expr {
+        let cast_span = cast.span;
         let expr = self.lower_expr(&cast.arg);
         let ty = self.get_semantic_type_from_tydef(&cast.ty, expr.ty.is_const());
-        let mut cast = self.cast_expr_to_type(&ty, &expr);
+        let mut cast = self.cast_expr_to_type_with_span(&ty, &expr, cast_span);
 
         // This is an explicit cast, so we know its span. If casting
         // succeded, override the default span with the cast span.
+        cast.span = cast_span;
         if let semantic::ExprKind::Cast(cast_ref) = cast.kind.as_mut() {
-            cast_ref.span = cast.span;
+            cast_ref.span = cast_span;
         }
 
         cast
@@ -3027,8 +3029,17 @@ impl Lowerer {
     }
 
     fn cast_expr_to_type(&mut self, ty: &Type, expr: &semantic::Expr) -> semantic::Expr {
+        self.cast_expr_to_type_with_span(ty, expr, expr.span)
+    }
+
+    fn cast_expr_to_type_with_span(
+        &mut self,
+        ty: &Type,
+        expr: &semantic::Expr,
+        span: Span,
+    ) -> semantic::Expr {
         let Some(cast_expr) = Self::try_cast_expr_to_type(ty, expr) else {
-            self.push_invalid_cast_error(ty, &expr.ty, expr.span);
+            self.push_invalid_cast_error(ty, &expr.ty, span);
             return expr.clone();
         };
         cast_expr
--- a/compiler/qsc_qasm/src/semantic/lowerer.rs
+++ b/compiler/qsc_qasm/src/semantic/lowerer.rs
@@ -635,10 +635,7 @@ impl Lowerer {
                 let rhs = self.lower_expr(&bin_op_expr.rhs);
                 self.lower_binary_op_expr(bin_op_expr.op, lhs, rhs, expr.span)
             }
-            syntax::ExprKind::Cast(_) => {
-                self.push_unimplemented_error_message("cast expr", expr.span);
-                err_expr!(Type::Err, expr.span)
-            }
+            syntax::ExprKind::Cast(expr) => self.lower_cast_expr(expr),
             syntax::ExprKind::Err => err_expr!(Type::Err, expr.span),
             syntax::ExprKind::FunctionCall(expr) => self.lower_function_call_expr(expr),
             syntax::ExprKind::Ident(ident) => self.lower_ident_expr(ident),
@@ -668,6 +665,20 @@ impl Lowerer {
             // We take this branch if we are trying to lower an array but reached a leaf node.
             self.cast_expr_to_type(ty, &expr)
         }
+    }
+
+    fn lower_cast_expr(&mut self, cast: &syntax::Cast) -> semantic::Expr {
+        let expr = self.lower_expr(&cast.arg);
+        let ty = self.get_semantic_type_from_tydef(&cast.ty, expr.ty.is_const());
+        let mut cast = self.cast_expr_to_type(&ty, &expr);
+
+        // This is an explicit cast, so we know its span. If casting
+        // succeded, override the default span with the cast span.
+        if let semantic::ExprKind::Cast(cast_ref) = cast.kind.as_mut() {
+            cast_ref.span = cast.span;
+        }
+
+        cast
     }
 
     fn lower_ident_expr(&mut self, ident: &syntax::Ident) -> semantic::Expr {
@@ -2787,21 +2798,34 @@ impl Lowerer {
         let span = rhs.span;
 
         if matches!(lhs_ty, Type::Bit(..)) {
-            if let semantic::LiteralKind::Int(value) = kind {
-                // can_cast_literal_with_value_knowledge guarantees that value is 0 or 1
-                return Some(semantic::Expr {
-                    span,
-                    kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(
-                        *value != 0,
-                    ))),
-                    ty: lhs_ty.as_const(),
-                });
-            } else if let semantic::LiteralKind::Bool(value) = kind {
-                return Some(semantic::Expr {
-                    span,
-                    kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(*value))),
-                    ty: lhs_ty.as_const(),
-                });
+            match kind {
+                semantic::LiteralKind::Int(value) => {
+                    // can_cast_literal_with_value_knowledge guarantees that value is 0 or 1
+                    return Some(semantic::Expr {
+                        span,
+                        kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(
+                            *value != 0,
+                        ))),
+                        ty: lhs_ty.as_const(),
+                    });
+                }
+                semantic::LiteralKind::Bool(value) => {
+                    return Some(semantic::Expr {
+                        span,
+                        kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(*value))),
+                        ty: lhs_ty.as_const(),
+                    });
+                }
+                &semantic::LiteralKind::Angle(value) => {
+                    return Some(semantic::Expr {
+                        span,
+                        kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(
+                            value.value != 0,
+                        ))),
+                        ty: lhs_ty.as_const(),
+                    });
+                }
+                _ => (),
             }
         }
         // if lhs_ty is 1 dim bitarray and rhs is int/uint, we can cast
@@ -3036,11 +3060,11 @@ impl Lowerer {
             | (Type::Float(w1, _), Type::Float(w2, _))
             | (Type::Complex(w1, _), Type::Complex(w2, _)) => {
                 if w1.is_none() && w2.is_some() {
-                    return Some(wrap_expr_in_implicit_cast_expr(ty.clone(), expr.clone()));
+                    return Some(wrap_expr_in_cast_expr(ty.clone(), expr.clone()));
                 }
 
                 if *w1 >= *w2 {
-                    return Some(wrap_expr_in_implicit_cast_expr(ty.clone(), expr.clone()));
+                    return Some(wrap_expr_in_cast_expr(ty.clone(), expr.clone()));
                 }
             }
             _ => {}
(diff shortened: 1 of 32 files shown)


compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -74,10 +109,38 @@ function IntAsResultArrayBE(number : Int, bits : Int) : Result[] {
...
+
 /// The ``RESULT_ARRAY_AS_INT_BE`` function is used to implement the cast expr in QASM for bit[] to uint.
 /// with big-endian order. This is needed for round-trip conversion for bin ops.
 function ResultArrayAsIntBE(results : Result[]) : Int {
     Std.Convert.ResultArrayAsInt(Std.Arrays.Reversed(results))
 }
 
+/// The ``DOUBLE_AS_BIT`` function is used to implement the cast expr in QASM for float to bit.
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function DoubleAsBit(value : Double) : Result {
```
DmitryVasilevsky, line 138: Other functions use 'Result' in this case. Should it be DoubleAsResult?

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -74,10 +109,38 @@ function IntAsResultArrayBE(number : Int, bits : Int) : Result[] {
...
+}
+
 /// The ``RESULT_ARRAY_AS_INT_BE`` function is used to implement the cast expr in QASM for bit[] to uint.
 /// with big-endian order. This is needed for round-trip conversion for bin ops.
 function ResultArrayAsIntBE(results : Result[]) : Int {
     Std.Convert.ResultArrayAsInt(Std.Arrays.Reversed(results))
 }
 
+/// The ``DOUBLE_AS_BIT`` function is used to implement the cast expr in QASM for float to bit.
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
```
DmitryVasilevsky, line 137: "with big-endian order" seems irrelevant here.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -62,6 +65,38 @@ function ResultAsBigInt(input : Result) : BigInt {
...
+    false
+}
+
+/// The ``RESULT_ARRAY_AS_RESULT_BE`` function is used to implement the cast expr in QASM for bit[n] to bit.
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function ResultArrayAsResultBE(array: Result[]) : Result {
+    BoolAsResult(ResultArrayAsBoolBE(array))
+}
+
+/// The ``RESULT_ARRAY_AS_RESULT_BE`` function is used to implement the cast expr in QASM for bit[n] to angle[n].
```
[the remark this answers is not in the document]
orpuente-MS: That was the intention, `n` should be the same for both of them.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -74,10 +109,38 @@ function IntAsResultArrayBE(number : Int, bits : Int) : Result[] {
...
 /// The ``RESULT_ARRAY_AS_INT_BE`` function is used to implement the cast expr in QASM for bit[] to uint.
 /// with big-endian order. This is needed for round-trip conversion for bin ops.
 function ResultArrayAsIntBE(results : Result[]) : Int {
     Std.Convert.ResultArrayAsInt(Std.Arrays.Reversed(results))
 }
 
+/// The ``DOUBLE_AS_BIT`` function is used to implement the cast expr in QASM for float to bit.
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function DoubleAsBit(value : Double) : Result {
+    if Truncate(value) == 0 {
```
DmitryVasilevsky, line 139: Is this proper semantics? Which numbers you expect to result in Zero and which ones in One?

orpuente-MS: I am not sure. I was expecting to get your feedback on this one. I am just mimicking the logic we already had to cast from `Double` to `Bool`.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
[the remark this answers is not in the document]
DmitryVasilevsky: 'bits' are provided as a parameter to this function. So bits can be different as far as I can tell.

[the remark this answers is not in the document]
orpuente-MS: Yes, that makes sense. I changed it to say `bit[bits] to angle[bits]`.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
DmitryVasilevsky: Do you know where that logic is? Can't seem to find it.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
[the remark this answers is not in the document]
DmitryVasilevsky: I still don't get it. If the number of bits for the angle is always the same as the number of elements in the array, why does this function take a separate parameter 'bits'? It should just use array length for bits.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
DmitryVasilevsky: OK, Casting from float needs to truncate according to this: https://openqasm.com/language/types.html#casting-from-float
But the name should probably be DoubleAsResult.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -74,10 +109,38 @@ function IntAsResultArrayBE(number : Int, bits : Int) : Result[] {
...
+/// The ``BOOL_AS_RESULT_ARRAY_BE`` function is used to implement the cast expr in QASM for bool to bit[bits].
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function BoolAsResultArrayBE(value : Bool, bits: Int) : Result[] {
+    IntAsResultArrayBE(BoolAsInt(value), bits)
+}
+
+/// The ``RESULT_AS_RESULT_ARRAY_BE`` function is used to implement the cast expr in QASM for bit to bit[bits].
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function ResultAsResultArrayBE(value : Result, bits: Int) : Result[] {
+    IntAsResultArrayBE(ResultAsInt(value), bits)
```
DmitryVasilevsky, line 127: I general we want to postpone Result evaluation as long as possible. ResultAsInt will evaluate 'value' here. Technically we don't need it. We just need to stick it as one element of the array without evaluation (first element?). So we can do
something like "return [value] + Repeated(Zero, bits-1)". You need to verify,

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
```diff
@@ -62,6 +65,38 @@ function ResultAsBigInt(input : Result) : BigInt {
...
+    if value == One {
+        1.
+    } else {
+        0.
+    }
+}
+
+/// The ``RESULT_ARRAY_AS_BOOL_BE`` function is used to implement the cast expr in QASM for bit[bits] to bool.
+/// with big-endian order. This is needed for round-trip conversion for bin ops.
+function ResultArrayAsBoolBE(array: Result[]) : Bool {
```
DmitryVasilevsky, line 79: Consider renaming this to just ResultArrayAsBool (without BE). And also, ResultArrayAsResult.
The result of the function doesn't depend on the order of elements. So, the BE suffix is just getting in the way.
There's a subtle difference in the order of evaluation - i.e. some results may be evaluated earlier, and some would not be evaluated at all. But I think this is just and implementation detail. Up to you.

DmitryVasilevsky (approved):
Signing off on Convert.qs

compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_angle.rs
```diff
@@ -0,0 +1,312 @@
...
+            init_expr: Expr [0-0]:
+                ty: Angle(None, true)
+                kind: Lit: Angle(0)
+        ExprStmt [26-35]:
+            expr: Expr [32-33]:
+                ty: Angle(None, false)
+                kind: SymbolId(8)
+    "#]],
+    );
+}
```
idavis, line 177: missing truncating and expanding cast tests

compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_angle.rs
```diff
@@ -0,0 +1,312 @@
...
+                ty: BitArray(8, false)
+                kind: Cast [0-0]:
+                    ty: BitArray(8, false)
+                    expr: Expr [33-34]:
+                        ty: Angle(None, false)
+                        kind: SymbolId(8)
+    "#]],
+    );
+}
+
```
idavis, line 232: missing truncating and expanding cast tests

compiler/qsc_qasm/src/tests/expression/implicit_cast_from_bit.rs
```diff
@@ -143,11 +143,12 @@ fn to_implicit_float_implicitly_fails() {
...
-        .assert_eq(&error[0].to_string());
+    check_qasm_to_qsharp(
+        source,
+        &expect![[r#"
+        import QasmStd.Intrinsic.*;
+        mutable x = One;
+        mutable y = QasmStd.Convert.ResultAsDouble(x);
+    "#]],
+    );
 }
```
idavis, line 154: Test name doesn't match what is tested

compiler/qsc_qasm/src/tests/statement/const_eval.rs
```diff
@@ -1772,29 +1772,18 @@ fn cast_to_float() -> miette::Result<(), Vec<Report>> {
...
-        let a2 = One;
         let b1 = QasmStd.Angle.DoubleAsAngle(a1, 32);
-        let b2 = QasmStd.Angle.ResultAsAngle(a2);
         let s1 = QasmStd.Angle.AngleAsResult(b1);
-        let s2 = QasmStd.Angle.AngleAsResult(b2);
         mutable r1 = [Zero];
-        mutable r2 = [Zero];
     "#]]
     .assert_eq(&qsharp);
     Ok(())
```
idavis, line 1789: should bit arrays be in here?

compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_float.rs
```diff
@@ -0,0 +1,274 @@
...
+    );
+}
+
+/// Even though the spec doesn't say it, we need to allow
+/// casting from float to complex, else this kind of expression
+/// would be invalid: 2.0 + sin(pi) + 1.0i.
+#[test]
+fn float_to_complex() {
+    let input = "
+        float x;
```
idavis, line 252: This comment is duplicated but I don't see this as a test case anywhere.

compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_float.rs
```diff
@@ -0,0 +1,274 @@
...
+        ClassicalDeclarationStmt [9-17]:
+            symbol_id: 8
+            ty_span: [9-14]
+            init_expr: Expr [0-0]:
+                ty: Float(None, true)
+                kind: Lit: Float(0.0)
+        ExprStmt [26-34]:
+            expr: Expr [31-32]:
+                ty: UInt(None, false)
+                kind: Cast [0-0]:
```
idavis, line 79: The spans on all of the explicit casts are missing

compiler/qsc_qasm/src/tests/statement/const_eval.rs
orpuente-MS: Bitarrays are only here to use the other expressions as their size, that forces the lowerer to perform the const evaluation.

compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_float.rs
orpuente-MS: Nice catch. There were some tests testing this as a side effect of what they were actually testing.

I added some explicit tests for this.

compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -2774,21 +2777,34 @@ impl Lowerer {
...
+                }
+                &semantic::LiteralKind::Angle(value) => {
+                    return Some(semantic::Expr {
+                        span,
+                        kind: Box::new(semantic::ExprKind::Lit(semantic::LiteralKind::Bit(
+                            value.value != 0,
+                        ))),
+                        ty: lhs_ty.as_const(),
+                    });
+                }
```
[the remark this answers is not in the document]
orpuente-MS: There are three cases covering this:
 - compiler/qsc_qasm/src/tests/statement/const_eval.rs::cast_to_bit
 - compiler/qsc_qasm/src/tests/expression/explicit_cast_from_angle.rs::angle_to_bit
 - compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_angle.rs::angle_to_bit

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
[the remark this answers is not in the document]
orpuente-MS: Yes, you are right. I modified the functions to remove the `bits` argument.

compiler/qsc_qasm/src/stdlib/QasmStd/src/QasmStd/Convert.qs
orpuente-MS: I think is the other way around. Since we are in big endian notation, the most significant bit is stored first, in other words the least significant bit is at the end.

idavis (approved):

PR #2380, opened 2025-05-08
[WIP] Copilot tools for run, estimate, circuit

minestarks:


vscode/package.json
```diff
@@ -830,6 +830,116 @@
...
+        "displayName": "Run Q# Program",
+        "modelDescription": "Executes a Q# program. The path to a .qs file must be specified in the `filePath` parameter. A quantum simulator will be run locally. Q# does not provide any CLI tools, so call this tool whenever you need to execute Q#, instead of running any CLI commands. The `shots` parameter controls the number of times to repeat the simulation. If the number of shots is greater than 1, a histogram will be generated, and the results will be displayed in a dedicated panel.",
+        "canBeReferencedInPrompt": true,
+        "icon": "./resources/file-icon-light.svg",
+        "inputSchema": {
+          "type": "object",
+          "properties": {
+            "filePath": {
+              "type": "string",
+              "description": "The absolute path to the .qs file. If this file is part of a project, as defined by being in a folder with a qsharp.json file, the whole project will be compiled as the program."
```
ScottCarda-MS, line 851: This isn't *technically* true, but does it matter to the copilot? Files are part of a project if an ancestor in their path contains a qsharp.json file.

minestarks: No you're right, the more specific and technically correct we are the better Copilot deals with it. I'll fix the description!

vscode/package.json
```diff
@@ -830,6 +830,116 @@
...
+      {
+        "name": "qsharp-generate-circuit",
+        "tags": [
+          "azure-quantum",
+          "qsharp",
+          "qdk"
+        ],
+        "toolReferenceName": "qsharpGenerateCircuit",
+        "displayName": "Generate Q# Circuit Diagram",
+        "modelDescription": "Generates a visual circuit diagram for a Q# program. The path to a .qs file must be specified in the `filePath` parameter. This tool will compile the Q# program and generate a quantum circuit diagram that visually represents the quantum operations in the code. The diagram will be displayed in a dedicated circuit panel.",
```
ScottCarda-MS, line 874: Does this need to know the difference between a trace circuit and a non-trace circuit? The way it is written here indicates that this will generate a non-trace circuit, though I think the behavior is to generate a trace circuit that results from running the Q# program.

minestarks: It doesn't _need_ to know that detail to call the tool, but this is a good opportunity to give Copilot context so it knows what's going on in general. I'll add that to the description.

I'm generally going by these best practices, and they seem to imply that this sort of description would be helpful.
https://docs.anthropic.com/en/docs/build-with-claude/tool-use/implement-tool-use#best-practices-for-tool-definitions

vscode/package.json
minestarks:
```suggestion
              "description": "The absolute path to the .qs file. If this file is part of a project, the whole project will be compiled as the program. A .qs file belongs to a project if it resides anywhere under a src/ directory whose parent directory also contains a qsharp.json manifest."
```

vscode/package.json
minestarks:
```suggestion
        "modelDescription": "Generates a visual circuit diagram for a Q# program. The path to a .qs file must be specified in the `filePath` parameter. This tool will compile the Q# program and generate a quantum circuit diagram that visually represents the quantum operations in the code. If the program contains any dynamic behavior such as comparing measurement results, then the program will be executed using a quantum simulator, and the resulting circuit diagram will be a simply a trace of the operations that were actually executed during simulation. The diagram will be displayed in a dedicated circuit panel.",
```

vscode/package.json
```diff
@@ -830,6 +830,116 @@
...
+        "canBeReferencedInPrompt": true,
+        "icon": "./resources/file-icon-light.svg",
+        "inputSchema": {
+          "type": "object",
+          "properties": {
+            "filePath": {
+              "type": "string",
+              "description": "The absolute path to the .qs file. If this file is part of a project, as defined by being in a folder with a qsharp.json file, the whole project will be compiled as the program."
```
minestarks, line 882:
```suggestion
              "description": "The absolute path to the .qs file. If this file is part of a project, the whole project will be compiled as the program. A .qs file belongs to a project if it resides anywhere under a src/ directory whose parent directory also contains a qsharp.json manifest."
```

vscode/package.json
```diff
@@ -830,6 +830,116 @@
...
+        "displayName": "Run Q# Resource Estimator",
+        "modelDescription": "Runs the quantum resource estimator on a Q# program to calculate the required physical resources. The path to a .qs file must be specified in the `filePath` parameter. This tool will analyze the Q# program and generate estimates of the quantum resources required to run the algorithm, such as the number of qubits, T gates, and other quantum operations. Results will be displayed in a dedicated resource estimator panel.",
+        "canBeReferencedInPrompt": true,
+        "icon": "./resources/file-icon-light.svg",
+        "inputSchema": {
+          "type": "object",
+          "properties": {
+            "filePath": {
+              "type": "string",
+              "description": "The absolute path to the .qs file. If this file is part of a project, as defined by being in a folder with a qsharp.json file, the whole project will be compiled as the program."
```
minestarks, line 908:
```suggestion
              "description": "The absolute path to the .qs file. If this file is part of a project, the whole project will be compiled as the program. A .qs file belongs to a project if it resides anywhere under a src/ directory whose parent directory also contains a qsharp.json manifest."
```

vscode/package.json
ScottCarda-MS: typo: "circuit diagram will be a simply a trace of"

vscode/src/gh-copilot/qsharpTools.ts
```diff
@@ -0,0 +1,289 @@
...
+        finalHistogram = histogram;
+        const uniqueFailures = new Set<string>();
+        sampleFailures = [];
+        for (const failure of failures) {
+          const failureKey = `${failure.message}-${failure.range?.start.line}-${failure.range?.start.character}`;
+          if (!uniqueFailures.has(failureKey)) {
+            uniqueFailures.add(failureKey);
+            sampleFailures.push(failure);
+          }
+          if (sampleFailures.length === 3) {
```
billti, line 77: What is this for and why 3? Do we bail in general when running a histogram/simulation? If not, why do it when invoked via Copilot?

billti: Oh I see. This is just to report a maximum of 3 failure?

vscode/src/gh-copilot/qsharpTools.ts
```diff
@@ -0,0 +1,289 @@
...
+          failures.push(toVsCodeDiagnostic(results[i].result as VSDiagnostic));
+        }
+      }
+      histogram = {
+        buckets: Array.from(buckets.entries()),
+        shotCount: resultCount,
+      };
+      resultUpdate(histogram, failures);
+      if (shots === resultCount || failures.length > 0) {
+        // TODO: ugh
```
billti, line 257: ??

ScottCarda-MS (approved):

vscode/src/gh-copilot/qsharpTools.ts
minestarks: Yes, this truncates the failure count. I'll add clarification in my next PR

vscode/src/gh-copilot/qsharpTools.ts
minestarks: Oh no, I forgot about this :(  Behaviorally this should be ok, but it's a hack.

The problem was in the way the `run` function is designed. We can receive this result event *after* the function completes, but we do need to wait for all the results. We essentially never know when we're done. Originally I had this check `shots === resultCount` in here to understand we're done (i.e. we've run all the shots we set out to run), but that check would never be `true` if we had a compile error. So I had to add `failures.length > 0` check to cover the error case.

I need to revisit this.

Issue #2353, opened 2025-04-30
Need better docs/samples for access items in a tuple


billti:
Copilot struggled to generate code that accessed the elements of a tuple. Looking at our docs and samples we have very little covering this (basically one sentance on <https://learn.microsoft.com/en-us/azure/quantum/user-guide/language/statements/variabledeclarationsandreassignments>). We should add more content explicitly covering this.

PR #2381, opened 2025-05-08
Added tuple unpacking samples

DmitryVasilevsky:
We didn't have much sample code unpacking tuples. Added now.
Addresses https://github.com/microsoft/qsharp/issues/2353

### DIFF
--- a/samples/language/Tuple.qs
+++ b/samples/language/Tuple.qs
@@ -14,10 +14,40 @@ function Main() : (Int, String) {
     let myTuple = ("Id", 0, 1.);
     Message($"Tuple: {myTuple}");
 
+    // A tuple may be unpacked by assigning it to a tuple of variables
+    let (name, n, d) = myTuple;
+    Message($"Unpacked: {name}, {n}, {d}");
+
+    // If not all items are needed, a tuple may be unpacked into
+    // a tuple of variables with a wildcard `_` for the unwanted items.
+    let (name, _, _) = myTuple;
+    Message($"Name: {name}");
+
+    // A "tuple" may be created with a single item, but it is treated
+    // as the item itself enclosed in parenthesis, which is called
+    // singleton tuple equivalence.
+    let item = (0);
+    Message($"Item: {item}");
+    // Note that the type of `item` is `Int`, not `(Int)`.
+
+    // A tuple with a single item may be created with a trailing comma.
+    let myTuple = (0,);
+    Message($"myTuple: {myTuple}");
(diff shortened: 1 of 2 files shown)


samples/language/Tuple.qs
```diff
@@ -14,10 +14,40 @@ function Main() : (Int, String) {
...
+    Message($"Unpacked: {name}, {n}, {d}");
+
+    // If not all items are needed, a tuple may be unpacked into
+    // a tuple of variables with a wildcard `_` for the unwanted items.
+    let (name, _, _) = myTuple;
+    Message($"Name: {name}");
+
+    // A "tuple" may be created with a single item, but it is treated
+    // as the item itself enclosed in parenthesis, which is called
+    // singleton tuple equivalence.
```
swernli, line 28: I think we should avoid mentions of singleton tuple equivalence, since we don't technically have that anymore. Classic QDK would treat `3`, `(3)`, and `(3,)` as the same value of type `Int` because it had true singleton tuple equivalence (there was no such thing as a tuple of arity 1 as it always got turned into a singleton). With the modern QDK we actually do support tuples of arity 1 if they explicitly made a tuple with the trailing comma like the example below this shows.

Likewise, the comment at the top of the file that says "Tuples containing a single item are treated as identical to the item itself, both in tpye and value, which is called singleton tuple equivalence" should be removed.

swernli (approved):

samples/language/Tuple.qs
```diff
@@ -5,19 +5,44 @@
...
     // A tuple of type `String`, `Int`, and `Double`
     let myTuple = ("Id", 0, 1.);
     Message($"Tuple: {myTuple}");
 
+    // A tuple may be unpacked by assigning it to a tuple of variables
+    let (name, n, d) = myTuple;
+    Message($"Unpacked: {name}, {n}, {d}");
+
+    // If not all items are needed, a tuple may be unpacked into
+    // a tuple of variables with a wildcard `_` for the unwanted items.
```
minestarks, line 18:
```suggestion
    // a tuple of variables with a discard binding `_` for the unneeded items.
```

samples/language/Tuple.qs
```diff
@@ -5,19 +5,44 @@
...
+    let (name, _, _) = myTuple;
+    Message($"Name: {name}");
+
+    // A single item in a round brackets does not denote a tuple. In this case,
```
minestarks, line 22:
```suggestion
    // A single item in parentheses does not denote a tuple. In this case,
```

minestarks (approved):
Just a couple of suggestions for consistency of terms

Issue #2387, opened 2025-05-08
Support dynamic early return in QIR generation via inserted mutable variable updates


swernli:
This is follow on from #2290. To fully support early returns from dynamic contexts, we need partial evaluation to do additional book keeping on each callable invocation (including the entry point). Specifically, when an explicit return is hit, a store into a special return variable should be inserted into the RIR. Then if there is an implicit return from the callable when any such early return was previously triggered the implicit return should be treated the same as an explicit return by inserting the same kind of store instructions. This special return variable would need to be created and managed for every call scope, with potential for optimizing it out in cases where it isn't used (such as explicitly returning from a `Unit`-typed callable).

PR #2388, opened 2025-05-08
Fix panic in QIR generation when conditional branches use early return

### LINKED ISSUE #2290: QIR Generation panics when conditional branches use early return
The following code causes a panic in QIR generation:

```qsharp
function Branch(value : Bool) : Int {
    if (value) {
        return 3;
    } else {
        return 4;
    }
}
@EntryPoint()
operation Main() : Int {
    use qubits = Qubit[2];
    H(qubits[0]);
    CNOT(qubits[0], qubits[1]);
    let result = Branch(MResetZ(qubits[0]) == Zero);
    result
}
```

However, the same code with implicit returns in the function works as expected:

```qsharp
function Branch(value : Bool) : Int {
    if (value) {
        3
    } else {
        4
    }
}
@EntryPoint()
operation Main() : Int {
    use qubits = Qubit[2];
    H(qubits[0]);
    CNOT(qubits[0], qubits[1]);
    let result = Branch(MResetZ(qubits[0]) == Zero);
    result
}
```

This appears to be because of how the `BranchControlFlow::Return` is used to short circuit the handling of the rest of the block:

https://github.com/microsoft/qsharp/blob/cd5c2e3b526ded2d01f589e9333890f3832ab6f0/compiler/qsc_partial_eval/src/lib.rs#L1837-L1839

This causes the generated RIR program to be missing branch and jump terminator instructions at the end of blocks. The empty block (should have one terminating jump) is what triggers the specific panic, but other non-empty blocks also are missing terminators that mean the program is invalid.

Either the early return should be fully handled by refactoring the control flow handling for dynamic branches (which must correctly handling inlining for functions like the example above) or the `RuntimeFeatureFlags::ReturnWithinDynamicScope` should be updated to require more than just `TargetCapabilityFlags::Adaptive` so that RCA rejects this pattern at compile time.

swernli:
This change avoids the panic in early returns by blocking them in RCA and providing the user with compile-time feedback before failing during QIR generation. It also includes updates to partial evaluation to ensure that the panicking case is now treated as a unimplemented graceful failure, since we don't have proper support for it.

Since QIR generation for programs with early return have never properly been handled, this doesn't take any functionality away but rather ensures earlier indication of the unsupported patterns at compile time. I filed #2387 to cover the additional work needed to fully support dynamic explicit returns.

Fixes #2290

### DIFF
--- a/compiler/qsc_partial_eval/src/tests/branching.rs
+++ b/compiler/qsc_partial_eval/src/tests/branching.rs
@@ -2311,3 +2311,95 @@ fn if_expression_with_dynamic_operand_from_hybrid_doubles_array() {
                 Jump(6)"#]],
     );
 }
+
+#[test]
+fn if_expression_with_implicit_return_in_callable_supported() {
+    let program = get_rir_program(indoc! {r#"
+        function Choose(r : Result) : Int {
+            if r == One {
+                1
+            } else {
+                0
+            }
+        }
+        @EntryPoint()
+        operation Main() : Int {
+            use q = Qubit();
+            Choose(MResetZ(q))
+        }
+        "#
+    });
+
+    assert_blocks(
+        &program,
+        &expect![[r#"
+            Blocks:
+            Block 0:Block:
+                Call id(1), args( Qubit(0), Result(0), )
+                Variable(0, Boolean) = Call id(2), args( Result(0), )
+                Variable(1, Boolean) = Store Variable(0, Boolean)
+                Branch Variable(1, Boolean), 2, 3
+            Block 1:Block:
+                Variable(3, Integer) = Store Variable(2, Integer)
+                Call id(3), args( Variable(3, Integer), Pointer, )
+                Return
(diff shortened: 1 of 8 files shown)


compiler/qsc_partial_eval/src/tests/branching.rs
```diff
@@ -2311,3 +2311,95 @@ fn if_expression_with_dynamic_operand_from_hybrid_doubles_array() {
                 Jump(6)"#]],
     );
 }
+
```
DmitryVasilevsky, line 2314: We didn't have any "positive" tests for this, did we?

DmitryVasilevsky (approved):

compiler/qsc_partial_eval/src/tests/branching.rs
swernli: We did not, because it turns out we never properly supported it. That's why it makes sense to reject it at compile time.

PR #2390, opened 2025-05-09
Unify std libs

### LINKED ISSUE #2383: References to QasmStd Items don't Support GoTo Def
**Describe the bug**

When trying to use the language server's go-to-definition feature of references to QasmStd items, an error occurs.

**To Reproduce**

```qsharp
import QasmStd.Convert.BoolAsBigInt;
operation Main() : Unit {
    BoolAsBigInt(true);
}
```

The try to use go-to-definition on `BoolAsBigInt`. It will try to open a readonly view of Convert.qs, as it would with any library item, but the view displays a generic "unexpected error" message.

**Expected behavior**

The go-to-definition feature should ideally open a readonly view of the definition of the QasmStd item, or at least use the more graceful "no definition found" error message the user would normally get when trying to go-to-def on invalid items.

**Screenshots**

Image

### LINKED ISSUE #2384: QasmStd Items Should Not Show Up In Completions
**Describe the bug**

QasmStd items should be hidden from normal usage of Q# by not showing up in the language server's Completions suggestions.

**Additional context**

The namespace should also not show up in the API documentation.

### LINKED ISSUE #2385: QasmStd Items aren't Recognized in Dependencies
**Describe the bug**

While QasmStd items are recognized when righting code for a project, they are suddenly not recognizable when another project uses the first as a dependency.

**To Reproduce**

1. Create Q# project Foo, which uses any QasmStd item, like `QasmStd.Convert.BoolAsBigInt`:
```qsharp
import QasmStd.Convert.BoolAsBigInt;
operation FooOp() : Unit {
    BoolAsBigInt(true);
}
export FooOp;
```
2. Create Q3 project Bar, which uses Foo as a dependency:
```json
{
  "dependencies": {
    "Foo": {
      "path": "../Foo"
    }
  }
}
```
```qsharp
import Foo.FooOp
operation Main() : Unit {
    FooOp();
}
```
3. Try to run Main from Bar.
4. You should see an error message about how `BoolAsBigInt` is not found.

**Expected behavior**

Should be able to use dependencies that make use of QasmStd.

idavis:
This change should:
- Fix #2383
- Fix #2384
- Fix #2385


library/std/src/Std/Qasm/Intrinsic.qs
```diff
@@ -73,10 +77,10 @@ operation U(theta : Angle, phi : Angle, lambda : Angle, qubit : Qubit) : Unit is
         let phi = AngleAsDouble(phi);
         let lambda = AngleAsDouble(lambda);
 
-        Rz(lambda, qubit);
-        Ry(theta, qubit);
-        Rz(phi, qubit);
-        R(PauliI, -lambda - phi - theta, qubit);
+        Std.Intrinsic.Rz(lambda, qubit);
```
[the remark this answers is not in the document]
idavis: I didn't want anything accidentally visible to anythining else or spending time resolving.

language_service/src/completion/global_items.rs
```diff
@@ -281,6 +281,10 @@ impl<'a> Globals<'a> {
                     if !is_user_package && candidate_ns == ["Main".into()] {
                         return None;
                     }
+                    // filter out QASM namespaces
+                    if !is_user_package && namespace.name().to_lowercase().contains("qasm") {
```
ScottCarda-MS, line 285: I think I'd feel safer with something a little more specific. If the name started with "Std.Qasm." I think would be a safer check. I worry that the check as-is could unintentionally trigger on namespaces in user-defined library packages.

idavis: I left it as-is waiting for naming bikeshedding to be finished

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -22,9 +22,13 @@ export dcx, ecr, r, rzx, cs, csdg, sxdg, csx, cu1, cu3, rccx, c3sqrtx, c3x, rc3x
```
DmitryVasilevsky, line 22: It's about a function in Convert.qs, but I can't add a comment there. There's a function

```
/// The ``RESULT_AS_BOOL`` function is used to implement the cast expr in QASM for bit to bool.
/// This already exists in the Q# library, but is defined as a marker for casts from QASM.
function ResultAsBool(input : Result) : Bool {
    Std.Convert.ResultAsBool(input)
}
```

Do we still need now that we are moving everything to the standard library?

ScottCarda-MS (approved):

library/std/src/Std/OpenQASM/Intrinsic.qs
idavis: Yes, it is a marker for now.

PR #2395, opened 2025-05-14
Fix bug with Circuit CSS not being applied to notebooks

ScottCarda-MS:
Nests the circuit-specific classes under the `qs-circuit` class, which is universal to the circuit widget, instead of the `qs-circuit-panel` class, which is exclusive to the vscode panel. This fixes an issue where CSS classes were not being applied in the jupyter notebook scenario.

### DIFF
--- a/npm/qsharp/ux/qsharp-circuit.css
+++ b/npm/qsharp/ux/qsharp-circuit.css
@@ -9,7 +9,10 @@
 }
 
 /* Circuit widget */
-.qs-circuit-widget {
+.qs-circuit {
+  /* Background color for the circuit diagram */
+  background: var(--main-background);
+
   /* Error div in circuit panel */
   .qs-circuit-error {
     padding: 10px 0px;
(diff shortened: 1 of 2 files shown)


npm/qsharp/ux/qsharp-circuit.css
```diff
@@ -6,7 +6,10 @@
   p {
     padding: 10px 0px;
   }
+}
 
+/* Circuit widget */
+.qs-circuit-widget {
```
billti, line 12: I'm not sure you need another layer of CSS classes here. Ultimately, everything you care about styling in this block is already in a `qs-circuit` styled DIV. A lot of the styles below are already prefixed with `.qs-circuit`. Maybe you just need to elevate `.qs-circuit` to be the outer CSS scope here.

ScottCarda-MS: done

billti (approved):

PR #2396, opened 2025-05-14
Don't use the start of the target as the provider id

billti:
For targets coming online, the providerId is not necessarily the first part of the target string up to the first ".". Thus, saving the providerId along with the target (which is easier than having each child node store a reference to it's parent also)

### DIFF
--- a/vscode/src/azure/networkRequests.ts
+++ b/vscode/src/azure/networkRequests.ts
@@ -25,6 +25,7 @@ export async function azureRequest(
 
   try {
     log.debug(`Fetching ${uri} with method ${method}`);
+    log.trace("Request headers & body: ", headers, body);
     const response = await fetch(uri, {
       headers,
       method,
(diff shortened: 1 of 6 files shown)


vscode/src/azure/networkRequests.ts
```diff
@@ -25,6 +25,7 @@ export async function azureRequest(
 
   try {
     log.debug(`Fetching ${uri} with method ${method}`);
+    log.trace("Request headers & body: ", headers, body);
```
billti, line 28: I added this to help me troubleshoot the issue, so leaving it in for future use. We should capture network requests in 'trace' logging.

minestarks: I love debug/trace logging - thanks for adding this

minestarks (approved):

vscode/src/azure/networkRequests.ts
DmitryVasilevsky: Any potential privacy issues here?

DmitryVasilevsky (approved):

PR #2409, opened 2025-05-15
Updates some of the CSS colors for Circuits

ScottCarda-MS:
Made some suggested changes to the color schemes of some of the details in the Circuit Editor:
 - The disabled "OK" button now uses the same colors as other buttons and just uses an opacity to indicate its unavailability.
image
 - The default color for button boarders is now transparent so if a theme doesn't define a border color, boarders do not appear.
 - The argument on gates will remain underlined and have the "clickable" cursor on hover, but the color will not be hyperlink blue as that did not mesh well with certain themes. The color is now the same as the regular text color.
image

### DIFF
--- a/npm/qsharp/ux/qsharp-circuit.css
+++ b/npm/qsharp/ux/qsharp-circuit.css
@@ -83,16 +83,16 @@
 
   /* Edit mode style for arg-button (hyperlink appearance) */
   .arg-button.edit-mode {
-    fill: var(--vscode-textLink-foreground, #0066cc);
+    /* fill: var(--vscode-textLink-foreground, #0066cc); */
     text-decoration: underline;
     cursor: pointer;
     pointer-events: all;
   }
 
   /* Hover style for arg-button in edit mode */
-  .arg-button.edit-mode:hover {
-    fill: var(--vscode-textLink-activeForeground, #004499); /* Hover color */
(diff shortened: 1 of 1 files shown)


npm/qsharp/ux/qsharp-circuit.css
```diff
@@ -83,17 +83,11 @@
...
-    fill: var(--vscode-textLink-foreground, #0066cc);
     text-decoration: underline;
     cursor: pointer;
     pointer-events: all;
   }
 
-  /* Hover style for arg-button in edit mode */
-  .arg-button.edit-mode:hover {
-    fill: var(--vscode-textLink-activeForeground, #004499); /* Hover color */
-  }
```
minestarks, line 95: I wonder if we can do something else with hover, like put a box around it or something... to make it obvious that it's interactable. I may be overthinking it.

minestarks (approved):

npm/qsharp/ux/qsharp-circuit.css
ScottCarda-MS: Not sure what else to do for this here. Copilot gave some suggestions but they weren't what were going for. I'll move forward with this PR as-is and if we come up with something clever in the future we can always add it in then.

Issue #2378, opened 2025-05-07
Support simpler array update syntax


swernli:
A major pain point for folks learning Q# (or AI generating it) is the copy-update syntax used for modifying arrays:

```qsharp
mutable a = [1, 2, 3];
let idx = 0;
let update = 0;
a w/= idx <- update; // Sets the first entry to be `0`
```

We should consider a simpler syntax that is more intuitive while either conforming to current memory copying patterns or introducing new patterns.

One possible solution: we introduce new syntactic suger that is lowered into the same `hir::ExprKind::AssignIndex` structure such that runtime behavior and analysis via Runtime Capabilities Analysis remains the same. We could support familiar syntax like:

```qsharp
mutable a = [1, 2, 3];
let idx = 0;
let update = 0;
a[idx] = update; // Sets the first entry to be `0`
```

However, we'd have to be careful about how to handle different forms of indexing (like range-based updates) and arrays of arrays. So something like:

```qsharp
mutable a = [ [1, 2, 3], [1, 2, 3] ];
a[0][0] = 0;
```

would have to be treated as the equivalent of the older syntax:

```qsharp
mutable a = [ [1, 2, 3], [1, 2, 3] ];
a w/= 0 <- (a[0] w/ 0 <- 0);
```

Where the older syntax more clearly communicates that a copy of the entire inner array is incurred.

PR #2414, opened 2025-05-15
Support Array Update Syntax

ScottCarda-MS:
This PR supports array syntax as described in #2378.
```qsharp
arr[0][1][2] = 3
```
gets transformed in the HIR to
```qsharp
arr w/= 0 <- (arr[0] w/ 1 <- (arr[0][1] w/ 2 <- 3))
```
This allows user to not have to use the `w/=` syntax to update arrays. This also supports range expressions for the index. It is also careful to make sure the index expressions are only evaluated once and in order.

This PR also adds lints for discouraging the use of `w/=` and `w/` syntax, with a code action for the `w/=` to convert it to the new syntax here. Both of those lints are set to `Allow` by default, but can be changed to `Warn` in the future.

This also supports compound assignment operators, so you can do `arr[0][1][2] += 3` and it will be converted to the new syntax.

### DIFF
--- a/compiler/qsc_linter/src/lints/ast.rs
+++ b/compiler/qsc_linter/src/lints/ast.rs
@@ -32,7 +32,8 @@ declare_ast_lints! {
     (DeprecatedNewtype, LintLevel::Allow, "deprecated `newtype` declarations", "`newtype` declarations are deprecated, use `struct` instead"),
     (DeprecatedSet, LintLevel::Allow, "deprecated use of `set` keyword", "the `set` keyword is deprecated for assignments and can be removed"),
     (DiscourageChainAssignment, LintLevel::Warn, "discouraged use of chain assignment", "assignment expressions always return `Unit`, so chaining them may not be useful"),
-    (DiscourageUpdateExpr, LintLevel::Warn, "discouraged use of update expressions", "update expressions (w/=, w/) are discouraged; consider using explicit assignment instead"),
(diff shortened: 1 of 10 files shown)


ScottCarda-MS:
Making this a draft again to rework the AST lint so that it won't collide with the existing HIR lint for structs.

compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -30,6 +33,8 @@ declare_ast_lints! {
     (DeprecatedNewtype, LintLevel::Allow, "deprecated `newtype` declarations", "`newtype` declarations are deprecated, use `struct` instead"),
     (DeprecatedSet, LintLevel::Allow, "deprecated use of `set` keyword", "the `set` keyword is deprecated for assignments and can be removed"),
     (DiscourageChainAssignment, LintLevel::Warn, "discouraged use of chain assignment", "assignment expressions always return `Unit`, so chaining them may not be useful"),
+    (DeprecatedAssignUpdateExpr, LintLevel::Allow, "deprecated use of update assignment expressions", "update assignment expressions \"a w/= b <- c\" are deprecated; consider using explicit assignment instead \"a[b] = c\""),
+    (DeprecatedUpdateExpr, LintLevel::Allow, "deprecated use of update expressions", "update expressions \"a w/ b <- c\" are deprecated; consider using explicit assignment instead"),
```
minestarks, line 37: I don't immediately see it in this PR, but we should have a lint group for deprecations that @orpuente-MS  recently added, right? Should these go under there?

swernli (approved):

compiler/qsc_linter/src/lints/ast.rs
ScottCarda-MS: done

PR #2422, opened 2025-05-16
Drop `ms.topic` Metadata Field for all Non-Index Doc Files

ScottCarda-MS:
As requested by the doc team, they now have a script that will add the `ms.topic` field with a value that they determine, so we shouldn't generate `ms.topic` fields.

### DIFF
--- a/compiler/qsc_doc_gen/src/generate_docs.rs
+++ b/compiler/qsc_doc_gen/src/generate_docs.rs
@@ -29,7 +29,6 @@ type ToC = FxHashMap<Rc<str>, Vec<Rc<Metadata>>>;
 struct Metadata {
     uid: String,
     title: String,
-    topic: String,
     kind: MetadataKind,
     package: PackageKind,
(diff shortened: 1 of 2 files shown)


compiler/qsc_doc_gen/src/generate_docs.rs
```diff
@@ -60,9 +59,9 @@ description: {}
 author: {{AUTHOR}}
 ms.author: {{MS_AUTHOR}}
 ms.date: {{TIMESTAMP}}
-ms.topic: {}
+ms.topic: landing-page
```
DmitryVasilevsky, line 62: So landing-page is the only ms.topic that we leave to table of content?

DmitryVasilevsky (approved):

compiler/qsc_doc_gen/src/generate_docs.rs
ScottCarda-MS: Yes, that's right

PR #2423, opened 2025-05-16
Remove `w/=` and `w/` operations in our repo's Q# code

ScottCarda-MS:


swernli:
Of note with this one, we have a bit of a timing challenge: we should merge after the release with the feature goes out since the samples are treated as a live resource BUT we must merge before the release with the feature goes out so that the embedded samples have the new syntax. I think this means we have to hit that window of merging just before we know we are going to do an official release.

samples/language/CopyAndUpdateOperator.qs
minestarks, line 1: This sample also shows the copy and update syntax for structs. Is this intentionally deleted?

minestarks (approved):

minestarks:
@swernli  why, what do you have in mind when you say "live resource"? Do we have direct links to the `main` of this repo from documentation?

swernli:
> @swernli why, what do you have in mind when you say "live resource"? Do we have direct links to the `main` of this repo from documentation?

Yeah, last I checked we do. We still get referrals from learn.microsoft.com and quantum.microsoft.com, and I believe the learn site still deep links to the samples folder. It's possible those links have changed since I last looked though.

samples/language/CopyAndUpdateOperator.qs
ScottCarda-MS: We should probably have two new samples to replace this file: one for the copy syntax of structs and a separate one for the array index assignment.

swernli (approved):
Changes look good to me!

swernli:
Closing for now, keeping the branch and will reopen when/if this is rebased and ready to move forward.

PR #2425, opened 2025-05-16
Add documentType to telemetry events

minestarks:
This adds a `documentType` property to relevant telemetry events to differentiate between the various language types we now support.


vscode/src/copilot/azqTools.ts
```diff
@@ -433,7 +431,7 @@ export async function submitToTarget(
 
   let qir = "";
   try {
-    qir = await getQirForVisibleSource(supportsAdaptive(target.id));
+    qir = "";
   } catch (e: any) {
```
idavis, line 435: I think you forgot to put this back in.

minestarks: This was indeed a mistake. Let me put it back in

idavis (approved):

PR #2427, opened 2025-05-18
Only surface Q# errors if there are no OpenQASM compilation errors

idavis:


compiler/qsc/src/qasm.rs
```diff
@@ -83,16 +84,28 @@ pub fn compile_with_config<R: SourceResolver, S: Into<Arc<str>>>(
...
 
-    let mut compile_errors = compile_errors;
-    for error in errors {
-        let err = WithSource::from_map(
-            &source_map,
-            crate::compile::ErrorKind::OpenQasm(error.into_error()),
-        );
-        compile_errors.push(err);
-    }
+    // We allow the best effort compliation, but for errors we only
```
orpuente-MS, line 87:
```suggestion
    // We allow the best effort compilation, but for errors we only
```

orpuente-MS (approved):

PR #2432, opened 2025-05-19
Remove warnings and update module name.

idavis:


pip/qsharp/openqasm/__init__.py
```diff
@@ -4,15 +4,15 @@
...
+from ._import import import_openqasm
 from ._run import run
 from .._native import ProgramType, OutputSemantics, QasmError  # type: ignore
 
 __all__ = [
     "circuit",
     "compile",
     "estimate",
-    "import_qasm",
+    "import_openqasm",
```
swernli, line 15: It's too bad this can't just be called `import` but I'm guessing that's because that is already a Python keyword, right?

swernli (approved):

pip/qsharp/openqasm/__init__.py
idavis: Yup, that was my first choice.

PR #2433, opened 2025-05-19
Adds a python integration test for circuits

ScottCarda-MS:
Also fixes a small capitalization inconsistency in the context menu for circuit editor.

### DIFF
--- a/npm/qsharp/ux/circuit-vis/contextMenu.ts
+++ b/npm/qsharp/ux/circuit-vis/contextMenu.ts
@@ -78,14 +78,14 @@ const addContextMenuToHostElem = (
         circuitEvents.renderFn();
       });
 
-      const addControlOption = _createContextMenuItem("Add control", () => {
+      const addControlOption = _createContextMenuItem("Add Control", () => {
         if (selectedOperation.kind !== "unitary") return;
         circuitEvents._startAddingControl(selectedOperation, selectedLocation);
       });
 
       let removeControlOption: HTMLDivElement | undefined;
       if (selectedOperation.controls && selectedOperation.controls.length > 0) {
-        removeControlOption = _createContextMenuItem("Remove control", () => {
+        removeControlOption = _createContextMenuItem("Remove Control", () => {
           circuitEvents._startRemovingControl(selectedOperation);
         });
         contextMenu.appendChild(removeControlOption);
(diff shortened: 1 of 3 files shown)


npm/qsharp/ux/circuit-vis/contextMenu.ts
minestarks, line 1: Looks like you're going for consistency, which is great, but I feel like sentence case makes more sense than title case?

pip/tests/test_project.py
```diff
@@ -74,6 +74,18 @@ def test_github_dependency(qsharp) -> None:
...
 
+def test_circuit(qsharp) -> None:
+    qsharp.init(project_root="/circuit")
+    result = qsharp.eval("Test.TestCircuit()")
+    assert len(result) == 2
+    assert result[0] == result[1]
+
+
+with open(
+    os.path.join(os.path.dirname(__file__), "circuit.qsc"), "r", encoding="utf-8"
```
minestarks, line 85: I think it'd be nicer if you just inlined this file into a string (there's no I/O at all in these unit tests at the moment, which I quite enjoy )

minestarks (approved):

pip/tests/test_project.py
ScottCarda-MS: 😬 Are you sure about that being nicer? I'd be hard-coding the circuit JSON format into the test, which is a lot of text, more than any other test. It also means that devs can't look at/edit that circuit via the circuit editor.

npm/qsharp/ux/circuit-vis/contextMenu.ts
ScottCarda-MS: Most context menu items use Title Case over Sentence case. I think, for consistency, we should stick with Title Case here.

pip/tests/test_project.py
minestarks: Ehh, I can see both sides. I personally prefer keeping the directory structure uncluttered and tests consistent (all the other tests in this file hardcode the source content into strings). The JSON doesn't have to be this long either - you could have 1 qubit and 1 measurement, since the point of the test is validating project compilation, not circuit correctness.

But if you prefer this I don't mind if you choose to keep it.

ScottCarda-MS: I've minimized the circuit to just a single measurement as you suggested. You are right that the test really shouldn't be checking for circuit correctness. I'll stick with the separate circuit file though as I'm against hardcoding raw circuit JSON because I think that will create maintainability issues.

billti (approved):

PR #2437, opened 2025-05-20
Add sample notebooks

idavis:
This PR adds sample Python integration and resource estimation notebooks for OpenQASM. There is more here than just those changes as existing external issues require workarounds.

- Qiskit and existing OpenQASM code incorrectly specify 0 for angles which isn't allowed in the spec. So this PR adds the ability to cast literals to angles with a value of 0 for compatibility. This also removes some casting calls as we can now directly create the angle from the literal.
- There is a util bitstring function which allows the user to specify that they want the output of simulation to be converted into a bitstring if any of the outputs are bit registers. This matches what Qiskit output would be like. For the moment the code defaults to Q#'s typicaly output rendering.

### DIFF
--- a/samples/estimation/estimation-openqasm.ipynb
+++ b/samples/estimation/estimation-openqasm.ipynb
@@ -1,0 +1,609 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Getting Started with QDK Resource Estimation using OpenQASM\n",
+        "\n",
+        "👋 Welcome to the QDK Resource Estimator using OpenQASM. In this notebook we will\n",
+        "guide you how to estimate and analyze the physical resource estimates of a\n",
+        "quantum program targeted for execution based on the architecture design of a\n",
+        "fault-tolerant quantum computer. As a running example we are using a multiplier."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "## Implementing the algorithm"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "As a first step, we will create a sample application which will be used throughout this Resource Estimation notebook. To start, we'll import some utilities from the `qiskit` Python package eventually generating the OpenQASM program."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "tags": [
+          "parameters"
+        ]
+      },
+      "outputs": [],
+      "source": [
+        "# These are the Python imports that we use in this Qiskit-based example\n",
+        "from qiskit.circuit.library import RGQFTMultiplier"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "We are creating a quantum circuit for a multiplier based on the construction presented in [arXiv:1411.5949](https://arxiv.org/abs/1411.5949) which uses the Quantum Fourier Transform to implement arithmetic. You can adjust the size of the multiplier by changing the `bitwidth` variable. The circuit generation is wrapped in a function that can be called with the bitwidth of the multiplier. The circuit will have two input registers with that bitwidth, and one output register with the size of twice the bitwidth."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "def create_algorithm(bitwidth):\n",
+        "    print(f\"[INFO] Create a QFT-based multiplier with bitwidth {bitwidth}\")\n",
+        "\n",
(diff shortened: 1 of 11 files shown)


samples/notebooks/openqasm.ipynb
```diff
@@ -0,0 +1,329 @@
...
+      "source": [
+        "# QDK Interop with OpenQASM"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "2b838c23",
+      "metadata": {},
+      "source": [
+        "The modern QDK provides interoperability with OpenQASM 3 programs built upon the core QDK compiler infrastructure.\n",
```
billti, line 16: It's just 'the QDK', we don't use 'modern' any more when refering to it.

samples/notebooks/openqasm.ipynb
```diff
@@ -0,0 +1,329 @@
...
+        "\n",
+        "This core enables integration and local resource estimation without relying on external tools. Users are able to estimate resources for their OpenQASM programs locally (see the [resource estimation with OpenQASM sample notebook](../../estimation/estimation-openqasm.ipynb)), leveraging the QDK compiler's capabilities for analysis, transformation, code generation, and simulation. This also enables the generation of QIR from OpenQASM progams leveraging the [modern QDKs advanced code generation capabilities](https://devblogs.microsoft.com/qsharp/integrated-hybrid-support-in-the-azure-quantum-development-kit/).\n",
```
billti, line 18: Isn't `../../estimation` one level to high if you're already in `samples/notebooks`?

samples/notebooks/openqasm.ipynb
```diff
@@ -0,0 +1,329 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "981d489a",
+      "metadata": {},
+      "source": [
+        "### Compiling OpenQASM to Quantum Intermediate Representation (QIR)\n",
+        "\n",
+        "We can directly compile OpenQASM to QIR with the `compile` function. This "
```
billti, line 76: The final sentance appears truncated.

samples/notebooks/openqasm.ipynb
```diff
@@ -0,0 +1,351 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "id": "5bbd6d92",
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "from qsharp import run, DepolarizingNoise\n",
+        "from qsharp_widgets import Histogram\n",
+        "\n",
+        "Histogram(run(\"bell()\", shots=1000, noise=DepolarizingNoise(0.01)))"
```
swernli, line 226: This works fine, but why not use the callable object like the other samples? I think there was a bug previously where this didn't work as expected, but now it seems to work just fine.
```suggestion
        "Histogram(run(bell, shots=1000, noise=DepolarizingNoise(0.01)))"
```

swernli (approved):

samples/notebooks/openqasm.ipynb
idavis: The text above this line talks about how to call it just like you would in a %%qsharp cell. We can change this but we need to update more for consistency.

samples/estimation/estimation-openqasm.ipynb
```diff
@@ -0,0 +1,609 @@
...
+      "source": [
+        "As a first step, we will create a sample application which will be used throughout this Resource Estimation notebook. To start, we'll import some utilities from the `qiskit` Python package eventually generating the OpenQASM program."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "tags": [
+          "parameters"
```
billti, line 34: What is this metadata/tags/parameters for? I think that's why I'm seeing the below when I open the notebook

image

samples/estimation/estimation-qiskit.ipynb
```diff
@@ -95,7 +95,7 @@
         "from qsharp.interop.qiskit import estimate\n",
         "\n",
         "params = EstimatorParams()\n",
-        "result = estimate(circ, params)"
+        "result = estimate(circ, params, skip_transpilation=True)"
```
billti, line 98: Why are these parameters now needed in this notebook? Is this a breaking change for existing notebooks customers might have? (Do we need a doc update for existing RE APIs to cover this?)

samples/notebooks/openqasm.ipynb
```diff
@@ -0,0 +1,359 @@
...
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.13.3"
+    }
+  },
+  "nbformat": 4,
+  "nbformat_minor": 5
+}
```
billti, line 359: Nice! I really like the clarity and simplicity of this notebook. It flows really well.

billti (approved):
Couple minor comments/issues, then good to go. Thanks!

PR #2438, opened 2025-05-20
Update branding to Microsoft Quantum Development Kit

billti:


jupyterlab/README.md
```diff
@@ -16,7 +16,7 @@ To run Q# in Jupyter notebooks, remember to also install the `qsharp` package: [
 
 ## Support
 
-For more information about the Azure Quantum Development Kit, visit [https://aka.ms/AQ/Documentation](https://aka.ms/AQ/Documentation).
```
billti, line 19: Switching a few links from the generic "Azure Quantum" landing page to the QDK specific page.

samples/estimation/estimation-qiskit.ipynb
```diff
@@ -590,7 +590,7 @@
       "name": "python3"
     },
     "kernelspec": {
-      "display_name": "Python 3 (ipykernel)",
```
ScottCarda-MS, line 593: This display name (and python version number below) change every time I open a notebook, but they change to a different value than the one you list here. We should probably have a standard we agree on for the name and version across these notebooks. I think "Python 3" is a better display name than ".venv".

.gitignore
```diff
@@ -20,3 +20,4 @@ __pycache__/
 .pytest_cache/
 .idea/
 *.so
+scratch/
```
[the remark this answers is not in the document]
minestarks: I don't mind this, but have you considered adding a personal .gitignore file with `git config --local core.excludesFile .gitignore-local` ?

vscode/README-DEV.md
```diff
@@ -1,9 +1,9 @@
...
 
-This is a development build of the Azure Quantum Development Kit, and is intended for early testing
+This is a development build of the Microsoft Quantum Development Kit, and is intended for early testing
 of upcoming features.
 
 To install the tested version of this extension, visit <https://marketplace.visualstudio.com/items?itemName=quantum.qsharp-lang-vscode>.
 The README for the stable build contains details on features and usage.
 
-For more information about the Modern QDK and Azure Quantum, visit [https://aka.ms/AQ/Documentation](https://aka.ms/AQ/Documentation).
+For more information about the Modern QDK and Microsoft Quantum, visit [https://aka.ms/qdk](https://aka.ms/qdk).
```
minestarks, line 9: How do we feel about the "Modern QDK" now?

vscode/package.json
```diff
@@ -432,12 +432,12 @@
...
         "command": "qsharp-vscode.setTargetProfile",
         "category": "QDK",
-        "title": "Set the Azure Quantum QIR target profile"
+        "title": "Set the QIR target profile"
       },
       {
         "command": "qsharp-vscode.createNotebook",
         "category": "QDK",
-        "title": "Create an Azure Quantum notebook"
+        "title": "Create a Microsoft Quantum notebook"
```
minestarks, line 440: This one creates a sample that _actually_ uses Azure Quantum, so the name may actually be appropriate?

vscode/package.json
```diff
@@ -31,13 +31,13 @@
     "walkthroughs": [
       {
         "id": "qsharp-vscode.welcome",
-        "title": "The Azure Quantum Development Kit",
-        "description": "Getting started with the Azure Quantum Development Kit in VS Code",
+        "title": "The Microsoft Quantum Development Kit",
```
minestarks, line 34: Maybe just "Microsoft Quantum Development Kit"? (or are we going for something like "The Batman" 😜)

vscode/src/notebookTemplate.ts
```diff
@@ -11,7 +11,7 @@ export const notebookTemplate: vscode.NotebookData = {
     {
       kind: vscode.NotebookCellKind.Markup,
       languageId: "markdown",
-      value: `## Azure Quantum Q# notebook
+      value: `## Microsoft Quantum Q# notebook
```
minestarks, line 14: Same as above - the notebook content does include Azure Quantum so the name may be fine?

minestarks (approved):

.gitignore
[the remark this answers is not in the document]
DmitryVasilevsky: Maybe we can add a readme file in this folder, check it in and then add this git ignore. I have a need for this as well, but I often use samples folder instead.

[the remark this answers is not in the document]
swernli: wouldn't that make our already busy top-level folder structure even more busy?

For my part, I don't think this is change makes sense to do here, but if others are fine with it then it's not the worst.

vscode/README-DEV.md
billti: Good catch - we hate it! Will delete and look for other occurence :-)

.gitignore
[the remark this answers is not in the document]
billti: > wouldn't that make our already busy top-level folder structure even more busy?

But nothing is going to get checked in there - that's the point of it being in the .gitignore. I just like to have a 'working folder' for experiements when I have the qsharp workspace open but not have the git change list cluttered with things I don't want to actually check-in (also avoids accidentally checking them in). (I often put them under `samples` also as Dmitry alluded to, so that's an option).

I can take it out of here and discuss separately. It's unrelated.

samples/estimation/estimation-qiskit.ipynb
billti: It's going to be whatever kernel was active when the notebook was saved. If using a virutal environment, (which folks should be), then it was show that, and ".venv" is a very common pattern (e.g. the PR Ian just opened shows the same change - https://github.com/microsoft/qsharp/pull/2437/files#diff-09a59e4bf67ef253c4865a8c0a11f945967509461947ddca865017d159064bbb )

I don't think the value set here can be generic in an effective or cross platform way. Sadly, I think it's best just to ignore it.

.gitignore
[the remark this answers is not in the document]
minestarks: I thought my personal gitinore suggestion was pretty good.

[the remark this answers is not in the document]
idavis: 🚲 🏚️ 🤣

PR #2444, opened 2025-05-21
Fix include error span

orpuente-MS:
Include error spans were set to the beginning of the file, since include errors were being pushed in resolution, where there is no knowledge of spans. This PR delays pushing `IO` errors until lowering, so that we can construct the error with the correct span.


idavis:
Superseded by #2512

PR #2445, opened 2025-05-21
Create copilot-setup-steps.yml

minestarks:
This sets up the environment for [Copilot Coding Agent](https://github.blog/news-insights/product-news/github-copilot-meet-the-new-coding-agent/) to work in.

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,30 @@
+name: Copilot Custom Setup Steps
+
+on:
+  workflow_dispatch:
+
+jobs:
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: "true"
+      - name: Setup rust toolchain
+        uses: ./.github/actions/toolchains/rust
+        with:
+          toolchain: "1.84"
+          components: rustfmt clippy
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20.18.2"
+      - uses: Swatinem/rust-cache@v2
+      - name: Prereqs
+        run: python ./prereqs.py --install
+      - name: First build
+        run: python ./build.py --no-check --no-test


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,30 @@
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: "true"
+      - name: Setup rust toolchain
+        uses: ./.github/actions/toolchains/rust
+        with:
+          toolchain: "1.84"
```
swernli, line 18: I think this is the wrong toolchain version... we are on 1.86 now in the repo.

swernli:
I wonder if there is a way to have this build off of the devcontainer definition we already have in the repo (https://github.com/microsoft/qsharp/blob/main/.devcontainer/devcontainer.json). I try to keep that up to date for internal or external contributors, and it seems like it could be opportunity to take advantage of existing setup.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,30 @@
...
+          components: rustfmt clippy
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20.18.2"
+      - uses: Swatinem/rust-cache@v2
```
idavis, line 26: I don't think we need rust-cache here, it is just for lowering CI build times

minestarks: I do a `./build.py` below to set up the environment before Copilot gets going.... would this cache not help speed up that step? I don't actually know how the cache works.

idavis: Maybe, we can leave for now and see.

.github/workflows/copilot-setup-steps.yml
idavis, line 18:
```suggestion
        with:
          toolchain: "1.86"
```

minestarks:
@swernli I like the idea, but I'll throw the question back at you - can we use that devcontainer for `ci.yml` too? I tried to align this workflow with `ci.yml` since I consider that the source of truth but maybe there's an opportunity for all of it to be aligned?

idavis (approved):

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,33 @@
+name: Copilot Setup Steps
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
```
minestarks, line 7: Will remove the PR trigger. Needed to add this, just so I can test the workflow once before merging.

PR #2452, opened 2025-05-22
Fix panic due to missing `Unit` value from assignment

swernli:
Assignments are expressions that should result in unit, but they don't push unit into the exec graph, which can cause problems if those expressions have their results bound to some other variable or expression.

### DIFF
--- a/compiler/qsc_lowerer/src/lib.rs
+++ b/compiler/qsc_lowerer/src/lib.rs
@@ -728,6 +728,18 @@ impl Lowerer {
             _ => self.exec_graph.push(ExecGraphNode::Expr(id)),
         }
 
+        if matches!(
+            kind,
+            fir::ExprKind::Assign(..)
+                | fir::ExprKind::AssignField(..)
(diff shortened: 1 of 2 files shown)


compiler/qsc_lowerer/src/lib.rs
```diff
@@ -728,6 +728,18 @@ impl Lowerer {
             _ => self.exec_graph.push(ExecGraphNode::Expr(id)),
         }
 
+        if matches!(
```
idavis, line 731: Why not make this park of the match above?

DmitryVasilevsky: I find the single match a bit clearer. That way I see that Expression Id is actually pushed instead of relying on fairly obscure "_" case before that.

swernli: Seems reasonable. I'll make the change.

DmitryVasilevsky (approved):

idavis (approved):

PR #2458, opened 2025-05-27
Added OpenQASM samples into the VSCode Playground

DmitryVasilevsky:
Added OpenQASM samples into the VSCode Playground
- OpenQASM samples added to "Samples (OpenQASM)" folder
- Existing folder with Q# samples is renamed to "Samples (Q#)"

image

### DIFF
--- a/vscode/src/memfs.ts
+++ b/vscode/src/memfs.ts
@@ -70,7 +70,7 @@ function populateSamplesFolder(
     const sanitized_file_name = sample.title.replace(/\W/g, ""); // Remove non-word characters
     vfs.writeFile(
       playgroundRootUri.with({
-        path: `/${folder}/${sanitized_file_name}.${extension}`,
+        path: `/${folder}/${sanitized_file_name}${extension}`,
       }),
       encoder.encode(sample.code),
       { create: true, overwrite: true },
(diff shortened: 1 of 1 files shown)


vscode/src/memfs.ts
```diff
@@ -33,28 +34,49 @@ For more information about the Azure Quantum Development Kit,
...
+  extension: string,
+) {
+  vfs.createDirectory(playgroundRootUri.with({ path: `/${folder}` }));
   samples.forEach((sample) => {
     const sanitized_file_name = sample.title.replace(/\W/g, ""); // Remove non-word characters
     vfs.writeFile(
-      playgroundRootUri.with({ path: `/samples/${sanitized_file_name}.qs` }),
+      playgroundRootUri.with({
+        path: `/${folder}/${sanitized_file_name}${extension}`,
+      }),
```
swernli, line 75: would it be reasonable to use the structure `samples/qsharp/*.qs` and `samples/openqasm/*.qasm` instead? I think that will look cleaner than having `samples_qsharp` and `samples_openqasm`.

DmitryVasilevsky: I renamed folders, probably better with these names. Looks like the virtual file system supports fancy names.

swernli (approved):

PR #2461, opened 2025-05-27
Display Error Message when Circuit is Unrenderable

ScottCarda-MS:


npm/qsharp/src/shared/circuit.ts
minestarks, line 1: I had missed this the first time, but can we pull all this code out of `shared`? It should live in either `ux` or elsewhere in `src`, depending on where it's used - but not `shared` since it's not actually shared between the two components.

`legacyCircuitUpdate.ts` should  come out of this folder entirely.

`shared` was intended to stay minimal and only contain basic constants and types. If it grows, I worry it may start pulling in dependencies accidentally, and create the problems I'd hoped to avoid by creating this directory in the first place. Long story, I had to create this folder when we needed a solution for common types that needed to be referenced _both_ from the UX code in `ux`, and non-UX code under `src/` . Non-UX code (which typically gets loaded in the VS Code extension side) gets compiled with different options than UX code, and when the two kinds of modules end up referencing each other, you end up with horrible TypeScript module errors.

minestarks (approved):

npm/qsharp/src/shared/circuit.ts
ScottCarda-MS: Discussed offline: decided to rename `shared/` to `data-structures/` to make it clearer that that was being treated as a data layer.

PR #2462, opened 2025-05-27
Generating files should be usable more often

idavis:


swernli (approved):
This worked in my testing!

billti (approved):

PR #2465, opened 2025-05-28
Add --set option to version.py

billti:
Simple utility to make bumping the version before release easy. Just run with something like `./version.py --set 1.18.1` to bump in all the necessary places.

### DIFF
--- a/version.py
+++ b/version.py
@@ -4,9 +4,73 @@
 # Licensed under the MIT License.
 
 import os
+import re
 
 # To be updated every time we start a new major.minor version.
 major_minor = "1.17"
+
+root_dir = os.path.dirname(os.path.abspath(__file__))
+
+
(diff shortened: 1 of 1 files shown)


version.py
```diff
@@ -4,10 +4,74 @@
...
+    # Collect any file named "qsharp.json" under the /library directory
+    for root, dirs, files in os.walk(os.path.join(root_dir, "library")):
+        update_list += [
+            os.path.join(root, file) for file in files if file == "qsharp.json"
+        ]
+
+    for file in update_list:
+        update_file(
+            file,
+            r'"ref": "v[0-9.]+"',
```
billti, line 65: Should maybe improve this matching to only match versions with the prior major_minor, but for current manifest this works fine.

minestarks (approved):

PR #2473, opened 2025-05-30
Added ApplyQPE and ApplyOperationPowerCA to Canon namespace

DmitryVasilevsky:
- Added ApplyQPE and ApplyOperationPowerCA to Canon namespace.
- Added doc comments
- Added tests

ApplyQPE along with the previously implemented ApplyQFT are useful primitives for quantum algorithms. To keep these implementations simple and avoid performance penalty no swaps are performed on arguments or results. In most cases such results can be used without problems. Note, that ApplyQFT and ApplyQPE work with little-endian registers.

### DIFF
--- a/compiler/qsc_partial_eval/src/tests/qubits.rs
+++ b/compiler/qsc_partial_eval/src/tests/qubits.rs
@@ -377,6 +377,6 @@ fn qubit_relabel_in_dynamic_block_triggers_capability_error() {
 
     assert_error(
         &error,
-        &expect!["CapabilityError(UseOfDynamicQubit(Span { lo: 62785, hi: 62798 }))"],
+        &expect!["CapabilityError(UseOfDynamicQubit(Span { lo: 62782, hi: 62795 }))"],
     );
 }


library/src/tests/resources/src/qpe.qs
```diff
@@ -0,0 +1,49 @@
...
+        Fact(CheckAllZero(phase), "Expected state |0000⟩");
+
+        // Estimate eigenvalue of gate on eigenvector of |1⟩
+        X(state);
+        ApplyQPE(ApplyOperationPowerCA(_, qs => gate(qs[0]), _), [state], phase);
+        // The eigenvalue of gate on |1⟩ is exp(i * π / 2^power),
+        // which corresponds to the binary fraction 0.0...01 of 2π,
+        // where the 1 is at the position power counting from 0.
+        // The state should be |0...01⟩, or, in a little-endian register
+        // phase[0] = 0, phase[1] = 0, ..., phase[N-power-1] = 1
```
ScottCarda-MS, line 20: This sounds like it is suggesting that the ket notation is expressed in little-endian. I thought ket notations was big-endian.

library/src/tests/resources/src/qpe.qs
```diff
@@ -0,0 +1,49 @@
+namespace Test {
+    import Std.Diagnostics.*;
+
+    operation TestQPE_Phase(gate : Qubit => Unit is Ctl + Adj, power : Int) : Unit {
+        use state = Qubit();
+        use phase = Qubit[4];
+
+        // Estimate eigenvalue of gate on eigenvector of |0⟩
+        ApplyQPE(ApplyOperationPowerCA(_, qs => gate(qs[0]), _), [state], phase);
+        // It should be 1, which corresponds to the binary fraction 0.0000 of 2π.
```
ScottCarda-MS, line 10: "Binary fraction X of Y" is not a phrasing I'm familiar with. Can you explain what that means?

library/std/src/Std/Canon.qs
```diff
@@ -599,6 +599,72 @@ operation ApplyOperationPowerA<'T>(
...
+///
+/// # Remarks
+/// All eigenvalues of a unitary operator are of magnitude 1 and therefore can be represented as
+/// $e^{i\phi}$ for some $\phi \in [0, 2\pi)$. Finding the phase of an eigenvalue is therefore
+/// equivalent to finding the eigenvalue itself. Passing the eigenvector as `targetState`
+/// to the phase estimation operation allows finding the eigenvalue to the desired precision
+/// determined by the length of the `phase` register.
+///
+/// # Reference
+/// - [Quantum phase estimation algorithm](https://en.wikipedia.org/wiki/Quantum_phase_estimation_algorithm)
```
ScottCarda-MS, line 644: I remember getting automated feedback in a different context in MS saying that we should avoid putting the "en." in hyperlinks. I think it allows the user's default language to possible be applied, if it can be for the site.

ScottCarda-MS (approved):

library/std/src/Std/Canon.qs
DmitryVasilevsky: Keep forgetting about this, will update.

library/src/tests/resources/src/qpe.qs
DmitryVasilevsky: This comment is part of a particular test and is very much in the context of ApplyQPE operation. I didn't do a good job explaining it here thoroughly. I'll provide a better explanation.

DmitryVasilevsky: Updated.

library/std/src/Std/Canon.qs
DmitryVasilevsky: Updated.

PR #2476, opened 2025-05-30
Properly expose project load errors in VS Code, improve error handling in WASM/JS

minestarks:
This PR improves how project loading and QIR generation errors are handled in the VS Code extension. Additionally, there are internal changes normalizing how we raise and handle errors coming from the WASM component.

## User facing changes

Previously, whenever project load failed while performing some task, we would show a modal dialog. Additionally we would populate the Problems view with sticky "QDK command error:" errors.

- Removed excessive modal dialogs on project load errors for tasks that don't need them:

  - Test explorer
  - Debugger
  - Copilot tools for Q#
    - Also, more detailed project load errors are communicated to Copilot so it can respond appropriately.

- These user-invoked commands continue to show the modal dialog when project load fails:

  - Histogram
  - Show circuit
  - Generate QIR / Submit to Azure
  - Estimate
  - Generate documentation

- Removed the sticky "QDK command error:" errors for project load, since the language service already reports those errors. We don't want to litter the Problems view with duplicates.

- _Kept_ the sticky errors for when QIR generation fails at runtime. These are runtime errors and therefore are not reported by the language service, so no risk of duplication.

- Improved the contents of these modal error dialogs to be more user-friendly.

## Internals

- Added a new `QdkDiagnostics` class that properly extends JavaScript's `Error` type and contains VS Code-ish diagnostics

- Many WASM functions use a `string` as an error type, which is really a JSON-serialized `IQSharpError[]` array. Given the limitations of wasm-bindgen, and also the fact that this error has to travel across worker threads, we can't really use a more sophisticated error type here.

- So instead, we convert these into `QdkDiagnostics` in JavaScript.

- To make sure the error type flows properly from the worker thread to the main thread, we serialize the errors to cloneable objects.

- Going forward, when adding new WASM functions, we should try to conform to this pattern of error handling. WASM functions should throw `IQSharpError[]` strings wherever appropriate, and the appropriate wrappers should be used in JavaScript to convert it into a `QdkDiagnostics`.

## Before/After

For my own sanity, here's a visual list of the places where error UX changed.

### Project load error in Run / Debug

| Before                                                      | After                                                       |
| ----------------------------------------------------------- | ----------------------------------------------------------- |
| run_projectload_old | run_projectload_new  |

---

### Project load error in Histogram / Circuit / Estimate / QIR generation / Submit to Azure

| Before                                                                  | After                                                                   |
| ----------------------------------------------------------------------- | ----------------------------------------------------------------------- |
| histogram_projectload_old | histogram_projectload_new |

---

### Project load error in Copilot Q# tools / Submit to Azure

| Before                                                              | After                                                               |
| ------------------------------------------------------------------- | ------------------------------------------------------------------- |
| copilot_projectload_old | copilot_projectload_new |

---

### Project load error in Documentation generation

| Before                                                                          | After                                                                           |
| ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| documentation_projectload_old | documentation_projectload_new |

---

### Project load error in Test case run (`vscode/src/language-service/testExplorer.ts`)

| Before                                                        | After                                                         |
| ------------------------------------------------------------- | ------------------------------------------------------------- |
| test_projectload_old | test_projectload_new |

---

### Runtime error in QIR generation

| Before                                                            | After                                                             |
| ----------------------------------------------------------------- | ----------------------------------------------------------------- |
| qirgen_runtimefail_old | qirgen_runtimefail_new |

---

### Errors in Playground QIR/RIR panels

| Before                                                                    | After                                                                     |
| ------------------------------------------------------------------------- | ------------------------------------------------------------------------- |
| playground_runtimefail_old | playground_runtimefail_new |

### DIFF
--- a/npm/qsharp/test/diagnostics.js
+++ b/npm/qsharp/test/diagnostics.js
@@ -33,7 +33,7 @@ function getInvalidQirProgramConfig() {
   };
 }
 
-test("getQir throws QdkDiagnostics on invalid program", async () => {
+test("getQir throws QdkDiagnostics", async () => {
   const compiler = getCompiler();
   const invalidConfig = getInvalidQirProgramConfig();
   await assert.rejects(
@@ -48,7 +48,7 @@ test("getQir throws QdkDiagnostics on invalid program", async () => {
   );
 });
 
-test("getQir throws QdkDiagnostics on invalid program - worker", async () => {
+test("getQir throws QdkDiagnostics - worker", async () => {
   const compiler = getCompilerWorker();
   const invalidConfig = getInvalidQirProgramConfig();
   try {
@@ -76,7 +76,7 @@ const dummyHost = {
   findManifestDirectory: async () => null,
 };
 
-test("ProjectLoader.loadQSharpProject throws QdkDiagnostics on invalid directory", async () => {
+test("loadQSharpProject throws QdkDiagnostics", async () => {
   const loader = getProjectLoader(dummyHost);
   await assert.rejects(
     () => loader.loadQSharpProject("/not/a/real/dir"),


minestarks:
I think request ID is `67f28fe3-6808-40f2-9539-69eecbe77d72` , but not sure I'm looking at the right thing - feel free to reach out privately

vscode/src/diagnostics.ts
```diff
@@ -136,65 +136,75 @@ export function clearCommandDiagnostics() {
...
     clearCommandDiagnostics();
     return await fn();
   } catch (e: unknown) {
-    reportIfQSharpErrors(e);
+    reportIfQSharpErrors(e, options);
     throw e;
   }
 }
 
 /**
```
idavis, line 154: Should this be `reportIfQdkErrors` and should we rename `IQSharpError` or will that have external breaking changes?

minestarks: I can rename this function. `IQSharpError` is theoretically public facing, but I could check usage in QCOM and rename. I was trying to avoid PR creep

PR #2480, opened 2025-06-02
Prevent Stepping into Circuit Files during Debug

ScottCarda-MS:
This adds a de-spanning pass to the compiler to remove the spans for the contents of operations defined via circuit files. The operation declarations themselves retain their spans to support LS features, but the statements in the body of the operations are de-spanned. This prevents the debugger from stepping into the circuit files.

### DIFF
--- a/compiler/qsc_frontend/src/compile/preprocess.rs
+++ b/compiler/qsc_frontend/src/compile/preprocess.rs
@@ -3,13 +3,17 @@
 
 use core::str::FromStr;
 use qsc_ast::{
-    ast::{Attr, ExprKind, Idents, ItemKind, Namespace, PathKind, Stmt, StmtKind, UnOp},
-    mut_visit::MutVisitor,
+    ast::{
+        Attr, ExprKind, Idents, ItemKind, Namespace, Package, PathKind, Stmt, StmtKind,
+        TopLevelNode, UnOp,
+    },
+    mut_visit::{walk_stmt, MutVisitor},
 };
+use qsc_data_structures::span::Span;
 use qsc_hir::hir;
 use std::rc::Rc;
 
-use super::TargetCapabilityFlags;
+use super::{SourceMap, TargetCapabilityFlags};
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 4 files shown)


compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -193,3 +197,75 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
...
+                    .iter()
+                    .any(|s| s.lo <= callable.span.lo && s.hi >= callable.span.hi)
+                    .then(|| {
+                        self.visit_callable_decl(callable);
+                    });
+            }
+        });
+    }
+
+    fn visit_stmt(&mut self, stmt: &mut Stmt) {
```
idavis, line 254: This logic isn't generally safe and is making assumptions about the AST which may or may not be safe depending on the qsc compilatation of the circuits.

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -193,3 +197,75 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
...
+        walk_stmt(self, stmt);
+        self.inside_stmt = false;
+    }
+
+    fn visit_span(&mut self, span: &mut Span) {
+        // Clear the span if it is inside a statement
+        if self.inside_stmt {
+            span.lo = Span::default().lo;
+            span.hi = Span::default().hi;
+        }
```
idavis:
```suggestion
        if self.inside_stmt {
            span.lo = 0;
            span.hi = 0;
        }
```

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -193,3 +197,75 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
...
+        });
+    }
+
+    fn visit_stmt(&mut self, stmt: &mut Stmt) {
+        self.inside_stmt = true;
+        walk_stmt(self, stmt);
+        self.inside_stmt = false;
+    }
+
+    fn visit_span(&mut self, span: &mut Span) {
```
idavis: Why is this only attempting to remove statement spans?

compiler/qsc_frontend/src/compile/preprocess.rs
```diff
@@ -193,3 +197,75 @@ fn matches_config(attrs: &[Box<Attr>], capabilities: TargetCapabilityFlags) -> b
...
+                    });
+            }
+        });
+    }
+
+    fn visit_stmt(&mut self, stmt: &mut Stmt) {
+        self.inside_stmt = true;
+        walk_stmt(self, stmt);
+        self.inside_stmt = false;
+    }
```
idavis:
```suggestion
    fn visit_stmt(&mut self, stmt: &mut Stmt) {
        stmt.span.lo = 0;
        stmt.span.hi = 0;
        walk_stmt(self, stmt);
    }
```

ScottCarda-MS: Is it safe enough for the debugger to remove spans only from the statement node but keep the spans for the underlying expressions, or are there circumstances where we want to remove the spans of the underlying expressions also?

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: It should remove spans from statements and their underlying expressions. We don't want to remove spans elsewhere to avoid invalidating language server functionality.

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: The default for spans is 0-0. I had thought it would be safer to reference that then use hard-coded values, but if the debugger's logic also uses hard-coded zeros then maybe it makes sense to have zeros here for consistency?

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: Can you describe more what you mean when you say it isn't safe? Perhaps you are referring to the fact that there are statements that we don't want to remove spans from, and this relies on the other visit functions to make sure this only gets called for circuit operations.

ScottCarda-MS: I see what you mean now. The logic in this function shouldn't be setting `inside_stmt` to `true` and `false`. It should be saving the value as it is, setting it to true, then afterwards restoring the origin value. This avoids the issues of when a statement existing inside another statement.

ScottCarda-MS: Though this might be unnecessary if all we want to do is de-span statements and not their underlying expressions.

ScottCarda-MS: Decided to only de-span statements.

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: removed

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: removed

compiler/qsc_frontend/src/compile/preprocess.rs
ScottCarda-MS: Went with only de-spanning statements as described here. It makes for simpler, easier to maintain code. One difference from the suggested code, however, is that I used the `Span::default()` instead of setting 0. This is because the debugger logic to check if it should skip a statement checks against `Span::default()`, so it is safer to be consistent with that.

PR #2505, opened 2025-06-03
Simple Quantum Phase estimation sample via ApplyQPE (Base profile)

DmitryVasilevsky:
This adds a sample for Simple Quantum Phase Estimation using ApplyQPE.
Sample is suitable for base profile.

### DIFF
--- a/samples/algorithms/SimplePhaseEstimation.qs
+++ b/samples/algorithms/SimplePhaseEstimation.qs
@@ -7,14 +7,15 @@
 /// Rz gate is used as the unitary operation,
 /// and the eigenvector |1⟩ is chosen for the phase estimation.
 /// Run this sample and check the histogram of results.
+/// This sample is suitable for base profile.
 
 /// # Summary
 /// Estimate the phase of the eigenvalue of the unitary gate U
 /// using Quantum Phase Estimation (QPE) algorithm.
-/// The result is returned as an array of {0,1} integers
+/// The result is returned as an array of `Result` values
 /// representing the binary fraction of the phase in big-endian order.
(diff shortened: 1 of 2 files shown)


samples/algorithms/SimplePhaseEstimation.qs
```diff
@@ -0,0 +1,66 @@
...
+/// Quantum Phase Estimation of a simple unitary operation.
+/// Rz gate is used as the unitary operation,
+/// and the eigenvector |1⟩ is chosen for the phase estimation.
+/// Run this sample and check the histogram of results.
+
+/// # Summary
+/// Estimate the phase of the eigenvalue of the unitary gate U
+/// using Quantum Phase Estimation (QPE) algorithm.
+/// The result is returned as an array of {0,1} integers
+/// representing the binary fraction of the phase in big-endian order.
```
billti, line 16: Is this typical? What's the down-side to just returning an actual float?

DmitryVasilevsky: This is a good question and I've seen different ways of doing this. Originally, I thought I'd just return an array of Results for this sample, so that it's suitable for the base profile. But seeing that on the histogram it wasn't super easy to decipher. I changed it to 0s and 1s so that it looked like binary at least for the time I worked with it. But left it that way. Now I think it should be back to array of results.
I'm preparing another sample to estimate the actual eigenvalue, and it will deal with actual floats.

DmitryVasilevsky: Updated.

msoeken (approved):

PR #2506, opened 2025-06-04
Quantum Phase estimation sample via ApplyQPE

DmitryVasilevsky:
This is a more complex quantum phase estimation sample. It is not suitable for base profile.

### DIFF
--- a/samples/algorithms/PhaseEstimation.qs
+++ b/samples/algorithms/PhaseEstimation.qs
@@ -1,0 +1,86 @@
+/// # Sample
+/// Quantum Phase Estimation
+///
+/// # Description
+/// This sample demonstrates how to use Quantum Phase Estimation (QPE) algorithm
+/// to estimate the eigenvalue of a specific unitary operation.
+/// QPE algorithm is performed by calling `ApplyQPE` library operation.
+/// Run this sample and check the histogram of results.
+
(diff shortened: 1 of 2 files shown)


samples/algorithms/PhaseEstimation.qs
```diff
@@ -0,0 +1,86 @@
...
+/// # Summary
+/// Estimate the eigenvalue of a unitary operation `U` with certain `precision`
+/// for a given eigenvector using Quantum Phase Estimation (QPE) algorithm.
+operation EstimateEigenvalue(
+    U : Qubit[] => Unit is Adj + Ctl,
+    eigenstate : Qubit[],
+    precision : Int
+) : ComplexPolar {
+    // Allocate qubits to be used in phase estimation
+    use phase = Qubit[precision];
```
msoeken, line 42: It would be useful to add an example for iterative phase estimation that only uses a single helper qubit. It's more commonly used than this version of phase estimation that applies the QFT all at once on multiple helper qubits.

DmitryVasilevsky: Yes! More phase estimation approaches and samples are in the backlog. However, I added a standard one to the Canon namespace as it is unlikely to change. Other approaches will be in a separate library later.

PR #2507, opened 2025-06-04
Fix bug in `Controlled SX` with empty controls

swernli:
The decomposition for `Controlled SX` when the controls list is empty incorrectly performed an adjoint SX instead of a normal SX. This fixes the bug and adds tests for this case.


DmitryVasilevsky (approved):
Wow!

PR #2509, opened 2025-06-05
Clarified search goal in the Grover sample comments

DmitryVasilevsky:
Updated comments in the Grover search sample to clarify what we are searching for. In this particular case we are looking for the marked state |01010⟩. I still think we should keep this sample simple by looking for a specific state represented by a hard-coded operation rather than somehow providing arbitrary marked state as a parameter.

### DIFF
--- a/samples/algorithms/Grover.qs
+++ b/samples/algorithms/Grover.qs
@@ -22,15 +22,16 @@ operation Main() : Result[] {
     // Grover's algorithm relies on performing a "Grover iteration" an
     // optimal number of times to maximize the probability of finding the
     // value we are searching for.
-    // You can set the number iterations to a value lower than optimal to
-    // intentionally reduce precision.
-    let iterations = CalculateOptimalIterations(nQubits);
-    Message($"Number of iterations: {iterations}");
+    // You can intentionally set the number iterations to a value lower
+    // than optimal. In this case, the algorithm will find the marked state
+    // with lower probability.
+    let nIterations = IterationsToMarked(nQubits);
+    Message($"Number of iterations: {nIterations}");
 
     // Use Grover's algorithm to find a particular marked state. The marked state
     // we are looking for in this sample is |01010⟩, which is represented
     // by the operation ReflectAboutMarked.
-    let results = GroverSearch(nQubits, iterations, ReflectAboutMarked);
+    let results = GroverSearch(nQubits, nIterations, ReflectAboutMarked);
 
     // Expected result is [Zero, One, Zero, One, Zero] with very high probability.
     return results;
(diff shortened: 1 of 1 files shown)


samples/algorithms/Grover.qs
```diff
@@ -24,8 +27,12 @@ operation Main() : Result[] {
     let iterations = CalculateOptimalIterations(nQubits);
```
minestarks: About the comment just above this: Is it worth mentioning what would happen if I set the number of iterations to a _higher_ value? Is that interesting?

minestarks (approved):

samples/algorithms/Grover.qs
DmitryVasilevsky: You will start going away from the target value as far as I understand. Not very useful, but if it helps understanding...

DmitryVasilevsky: I updated the comments and the names after discussion.

samples/algorithms/Grover.qs
```diff
@@ -22,15 +22,16 @@ operation Main() : Result[] {
     // Grover's algorithm relies on performing a "Grover iteration" an
     // optimal number of times to maximize the probability of finding the
     // value we are searching for.
-    // You can set the number iterations to a value lower than optimal to
-    // intentionally reduce precision.
-    let iterations = CalculateOptimalIterations(nQubits);
-    Message($"Number of iterations: {iterations}");
+    // You can intentionally set the number iterations to a value lower
```
minestarks, line 25: Do we want to say "lower or higher" here, since _any_ divergence from this number will reduce the probability of finding the right answer?

samples/algorithms/Grover.qs
```diff
@@ -63,8 +64,9 @@ operation GroverSearch(
 
 /// # Summary
 /// Returns the optimal number of Grover iterations needed to find a marked
-/// item, given the number of qubits in a register.
-function CalculateOptimalIterations(nQubits : Int) : Int {
+/// item, given the number of qubits in a register. Setting the number of
+/// itertions to a different number may undershoot or overshoot the marked state.
```
minestarks:
```suggestion
/// iterations to a different number may undershoot or overshoot the marked state.
```

minestarks (approved):
Sounds clearer, thanks!

samples/algorithms/Grover.qs
DmitryVasilevsky: There's not much value in making this number higher - you'll use more computation for nothing - so I think it's not a good idea to suggest this. Besides, some higher numbers may bring you to the right solutions after going around the circle.
If you are curious, there's another comment on the `IterationsToMarked` function, which tells you what happens if the number of iterations is different. So I think I should leave this comment as is.

PR #2512, opened 2025-06-05
Consolidate OpenQASM project loading and include handling

idavis:
Replaces #2444
 
- **Unified project loading**: Consolidated OpenQASM project loading across all components (compiler, language service, Python pip, WASM) to use a single `load_openqasm_project` method instead of multiple inconsistent approaches
- **Improved error handling**: Enhanced include file error propagation by properly associating errors with their source spans and include statements in the parent file
- **Better error reporting**: Include file errors now display with proper source location information, showing the problematic include statement in context instead of generic "Not Found" messages
- **API simplification**:
  - Removed custom `ImportResolver` from Python interop in favor of unified project loading
  - Replaced multiple `compile_to_qsharp_ast_with_config` calls with `parse_and_compile_to_qsharp_ast_with_config`. With parsing, compilation, and project loading being separated, trying to clear up intent and purpose.
  - Added `parse_sources` utility function for handling multiple source files
- **Enhanced debugging support**: Added test coverage for setting breakpoints in multi-file OpenQASM programs with included files
- **Code structure improvements**:
  - Added `Debug` and `Clone` derives to semantic analysis types for better tooling support
  - Consolidated error handling logic into reusable helper functions
  - Improved span tracking for IO errors through the compilation pipeline

The changes ensure consistent OpenQASM project handling across VS Code extension, Python package, WASM builds, and language service while providing better error messages and debugging capabilities for multi-file projects.

### DIFF
--- a/compiler/qsc_project/src/error.rs
+++ b/compiler/qsc_project/src/error.rs
@@ -4,7 +4,6 @@
 use miette::Diagnostic;
 use thiserror::Error;
 
-#[cfg(feature = "fs")]
 #[derive(Error, Debug, Diagnostic)]
 pub enum StdFsError {
     #[error("found a qsharp.json file, but it was invalid: {0}")]
--- a/compiler/qsc_qasm/src/io.rs
+++ b/compiler/qsc_qasm/src/io.rs
@@ -146,13 +146,6 @@ pub struct Cycle {
     pub paths: Vec<Arc<str>>,
 }
 
-impl Cycle {
-    #[must_use]
-    pub fn new(paths: Vec<Arc<str>>) -> Self {
-        Self { paths }
-    }
-}
-
 impl std::fmt::Display for Cycle {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let parents = self.paths[0..(self.paths.len() - 1)].iter();
--- a/compiler/qsc_qasm/src/io.rs
+++ b/compiler/qsc_qasm/src/io.rs
@@ -17,7 +17,11 @@ use rustc_hash::FxHashMap;
 pub trait SourceResolver {
     fn ctx(&mut self) -> &mut SourceResolverContext;
 
-    fn resolve(&mut self, path: &Arc<str>) -> miette::Result<(Arc<str>, Arc<str>), Error>;
+    fn resolve(
+        &mut self,
+        path: &Arc<str>,
+        original_path: &Arc<str>,
+    ) -> miette::Result<(Arc<str>, Arc<str>), Error>;
 }
 
 pub struct IncludeGraphNode {
@@ -189,12 +193,16 @@ impl SourceResolver for InMemorySourceResolver {
         &mut self.ctx
     }
 
-    fn resolve(&mut self, path: &Arc<str>) -> miette::Result<(Arc<str>, Arc<str>), Error> {
+    fn resolve(
+        &mut self,
+        path: &Arc<str>,
+        original_path: &Arc<str>,
+    ) -> miette::Result<(Arc<str>, Arc<str>), Error> {
         match self.sources.get(path) {
             Some(source) => Ok((path.clone(), source.clone())),
             None => Err(Error(ErrorKind::NotFound(
                 Span::default(),
-                format!("Could not resolve include file: {path}"),
+                format!("Could not resolve include file: {original_path}"),
             ))),
         }
     }
--- a/compiler/qsc_qasm/src/io.rs
+++ b/compiler/qsc_qasm/src/io.rs
@@ -4,6 +4,7 @@
 mod error;
 pub use error::Error;
 pub use error::ErrorKind;
+use qsc_data_structures::span::Span;
 
 use std::sync::Arc;
 
@@ -33,11 +34,15 @@ pub struct SourceResolverContext {
 }
 
 impl SourceResolverContext {
-    pub fn check_include_errors(&mut self, path: &Arc<str>) -> miette::Result<(), Error> {
+    pub fn check_include_errors(
+        &mut self,
+        path: &Arc<str>,
+        span: Span,
+    ) -> miette::Result<(), Error> {
         // If the new path makes a cycle in the include graph, we return
         // an error showing the cycle to the user.
         if let Some(cycle) = self.cycle_made_by_including_path(path) {
-            return Err(Error(ErrorKind::CyclicInclude(cycle)));
+            return Err(Error(ErrorKind::CyclicInclude(span, cycle)));
         }
 
         // If the new path doesn't make a cycle but it was already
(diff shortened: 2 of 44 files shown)


compiler/qsc_qasm/src/parser.rs
```diff
@@ -219,18 +229,90 @@ impl QasmSource {
     }
 }
 
+fn strip_scheme(path: &str) -> (Option<Arc<str>>, Arc<str>) {
+    if let Some(scheme_end) = path.find("://") {
```
minestarks, line 233: Strictly speaking schemes can be separated by just `:` , e.g. `untitled:Untitled-1`

Sadly, there's no strict way to tell whether a string is a file path or URL, so I'm not really suggesting you try :P

But you may want to handle the below case more gracefully (for untitled files)?

```
2025-06-06 14:08:05.117 [warning] Failed to resolve path  and nonexistent.qasm: Error: [UriError]: Scheme is missing: {scheme: "", authority: "", path: "", query: "", fragment: ""}
```

compiler/qsc_qasm/src/io.rs
```diff
@@ -180,12 +197,12 @@ impl SourceResolver for InMemorySourceResolver {
...
     fn resolve(&mut self, path: &Arc<str>) -> miette::Result<(Arc<str>, Arc<str>), Error> {
-        self.ctx().check_include_errors(path)?;
         match self.sources.get(path) {
             Some(source) => Ok((path.clone(), source.clone())),
-            None => Err(Error(ErrorKind::NotFound(format!(
-                "Could not resolve include file: {path}"
-            )))),
+            None => Err(Error(ErrorKind::NotFound(
+                Span::default(),
+                format!("Could not resolve include file: {path}"),
```
minestarks: In VS Code we get the full URL here. (e.g. `file:///c%3A/src/qsharp/compiler/qsc_project/src/tests/openqasm_projects/nonexistent.qasm`) . I wonder if we should pass in the original (relative) path to this function so the error looks nicer?

compiler/qsc_qasm/src/io.rs
```diff
@@ -133,7 +143,14 @@ impl SourceResolverContext {
 /// We use this struct to print a nice error message when we find a cycle.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Cycle {
-    paths: Vec<Arc<str>>,
+    pub paths: Vec<Arc<str>>,
+}
+
+impl Cycle {
+    #[must_use]
+    pub fn new(paths: Vec<Arc<str>>) -> Self {
```
minestarks: This is unused since you also made `paths` public and used that in the initialization.

compiler/qsc_qasm/src/io.rs
minestarks: Same goes for the Cyclic Include error:

```
Cyclic include:
  file:///c%3A/src/qsharp/compiler/qsc_project/src/tests/openqasm_projects/circular_b.qasm includes file:///c%3A/src/qsharp/compiler/qsc_project/src/tests/openqasm_projects/circular_a.qasm
  file:///c%3A/src/qsharp/compiler/qsc_project/src/tests/openqasm_projects/circular_a.qasm includes file:///c%3A/src/qsharp/compiler/qsc_project/src/tests/openqasm_projects/circular_b.qasm
```

I think the literal include paths would suffice.

vscode/test/suites/debugger/openqasm.test.ts
```diff
@@ -14,11 +14,19 @@ suite("OpenQASM Debugger Tests", function suite() {
...
   const selfContainedName = "self-contained.qasm";
+  const multifileName = "multifile.qasm";
+  const multifileIncludeName = "imports.inc";
 
   const selfContainedUri = vscode.Uri.joinPath(
     workspaceFolder.uri,
     selfContainedName,
   );
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const multifileUri = vscode.Uri.joinPath(workspaceFolder.uri, multifileName);
```
minestarks: Any reason you need this const?

compiler/qsc_project/src/error.rs
```diff
@@ -4,7 +4,6 @@
 use miette::Diagnostic;
 use thiserror::Error;
```
minestarks, line 6: I think you can revert this?

minestarks: Oh, I see the cfg moved to lib.rs. Is that better somehow?

minestarks (commented):
Posting a round of comments. Still need to review language service and pip folders

language_service/src/compilation.rs
```diff
@@ -256,24 +259,21 @@ impl Compilation {
...
 
+        let config = CompilerConfig::new(
+            QubitSemantics::Qiskit,
+            OutputSemantics::OpenQasm,
+            ProgramType::File,
+            Some("program".into()),
+            None,
+        );
+        let res = qsc::qasm::semantic::parse_sources(&sources);
+        let unit = compile_to_qsharp_ast_with_config(res.clone(), config);
```
minestarks: Is `.clone()` necessary?

language_service/src/compilation.rs
```diff
@@ -256,24 +259,21 @@ impl Compilation {
...
         let CompileRawQasmResult(store, source_package_id, dependencies, _sig, mut compile_errors) =
-            qsc::qasm::compile_raw_qasm(
-                source.clone(),
-                path.clone(),
-                Some(&mut resolver),
-                package_type,
-                capabilities,
-            );
+            qsc::qasm::compile_openqasm(unit.clone(), package_type, capabilities);
```
minestarks: Is `.clone()` necessary?

language_service/src/compilation.rs
```diff
@@ -421,12 +421,12 @@ impl Compilation {
...
-                ref sources,
+                sources: ref source,
                 ref friendly_name,
             } => Self::new_qasm(
                 package_type,
                 target_profile,
-                sources.clone(),
+                source.clone(),
                 Vec::new(), // project errors will stay the same
                 friendly_name,
```
minestarks: Maybe revert? This looks more confusing than the original?

language_service/src/completion/tests/openqasm.rs
minestarks, line 1: These test changes seem unnecessary (and a bit scary)?

compiler/qsc_project/src/error.rs
idavis: It is simpler and follows other patterns like including test files only when test is defined.

language_service/src/compilation.rs
idavis: Not any more

language_service/src/compilation.rs
idavis: Same as above, left over from different impl. Nice catch.

language_service/src/compilation.rs
idavis: Yeah, leftover from type changes and then going back to sources.

compiler/qsc_qasm/src/io.rs
idavis: We don't have the original include path. We could change the trait fn resolve to take an extra param. Same with `check_include_errors`, we would have to add the extra parameter.

minestarks (approved):

PR #2516, opened 2025-06-06
Add EC trait function to adjust code parameter after selection

msoeken:
This adds an EC trait function to provide the possibility to adjust the code parameter after it has been computed. This is required for special EC functionality through the extensibility API. The default implementation of the trait function returns the original code parameter and the EC implementation inside the QDK does not change that function.

### DIFF
--- a/resource_estimator/src/estimates/error_correction.rs
+++ b/resource_estimator/src/estimates/error_correction.rs
@@ -156,6 +156,21 @@ pub trait ErrorCorrection {
             .ok_or_else(|| format!("No code parameter achieves required logical error rate {required_logical_error_rate:.3e}"))
     }
 
+    /// Adjusts code parameter after computing it
+    ///
+    /// This function is called after the initial code parameter has been
+    /// computed.  The initial code parameter will fit the required logical
+    /// error rate, but it's possible to further refine it, if necessary.
+    ///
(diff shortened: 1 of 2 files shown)


resource_estimator/src/estimates/error_correction.rs
```diff
@@ -156,6 +156,21 @@ pub trait ErrorCorrection {
             .ok_or_else(|| format!("No code parameter achieves required logical error rate {required_logical_error_rate:.3e}"))
     }
 
+    /// Adjusts code parameter after computing it
+    ///
+    /// This function is called after the initial code parameter has been
+    /// computed.  The initial code parameter will fit the required logical
+    /// error rate, but it's possible to further refine it, if necessary.
+    ///
+    /// Note, that when implementing this function wrong, it may return a code
```
DmitryVasilevsky: When I first looked at this, I thought it was an API for clients to call. Maybe it's worth mentioning that this is for implementers of this (ErrorCorrection) trait explicitly. Something like 'when implementing ErrorCorrection trait and providing custom implementation of this function..."

DmitryVasilevsky (approved):

resource_estimator/src/estimates/error_correction.rs
msoeken: Thanks, addressed in last commit

PR #2519, opened 2025-06-10
Generic resource estimation using Python models

msoeken:
This adds a new Python function `estimate_generic` to the Q# Python package. It performs resource estimation on generic input models for logical application and layout overhead, quantum error correction, qubits, and magic state factories. Application, QEC, and factory overhead are modeled by means of Python classes, where as qubits are Python dictionaries to hold their properties.

### DIFF
--- a/pip/qsharp/__init__.py
+++ b/pip/qsharp/__init__.py
@@ -23,7 +23,7 @@ from ._qsharp import (
 
 telemetry_events.on_import()
 
-from ._native import Result, Pauli, QSharpError, TargetProfile
+from ._native import Result, Pauli, QSharpError, TargetProfile, estimate_generic
 
 # IPython notebook specific features
 try:
@@ -47,6 +47,7 @@ __all__ = [
     "compile",
     "circuit",
     "estimate",
+    "estimate_generic",
     "Result",
     "Pauli",
     "QSharpError",
--- a/pip/src/generic_estimator.rs
+++ b/pip/src/generic_estimator.rs
@@ -1,0 +1,134 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::{rc::Rc, time::Instant};
+
+use pyo3::{prelude::*, types::PyDict};
+use resource_estimator::estimates::{ErrorBudget, ErrorBudgetStrategy, PhysicalResourceEstimation};
+use serde_json::json;
+
+use crate::generic_estimator::{
+    code::PythonQEC,
+    counts::PythonCounts,
+    factory::{PythonFactoryBuilder, PythonFactoryBuilderDispatch},
+    utils::json_map_to_python_dict,
+};
+
+mod code;
+mod counts;
+mod factory;
+mod utils;
+
+#[cfg(test)]
+mod tests;
+
+pub(crate) fn register_generic_estimator_submodule(m: &Bound<'_, PyModule>) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(estimate_generic, m)?)?;
+    Ok(())
+}
+
+/// Estimates quantum resources for a given algorithm, qubit, and code.
+///
+/// Args:
+///     algorithm: Python object representing the algorithm.
+///     qubit: The qubit properties as a dictionary.
+///     qec: Python object representing the quantum error correction code.
+///     factories (list, optional): List of python objects representing factories. Default: [].
+///     error_budget (float, optional): The total error budget, which is uniformly distributed. Default: 0.01.
+///     max_factories (int, optional): Constrains the number of factories. Default: None.
+///     logical_depth_factor (float, optional): Extends algorithmic logical depth by a factor >= 1. Default: None.
+///     max_physical_qubits (int, optional): Forces estimator to not exceed provided number of physical qubits, may fail. Default: None.
+///     max_duration (int, optional): Allows estimator to run for given runtime in nanoseconds, may fail. Default: None.
+///     error_budget_pruning (bool, optional): Will try to prune the error budget to increase magic state error budget. Default: False.
+///
+/// Returns:
+///     dict: A dictionary with resource estimation results.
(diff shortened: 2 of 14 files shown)


pip/src/generic_estimator.rs
```diff
@@ -0,0 +1,134 @@
...
+///     max_physical_qubits (int, optional): Forces estimator to not exceed provided number of physical qubits, may fail. Default: None.
+///     max_duration (int, optional): Allows estimator to run for given runtime in nanoseconds, may fail. Default: None.
+///     error_budget_pruning (bool, optional): Will try to prune the error budget to increase magic state error budget. Default: False.
+///
+/// Returns:
+///     dict: A dictionary with resource estimation results.
+#[allow(clippy::too_many_arguments)]
+#[pyfunction]
+#[pyo3(signature = (algorithm, qubit, qec, factories = vec![], *, error_budget = 0.01, max_factories = None, logical_depth_factor = None, max_physical_qubits = None, max_duration = None, error_budget_pruning = false))]
+fn estimate_generic<'py>(
```
swernli: I've started reviewing the code, but one sort of "top-level" comment is about the naming (which is always so hard to do for APIs). While conceptually I see how this is a more generic version of resource estimation, `qsharp.estimate_generic` seems like an odd name to me. Could it be something like `estimate_custom` or `estimate_with_algorithm`?

pip/qsharp/__init__.py
```diff
@@ -23,7 +23,7 @@
 
 telemetry_events.on_import()
 
-from ._native import Result, Pauli, QSharpError, TargetProfile
+from ._native import Result, Pauli, QSharpError, TargetProfile, estimate_generic
```
swernli: Looks like the .pyi entry for the new python API is missing. We use tht as the place to put the doc-comments and type signatures for the Python objects and functions we expose. Is that intentionally ommitted because the function is not meant to be documented yet?

msoeken: I added the comments to the Rust file and used pyo3 to generate the signature. The Rust docstring is automatically be used as docstring for the Python function. But if it is preferred, I can move the documentation and signature to the `pyi` file. But note that `estimate_custom` is not a function of the `_native` submodule, but directly implemented in the `qsharp` topmodule.

pip/src/generic_estimator.rs
msoeken: Thanks for the suggestions. I used `estimate_custom`, because I find `estimate_with_algorithm` a bit misleading, as we estimate with custom models not only for the application input, but also for the qubits, QEC codes, and magic state factories.

An alternative is to overload the `estimate` function and check based on the types which function to call. But I am afraid that this makes it difficult to properly document the function and render a comprehensible signature.

pip/qsharp/__init__.py
swernli: That's a good point... it doesn't really show up in the same place as the others. That said, I'd like to try the package locally and see how this works with goto-def in Python. One advantage of the .pyi is it gives the editor something to jump to. I'm not sure if pyO3 generates a stub of some sort to fill this role.

pip/src/generic_estimator.rs
swernli: Yeah, agreed, overloading the existing estimate would get messy fast. I can't really think of better options, so perhaps `estimate_generic` is descriptive enough.

msoeken: I changed it to `estimate_custom`. Do you prefer to change it back to `estimate_generic`?

swernli: I personally like `estimate_custom` a little more, thanks for changing that. As always, naming is hard and different names resonate with different people!

pip/qsharp/__init__.py
swernli: I finally got the chance to try this, and unfortunately it does run into the problems I feared: there is no help or guidance in the editor from Python extensions. By way of comparison, this is how the existing `estimate` shows up:
image
Note the icon is different from that of the `estimate_custom` function, because of how they are detected by the extension. This means that while `estimate` gets parameter help and goto-def:
image
none of the comments or help text for `estimate_custom` show up:
image
and even when invoked properly it is highlighted differently than other function calls.

I think what I'd reccommend here to give the function the same user friendly features as the existing `estimate` is to expose it the same way (a wrapper function in interpreter.rs with a `pyO3` signature and matching interface in _native.pyi). It's a couple extra layers, but the usability wins seem worth it.

pip/qsharp/__init__.py
```diff
@@ -47,6 +47,7 @@
     "compile",
     "circuit",
     "estimate",
+    "estimate_common",
```
swernli:
```suggestion
    "estimate_custom",
```

pip/qsharp/__init__.py
msoeken: Thanks for this insight. I also double checked now, and also the "go to type definition / implementation" command now jumps to the installed pyi file.

I will wrap this in the `_native.pyi` file, but I don't add it to `interpreter.rs`. The functions in there seem to be relying on an interpreter (also in `_qsharp.py` all functions defined in there call `get_interpreter()`). I think that `estimate_custom` is to `physical_estimates` or `resource_estimate_qasm_program` which is directly declared in `_native.pyi`. I noticed that some of these functions have both the documentation inside the Rust file and in the `_native.pyi` file, and some don't. I prefer to only have the documentation once in order to avoid duplicate code that may lead to inconsistent doc strings. So, I only added the doc string to the pyi file.

swernli: Great, thanks! I can confirm I see the doc comments and goto def behavior on my end too:
image

swernli (approved):
Looks good! Of note, #2523 moved a bunch of folders as part of a effort to restructure the top level of the repo and make it more friendly to users. That's going to be a big merge conflict but hopefully not too bad since the files just moved but were not otherwise significantly changed. Once you've got the rebase on main done with that you should be good to merge.

PR #2521, opened 2025-06-10
Implement builtin functions

orpuente-MS:
Add builtin functions to OpenQASM.

Note: This PR is a follow up of #2508, which added the necessary infrastructure to support builtin functions.

### DIFF
--- a/compiler/qsc_qasm/src/semantic/lowerer.rs
+++ b/compiler/qsc_qasm/src/semantic/lowerer.rs
@@ -36,6 +36,7 @@ use crate::semantic::types::can_cast_literal;
 use crate::semantic::types::can_cast_literal_with_value_knowledge;
 use crate::stdlib::angle::Angle;
 use crate::stdlib::builtin_functions;
+use crate::stdlib::complex::Complex;
 
 use super::ast as semantic;
 use crate::parser::ast as syntax;
@@ -807,7 +808,7 @@ impl Lowerer {
                 Type::Float(None, true),
             ),
             syntax::LiteralKind::Imaginary(value) => (
-                semantic::ExprKind::Lit(semantic::LiteralKind::Complex(0.0, *value)),
+                semantic::ExprKind::Lit(Complex::imag(*value).into()),
                 Type::Complex(None, true),
             ),
             syntax::LiteralKind::String(_) => {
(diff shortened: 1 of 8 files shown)


compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -1668,12 +1669,22 @@ impl Lowerer {
             .collect();
 
         let output = match name {
```
idavis: Does it make sense to have an enum with a from_str implementation? Then you can pattern match and do exhaustive pattern detection.

compiler/qsc_qasm/src/semantic/tests/expression/builtin_functions.rs
```diff
@@ -244,99 +244,537 @@ fn nested_builtin_call_succeeds() {
     );
 }
 
+// ----------------------------------
```
idavis: Does it make sense to break these into a test file per function?

compiler/qsc_qasm/src/stdlib/complex.rs
```diff
@@ -0,0 +1,151 @@
...
+use crate::semantic::ast::LiteralKind;
+
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Complex {
+    pub real: f64,
+    pub imag: f64,
+}
+
+impl Complex {
+    pub fn new(real_value: f64, imaginary_value: f64) -> Self {
```
idavis: can these be named `real` and `imag`?

orpuente-MS: Yes, that makes sense. Changed.

compiler/qsc_qasm/src/semantic/tests/expression/builtin_functions.rs
orpuente-MS: Yes, it totally makes sense considering this file has 2k lines. Let me change that.

compiler/qsc_qasm/src/semantic/lowerer.rs
orpuente-MS: Actually, yes. And it would save me the `is_builtin_function` predicate as well, reducing the number of string matches I would need to do by half. Nice catch!

compiler/qsc_qasm/src/stdlib/builtin_functions.rs
```diff
@@ -185,14 +242,303 @@ pub fn int() -> Type {
...
+        Type::Angle(w, _)
+        | Type::Complex(w, _)
+        | Type::Float(w, _)
+        | Type::Int(w, _)
+        | Type::UInt(w, _) => *w,
+        Type::BitArray(w, _) => Some(*w),
+        _ => None,
+    }
+}
+
```
idavis, line 346: can't we call ty.width() instead?

compiler/qsc_qasm/src/stdlib/builtin_functions.rs
```diff
@@ -185,14 +242,303 @@ pub fn int() -> Type {
...
+// arccos
+
+pub(crate) fn arccos(
+    inputs: &[Expr],
+    name_span: Span,
+    call_span: Span,
+    ctx: &mut Lowerer,
+) -> PolymorphicFunctionOutput {
+    let fn_table: FnTable = vec![(fun(&[float()], float()), Box::new(arccos_float))];
+    dispatch("arccos", name_span, call_span, inputs, fn_table, ctx)
```
idavis: Should we use the new enum here instead of the string names?

idavis (approved):

PR #2527, opened 2025-06-12
Adding support for QASM 2.0 in the OpenQASM compiler

idavis:
This PR adds promotion of OpenQASM 2.0 programs into OpenQASM 3.0 with some caveats
- Global phase is not adjusted so any OpenQASM 2.0 programs will have OpenQASM 3.0 global phase
- Opaque gates are not parsed. Any OpenQASM 2.0 programs with them will have parser errors. This concept was not migrated to OpenQASM 3.0
- OpenQASM 2.0 programs can import `"qelib1.inc"` and trying to import `"stdgates.inc"` will give an error.


orpuente-MS (approved):
Looks good. Thanks!

PR #2529, opened 2025-06-16
Remove implicit modifier handling during OpenQASM lowering

idavis:
In the original implementation of the OpenQASM compiler, gates with implicit modifiers needed to be specially handled by redefining them as their unmodiefied variants and collecting the modifier. For example, `ch` would be converted to `h` with a `ctrl(1)` modifier and would be used to build up the gate call with the additional explicit modifiers.

With the introduction of the `Std.OpenQASM` library, we can now call those gates directly without needing to decompose them. This also cleans up error messages as any errors would have been associated with the unmodified gate name prior.

### DIFF
--- a/library/std/src/Std/OpenQASM/Intrinsic.qs
+++ b/library/std/src/Std/OpenQASM/Intrinsic.qs
@@ -105,7 +105,7 @@ operation U(theta : Angle, phi : Angle, lambda : Angle, qubit : Qubit) : Unit is
     controlled adjoint auto;
 }
 operation CX(ctrl : Qubit, qubit : Qubit) : Unit is Adj + Ctl {
-    Std.Intrinsic.CNOT(ctrl, qubit);
+    Std.Canon.CX(ctrl, qubit);
 }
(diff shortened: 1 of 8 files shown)


orpuente-MS (approved):

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -146,10 +167,37 @@ operation cx(ctrl : Qubit, qubit : Qubit) : Unit is Adj + Ctl {
     Std.Intrinsic.CNOT(ctrl, qubit);
 }
 
+operation cy(ctrl : Qubit, qubit : Qubit) : Unit is Adj + Ctl {
+    Controlled Std.Intrinsic.Y([ctrl], qubit);
```
DmitryVasilevsky: I'm wondering if CY and CZ from Canon namespace is any better. @swernli , what do you think?

DmitryVasilevsky (approved):

PR #2531, opened 2025-06-16
Added div class table-with-latex for official web site

DmitryVasilevsky:
As requested, added the `div` class `table-with-latex` around the two tables to aid proper rendering.

### DIFF
--- a/katas/content/single_qubit_gates/index.md
+++ b/katas/content/single_qubit_gates/index.md
@@ -394,6 +394,7 @@ The identity gate is mostly here for completeness, at least for now. It will com
 The **Hadamard** gate is an extremely important quantum gate. Unlike the previous gates, applying the Hadamard gate to a qubit in a computational basis state puts that qubit into a superposition.
 Like the Pauli gates, the Hadamard gate is self-adjoint.
 
+<div class="table-with-latex">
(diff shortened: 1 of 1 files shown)


katas/content/single_qubit_gates/index.md
```diff
@@ -394,6 +394,7 @@ The identity gate is mostly here for completeness, at least for now. It will com
 The **Hadamard** gate is an extremely important quantum gate. Unlike the previous gates, applying the Hadamard gate to a qubit in a computational basis state puts that qubit into a superposition.
 Like the Pauli gates, the Hadamard gate is self-adjoint.
 
+<div class="table-with-latex">
```
ScottCarda-MS, line 405: nit: I'm unsure why this, as well as the table above, don't have proper indentation for their inner elements. Not a big deal, but would help with readability of the raw file. Feel free to fix if you want, or dismiss this comment if you don't want to. Just wanted to point it out.

DmitryVasilevsky: Yes, I was also wondering! But since there're two of us now... I formatted this table and another one. I also added this class to other tables in this kata - result of discussion with web site maintainers.

PR #2539, opened 2025-06-17
Update workflow paths

ScottCarda-MS:
Updates the paths in the .github/workflows files to respect the new repo structure.


ScottCarda-MS:
The deployment error listed above is from when workflows were being manually run on this branch. The Playground deployment failed because only main is allowed to deploy the Playground, so this is expected.

swernli (approved):

PR #2557, opened 2025-06-18
Add new qasm AST node for dynamic arrays

orpuente-MS:


source/compiler/qsc_qasm/src/parser/ast.rs
```diff
@@ -1020,19 +1020,61 @@ impl Display for ArrayType {
 }
 
 #[derive(Clone, Debug)]
-pub struct ArrayReferenceType {
+pub enum ArrayReferenceType {
+    Static(StaticArrayReferenceType),
+    Dyn(DynArrayReferenceType),
+}
```
orpuente-MS, line 1026: This is an enum instead of factoring the span, mutability, and ident into a struct so that we can nicely send it down different lowering paths where all this information is necessary.

ScottCarda-MS (approved):
Thanks for walking me through the PR.

PR #2562, opened 2025-06-23
J

Metatron3699:


Metatron3699:
> The build actions already handle the Rust components, so a new yaml is not needed and redundant.

You are Redundant. @minestarks will handle it ~~~~

PR #2564, opened 2025-06-24
Classical types indexing in OpenQASM

orpuente-MS:
This PR adds classical type indexing support to our OpenQASM compiler and fixes two bugs.

## Highlights
The major challenge of this PR was supporting classical type indexing in the left-hand side of assignment expressions. The idea to make it work is compiling
```
int[4] a = 0b1111;
a[1:][1:] = "01";
```
to this
```
mutable a = 15;
a = {
    mutable bitarray = Std.OpenQASM.Convert.IntAsResultArrayBE(a, 4);
    bitarray[1...][1...] = [Zero, One];
    Std.OpenQASM.Convert.ResultArrayAsIntBE(bitarray)
};
```
so that we can leverage Q#'s array semantics.
 
The indexing infrastructure is now simpler. Instead of having different codepaths for indexed identifiers and indexed expressions, now we just have a codepath for indexed expressions. (The codepath for indexed identifiers was redundant).

Multiple indexing is now expressed as nested IndexedExpr structs with a single index each, instead of a single `IndexedExpr` with a list of indices. This makes thinking about indexing simpler, since now the utility functions to handle indexing just need to handle a single index. It also makes compilation simpler, since this is how the Q# AST is structured.

## Bug fixes:
 1. Consecutive slices into an array now act on the same dimension, instead of acting on consecutive dimensions.
 2. `AngleAsResultArrayBE` is now correctly exported in the `Angle.qs` file instead of `AngleAsResultArray`.


source/compiler/qsc_qasm/src/semantic/const_eval.rs
```diff
@@ -132,21 +131,8 @@ impl LiteralKind {
...
-            // A bit array accepts a single index.
-            if indices.len() != 1 {
-                ctx.push_const_eval_error(ConstEvalError::TooManyIndices(collection_span));
-                return None;
-            }
-
-            let index = indices.first().expect("there is exactly one index");
-
             Self::index_bitarray(ctx, value, size, index)
         } else {
```
idavis, line 137: In what case can the bitarray take multiple indices?

orpuente-MS: You can apply multiple consecutive slices to a bitarray
```
const bit[4] a = "1010";
a[1:][1:]
```

source/compiler/qsc_qasm/src/semantic/const_eval.rs
```diff
@@ -205,21 +191,10 @@ impl LiteralKind {
...
-impl IndexExpr {
-    #[allow(clippy::unused_self)]
+impl IndexedExpr {
     fn const_eval(&self, ctx: &mut Lowerer, _ty: &Type) -> Option<LiteralKind> {
         let value = self.collection.const_eval(ctx)?;
-        let indices: Vec<_> = self.indices.iter().map(|idx| (**idx).clone()).collect();
-        value.index_array(ctx, &indices, self.collection.span)
+        value.index_array(ctx, &self.index, self.collection.span)
     }
 }
```
idavis, line 199: I don't see any tests for const eval of indexed sized types and I don't think they'll currently work. Can you please add tests to verify?

orpuente-MS: The tests are in tests/expression/indexed.rs and tests/statement/assignment.rs

PR #2566, opened 2025-06-26
QIR Profiles Selection via Project Field

ScottCarda-MS:
This removes the workspace setting for the QIR profile and adds a setting for this for Q# project files instead. Single-file Q# and QASM files are handled by assuming the profile based off of the target, or "base" profile when their isn't a target.


ScottCarda-MS:
Closed in favor of #2591

PR #2567, opened 2025-06-26
Support for qubit loss

### LINKED ISSUE #2483: Library and/or guidance for dealing with atom loss
Some modalities (e.g. neutral atom) can report atom loss from a measurement. We should provide a library, or if not guidance, on how users could write code for this.

### LINKED ISSUE #2484: Simulation of atom loss
Related #2483

Some modalities encounter qubit loss (e.g. atom loss in neutral atom hardware). We should provide tooling and/or guidance on how to simulate this (be it in our default simulator, or if wiring up another 'noisy simulator').

swernli:
This PR adds support for qubit loss to the QDK. In particular, it adds the following features:

- Simulator with configured qubit loss will probabilistically "lose" qubits during application of noise (ie: any gates, measurements, or calls to `ApplyIdleNoise`), treating futher operations on that qubit as no-op.
- Qubits that are lost return neither `Zero` nor `One` during measurement but a special `Loss` value that cannot be used in comparisons, triggering a runtime failure if it is.
- `Std.Diagnostics.ConfigureQubitLoss` to configure the probability of qubit loss on any operation.
- `Std.Measurement.MResetZChecked` which returns a tuple of `(Result, Bool)` by checking for qubit loss as part of measurement.
- Loss checks generate into QIR by emitting the `__quantum__rt__read_loss` runtime function (may not be supported on all QIR targets).

Additional features:
 - [x] Extension setting alongside Pauli noise for configuring loss probability during histogram generation
 - [x] OpenQASM support for special measurement that returns loss bit, exposed via new "qdk.inc" special include
 - [x] Python API for configuring loss probability on calls to `qsharp.run`
 - [x] Samples for configuring and detecting loss

 Fixes #2484, fixes #2483


library/std/src/Std/Measurement.qs
```diff
@@ -160,5 +161,27 @@ operation MeasureInteger(target : Qubit[]) : Int {
...
+/// A single qubit to be measured.
+///
+/// # Output
+/// A tuple containing the measurement result and a Boolean `true` if the result indicates a loss, otherwise `false`.
+///
+/// # Remarks
+/// This operation is useful for detecting qubit loss during execution. During simulation, qubit loss probability can be
+/// configured via the `ConfigureQubitLoss` operation. When compiled to QIR, this uses the `__quantum__rt__read_loss` intrinsic,
+/// which may not be supported on all hardware targets and could result in compilation errors when submitting to those targets.
+operation MResetZChecked(target : Qubit) : (Result, Bool) {
```
DmitryVasilevsky, line 186: I do not agree with the separate measurement function for getting qubit loss. I strongly believe that a function that checks the result for loss is more appropriate for this. With a separate function for a check:
1. Fewer code changes required to switch from one to another - measurements and result checks are separated, and only one Result is passed between them.
2. You can delegate measurements in the same manner as before. For example, you'll still be able to use MResetEachZ and all other functions that handle results.
3. It's a natural place to add handling code. Even if you forget to check, and you get an error, you'll add more handling where the result is already (miss)handled.
4. This separate function gives name to this result. Returning a Boolean value in a tuple doesn't give it a name and is therefore requires documentation lookup.
5. QIR function __quantum__rt__read_loss is already a separate function, and Result already has this information because it needs to be propagated out of a shot.

DmitryVasilevsky: I don't have a strong opinion on the actual name, two approaches seem reasonable. One - specific to the loss. Something like IsQubitLost(r). Another one is more abstract, something like IsResultAvailable(r). With the second one, other potential errors on measurements can be folded in. (Do not take these specific names as suggestions.)

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -1776,6 +1782,11 @@ fn eval_binop_eq(lhs_val: Value, rhs_val: Value, rhs_span: PackageSpan) -> Resul
             // to prevent executing programs that do result comparisons.
             Err(Error::ResultComparisonUnsupported(rhs_span))
         }
+        (Value::Result(val::Result::Loss), _) | (_, Value::Result(val::Result::Loss)) => {
```
DmitryVasilevsky, line 1797: So basically loss != loss (and causes a runtime error). Similar to a NAN handling. l like this.

source/compiler/qsc_eval/src/intrinsic/tests.rs
```diff
@@ -1910,6 +1910,871 @@ fn check_pauli_noise() {
...
+}
+
+#[test]
+fn comparison_of_loss_result_runtime_error() {
+    check_intrinsic_result(
+        "",
+        indoc! {"{
+            use q = Qubit();
+            Std.Diagnostics.ConfigureQubitLoss(1.0);
+            MResetZ(q) == One
```
DmitryVasilevsky, line 1985: Can you add another test that compares loss with itself?

swernli: That's a good idea, I'll include a test to confirm that behavior!

library/std/src/Std/Measurement.qs
swernli: I've kept the `MResetZChecked` operation but now also exposed an `IsLossResult` to make it easier to detect loss from existing measurement functions.

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -633,6 +636,11 @@ operation ccz(ctrl1 : Qubit, ctrl2 : Qubit, target : Qubit) : Unit is Adj + Ctl
     h(target);
 }
 
+operation mresetzchecked(q : Qubit) : Result[] {
```
swernli: I'm not a fan of `mresetzchecked` as the name here, but I'm not familiar enough with the naming conventions in OpenQASM to suggest an alternative. @idavis and @orpuente-MS, any suggestions? Something like `mresetz_checked` or other?

idavis: I don't think there is a real naming standard for defs

orpuente-MS: `mresetz_checked` is easier to read. I would go with that instead. About the naming standard, they use both snake_case and camelCase in the official spec. So, I would just pick one (snake case is my personal preference).

source/compiler/qsc_eval/src/val.rs
```diff
@@ -60,7 +61,7 @@ impl Result {
     pub fn unwrap_bool(self) -> bool {
         match self {
             Self::Val(v) => v,
-            Self::Id(_) => panic!("cannot unwrap Result::Id as bool"),
+            Self::Id(_) | Self::Loss => panic!("cannot unwrap Result as bool"),
```
orpuente-MS: Should we have this as two different panics so that we know if we were trying to unwrap a `Result::Id` or a `Result::Loss` if we ever hit this? Same for `unwrap_id` a few lines below.

swernli: Sure, that makes sense.

source/compiler/qsc_eval/src/backend.rs
```diff
@@ -178,54 +214,103 @@ impl SparseSim {
...
     fn ccx(&mut self, ctl0: usize, ctl1: usize, q: usize) {
-        self.sim.mcx(&[ctl0, ctl1], q);
+        match (
+            self.is_qubit_lost(ctl0),
+            self.is_qubit_lost(ctl1),
+            self.is_qubit_lost(q),
+        ) {
+            (true, true, _) | (_, _, true) => {
+                // If the target qubit is lost or both controls are lost, skip the operation,
+                // but apply
```
orpuente-MS: I think this comment is incomplete.

swernli: Whoops! Good catch.

orpuente-MS (approved):
Looks good. Thanks!

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -1,3 +1,4 @@
+import Std.OpenQASM.Convert.ResultAsInt;
```
idavis: Remove this (bad location) and fully qualify the usage.

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -47,6 +50,7 @@ import Std.OpenQASM.Angle.AddAngles;
 import Std.OpenQASM.Angle.SubtractAngles;
 import Std.OpenQASM.Angle.DivideAngleByInt;
 import Std.OpenQASM.Angle.NegAngle;
+import Std.Convert.BoolAsResult;
```
idavis: Unused, remove please

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -633,6 +637,18 @@ operation ccz(ctrl1 : Qubit, ctrl2 : Qubit, target : Qubit) : Unit is Adj + Ctl
     h(target);
 }
 
+/// A resetting measurement operation that checks for qubit loss.
+/// Returns 0 if the qubit measurement was `Zero`, 1 if it was `One`,
+/// and 2 if the measurement indicated qubit loss.
+operation mresetz_checked(q : Qubit) : Int {
+    let (r, b) = MResetZChecked(q);
```
idavis: Fully qualify usage

library/std/src/Std/OpenQASM/Intrinsic.qs
```diff
@@ -633,6 +637,18 @@ operation ccz(ctrl1 : Qubit, ctrl2 : Qubit, target : Qubit) : Unit is Adj + Ctl
...
+    if b {
+        2
+    } else {
+        ResultAsInt(r)
```
idavis: fully qualify

library/core/qir.qs
```diff
@@ -27,5 +27,9 @@ namespace QIR.Runtime {
         }
     }
 
-    export __quantum__rt__qubit_allocate, __quantum__rt__qubit_release, AllocateQubitArray, ReleaseQubitArray;
+    function __quantum__rt__read_loss(r : Result) : Bool {
+        body intrinsic;
+    }
```
idavis, line 32: Feels like we should document std functions

swernli: These are techinically in "core" and not "std" and we don't really expect anyone to call these manually. That said, we don't supress them from suggestions and tab completions, so it feels like we are in a halfway state. For now, I'm just matching the pattern we have, which is to skip documenting these kind of functions.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -161,6 +161,11 @@ pub enum Error {
     #[diagnostic(help("comparing measurement results is not supported when performing circuit synthesis or base profile QIR generation"))]
     ResultComparisonUnsupported(#[label("cannot compare to result")] PackageSpan),
 
+    #[error("cannot compare measurement result from qubit loss")]
+    #[diagnostic(code("Qsc.Eval.ResultLossComparisonUnsupported"))]
+    #[diagnostic(help("use of a measurement result from a qubit that was lost is not supported, use `CheckLoss` to check for loss instead"))]
+    ResultLossComparisonUnsupported(#[label("cannot compare result from qubit loss")] PackageSpan),
```
idavis, line 177:
```suggestion
    ResultLossComparisonUnsupported(#[label] PackageSpan),
```

source/compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -191,23 +191,25 @@ pub(crate) fn call(
...
-        )),
+                .0,
+        ) {
+            Some(r) => Ok(Value::Result(r.into())),
+            None => Ok(Value::Result(val::Result::Loss)),
+        },
+        "__quantum__rt__read_loss" => Ok(Value::Bool(arg == Value::Result(val::Result::Loss))),
         _ => {
             let qubits = arg.qubits();
             let qubits_len = qubits.len();
```
idavis: can we refctor these into fuctions for clarity?

source/compiler/qsc_eval/src/intrinsic.rs
```diff
@@ -191,23 +191,25 @@ pub(crate) fn call(
...
-                    .ok_or(Error::QubitUsedAfterRelease(arg_span))?
-                    .0,
-            )
-            .into(),
-        )),
+                .0,
+        ) {
+            Some(r) => Ok(Value::Result(r.into())),
+            None => Ok(Value::Result(val::Result::Loss)),
+        },
```
idavis: does it make sense for ResultType to be `impl Into<Result>`?

source/pip/tests/test_qsharp.py
```diff
@@ -306,6 +306,17 @@ def test_run_with_noise_produces_noisy_results() -> None:
...
+    qsharp.init()
+    qsharp.set_quantum_seed(0)
+    result = qsharp.run(
+        "{ use q = Qubit(); X(q); MResetZ(q) }",
+        shots=1,
+        qubit_loss=1.0,
+    )
+    assert result[0] == qsharp.Result.Loss
+
+
```
idavis, line 319: Does it make sense to add a test for comparing a lost qubit result to make sure we surface a nice error in Python?

source/pip/src/interpreter.rs
```diff
@@ -1084,10 +1096,11 @@ impl<'py> IntoPyObject<'py> for ValueWrapper {
             Value::Double(val) => val.into_bound_py_any(py),
             Value::Bool(val) => val.into_bound_py_any(py),
             Value::String(val) => val.into_bound_py_any(py),
-            Value::Result(val) => if val.unwrap_bool() {
-                Result::One
-            } else {
-                Result::Zero
+            Value::Result(val) => match val {
+                qsc::interpret::Result::Id(_) => panic!("unexpected Result::Id in ValueWrapper"),
```
idavis, line 1098: Is this a compiler error? This function returns a `Result` so we can return a nice error rather than panic.

source/npm/qsharp/src/compiler/compiler.ts
```diff
@@ -202,6 +203,7 @@ export class Compiler implements ICompiler {
     expr: string,
     shots: number,
     pauliNoise: number[],
+    qubitLoss: number,
```
idavis, line 206: This functions name doesn't really match it's new capabilities.

source/compiler/qsc_eval/src/backend.rs
```diff
@@ -164,7 +180,27 @@ impl SparseSim {
...
 
     fn apply_noise(&mut self, q: usize) {
+        if self.is_qubit_lost(q) {
+            // If the qubit is already lost, we don't apply noise.
+            return;
+        }
         if let Some(rng) = &mut self.rng {
+            // First, check for loss.
+            let p = rng.gen_range(0.0..1.0);
+            if p < self.loss {
```
idavis, line 190: should this be `0.0..=1.0` and `p <= aelf.loss`?

source/pip/src/interpreter.rs
swernli: Yes, this is a compiler error. A `Result::Id` should never make it here, as those are only used by partial evaluation and not returned as output.

source/compiler/qsc_eval/src/lib.rs
swernli: Why remove the label? A lot of our errors use them, and it looks nice in the output (though generating this made me realize it should use the whole expression span rather than the right-hand side):
image

source/npm/qsharp/src/compiler/compiler.ts
swernli: Updated to `runWithNoise`

source/pip/tests/test_qsharp.py
swernli: I could, but since it's a runtime error like any other I don't think it needs special callout here. We verify that it behaves like other runtime errors in our Rust unit tests.

source/compiler/qsc_eval/src/backend.rs
swernli: I thought about this too, and I'm pretty sure that range would be too large. Easier to think about with integers... Imagine the range was `0..=100`, and that would actually be 101 possible results. To keep the range of probabilities in the expected zone with the right number of members, you either have to exclude `0.0` or `1.0`. By combining with `p < self.loss` we see that excluding `1.0` gives us the behavior we want: `0.0` means no `p` can ever satisfy the less-than check, and `1.0` means every `p` will satisfy the less-than check. This is the same pattern thats used for the Pauli noise below.

source/compiler/qsc_eval/src/intrinsic.rs
swernli: I had to do this:
```rust
impl<R> From<Option<R>> for val::Result
where
    R: Into<val::Result>
```
but that did indeed make the code much simpler.

source/compiler/qsc_eval/src/intrinsic.rs
swernli: I think with the addition of the `impl` this code now stays how it was before, so should no longer require the refactor.

source/compiler/qsc_eval/src/lib.rs
idavis: oh, I thought it would pick up the label from the `#[error...]`

swernli: Yeah, it's one of those miette subtleties... the `#[error...]` is what determines the text that appears with the red arrow in the screenshot, while the `label` text (if any) determines what appears with the purple underline. Many but not all of our errors have additional text in the label, but those that don't just show up with an underline on the span and no additional explanatory text.

idavis: In this case, should the error span be the comparison operation span or the variable span. Doesn't seem right for the span to be a constant.

source/compiler/qsc_eval/src/backend.rs
```diff
@@ -178,54 +214,102 @@ impl SparseSim {
...
     fn h(&mut self, q: usize) {
-        self.sim.h(q);
+        if !self.is_qubit_lost(q) {
+            self.sim.h(q);
+        }
         self.apply_noise(q);
     }
 
-    fn m(&mut self, q: usize) -> Self::ResultType {
+    fn m(&mut self, q: usize) -> Option<Self::ResultType> {
```
idavis: I don't think the measurements should return `Option`.

source/compiler/qsc_eval/src/lib.rs
swernli: I took a look at this, and it's enough of a refactor to pull in the additional spans that I think it deserves its own issue (and a review of existing errors to see which other ones can be improved):
#2581

source/compiler/qsc_eval/src/backend.rs
swernli: Oh, I think I see what you mean in your other comment about the impls now. I'll mull on this one, but you're right that there may be a way to get away with not changing this signature and reducing the churn of this change.

samples/notebooks/noise.ipynb
```diff
@@ -182,6 +182,69 @@
...
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%%qsharp\n",
+    "\n",
+    "operation CheckForLoss() : (Bool, Bool) {\n",
+    "    use q = Qubit();\n",
+    "    H(q);\n",
+    "    let res = MResetZ(q);\n",
```
billti, line 229: Was there a reason not to use `MResetZChecked` here? Or either works fine.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -169,6 +169,13 @@ pub enum Error {
     ))]
     ResultComparisonUnsupported(#[label("cannot compare to result")] PackageSpan),
 
+    #[error("cannot compare measurement result from qubit loss")]
+    #[diagnostic(code("Qsc.Eval.ResultLossComparisonUnsupported"))]
+    #[diagnostic(help(
+        "use of a measurement result from a qubit that was lost is not supported, use `CheckLoss` to check for loss instead"
```
billti: The end of this error reads a little weird. I assume you should `use "CheckLoss" to ensure the qubit is valid before using the measurement result`, not just use `CheckLoss` __instead__ of the measurement result.

source/vscode/package.json
```diff
@@ -141,6 +141,14 @@
             }
           }
         },
+        "Q#.simulation.qubitLoss": {
+          "markdownDescription": "The probability of a qubit loss occurring when running multiple shots via the Histogram command. This is applied for every qubit referenced.\n\nProbability values are in the range [0, 1].",
```
billti: Re `This is applied for every qubit referenced.`. This makes it sounds like the probability is applied once per qubit, but it is applied on every operation on a qubit, is it not?

billti (approved):

idavis (approved):

samples/notebooks/noise.ipynb
swernli: Either works. I wanted to show the way to use an existing result from a "normal" measurement function.

source/compiler/qsc_eval/src/lib.rs
swernli: Yeah, I can clean this up. Also, I'm glad you commented here, becuase I wrote the guidance before the name change, so `CheckLoss` should be `IsLossResult`. While I'm fixing that, it occurs to me that I could add something about loss to `ApplyIdleNoise` since that now can trigger loss if it's configured.

source/vscode/package.json
swernli: Good catch, I'll clean up the wording for clarity.

PR #2571, opened 2025-06-30
Adding box pragma support to QASM compiler

idavis:
Adding support for [`boxed expressions`](https://openqasm.com/language/delays.html#boxed-expressions) and pragmas for them when integrating with the Q# compiler.

When working in OpenQASM with the QDK, we can now use pragmas to denote parameterless void functions to call at the beginning and/or end of boxed expressions. They can additionally leverage simulatable intrinsic functions which will flow through to code generation.

```openqasm
OPENQASM 3.0;
include "stdgates.inc";
#pragma qdk.box.open box_begin
#pragma qdk.box.close box_end

@SimulatableIntrinsic
def box_begin() {}

@SimulatableIntrinsic
def box_end() {}

qubit q;
box {
    x q;
}
output bit c;
c = measure q;
```

```llvm
define void @ENTRYPOINT__main() #0 {
block_0:
  call void @box_begin()
  call void @__quantum__qis__x__body(%Qubit* inttoptr (i64 0 to %Qubit*))
  call void @box_end()
  call void @__quantum__qis__m__body(%Qubit* inttoptr (i64 0 to %Qubit*), %Result* inttoptr (i64 0 to %Result*))
  call void @__quantum__rt__tuple_record_output(i64 0, i8* null)
  ret void
}
```

### DIFF
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -4,12 +4,13 @@
 pub mod error;
 
 use core::f64;
-use std::{rc::Rc, sync::Arc};
+use std::{rc::Rc, str::FromStr, sync::Arc};
 
 use error::CompilerErrorKind;
 use num_bigint::BigInt;
 use qsc_data_structures::span::Span;
 use qsc_frontend::{compile::SourceMap, error::WithSource};
+use rustc_hash::FxHashMap;
 
 use crate::{
     CompilerConfig, OperationSignature, OutputSemantics, ProgramType, QasmCompileUnit,
@@ -18,9 +19,10 @@ use crate::{
         build_adj_plus_ctl_functor, build_angle_cast_call_by_name,
         build_angle_convert_call_with_two_params, build_arg_pat, build_array_reverse_expr,
         build_assignment_statement, build_attr, build_barrier_call, build_binary_expr,
-        build_call_no_params, build_call_with_param, build_call_with_params, build_classical_decl,
-        build_complex_from_expr, build_convert_call_expr, build_convert_cast_call_by_name,
-        build_end_stmt, build_expr_array_expr, build_for_stmt, build_function_or_operation,
+        build_call_no_params, build_call_stmt_no_params, build_call_with_param,
+        build_call_with_params, build_classical_decl, build_complex_from_expr,
+        build_convert_call_expr, build_convert_cast_call_by_name, build_end_stmt,
+        build_expr_array_expr, build_for_stmt, build_function_or_operation,
         build_gate_call_param_expr, build_gate_call_with_params_and_callee,
         build_if_expr_then_block, build_if_expr_then_block_else_block,
         build_if_expr_then_block_else_expr, build_if_expr_then_expr_else_expr,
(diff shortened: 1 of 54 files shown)


source/compiler/qsc_qasm/src/parser/stmt.rs
```diff
@@ -408,49 +412,105 @@ fn parse_include(s: &mut ParserContext) -> Result<StmtKind> {
...
+    };
+
+    let value_span = if value.is_none() {
+        None
+    } else {
+        Some(Span {
+            lo: value_lo + ident_lo,
+            hi: stmt_hi,
+        })
+    };
```
swernli: probably just code golf, but this seems like it could be a `map` or similar, rather than needing an explicit if-none-then-none-else pattern.

source/compiler/qsc_qasm/src/tests/statement/pragma/box.rs
```diff
@@ -0,0 +1,221 @@
...
+            sample_box_target();
+            sample_box_target();
+        };
+    "#]]
+    .assert_eq(&qsharp);
+    Ok(())
+}
+
+#[test]
+fn pragmas_can_have_separate_targets() -> miette::Result<(), Vec<Report>> {
```
swernli, line 71: Might be worth adding a test that sets two pragmas for box open/close and confirms only the last one "wins"

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -916,9 +1004,58 @@ impl QasmCompiler {
...
+                    }
+                }
+                self.push_compiler_error(CompilerErrorKind::InvalidBoxPragmaTarget(
+                    value.to_string(),
+                    stmt.value_span.unwrap_or(stmt.span),
+                ));
+            }
+            (PragmaKind::QdkBoxOpen | PragmaKind::QdkBoxClose, None) => {
+                self.push_compiler_error(CompilerErrorKind::MissingBoxPragmaTarget(stmt.span));
+            }
```
swernli: nitpick: could you move the error case to the top, that way the two "good" cases are together? I find that makes it easier to read this sort of pattern.

source/compiler/qsc_qasm/src/parser/stmt.rs
```diff
@@ -408,49 +412,101 @@ fn parse_include(s: &mut ParserContext) -> Result<StmtKind> {
...
+
+    // update the path span based on the original lo
+    path.offset(ident_lo);
+
+    // we need to get to the first non-whitespace token
+    // after the identifier, which should be the value.
+
+    // This value_lo offset is the start of the value in the pragma line content.
+    let value_lo = c.skip_trivia().hi;
+    let value = shorten_safely(value_lo.try_into().expect("valid size"), 0, pragma_content);
```
orpuente-MS: `shorten_safely` is only used with `from_end = 0`. Do we need the `from_end` argument?

source/compiler/qsc_qasm/src/parser/stmt.rs
```diff
@@ -408,49 +412,101 @@ fn parse_include(s: &mut ParserContext) -> Result<StmtKind> {
...
+    // advance the parser to the next token for the next token in the program
+    s.advance();
+
+    // create a sub-parser context just for the line content
+    // This will tokenize the rest of the line which should be the identifier
+    // and the value, if any.
+
+    let mut c = ParserContext::new(pragma_content);
+
+    println!("pragma content : {pragma_content}");
```
orpuente-MS: A leftover from debugging?

orpuente-MS (approved):

idavis:

idavis:

PR #2572, opened 2025-07-01
Show local variables of selected frame when debugging.

### LINKED ISSUE #2494: Add frame variables while debugging
Currently the debugger only shows the local variable for the top frame, even when you navigate to another frame on the stack. It would be useful, and consistent with other debuggers, to show locals for the selected stack frame.

orpuente-MS:
Fixes #2494 

This PR modifies the debugger behavior to show local variables for the selected stack frame when debugging.

### DIFF
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -607,13 +607,9 @@ impl State {
     }
 
     fn push_scope(&mut self, env: &mut Env) {
-        // The `EntryPoint` is the first and only function called from the entry.
-        // The `EntryPoint` should have `frame_id := 0`.
         // `push_frame`, which increments the length of `self.call_stack` by 1,
         // is called before `self.push_scope`.
-        // Therefore, when assigning the `frame_id` to the `EntryPoint`,
-        // `self.call_stack.len()` is already 1, and we need to correct
-        // by substracting 1, so that the `frame_id` of the `EntryPoint` is 0.
+        // Since the first `frame_id` should be 0, we substract 1 here.
         env.push_scope(self.call_stack.len() - 1);
     }
 
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -607,13 +607,13 @@ impl State {
     }
 
     fn push_scope(&mut self, env: &mut Env) {
-        // `Main` is the first and only function called from `entry_point`.
-        // `Main` should have `frame_id := 0`.
+        // The `EntryPoint` is the first and only function called from the entry.
+        // The `EntryPoint` should have `frame_id := 0`.
         // `push_frame`, which increments the length of `self.call_stack` by 1,
         // is called before `self.push_scope`.
-        // Therefore, when assigning the `frame_id` to `Main`,
+        // Therefore, when assigning the `frame_id` to the `EntryPoint`,
         // `self.call_stack.len()` is already 1, and we need to correct
-        // by substracting 1, so that the `frame_id` of `Main` is 0.
+        // by substracting 1, so that the `frame_id` of the `EntryPoint` is 0.
         env.push_scope(self.call_stack.len() - 1);
     }
 
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -607,7 +607,14 @@ impl State {
     }
 
     fn push_scope(&mut self, env: &mut Env) {
-        env.push_scope(self.call_stack.len());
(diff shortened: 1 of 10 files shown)


source/vscode/src/debugger/session.ts
```diff
@@ -748,15 +750,19 @@ export class QscDebugSession extends LoggingDebugSession {
...
-          this.variableHandles.create("quantum"),
+          this.variableHandles.create(["quantum", args.frameId]),
           true, // expensive - keeps scope collapsed in the UI by default
         ),
         new Scope(
           "Quantum Circuit",
-          this.variableHandles.create("circuit"),
+          this.variableHandles.create(["circuit", args.frameId]),
           true, // expensive - keeps scope collapsed in the UI by default
         ),
```
idavis: Does it make sense to pass 0/-1 for quantum and circuit? The frame id doesn't make sense. Or Should the type on the handle be an explicit enum where the locals variant has an int for frame?

source/vscode/src/debugger/session.ts
```diff
@@ -776,11 +782,12 @@ export class QscDebugSession extends LoggingDebugSession {
       variables: [],
     };
 
-    const handle = this.variableHandles.get(args.variablesReference);
+    const [handle, frameID] = this.variableHandles.get(args.variablesReference);
```
idavis, line 785: See comment above, this would become a real type so that we don't have an extra int in invalid cases

source/vscode/test/suites/debugger/openqasm.test.ts
```diff
@@ -339,13 +339,76 @@ suite("OpenQASM Debugger Tests", function suite() {
...
+    ]);
+
+    // Step over to prepare the tracker to detect a new variable.
+    await vscode.commands.executeCommand("workbench.action.debug.stepOver");
+
+    // Go one frame down the call stack.
+    await vscode.debug.activeDebugSession?.customRequest("scopes", {
+      frameId: 1,
+    });
+
```
idavis: This comment isn't what actually happens. What does this do and what are the side effects?

source/vscode/test/suites/debugger/qsharp.test.ts
```diff
@@ -400,13 +400,84 @@ suite("Q# Debugger Tests", function suite() {
...
 
+  test("Show local variables of selected frame", async () => {
+    // Set a breakpoint on line 16 of foo.qs (7 when 0-indexed)
+    // This will be in the function `AnotherCallFrame` after `b`
+    // has been defined.
+    vscode.debug.addBreakpoints([
+      new vscode.SourceBreakpoint(
+        new vscode.Location(fooUri, new vscode.Position(15, 0)),
+      ),
+    ]);
```
idavis: Line numbers in code and comments are inconsistent

source/vscode/test/suites/debugger/qsharp.test.ts
```diff
@@ -400,13 +400,84 @@ suite("Q# Debugger Tests", function suite() {
...
+      { name: "b", type: undefined, value: "3", variablesReference: 0 },
+    ]);
+
+    // Step over to prepare the tracker to detect a new variable.
+    await vscode.commands.executeCommand("workbench.action.debug.stepOver");
+
+    // Go one frame down the call stack.
+    await vscode.debug.activeDebugSession?.customRequest("scopes", {
+      frameId: 0,
+    });
```
idavis: See comment in openqasm test

source/vscode/src/debugger/session.ts
orpuente-MS: I like the idea of passing -1 better. It will make handling the argument in `variableHandles.create` easier.

orpuente-MS: I think even better would be having three different helper functions `localsRequest`, `quantumStateRequest`, and `circuitRequest` instead of sending three different kinds of input to the same function and then switching on the input.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -607,7 +607,10 @@ impl State {
     }
 
     fn push_scope(&mut self, env: &mut Env) {
-        env.push_scope(self.call_stack.len());
+        // `push_frame`, which increments the length of `self.call_stack` by 1,
+        // is called before `self.push_scope`.
+        // Since the first `frame_id` should be 0, we substract 1 here.
+        env.push_scope(self.call_stack.len() - 1);
```
swernli, line 613: Ah, this is where the off-by-one error is coming from! Glad you found this, now the different env scopes will maintain a frame_id that matches the index of that frame in the call_stack.

swernli (approved):

Issue #2497, opened 2025-06-03
Add a "What's new" alert and changelog for the VS Code extension


billti:
When we add significant new features, having a 'what's new' notification in the VS Code extension could help awareness. For example, the CodeLLDB extension does this.

Also, having a changelog.md in our VS Code package should I believe have it show up for the extension in the marketplace.

ScottCarda-MS:
Closed with #2576

PR #2576, opened 2025-07-03
Changelog View Added

ScottCarda-MS:
Adds a changelog view that can be accessed via a new command: "QDK: Show Changelog". The view is populated by the contents of the CHANGELOG.md file, and the user is given a notification of a new changelog when the CHANGELOG_VERSION constant defined in changelog.ts is updated. The user can opt out of these notifications after the first one, and can opt back in by changing the new Q# setting.

The CHANGELOG_VERSION is currently set to `undefined` which means that the changelog page will not automatically appear for users. The first time users will be notified of a new changelog page will be when the constant is manually updated to a new value.

Manually Validated the following scenarios:
- With setting = false:
  - CHANGELOG_VERSION = undefined : no popup
  - CHANGELOG_VERSION = undefined -> "defined" : no popup but changelog automatically shows
  - CHANGELOG_VERSION = "defined" : no popup
  - CHANGELOG_VERSION = "defined" -> "defined2" : popup
  - CHANGELOG_VERSION = "defined2" : no popup
  - CHANGELOG_VERSION = "defined2" -> undefined : popup (this is unrealistic scenario)

- With setting = true: (no popup for all cases)
  - CHANGELOG_VERSION = undefined
  - CHANGELOG_VERSION = undefined -> "defined"
  - CHANGELOG_VERSION = "defined"
  - CHANGELOG_VERSION = "defined" -> "defined2"
  - CHANGELOG_VERSION = "defined2"
  - CHANGELOG_VERSION = "defined2" -> undefined (this is unrealistic scenario)

Addresses: #2497


source/vscode/WHATSNEW.md
```diff
@@ -0,0 +1,7 @@
+# What's New in QDK
```
billti: This should be a CHANGELOG.md, not a WHATSNEW.md. For example, see the repo at https://github.com/vadimcn/codelldb/blob/master/CHANGELOG.md.

By adding this, you also get a change log show up in the marketplace (e.g. see https://marketplace.visualstudio.com/items/vadimcn.vscode-lldb/changelog) as well as in the VS Code extension gallary (e.g. see below)

image

ScottCarda-MS: Okay I've changed "What's New" to "Changelog" everywhere, except I kept "What's New" for the popup notification as I felt that was more user friendly. I also didn't change the contents of the CHANGELOG.md because that is dummy content.

source/vscode/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# What's New in QDK
+
+## v1.2.3
+
+- Added feature X
+- Improved Y
+- Fixed bug Z
```
billti: We can't merge it with a mock changelog, as this is what will automatically show up for the extension as soon as the next build is published. Can you add the contents for the past couple of releases? You can use the releases page on GitHub for reference.

ScottCarda-MS: I've filled in this changelog with the changelogs of 1.18.0, 1.17.0, and 1.16.0.

source/vscode/src/changelog.ts
```diff
@@ -0,0 +1,72 @@
...
+    vscode.commands.registerCommand("qsharp-vscode.showChangelog", async () => {
+      const changelogUri = vscode.Uri.joinPath(
+        context.extensionUri,
+        "CHANGELOG.md",
+      );
+      await vscode.commands.executeCommand(
+        "markdown.showPreview",
+        changelogUri,
+        vscode.ViewColumn.One,
+        { locked: true },
```
billti: I can't find docs for this API/argument. What does `{locked: true}` do and why do we need it here?

source/vscode/src/changelog.ts
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as vscode from "vscode";
+
+// The latest version for which we want to show the changelog page
+const CHANGELOG_VERSION = undefined; // <-- Update this when you want to show a new changelog to users
```
billti: Let's set this to `"v1.18.0"` now then (and bump when we release) so that it starts showing up and we can start testing it.

source/vscode/src/extension.ts
```diff
@@ -89,6 +90,9 @@ export async function activate(
   // fire-and-forget
   registerGhCopilotInstructionsCommand(context);
 
+  // Show prompt after update if not suppressed
+  await maybeShowChangelogPrompt(context);
```
billti: I'm not sure we should `await` here. This is in the activation flow, and it awaits the user response to the popup message - thus completing activation will be blocked until the uses dismisses the popup (which may trigger warnings in VS Code based on extension activation taking too long).

source/vscode/CHANGELOG.md
```diff
@@ -0,0 +1,182 @@
...
+- Add reset broadcast by [@idavis](https://github.com/idavis) in [#2404](https://github.com/microsoft/qsharp/pull/2404)
+- Limit inferred type sizes by [@swernli](https://github.com/swernli) in [#2382](https://github.com/microsoft/qsharp/pull/2382)
+- Add measurement broadcast by [@idavis](https://github.com/idavis) in [#2405](https://github.com/microsoft/qsharp/pull/2405)
+- bitstring processing is more efficient by [@idavis](https://github.com/idavis) in [#2408](https://github.com/microsoft/qsharp/pull/2408)
+- Restore fancy error reporting in Python by [@swernli](https://github.com/swernli) in [#2410](https://github.com/microsoft/qsharp/pull/2410)
+- OpenQASM Grover's algorithm sample by [@DmitryVasilevsky](https://github.com/DmitryVasilevsky) in [#2398](https://github.com/microsoft/qsharp/pull/2398)
+- OpenQASM Bernstein-Vazirani sample by [@DmitryVasilevsky](https://github.com/DmitryVasilevsky) in [#2403](https://github.com/microsoft/qsharp/pull/2403)
+- Users/merlinbot/1es pt auto baselining pr by [@idavis](https://github.com/idavis) in [#2371](https://github.com/microsoft/qsharp/pull/2371)
```
billti: We should trim this changelog to list things that are new that users would be relevant and interesting. Listing every commit re dependecy updates, pipline changes, etc. is too noisy.

source/vscode/src/changelog.ts
```diff
@@ -0,0 +1,72 @@
...
+      );
+    }),
+  ];
+}
+
+export async function maybeShowChangelogPrompt(
+  context: vscode.ExtensionContext,
+) {
+  const lastChangelogVersion = context.globalState.get<string>(
+    "qdk.lastChangelogVersion",
```
billti: Can this name be a bit more explicit/accurate? Maybe `qdk.lastChangelogNotificationVersion`.

source/vscode/src/changelog.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (
+    lastChangelogVersion !== CHANGELOG_VERSION &&
+    !suppressUpdateNotifications
+  ) {
+    await context.globalState.update(
+      "qdk.lastChangelogVersion",
+      CHANGELOG_VERSION,
+    );
+    // Only show prompt if not first install (i.e., lastChangelogVersion is not undefined/null)
+    if (lastChangelogVersion !== undefined) {
```
billti: Is this on the assumption that if they just installed the QDK, they're likely aware of the latest features, and/or showing a notification as soon as you install is a bit noisy? Makes sense. I like it. Can you add a comment to outline that.

ScottCarda-MS: This just suppresses the prompt for seeing the changelog. On a freash install, you will just get the changelog view unprompted.

ScottCarda-MS: I updated the comment to try to make that a bit clearer. It _is_ to reduce noise for the user on install.

source/vscode/src/changelog.ts
ScottCarda-MS: done

source/vscode/src/extension.ts
ScottCarda-MS: removed

source/vscode/src/changelog.ts
ScottCarda-MS: done, the changelog will show up (for the first time) immediately after running a build with this in it.

source/vscode/src/changelog.ts
ScottCarda-MS: removed

source/vscode/CHANGELOG.md
ScottCarda-MS: What is the vision for this view? I had originally imagined it would just have the info for the latest version, but I just put the changelogs for the last couple in here as a placeholder until we had that next version to talk about here. Are we wanting this to be a file that is continually added to version after version?

ScottCarda-MS: I've trimmed off the items that seemed very implementation specific and tried to keep the user-facing items.

billti (approved):

PR #2577, opened 2025-07-07
Upgrade rust edition to 2024

orpuente-MS:
This PR upgrades the rust edition to 2024.


source/compiler/qsc_hir/src/hir.rs
```diff
@@ -382,6 +380,7 @@ impl Display for Item {
 }
 
 /// An item kind.
+#[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug, PartialEq)]
 pub enum ItemKind {
     /// A `function` or `operation` declaration.
```
idavis: we should fix this rather than allow

source/compiler/qsc_hir/src/hir.rs
```diff
@@ -1115,6 +1114,7 @@ impl Display for FieldAssign {
 }
 
 /// A string component.
+#[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug, PartialEq)]
 pub enum StringComponent {
     /// An expression.
```
idavis: We should fix this rather than allow

source/compiler/qsc_qasm/src/semantic/ast.rs
```diff
@@ -990,6 +990,7 @@ impl Display for ForStmt {
 }
 
 #[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum EnumerableSet {
     Set(Set),
     Range(Range),
```
idavis: Fix instead of allow

source/compiler/qsc_qasm/src/semantic/ast.rs
```diff
@@ -1358,6 +1359,7 @@ impl fmt::Display for Version {
 }
 
 #[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum Index {
     Expr(Expr),
     Range(Range),
```
idavis: Fix instead of allow

idavis (approved):

PR #2582, opened 2025-07-08
OpenQASM VSCode Spec Mode

idavis:
This PR adds a new configuration setting allowing the user to disconnect the Q# integration features of the language service (such as code lenses, debugging, execution, etc). The use still get syntax and semantic errors.


idavis:

PR #2583, opened 2025-07-08
Upgrade the rust version to 1.88

orpuente-MS:
This PR upgrades the rust toolchain version from 1.86 to 1.88 across multiple workflows and files. To make the upgrade possible, we had to fix a linker issue and a wasm-config flag that changed in 1.87.

## Linker issue
Starting on rust 1.87, the Std doesn't link advapi32 on windows anymore. So, C libraries that depend on it, like mimalloc, now need to link it manually. [Link to the Rust PR](https://github.com/rust-lang/rust/pull/138233).

## wasm config flag
Rust 1.87 upgraded to LLVM 20, which enables bulk-memory by default. Therefore, `wasm-opt` needs to be configured to expect this setting. This behavior is document [here](https://doc.rust-lang.org/rustc/platform-support/wasm32-unknown-unknown.html#enabled-webassembly-features).


package.json
```diff
@@ -54,5 +54,10 @@
     "url": "^0.11.3",
     "user-agent-data-types": "^0.4.2",
     "vscode-uri": "^3.0.8"
+  },
+  "npm": {
```
billti: What does this do and why is it needed? I'm not familiar with it and a quick search isn't helping much.

source/wasm/Cargo.toml
```diff
@@ -45,3 +45,13 @@ indoc = { workspace = true }
...
+
+# Rust 1.87 upgraded to LLVM 20, which enables bulk-memory by default.
+# Therefore, `wasm-opt` needs to be configured to expect this setting.
+# This behavior is document in:
+# <https://doc.rust-lang.org/rustc/platform-support/wasm32-unknown-unknown.html#enabled-webassembly-features>.
+[package.metadata.wasm-pack.profile.profiling]
+wasm-opt = ['--enable-bulk-memory', '--enable-nontrapping-fptoint']
+
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = ['--enable-bulk-memory', '--enable-nontrapping-fptoint']
```
billti: Does the watch.mjs script at line 73 also need these flags specified?

billti: I tried. It doesn't.

idavis (approved):

PR #2584, opened 2025-07-09
Allow unrestricted capabilities in `@Test` callables

### LINKED ISSUE #2504: How to write tests in base profile projects
Possibly related to any solution for #1198

Test code is often written in 'unrestricted' code, as it wants to take measurements and compare them, etc. This means when you switch the profile to 'base' to submit to hardware your project has a bunch of errors due to the tests. We should figure out how best to solve this.

swernli:
This change creates a special exception for callables annoted with `@Test` in Runtime Capabilities Analysis, making it easier to write and run unit tests with profiles other than "Unrestricted." As part of this change, the `CheckZero`, `CheckAllZero`, and `CheckOperationsAreEqual` callables are no longer marked as `@Config(Unrestricted)` so that they can be used from tests. Any calls into these will trigger updated codegen error cases. In addition, calls into `@Test` callables also trigger a new codegen-time error.

Fixes #2504

### DIFF
--- a/library/std/src/Std/Diagnostics.qs
+++ b/library/std/src/Std/Diagnostics.qs
@@ -122,7 +122,6 @@ function DumpMatrix(qs : Qubit[]) : Unit {
 /// # Remarks
 /// This operation is useful for checking whether a qubit is in the |0⟩ state during simulation. It is not possible to check
 /// this on hardware without measuring the qubit, which could change the state.
-@Config(Unrestricted)
 operation CheckZero(qubit : Qubit) : Bool {
     body intrinsic;
 }
@@ -144,7 +143,6 @@ operation CheckZero(qubit : Qubit) : Bool {
 /// # Remarks
 /// This operation is useful for checking whether a qubit is in the |0⟩ state during simulation. It is not possible to check
 /// this on hardware without measuring the qubit, which could change the state.
-@Config(Unrestricted)
 operation CheckAllZero(qubits : Qubit[]) : Bool {
     for q in qubits {
         if not CheckZero(q) {
@@ -201,7 +199,6 @@ function Fact(actual : Bool, message : String) : Unit {
 /// Operation defining the expected behavior for the operation under test.
 /// # Output
 /// True if operations are equal, false otherwise.
-@Config(Unrestricted)
 operation CheckOperationsAreEqual(
     nQubits : Int,
     actual : (Qubit[] => Unit),
--- a/source/compiler/qsc_partial_eval/src/lib.rs
+++ b/source/compiler/qsc_partial_eval/src/lib.rs
@@ -122,6 +122,20 @@ pub enum Error {
     #[error("failed to evaluate: {0} is not supported")]
     #[diagnostic(code("Qsc.PartialEval.Unimplemented"))]
     Unimplemented(String, #[label] PackageSpan),
+
+    #[error("unsupported call into test callable")]
+    #[diagnostic(code("Qsc.PartialEval.UnsupportedTestCallable"))]
+    #[diagnostic(help(
+        "callables with the `@Test` annotation are not supported in partial evaluation."
+    ))]
+    UnsupportedTestCallable(#[label] PackageSpan),
+
+    #[error("unsupported use of simulation-only intrinsic `{0}`")]
+    #[diagnostic(code("Qsc.PartialEval.UnsupportedSimulationIntrinsic"))]
+    #[diagnostic(help(
+        "intrinsic callables that require a running quantum simulator cannot be invoked in partial evaluation"
+    ))]
+    UnsupportedSimulationIntrinsic(String, #[label] PackageSpan),
 }
 
 impl From<EvalError> for Error {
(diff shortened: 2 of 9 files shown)


source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -122,6 +122,20 @@ pub enum Error {
     #[error("failed to evaluate: {0} is not supported")]
     #[diagnostic(code("Qsc.PartialEval.Unimplemented"))]
     Unimplemented(String, #[label] PackageSpan),
+
+    #[error("unsupported call into test callable")]
+    #[diagnostic(code("Qsc.PartialEval.UnsupportedTestCallable"))]
+    #[diagnostic(help(
+        "callables with the `@Test` annotation are not supported in partial evaluation."
```
billti: If this is a user facing message, can we making to 'friendlier'? I expect very few users would know what "partial evaluation" is. Is something like "A compiled program may not include code that calls into operations marked with a `@Test` attribute" reasonable?

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -122,6 +122,20 @@ pub enum Error {
...
+    ))]
+    UnsupportedTestCallable(#[label] PackageSpan),
+
+    #[error("unsupported use of simulation-only intrinsic `{0}`")]
+    #[diagnostic(code("Qsc.PartialEval.UnsupportedSimulationIntrinsic"))]
+    #[diagnostic(help(
+        "intrinsic callables that require a running quantum simulator cannot be invoked in partial evaluation"
```
billti: Same comment as above. Let's make this error understandable beyond folks that work on our team :-)

library/std/src/Std/Diagnostics.qs
```diff
@@ -122,7 +122,6 @@ function DumpMatrix(qs : Qubit[]) : Unit {
 /// # Remarks
 /// This operation is useful for checking whether a qubit is in the |0⟩ state during simulation. It is not possible to check
 /// this on hardware without measuring the qubit, which could change the state.
-@Config(Unrestricted)
 operation CheckZero(qubit : Qubit) : Bool {
```
DmitryVasilevsky, line 125: I'm not sure I understand. Suppose I use Base profile. If I call this from a regular operation where will it fail compilation? And if I call it from a `@Test` operation it will run correctly and perform the check? Can `@Test` function be called from a regular one?

source/compiler/qsc_partial_eval/src/lib.rs
swernli: I'll make the adjustment. For what it's worth, we have other error messages and notifications that use "partial evaluation" so we might want to take a pass at cleaning all of those up if this is a term we want to avoid.

library/std/src/Std/Diagnostics.qs
swernli: No, a `@Test` operation being called from a regular one will fail during QIR generation with the new error I added, `UnsupportedTestCallable`.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1614,9 +1661,13 @@ impl<'a> PartialEvaluator<'a> {
...
             | "ApplyIdleNoise"
             | "GlobalPhase" => Ok(Value::unit()),
+            "CheckZero" => Err(Error::UnsupportedSimulationIntrinsic(
+                "CheckZero".to_string(),
+                callee_expr_span,
+            )),
             // The following intrinsic functions and operations should never make it past conditional compilation and
             // the capabilities check pass.
-            "CheckZero" | "DrawRandomInt" | "DrawRandomDouble" | "DrawRandomBool" | "Length" => {
+            "DrawRandomInt" | "DrawRandomDouble" | "DrawRandomBool" | "Length" => {
```
billti, line 1667: Wording comments re "partial evaluation" aside, is there much value in changing it from this error just to add a new error which then finishes with "...cannot be invoked in partial evaluation" anyway?

source/compiler/qsc_partial_eval/src/lib.rs
swernli: I'll need a paragraph or two to explain the concepts then... is the request here to improve the help text or the error text? I think the error text is already the concise, easy to understand version: "unsupported use of simuilation only intrinsic `<name>`."

swernli: Actually, for this one there isn't really help text that is helpful without getting into internal details, so I'll just remove this help.

source/compiler/qsc_partial_eval/src/lib.rs
swernli: It makes a really big difference. I ran into this in testing and the existing code path is really meant for things that shouldn't happen but we didn't want to panic and crash on. Here's what it looks like before the new error:
image
Here's what it looks like with this error (minus the help text I mentioned above that I'd remove):
image

library/std/src/Std/Diagnostics.qs
DmitryVasilevsky: OK, I cannot call `@Test` operations from regular ones, good.
What about CheckZero? Can I call it now in base profile from a normal operation? Before this change it wouldn't be available.

swernli: No, it will generate an error in Base profile:
image
For Adaptive, it won't be a compile time error but an error during QIR code generation (see the image shared with Bill here: https://github.com/microsoft/qsharp/pull/2584#discussion_r2195671191)

billti (approved):

Issue #1104, opened 2024-02-01
State preparation could be better


DmitryVasilevsky:
Postponed items from the code review (https://github.com/microsoft/qsharp/pull/1068):

internal operation ApproximatelyMultiplexZ:
This cost of this can be cut in half, by applying the first part of its relative phase decomposition, which is then uncomputed in the adjoint call to the within block.
(by Mathias)

internal operation ApproximatelyMultiplexPauli:
This can be made more concise with Microsoft.Quantum.Intrinsic.MapPauli, but currently this operation is internal.
(by Mathias)

DmitryVasilevsky:
One of the issues should be resolved with this: https://github.com/microsoft/qsharp/pull/1461

DmitryVasilevsky:
#2 is resolved by https://github.com/microsoft/qsharp/pull/2585

PR #2585, opened 2025-07-10
Making MapPauliAxis public with updates to functionality and comments

DmitryVasilevsky:
This adds a function to remap Pauli axes `MapPauliAxis` to the new BlochSphere.qs file. It is based on the previously internal MapPauli and brings some updates and extensive doc comments. This is useful to apply operations in different Pauli bases.

For example, rotation around the X axis can be expressed as a rotation around the Z axis by remapping the Pauli axes so that X axis becomes Z axis. The following two operations are equivalent:

```
Rx(0.1, q)
```
and
```
within {
    MapPauliAxis(PauliZ, PauliX, q);
} apply {
    Rz(0.1, q);
}
```

This function is used internally in some rotations and state preparation.
Resolves part 2 of https://github.com/microsoft/qsharp/issues/1104


library/std/src/Std/BlochSphere.qs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// # Description
+/// This function applies single-qubit Clifford transformations that remap Pauli axes on a Bloch sphere.
+/// These mappings are useful to apply operations known for one Pauli basis in different Pauli bases.
+/// Provide `from` and `to` parameters in terms of a passive transformation. For example,
+/// when a rotation around the X axis is desired and a rotation around the Z axis is available,
+/// the Z axis gets pointed in the direction of the X axis so that the rotation around Z axis can be used.
+/// Out of several transformations that achieve the requested mapping, the one with fewer gates is used.
+///
+/// # Remarks
```
swernli: In our other doc comments, the `# Remarks` section goes after the `# Input` section, I believe.

library/std/src/Std/BlochSphere.qs
```diff
@@ -0,0 +1,95 @@
...
+/// +X+Y+Z ↦ -Z+X-Y: HXS⁻¹, HSX, ZHS⁻¹, YHS, HYS, HS⁻¹Y      \
+/// +X+Y+Z ↦ +X-Z+Y: SHS, HS⁻¹H                              \
+/// +X+Y+Z ↦ -Y+Z-X: SHY, S⁻¹HZ, YS⁻¹H, SYH, S⁻¹XH, XSH      \
+/// +X+Y+Z ↦ -X+Y-Z: Y                                       \
+/// +X+Y+Z ↦ -Z+Y+X: HX, ZH                                  \
+/// +X+Y+Z ↦ -Y-X-Z: YS, SX, S⁻¹Y, XS⁻¹                      \
+/// +X+Y+Z ↦ +Y+X-Z: YS⁻¹, SY, S⁻¹X, XS                      \
+/// +X+Y+Z ↦ +Z+Y-X: XH, HZ                                  \
+/// +X+Y+Z ↦ +X+Z-Y: S⁻¹HS⁻¹, HSH                            \
+/// +X+Y+Z ↦ -Z-X+Y: HXS, HSY, ZHS, YHS⁻¹, HYS⁻¹, HS⁻¹X      \
```
swernli: I think it's worth double checking that these unicode characters show up as expected in the VS Code renering of stdlib docs.

library/std/src/Std/BlochSphere.qs
```diff
@@ -0,0 +1,95 @@
...
+/// } apply {
+///    Rz(0.1, q);
+/// }
+/// ```
+///
+/// # References
+/// - [Wikipedia: Bloch sphere](https://wikipedia.org/wiki/Bloch_sphere)
+/// - [Wikipedia: Clifford group](https://wikipedia.org/wiki/Clifford_group)
+/// - [Wikipedia: Active and passive transformation](https://wikipedia.org/wiki/Active_and_passive_transformation)
+operation MapPauliAxis(from : Pauli, to : Pauli, q : Qubit) : Unit is Adj + Ctl {
```
swernli: Now that this is just one callable rather than a family, I wonder if it's worth folding it into an existing namespace, such as `Std.Canon` where things like `CY` and `ApplyPauli` live?

swernli: One advantage of it being in `Std.Canon` is that you won't need to fully qualify it as much; that namespace is part of our default pre-amble, so it's imported by default.

library/std/src/Std/BlochSphere.qs
DmitryVasilevsky: Updated.

library/std/src/Std/BlochSphere.qs
DmitryVasilevsky: In my VSCode they show correctly. I did try this before using them. But the concern is valid. However, we already have unusual Unicode characters in other doc comments...

library/std/src/Std/BlochSphere.qs
DmitryVasilevsky: I was thinking if I should add this to Canon or create a new file. I leaned towards the new file because other utilities may appear there with time. With time doesn't always work so I wasn't strongly convinced... Since there're now two of you now, I moved it to Canon.

PR #2586, opened 2025-07-10
Add sizeof bultin to qasm

orpuente-MS:
This PR adds support for the `sizeof` builtin to OpenQASM. As part of the necessary things to make `sizeof` work in all supported cases, it also adds support for `readonly array[_, ...]` input types in subroutines.

### DIFF
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -791,7 +791,13 @@ impl QasmCompiler {
         let array_dims = expr.array_dims;
         assert!((1..=7).contains(&array_dims));
         let fn_name = format!("sizeof_{array_dims}");
-        build_call_with_params(&fn_name, &[], operands, name_span, span)
+        build_call_with_params(
+            &fn_name,
+            &["Std", "OpenQASM", "Builtin"],
+            operands,
+            name_span,
+            span,
+        )
     }
 
     fn compile_gate_call_stmt(&mut self, stmt: &semast::GateCall) -> Option<qsast::Stmt> {
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -785,15 +785,13 @@ impl QasmCompiler {
     fn compile_sizeof_call_expr(&mut self, expr: &semast::SizeofCallExpr) -> qsast::Expr {
         let span = expr.span;
         let name_span = expr.fn_name_span;
-        let mut array = self.compile_expr(&expr.array);
-
-        // Index into the array `dim` times to access the `dim` dimension.
-        for _ in 0..expr.dim {
-            array = build_index_expr(array, self.compile_expr(&Expr::uint(0, span)), span);
-        }
-
-        // Call Q#'s `Length` with the indexed array.
-        build_call_with_param("Length", &[], array, name_span, expr.array.span, span)
+        let array = self.compile_expr(&expr.array);
+        let dim = self.compile_expr(&expr.dim);
+        let operands = vec![array, dim];
+        let array_dims = expr.array_dims;
+        assert!((1..=7).contains(&array_dims));
+        let fn_name = format!("sizeof_{array_dims}");
+        build_call_with_params(&fn_name, &[], operands, name_span, span)
     }
 
     fn compile_gate_call_stmt(&mut self, stmt: &semast::GateCall) -> Option<qsast::Stmt> {
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -789,7 +789,10 @@ impl QasmCompiler {
         let dim = self.compile_expr(&expr.dim);
         let operands = vec![array, dim];
         let array_dims = expr.array_dims;
-        assert!((1..=7).contains(&array_dims), "array dimension should be between 1 and 7");
+        assert!(
+            (1..=7).contains(&array_dims),
+            "array dimension should be between 1 and 7"
+        );
         let fn_name = format!("sizeof_{array_dims}");
         build_call_with_params(
             &fn_name,
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -789,7 +789,7 @@ impl QasmCompiler {
         let dim = self.compile_expr(&expr.dim);
         let operands = vec![array, dim];
         let array_dims = expr.array_dims;
-        assert!((1..=7).contains(&array_dims));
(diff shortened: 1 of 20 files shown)


source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -1801,13 +1824,93 @@ impl Lowerer {
...
+        let first_arg = inputs_iter.next().expect("there is at least one argument");
+        let second_arg = inputs_iter
+            .next()
+            .unwrap_or_else(|| Expr::uint(0, expr.span));
+
+        // Check that the 2nd argument is a const expr.
+        let Some(second_arg) = second_arg.with_const_value(self).get_const_u32() else {
+            self.push_const_eval_error(sizeof_invalid_args_error(expr.span, &inputs));
+            return err_expr!(Type::Err, expr.span);
+        };
```
idavis: Is this only needed to be true in the const int case?

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -1801,13 +1824,93 @@ impl Lowerer {
...
     fn lower_function_call_expr(&mut self, expr: &syntax::FunctionCall) -> semantic::Expr {
         // 1. If the name refers to a builtin function, we defer
         //    the lowering to `lower_builtin_function_call_expr`.
         if let Ok(builtin_fn) = BuiltinFunction::from_str(&expr.name.name) {
             return self.lower_builtin_function_call_expr(expr, builtin_fn);
         }
 
+        if &*expr.name.name == "sizeof" {
+            return self.lower_sizeof_call_expr(expr);
+        }
```
idavis, line 1901: Should this be part of `BuiltinFunction::from_str(&expr.name.name)` above and call `lower_sizeof_call_expr` from there?

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -4319,3 +4436,33 @@ impl FromStr for BuiltinFunction {
         }
     }
 }
+
+fn sizeof_invalid_args_error(call_span: Span, inputs: &[Expr]) -> ConstEvalError {
+    let mut error_msg = String::new();
+    write!(
+        error_msg,
+        "There is no valid overload of `sizeof` for inputs: "
```
idavis, line 4474:
```suggestion
        "There is no valid overload of `sizeof` for inputs: "
```

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -4319,3 +4436,33 @@ impl FromStr for BuiltinFunction {
         }
     }
 }
+
+fn sizeof_invalid_args_error(call_span: Span, inputs: &[Expr]) -> ConstEvalError {
```
idavis, line 4470: Should this be with the builtin functions code?

source/compiler/qsc_qasm/src/types.rs
```diff
@@ -59,74 +59,57 @@ impl Display for CallableKind {
...
+            ArrayDimensions::One => 1,
+            ArrayDimensions::Two => 2,
+            ArrayDimensions::Three => 3,
+            ArrayDimensions::Four => 4,
+            ArrayDimensions::Five => 5,
+            ArrayDimensions::Six => 6,
+            ArrayDimensions::Seven => 7,
+        }
+    }
+}
```
swernli: This can be a little more tersely written as:

```suggestion
pub enum ArrayDimensions {
    One = 1,
    Two = 2,
    Three = 3,
    Four = 4,
    Five = 5,
    Six = 6,
    Seven = 7,
}

impl From<ArrayDimensions> for u32 {
    fn from(value: ArrayDimensions) -> Self {
        value as u32
    }
}
```

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -785,6 +782,24 @@ impl QasmCompiler {
...
     }
 
+    fn compile_sizeof_call_expr(&mut self, expr: &semast::SizeofCallExpr) -> qsast::Expr {
+        let span = expr.span;
+        let name_span = expr.fn_name_span;
+        let array = self.compile_expr(&expr.array);
+        let dim = self.compile_expr(&expr.dim);
+        let operands = vec![array, dim];
+        let array_dims = expr.array_dims;
+        assert!((1..=7).contains(&array_dims));
```
swernli: niptick: it's helpful to have some text with the assert.

```suggestion
        assert!((1..=7).contains(&array_dims), "array dimension should be between 1 and 7");
```

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -1096,11 +1098,38 @@ impl Lowerer {
         }
     }
 
+    fn make_qsharp_array_ty(
+        base_ty: &crate::semantic::types::ArrayBaseType,
+        dims: crate::types::ArrayDimensions,
+    ) -> crate::types::Type {
+        use crate::semantic::types::ArrayBaseType;
```
swernli: I think we've preferred keeping `use` stmts together at the top of the file.

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -1096,11 +1098,38 @@ impl Lowerer {
...
+            }
+        }
+    }
+
     fn convert_semantic_type_to_qsharp_type(
         &mut self,
         ty: &super::types::Type,
         span: Span,
     ) -> crate::types::Type {
+        use crate::semantic::types::ArrayBaseType;
```
swernli: Here too it would be good to move this to the top of the file.

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -2802,70 +2892,140 @@ impl Lowerer {
...
+
+                Type::make_static_array_ref_ty(&dims, &base_ty, is_mutable)
+            }
+            syntax::ArrayReferenceType::Dyn(ref_ty) => {
+                let is_mutable = matches!(ref_ty.mutability, syntax::AccessControl::Mutable);
+                let base_ty =
+                    self.get_semantic_type_from_array_base_ty(&ref_ty.base_type, ref_ty.span);
+                let Some(num_dims) = self.const_eval_array_size_designator_expr(&ref_ty.dimensions)
+                else {
+                    return Type::Err;
```
swernli, line 3008: This case seems to be missing the push of an error message, where all the others have it.

orpuente-MS: Oh. In this case `self.const_eval_array_size_designator_expr(&ref_ty.dimensions)` already pushed the relevant error message. But I will add a comment explaining that.

idavis (approved):

PR #2587, opened 2025-07-11
Remove wasm-pack dependency

billti:
wasm-pack has been a bit of a pain, and seems largely unmaintained at this point (has several hundred issues open and hasn't had a commit in 9 months). We don't use many of its features anyway (such as testing and packaging for npm).

With this change, we use wasm-bindgen (for the .JS files) and wasm-opt (to optimize the wasm) directly, instead of via wasm-pack. All wasm-pack was doing under the covers was installing these tools from the internet and invoking them, so this just makes their usage more explicit.

With this we can:

- Directly control the versions of the wasm utilities used
- Hopefully build wasm on Windows ARM64 (haven't tried yet, but wasm-pack didn't work there)
- Avoid streaming a .sh script from the internet to install wasm-pack at build time

This requires both wasm-bindgen and wasm-opt be on the path, but 'python prereqs.py --install' will add those if needed. (They are handy to have around if working with wasm anyway).

Open to feedback. It wasn't as simple as I hoped, but should be a one-off change.

### DIFF
--- a/build.py
+++ b/build.py
@@ -13,7 +13,7 @@ import shutil
 import subprocess
 import functools
 
-from prereqs import check_prereqs
+from prereqs import check_prereqs, get_wasm_env
 
 # Disable buffered output so that the log statements and subprocess output get interleaved in proper order
 print = functools.partial(print, flush=True)
@@ -69,6 +69,13 @@ parser.add_argument(
 )
 
 parser.add_argument(
+    "--web-only",
+    action=argparse.BooleanOptionalAction,
+    default=False,
+    help="Build only the web version of the wasm package",
+)
+
+parser.add_argument(
     "--ci-bench",
     action=argparse.BooleanOptionalAction,
     default=False,
@@ -110,6 +117,7 @@ npm_install_needed = (
 npm_cmd = "npm.cmd" if platform.system() == "Windows" else "npm"
 
 build_type = "debug" if args.debug else "release"
+wasm_targets = ["web", "nodejs"] if not args.web_only else ["web"]
 run_tests = args.test
 
 root_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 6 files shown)


build.py
```diff
@@ -68,6 +68,13 @@
     help="Build and run the integration tests (default is --no-integration-tests)",
 )
 
+parser.add_argument(
```
billti, line 71: Added this for the watch.mjs script to use, so it can just build the web version for speed (else it will run the nodejs wasm build also)

build.py
```diff
@@ -401,44 +409,92 @@ def run_ci_historic_benchmark():
...
 if build_wasm:
-    step_start("Building the wasm crate")
-    # wasm-pack can't build for web and node in the same build, so need to run twice.
-    # Hopefully not needed if https://github.com/rustwasm/wasm-pack/issues/313 lands.
-    build_flag = "--release" if build_type == "release" else "--dev"
-
-    wasm_pack_args = ["wasm-pack", "build", build_flag]
-    web_build_args = ["--target", "web", "--out-dir", os.path.join(wasm_bld, "web")]
-    node_build_args = [
+    env = get_wasm_env()
```
billti: This is to put the 'default' directories where 'prereqs.py --install' downloads the utilities to on the end of the PATH. (It will still pick up any versions you have earlier on your path if manually installled).

prereqs.py
```diff
@@ -54,15 +66,12 @@ def check_prereqs(install=False, skip_wasm=False):
...
-    version_match = re.search(r"v(\d+)\.(\d+)\.\d+", node_version.decode())
-    if version_match:
-        node_major = int(version_match.group(1))
-        node_minor = int(version_match.group(2))
-        if node_major < node_ver[0] or (
-            node_major == node_ver[0] and node_minor < node_ver[1]
-        ):
+    ver_match = re.search(r"v(\d+)\.(\d+)\.(\d+)", node_version.decode())
+    if ver_match:
+        found_ver = tuple(int(g) for g in ver_match.groups())
```
billti, line 77: I simplified several overly-verbose version check logic throughout to this style.

prereqs.py
```diff
@@ -213,6 +215,69 @@ def wasm_checks(install, installed_rust_targets):
...
+        "linux": {
+            "arm64": f"binaryen-version_{binaryen_ver}-aarch64-linux.tar.gz",
+            "x64": f"binaryen-version_{binaryen_ver}-x86_64-linux.tar.gz",
+        },
+        "windows": {
+            "arm64": f"binaryen-version_{binaryen_ver}-x86_64-windows.tar.gz",
+            "x64": f"binaryen-version_{binaryen_ver}-x86_64-windows.tar.gz",
+        },
+    }
+    binaryen_base_url = f"https://github.com/WebAssembly/binaryen/releases/download/version_{binaryen_ver}/"
```
billti, line 281: I did mess around with using the cross-platform 'binaryen' npm package, but running the wasm optimization was taking about 47sec on my mac, versus about 3sec if using the platform native binaries - so installing the binaries it is!

source/npm/qsharp/generate_docs.js
```diff
@@ -7,7 +7,7 @@ import { existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 
-import { generate_docs } from "./lib/node/qsc_wasm.cjs";
+import { generate_docs } from "./lib/nodejs/qsc_wasm.cjs";
```
billti, line 10: Normalized the subdir name to match the 'wasm-bindgen' target option. This should only affect us, and we only use it for node unit tests and doc generation.

watch.mjs
```diff
@@ -54,42 +53,24 @@ const release = process.argv.includes("--release");
...
-  console.log("wasm-pack done! ", result.stderr.toString());
-
-  console.log("Copying the wasm-pack output files to the npm package");
-  const npmLibDir = join(npmDir, "lib", "web");
-
-  ["qsc_wasm_bg.wasm", "qsc_wasm.d.ts", "qsc_wasm.js"].forEach((file) =>
-    copyFileSync(join(buildDir, file), join(npmLibDir, file)),
+    "python",
+    [
+      "build.py",
```
billti, line 64: I replaced the repeating of all the wasm build logic here with just calling out to build.py to rebuild the wasm, which is much simpler and avoids duplicating some tricky subprocess invocations.

watch.mjs
```diff
@@ -184,11 +165,6 @@ runWatcher(npmDir, "npm");
 // VSCode and playground are built by esbuild, but run the type checker in watch mode
 runWatcher(join(qdkSrcDir, "vscode"), "vscode");
 runWatcher(join(qdkSrcDir, "vscode"), "vscode webview", "tsc:watch:view");
-runWatcher(
-  join(qdkSrcDir, "vscode"),
-  "vscode copilot webview",
-  "tsc:watch:copilotView",
-);
```
billti, line 191: This was dead code since we removed the custom copilot UX (and was causing errors in the existing code)

build.py
```diff
@@ -401,44 +409,95 @@ def run_ci_historic_benchmark():
...
+        #
+        # Note: wasm-opt is not needed for debug builds, so we only run it for release builds.
+        if build_type == "release":
+            wasm_file = os.path.join(out_dir, "qsc_wasm_bg.wasm")
+            wasm_opt_args = [
+                "wasm-opt",
+                "-Oz",
+                '--enable-bulk-memory',
+                '--enable-nontrapping-float-to-int',
+                "--output",
```
orpuente-MS: I think this comment is from when you had `--all-features` a few commits ago. Can we update it to include the info deleted in wasm/Cargo.toml explaining why we enabled these specific features?

```python
# Rust 1.87 upgraded to LLVM 20, which enables bulk-memory and non-trapping float by default.
# (See https://github.com/rust-lang/rust/pull/137322)
# Therefore, `wasm-opt` needs to be configured to expect this setting.
# This behavior is document in:
# <https://doc.rust-lang.org/rustc/platform-support/wasm32-unknown-unknown.html#enabled-webassembly-features>.
```

billti: Yeah, I saw that but didn't want to push again yet while the CI is showing green and have it retrigger :-D

FWIW, I did try with --all-features, but it actually fails to then run, so changed it back. We do need to be explicit and keep that list up to date.

I already fixed the comment locally, and will push if/when I have any other feedback to address.

build.py
```diff
@@ -401,44 +409,95 @@ def run_ci_historic_benchmark():
...
+    cargo_args = [
+        "cargo",
+        "build",
+        "--lib",
         "--target",
-        "nodejs",
-        "--out-dir",
-        os.path.join(wasm_bld, "node"),
+        "wasm32-unknown-unknown",
+        "--target-dir",
```
idavis: Why does the target dir need to be explicitly added?

prereqs.py
```diff
@@ -157,60 +156,134 @@ def check_prereqs(install=False, skip_wasm=False):
...
+    # Note: Using curl and tar as subprocesses rather than Python libraries for features such as --strip-components
+    subprocess.run(["curl", "-L", "-o", temp_file, url_base + tar_file], check=True)
+    subprocess.run(
+        ["tar", "-xzf", temp_file, "--strip-components=1", "-C", out_dir], check=True
+    )
+
+    os.remove(temp_file)  # Clean up the tar file
+
+
+def install_wasm_bindgen():
```
idavis, line 239: What about `cargo install wasm-bindgen-cli` with `--locked` and a specific version?

billti: That builds from source, which is both a lot slower, and failed on Windows ARM64 (hence the comment on line 241 "...we want x64 builds on Windows ARM64")

Note you can still install wasm-bindgen yourself using `cargo install wasm-bindgen-cli`. That's what I've always done. The scripts will find it and use it and not install anything if it is already present and found.

build.py
billti: Maybe it doesn't. Didn't seem bad to be explicit. I can remove it if you'd rather it was implicit.

billti: Changed.

prereqs.py
idavis: You can also use `cargo binstall wasm-bindgen-cli` I trust building from source the most, and it's a one time cost except on CI

billti: `cargo binstall` isn't availalbe by default though, so that's another prereq instead.

> I trust building from source the most

Maybe we should recommend users get the QDK by building the repo themselves then instead of installing from the marketplace 😛

billti: In all seriousness though, the bininstall page states _"Binstall works by fetching the crate information from crates.io and searching the linked repository for matching releases and artifacts"_, so even if we did the work to use it, it's probably unpacking exactly the same tarball as an end result.

prereqs.py
```diff
@@ -35,6 +36,23 @@ def get_installed_rust_targets() -> str:
         raise Exception(message)
 
 
+def add_wasm_tools_to_path():
```
idavis, line 39: Should we only modify the path if the binary is not already on the path?

idavis (approved):

prereqs.py
billti: It's kind of benign to add it anyway, just seemed pointless if it's already there.

I specifically add it to the end of the path, so it will only find the tool in the added location if it wasn't already on the PATH earlier.

PR #2590, opened 2025-07-11
Emit spec compliant QIR

swernli:
This updates the QIR code generation logic to comply with the Base and Adaptive Profile specifications. In particular, it makes the following updates:

 - The entry point function returns `i64`.
 - Module level flags for computation extensions are updated to be consistent with spec fixes for LLVM flags.
 - QIR now includes global strings in the data section that are used as tags for the output recording calls, with tag values chosen to be consistent with the Azure Quantum service output translation.
 - The `__quantum__qis__read_result__body` function is renamed to `__quantum__rt__read_result`.
- Every program begins with a call to the `__quantum__rt__initialize` function.


source/compiler/qsc_codegen/src/qir.rs
```diff
@@ -728,8 +728,8 @@ fn get_module_metadata(program: &rir::Program) -> String {
                     let name = "int_computations";
                     writeln!(
                         flags,
```
idavis, line 753: probably worth a note of what 5 means. Or introduce an enum that impls `From<i64>`.

swernli: Added a comment with a link to the metadata flags section of the LLVM spec.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -3233,11 +3273,17 @@ impl<'a> PartialEvaluator<'a> {
...
     ) -> Result<(), ()> {
         let Ty::Tuple(elem_tys) = ty else {
             panic!("expected tuple type for tuple value");
         };
+        let new_tag_root = format!("{tag_root}t");
         let tuple_record_callable_id = self.get_tuple_record_callable();
+        // let idx = self.program.tags.len();
+        // let tag = format!("{idx}_{new_tag_root}");
+        // let len = tag.len();
+        // self.program.tags.push(tag);
```
idavis: Is is refactoring leftovers?

swernli: Whoops, yes these can be removed.

PR #2591, opened 2025-07-14
Single File Profiles

ScottCarda-MS:


source/compiler/qsc/src/interpret/package_tests.rs
```diff
@@ -53,7 +53,7 @@ fn import_and_call_reexport() {
     };
 
     // This builds all the dependencies
-    let buildable_program = BuildableProgram::new(TargetCapabilityFlags::all(), pkg_graph);
+    let buildable_program = BuildableProgram::new(TargetCapabilityFlags::all(), pkg_graph, false);
```
idavis: An enum is more desciptive and readable than passing a bool

source/compiler/qsc/src/packages.rs
```diff
@@ -78,16 +82,41 @@ fn convert_circuit_sources(
...
+
+    // If the entry profile is set, we need to ensure that the user code is compiled with it.
+    if let Some((profile, span)) = entry_profile {
+        // If this is a project (qsharp.json present), emit error with span
+        if is_single_file {
+            capabilities = profile.into();
+        } else {
+            dependency_errors.push(Error::from_map(
+                &SourceMap::new(sources, None),
+                ErrorKind::Frontend(entrypoint_profile_in_project_error(span)),
```
swernli: Rather than defining the new error inside of the parser for it to only be used here, I think it would just make sense to add it to the `qsc::compile::ErrorKind`. It's fine to add a concrete error case (as in, doesn't wrap another error) to that enum, and then the error is defined in the same layer that detects and reports it, avoiding the need for extra plumbing.

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -376,10 +380,19 @@ impl With<'_> {
...
+                {
+                    Some(hir::Attr::EntryPoint)
+                }
+                // Any other form is not valid so generates an error.
                 _ => {
-                    self.lowerer
-                        .errors
-                        .push(Error::InvalidAttrArgs("()".to_string(), attr.arg.span));
+                    self.lowerer.errors.push(Error::InvalidAttrArgs(
+                        "profile name".to_string(),
```
swernli: Since `()` is still valid, the error should not imply that profile name is required.

```suggestion
                        "empty or profile name ".to_string(),
```

source/compiler/qsc/src/interpret/package_tests.rs
swernli: Agreed... it's a little more hoop-jumping, but something like `ProjectType::SingleFile` vs `ProjectType::Project` would make it abundantly clear what is being used here and in other places where the option is passed.

swernli: Ah, there is already a `ProjectType`, so another name, maybe `ProjectStructure` or something like that.

source/compiler/qsc_project/src/project.rs
```diff
@@ -46,6 +47,10 @@ pub struct Project {
     pub errors: Vec<Error>,
     /// The type of project. This is used to determine how to load the project.
     pub project_type: ProjectType,
+    /// QIR target profile for this project (from manifest or default)
+    pub target_profile: Profile,
+    /// Whether this project is a single file (i.e. no manifest file).
+    pub is_single_file: bool,
```
minestarks: Just make `is_single_file` a field of `PackageGraphSources` . It logically and neatly belongs there. I think it'll save you a bunch of plumbing.

Also, personally I'd reverse the logic and call it `has_manifest` since I feel that that would be more accurate. (A project with a `qsharp.json` could still have a single file... so this name is a bit confusing IMO)

source/language_service/src/state/tests.rs
minestarks, line 1: I can't understand why these tests had to change so much. I can't find any meaningful changes in the language service that would have made these changes necessary?

source/language_service/src/compilation.rs
minestarks, line 1: Not related to this file, but commenting here out of convenience - shall we add hardcoded completions for the inside of the `@EntryPoint()` attribute? (I remember we opted not to for `@Config()`, but this one may be simpler to do, since you can only put an identifier in the attribute, not expressions)

source/wasm/src/project_system.rs
```diff
@@ -29,6 +29,7 @@ export interface IProgramConfig {
     packageGraphSources: IPackageGraphSources;
     profile: TargetProfile;
     projectType: ProjectType;
+    isSingleFile: boolean;
```
minestarks: As I commented elsewhere (for the corresponding native struct) I think `isSingleFile` simply belongs in `PackageGraphSources` and the rest of the changes should fall into place

source/vscode/test/suites/language-service/notebook.test.ts
```diff
@@ -117,41 +116,4 @@ suite("Q# Notebook Tests", function suite() {
     assert.equal(location.range.start.line, 2);
     assert.equal(location.range.start.character, 10);
   });
-
-  test("Notebook uses Unrestricted profile by default even when workspace is Base", async function () {
```
minestarks, line 121: I love that this test went away. Thank you 🙏

source/vscode/src/qirGeneration.ts
```diff
@@ -114,11 +124,40 @@ async function getQirForProgram(
     );
     if (result?.action !== "set") {
       throw new QirGenerationError(
-        error_msg +
-          " Please update the QIR target via the status bar selector or extension settings.",
+        error_msg + " Please update the QIR target via the qsharp.json.",
       );
     } else {
-      await setTarget(targetSupportsAdaptive ? "adaptive_ri" : "base");
+      // Try to update the qsharp.json manifest automatically.
```
minestarks: Was this behavior decided as part of a team discussion or something? It's weird. Can we just open the qsharp.json and let the user change the file if they want to?

source/language_service/src/state/tests.rs
ScottCarda-MS: It is because before the changes, we were passing in the target profile as an extra argument to the compilation. Now target profile needs to be specified by the source code, either in the manifest or in an entrypoint argument.

source/compiler/qsc/src/interpret/package_tests.rs
ScottCarda-MS: Since moving this into the packageGraphSources struct, we only ever see this Boolean accompanied by the field name `has_manifest`. I'm hoping that is sufficient to provide clarity and readability now.

source/compiler/qsc/src/packages.rs
ScottCarda-MS: Okay I've moved this error to the qsc and removed the parser-specific error kinds/constructors.

source/compiler/qsc_frontend/src/lower.rs
ScottCarda-MS: done

source/compiler/qsc_project/src/project.rs
ScottCarda-MS: I've move that into the package graph sources as you suggested and made it the `has_manifest`

source/wasm/src/project_system.rs
ScottCarda-MS: done

source/vscode/src/qirGeneration.ts
ScottCarda-MS: This was reflecting the existing behavior of that option, which updates the global profile setting. A button that says "Set the QIR profile to Base to continue" should do the actual setting, imho

source/language_service/src/compilation.rs
ScottCarda-MS: I've added those completions. Nice suggestion!

source/compiler/qsc/src/interpret.rs
```diff
@@ -99,6 +99,7 @@ pub enum Error {
     NoEntryPoint,
     #[error("unsupported runtime capabilities for code generation")]
     #[diagnostic(code("Qsc.Interpret.UnsupportedRuntimeCapabilities"))]
+    #[diagnostic(help("@EntryPoint attribute argument should be 'Base' or 'Adaptive_RI'"))]
```
billti: We also have Adaptive_RIF, right @idavis?

ScottCarda-MS: We do, but that one doesn't seem to get mentioned much. Seems like we don't advertise it like `Adaptive_RI`. I can add it into the message though if we want to point it out here.

source/vscode/src/qirGeneration.ts
ScottCarda-MS: Changed to just open the qsharp.json to encourage the user to change the field, as suggested.

source/compiler/qsc/src/interpret.rs
ScottCarda-MS: Updated to include `Adaptive_RIF`

orpuente-MS: I think the message should be comprehensive. So, I lean in favor of including `Adaptive_RIF`.

PR #2592, opened 2025-07-14
Improve runtime error debugging

swernli:
This change updates the way that runtime errors are reported from debug execution, allowing the VS Code debugger to break in and treat those as "exceptions" with detailed information. The tweaks avoid extra checks during the tight loop for evaluation, and keeps behavior the same for non-debug execution (like what Python native binaries do).


source/compiler/qsc_eval/src/lib.rs
```diff
@@ -785,6 +826,9 @@ impl State {
             return Ok(res);
         }
 
+        // Throw any error that was encountered during execution.
+        self.get_last_error()?;
```
idavis: This comment doesn't match the behavior.

PR #2594, opened 2025-07-14
Annotations should now be parsed correctly

idavis:
This PR fixes annotation parsing and renames existing annotations to comply better with namespaced intent from the spec. The old annotation names have been preserved for backwards compatibility but should be removed in the future.
- `@SimulatableIntrinsic` is now `@qdk.qir.intrinsic`
- `@Config` is now `@qdk.qir.profile`

### DIFF
--- a/source/compiler/qsc_qasm/src/parser/stmt.rs
+++ b/source/compiler/qsc_qasm/src/parser/stmt.rs
@@ -380,7 +380,12 @@ pub fn parse_annotation(s: &mut ParserContext) -> Result<Annotation> {
     // This will tokenize the rest of the line which should be the identifier
     // and the value, if any.
 
-    let mut c = ParserContext::new(annotation_content);
+    let mut c = match s.word_collector {
+        Some(ref mut collector) => {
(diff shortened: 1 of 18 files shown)


source/compiler/qsc_qasm/src/parser/stmt.rs
```diff
@@ -335,55 +335,95 @@ fn default(span: Span) -> Stmt {
...
+    // annotation_line_content:
+    // @ a.b.c "some value" more value \nmore content
+    // |.      |     |                |
+    // |.      |     |                |
+    // |.      |     ^------ value ---|
+    // |.      ^-- ident_lo           ^-- stmt_hi
+    // |       |                      |
+    // |       ^- annotation_content -^
+    // |      ^-- annotation_kw_offset
+    // ^-- stmt_lo
```
ScottCarda-MS: I really like this diagram!

ScottCarda-MS (approved):
The name change (as well as the rest of the changes) looks good to me. Thanks!

orpuente-MS (approved):

PR #2600, opened 2025-07-15
Fix endianness bug when indexing into OpenQASM classical types

orpuente-MS:


source/compiler/qsc_qasm/src/semantic/const_eval.rs
```diff
@@ -165,14 +165,24 @@ impl LiteralKind {
...
+                // We need to change the endianness of the index.
+                // Our bitarrays are stored as BigInts, but indexing into OpenQASM bit registers
+                // behaves like indexing in quantum-register or a normal array, in that doing
+                // a[0] gets the left-most element. So, instead of doing BigInt[idx] we need to
+                // do BigInt[size - idx - 1].
+                let mask = BigInt::from(1) << (u64::from(size) - idx - 1);
                 Some(Self::Bit((value & mask) != BigInt::ZERO))
             }
             Index::Range(range) => {
                 let (start, step, end) = compute_slice_components(range, size);
```
swernli, line 178: it looks like `step` always defaults to 1 when not specified, but if the original OpenQASM range was negative, like `[-16:-1]`, is that an empty range (because step defaults to `1`) or is step expected to automatically be `-1` when both the start and end of the range are negative?

orpuente-MS: It always defaults to 1.

The [Spec says](https://openqasm.com/versions/3.1/language/types.html#register-concatenation-and-slicing) "Ranges are written as a:b or a:c:b where a, b, and c are integers... If c is not given, it is assumed to be one, and cannot be zero."

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -4118,6 +4130,48 @@ impl Lowerer {
...
+            semantic::Index::Range(range) => {
+                // When changing the endianness of a range, we need to swap the start and end.
+                let start = range.end.map(width_minus_expr_minus_one);
+                let end = range.start.map(width_minus_expr_minus_one);
+                semantic::Index::Range(Box::new(semantic::Range {
+                    span: range.span,
+                    start,
+                    end,
+                    step: range.step,
+                }))
```
swernli: I'm still a little confused on how `step` factors in here. Does it also need to be transformed, or can we safely assume that the step is still correct even after this transformation has taken place?

orpuente-MS: Yes, there is a bug on this one. Nice catch! I was trying to account for it by swapping the start and end, but that only works if the number of elements in the range is an exact multiple of the step. The correct solution is leaving the start and end in the original order, and instead negating the step.

To give a more concrete example, if you have `int[8] a`, and you want to evaluate `a[0:2:5]`. The intent is accessing the indices `0, 2, 4`. After applying the `width - idx - 1` transformation to each of those indices, we get `7, 5, 3`, which is equivalent to the range `7:-2:2` resulting from applying the `width - idx - 1` transformation to the start and end of the original range and then negating the step.

PR #2604, opened 2025-07-16
Added telemetry to changelog prompt

ScottCarda-MS:
Added two telemetry events, `ChangelogPromptStart` and `ChangelogPromptEnd` for tracking the use of the changelog prompt. `ChangelogPromptStart` will fire when we show the popup and has a `qdkVersion` field for what version of the QDK the popup is being shown for. `ChangelogPromptEnd` will fire when the user clicks one of the buttons on the changelog popup and has an `action` field with possible values `"showChangelog"` or `"suppressChangelog"` for discerning which button was pressed. Both events have and `associationId` field to correlate instances of the two.

Also removed the restriction that prevents the popup from appearing on a fresh install, as we want the popup to appear in that scenario now.

### DIFF
--- a/source/vscode/src/changelog.ts
+++ b/source/vscode/src/changelog.ts
@@ -48,7 +48,7 @@ export async function maybeShowChangelogPrompt(
     const associatedId = getRandomGuid();
     sendTelemetryEvent(EventType.ChangelogPromptStart, {
       associationId: associatedId,
-      qdkVersion: CHANGELOG_VERSION,
+      changelogVersion: CHANGELOG_VERSION,
     });
     const choice = await vscode.window.showInformationMessage(
       "The Azure Quantum Development Kit has been updated.",
--- a/source/vscode/src/telemetry.ts
+++ b/source/vscode/src/telemetry.ts
@@ -328,7 +328,7 @@ type EventTypes = {
   [EventType.ChangelogPromptStart]: {
     properties: {
       associationId: string;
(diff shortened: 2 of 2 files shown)


source/vscode/src/changelog.ts
```diff
@@ -42,27 +44,37 @@ export async function maybeShowChangelogPrompt(
...
-            true,
-            vscode.ConfigurationTarget.Global,
-          );
-        vscode.window.showInformationMessage(
-          'You will no longer receive "What\'s New" notifications. You can re-enable them from the Q# settings.',
+    const buttons = ["What's New?", "Don't show this again"];
+    const associatedId = getRandomGuid();
+    sendTelemetryEvent(EventType.ChangelogPromptStart, {
+      associationId: associatedId,
+      qdkVersion: CHANGELOG_VERSION,
```
billti: Not to be nitpicky, but this should be "changelogVersion" as the property name. As discussed, we'll only bump the CHANGELOG_VERSION when we have notable new changes, so there may be months where we ship a new QDK but don't bump this if there's only minor fixes and updates.

Also, the VS Code telemtry sends through the extension version anyway, so we always get that (i.e. all events have fields like `"ExtensionName":"quantum.qsharp-lang-vscode","ExtensionVersion":"1.18.0"`)

ScottCarda-MS: done

PR #2607, opened 2025-07-16
Update expect formatting in unit tests to be more readable

minestarks:
In anticipation of some bugfixes I'm making in this area, I wanted to clean up the unit tests a bit, so that the upcoming PR is less headache-inducing.

Changing how we format the expected output in completions tests. This way:

- They're more readable

- There's less churn when sort texts change due to unrelated changes, (e.g. [this PR](https://github.com/microsoft/qsharp/pull/2345/files#diff-474255c7ce3b4fd2763734186079ae268564b2bc4f9377c266de1ed6e8f9df65)). _Relative_ order is still validated, but the exact sort text (e.g. "`0300DumpMachine`") is not.

### DIFF
--- a/source/language_service/src/completion/tests.rs
+++ b/source/language_service/src/completion/tests.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use super::{CompletionItem, get_completions};
+use super::{Compilation, CompletionItem, get_completions};
 use crate::{
     Encoding,
     protocol::CompletionList,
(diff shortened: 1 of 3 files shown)


source/language_service/src/completion/tests.rs
```diff
@@ -150,36 +158,80 @@ fn assert_no_duplicates(mut actual_completions: CompletionList) {
...
+                if let Some(detail) = &c.detail {
+                    writeln!(output, "    detail: {detail:?}")?;
+                }
+
+                if let Some(edits) = &c.additional_text_edits {
+                    writeln!(output, "    additional_text_edits:")?;
+                    for edit in edits {
+                        writeln!(
+                            output,
+                            "      [{}:{}-{}:{}] {:?}",
```
orpuente-MS: Nitpick: there is a blank line at the end of each test. You can remove it by doing `write!(output, "\n ...")` in these places instead of using `writeln!`.

orpuente-MS (approved):
Love this changes!

ScottCarda-MS (approved):
lgtm, thanks!

PR #2610, opened 2025-07-17
Restrict OpenQASM indices to range `[-N, N - 1]`.

orpuente-MS:
We initially implemented OpenQASM indexing using arithmetic mod N, where N is the size of the array. We are now restricting indices to be in the range [-N, N - 1].

### DIFF
--- a/source/compiler/qsc_qasm/src/semantic/types.rs
+++ b/source/compiler/qsc_qasm/src/semantic/types.rs
@@ -8,7 +8,7 @@ use qsc_data_structures::span::Span;
 
 use super::ast::{BinOp, Expr, ExprKind, Index, LiteralKind, Range};
 use crate::parser::ast as syntax;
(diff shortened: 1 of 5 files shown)


source/compiler/qsc_qasm/src/semantic/types.rs
```diff
@@ -555,12 +559,31 @@ fn indexed_type_builder(
 ///
```
swernli, line 558: I can't comment on the exact line, but I think it's worth updating this comment to indicate how the implementation interprets it now, ie: not wrapping.

idavis (approved):

source/compiler/qsc_qasm/src/semantic/types.rs
orpuente-MS: Yes, nice catch. Updated.

PR #2611, opened 2025-07-17
Lower duration and stretch

idavis:
This PR attempts to lower duration, and stretch values in OpenQASM to try to make them spec 'compliant'. The duration and stretch types are supposed to be compile-time constant which would imply they are actually const decls but not declared in const decl statements.

This is complicated by the `durationof` which is supposed to return a duration. The `durationof` call will accumulate the durations of boxes, gate calls, and delays.

**NOTE**
In draft while getting spec clarification. Where we are at:
- Stetches are always const. Default value is `0`. 
- Durations can be but aren't required to be const
- `durationof` is problematic. No real way calculate duration without defcals. We may need two different calls in the spec. One where it is const and other call where it returns a non-const duration.

### DIFF
--- a/source/compiler/qsc_qasm/src/semantic/tests/expression/binary/duration.rs
+++ b/source/compiler/qsc_qasm/src/semantic/tests/expression/binary/duration.rs
@@ -75,7 +75,7 @@ fn addition_of_two_durations_returns_duration() {
 }
 
 #[test]
-fn addition_with_duration() {
+fn addition_of_duration_and_int_errors() {
     let input = "
         duration a;
         int b;
@@ -583,7 +583,7 @@ fn multiplication_float_by_duration() {
 }
 
 #[test]
-fn multiplication_assign_op() {
+fn multiplication_assign_op_not_support() {
     let input = "
         duration a = 2 ns;
         duration b = 3 ns;
@@ -854,7 +854,7 @@ fn division_assign_op() {
 #[test]
 fn division_assign_op_errors_when_duration_is_const() {
     let input = "
-        duration a = 12 ns;
+        const duration a = 12 ns;
         const float b = 3.0;
         a /= b;
     ";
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_qasm/src/semantic/tests/expression/explicit_cast_from_stretch.rs
```diff
@@ -0,0 +1,669 @@
...
+        "#]],
+    );
+}
+
+//===================
+// Casts to stretch
+//===================
+
+#[test]
+fn stretch_to_stretch() {
```
orpuente-MS, line 96: `stretch_to_stretch` and `stretch_to_stretch_is_no_op` are the same test. I think you can remove one of them.

source/compiler/qsc_qasm/src/semantic/tests/statements/delay_stmt.rs
```diff
@@ -0,0 +1,128 @@
...
+    );
+}
+
+#[test]
+fn on_a_single_qubit_with_variables() {
+    check_stmt_kinds(
+        "qubit q;
+        const duration stride = 5ns;
+        const int p = 2;
+        delay [p * stride] q;",
```
orpuente-MS, line 91: Do delay statements only work with const durations? Can we add a test like this one?

```
qubit q;
duration d = 5ns;
delay [d] q;
```

source/compiler/qsc_qasm/src/semantic/tests/expression/binary/duration.rs
```diff
@@ -0,0 +1,894 @@
...
+                            kind: SymbolId(8)
+                        rhs: Expr [71-72]:
+                            ty: const float
+                            kind: SymbolId(9)
+        "#]],
+    );
+}
+
+#[test]
+fn division_assign_op_errors_when_duration_is_const() {
```
orpuente-MS, line 855: This test is the same one as `division_assign_op`

source/compiler/qsc_qasm/src/semantic/tests/expression/binary/duration.rs
```diff
@@ -0,0 +1,894 @@
...
+}
+
+#[test]
+fn multiplication_assign_op() {
+    let input = "
+        duration a = 2 ns;
+        duration b = 3 ns;
+        a *= b;
+    ";
+
```
orpuente-MS: Can we sufix the name of this test with `_errors`?

source/compiler/qsc_qasm/src/semantic/tests/expression/binary/duration.rs
```diff
@@ -0,0 +1,894 @@
...
+    );
+}
+
+#[test]
+fn addition_with_duration() {
+    let input = "
+        duration a;
+        int b;
+        duration x = a + b;
+    ";
```
orpuente-MS: Can we rename it to `addition_of_duration_and_int_errors`?

source/compiler/qsc_qasm/src/semantic/tests/statements/delay_stmt.rs
idavis: As far as I know they don't have to be constant. Given all code I've seen also declares the durations as non-const var or const literals.

PR #2612, opened 2025-07-17
QASM Profile Pragma

ScottCarda-MS:
Adds the `#pragma qdk.qir.profile` pragma to our QASM compiler. This pragma is used to specify the profile to compile the QASM file with, i.e. `#pragma qdk.qir.profile Base` to have the file compile with the `Base` profile. QASM files are always compiled with the profile specified by their source with this pragma. If no pragma is provided, the compilation will default to `Unrestricted`.

The PR is built on top of and relies on #2591

### DIFF
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -8,7 +8,7 @@ use std::{rc::Rc, str::FromStr, sync::Arc};
 
 use error::CompilerErrorKind;
 use num_bigint::BigInt;
-use qsc_data_structures::span::Span;
+use qsc_data_structures::{span::Span, target::Profile};
 use qsc_frontend::{compile::SourceMap, error::WithSource};
 use rustc_hash::FxHashMap;
 
@@ -114,6 +114,7 @@ pub fn compile_to_qsharp_ast_with_config(
 pub enum PragmaKind {
     QdkBoxOpen,
     QdkBoxClose,
+    QdkProfile,
 }
 
 impl FromStr for PragmaKind {
@@ -123,6 +124,7 @@ impl FromStr for PragmaKind {
         match s.to_lowercase().as_str() {
             "qdk.box.open" => Ok(PragmaKind::QdkBoxOpen),
             "qdk.box.close" => Ok(PragmaKind::QdkBoxClose),
+            "qdk.qir.profile" => Ok(PragmaKind::QdkProfile),
             _ => Err(()),
         }
     }
@@ -1113,6 +1115,20 @@ impl QasmCompiler {
             }
             (PragmaKind::QdkBoxOpen | PragmaKind::QdkBoxClose, None) => {
                 self.push_compiler_error(CompilerErrorKind::MissingBoxPragmaTarget(stmt.span));
+            }
+            (PragmaKind::QdkProfile, Some(profile)) => {
+                if Profile::from_str(profile).is_ok() {
+                    self.pragma_config
+                        .insert(PragmaKind::QdkProfile, profile.clone());
+                    return;
+                }
+                self.push_compiler_error(CompilerErrorKind::InvalidProfilePragmaTarget(
+                    profile.to_string(),
+                    stmt.value_span.unwrap_or(stmt.span),
+                ));
+            }
+            (PragmaKind::QdkProfile, None) => {
+                self.push_compiler_error(CompilerErrorKind::MissingProfilePragmaTarget(stmt.span));
             }
         }
     }
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -114,7 +114,7 @@ pub fn compile_to_qsharp_ast_with_config(
 pub enum PragmaKind {
     QdkBoxOpen,
     QdkBoxClose,
-    QdkProfile,
+    QdkQirProfile,
 }
 
 impl FromStr for PragmaKind {
@@ -124,7 +124,7 @@ impl FromStr for PragmaKind {
         match s.to_lowercase().as_str() {
             "qdk.box.open" => Ok(PragmaKind::QdkBoxOpen),
             "qdk.box.close" => Ok(PragmaKind::QdkBoxClose),
-            "qdk.qir.profile" => Ok(PragmaKind::QdkProfile),
+            "qdk.qir.profile" => Ok(PragmaKind::QdkQirProfile),
             _ => Err(()),
         }
     }
@@ -190,7 +190,7 @@ impl QasmCompiler {
         let target_profile = self.pragma_config.pragmas
             .iter()
             .find_map(|(kind, value)|
-                if matches!(kind, PragmaKind::QdkProfile) {
+                if matches!(kind, PragmaKind::QdkQirProfile) {
                     Some(Profile::from_str(value).expect("Invalid profile pragma; only a valid profile should be store in pragma_config."))
                 } else {
                     None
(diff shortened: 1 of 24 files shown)


source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -114,6 +114,7 @@ pub fn compile_to_qsharp_ast_with_config(
 pub enum PragmaKind {
     QdkBoxOpen,
     QdkBoxClose,
+    QdkProfile,
```
idavis: To match others, should be QdkQirProfile

ScottCarda-MS: Done

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -184,14 +186,31 @@ impl QasmCompiler {
...
+        let target_profile = self.pragma_config.pragmas
+            .iter()
+            .find_map(|(kind, value)|
+                if matches!(kind, PragmaKind::QdkQirProfile) {
+                    Some(Profile::from_str(value).expect("Invalid profile pragma; only a valid profile should be store in pragma_config."))
+                } else {
+                    None
+                }
+            )
+            .unwrap_or(Profile::Unrestricted);
```
idavis: Refactor into a function call. Expect message also needs to be updated.

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -1114,6 +1133,20 @@ impl QasmCompiler {
...
+                    return;
+                }
+                self.push_compiler_error(CompilerErrorKind::InvalidProfilePragmaTarget(
+                    profile.to_string(),
+                    stmt.value_span.unwrap_or(stmt.span),
+                ));
+            }
+            (PragmaKind::QdkQirProfile, None) => {
+                self.push_compiler_error(CompilerErrorKind::MissingProfilePragmaTarget(stmt.span));
+            }
```
idavis: Are these really different cases? When a user is typing they will get one message for a moment, then the other. I think missing value is just a subcase of invalid value.

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
```diff
@@ -0,0 +1,291 @@
...
+
+use crate::tests::{compile, compile_qasm_to_qir};
+use expect_test::expect;
+use miette::Report;
+
+#[test]
+fn profile_pragma_compiles_with_adaptive() -> miette::Result<(), Vec<Report>> {
+    let source = r#"
+        include "stdgates.inc";
+        #pragma qdk.qir.profile Adaptive_RI
```
orpuente-MS: Can we have a test like this one for `Adaptive_RIF`?

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
```diff
@@ -0,0 +1,291 @@
...
+        ctrl(5) @ x qs[0], qs[1], qs[2], qs[3], qs[4], aux;
+
+        // Measure the qubits
+        results = measure qs;
+    "#;
+
+    let unit = compile(source)?;
+
+    assert!(unit.has_errors(), "Expected compilation to fail");
+
```
idavis: I'd suggest updating theses tests to show the the error display so that the full error message and selected span for the error is saved in the test.

source/compiler/qsc_qasm/src/tests/statement/gate_call.rs
```diff
@@ -232,6 +231,7 @@ fn barrier_can_be_called_without_qubits() -> miette::Result<(), Vec<Report>> {
 fn barrier_generates_qir() -> miette::Result<(), Vec<Report>> {
     let source = r#"
         include "stdgates.inc";
+        #pragma qdk.qir.profile Adaptive_RI
         bit[1] c;
         qubit[2] q;
         barrier q[0], q[1];
```
idavis, line 237: Does the profile support any casing or does it have to be exactly cased like all of these samples? The error message shows the values lower case.

source/compiler/qsc_qasm/src/compiler.rs
ScottCarda-MS: I'm not sure what the expect message should be if not this. It seems like the correct expect message. An invalid profile should not make it this far into the compilation, so we shouldn't find an invalid profile in the `pragma_config.pragmas`.

source/compiler/qsc_qasm/src/tests/statement/gate_call.rs
ScottCarda-MS: Any casing should work, though the error message should be in CamelCase (with the underscore) to be consistent with references to profile names elsewhere.

source/compiler/qsc_qasm/src/compiler.rs
ScottCarda-MS: They definitely could be merged, but I was trying to follow the pattern for the box pragma. I can see a notable difference between the two different pragmas though and why it makes more sense to have a separate error for invalid vs missing for box pragma, but not for profile pragma. I'll merge the two errors together.

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
ScottCarda-MS: Hmm, I can convert the error to a string with to_string() to show the error message, but I'm not sure how to get access to the code snippet display programmatically.

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
ScottCarda-MS: I wasn't really sure how best to write tests for this feature. The solution here is that, specifically for the Adaptive_RI profile, the generated QIR is significantly different that other profiles, which shows that it is correctly using that profile. I wouldn't get the same kind of test with a different profile. I'm open to ideas for how to tests this better.

source/compiler/qsc_qasm/src/tests/statement/gate_call.rs
ScottCarda-MS: The casing in the error message is fixed now.

source/compiler/qsc_qasm/src/compiler.rs
ScottCarda-MS: I've merged the two error kinds and tried to rephrase the error message to work better for both cases.

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
orpuente-MS: You can use the `check_qasm_to_qsharp` function. It takes the source directly as input, so you don't even need to call `compile(source)`.

orpuente-MS: Also, can we suffix the test names that are expected to throw an error with `_errors`? For example, this one would become `invalid_profile_target_errors`.

ScottCarda-MS: That worked great, thanks! I've also added those `_error` suffixes.

source/compiler/qsc_qasm/src/compiler.rs
ScottCarda-MS: I've factored this out into its own function, but left the expect message. I'll close the convo for now, but feel free to reopen if you have further suggestions on that.

source/compiler/qsc_qasm/src/tests/statement/pragma/profile.rs
ScottCarda-MS: I've added the Adaptive_RIF test. It is very similar to the Adaptive_RI test, but there is a difference in the generated QIR, so it works as a test.

idavis (approved):

PR #2615, opened 2025-07-18
Add noise telemetry to `qsharp.run`

swernli:
This adds two Booleans to show whether `qsharp.run` or `qsharp.openqasm.run` were invoked with noise and/or qubit loss. The bools show up in the custom data along with the bucketized shot count.

### DIFF
--- a/source/pip/qsharp/_qsharp.py
+++ b/source/pip/qsharp/_qsharp.py
@@ -490,7 +490,9 @@ def run(
     if shots < 1:
         raise ValueError("The number of shots must be greater than 0.")
 
-    telemetry_events.on_run(shots)
+    telemetry_events.on_run(
+        shots, noise=noise is not None, qubit_loss=qubit_loss is not None
+    )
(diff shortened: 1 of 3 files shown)


source/pip/qsharp/_qsharp.py
```diff
@@ -490,7 +490,9 @@ def run(
     if shots < 1:
         raise ValueError("The number of shots must be greater than 0.")
 
-    telemetry_events.on_run(shots)
+    telemetry_events.on_run(
+        shots, noise=noise is not None, qubit_loss=qubit_loss is not None
```
DmitryVasilevsky: What if it's a noiseless noise?

DmitryVasilevsky (approved):

source/pip/qsharp/_qsharp.py
swernli: Good point... rather than lump that in as "noise usage" I'll differentiate here.

PR #2625, opened 2025-07-24
Better error messages for expected user scenarios

idavis:
This PR looks at expected user scenarios where they've done something incorrect, but we have enough information to help guide them.

### DIFF
--- a/source/compiler/qsc_qasm/src/semantic/error.rs
+++ b/source/compiler/qsc_qasm/src/semantic/error.rs
@@ -116,19 +116,21 @@ pub enum SemanticErrorKind {
     #[error("for statements must have a body or statement")]
     #[diagnostic(code("Qasm.Lowerer.ForStatementsMustHaveABodyOrStatement"))]
     ForStatementsMustHaveABodyOrStatement(#[label] Span),
-    #[error("Function called like gate: {0}")]
+    #[error("function called like gate: {0}")]
     #[diagnostic(help("function parameters must be in parentheses"))]
     #[diagnostic(code("Qasm.Lowerer.FuncCalledLikeGate"))]
     FuncCalledLikeGate(String, #[label] Span),
-    #[error("Function call missing parameters: {0}")]
-    #[diagnostic(help("function parameters must be in parentheses"))]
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_qasm/src/semantic/error.rs
```diff
@@ -116,6 +116,24 @@ pub enum SemanticErrorKind {
     #[error("for statements must have a body or statement")]
     #[diagnostic(code("Qasm.Lowerer.ForStatementsMustHaveABodyOrStatement"))]
     ForStatementsMustHaveABodyOrStatement(#[label] Span),
+    #[error("Function called like gate: {0}")]
+    #[diagnostic(help("function parameters must be in parentheses"))]
+    #[diagnostic(code("Qasm.Lowerer.FuncCalledLikeGate"))]
+    FuncCalledLikeGate(String, #[label] Span),
+    #[error("Function call missing parameters: {0}")]
+    #[diagnostic(help("function parameters must be in parentheses"))]
```
swernli: nitpick: I wonder if there is a slightly different phrasing here that clarifies parenthesis are required even when there are no arguments? Maybe something like, "a function call must use parenthesis, with any parameters inside those parenthsis." That feels verbose, but imagining an LLM responding to the help text, I can't help but feel like more specific (and different than the previous error) could be better.

source/compiler/qsc_qasm/src/semantic/error.rs
```diff
@@ -116,6 +116,24 @@ pub enum SemanticErrorKind {
     #[error("for statements must have a body or statement")]
     #[diagnostic(code("Qasm.Lowerer.ForStatementsMustHaveABodyOrStatement"))]
     ForStatementsMustHaveABodyOrStatement(#[label] Span),
+    #[error("Function called like gate: {0}")]
```
swernli: The new errors start with capital letters, but our convention is to use lower case (since they can appear in the middle of a longer message).

swernli (approved):
Two minor things on the error phrasing, otherwise the implementation looks great.

PR #2635, opened 2025-07-29
Python - Q# UDT interop

orpuente-MS:
Currently we support sending primitive types, arrays, and tuples between Python and Q#. This PR expands those capabilities to allow sending structs across the interop layer.

## What it looks like

```python
import qsharp

qsharp.eval("""
    struct Data { a : Int, b : Int }

    function SwapData(data : Data) : Data {
        new Data { a = data.b, b = data.a }
    }
""")

# You can import your custom structs defined in Q# from Python.
from qsharp.code import Data, SwapData
SwapData(Data(2, 3))

# You can pass a dict that matches the same field values and types as the expected struct
SwapData({ 'a': 2, 'b': 3 })

# You can also create a custom class in a number of ways
# A regular Python class
class MyData:
    def __init__(self, a, b):
        self.a = a
        self.b = b
SwapData(MyData(2, 3))

# Using dataclasses
from dataclasses import dataclass
@dataclass
class MyData:
    a: int
    b: int
SwapData(MyData(2, 3))

# Class with slots
class MyData:
    __slots__ = ['a', 'b']
    def __init__(self, a, b):
        self.a = a
        self.b = b
SwapData(MyData(2, 3))

# Class with slots and dynamic values
class MyData:
    __slots__ = ['a', '__dict__']
    def __init__(self, a):
        self.a = a
data = MyData(2)
data.b = 3
SwapData(data)
```

As a special case of this functionality, we now support sending complex numbers between Python and Q#.

```python
import qsharp

qsharp.eval("""
    import Std.Math.Complex;

    function SwapComponents(c: Complex) : Complex {
        new Complex { Real = c.Imag, Imag = c.Real }
    }
    """)

from qsharp.code import SwapComponents
assert SwapComponents(2 + 3j) == 3 + 2j
```


source/compiler/qsc/src/interpret.rs
```diff
@@ -389,6 +393,71 @@ impl Interpreter {
         }
     }
 
+    /// Given a package ID, returns all the types in the package.
+    /// Note this does not currently include re-exports.
+    #[allow(clippy::type_complexity)]
```
minestarks: I agree with clippy ;) A struct with named fields for the return value would go a long way towards making this readable. (And yes I see you had to do this to match existing code, sorry!)

source/compiler/qsc_hir/src/global.rs
```diff
@@ -189,6 +190,27 @@ impl PackageIter<'_> {
                     kind: Kind::Ty(Ty { id }),
                 })
             }
+            (ItemKind::Ty(name, def), None) => {
+                self.next = Some(Global {
```
minestarks: Just curious if this is causing any trouble - since the same UDT definition is returned twice, once as a "term" and once as a "ty", is the existing code that finds global callables now "seeing" this as a callable?

(Note in my other PR I'm changing this _not_ to return the same UDT as a "term"!. So let's verify that the changes make sense together)

source/compiler/qsc/src/interpret.rs
```diff
@@ -389,6 +393,71 @@ impl Interpreter {
...
+            &self.compiler.package_id()
+        };
+
+        let unit = self
+            .compiler
+            .package_store()
+            .get(*package_id)
+            .expect("package should exist in the package store");
+
+        let item = unit.package.items.get(*local_item_id)?;
```
swernli: When would this return `None`? How could the item be missing from the package it's supposed to be in? Just wondering if it is possible this could be treated as a panic and the signature simplified.

source/compiler/qsc/src/interpret.rs
```diff
@@ -488,9 +557,23 @@ impl Interpreter {
             .map_err(into_errors)?;
         // Clear the entry expression, as we are evaluating fragments and a fragment with a `@EntryPoint` attribute
         // should not change what gets executed.
+
```
swernli: nitpick: this newline doesn't really make sense, as the comment applies to the line below. If anything, the newline should be before the comment.

source/pip/src/interpreter.rs
```diff
@@ -613,7 +655,13 @@ impl Interpreter {
         };
 
         match result {
-            Ok(value) => Ok(ValueWrapper(value).into_pyobject(py)?.unbind()),
+            Ok(value) => {
+                if let Some(ty) = result_ty {
+                    typed_value_to_python_obj(&self.interpreter, py, &value, &ty)
+                } else {
+                    Ok(ValueWrapper(value).into_pyobject(py)?.unbind())
```
swernli: Hmm... I don't think this will work as expected if the expression string passed to `qsharp.run` includes a definition of a struct inside of it. For example:
```python
qsharp.run("{ struct Foo { Res : Result}; use q = Qubit(); H(q); new Foo { Res = MResetZ(q) } }", shots = 10)
```
That said, perhaps this is a corner case we don't want to be concerned with too much? In Q# these structs would be locally defined and not usable outside of the scope they were defined in.

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+        let qsc::hir::ty::UdtDefKind::Field(real) = &fields[0].kind else {
+            return false;
+        };
+        let qsc::hir::ty::UdtDefKind::Field(imag) = &fields[1].kind else {
+            return false;
+        };
+        return matches!(real.ty, Ty::Prim(Prim::Double))
+            && matches!(imag.ty, Ty::Prim(Prim::Double))
+            && &*udt.name == "Complex";
```
swernli: Since the `UdtField` also have names, is it worth an extra layer of protection to match on them too?

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,432 @@
...
+{
+    match &udt_def.kind {
+        qsc::hir::ty::UdtDefKind::Field(udt_field) => {
+            if let Some(name) = udt_field.name.as_ref() {
+                buffer.push((name.clone(), &udt_field.ty));
+                Ok(())
+            } else {
+                Err(PyTypeError::new_err(format!(
+                    "structs with anonymous fields are not supported: {udt_name}"
+                )))
```
swernli, line 173: What would hit this error case? Would it happen if someone used the older UDT syntax instead of the Struct syntax and tried to return something of that type? If so, is it worth a test case confirming that triggers this expected failure case?

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+        Ty::Udt(_, res) => {
+            let qsc::hir::Res::Item(item_id) = res else {
+                panic!("Udt should be an item");
+            };
+            let Some(udt) = ctx.udt_ty(item_id) else {
+                unreachable!(
+                    "we verified that the udt is defined in `first_unsupported_interop_ty`"
+                );
+            };
```
swernli: I think these two should be graceful errors rather than panics for a specific corner case. Someone can run some code that defines UDTs, get back objects with IR corresponding to those UDTs, then call `qsharp.init` which resets the compilation. Now the item IDs cached in the IR might not correspond to any entries in the compilation or might happen to correspond to different entries in the compilation. It's definitely a failure case, but it shouldn't crash.

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,432 @@
...
+
+            // Handle `Complex` special case.
+            if let Ok(v) = obj.extract::<num_complex::Complex64>(py) {
+                if is_complex_udt(udt) {
+                    let tuple = Value::Tuple(Rc::new([Value::Double(v.re), Value::Double(v.im)]));
+                    return Ok(tuple);
+                }
+            }
```
swernli: Is it possible that pyO3's attempts at type coercion may trigger this unexpectedly? I don't know how many other types coerce into `Complex64`, so it's probably fine, just thinking through edge cases.

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+        Ty::Udt(name, res) => {
+            let qsc::hir::Res::Item(item_id) = res else {
+                panic!("Udt should be an item");
+            };
+            let Some(udt) = ctx.udt_ty(item_id) else {
+                unreachable!(
+                    "we verified that the udt is defined in `first_unsupported_interop_ty`"
+                );
+            };
```
swernli: Here the panic should be fine... these types came from inside the evaluator, so they should always line up.

swernli: (feel free to resolve)

source/pip/qsharp/_qsharp.py
```diff
@@ -20,6 +26,7 @@
     TypedDict,
     Union,
     List,
+    Self,
     overload,
```
swernli: What is `overload` used for here?

source/pip/qsharp/_qsharp.py
```diff
@@ -28,6 +35,66 @@
 import sys
 import types
 from time import monotonic
+from dataclasses import make_dataclass, field
```
swernli: Is `field` here used?

source/pip/tests/test_qasm.py
```diff
@@ -543,24 +543,6 @@ def test_callables_with_unsupported_types_raise_errors_on_call() -> None:
         code.Unsupported()
 
 
-def test_callables_with_unsupported_udt_types_raise_errors_on_call() -> None:
```
swernli, line 546: Now that these don't trigger unsupported errors, should we have positive tests for these? Specifically, something that takes and/or returns the OpenQASM `complex` works as expected.

swernli (commented):
Changes are looking good, a few comments and corner cases that are worth exploring before sign off!

source/pip/qsharp/_qsharp.py
orpuente-MS: Not anymore, nice catch.

source/pip/qsharp/_qsharp.py
orpuente-MS: I think it's dead code. I will remove it.

source/pip/src/interpreter/data_interop.rs
orpuente-MS: I think that if PyO3 can coerce it to a complex number, then it is the expected behavior for a Python user. For example, if I were a Python user and I were using a function taking a complex number, I would expect that passing it an int or a float would also work.

source/pip/tests/test_qasm.py
idavis: Agreed, we should have positive tests for this now.

source/pip/qsharp/_qsharp.py
```diff
@@ -28,6 +35,66 @@
...
+            if name == "__dict__":
+                for name, val in obj.__dict__.items():
+                    fields[name] = lower_python_obj(val, visited)
+            else:
+                val = getattr(obj, name)
+                fields[name] = lower_python_obj(val, visited)
+    elif hasattr(obj, "__dict__"):
+        fields = {
+            name: lower_python_obj(val, visited) for name, val in obj.__dict__.items()
+        }
```
idavis, line 77: Should we raise en error case it an object has `__slots__` and `__dict__` or has a base class that isn't object?

source/pip/qsharp/_qsharp.py
```diff
@@ -435,12 +506,10 @@ def callback(output: Output) -> None:
                     pass
             print(output, flush=True)
 
-        if len(args) == 1:
-            args = args[0]
-        elif len(args) == 0:
-            args = None
+        args = python_args_to_interpreter_args(args)
```
idavis, line 507: All of the uses don't map to the new calls semantics when args > 1. Is there any logic that was making assumptions when args > 1?

source/pip/src/interpreter.rs
```diff
@@ -821,64 +876,29 @@ fn first_unsupported_interop_ty(ty: &Ty) -> Option<&Ty> {
...
-        Ty::Array(ty) => {
-            let obj = obj.extract::<Vec<PyObject>>(py)?;
-            let mut values = Vec::with_capacity(obj.len());
-            for item in &obj {
-                values.push(convert_obj_with_ty(py, item, ty)?);
-            }
-            Ok(Value::Array(values.into()))
-        }
-        _ => unimplemented!("input type: {ty}"),
+        _ => Some(ty),
```
idavis: Does it make sense to explicitly add the cases that result in unsupported tys here instead of `_`?

source/pip/tests/test_interpreter.py
```diff
@@ -178,6 +185,105 @@ def test_value_array() -> None:
     assert value == [1, 2, 3]
 
 
+def test_value_udt() -> None:
```
idavis, line 235: Missing these kinds of tests for invoke, run, circuit, qir, etc.

source/pip/src/interpreter.rs
orpuente-MS: Solved in [this commit](https://github.com/microsoft/qsharp/pull/2635/commits/28a971e03ce0a39f76a5a33276e7cb4a73043b3b).

source/compiler/qsc/src/interpret.rs
orpuente-MS: Yes, simplifying the signature also simplified the code calling this function. Nice catch.

source/pip/src/interpreter/data_interop.rs
orpuente-MS: Since we do duck typing and don't store the item ID when building the Python types, we never hit this scenario.

source/pip/qsharp/_qsharp.py
orpuente-MS: We can have objects with `__slots__` and `__dict__`, it's a special case of the `__slots__` class where you can also have dynamic values.

For the base class not being object, I think everything in python is an instance of `object`, at least AFAIK.

source/pip/qsharp/_qsharp.py
orpuente-MS: Oh, It's the same mapping semantics, the only difference is that now we lower all the python objects.

Before, all of the uses of this logic would:
1. Map an empty args list to None.
2. Map a list with one element to the element (i.e. unwrapping the element).
3. Leave a list with more than one element untouched.

Now, we need to keep a similar behavior but lower all python objects with `lower_python_obj`. So we:
1. Map an empty args list to None.
2. Map a list with one element to `lower_python_obj(element)`
3. Map a list with more than one element to another list but applying `lower_python_obj` to all elements.

source/compiler/qsc/src/interpret.rs
```diff
@@ -389,6 +397,73 @@ impl Interpreter {
...
+            if let global::Kind::Ty(ty) = global.kind {
+                exported_items.push((global.namespace, global.name, ty.id));
+            }
+        }
+        exported_items
+    }
+
+    #[allow(clippy::type_complexity)]
+    pub fn user_types(&self) -> Vec<(Vec<Rc<str>>, Rc<str>, qsc_hir::hir::ItemId)> {
+        self.package_types(self.source_package)
```
minestarks, line 445: Did the names get flipped here? What's typically called the "source package" in the interpreter is the package indicated by `source_package` . And the "current package" is the one indicated by `package`. Elsewhere, the "current package" is also called the "user package".

source/pip/src/interpreter/data_interop.rs
orpuente-MS: Added in [this commit](https://github.com/microsoft/qsharp/pull/2635/commits/79a5100641d457c642422598a13df879c7541a46).

source/pip/tests/test_qasm.py
orpuente-MS: Added the positive tests for `complex`. Still missing the handling of `angle` in the "right way", in which we cast Python floats to Q# `Std.OpenQASM.Angle.Angle` and vice versa.

source/pip/src/interpreter.rs
orpuente-MS: Yes, that should protect us from accidently missing this spot in the future if we add a new type. Changed.

minestarks (commented):
I haven't reviewed everything, but posting these outstanding comments before they gate stale.

Besides that, our side conversation about pushing  `data_interop` deeper into the `Interpreter`, so even thing like the debugger could take advantage of the rich type information, still feels like a good idea to me

source/compiler/qsc_hir/src/global.rs
orpuente-MS: I undid my changes in this file and just merged main after your PR, and it all works fine!

source/compiler/qsc_hir/src/global.rs
```diff
@@ -118,6 +118,7 @@ pub struct PackageIter<'a> {
 }
 
 impl PackageIter<'_> {
+    #[allow(clippy::too_many_lines)]
     fn global_item(&mut self, item: &Item) -> Option<Global> {
```
swernli: Looks like this `allow` got left behind but isn't really needed anymore.

source/compiler/qsc/src/interpret.rs
orpuente-MS: Probably. I copied the naming from `user_globals` and `source_globals` defined a few lines above, but it did seem a bit odd to me. I can fix it there too.

source/pip/tests/test_qasm.py
```diff
@@ -543,22 +543,36 @@ def test_callables_with_unsupported_types_raise_errors_on_call() -> None:
...
+    code.ComplexInput(3.14)
+
+
+def test_callable_with_angle_output() -> None:
+    init()
+    import_openqasm(
+        "def ComplexOutput() -> angle { return pi; }",
+        program_type=ProgramType.Fragments,
+    )
+    assert code.ComplexOutput() == pi
```
swernli: nitpick: the functions defined are for angles, not complex

```suggestion
def test_callable_with_angle_input() -> None:
    init()
    import_openqasm("def AngleInput(angle a) { }", program_type=ProgramType.Fragments)
    code.AngleInput(3.14)


def test_callable_with_angle_output() -> None:
    init()
    import_openqasm(
        "def AngleOutput() -> angle { return pi; }",
        program_type=ProgramType.Fragments,
    )
    assert code.AngleOutput() == pi
```

source/pip/src/interpreter/data_interop.rs
```diff
@@ -0,0 +1,426 @@
...
+            match kind {
+                interpret::UdtKind::Angle => {
+                    let value = values[0].clone().unwrap_int();
+                    let size = values[1].clone().unwrap_int();
+                    let value = u64::try_from(value).expect("value should fit in u64");
+                    let size = u32::try_from(size).expect("size should fit in u32");
+                    let angle = qsc::qasm::stdlib::angle::Angle::new(value, size);
+                    let angle: f64 = angle
+                        .try_into()
+                        .map_err(|_| QasmError::new_err("cannot cast angle to 64-bit float"))?;
```
swernli: nitpick: "cannot" sounds like it's not allowed, but here I think the error is that we are failing to do it.

```suggestion
                        .map_err(|_| QasmError::new_err("failed to cast angle to 64-bit float"))?;
```

swernli (approved):

source/pip/src/interpreter/data_interop.rs
orpuente-MS: Yes, that wording was definitely misleading. Thanks!

source/pip/tests/test_interpreter.py
orpuente-MS: Added in [this commit](https://github.com/microsoft/qsharp/pull/2635/commits/8cb9a05a9c3922ea472bc3e3e135ff53e2c25353).

source/pip/tests/test_interpreter.py
```diff
@@ -178,6 +232,169 @@ def test_value_array() -> None:
...
+    callable = "function makeData() : Data { new Data { a = 2, b = 3 } }"
+    output = "Data(a=2, b=3)"
+
+    check_interpret(source, output)
+    check_run(expr, output)
+    check_invoke(source, callable, output)
+    check_circuit(expr, "")
+    check_estimate(expr)
+    # This test triggers an existing panic in Partial Evaluation
+    # check_qir(expr, "")
```
orpuente-MS: The new QIR tests for UDTs trigger a panic in Partial Evaluation. A few of extra details:
1. The panic happens when returning an UDT from the entry expression.
2. It only happens when triggering QIR gen from Python, it doesn't happen in VS-Code or the Playground
3. I double checked, and the panic is present in `main`, it wasn't introduced in this PR.

minestarks:
@orpuente-MS  Do we need an extra step to handle notebook cell outputs? I would expect the second and third cells in the below screenshot  to have the same output. 

image

minestarks:
@orpuente-MS  while we're looking at interop, I wonder if we should address this (preexisting) case where we inconsistently treat `Qubit` as an "unsupported" input/output type, depending on how you call the function. Can the data interop layer "reject" this output and just not return it?

```python
import qsharp
qsharp.eval("""
operation allocate() : Qubit {
    use q = Qubit();
    return q;
}
""")
qsharp.code.allocate() // complains about unsupported output type :)
qsharp.eval("allocate()") // returns a useless released qubit object :/
```

image

minestarks:
@orpuente-MS I also wonder if we could provide a more helpful error message (similar to what the Q# compiler gives when you pass the wrong arguments to a function)

```python
qsharp.eval("""
struct Data { a : Int, b : Int }

function takeData(d : Data) : Int {
    return d.a + d.b
}

""")

qsharp.eval("takeData(1)") # nice looking exception ("expected Data, found Int")

qsharp.code.takeData(1) # less helpful exception ("TypeError: 'int' object cannot be converted to 'PyDict'")
```

minestarks (approved):

source/pip/tests/test_qasm.py
```diff
@@ -543,22 +543,36 @@ def test_callables_with_unsupported_types_raise_errors_on_call() -> None:
...
     init()
     import_openqasm(
-        "def Unsupported() -> complex { end; }", program_type=ProgramType.Fragments
+        "def ComplexOutput() -> complex { return 2 + 3im; }",
+        program_type=ProgramType.Fragments,
     )
-    with pytest.raises(QSharpError, match='unsupported output type: `UDT<"Complex":'):
-        code.Unsupported()
+    assert code.ComplexOutput() == 2 + 3j
+
```
idavis, line 561:
```suggestion

def test_callable_with_complex_input_output() -> None:
    init()
    import_openqasm(
        "def ComplexOutput(complex a) -> complex { return 2 * a; }",
        program_type=ProgramType.Fragments,
    )
    assert code.ComplexOutput(2 + 3j) == 4 + 6j

```

source/pip/src/interpreter.rs
```diff
@@ -63,6 +69,12 @@ fn verify_classes_are_sendable() {
     is_send::<Output>();
     is_send::<StateDumpData>();
     is_send::<Circuit>();
+    is_send::<UdtValue>();
+    is_send::<UdtFields>();
```
idavis, line 73: Why does this need to be sendable but not exported in the module?

source/pip/src/interpreter.rs
```diff
@@ -1041,6 +1062,12 @@ pub(crate) enum Result {
     Loss,
 }
 
+impl From<Result> for qsc::interpret::Result {
+    fn from(value: Result) -> Self {
+        qsc::interpret::Result::Val(value == Result::One)
```
idavis: @swernli - do we really want to lose `Loss` here?

swernli: That's a good question. This is the code that handles values coming back into Q# from Python, so while `Loss` can be returned up just fine passing `Loss` back into Q# would be interpreted as a `Zero`. While I doubt there is much need for this kind of round-tripping for `Loss` values, for correctness and to avoid confusion in the corner case where someone does round-trip them this should be a match:

```suggestion
        match value {
            Result::Loss => qsc::interpret::Result::Loss,
            Result::One | Result::Zero => qsc::interpret::Result::Val(value == Result::One)
        }
```

source/pip/src/interpreter.rs
orpuente-MS: Because we send `UdtValue` from rust to Python as the output of `interpret`, `run`, etc., but we never need to do `_native.UdtValue(...)` to construct one directly.

source/pip/qsharp/_qsharp.py
```diff
@@ -198,11 +263,13 @@ def init(
...
+    # Q# callables or structs. This is necessary to avoid conflicts with the new interpreter instance.
     keys_to_remove = []
     for key in code.__dict__:
-        if hasattr(code.__dict__[key], "__global_callable") or isinstance(
-            code.__dict__[key], types.ModuleType
+        if (
+            hasattr(code.__dict__[key], "__global_callable")
+            or hasattr(code.__dict__[key], "__qsharp_class")
+            or isinstance(code.__dict__[key], types.ModuleType)
         ):
```
idavis:
```suggestion
    for (key, val) in code.__dict__.items():
        if (
            hasattr(val, "__global_callable")
            or hasattr(val, "__qsharp_class")
            or isinstance(val, types.ModuleType)
        ):
```

source/pip/qsharp/_native.pyi
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 from enum import Enum
-from typing import Any, Callable, Optional, Dict, List, Tuple
+from typing import Any, Callable, Optional, Dict, List, Tuple, Self
```
idavis: Is this used?

orpuente-MS: Nope. Removed.

idavis (approved):

PR #2636, opened 2025-07-29
Profile Selection Redesign

ScottCarda-MS:
Reintroduces the profile selection feature with an updated design. This PR:
- Removes global setting for target profile.
- Removes the status bar selection for target profile.
- Removes the Playground selection for target profile.
- Adds an optional field to qsharp.json, "target_profile" where the target profile can be set for Q# projects.
- Adds an optional argument to the EntryPoint attribute where the target profile can be set for non-project Q# files.
- Adds an optional pragma to set target profile for QASM files.
- Adds an error that will highlight the EntryPoint attribute if one occurs in a Q# project.
- Adds logic to assume the target profile based on the target's capabilities when no target profile is specified.


source/vscode/src/projectSystem.ts
```diff
@@ -225,6 +225,11 @@ async function singleFileProject(
   documentUri: vscode.Uri,
 ): Promise<IProjectConfig> {
   const file = await vscode.workspace.openTextDocument(documentUri);
+  // ToDo: the `getEntryProfile` doesn't acutally use the project loader object. Can I call it without it?
```
ScottCarda-MS: It would be great if we didn't have to have the projectLoader loaded in order to use this function, as it doesn't really make use of it anyways. But I'm not sure right now how to get it to be exposed from the npm to the vscode without attaching it to an object like this.

source/vscode/src/qirGeneration.ts
```diff
@@ -52,25 +57,82 @@ export async function getQirForActiveWindow(
...
   targetSupportsAdaptive?: boolean,
+  documentUri?: vscode.Uri,
 ): Promise<string> {
   let result = "";
   const isLocalQirGeneration = targetSupportsAdaptive === undefined;
-  const targetProfile = config.profile;
-  const isUnrestricted = targetProfile === "unrestricted";
+  const hasManifest = config.packageGraphSources.hasManifest;
+  if (!hasManifest) {
+    // ToDo: this is a lot of weird work to do just to get the profile,
```
ScottCarda-MS: This might be too big of an ask, but it is really awkward that there are two separate code paths that seem like they do the same thing (get a profile from source), one for the language service and another for qir generation. It would be much better if we could consolidate those paths into one.

source/vscode/src/projectSystem.ts
ScottCarda-MS: I changed the way this is exposed through the wasm/npm/vscode layers. I'm not super confident it is the correct way to do it, and would appreciate someone double-checking this.

source/vscode/src/circuit.ts
```diff
@@ -144,6 +144,11 @@ export async function generateCircuit(
   extensionUri: Uri,
   params: CircuitParams,
 ): Promise<CircuitOrError> {
+  // Make sure target profile is set
+  if (!params.program.profile) {
+    params.program.profile = "unrestricted";
+  }
+
```
minestarks: This shouldn't be needed as our other changes should have guaranteed that `FullProgramConfig` always has a valid profile set.

source/compiler/qsc/src/interpret.rs
```diff
@@ -99,6 +99,9 @@ pub enum Error {
     NoEntryPoint,
     #[error("unsupported runtime capabilities for code generation")]
     #[diagnostic(code("Qsc.Interpret.UnsupportedRuntimeCapabilities"))]
+    #[diagnostic(help(
+        "@EntryPoint attribute argument should be 'Base', 'Adaptive_RI' or 'Adaptive_RIF'"
+    ))]
```
minestarks: Is this help text seen anywhere now? We should have made enough changes that we're not even calling into the compiler with an unsupported target profile anymore (and bailing out with a user-friendly error instead). Maybe we can revert this.

source/samples_test/src/tests.rs
```diff
@@ -110,7 +110,7 @@ fn compile_and_run_qasm_internal(source: &str, debug: bool) -> String {
         Option::<&mut InMemorySourceResolver>::None,
         config,
     );
-    let (source_map, errors, package, sig) = unit.into_tuple();
+    let (source_map, errors, package, sig, _) = unit.into_tuple();
```
minestarks: We should probably pass this into the compiler below.

source/vscode/src/language-service/activate.ts
```diff
@@ -327,24 +327,12 @@ function registerConfigurationChangeHandlers(
 async function updateLanguageServiceConfiguration(
```
minestarks, line 324: Up above, you need to also update the listeners that kick this if the targetProfile setting has changed.

source/vscode/src/qirGeneration.ts
ScottCarda-MS: resolved

source/vscode/src/qirGeneration.ts
```diff
@@ -167,12 +176,12 @@
     );
     clearTimeout(compilerTimeout);
   } catch (e: any) {
-    log.error("Codegen error. ", e.toString());
     if (e.toString() === "terminated") {
       throw new QirGenerationError(
         "QIR generation was cancelled or timed out.",
       );
     } else {
+      // TODO: probably this message should be updated
```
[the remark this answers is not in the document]
ScottCarda-MS: resolved

source/vscode/src/qirGeneration.ts
```diff
@@ -32,94 +34,100 @@
...
-    if (result?.action !== "set") {
-      throw new QirGenerationError(
-        error_msg +
-          " Please update the QIR target via the status bar selector or extension settings.",
-      );
-    } else {
-      await setTarget(targetSupportsAdaptive ? "adaptive_ri" : "base");
+  const compatible = checkCompatibility(config.profile, preferredTargetProfile);
+  if (!compatible) {
+    // TODO: this error message could be made more helpful by checking `config.packageGraphSources.hasManifest`
```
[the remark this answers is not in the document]
ScottCarda-MS: resolved

source/vscode/src/circuit.ts
ScottCarda-MS: removed

source/vscode/src/language-service/activate.ts
ScottCarda-MS: updated

source/samples_test/src/tests.rs
ScottCarda-MS: done

source/compiler/qsc/src/interpret.rs
ScottCarda-MS: removed

source/compiler/qsc/src/compile.rs
```diff
@@ -45,6 +47,11 @@ pub enum ErrorKind {
     /// `OpenQASM` compilation errors.
     #[diagnostic(transparent)]
     OpenQasm(#[from] crate::qasm::error::Error),
+
+    #[error(
+        "The @EntryPoint attribute with a profile argument is not allowed in a Q# project (with qsharp.json). Please specify the profile in qsharp.json instead."
+    )]
+    EntryPointProfileInProject(#[label] Span),
```
minestarks, line 54: Please add a unit test demonstrating where this is raised. the `packages` unit tests seem like a good spot?

source/compiler/qsc/src/packages.rs
```diff
@@ -75,19 +76,70 @@ fn convert_circuit_sources(
...
+            let converted_source = converted_source_map.find_by_offset(span.hi);
+            let original_source =
+                converted_source.and_then(|s| original_source_map.find_by_name(&s.name));
+            if let (Some(converted), Some(original)) = (converted_source, original_source) {
+                // Adjust the span to account for the offset of the original source
+                span = span - converted.offset + original.offset;
+            }
+
+            dependency_errors.push(Error::from_map(
+                &SourceMap::new(package_graph_sources.root.sources.clone(), None),
```
minestarks: I think you should be able to just pass in `converted_source_map` here and avoid the re-conversion?

source/compiler/qsc/src/packages.rs
```diff
@@ -75,19 +76,70 @@ fn convert_circuit_sources(
...
 }
 
+#[must_use]
+pub fn get_target_profile_from_entry_point(
+    package_graph_sources: &PackageGraphSources,
+    dependency_errors: &mut Vec<WithSource<ErrorKind>>,
+) -> Option<Profile> {
+    // Convert circuit files in user code to generated Q# before entry profile check
+    let mut sources = package_graph_sources.root.sources.clone();
+    sources = convert_circuit_sources(sources, dependency_errors);
```
minestarks: Since you already call `convert_circuit_sources` in `prepare_package_store`, could you perhaps perform this check on those converted sources and raise the error there?

source/npm/qsharp/src/compiler/compiler.ts
```diff
@@ -188,6 +172,7 @@ export class Compiler implements ICompiler {
     // All results are communicated as events, but if there is a compiler error (e.g. an invalid
     // entry expression or similar), it may throw on run. The caller should expect this promise
     // may reject without all shots running or events firing.
+    log.info(`Running run with profile ${program.profile}`);
```
minestarks: Was this just for debugging? I worry they'll clutter up the Output UI (if you still need it, consider putting it at the `debug` level)

source/vscode/src/qirGeneration.ts
```diff
@@ -167,15 +184,13 @@ async function getQirForProgram(
...
-    log.error("Codegen error. ", e.toString());
     if (e.toString() === "terminated") {
       throw new QirGenerationError(
         "QIR generation was cancelled or timed out.",
       );
     } else {
       throw new QirGenerationError(
-        `QIR generation failed due to error: "${e.toString()}". Please ensure the code is compatible with a QIR profile ` +
-          "by setting the target QIR profile to 'base' or 'adaptive_ri' and fixing any errors.",
+        `QIR generation failed due to error: "${e.toString()}". Please ensure the code is compatible with the selected QIR profile.`,
```
minestarks: Does this look user-friendly in the UI or do want to further refine this messaging?

minestarks (approved):

source/wasm/src/project_system.rs
```diff
@@ -430,15 +445,14 @@ pub(crate) fn into_qsc_args(
...
-        .into();
-
     let pkg_graph: PackageGraphSources = program.packageGraphSources().into();
     let pkg_graph: qsc_project::PackageGraphSources = pkg_graph.into();
 
-    // this function call builds all dependencies as a part of preparing the package store
-    // for building the user code.
+    // Use the project-level target_profile from ProgramConfig
+    let profile = qsc::target::Profile::from_str(&program.profile())
+        .unwrap_or_else(|()| panic!("Invalid target : {}", program.profile()));
```
idavis, line 453: Why does this `panic` instead of returning `Err`?

source/wasm/src/project_system.rs
```diff
@@ -466,18 +480,15 @@ pub(crate) fn into_qsc_args(
...
 #[allow(clippy::type_complexity)]
-pub(crate) fn into_openqasm_args(
-    program: ProgramConfig,
-) -> (Vec<(Arc<str>, Arc<str>)>, qsc::TargetCapabilityFlags) {
-    let capabilities = qsc::target::Profile::from_str(&program.profile())
-        .unwrap_or_else(|()| panic!("Invalid target : {}", program.profile()))
-        .into();
+pub(crate) fn into_openqasm_arg(program: ProgramConfig) -> (Vec<(Arc<str>, Arc<str>)>, Profile) {
+    let profile = qsc::target::Profile::from_str(&program.profile())
+        .unwrap_or_else(|()| panic!("Invalid target : {}", program.profile()));
```
idavis: Related to the question above, we don't return result here, but the value isn't user input. We should probably do the same in both scenarios, whichever way we go.

source/wasm/src/lib.rs
```diff
@@ -672,21 +672,21 @@ pub fn generate_docs(additional_program: Option<ProgramConfig>) -> Vec<IDocFile>
...
-    get_configured_interpreter_from_openqasm(sources, capabilities, false)
+    get_configured_interpreter_from_openqasm(sources, profile, false)
 }
 
 fn get_configured_interpreter_from_openqasm(
     sources: &[(Arc<str>, Arc<str>)],
-    capabilities: TargetCapabilityFlags,
+    profile: Profile,
     dbg: bool,
 ) -> Result<(String, interpret::Interpreter), String> {
```
idavis: Why the change to profile from capablilties?

source/wasm/src/lib.rs
```diff
@@ -720,7 +719,7 @@ fn get_configured_interpreter_from_openqasm(
         dbg,
         store,
         source_package_id,
-        capabilities,
+        profile.into(),
```
idavis: See above, we always have to turn it into caps, so why take profile as an arg?

source/vscode/test/suites/language-service/notebook.test.ts
```diff
@@ -117,41 +116,4 @@ suite("Q# Notebook Tests", function suite() {
...
-    } finally {
-      // Restore the original workspace configuration using VS Code API
-      await config.update(
-        "qir.targetProfile",
-        originalTarget,
-        vscode.ConfigurationTarget.Global,
-      );
-    }
-  });
 });
```
idavis: Is this test case invalid now?

source/compiler/qsc/src/compile.rs
ScottCarda-MS: added

source/compiler/qsc/src/packages.rs
ScottCarda-MS: Yes, I was able to simplify the logic for handling circuit files by filtering them out of the parsing that looks for entry point target profiles.

source/compiler/qsc/src/packages.rs
ScottCarda-MS: I couldn't quite get the logic to a place where the error could be thrown where you suggested. The best I'm able to do is not do the circuit conversion when looking for target profile. Because circuit files won't have such a thing, I filtered them out of that search.

source/npm/qsharp/src/compiler/compiler.ts
ScottCarda-MS: removed

source/compiler/qsc_project/src/openqasm.rs
```diff
@@ -5,9 +5,10 @@
 mod integration_tests;
 
 use super::{FileSystemAsync, Project};
-use qsc_qasm::parser::ast::{Program, StmtKind};
+use qsc_data_structures::target::Profile;
+use qsc_qasm::parser::ast::{PathKind, Program, StmtKind};
 use rustc_hash::FxHashSet;
-use std::{path::Path, sync::Arc};
+use std::{path::Path, str::FromStr as _, sync::Arc};
```
idavis, line 11: what is ` str::FromStr as _, ` doing?

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -184,14 +186,37 @@ impl QasmCompiler {
...
+            .iter()
+            .find_map(|(kind, value)|
+                if matches!(kind, PragmaKind::QdkQirProfile) {
+                    Some(Profile::from_str(value).expect("Invalid profile pragma; only a valid profile should be store in pragma_config."))
+                } else {
+                    None
+                }
+            )
+            .unwrap_or(Profile::Unrestricted)
     }
```
idavis: Shouldn't iterate, this is a hash lookup.
```rust
self
    .pragma_config
    .pragmas
    .get(&PragmaKind::QdkQirProfile)
    .map(...)
    .unwrap_or(Profile::Unrestricted)
```

source/language_service/src/completion/ast_context.rs
```diff
@@ -31,7 +31,7 @@ enum Context<'a> {
         record: &'a Expr,
     },
     /// The cursor is on an attribute.
-    AttrArg,
+    AttrArg(String),
```
idavis: Should this be an `Rc<str>`?

source/wasm/src/project_system.rs
ScottCarda-MS: There shouldn't be a way for there to be an invalid profile in the program at this point. If this panic happens, it reflects something wrong with our code, not with something the user did. My understanding is that we only create and return formal errors for issues caused by the user doing something wrong, and that we use panics for issues caused by compiler logic.

source/wasm/src/project_system.rs
ScottCarda-MS: I'm not following. You are comparing this to the `into_qsc_args` function? That function can encounter dependency errors while it is calculating the arguments that qsc compilation needs, whereas this function can't encounter such errors in articulating the arguments that qasm compilation needs. I can have this function return a Result type that will always be the Ok value, but is that really useful?

source/vscode/test/suites/language-service/notebook.test.ts
ScottCarda-MS: Insofar as this test was about the workspace profile, it became invalid. But, I think I found a way to restore this test and have it test notebook defaulting to unrestricted in the new system using the entry point argument. I've added that restored test back in.

source/compiler/qsc_project/src/openqasm.rs
ScottCarda-MS: This enables me to use the `Profile::from_str` on line 156.

source/language_service/src/completion/ast_context.rs
ScottCarda-MS: yes, that can be an `Rc<str>`. Done

source/wasm/src/lib.rs
ScottCarda-MS: Oh then _can_ remain as capabilities. That should cut down the number of changes. Thanks!

source/wasm/src/lib.rs
ScottCarda-MS: reverted

source/compiler/qsc_qasm/src/compiler.rs
ScottCarda-MS: done

source/vscode/src/qirGeneration.ts
ScottCarda-MS: It can be a bit much:
image
Should we just chop off the end, removing the "Please ensure the code is compatible ..." bit?

source/wasm/src/project_system.rs
ScottCarda-MS: Ah I see, this was meant conditionally on the panic being turned into a formal error. Since the panic is meant to catch issues with our internal logic, we should keep this logic as using a panic instead of a formal error.

idavis:
Approved openqasm related changes.

source/vscode/src/qirGeneration.ts
minestarks: Yeah... given that this exception is raised for a variety of different errors, I don't think wrapping it in anything "helpful" is really that helpful. Maybe we should just surface `e.message` and be done with it

ScottCarda-MS: updated error messages

PR #2638, opened 2025-08-02
Import and export resolution improvements

minestarks:
### Fixes

- **Fixed package exports** - Export items weren't being generated in HIR for non-aliased exports, breaking libraries like FixedPoint
- **Fixed relative namespace imports** - `import Bar` from within `namespace Foo` now correctly resolves to `Foo.Bar`, matching the behavior of `open Bar`
- **Fixed namespace reexports from dependency packages** - Namespace aliases exported from a dependency's `Main` namespace (like `export A as Alias`) are now properly accessible to consuming packages
- **Fixed declaration order dependencies** - Import/export resolution now works regardless of the order namespaces are declared, using an iterative resolution approach

### Breaking changes

- Primitive type exports (`Unit`, `Qubit`, `Int`) removed from legacy APIs - these never actually worked
- Parent namespace exports now explicitly banned (e.g. `export Parent` where `Parent.Foo` exists) - never properly worked
- Cross-namespace exports from external packages disallowed as they serve no purpose
- Duplicate imports/exports in namespace scope now consistently error - previously only some collisions would error
- Import name collisions with callable/UDT declarations now error

### Library Cleanup

- Removed workaround aliases from FixedPoint and Rotations libraries that are no longer needed
- Cleaned up Microsoft.Quantum.Core legacy API exports

### Implementation Notes

**HIR:**
- Removed redundant `namespaces` field from HIR - only resolver needs the namespace tree
- All import/export items now get item IDs; all exports are lowered while imports are not
- Export items now use `Res` type instead of `ItemId` to preserve even failed export items
- Simplified globals enumeration in `global.rs`

**Parser & AST:**
- Invalid patterns like `import A.* as B`, `export C.*`, and empty `import;` are now reported in the parser, reducing the validation that the resolver has to do
- Changed from "glob" to "wildcard" throughout codebase to match public documentation

**Resolver:**
- Iterative fixed-point algorithm resolves dependency chains between import/exports
- Simplified lowerer by moving more logic to resolver
- New public `resolve()` function encapsulates resolver API complexity

**Language Service:**
- Updated completion logic to work with new `ImportKind` structure, but no real fixes yet (next PR)


source/compiler/qsc_passes/src/common.rs
```diff
@@ -97,7 +97,7 @@ pub(crate) fn create_gen_core_ref(
     span: Span,
 ) -> Expr {
     let term = core
-        .resolve_term(namespace, name)
+        .resolve_callable(namespace, name)
         .expect("term should resolve");
```
idavis: should the `expect` still reference `term` or that it is trying to resolve a `callable`?

source/compiler/qsc/src/interpret.rs
```diff
@@ -334,7 +334,7 @@ impl Interpreter {
             .expect("package should exist in the package store")
             .package;
         for global in global::iter_package(Some(map_fir_package_to_hir(package_id)), package) {
-            if let global::Kind::Term(term) = global.kind {
+            if let global::Kind::Callable(term) = global.kind {
                 let store_item_id = fir::StoreItemId {
```
idavis, line 338: is this still true?
```
    /// Given a package ID, returns all the global items in the package.
    /// Note this does not currently include re-exports.
```

source/compiler/qsc_ast/src/visit.rs
```diff
@@ -122,7 +122,10 @@ pub fn walk_item<'a>(vis: &mut impl Visitor<'a>, item: &'a Item) {
...
             for item in &decl.items {
                 vis.visit_path_kind(&item.path);
-                if let Some(ref alias) = item.alias {
+                if let ImportKind::Direct {
+                    alias: Some(ref alias),
+                } = item.kind
+                {
                     vis.visit_ident(alias);
                 }
             }
```
idavis: Would it make sense to pull this out? We tend to do that for other collections:
```rust
decl.items.iter().for_each(|item| vis.visit_import_or_export(item));
```

source/compiler/qsc_ast/src/mut_visit.rs
```diff
@@ -133,7 +133,7 @@ pub fn walk_item(vis: &mut impl MutVisitor, item: &mut Item) {
             for item in &mut *export.items {
                 vis.visit_span(&mut item.span);
                 vis.visit_path_kind(&mut item.path);
-                if let Some(ref mut alias) = item.alias {
+                if let ImportKind::Direct { alias: Some(alias) } = &mut item.kind {
                     vis.visit_ident(alias);
                 }
             }
```
idavis: see comment on `visit.rs`

source/language_service/src/completion/global_items.rs
```diff
@@ -805,15 +806,17 @@ impl ImportItem {
...
-            let is_glob = item.is_glob;
+
+            let is_glob = matches!(item.kind, ImportKind::Wildcard);
 
             buf.push(ImportItem {
-                path: path.rc_str_iter().cloned().collect(),
+                path: item.path.rc_str_iter().cloned().collect(),
                 alias,
                 is_glob,
             });
```
orpuente-MS, line 822: Should we rename `is_glob` to `is_wildcard` now? Same for the `is_glob` field in `ImportItem`.

orpuente-MS: Actually, I saw that you changed the `alias` and `is_glob` fields in the `ImportOrExportItem` struct by just the kind, does the same change makes sense here?

source/compiler/qsc_frontend/src/compile/tests/multiple_packages.rs
```diff
@@ -643,3 +681,366 @@ fn old_syntax_udt_reexported() {
...
+            "A",
+            r#"
+            @EntryPoint()
+            operation Main() : Unit {
+                B.Bar.NamespaceBAlias.Foo();
+            }
+            "#,
+        ),
+    ]);
+}
```
idavis, line 710: Do we need a test that this doesn't work? Sorry, just don't know what is covered in other tests.
```
@EntryPoint()
operation Main() : Unit {
    B.Bar.NamespaceB.Foo();
}
```

source/compiler/qsc_frontend/src/incremental/tests.rs
```diff
@@ -58,6 +58,7 @@ fn one_callable() {
                         opens: {},
                         tys: {},
                         terms: {},
+                        importables: {},
                         vars: {},
                         ty_vars: {},
```
idavis, line 63: We have tests here where the `importables` is empty. Are tests where this is set is covered in other kinds of tests from incremental?

source/compiler/qsc_frontend/src/typeck/check.rs
```diff
@@ -90,6 +90,7 @@ impl GlobalTable {
                 // and recursively resolve it (it could be another export, i.e. a chain of exports.
                 self.handle_item(resolved_export, package_id, store);
             }
+            hir::ItemKind::Export(..) => {}
```
idavis: Is it worth a comment here what cases fall into this pattern?

source/compiler/qsc_frontend/src/resolve/imports.rs
```diff
@@ -0,0 +1,524 @@
...
+/// as a signal that we should continue iterating.
+fn try_resolve_imports_and_exports<'a, I>(
+    resolver: &mut Resolver,
+    scope_items_iter: impl Fn() -> I,
+    attempted_imports: &mut FxHashMap<NodeId, Result<(), Error>>,
+) -> bool
+where
+    I: Iterator<Item = &'a Item>,
+{
+    if scope_items_iter().count() == 0 {
```
idavis, line 122: Assuming we need separate ordered loops here, I think we should collect the iterator once into a vec and update the usages for the vec.

ScottCarda-MS:
I think I found a bug:

file A:
```qsharp
operation Original() : Unit {}
```

file B:
```qsharp
import A.Original as Other;
```

file C:
```qsharp
import B.Other;
```

File C will error saying it can't fine `B.Other`. This is avoided if I put an export in file B: `export Other as ExportedOther`. Then file C can `import B.ExportedOther;`, but that seems like it shouldn't be necessary. My understanding was that exporting should really only be necessary to expose content across the package boundary.

source/compiler/qsc_frontend/src/resolve.rs
```diff
@@ -71,20 +126,30 @@ pub enum Res {
...
 }
 
+/// Kind of importable item - callable, UDT or namespace.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum Importable {
+    Callable(ItemId, ItemStatus),
+    Ty(ItemId, ItemStatus),
+    Namespace(NamespaceId, Option<ItemId>),
+}
+
```
idavis: Can derive `Copy`

source/compiler/qsc_ast/src/ast.rs
```diff
@@ -1968,19 +1965,34 @@ impl ImportOrExportItem {
...
 
+/// The kind of import being performed in an `ImportOrExportItem`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum ImportKind {
+    /// A wildcard import: `import A.*`
+    Wildcard,
+    /// A direct import or export: `import A.B`, `export A`, etc.
+    Direct {
+        /// An optional alias for the item being imported.
+        alias: Option<Ident>,
```
ScottCarda-MS, line 1986: I'm not familiar with having enum values with these named fields, but I like it for the clarity it provides. Usually it is something like `Direct(Option<Ident>)` and you there is nothing that explicitly tells you what the field is for. Is this a newer Rust feature?

source/compiler/qsc_ast/src/visit.rs
ScottCarda-MS: This would make it easier to write visitors that want to do things with imports/exports.

source/compiler/qsc_hir/src/global.rs
```diff
@@ -125,95 +127,80 @@ impl PackageIter<'_> {
...
+                    .0
+                    .split_last()
+                    .expect("namespace name should not be empty");
+                // Parent namespace can be empty
                 Some(Global {
-                    namespace: namespace.into(),
-                    name: Rc::clone(&name.name),
-                    visibility,
+                    namespace: parent.iter().map(|i| i.name.clone()).collect(),
+                    name: name.name.clone(),
```
swernli: super nitpick: I like that the code above uses explicit `Rc::clone` so it would be nice if this did too.

```suggestion
                    namespace: parent.iter().map(|i| Rc::clone(&i.name)).collect(),
                    name: Rc::clone(&name.name),
```

source/compiler/qsc_hir/src/global.rs
```diff
@@ -125,95 +127,80 @@ impl PackageIter<'_> {
...
-                        name: name.name.clone(),
-                        visibility,
-                        status,
-                        kind: Kind::Export(id),
-                    })
-                }
-            }
+            ) => Some(Global {
+                namespace: namespace.into(),
+                name: name.name.clone(),
```
swernli: here too:

```suggestion
                name: Rc::clone(&name.name),
```

source/compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -65,6 +66,16 @@ impl<'a> Renamer<'a> {
     fn rename(&self, input: &mut String) {
         for (span, change) in self.changes.iter().rev() {
             let name = match change {
+                Change::Import(res_source, res) => {
+                    format!(
+                        "{} as {}",
```
swernli, line 71: nitpick: I was a little confused by this phrasing at first. I think I understand now that it's meant to convey "<item id of the target of the export> as <item id of the export itself>" but at first I thought it was related to aliases. Is there a different phrasing that could make it less overlapping with the existin alias syntax?

source/language_service/src/completion/tests.rs
```diff
@@ -2083,19 +2083,6 @@ fn path_segment_before_glob() {
     );
 }
 
-#[test]
```
ScottCarda-MS, line 2086: This is a test for something that you are working on implementing, right? It would be better if we just ignored the test for now and cited a reason being that this functionality is in progress, rather than removing the test.

source/language_service/src/completion/ast_context.rs
```diff
@@ -65,13 +65,7 @@ impl<'a> Visitor<'a> for AstContext<'a> {
...
                 self.set_context(Context::Path(PathKind::Import));
                 for item in &decl.items {
-                    if item.is_glob
-                        && item.span.touches(self.offset)
-                        && item
-                            .alias
-                            .as_ref()
-                            .is_none_or(|a| !a.span.touches(self.offset))
-                    {
+                    if ImportKind::Wildcard == item.kind && item.span.touches(self.offset) {
```
ScottCarda-MS: Would using `span.contains` be more effective here?

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -626,15 +626,8 @@ impl<'a> Context<'a> {
                         .expect("local should have type")
                         .clone(),
                 ),
-                Some(Res::ExportedItem(item, _)) => {
```
ScottCarda-MS, line 629: Was this rule here to ensure that the exported item actually exists as a global item? Why is this rule not necessary anymore?

source/compiler/qsc_frontend/src/resolve/imports.rs
```diff
@@ -0,0 +1,524 @@
...
+/// The iterative approach resolves these dependencies in multiple passes.
+#[must_use]
+fn iterate_until_done<F>(mut resolver_fn: F) -> Vec<Error>
+where
+    F: FnMut(&mut FxHashMap<NodeId, Result<(), Error>>) -> bool,
+{
+    let mut errors = Vec::new();
+    let mut attempted_items = FxHashMap::default();
+
+    for i in 1..=100 {
```
idavis: Should probably have this declared as a const at the top of the file. I think that is the pattern we use for other best effect limiting passes in other code.

source/compiler/qsc_frontend/src/resolve.rs
```diff
@@ -1638,52 +1234,90 @@ impl GlobalTable {
...
                 }
-                (global::Kind::Namespace, hir::Visibility::Public) => {
-                    self.scope.insert_or_find_namespace(global.namespace);
+                (global::Kind::Namespace(item_id), hir::Visibility::Public) => {
+                    let this_namespace = self
+                        .scope
+                        .insert_or_find_namespace(vec![global.name.clone()], Some(namespace));
+
+                    self.scope.importables.get_mut_or_default(namespace).insert(
+                        global.name.clone(),
```
swernli: This clone is technically not needed:

```suggestion
                        global.name,
```

source/compiler/qsc_frontend/src/resolve.rs
```diff
@@ -1638,52 +1234,90 @@ impl GlobalTable {
...
+
                             self.scope
-                                .insert_or_find_namespace(ns.iter().map(|s| s.name.clone()));
-                        }
-                        hir::ItemKind::Ty(..) => {
-                            self.scope.tys.get_mut_or_default(namespace).insert(
+                                .insert_alias_for_namespace(orig_id, &global.name, namespace);
+
+                            self.scope.importables.get_mut_or_default(namespace).insert(
                                 global.name.clone(),
```
swernli: same here:

```suggestion
                                global.name,
```

source/compiler/qsc_frontend/src/resolve.rs
```diff
@@ -2366,11 +1982,14 @@ fn resolve_scope_locals(
                     });
                 }
             }
+            NameKind::Importable => {
+                // local variables are not importable
```
swernli, line 1989: Should this be a panic if it can never happen? Just to help a future refactor that might violate this invariant.

source/compiler/qsc_hir/src/global.rs
minestarks: Curious, is there any reason to prefer `Rc::clone` over `.clone()` besides consistency?

source/language_service/src/completion/tests.rs
minestarks: No -- I just thought this test case didn't make sense anymore. After the strict validation I added in the parser, this test broke. I don't think it's worth any extra effort to support completions inside this syntactically invalid construct.

source/compiler/qsc_frontend/src/resolve/tests.rs
minestarks: Well... my "big brain" moment was when I realized there's nothing special about an alias. Every import/export has an alias. It's always* declaring a new name for an existing thing. `import A.B` is just shorthand for `import A.B as B`. So I eliminated almost all code that treats aliases as a special case. An import statement can always be expressed in its "expanded" form as `import X as Y`.

`import A.B` is an item declaration. `A.B` is your source item, and `B` is the name for the new item you're declaring in the current scope.  Both of those things end up getting assigned item IDs. So in these tests, `import item3 as item6` is meant to express what the names `A.B` and `B` get mapped to, respectively.

So I guess what I'm trying to say is "it looks like an alias because it _is_ an alias". 


(*except in self-exports, which have to be special-cased because `operation Foo() : Unit {} \ export Foo as Foo;` in the same scope creates a naming collision. But that's the language design choice we made, so... )

source/compiler/qsc_ast/src/ast.rs
minestarks: Struct-like enum variants have been in Rust since 1.0! I love them.

source/compiler/qsc_frontend/src/resolve/tests.rs
swernli: Given this context, I now love this change. They look like aliases because they are aliases.

source/compiler/qsc_hir/src/global.rs
swernli: > We could have called `a.clone()` rather than `Rc::clone(&a)`, but Rust’s convention is to use `Rc::clone` in this case. The implementation of `Rc::clone` doesn’t make a deep copy of all the data like most types’ implementations of clone do. The call to `Rc::clone` only increments the reference count, which doesn’t take much time. Deep copies of data can take a lot of time. By using `Rc::clone` for reference counting, we can visually distinguish between the deep-copy kinds of clones and the kinds of clones that increase the reference count. When looking for performance problems in the code, we only need to consider the deep-copy clones and can disregard calls to `Rc::clone`.

From https://doc.rust-lang.org/book/ch15-04-rc.html#using-rct-to-share-data

So really, it is for differentiating cheap clone from expensive clone. And then later when looking for places to optimize we can search for ".clone()" and not get false positives for cheap `Rc::clone`.

source/compiler/qsc/src/interpret.rs
minestarks: It is. We'd have to match `global::Kind::Export` for that

source/compiler/qsc_frontend/src/resolve.rs
minestarks: Oh it can definitely happen. This block deals with local variables / type parameters. This code path is just saying "if you're looking for an importable, you're not going to find it among the local variables". But this function definitely _can_ get called with `NameKind::Importable`. In that case, `scope.item()` in the call below should return the item you're looking for.

source/compiler/qsc_frontend/src/incremental/tests.rs
minestarks: So... I think essentially what you're asking is, whether we test importing items that've been declared in incremental compilation (interpreter / notebook etc.) scenarios. I checked, and nope we don't have specific coverage for that. I had convinced myself that name resolution in the incremental compiler isn't meaningfully different than name resolution in "normal" local scopes, so I didn't go about writing specific tests for incremental. But on second thought, it's probably wise to throw a couple of integration test cases in the `Interpreter` to cover everything.

(In case you're wondering why I removed a ton of test cases from this file -- the existing tests had nothing to do with incremental compilation. They went to `multiple_packages.rs` where they belong.)

source/compiler/qsc_frontend/src/compile/tests/multiple_packages.rs
minestarks: Hmm... what would that be testing? That `export A.B as C` doesn't export the same item as both `B` and `C`? Maybe I'm too close to the implementation, but that feels to me like kind of an arbitrary thing to test (we already validate here the export _was_ named `C`, so it's not named `B`). If you don't feel too strongly, I prefer to skip it.

source/compiler/qsc_frontend/src/resolve/imports.rs
minestarks: OK, we're probably splitting hairs here, but if the concern is performance, I had convinced myself that iterating over the same data 3 times is definitely cheaper than allocating a vector. What do you think?

source/compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -4868,6 +4972,7 @@
 
 #[test]
 fn export_namespace_contains_children() {
+    // TODO: this is unlowerable, is there a test case for it?
```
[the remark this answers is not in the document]
minestarks: Yes. It's called `export_parent_namespace_is_err`

source/compiler/qsc_frontend/src/typeck/rules.rs
minestarks: This enum variant (`ExportedItem`) simply doesn't exist anymore after my redesign. By the time we reach the type checker, any import/export usages in any expressions should have been resolved to the original item declarations.

source/language_service/src/completion/ast_context.rs
minestarks: Yeah, that's probably more accurate!

source/compiler/qsc_frontend/src/incremental/tests.rs
minestarks: Oh wait I found one. https://github.com/microsoft/qsharp/blob/305225f8117d86b5989cea56fdc38e755fc10fb0/source/compiler/qsc/src/interpret/tests.rs#L2194-L2195

With that one case, I honestly think we're covered. Test validates that imports work in notebook cells. The other intricacies of import behavior is covered in the name resolution tests.

source/compiler/qsc_frontend/src/compile/tests/multiple_packages.rs
idavis: The goal of the test was to ensure that the original name wasn't exported as well.

minestarks: Added. Guess that wasn't too hard.

idavis (approved):

source/language_service/src/completion/global_items.rs
minestarks: Definitely - this field is disappearing altogether in #2640 , so I'll leave this one for now.

PR #2640, opened 2025-08-05
 Fix completions for import and exports

### LINKED ISSUE #2142: Completion list is missing re-exports from dependencies
In many cases, the completion list is missing reexports from dependencies, which reduces the usability of libraries such as `fixed_point`.

**REPRO**

**`MyProj/src/Main.qs`**

```qsharp
namespace Test {
    open MyDep;
    
    operation Foo() : Unit {
        ExportedItem;    // Appears, but with auto-import edit ❌
        ReexportAlias;   // Does not appear ❌
        ReexportedItem;  // Appears, but with auto-import ❌
                         // Also, doesn't compile (#1955) ❌
    }
}

namespace Test2 {
    open SomeNS; // Does not appear ❌
}

namespace Test3 {    
    open MyDep.C.ReexportedNS; // `B` does not appear after `.` ❌
}

namespace Test4 {
    open ParentNS; // Does not appear ❌ 
}
```

**`MyProj/qsharp.json`**

```json
{
  "dependencies": {
    "MyDep": {
      "path": "../MyDep"
    }
  }
}
```

**`MyDep/src/Main.qs`**

```qsharp
namespace SomeNS {
    operation ReexportedItem() : Unit {}
    export ReexportedItem;
}

namespace Main {
    operation ExportedItem() : Unit {}
     
    export ExportedItem, SomeNS.ReexportedItem, SomeNS.ReexportedItem as ReexportAlias;
}

namespace C {
    export ParentNS.ReexportedNS;
}

namespace ParentNS.ReexportedNS {
    operation ABOp() : Unit {}
}
```

**`MyDep/qsharp.json`**

```json
{}
```

Seen in  3b32013a6e1ff26f5d12c12116b1946eeda4bb0d

**Additional context**

Related: #1955

To capture some historical context -- there was an attempt to cover these cases in the original Completions implementation, but it required quite a few changes to the resolver. Currently, neither the HIR nor the resolver contains enough information about the reexports for the Completions module to include the reexports.

At the time I made some changes to the resolver to address _some_ of the cases from above, but hit blockers on others. So that effort's been abandoned for a few months now.

Image

minestarks:
This pull request refactors the completion logic for imports and exports to be more robust and accurate.

The completion provider in the language service (`global_items.rs`) has been rewritten to leverage the `GlobalScope` and `Locals` structs already available from the resolver. This allows completions to use the same "view" as the name resolution code, rather than trying to recreate it by manually analyzing imports, enumerating items, etc. All of that work has already been done by the resolver.

The previous implementation for completions had difficulty handling more complex import and export scenarios, such as items re-exported from dependencies or items under aliased namespaces. This could lead to incomplete or incorrect suggestions. The new approach provides a more solid foundation, giving the language service a complete and accurate view of all available symbols.

Fixes #2142

### DIFF
--- a/source/compiler/qsc_frontend/src/resolve/tests.rs
+++ b/source/compiler/qsc_frontend/src/resolve/tests.rs
@@ -3336,6 +3336,29 @@ fn get_locals_block_scope_boundary_begin() {
 }
 
 #[test]
+fn get_locals_namespace_imports() {
+    check_locals(
+        indoc! {"
+            namespace Bar {}
+            namespace Foo {
+                import Bar;
+                import Bar as Baz;
+                import Bar.*;
+                function A() : Int {
+                    ↘
+                }
+            }
+        "},
+        &expect![[r#"
+            namespace 4
+            namespace 3
+            Bar (namespace 3)
+            Baz (namespace 3)
+        "#]],
+    );
+}
+
+#[test]
 fn use_after_scope() {
     check(
         indoc! {"
--- a/source/compiler/qsc_frontend/src/resolve/tests.rs
+++ b/source/compiler/qsc_frontend/src/resolve/tests.rs
@@ -3336,6 +3336,29 @@ fn get_locals_block_scope_boundary_begin() {
 }
 
 #[test]
+fn get_locals_item_imports() {
+    check_locals(
+        indoc! {"
+            namespace Bar {
+                function A() : Unit {}
+            }
+            namespace Foo {
+                import Bar.A;
+                import Bar.A as B;
+                function C() : Int {
+                    ↘
+                }
+            }
+        "},
+        &expect![[r#"
+            A (Item 1)
+            B (Item 1)
+            namespace 4
+        "#]],
+    );
+}
+
+#[test]
 fn get_locals_namespace_imports() {
     check_locals(
         indoc! {"
--- a/source/compiler/qsc_frontend/src/resolve/tests.rs
+++ b/source/compiler/qsc_frontend/src/resolve/tests.rs
@@ -4,7 +4,7 @@
 use super::{Error, Locals, Names, Res};
 use crate::{
     compile,
-    resolve::{Importable, LocalKind, Resolver, imports::iter_valid_items},
+    resolve::{Importable, Local, Resolver, imports::iter_valid_items},
 };
 use expect_test::{Expect, expect};
 use indoc::indoc;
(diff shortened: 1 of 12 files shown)


source/compiler/qsc_frontend/src/resolve/tests.rs
```diff
@@ -3029,16 +3029,17 @@ fn check_locals(input: &str, expect: &Expect) {
...
-            output,
-            "{} ({})",
-            l.name,
-            match l.kind {
-                LocalKind::Item(item_id) => item_id.to_string(),
-                LocalKind::TyParam(param_id) => format!("ty_param {param_id}"),
-                LocalKind::Var(node_id) => format!("var {node_id}"),
-            }
-        );
+        let _ = match l {
```
ScottCarda-MS, line 3032: I'm not seeing any tests in this file that make use of this new printing behavior. Could you add a resolver test that resolves an alias, which should creates a different in printed output other than "namespace 3"?

source/language_service/src/completion/tests.rs
```diff
@@ -2379,3 +2380,170 @@ fn in_trailing_comment() {
...
+        r"
+            namespace Test {
+                operation Test() : Unit {
+                    MyDep.↘
+                }
+            }",
+        "MyDep",
+        "namespace Foo {
+                operation Baz() : Unit {}
+                export Baz;
```
ScottCarda-MS, line 2485: I know we've talked before about this, but it is really confusing to me. Do we need to export `Baz` here if it is already exported in `Main` as `Foo.Baz`?

source/language_service/src/test_utils.rs
```diff
@@ -290,13 +287,14 @@ where
     let source_package_id = compiler.source_package_id();
     let (package_store, package_id) = compiler.into_package_store();
 
+    dependencies.push((source_package_id, None));
```
orpuente-MS: I think `dependencies` isn't being used after this line. Do we need this line?

source/language_service/src/completion/tests.rs
minestarks: Yes, `namespace Main { export Foo.Baz }`  exports the name `Baz` at the top level, while `namespace Foo { export Baz }` exports the name `Foo.Baz`.  So, the first one will be referenced as `MyDep.Baz` and the second one `MyDep.Foo.Baz`

source/language_service/src/test_utils.rs
minestarks: Yay, more unused code!

source/compiler/qsc_frontend/src/resolve/tests.rs
minestarks: Added a couple of tests specifcally for imports

source/language_service/src/completion/tests.rs
minestarks: This is kind of like how in Rust:

```rs
pub mod foo {
   pub use other_mod::bar;
}
```

Would create `foo::bar`, _not_ `foo::other_mod::bar`.

minestarks: Or, maybe an even closer example, since we're in `Main.qs` :

If you are in `lib.rs`:

```rs
pub use other_mod::bar;
```

would expose `bar` at the crate level.

ScottCarda-MS: Thanks for the explanation!

PR #2641, opened 2025-08-05
Definition, References, Hover, Rename for import/export decls

minestarks:
Adds Go To Definition, Hover, Go To References, Rename support for import/export declarations. Also, for References and Rename, results when these commands are invoked on any items, now include any import/export declarations.

- Go To Definition on any imports/exports, and any usages of those imports/exports, resolve back to the original item.
- Hover shows hover info of original item.
- Go To References finds references to an item including any imports, exports, and usages across packages, including any aliased usages.
- Rename renames the declaration and usages, as long as they match the alias that the cursor is on.

### DIFF
--- a/source/language_service/src/definition/tests.rs
+++ b/source/language_service/src/definition/tests.rs
@@ -695,13 +695,26 @@ fn item_export() {
 }
 
 #[test]
-fn item_export_with_alias() {
+fn item_export_with_alias_on_path() {
     assert_definition(
         r#"
         namespace Test {
             operation ◉Foo◉() : Unit {
             }
             export Fo↘o as Bar;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_export_with_alias_on_alias() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+            export Foo as B↘ar;
         }
     "#,
     );
--- a/source/language_service/src/definition/tests.rs
+++ b/source/language_service/src/definition/tests.rs
@@ -680,3 +680,92 @@ fn notebook_local_from_later_cell() {
         ("cell2", "let z = ↘x + 2;"),
     ]);
 }
+
+#[test]
+fn item_export() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+            export Fo↘o;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_export_with_alias() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+            export Fo↘o as Bar;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_import() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+        }
+        namespace Other {
+            import X, Test.Fo↘o;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_import_incomplete() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+        }
+        namespace Other {
+            import X, Test.Foo↘
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_import_alias() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+        }
+        namespace Other {
+            import Test.Foo as B↘ar;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_import_alias_usage() {
+    assert_definition(
+        r#"
+        namespace Test {
+            operation ◉Foo◉() : Unit {
+            }
+        }
+        namespace Other {
+            import Test.Foo as Bar;
+            operation Baz() : Unit {
+                let x = Ba↘r();
+            }
+        }
+    "#,
+    );
+}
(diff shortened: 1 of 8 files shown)


idavis (approved):

source/language_service/src/name_locator.rs
```diff
@@ -186,7 +189,7 @@ impl<'package, T: Handler<'package>> Visitor<'package> for Locator<'_, 'package,
 
     // Handles callable, UDT, struct, and type param definitions
     fn visit_item(&mut self, item: &'package ast::Item) {
-        if item.span.contains(self.offset) {
+        if item.span.touches(self.offset) {
```
ScottCarda-MS, line 192: Are we sure changing this to "touches" is right? It seems like we'd only want to visit an item if we are _in_ the item. Is there a scenario that doesn't work correctly unless this is "touches"?

source/language_service/src/definition/tests.rs
```diff
@@ -680,3 +680,92 @@ fn notebook_local_from_later_cell() {
...
+            operation ◉Foo◉() : Unit {
+            }
+            export Fo↘o;
+        }
+    "#,
+    );
+}
+
+#[test]
+fn item_export_with_alias() {
```
ScottCarda-MS: Should there be a test exactly like this but on the export alias too?

source/language_service/src/references/tests.rs
```diff
@@ -928,3 +928,87 @@ fn notebook_local_reference() {
         ("cell2", "let z = ◉↘x◉ + 2;"),
     ]);
 }
+
+#[test]
+fn on_item_declaration_finds_exports_and_imports() {
```
ScottCarda-MS: This is a bit pedantic of me, but to cover the full cross product of possibilities with imports/exports:
```qsharp
namespace Base {
    operation ◉b1◉() : Unit {}
    export ◉b1◉, ◉b1◉ as ◉b2◉;
}
namespace Intermediate {
    import Base.◉b1◉, Base.◉b1◉ as ◉i1◉;
    import Base.◉b2◉, Base.◉b2◉ as ◉i2◉;
    export Base.◉b1◉, Base.◉b2◉, ◉i1◉, ◉i2◉;
}
namespace Usage1 {
    import Base.◉b1◉, Base.◉b1◉ as ◉u1◉;
	import Base.◉b2◉, Base.◉b2◉ as ◉u2◉;
	import Intermediate.◉b1◉ as ◉u3◉;
	import Intermediate.◉b2◉ as ◉u4◉;
	import Intermediate.◉i1◉, Intermediate.◉i1◉ as ◉u5◉;
	import Intermediate.◉i2◉, Intermediate.◉i2◉ as ◉u6◉;
	operation useOp() : Unit {
	    ◉b1◉();
		◉b2◉();
		◉i1◉();
		◉i2◉();
		◉u1◉();
		◉u2◉();
		◉u3◉();
		◉u4◉();
		◉u5◉();
		◉u6◉();
	}
}
namespace Usage2 {
    import Intermediate.◉b1◉, Intermediate.◉b2◉;
	operation useOp() : Unit {
	    ◉b1◉();
		◉b2◉();
	}
}
```
You could use this template for these tests and just put the ↘ in the appropriate place for each test.

minestarks: Well, you didn't cover cross-package exports ;)

Cheap enough to add the test, so I can do so... but I just see so much overlap here. An import-chain of 3 is not meaningfully different than an import-chain of 2. I tried to cover behaviors that could independently go wrong, rather than test every single permutation.

ScottCarda-MS: The reason why I'm going full verbose on this is that imports/exports with aliases have, in the past, been pretty tricky and certain hard-to-think-of scenarios listed here had buggy behavior. All of this seems to work great with this PR 😁 but I want it to ensure it keeps working in the future!

source/language_service/src/name_locator.rs
minestarks: Yeah, but only in case of minor syntax errors (missing semicolon). I had to change this to make `item_import_incomplete` pass. Not critical, but nice enough to support. I kept hitting this case because I wanted to check the definition/hover info _as I was typing the code_ to make sure I was importing the right item.

https://github.com/microsoft/qsharp/blob/0ce4c5122c49ef3770683cead3862694d35242db/source/language_service/src/definition/tests.rs#L725-L738

ScottCarda-MS: That makes sense. Thanks!

source/language_service/src/references/tests.rs
ScottCarda-MS: > An import-chain of 3 is not meaningfully different than an import-chain of 2.

iirc re-exports specifically had issues in the past. I think There is a meaningful distinction to test between an export and a re-export (chain of 2 vs chain of 3 as you put it), but I would agree that further chaining after that would be redundant to test.

source/language_service/src/definition/tests.rs
minestarks: Added.

source/language_service/src/references/tests.rs
minestarks: When you talk about the re-exports we had issues with, I believe you're referring to mostly cross-package exports. Otherwise, "re-exports" within a package, no matter how many levels removed from the original item, look exactly the same in the `Names` table we use. (I have decided to avoid the name re-export because of how ambiguous the definition actually is)

Covering cross-package exports is probably a good idea... but your suggested code didn't include that. No problem though, I can modify it to actually cover cross-package scenarios and still use it.

minestarks: I went down a deep rabbit hole of implementing cross-package Find References, but in the end I decided it was a lot of changes out of scope for this PR. I'll work on it as a separate feature.

For now, I added the test inputs the same as you suggested.

PR #2642, opened 2025-08-06
Playground Profile Selection

ScottCarda-MS:
Profile selection in the playground will now default to adaptive_rif for qir gen.
It will also properly search the user code for a profile specification via the EntryPoint argument.

### DIFF
--- a/source/npm/qsharp/src/compiler/compiler.ts
+++ b/source/npm/qsharp/src/compiler/compiler.ts
@@ -156,10 +156,7 @@ export class Compiler implements ICompiler {
   }
 
   async getRir(program: ProgramConfig): Promise<string[]> {
-    const config = toWasmProgramConfig(
-      program,
-      program.profile || "adaptive_ri",
-    );
+    const config = toWasmProgramConfig(program, "adaptive_rif");
     return callAndTransformExceptions(async () => this.wasm.get_rir(config));
   }
 
--- a/source/npm/qsharp/src/compiler/compiler.ts
+++ b/source/npm/qsharp/src/compiler/compiler.ts
@@ -204,7 +204,7 @@ export class Compiler implements ICompiler {
 
   async getQir(program: ProgramConfig): Promise<string> {
     return callAndTransformExceptions(async () =>
-      this.wasm.get_qir(toWasmProgramConfig(program, "base")),
+      this.wasm.get_qir(toWasmProgramConfig(program, "adaptive_rif")),
     );
   }
 
--- a/source/npm/qsharp/src/compiler/compiler.ts
+++ b/source/npm/qsharp/src/compiler/compiler.ts
@@ -156,7 +156,7 @@ export class Compiler implements ICompiler {
   }
 
   async getRir(program: ProgramConfig): Promise<string[]> {
-    const config = toWasmProgramConfig(program, "adaptive_rif");
+    const config = toWasmProgramConfig(program, "adaptive_ri");
     return callAndTransformExceptions(async () => this.wasm.get_rir(config));
   }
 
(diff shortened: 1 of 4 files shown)


source/npm/qsharp/src/compiler/compiler.ts
```diff
@@ -204,7 +204,7 @@ export class Compiler implements ICompiler {
 
   async getQir(program: ProgramConfig): Promise<string> {
     return callAndTransformExceptions(async () =>
-      this.wasm.get_qir(toWasmProgramConfig(program, "base")),
+      this.wasm.get_qir(toWasmProgramConfig(program, "adaptive_rif")),
```
swernli: I was about to suggest that this default match the behavior of VS Code and Jupyter where "unrestricted" is the default, but then I realized that might have the side effect of breaking QCOM since this is down in the npm layer. That said, even this change will make QCOM start producing adaptive code rather than base, right?

source/playground/src/editor.tsx
```diff
@@ -128,6 +129,7 @@ export function Editor(props: {
     const config = {
       sources: [["code", code]] as [string, string][],
       languageFeatures: [],
+      profile: await getTargetProfileFromEntryPoint("main.qs", code),
```
swernli: To get the "unrestricted" by default behavior, couldn't this playground specific code just use "unrestricted" if it didn't find the profile explicitly?
```suggestion
      profile: await getTargetProfileFromEntryPoint("main.qs", code) ?? "unrestricted",
```

source/npm/qsharp/src/compiler/compiler.ts
minestarks: The VS Code default for the "generate QIR" command is adaptive_rif, so this does match that.

minestarks: `getRir` has a similar default that looks kind of suspicious to me. It should probably match what we do in `getQir`.

ScottCarda-MS: The idea is that, because this is qir generation, the assumed behavior should be the minimal required, which here is adaptive_rif. This matches VS Code behavior, when the user tries to generate qir locally.

source/playground/src/editor.tsx
ScottCarda-MS: The profile at this point is supposed to faithfully report what the user has in their source code, which could be no profile specified. The data that no profile has been specified needs to be carried forward so that later logic can fill in that gap depending on what the user is trying to do, for example the default for qir generation, adaptive_rif, is different from the default for just running the code, unrestricted.

Trying to think through the reasoning, but there are a lot of moving parts to this feature, so it's easy for me to miss something. Do you see why we should always default to unrestricted for the playground?

ScottCarda-MS: I guess it means that, because the default is different for different actions, you could be looking at the playground, see the generated qir based off of an adaptive_rif compilation, and click "Run" triggering a compilation based off of unrestricted profile. If you haven't specified a profile in the EntryPoint. In VS Code that makes more sense, but maybe in the Playground it makes more sense to always default to unrestricted? Though, in that case you'd _have_ to specify the profile in the EntryPoint before you'd be able to see the qir or rir.

source/npm/qsharp/src/compiler/compiler.ts
ScottCarda-MS: What is the bare minimum required for rir generation? I'm not very familiar with what rir is or is used for.

ScottCarda-MS: Seems like it can handle adaptive_rif, so I'll change it to use that as the default for rir generation.

ScottCarda-MS: done

source/playground/src/editor.tsx
ScottCarda-MS: Setting this place in code to have default profile "adaptive_rif" as this profile is only used for qir and rir generation here. The default for running the code in the playground is still "unrestricted".

source/npm/qsharp/src/compiler/compiler.ts
ScottCarda-MS: Changing this back to what it was before to not risk breaking changes to dependents. The default for the playground will be set higher up the stack.

Issue #2419, opened 2025-05-16
Better user guidance on failure to capture non-const symbols.


swernli:
Following up on https://github.com/microsoft/qsharp/pull/2417#discussion_r2093301530, it looks like the error clean up exposed an issue with how `SemanticErrorKind::ExprMustBeConst` is used in `lower_ident_expr`. The error is very helpful, but no longer gets filed because symbol resolution doesn't return actual symbols that are non-const. We need to review the interaction with `try_get_existing_or_insert_err_symbol` to ensure the assumptions are compatible with the `lower_ident_expr`.

PR #2653, opened 2025-08-11
Better error message when capturing non-const const globals in OpenQASM

orpuente-MS:
Addresses #2419.

### DIFF
--- a/source/compiler/qsc_qasm/src/semantic/types.rs
+++ b/source/compiler/qsc_qasm/src/semantic/types.rs
@@ -356,6 +356,7 @@ impl Type {
             | Type::Float(_, is_const)
             | Type::Int(_, is_const)
             | Type::UInt(_, is_const) => *is_const,
+            Type::Function(..) | Type::Gate(..) | Type::Void => true,
             _ => false,
         }
     }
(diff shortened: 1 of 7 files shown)


source/compiler/qsc_qasm/src/semantic/types.rs
```diff
@@ -356,6 +356,7 @@ impl Type {
             | Type::Float(_, is_const)
             | Type::Int(_, is_const)
             | Type::UInt(_, is_const) => *is_const,
+            Type::Function(..) | Type::Gate(..) | Type::Void => true,
```
idavis: I think this needs to be something other than `is_const`. The symbol table is able to see symbols of this type from special scopes, but they aren't `const` as they argument for these being const could be applied to other types that aren't marked as `const`, but as can't see them from the special scopes.

orpuente-MS: Yes, I agree. And `Type::Void` shouldn't be there, since `void` can only be used as the output type of functions.

I created a new predicate called `is_callable` for `Type::Function` and `Type::Gate`.

idavis (approved):

PR #2654, opened 2025-08-12
Allow recursive calls to operations that return `Unit`

swernli:
This change updates RCA to allow a subset of cyclic operation calls. These operation calls are allowed as long as the operation returns `Unit`. The change involves:
 - updating the cyclic callable check to avoid populating the compute properties for cyclic operations that return `Unit`
 - updating the core analysis logic to check callee expressions for cycles in the active context, which are treated as "unresolved" to fall back to partial evaluation capabilities checks.

Some new test cases confirm the expected behavior for the corner cases, as well as updates to existing test cases to account for the change that adds `CallToUnresolvedCallee` capability.

### DIFF
--- a/source/compiler/qsc_rca/src/core.rs
+++ b/source/compiler/qsc_rca/src/core.rs
@@ -504,7 +504,6 @@ impl<'a> Analyzer<'a> {
             if quantum_properties
                 .runtime_features
                 .contains(RuntimeFeatureFlags::CyclicOperationSpec)
-                && Some(&callee.item) != self.get_current_context().get_current_item_id()
             {
                 quantum_properties
                     .runtime_features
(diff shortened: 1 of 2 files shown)


swernli:
@minestarks the changes are much simpler now with the further relaxation you proposed. Everything should be up-to-date and ready for rereview.

minestarks (approved):
Nice!!

minestarks:
@swernli This change is great, however now that I've been testing the relevant scenarios, I'm still unhappy with the way we surface infinite recursion issues in VS code QIR generation.  This is a preexisting problem (we always allowed recursive classical functions, so...) so I'll just file a separate issue.

minestarks:
Filed #2661

PR #2656, opened 2025-08-12
OpenQASM language service: add rename/definition/references, semantic passes, and VS Code wiring

idavis:
## Summary of changes

- Language service (OpenQASM)
  - Added OpenQASM support for prepare-rename, rename, go-to-definition, and find-references.
  - Added tests for OpenQASM rename. Other functionalities are transitively tested as they all use the same code underneath
- OpenQASM semantic layer (compiler)
  - Implemented semantic passes and visitors to support symbol lookup and reference finding (SymbolFinder, ReferenceFinder).
  - Fixed and standardized source span tracking (notably for casts) locations.
  - Introduced new semantic AST field indicating whether casts were implicit or explicit
- VS Code extension
  - Registered providers for OpenQASM: completion, definition, references, and rename.
- In VS Code, OpenQASM files now support:
  - Rename symbol (F2) for user-defined identifiers, gates, functions, etc; built-ins are not renameable.
  - Go to Definition (F12) and Find All References (Shift+F12).

### Notes

- Expressions in type designators are not support with the LS features. We need a follow up PR to add the support as it will be a significant set of changes to support this. There is an ignored test for this scenario for the follow up PR.
- The semantic span fixes (e.g., casts) were required to make rename/definition locations correct, but it turned out that we couldn't reuse the Q# AST based LS features, so they are now correct, but aren't used for LS features.

### DIFF
--- a/source/compiler/qsc_qasm/src/compiler.rs
+++ b/source/compiler/qsc_qasm/src/compiler.rs
@@ -346,7 +346,7 @@ impl QasmCompiler {
             is_qiskit,
         );
 
-        let ast_ty = map_qsharp_type_to_ast_ty(&output_ty);
+        let ast_ty = map_qsharp_type_to_ast_ty(&output_ty, whole_span);
         signature.output = format!("{output_ty}");
         // This can create a collision on multiple compiles when interactive
         // We also have issues with the new entry point inference logic.
@@ -370,7 +370,7 @@ impl QasmCompiler {
                     build_arg_pat(
                         s.name.clone(),
                         s.span,
-                        map_qsharp_type_to_ast_ty(&qsharp_ty),
+                        map_qsharp_type_to_ast_ty(&qsharp_ty, s.ty_span),
                     )
                 })
                 .collect(),
@@ -795,7 +795,7 @@ impl QasmCompiler {
                 let symbol = self.symbols[*arg].clone();
                 let name = symbol.name.clone();
                 let qsharp_ty = self.map_semantic_type_to_qsharp_type(&symbol.ty, symbol.ty_span);
-                let ast_type = map_qsharp_type_to_ast_ty(&qsharp_ty);
+                let ast_type = map_qsharp_type_to_ast_ty(&qsharp_ty, symbol.ty_span);
                 (
                     name.clone(),
                     ast_type.clone(),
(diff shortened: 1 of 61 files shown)


source/vscode/src/language-service/activate.ts
```diff
@@ -119,6 +119,13 @@ export async function activateLanguageService(
     ),
   );
 
+  subscriptions.push(
+    vscode.languages.registerDefinitionProvider(
+      openqasmLanguageId,
+      createDefinitionProvider(languageService),
```
minestarks, line 118: I think it's better to create the definition provider once, and then register it for both languages by passing an array into `registerDefinitionProvider` like `[qsharpLanguageId, openqasmLanguageId]` . (same comment applies to all openqasm language service features in this file)

source/language_service/src/test_utils.rs
```diff
@@ -397,3 +397,54 @@ fn target_offsets_to_spans(target_offsets: &[u32]) -> Vec<Span> {
     }
     spans
 }
+
+pub(crate) mod openqasm {
```
minestarks: In general, I find inline `mod`s confusing, and prefer modules to be their own files. Do you have a strong preference either way?

source/vscode/src/language-service/activate.ts
idavis: I'll try this out, didn't know that was possible.

source/compiler/qsc_qasm/src/compiler.rs
```diff
@@ -1447,6 +1452,19 @@ impl QasmCompiler {
...
 
+    fn compile_captured_ident_expr(&mut self, symbol_id: SymbolId, span: Span) -> qsast::Expr {
+        let symbol = &self.symbols[symbol_id];
+        // when closing over a constant value we will have a const value
+        // associated with the symbol, but due to scoping rule differences
+        // we have to "copy" the value into the usage.
+        if let Some(value) = symbol.get_const_value() {
+            self.compile_literal_expr(&value, span)
+        } else {
+            // todo: err?
```
minestarks: TODO?

source/language_service/src/openqasm.rs
```diff
@@ -0,0 +1,109 @@
...
+        position,
+        position_encoding,
+    );
+    let id = SymbolFinder::get_symbol_at_offset(&res.program, offset, &res.symbols);
+    (res, id)
+}
+
+/// Maps a source position from the user package
+/// to a package (`SourceMap`) offset.
+fn source_position_to_package_offset(
```
minestarks: This looks familiar... isn't it duplicated from somewhere?

orpuente-MS (approved):

minestarks (approved):

PR #2664, opened 2025-08-13
Nicer error messages for QIR generation

### LINKED ISSUE #2661: VS Code could show more helpful UI for infinite loops and recursions in user code (QIR generation)
Partial evaluation, and therefore QIR generation, is inherently susceptible to infinite loops and out-of-memory conditions since we're executing user code. So we can't prevent these fully.

However, the way these conditions are surfaced in the UI is not helpful. It looks like a bug in the extension, rather than a problem in the user code. I think we can do better by catching some of these fatal conditions and wrap them in a more helpful user message that tells them what to look for.

(If we wanted to get fancy, we can emit warnings from the evaluator itself if we detect that no progress is being made, but that would be pretty fancy. This issue is just about the UI.)

Current UI:

### infinite loops (after a rather long timeout):

```qsharp
while 0 == 0 {}
```

Image

### infinite recursion in classical `function`s:

```qsharp
function Foo() : Unit {
    Foo();
}
```

image

### infinite recursion in `operation`s:

```qsharp
operation Foo() : Unit {
    Foo();
}
```

image

minestarks:
Fixes #2661

### DIFF
--- a/source/vscode/src/qirGeneration.ts
+++ b/source/vscode/src/qirGeneration.ts
@@ -203,7 +203,8 @@ async function getQirForProgram(
         );
       }
       throw new QirGenerationError(
-        "Compiling to QIR timed out. Please check your source code for potential infinite loops or excessively long-running operations.",
+        `Compiling to QIR timed out after ${generateQirTimeoutMs / 1000} seconds. ` +
+          "Please check your source code for potential infinite loops or excessively long-running operations.",
       );
     }
     if (e instanceof QdkDiagnostics) {
--- a/source/vscode/src/qirGeneration.ts
+++ b/source/vscode/src/qirGeneration.ts
@@ -1,8 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { getCompilerWorker, log, TargetProfile } from "qsharp-lang";
+import {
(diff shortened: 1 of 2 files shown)


source/vscode/src/qirGeneration.ts
```diff
@@ -182,13 +189,34 @@ async function getQirForProgram(
...
     if (e.toString() === "terminated") {
+      if (cancelled) {
+        throw new QirGenerationError(
+          "Compiling to QIR was cancelled. If the operation is taking an unusually long time, " +
+            "please check your source code for potential infinite loops or excessively long-running operations.",
+        );
+      }
       throw new QirGenerationError(
-        "QIR generation was cancelled or timed out.",
+        `Compiling to QIR timed out after ${generateQirTimeoutMs / 1000} seconds. ` +
```
idavis, line 206: Does this do integer math for a round number of seconds?

minestarks: No... this is a value hardcoded by us. If ever we are possessed to code this value to something like 123456 milliseconds... I'm okay with it showing "123.456 seconds" in the UI. And we aren't really dealing in computed values or ranges where we have to worry about floating point precision.

orpuente-MS (approved):

PR #2666, opened 2025-08-13
Fix panic when running .qsc circuit file

orpuente-MS:
The `Run` button in the left panel of the .qsc circuit files started panicking after #2572 fixed an off-by-one error.

The panic happened because:
1. The Run button generates a block entry expression instead of just a call to an `@EntryPoint` function.
2. When pushing scopes, we tag them with the current `frameId`.
3. However, since no functions have been called when running the entry expression, we haven't pushed any `frameId`s. So, we were trying to tag that block entry expression with a non-existing `frameId`, which caused the panic.

From the ideas we discussed to address this issue, the one that seemed more future proof was reserving the `frameId` `0` for the entry expression (even though there is no real Frame pushed by the entry expr, since it is not a function call). And the first actual Frame, e.g. the one from the `@EntryPoint` function, will have `frameId` `1`.

### DIFF
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -615,6 +615,10 @@ impl State {
         }
     }
 
+    fn current_frame_id(&self) -> usize {
+        self.call_stack.len()
+    }
+
     fn push_frame(&mut self, exec_graph: ExecGraph, id: StoreItemId, functor: FunctorApp) {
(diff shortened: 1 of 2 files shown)


ScottCarda-MS (approved):
Glad the code changes ended up being nice and streamlined. Thanks for addressing this!

swernli (approved):

DmitryVasilevsky (approved):

billti (approved):

PR #2671, opened 2025-08-14
Support numpy arrays in the Python interop layer

orpuente-MS:
This PR changes the if statement that detects python lists to detect iterables in general, allowing is to process, among other things, numpy arrays.


swernli (approved):
code looks good, confirmed the behavior works as expected in my scenario too!

swernli:

PR #2672, opened 2025-08-14
Copilot Reads Summaries from DocGen Tool

ScottCarda-MS:
Adds a Copilot tool called `qsharpGetLibraryDescriptions` which gives Copilot updated descriptions of Q# Library API items, allowing it to make more accurate use of library items in generating Q# snippets and in answering user questions about the libraries.
The doc get tool is modified to include a way to generate these summaries in a JSON that is used by the Copilot tool.

### DIFF
--- a/source/vscode/package.json
+++ b/source/vscode/package.json
@@ -944,7 +944,7 @@
         ],
         "toolReferenceName": "qsharpGetLibraryDescriptions",
         "displayName": "Get Q# Library API Descriptions",
-        "modelDescription": "Returns a structured JSON description of all Q# standard library items, organized by namespace. Each item includes its name, namespace, kind, signature, summary, parameter descriptions, and output description. Use this tool to answer any questions about Q# library functions or to ensure generated Q# code uses library functions correctly.",
+        "modelDescription": "Returns a Markdown string summarizing all Q# standard library items, grouped by namespace. Each entry includes its signature and a short description extracted from doc comments when available. Use this tool to answer any questions about Q# library functions or to ensure generated Q# code uses library functions correctly.",
         "canBeReferencedInPrompt": true,
         "icon": "./resources/file-icon-light.svg",
         "inputSchema": {
--- a/source/vscode/src/gh-copilot/qsharpTools.ts
+++ b/source/vscode/src/gh-copilot/qsharpTools.ts
@@ -223,9 +223,9 @@ export class QSharpTools {
   }
 
   /**
-   * Copilot tool: Returns a structured JSON description of all Q# standard library items,
-   * organized by namespace. Each item includes its name, namespace, kind, signature, summary,
-   * parameter descriptions, and output description.
+   * Copilot tool: Returns a Markdown string summarizing all Q# standard library items,
+   * organized by namespace. Each entry includes its signature and a short description extracted
+   * from doc comments when available.
    */
   async qsharpGetLibraryDescriptions(): Promise<string> {
     const compilerRunTimeoutMs = 1000 * 5; // 5 seconds


source/vscode/package.json
```diff
@@ -925,6 +925,25 @@
           ],
           "additionalProperties": false
         }
+      },
+      {
+        "name": "qsharp-get-library-descriptions",
```
minestarks, line 939: So since you don't do any filtering, how much data does this end up being? Does it fit comfortably into our token budget, even when the conversation has gotten long? (We have debug logs to make manual testing a bit easier for this).

source/vscode/resources/chat-instructions/qsharp.instructions.md
```diff
@@ -127,6 +127,14 @@ and `azure-quantum` packages.
 
 To execute Q# code, use the provided tools.
 
+## Copilot tool usage for Q# libraries
```
minestarks: Nitpicking, but this section could technically be part of the above section which is encouraging Copilot to use the provided tools ("setup and tools").

Also, in the interest of not using up too many tokens.... providing the tool description again ("retrieve the authoritative list of available Q# library items, their signatures, summaries, parameters, and output descriptions.") seems a bit unnecessary to me. Does it make a difference? The tool descriptions should already be in the context, so this isn't adding information.

source/vscode/src/gh-copilot/qsharpTools.ts
```diff
@@ -234,6 +234,28 @@ export class QSharpTools {
...
+      worker.terminate();
+    }, compilerRunTimeoutMs);
+    const worker = loadCompilerWorker(this.extensionUri!);
+    const summaries = await worker.getLibrarySummaries();
+    clearTimeout(compilerTimeout);
+    worker.terminate();
+
+    const summariesObj = deepMapToObject(summaries);
+    const temp = JSON.stringify(summariesObj, null, 2);
+    log.info(`Summaries: ${temp}`);
```
billti: I don't think should be an 'info' level, maybe debug. It's not generally interesting. (Also, if that's only used for logging, which will rarely be on, maybe make the 2 lines above conditional on the log level if the temp string might be large/expensive)

source/vscode/src/gh-copilot/qsharpTools.ts
```diff
@@ -322,3 +344,17 @@ export class QSharpTools {
     worker.terminate();
   }
 }
+
+function deepMapToObject(value: any): any {
```
billti: I'm confused on the need for this. Isn't the API already returning an object? What's wrong with the return value you're already getting? And if what it's returning is not suitable, can you not just change what it is returning, being the API is being added with this PR and this is the only caller of it?

source/npm/qsharp/src/compiler/compiler.ts
```diff
@@ -76,6 +76,8 @@ export interface ICompiler {
 
   getDocumentation(additionalProgram?: ProgramConfig): Promise<IDocFile[]>;
 
+  getLibrarySummaries(): Promise<any>;
```
minestarks: This doesn't seem like an improvement over the existing `getDocumentation` - it returns the same information with different formatting, and is less strongly typed. Why not just use `getDocumentation`? Is there are reason for all this new code to serialize to JSON? Copilot should be able to handle the documentation as is.

source/vscode/src/gh-copilot/qsharpTools.ts
```diff
@@ -234,6 +234,28 @@ export class QSharpTools {
...
+   * Copilot tool: Returns a structured JSON description of all Q# standard library items,
+   * organized by namespace. Each item includes its name, namespace, kind, signature, summary,
+   * parameter descriptions, and output description.
+   */
+  async qsharpGetLibraryDescriptions(): Promise<any> {
+    const compilerRunTimeoutMs = 1000 * 5; // 5 seconds
+    const compilerTimeout = setTimeout(() => {
+      worker.terminate();
+    }, compilerRunTimeoutMs);
+    const worker = loadCompilerWorker(this.extensionUri!);
```
minestarks, line 235: After the 6th "run the compiler with a timeout" boilerplate... I'm starting to think it'd be nice to have a wrapper for this pattern.

source/compiler/qsc_doc_gen/src/generate_docs/tests.rs
```diff
@@ -207,3 +207,117 @@ fn top_index_file_generation() {
...
+              },
+              {
+                "description": "Length of the new array.",
+                "name": "length"
+              }
+            ],
+            "signature": "function Repeated<'T>(value : 'T, length : Int) : 'T[]",
+            "summary": "Creates an array of given `length` with all elements equal to given `value`. `length` must be a non-negative integer."
+          }
+        ]"#]]
```
billti: Is the above the volume of information you return for just 2 APIs? As discussed in the room on the whiteboard, I'm concerned about the amount of tokens this might use. Could it not just generate a response with two lines for each API, pretty much just the signature and summary lines (maybe with the function name prefixed by namespace in the signature)?

source/compiler/qsc_doc_gen/src/display.rs
```diff
@@ -813,3 +813,74 @@ pub fn parse_doc_for_param(doc: &str, param: &str) -> String {
     .trim()
     .to_string()
 }
+
+/// Takes a doc string from a Q# callable and returns all parameter names and their descriptions
+/// as a Vec<(String, String)>, extracted from the # Input(s) section.
+#[must_use]
+pub fn parse_doc_for_all_params(doc: &str) -> Vec<(String, String)> {
```
billti: We're ending up with quite a few different parsers in this file to extract various parts of doc comments. I wonder if we shouldn't just have one good doc comment parser that turns a doc comment into a strongly typed struct that every other helps can call and take the parts it needs from. (Maybe not for this PR)

source/vscode/src/gh-copilot/qsharpTools.ts
ScottCarda-MS: In the wasm, the data object is a BTreeMap. I've tried to use serde to convert it to something that will be recognized in the TS layer, but I've not been successful with that. This is handling the issue from the TS side by converting it into a JSON that can be properly read. I'd be happy to do this in the wasm if you know of a way to do this.

source/vscode/package.json
ScottCarda-MS: The returned string is 150879 characters long.

source/vscode/src/gh-copilot/qsharpTools.ts
ScottCarda-MS: The log isn't very seems useful after my changes, so I removed it.

source/vscode/resources/chat-instructions/qsharp.instructions.md
ScottCarda-MS: I move the content to the "Setup and tools" section and shortened the instructions a bit.

source/vscode/src/gh-copilot/qsharpTools.ts
ScottCarda-MS: I tried implementing an abstraction over this timeout logic and it got very complicated and ended up not simplifying the call sites either. I don't think this is worth it.

source/vscode/package.json
ScottCarda-MS: At a first glance, it looks like the tool has a budget of 41549 tokens and it used 27524 tokens. I will test it out with some longer conversations though.

source/compiler/qsc_doc_gen/Cargo.toml
```diff
@@ -18,6 +18,7 @@ qsc_frontend = { path = "../qsc_frontend" }
 qsc_ast = { path = "../qsc_ast" }
 qsc_hir = { path = "../qsc_hir" }
 rustc-hash = { workspace = true }
+serde_json = { workspace = true }
```
minestarks: Is this dependency still used?

source/vscode/src/gh-copilot/qsharpTools.ts
ScottCarda-MS: Not needed anymore as the return type is just simply string now.

source/compiler/qsc_doc_gen/src/display.rs
ScottCarda-MS: These parsers aren't needed anymore as we are just using the summaries from the doc string so I removed them, but it is a good idea to keep in mind if we find ourselves adding more parsers in the future.

source/compiler/qsc_doc_gen/Cargo.toml
ScottCarda-MS: Nope, removed

source/compiler/qsc_doc_gen/src/generate_docs/tests.rs
ScottCarda-MS: I changed this to just return markdown containing just the name, summary and signature, so it is much smaller now.

source/npm/qsharp/src/compiler/compiler.ts
ScottCarda-MS: Discussed offline: The API's are conceptually similar, but the functionality and the returned data structure is different enough to warrant a different API call. The return type for the getLibrarySummaries was updated to be more specific, though.

source/vscode/package.json
ScottCarda-MS: The token count seems to be consistently lower than the token budget.

source/vscode/package.json
```diff
@@ -934,6 +934,25 @@
...
+        "tags": [
+          "azure-quantum",
+          "qsharp",
+          "qdk"
+        ],
+        "toolReferenceName": "qsharpGetLibraryDescriptions",
+        "displayName": "Get Q# Library API Descriptions",
+        "modelDescription": "Returns a structured JSON description of all Q# standard library items, organized by namespace. Each item includes its name, namespace, kind, signature, summary, parameter descriptions, and output description. Use this tool to answer any questions about Q# library functions or to ensure generated Q# code uses library functions correctly.",
```
minestarks: I guess it doesn't need to say "structured JSON" anymore

ScottCarda-MS: fixed, thanks

minestarks (approved):

PR #2676, opened 2025-08-18
Add support for array concatenation in OpenQASM

orpuente-MS:
This PR adds support for [array concatenation in OpenQASM](https://openqasm.com/versions/3.0/language/types.html#array-concatenation-and-slicing).

### Notes

Array concatenation does not appear in the Grammar and is underspecified in the Spec. This implementation is conservative in two senses while we wait for Spec clarification:
1. We don't allow concatenating arrays of different base types, even if they differ only in width.
2. We don't allow mixing together arrays, static array references, and dynamic array references.

### DIFF
--- a/source/compiler/qsc_qasm/src/semantic/lowerer.rs
+++ b/source/compiler/qsc_qasm/src/semantic/lowerer.rs
@@ -3985,19 +3985,43 @@ impl Lowerer {
     fn cast_array_expr_to_type(ty: &Type, expr: &semantic::Expr) -> Option<semantic::Expr> {
         assert!(matches!(expr.ty, Type::Array(..)));
 
+        let Type::Array(array_ty) = &expr.ty else {
+            unreachable!("`expr` should be an array");
+        };
+
         match ty {
-            Type::StaticArrayRef(ref_ty) if !ref_ty.is_mutable => Some(Expr {
-                span: expr.span,
-                kind: expr.kind.clone(),
-                const_value: expr.const_value.clone(),
-                ty: ty.clone(),
-            }),
-            Type::DynArrayRef(ref_ty) if !ref_ty.is_mutable => Some(Expr {
-                span: expr.span,
-                kind: expr.kind.clone(),
-                const_value: expr.const_value.clone(),
-                ty: ty.clone(),
-            }),
+            Type::StaticArrayRef(ref_ty) if !ref_ty.is_mutable => {
+                if !types_equal_except_const(
+                    &array_ty.base_ty.clone().into(),
+                    &ref_ty.base_ty.clone().into(),
+                ) || array_ty.dims != ref_ty.dims
+                {
+                    return None;
+                }
+
+                Some(Expr {
+                    span: expr.span,
+                    kind: expr.kind.clone(),
+                    const_value: expr.const_value.clone(),
+                    ty: ty.clone(),
+                })
+            }
+            Type::DynArrayRef(ref_ty) if !ref_ty.is_mutable => {
+                if !types_equal_except_const(
+                    &array_ty.base_ty.clone().into(),
+                    &ref_ty.base_ty.clone().into(),
+                ) || array_ty.dims.num_dims() != u32::from(ref_ty.dims)
+                {
+                    return None;
+                }
+
+                Some(Expr {
+                    span: expr.span,
+                    kind: expr.kind.clone(),
+                    const_value: expr.const_value.clone(),
+                    ty: ty.clone(),
+                })
+            }
             _ => None,
         }
     }
--- a/source/compiler/qsc_qasm/src/tests/expression/function_call.rs
+++ b/source/compiler/qsc_qasm/src/tests/expression/function_call.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use crate::tests::{compile_qasm_to_qir, compile_qasm_to_qsharp};
+use crate::tests::{check_qasm_to_qsharp, compile_qasm_to_qir, compile_qasm_to_qsharp};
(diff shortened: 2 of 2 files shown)


source/compiler/qsc_qasm/src/parser/expr.rs
```diff
@@ -701,17 +704,37 @@ fn lit_array_element(s: &mut ParserContext) -> Result<Expr> {
 /// These are expressions allowed in classical declarations.
 /// Grammar: `arrayLiteral | expression | measureExpression`.
```
idavis, line 705: Do we have a bug reported for the grammar as this comment doesn't match the code now.

source/compiler/qsc_qasm/src/semantic/lowerer.rs
```diff
@@ -883,6 +892,7 @@ impl Lowerer {
         let rhs = match rhs {
             syntax::ValueExpr::Expr(expr) => self.lower_expr(expr),
             syntax::ValueExpr::Measurement(measure_expr) => self.lower_measure_expr(measure_expr),
+            syntax::ValueExpr::Concat(expr) => self.lower_array_concat_expr(expr),
         };
```
idavis: is ++= actually supported? This seems like a bug.

source/compiler/qsc_qasm/src/semantic/types.rs
```diff
@@ -757,12 +757,12 @@ impl ArrayDimensions {
...
-            | ArrayDimensions::Five(_, _, _, _, d)
-            | ArrayDimensions::Six(_, _, _, _, _, d)
-            | ArrayDimensions::Seven(_, _, _, _, _, _, d) => Some(*d),
+            | ArrayDimensions::Two(d, _)
+            | ArrayDimensions::Three(d, _, _)
+            | ArrayDimensions::Four(d, _, _, _)
+            | ArrayDimensions::Five(d, _, _, _, _)
+            | ArrayDimensions::Six(d, _, _, _, _, _)
+            | ArrayDimensions::Seven(d, _, _, _, _, _, _) => Some(*d),
             ArrayDimensions::Err => None,
```
idavis, line 766: Do we have a test that shows this was a bug?

source/compiler/qsc_qasm/src/semantic/error.rs
```diff
@@ -169,6 +169,11 @@ pub enum SemanticErrorKind {
     #[error("inconsistent types in alias expression: {0}")]
     #[diagnostic(code("Qasm.Lowerer.InconsistentTypesInAlias"))]
     InconsistentTypesInAlias(String, #[label] Span),
+
+    #[error("inconsistent types in array concatenation expression: {0}")]
+    #[diagnostic(code("Qasm.Lowerer.InconsistentTypesInArrayConcatenation"))]
+    InconsistentTypesInArrayConcatenation(String, #[label] Span),
+
     #[error("indexed must be a single expression")]
```
idavis:
```suggestion
    #[diagnostic(code("Qasm.Lowerer.InconsistentTypesInAlias"))]
    InconsistentTypesInAlias(String, #[label] Span),
    #[error("inconsistent types in array concatenation expression: {0}")]
    #[diagnostic(code("Qasm.Lowerer.InconsistentTypesInArrayConcatenation"))]
    InconsistentTypesInArrayConcatenation(String, #[label] Span),
    #[error("indexed must be a single expression")]
```

source/compiler/qsc_qasm/src/semantic/const_eval.rs
```diff
@@ -127,6 +127,11 @@ impl Expr {
                 // in [`Lowerer::lower_sizeof_call_expr`].
                 None
             }
+            ExprKind::Concat(_) => {
+                // Arrays are non-const, so we don't need to implement array
+                // concatenation in the const-evaluator.
+                None
+            }
```
idavis, line 134: I don't think this is true. Need to push on the TSC for clarity.

source/compiler/qsc_qasm/src/semantic/types.rs
orpuente-MS: This wasn't being used anywhere, it was dead code. Now we have code that uses it, and that code is covered by tests.

source/compiler/qsc_qasm/src/semantic/const_eval.rs
orpuente-MS: The grammar doesn't have `const array[...]`. And there is no mention of `const array` in any spec example. So, I think this spec-compliant for now.

```g4
constDeclarationStatement: CONST scalarType Identifier EQUALS declarationExpression SEMICOLON;
```

source/compiler/qsc_qasm/src/parser/expr.rs
orpuente-MS: Yes, here: <https://github.com/openqasm/openqasm/issues/620>.

source/compiler/qsc_qasm/src/semantic/lowerer.rs
orpuente-MS: This was an "operation not supported between types {} {}" error. But I changed it after your comment to say, "concatenation expressions are not allowed in the rhs of assignment operation statements".

source/compiler/qsc_qasm/src/semantic/const_eval.rs
idavis: We are corrrect, for now, but I think that const arrays are coming as a grammar update. We'll just have to keep our ears open.

source/compiler/qsc_qasm/src/parser/expr.rs
idavis: Can you please update this function to link to the issue?

PR #2677, opened 2025-08-19
Add missing UDT tag in Python - Q# interop layer

orpuente-MS:
There was an edge case in which passing a Python complex number to a Q# identity function, and getting that output back to Python would result in a tuple and not in a Python complex number. This was due to a missing type tag when constructing the Q# Complex UDT value from the Python object.

This PR adds that missing type tag.

### DIFF
--- a/source/compiler/qsc/src/interpret.rs
+++ b/source/compiler/qsc/src/interpret.rs
@@ -563,12 +563,12 @@ impl Interpreter {
             //              an error.
             //           2. The input type is `Std.OpenQASM.Angle.Angle`. To verify that
             //              the input type is indeed `Angle`, we call `udt_ty_from_item_id`,
-            //              which caches the `Angle` UDT `LocalItemId`.
+            //              which caches the `Angle` UDT's `LocalItemId`.
             //
             //         So, if we proceed to execute the function's body, it's guaranteed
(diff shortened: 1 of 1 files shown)


idavis (approved):

source/compiler/qsc/src/interpret.rs
```diff
@@ -528,6 +531,91 @@ impl Interpreter {
...
+            //           1. The input type is not `Std.OpenQASM.Angle.Angle` and we return
+            //              an error.
+            //           2. The input type is `Std.OpenQASM.Angle.Angle`. To verify that
+            //              the input type is indeed `Angle`, we call `udt_ty_from_item_id`,
+            //              which caches the `Angle` UDT `LocalItemId`.
+            //
+            //         So, if we proceed to execute the function's body, it's guaranteed
+            //         that we have already cached `Std.OpenQASM.Angle.Angle`'s `LocalItemId`.
+            //         Therefore, this else-branch is unreachable.
+            unreachable!()
```
minestarks: Thanks for the explanatory comment, very helpful. Could you add a string to the `unreachable!`, so that it's easy to search for in the source, if we do ever hit it?

minestarks (approved):

source/compiler/qsc/src/interpret.rs
orpuente-MS: Sure!

PR #2680, opened 2025-08-21
Adding extern compilation for OpenQASM

idavis:
This PR is in draft until we can get agreement for the behavior of extern compilation.


idavis:

PR #2687, opened 2025-08-26
Refactor Azure job submission

minestarks:
Refactoring Azure job submission so that this PR can be easier to read:  #2686 

There's a slight change in order of the telemetry events. We were sending `submitJob` twice when the Copilot tools were invoked.  This is the only behavior change I'm expecting.

### DIFF
--- a/source/vscode/src/projectSystem.ts
+++ b/source/vscode/src/projectSystem.ts
@@ -325,21 +325,3 @@ export async function fetchGithubRaw(
 
   return text;
 }
-
-/**
- * Opens the qsharp.json manifest in vscode for the given document URI.
- *
- * @param documentUri The URI of the document for which to update the manifest profile.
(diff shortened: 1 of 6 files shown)


source/vscode/src/projectSystem.ts
```diff
@@ -325,21 +325,3 @@ export async function fetchGithubRaw(
...
   return text;
 }
-
-/**
- * Opens the qsharp.json manifest in vscode for the given document URI.
- *
- * @param documentUri The URI of the document for which to update the manifest profile.
- * @throws Error if the manifest cannot be found or parsed.
- */
-export async function openManifestFile(documentUri: vscode.Uri): Promise<void> {
```
minestarks, line 335: Removing this, because I noticed that the caller should already have the `qsharp.json` URI available that they can use directly. We don't have to discover the manifest again. I updated the caller to open the manifest directly.

PR #2704, opened 2025-09-11
Import sparse sim into shared crate

idavis:
Moving sparse sim into shared simulators crate which re-exports it. This is an organization change to help future work.

### DIFF
--- a/source/simulators/Cargo.toml
+++ b/source/simulators/Cargo.toml
@@ -9,7 +9,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
-quantum-sparse-sim = { git = "https://github.com/qir-alliance/qir-runner", rev = "8cd6fa698521ecf96d459f88254e53dd6e5d6877" }
+quantum-sparse-sim = { workspace = true }
 
 [lints]
 workspace = true


source/simulators/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+quantum-sparse-sim = { git = "https://github.com/qir-alliance/qir-runner", rev = "8cd6fa698521ecf96d459f88254e53dd6e5d6877" }
```
swernli: Nitpick: this could use the workspace reference rather than defining its own.

source/simulators/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub use quantum_sparse_sim::QuantumSim;
```
swernli, line 4: In the future I imagine we'll want a more specific name, maybe something that identifies this as the sparse simulator, but this makes sense for now to minimize changes. Just noting it.

PR #2706, opened 2025-09-12
Don't build windows package twice on aarch64

idavis:


.ado/publish.yml
```diff
@@ -324,15 +324,19 @@ extends:
               python ./version.py
             displayName: Set version
 
+          # On anything except Windows aarch64, we can build and run tests normally
           - script: |
               python build.py --pip --no-check-prereqs --integration-tests --manylinux
             displayName: Build Platform-Dependent Py Packages
+            condition: and(ne(variables['Agent.OS'], 'Windows_NT'), ne(variables['arch'], 'aarch64'))
```
billti: This doesn't look right. It's not an inversion of the condition on line 337. If it's Linux and aarch64, this would be false, but that is not Windows aarch64.

PR #2707, opened 2025-09-13
Initial QDK Python Package

ScottCarda-MS:
Creates a `qdk` python package that acts as a lightweight meta-package for quantum-related python packages. This package wraps and re-exports these other package to provide a consistent bundle of tooling for users.
This package is still experimental, but includes the packages: qsharp, qsharp-widgets, qsharp-jupyterlab, azure-quantum, and qiskit.

The qdk package itself only contributes functions that let's users check if certain optional features are available and to require those features when writing python that depends on them. Other content found in the root qdk module are re-exports of several utilities functions and classes that get used across several different sub-modules.


billti:
See https://github.com/microsoft/qdk/pull/2708 for what might be a simpler approach. Let's discuss on Monday.

source/qdk_package/pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+license = { file = "LICENSE.txt" }
+requires-python = ">=3.9"
+dependencies = [
+  "qsharp==1.20.0",  # rely on existing package
+]
+
+[project.optional-dependencies]
+widgets = ["qsharp-widgets==1.20.0"]
+azure = ["azure-quantum==3.2.0"]
+qiskit = ["qiskit>=1.2.2,<2.0.0"]
```
minestarks, line 18: As it came up in our discussion earlier, I believe when the user specifies both optional features `qdk[azure, qiskit]`, we should also enable the `azure-quantum[qiskit]` optional feature (I think this installs more than just `qiskit`, based on what I'm seeing in [this file](https://github.com/microsoft/azure-quantum-python/blob/main/azure-quantum/requirements-qiskit.txt))  I have no clue how to express this in a `pyproject.toml` though.

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+
+"""
+
+from __future__ import annotations
+
+from importlib import import_module
+
+# Eagerly check for qsharp presence to provide a clear error message.
+try:
+    import_module("qsharp")
```
billti: This module has a hard dependency on the qsharp module. You shouldn't need to wrap any imports from qsharp in exception handling.

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+    "StateDump",
+    "ShotResult",
+    "PauliNoise",
+    "DepolarizingNoise",
+    "BitFlipNoise",
+    "PhaseFlipNoise",
+]
+
+
+def widgets_available() -> bool:
```
billti: I still don't understand what the value is for everything from this line down. Why do we need/want this?

source/qdk_package/src/qdk/estimator.py
```diff
@@ -0,0 +1,13 @@
...
+# Licensed under the MIT License.
+
+"""Shim exposing qsharp.estimator as qdk.estimator."""
+
+from __future__ import annotations
+
+try:
+    from qsharp.estimator import *
+except Exception as ex:
+    raise ImportError(
```
billti: Again, qsharp should always be installed if you have the qdk package installed.

source/qdk_package/src/qdk/qiskit.py
```diff
@@ -0,0 +1,18 @@
...
+"""Re-export shim for the optional qiskit package as qdk.qiskit.
+
+If qiskit is not installed (with the qdk[qiskit] extra), importing this
+module raises an ImportError describing how to enable it.
+"""
+
+from __future__ import annotations
+
+try:
+    from qiskit import *
```
billti: We shouldn't reexport everything from qiskit under the qdk. They should import from qiskit for the things they need from qiskit.

source/qdk_package/src/qdk/widgets.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Re-export shim for the optional widgets package as qdk.widgets.
+
+If widgets is not installed (with the qdk[widgets] extra), importing this
+module raises an ImportError describing how to enable it.
+"""
+
+from __future__ import annotations
```
billti: Why do we need this line in these modules? What does it do that we need?

source/qdk_package/README.md
```diff
@@ -0,0 +1,186 @@
...
+result = qsharp.run("{ use q = Qubit(); H(q); return MResetZ(q); }", shots=100)
+print(result)
+```
+
+Widgets (installed via jupyter extra):
+
+```python
+from qdk import widgets_available, require
+
+if widgets_available():
```
billti: I don't get this guidance at all. If the customer wants to use widgets then they should have installed that extra, then in their code they should just use `from qdk.widgets import Histogram, etc.`. There's little value in them doing a runtime check first to make sure something they should have installed is present.

source/qdk_package/README.md
```diff
@@ -0,0 +1,186 @@
...
+| -------------------- | -------- | --------------------------- | ------------------------------------------------------------------- |
+| `code`               | module   | `qsharp.code`               | Define inline Q# snippets / code objects.                           |
+| `set_quantum_seed`   | function | `qsharp.set_quantum_seed`   | Deterministic seed for quantum randomness (simulators).             |
+| `set_classical_seed` | function | `qsharp.set_classical_seed` | Deterministic seed for classical host RNG.                          |
+| `dump_machine`       | function | `qsharp.dump_machine`       | Emit a structured dump of full quantum state (simulator dependent). |
+| `dump_circuit`       | function | `qsharp.dump_circuit`       | Produce a circuit representation / diagram (when supported).        |
+| `Result`             | class    | `qsharp.Result`             | Measurement result token.                                           |
+| `TargetProfile`      | class    | `qsharp.TargetProfile`      | Target capability / profile descriptor.                             |
```
billti, line 108: `TargetProfile` and `init` should be present in the same (sub)module, as they are commonly used together. It will be annoying to have to `from qdk.qsharp import init` and then have to `from qdk import TargetProfile` so that you can call `init` with the profile specified.

ScottCarda-MS: This is confusing to me: TargetProfile is used in several different context not specific to qsharp, like openqasm. The `init` function, in contrast, is specifically for initializing the Q# interpreter. In that case, it wouldn't make sense for the `init` to be anywhere other than under `qsharp`, while the more general `TargetProfile` should go in the more general location under `qdk`. Because it is a re-export, you should also be able to access `TargetProfile` under `qdk.qsharp` as well.

billti: It's "the" interpreter, not specifically Q#. You can init, eval, and run OpenQASM code in the exact same interpreter.

billti: Take a look through https://github.com/microsoft/qdk/blob/main/samples/notebooks/openqasm.ipynb if that helps.

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: removed the import guard

source/qdk_package/src/qdk/estimator.py
ScottCarda-MS: removed the import guard

source/qdk_package/src/qdk/widgets.py
ScottCarda-MS: They are for future-proofing for forward references, but we aren't using any right now. To keep things simple I removed these.

source/qdk_package/README.md
ScottCarda-MS: I thought they would be useful to users, but if you, as someone who uses these packages more than me, doesn't find them useful then I'll defer to you on this. I've removed these helpers and the `require` function below. Users will still get the instructive error message if missing an optional feature via the try-catch's in the shims.

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: removed, see comment below

source/qdk_package/README.md
ScottCarda-MS: Ah that is news to me then. Based on that then it makes more sense for it to be lifted along with the TargetProfile class to the general location under qdk.

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+
+from . import qsharp as qsharp
+
+# Optional: use telemetry hook if present (skipped in stub/mock envs)
+try:
+    import qsharp.telemetry_events.on_qdk_import
+
+    qsharp.telemetry_events.on_qdk_import()
+except Exception:
+    pass
```
billti, line 23: Why would we ignore this exception? Just don't wrap it in exception handling at all and let it get raised if the qsharp package can't be loaded for some reason.

version.py
```diff
@@ -121,12 +124,36 @@ def update_file(file: str, old_text: str, new_text: str, is_regex: bool = False)
...
+    qdk_pyproject,
+    r"qsharp-widgets==0.0.0",
+    r"qsharp-widgets=={}".format(pip_version),
+    is_regex=True,
+)
+update_file(
+    qdk_pyproject,
+    r"qsharp-jupyterlab==0.0.0",
+    r"qsharp-jupyterlab=={}".format(jupyterlab_version),
+    is_regex=True,
```
billti: Why are the last two here `is_regex=True` but not the first two? The pattern looks the same to me.

ScottCarda-MS: removed

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: That's there so the python tests can run. The on_qdk_import doesn't exist for the existing qsharp python package, so the tests fail unless we ignore the exception when trying to load the function.

source/qdk_package/pyproject.toml
ScottCarda-MS: After several design discussions with folks, we decided to not have to deal with the dependencies unique to the intersection of azure-quantum and qiskit. If in the future we find we need to deal with these cases, we can update the package dependencies then.

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    PauliNoise,
+    DepolarizingNoise,
+    BitFlipNoise,
+    PhaseFlipNoise,
+)
+
+__all__ = [
+    "qsharp",
+    "estimator",
+    "openqasm",
```
billti: This doesn't make sense. You're not importing anything called 'openqasm' in this __init__ file. Likewise for  'estimator'. You have _submodules_ by that name, but you don't have anything in this module to reexport by that name.

source/qdk_package/src/qdk/qiskit.py
ScottCarda-MS: I don't know what the expected behavior here is, in terms of what kinds of import statements are expected to work vs not work, so I don't know what to test this against for "correctness". Are you asking me to change the star-import to `from qiskit import qiskit`?

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: removed

build.py
```diff
@@ -445,6 +449,60 @@ def run_ci_historic_benchmark():
...
+    # Reuse (or create) the pip environment so qsharp wheel can be built/installed consistently.
+    python_bin = use_python_env(qdk_python_src)
+
+    # Ensure qsharp wheel exists if the core pip package wasn't explicitly built in this run.
+    if not build_pip:
+        pip_env: dict[str, str] = os.environ.copy()
+        if platform.system() == "Darwin":
+            pip_env["ARCHFLAGS"] = "-arch x86_64 -arch arm64"
+        build_qsharp_wheel(
+            os.path.join(qdk_src_dir, "pip"), wheels_dir, python_bin, pip_env
```
billti: If the user specifies just `--qdk`, then we should only build that package. We don't do dependency tracking in the build script. That's on the user to add `--pip` if they want to (re)build the main package also.

source/qdk_package/src/qdk/__init__.py
billti: It exists in the qsharp python package build in this repo now with this PR thought, right? And the tests install the qsharp package that gets built. So why would it not exist when running the tests?

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        submodules (`qdk.qsharp`, `qdk.widgets`, etc.).
+
+Optional extras:
+    azure   -> installs `azure-quantum`, available as `qdk.azure`.
+    qiskit  -> installs `qiskit`, available as `qdk.qiskit`.
+    jupyter -> installs `qsharp-widgets` + `qsharp-jupyterlab`; exposes `qdk.widgets`.
+
+"""
+
+from . import qsharp as qsharp
```
billti: You don't need to import `qsharp as qsharp` - that's the same name so no `as ...` is necessary.

But that aside, why are you importing 'qsharp' here? It's already a submodule via the `./qsharp.py` file, and the items you want to hoist into the `qdk` module you already do via the `from...` statements on lines 28 and 29. Why is this needed?

source/qdk_package/src/qdk/qiskit.py
billti: You just need line 12. Why are your reexporting from the qiskit package? If the user needs to use things from qiskit then their code should import from qiskit, not from `qdk.qiskit`. The QDK module should be for the functionality the QDK adds.

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: removed

build.py
ScottCarda-MS: done, code block removed

source/qdk_package/src/qdk/qiskit.py
ScottCarda-MS: Okay the qiskit re-export is removed. `qdk.qiskit` now only contains the content re-exported from `qsharp.interop.qiskit`. It still requires the qiskit package be installed as that interop content depends on it. `pip install qdk[qiskit]` will still install the external qiskit package.

source/qdk_package/src/qdk/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+# utilities lifted from qsharp
+__all__ = [
+    "code",
+    "set_quantum_seed",
+    "set_classical_seed",
+    "dump_machine",
+    "dump_circuit",
+    "init",
```
billti, line 50: You are not importing `init` anywhere in this code so you can export it.

source/qdk_package/src/qdk/azure.py
```diff
@@ -0,0 +1,18 @@
...
+
+If azure is not installed (with the qdk[azure] extra), importing this
+module raises an ImportError describing how to enable it.
+"""
+
+try:
+    from azure.quantum import *
+    from azure.quantum import target as target
+    from azure.quantum import argument_types as argument_types
+    from azure.quantum import job as job
```
billti: In the 3 above lines you don't need the `.. as ..` syntax. You aren't renaming the import. Just import it.

source/qdk_package/src/qdk/__init__.py
ScottCarda-MS: fixed, properly importing `init`

source/qdk_package/src/qdk/azure.py
ScottCarda-MS: refactored this to be a sub-package instead, avoiding these aliased imports

source/qdk_package/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+Qiskit (if installed):
+
+```python
+try:
+    from qdk import qiskit  # requires: pip install qdk[qiskit]
+    qk = qiskit
+    # Example: qk.transpile(...)
+except ImportError:
+    qk = None
```
billti, line 67: I think this readme needs a lot of work, and I don't think the patterns shown here make sense (i.e. wrapping imports in exception handling and reassigning the module to a local). But lets get this in and clean it up after.

billti (approved):

PR #2709, opened 2025-09-15
Support Complex literals, arithmetic operations

swernli:


source/compiler/qsc_eval/src/lib.rs
```diff
@@ -2002,6 +2079,34 @@
...
+            let [real, imag] = array::from_fn(|i| v[i].clone());
+            let real = real.unwrap_double();
+            let imag = imag.unwrap_double();
+            match rhs_val {
+                Value::Tuple(v, Some(id)) if *id.as_ref() == StoreItemId::complex() => {
+                    let [rhs_real, rhs_imag] = array::from_fn(|i| v[i].clone());
+                    let rhs_real = rhs_real.unwrap_double();
+                    let rhs_imag = rhs_imag.unwrap_double();
+                    // (a + bi)^(c + di) = exp((c + di) * log(a + bi))
+                    // TODO: ln() uses unspecified precision, switch to hard-coded approximation?
```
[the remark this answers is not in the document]
swernli: Removed.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -2002,6 +2079,34 @@
...
+                    let log_re = 0.5 * (real * real + imag * imag).ln();
+                    let log_im = imag.atan2(real);
+                    // TODO: exp() uses unspecified precision, switch to hard-coded approximation?
```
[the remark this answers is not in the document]
swernli: Removed.

library/core/core.qs
```diff
@@ -57,5 +57,17 @@ namespace Std.Core {
         output
     }
 
-    export Length, Repeated;
+    /// # Summary
+    /// Represents a complex number by its real and imaginary components.
+    /// The first element of the tuple is the real component,
```
ScottCarda-MS: Oh just realizing this still refers to the struct as a tuple. Maybe that should be updated?

source/compiler/qsc_doc_gen/src/generate_docs/tests.rs
```diff
@@ -181,6 +181,7 @@ fn index_file_generation() {
 
         | Name | Description |
         |------|-------------|
+        | [Complex](xref:Qdk.Std.Core.Complex) | Represents a complex number by its real and imaginary components. The first element of the tuple is the real component, the second one - the imaginary component. |
```
ScottCarda-MS: Note that if you change the summary of Complex as suggested in the earlier comment, this will also need to be updated.

source/compiler/qsc_parse/src/lex/cooked/tests.rs
```diff
@@ -679,6 +680,86 @@ fn leading_zero_float() {
...
+                ),
+            ]
+        "#]],
+    );
+}
+
+#[test]
+fn leading_dot_trailing_i_int_ident() {
+    check(
+        ".1i",
```
ScottCarda-MS, line 726: This test shows that this didn't make an imaginary token. That seems wrong, no?

library/core/core.qs
swernli: Yeah, worth updating the wording here to be less about tuples and more about accessing properties of the struct.

source/compiler/qsc_parse/src/lex/cooked/tests.rs
swernli: While we do support leaving off the trailing zero for a double, as in `1.`, we do not support leaving of the leading zero. So `.1` from a parsing perspective is not a Double literal but rather a "dot" token followed by an Int literal. We have other unit tests in the file that confirm this behavior, I just wanted to make sure the interaction between that and the new literal `i` syntax produced the expected token stream, namely "dot" followed by Int literal followed by ident "i" tokens.

ScottCarda-MS: That makes sense. Thanks!

source/compiler/qsc_eval/src/tests.rs
```diff
@@ -3645,6 +3645,84 @@ fn controlled_operation_with_unique_controls_duplicate_targets_allowed() {
...
+
+#[test]
+fn complex_literal_supports_unary_positive() {
+    check_expr("", "+(3.0 + 4.0i)", &expect!["(3.0, 4.0)"]);
+}
+
+#[test]
+fn complex_literal_supports_exponentiation() {
+    check_expr("", "(3.0 + 0.0i) ^ (0.0 + 0.0i)", &expect!["(1.0, 0.0)"]);
+}
```
orpuente-MS, line 3724: Can we have a second exponentiation test with non-trivial values to assert that exponentiation works as expected?

orpuente-MS (approved):

source/compiler/qsc_eval/src/tests.rs
swernli: That is actually covered by this test over in the library: https://github.com/microsoft/qdk/blob/aed3cacf8eec8c8d91f1832395a11509f3d4cf59/library/src/tests/math.rs#L1162-L1169

The reason it needs to be there is that the test uses an approximate comparison which accounts for minor floating point differences across different machine/OS types. If we put a string comparison here with more interesting values it would fail. So the goal of this test is to ensure that eval supports the operation at all, then the library test ensures the values are correct within tolerance.

PR #2720, opened 2025-09-24
Skip qdk tests on publish

billti:


.ado/publish.yml
```diff
@@ -228,8 +228,11 @@ extends:
             python ./prereqs.py --skip-wasm && python ./version.py
           displayName: Install Prereqs and set version
 
+        # The jupyterlab and widgets packages have no tests. The qdk package has minimal tests but
+        # they also depend on the qsharp package which is platform-dependent. So we skip tests here
+        # and rely on the GitHub CI to run the few `qdk` tests on every PR.
```
DmitryVasilevsky, line 233: Can you briefly describe the immediate problem this solves, not just the justification.

DmitryVasilevsky (approved):

.ado/publish.yml
billti: It does describe it. The QDK tests depend on the qsharp package, but that is platform dependent and isn't built here in the platform independent job. (Hence the tests fail because the package doesn't exist in this build agent).

ScottCarda-MS:
Does this allow removing the guard around the telemetry call in the qdk package __init__.py? location: https://github.com/microsoft/qdk/blob/main/source/qdk_package/src/qdk/__init__.py#L17

PR #2722, opened 2025-09-24
try removing guard

ScottCarda-MS:


build.py
```diff
@@ -644,9 +644,9 @@ def run_ci_historic_benchmark():
     subprocess.run(pip_build_args, check=True, text=True, cwd=jupyterlab_src)
     step_end()
 
-if build_pip and build_widgets and args.integration_tests:
```
billti, line 654: I don't think any of these changes to build.py need to be here if you're not adding a new notebook.

billti (approved):

build.py
ScottCarda-MS: Will save that for when the sample is in PR. For now, removed

PR #2723, opened 2025-09-24
Fix panic in loop unification pass for short-circuiting expressions

### LINKED ISSUE #2695: Panic: "The type of the iterable must be either array or range, but it is an Tuple([])" at /compiler/qsc_passes/src/loop_unification.rs:356
If a short-circuiting expression is used as the iterable expression in a for-loop it panics the HIR transformations:
```qsharp
function Main() : Unit {
    let x = for i : Int in fail "" {};
}
```

swernli:
This avoids the panic in loop unification for the special case of short-circuiting expressions in the iterable of a for-loop by checking for the unit type and allowing transformation to progress using the iteration variable's type instead.

Fixes #2695


ScottCarda-MS (approved):
Interesting edge-case. Good catch!

PR #2729, opened 2025-09-30
Improved adjoint Select implementation

DmitryVasilevsky:
This check-in improves implementation of adjoin variant of Select operation (Unlookup operation). Previously it was based on May [2019 paper by Craig Gidney](https://arxiv.org/abs/1905.07682), and this proposed implementation is based on [2025 paper by Craig Gidney](https://arxiv.org/abs/2505.15917).

The implementation of Unlookup is measurement-based with phase correction via Phase Lookup operation described in the aforementioned paper. Phase Lookup is implemented by splitting address register and computing power products for both halves. Then the necessary phase corrections are applied and power products are uncomputed.

This implementation uses fast Boolean Mobius transform instead of matrix multiplication as in the paper to convert classical data from truth table representation to coefficients of an algebraic normal form over GF(2) field. This results in simpler code.

A test is added to check that address register remains in correct state after uncomputation.

This implementation may reduce required resources to uncompute lookup operation. As one example, the circuit to uncompute **one particular case** with previous and proposed approaches. This is a highly dynamic algorithm that uses results of measurements to apply gates conditionally; actual execution trace and actual savings may vary.

Before:
Unlookup
After:
Phaseup

### DIFF
--- a/library/std/src/Std/TableLookup.qs
+++ b/library/std/src/Std/TableLookup.qs
@@ -488,41 +488,21 @@ operation ApplyPhasingViaZandCZ(products1 : Qubit[], products2 : Qubit[], mask :
     Fact(Length(mask) == Length(products2) + 1, "Mask row count must match products2 length.");
     Fact(Length(mask[0]) == Length(products1) + 1, "Mask column count must match products1 length.");
 
-    // products1[0] doesn't include any qbits from the first half, so we need to apply Z instead of CZ.
-    ApplyMaskedMultitargetZ(products2, Rest(ColumnAt(0, mask)));
+    // ColumnAt(0, mask) doesn't correspond to any qubits from the first half,
+    // so we can apply Z (rather than CZ) based on mask values.
+    ApplyPauliFromBitString(PauliZ, true, Rest(ColumnAt(0, mask)), products2);
 
-    // products2[0] doesn't include any qubits from the second half, so we need to apply Z instead of CZ.
-    ApplyMaskedMultitargetZ(products1, Rest(mask[0]));
+    // mask[0] row doesn't correspond to any qubits from the second half,
+    // so we can apply Z (rather than CZ) based on mask values.
+    ApplyPauliFromBitString(PauliZ, true, Rest(mask[0]), products1);
 
-    // From the second row on, take control from the first half and apply multi-target CZ gates.
+    // From the second row on, take control from the first half and apply
+    // masked multi-target CZ gates via Controlled ApplyPauliFromBitString.
     for row in 0..Length(products1)-1 {
-        ApplyMaskedMultitargetCZ(products1[row], products2, Rest(ColumnAt(row + 1, mask)));
-    }
-}
-
-/// # Summary
-/// Applies classically controlled Z gates to multiple targets
-/// based on a mask. If mask[i] is true, Z(targets[i]) is applied.
-operation ApplyMaskedMultitargetZ(targets : Qubit[], mask : Bool[]) : Unit {
-    Fact(Length(mask) == Length(targets), "Mask length must match targets length.");
-
-    for i in 0..Length(targets)-1 {
-        if mask[i] {
-            Z(targets[i]);
-        }
-    }
-}
-
-/// # Summary
-/// Applies quantum and classically controlled Z gates to multiple targets
-/// based on a mask. If mask[i] is true, CZ(control, targets[i]) is applied.
-operation ApplyMaskedMultitargetCZ(control : Qubit, targets : Qubit[], mask : Bool[]) : Unit {
-    Fact(Length(mask) == Length(targets), "Mask length must match targets length.");
-
-    for i in 0..Length(targets)-1 {
-        if mask[i] {
-            CZ(control, targets[i]);
-        }
+        Controlled ApplyPauliFromBitString(
+            [products1[row]],
+            (PauliZ, true, Rest(ColumnAt(row + 1, mask)), products2)
+        );
     }
 }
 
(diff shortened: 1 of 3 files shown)


swernli:
Does this change the profile compatibility of the `Select` operation? By that I mean, since the new one uses measurement-based uncomputation it would only be usable in Adaptive profile programs. If `Select` was already only usable from Adaptive this isn't a change because it was never available for Base profile, but if it used to work then this could be considered a breaking change.

swernli:
>>  ... By that I mean, since the new one uses measurement-based uncomputation it would only be usable in Adaptive profile ...

>  Current code already uses measurement-based uncomputation.

Actually, I don't think that is the case. Without your changes, `Adjoint Select` was usable in Base profile:
image

But after with the changes it produces an error:
image

I think this is because previously the adjoint implementation called `Std.Intrinsic.AND` which has two implementations, one compatible with Base profile and one that uses the measurement-based uncomputation. If we want to avoid a breaking change but still have the updated version of `Select` we would need to use a similar strategy and define the operation twice using the `@Config` attribute to make sure a compatible version is present like how `AND` does.

DmitryVasilevsky:
> > > ... By that I mean, since the new one uses measurement-based uncomputation it would only be usable in Adaptive profile ...
> 
> > Current code already uses measurement-based uncomputation.
> 
> Actually, I don't think that is the case. 

OK, maybe we are both right. :) It was using measurement based uncomputation but maybe in a way suitable for base profile - i.e. nothing depended on those measurements. I will need to dig deeper.

swernli (commented):
Wow, you are right and I missed that! It turns out that this particular line you quote is exactly the problem; there's a class of code patterns where RCA can't definitively determine the compatibility of an operation (usually when it is passed as callable argument into another operation, as is done with `ForEach` and `MResetX` here). We used to err on the side of caution and reject these with an error, but too many valid programs were rejected that way. So instead we relaxed the restriction and deferred validation of these passed callables to QIR codegen (see https://github.com/microsoft/qdk/pull/1497). That means that even though the editor doesn't show an error, `Adjoint Select` with more than one data entry will fail Base profile QIR codegen.

Long story short, this is not a breaking change at all, and arguably is an improvement since the editor now shows the incompatibility at design time rather than at codegen time.

swernli (approved):

library/std/src/Std/TableLookup.qs
```diff
@@ -219,67 +268,260 @@ function MustBeFixed(result : Bool[], data : Bool[]) : Bool {
...
+
+    // products1[0] doesn't include any qbits from the first half, so we need to apply Z instead of CZ.
+    ApplyMaskedMultitargetZ(products2, Rest(ColumnAt(0, mask)));
+
+    // products2[0] doesn't include any qubits from the second half, so we need to apply Z instead of CZ.
+    ApplyMaskedMultitargetZ(products1, Rest(mask[0]));
+
+    // From the second row on, take control from the first half and apply multi-target CZ gates.
+    for row in 0..Length(products1)-1 {
+        ApplyMaskedMultitargetCZ(products1[row], products2, Rest(ColumnAt(row + 1, mask)));
```
msoeken:
```suggestion
        Controlled ApplyMaskedMultitargetZ([products1[row]], (products2, Rest(ColumnAt(row + 1, mask))));
```

You could do this and remove `ApplyMaskedMultitargetCZ` by also making `ApplyMaskedMultitargetZ` controlled. Or you do this with `Controlled ApplyPauliFromBitString`.

library/std/src/Std/TableLookup.qs
```diff
@@ -219,67 +268,260 @@ function MustBeFixed(result : Bool[], data : Bool[]) : Bool {
...
+/// The mask is obtained by applying Fast Möbius Transform to phase data
+/// and reshaping into a 2D matrix. This allows efficient quantum evaluation of
+/// the phase function using O(2^(n/2)) quantum resources instead of O(2^n).
+operation ApplyPhasingViaZandCZ(products1 : Qubit[], products2 : Qubit[], mask : Bool[][]) : Unit {
+    Fact(Length(mask) > 0, "Mask must be a non-empty array.");
+    Fact(Length(mask) == Length(products2) + 1, "Mask row count must match products2 length.");
+    Fact(Length(mask[0]) == Length(products1) + 1, "Mask column count must match products1 length.");
+
+    // products1[0] doesn't include any qbits from the first half, so we need to apply Z instead of CZ.
+    ApplyMaskedMultitargetZ(products2, Rest(ColumnAt(0, mask)));
```
msoeken:
```suggestion
    ApplyPauliFromBitString(PauliZ, true, Rest(ColumnAt(0, mask)), products2);
```

Same for the other ones.

library/std/src/Std/TableLookup.qs
```diff
@@ -219,67 +268,260 @@ function MustBeFixed(result : Bool[], data : Bool[]) : Bool {
...
+
+    // products2[0] doesn't include any qubits from the second half, so we need to apply Z instead of CZ.
+    ApplyMaskedMultitargetZ(products1, Rest(mask[0]));
```
msoeken: One could apply the operation once to `produts2 + products1` by also concatenating the masks.

library/std/src/Std/TableLookup.qs
DmitryVasilevsky: Indeed, thank you! Used ApplyPauliFromBitString and controlled version of it. ApplyMaskedMultitargetZ/CZ was used as a paper as a primitive supported by the underlying ECC, but we don't need to keep it here.

library/std/src/Std/TableLookup.qs
DmitryVasilevsky: Same as the other one.

library/std/src/Std/TableLookup.qs
DmitryVasilevsky: It is possible, makes a code a bit shorter... But doesn't change complexity and makes understanding harder. I'd prefer to keep the two calls.

PR #2731, opened 2025-09-30
Code Action for Parameterless Wrappers

ScottCarda-MS:
Adds a Refactor Code Action for creating parameterless wrappers for operations that have parameters. The wrapper will assign default values to each of the parameters, where it can, and call the original operation with those defaults, encouraging the user to fill in their desired values for these parameters.

For example:
```qsharp
operation Bar<'T>(q1 : Qubit, q2 : (Qubit, Qubit[]), d : Double, g : 'T) : Unit { ... }
```

will produce a wrapper:
```qsharp
operation BarWithDefaults() : Unit {
    // TODO: Fill out the values for the parameters
    use q1 = Qubit();
    use q2_q0 = Qubit();
    use q2_qs0 = Qubit[1];
    let q2 = (q2_q0, q2_qs0);
    let d = 0.0;
    // TODO: provide value for g (Generic parameter 'T)

    // Call original operation
    Bar(q1, q2, d, _);
}
```


source/language_service/src/code_action/wrapper_refactor.rs
minestarks, line 1: would love it if this lived under the `code_action` module instead of at this top level

source/language_service/src/code_action/wrapper_refactor.rs
```diff
@@ -0,0 +1,297 @@
...
+            }
+
+            // Determine indentation using source-local offset (package offset minus source base).
+            let local_lo = item.span.lo - source.offset;
+            let indent = line_indentation(&source.contents, local_lo);
+            let body_indent = if indent.contains('\t') {
+                format!("{indent}\t")
+            } else {
+                format!("{indent}    ")
+            };
```
minestarks, line 58: I think we have some code similar to this from auto-import text edits... can we reuse from there? 

https://github.com/microsoft/qdk/blob/95a6a6a91f4535508fa9185a3a2f00b81efb7294/source/language_service/src/completion/text_edits.rs

source/language_service/src/code_action/wrapper_refactor.rs
ScottCarda-MS: Oh in the code_action.rs file? I actually had it in there for most of the development and thought I should take it out and make it its own file to compartmentalize the code more. I was kind of thinking that individual refactors would/should be separated out to avoid that file getting bloated. Maybe I could put it and its tests under a "refactors" folder instead?

source/language_service/src/code_action/wrapper_refactor.rs
```diff
@@ -0,0 +1,297 @@
...
+    let source = source_map
+        .find_by_name(source_name)
+        .expect("source should exist");
+    let source_span = compilation.package_span_of_source(source_name);
+
+    for (_, item) in package.items.iter() {
+        if !source_span.contains(item.span.lo) || span.intersection(&item.span).is_none() {
+            continue;
+        }
+        if let ItemKind::Callable(decl) = &item.kind {
```
minestarks, line 43: Does this catch lambdas too?

source/language_service/src/wrapper_refactor.rs
```diff
@@ -0,0 +1,297 @@
...
+            }
+            body_lines.push(format!("{body_indent}// Call original operation"));
+            body_lines.push(call_line);
+
+            // We intentionally do NOT prefix the first line with `indent` because the insertion point
+            // inherits the existing line's leading whitespace. We DO append `{indent}` after the blank line
+            // so that the original operation keeps its indentation after the inserted block.
+            let wrapper_text = format!(
+                "operation {wrapper_name}() : {return_ty} {{\n{}\n{indent}}}\n\n{indent}",
+                &body_lines.join("\n")
```
minestarks: How do you respect the chosen line ending (`\r\n` or `\n`) of the existing document?

If this is difficult, I wonder if you can return this as multiple text edits (one for each line) and let VS Code stitch them together

source/language_service/src/wrapper_refactor/tests.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::{code_action, test_utils::compile_project_with_markers_no_cursor};
+use expect_test::expect;
+use qsc::line_column::{Encoding, Position, Range};
+
+fn get_wrapper_text(source: &str, op_name: &str) -> String {
+    let (compilation, _targets) =
+        compile_project_with_markers_no_cursor(&[("<source>", source)], false);
```
minestarks, line 11: You can just use the simple `compile_with_markers` helper here

source/language_service/src/code_action/wrapper_refactor/tests.rs
```diff
@@ -0,0 +1,331 @@
...
+            line: newline_count,
+            column: 0,
+        }
+    };
+    let range = Range {
+        start: Position { line: 0, column: 0 },
+        end,
+    };
+    let actions = code_action::get_code_actions(&compilation, "<source>", range, Encoding::Utf8);
+    let action = actions
```
minestarks, line 29: Do you want to validate range here too?

source/language_service/src/code_action/wrapper_refactor/tests.rs
minestarks, line 1: Could you add a test case with a `notebook` (you could use `compile_notebook_with_markers`) and see if the new code gets inserted in the correct cell and location?

source/language_service/src/code_action/wrapper_refactor/tests.rs
ScottCarda-MS: Are you talking about validating that the edit has zero length and its position is right before the existing operation?

source/language_service/src/code_action/wrapper_refactor.rs
ScottCarda-MS: I looked into trying to reuse that code for a while and wasn't finding a way to utilize it for the wrapper_refactor. The logic in the refactor is simple enough and I don't want to over complicate things in the process of trying to pull in logic from somewhere else.

source/language_service/src/code_action/wrapper_refactor.rs
ScottCarda-MS: It did, but I fixed that now so that lambdas with parameters won't trigger the code action.

source/language_service/src/wrapper_refactor.rs
ScottCarda-MS: I added some logic to detect the line ending used in the source and use that instead of just assuming '\n'.

source/language_service/src/wrapper_refactor/tests.rs
ScottCarda-MS: done

source/language_service/src/code_action/wrapper_refactor/tests.rs
ScottCarda-MS: I added some validation to the function here to assert that the edit in the action is positioned in the correct place and has zero length.

source/language_service/src/code_action/wrapper_refactor/tests.rs
ScottCarda-MS: added

source/language_service/src/wrapper_refactor/tests.rs
ScottCarda-MS: On second thought, this doesn't make sense. The tests don't use cursors (they don't use any markers), so using compile_with_markers gives an error. It works only with compile_project_with_markers_no_cursor.

source/language_service/src/code_action/wrapper_refactor.rs
minestarks: I meant inside the _module_, like it could simply go under a (new) `code_action` folder, even if it doesn't live in the same file as `code_action.rs`

source/language_service/src/wrapper_refactor/tests.rs
minestarks: I think using a marker would have worked too instead of hardcoding a position (then you would have also been testing the refactors appear in the location you meant to). 

I prefer uniformity but it's not that important

source/language_service/src/code_action/wrapper_refactor.rs
ScottCarda-MS: Ah okay, I like that organizational change. Done

PR #2732, opened 2025-10-01
Bump version to v1.21

billti:


source/vscode/changelog.md
```diff
@@ -1,5 +1,44 @@
 # QDK Changelog
 
+## v1.21.0
+
+Below are some of the highlights for the 1.21 release of the QDK.
+
+### QDK Python package
+
+With this release we are also publishing a `qdk` package to PyPI (see <https://pypi.org/project/qdk/>). This is still in the 'preview' stage as we lock down the API, but the goal is that going forward the QDK will be installed in Python via `pip install qdk`, with any optional extras needed (e.g. `pip install "qdk[jupyter,azure,qiskit]"` to add the Jupyter Notebooks, Azure Quantum, and Qiskit integration). Once installed, import from the necessary submodules (e.g. `from qdk.openqasm import compile`)
```
DmitryVasilevsky, line 9: 'also' sounds strange here. Maybe just remove it?

DmitryVasilevsky (approved):

source/vscode/changelog.md
```diff
@@ -1,5 +1,44 @@
...
+
+Please give it a try and open an issue if you have any feedback.
+
+### Complex literals
+
+The Q\# language added support for complex literals. For example,
+
+```qsharp
+function GetComplex() : Complex {
+    3. + 4.i
```
DmitryVasilevsky, line 19: 4.i, although correct, seems like a bad example to have. 3.0+4.0i would be better in my opinion.

source/vscode/changelog.md
billti: I changed the wording here a couple times before landing on this. The intent is to make it clear this is an _additional_ package, and that we'd not dropping the `qsharp` package.

source/vscode/changelog.md
billti: It uses the syntax you outline on the next line. I think it's useful to show both so folks are aware of the permutations of the float syntax they may see.

source/vscode/changelog.md
```diff
@@ -1,5 +1,44 @@
...
+}
+```
+
+This can be used in arithmetic expressions (e.g., `(2.0 + 3.0i) * (4.0 + 5.0i)`) and passed to and from Python.
```
swernli: I think it's worth adjusting this slightly to clarify that arithmetic support works with any complex, not just complex literals (GitHub doesn't like me trying to put a code sample into a suggestion block, so just writing it out below):

> Additionally, `Complex` values can now be used in arithmetic expressions directly:
> ```qsharp
> let x = 2.0 + 3.0i;
> let y = x + (4.0 - 5.0i);
> ```

swernli (approved):

PR #2734, opened 2025-10-06
Housekeeping: Tidy up spelling

ConradJohnston:
I spotted some spelling errors in comments, error messages, etc.

This PR tidies them up. 

The majority of typo fixes are in comments and so should be harmless.
One is in the output JSON format generated by resource estimation. This typo is in the value of a `description` key and so is unlikely to be matched downstream by a consumer. 

Some typos are in error messages - this may break error handling downstream, if an exact string is tested for. I'd still advocate for correct spelling being the expected condition.


DmitryVasilevsky (approved):
Great!

idavis (approved):

PR #2735, opened 2025-10-08
Better `compile` error when missing call to `init`

swernli:
We have a nice error for this case, but refactoring lost it. This makes sure the interpreter is initialized before we try to read the global config so the variable isn't empty.

### DIFF
--- a/source/pip/qsharp/_qsharp.py
+++ b/source/pip/qsharp/_qsharp.py
@@ -176,6 +176,12 @@ class Config:
     ) -> Dict[str, Dict[str, str]]:
         return {"application/x.qsharp-config": self._config}
 
+    def get_target_profile(self) -> str:
+        """
+        Returns the target profile as a string, or "unspecified" if not set.
+        """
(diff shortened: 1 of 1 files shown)


source/pip/qsharp/_qsharp.py
```diff
@@ -788,15 +788,16 @@ def compile(entry_expr: Union[str, Callable], *args) -> QirInputData:
     ipython_helper()
     start = monotonic()
     global _config
+    interpreter = get_interpreter()
```
minestarks, line 809: I think this is the online line (besides `init` and `get_interpreter()` itself) that tries to access this global directy like this... which feels like a risky pattern I'd rather not perpetuate. I wonder if instead `get_interpreter()` should return the config, and we always access it through that call to ensure it's initialized?

swernli: I adjusted this to use some friendly accessors. Turns out we only even get the config here to check if there is a currently configured taget profile, so adding an accessor for that made the code even more clear/streamlined.

PR #2739, opened 2025-10-10
Sample Notebook for Submitting Qiskit to Azure Quantum using `qdk` python

ScottCarda-MS:
Adds a sample notebook that demonstrates how to use the qdk package to submit qiskit to azure quantum.

### DIFF
--- a/samples/python_interop/submit_qiskit_circuit_to_azure.ipynb
+++ b/samples/python_interop/submit_qiskit_circuit_to_azure.ipynb
@@ -93,7 +93,6 @@
    "outputs": [],
    "source": [
     "from qiskit import QuantumCircuit\n",
-    "# import numpy as np\n",
     "\n",
     "circuit = QuantumCircuit(2, 2)\n",
     "\n",


build.py
```diff
@@ -653,8 +653,7 @@ def run_ci_historic_benchmark():
 
 if build_pip and build_widgets and args.integration_tests:
     step_start("Running notebook samples integration tests")
-    # Find all notebooks in the samples directory. Skip the basic sample and the azure submission sample, since those won't run
-    # nicely in automation.
+    # Find all notebooks in the samples directory. Skip the some of the samples since these won't run.
```
orpuente-MS:
```suggestion
    # Find all notebooks in the samples directory. Skip some of the samples since these won't run.
```

samples/python_interop/submit_qiskit_to_azure.ipynb
```diff
@@ -0,0 +1,255 @@
...
+    "3. Compile that OpenQASM 3 source to QIR with `qdk.openqasm.compile`.\n",
+    "4. Create (or attach to) an Azure Quantum workspace with `qdk.azure.Workspace`.\n",
+    "5. Select a target (e.g. a simulator such as `rigetti.sim.qvm`).\n",
+    "6. Submit the QIR payload and retrieve measurement results.\n",
+    "\n",
+    "Why use the QDK in this flow?\n",
+    "\n",
+    "- Local compilation: QIR generation happens locally—no external CLI needed.\n",
+    "- Consistency: The same compiler stack powers Q#, OpenQASM, and (via translation) Qiskit interoperability.\n",
+    "- Extensibility: The same `compile` step can later drive resource estimation or optimization passes.\n",
```
swernli: It's not quite "the same" compile step, because doing resource estimation you just call `qdk.openqasm.estimate` without using `compile` at all. It's more like the same "pattern" or "workflow." Also, I would avoid calling out optmization here, since that's something we don't currently offer through our tools.

samples/python_interop/submit_qiskit_to_azure.ipynb
```diff
@@ -0,0 +1,255 @@
...
+   "id": "d9a0f59b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qiskit import qasm3\n",
+    "from qdk.openqasm import compile\n",
+    "from qdk.azure import Workspace\n",
+    "from qdk import TargetProfile\n",
+    "\n",
+    "def submit_qiskit_to_azure(circuit, target_name, name, shots=100):\n",
```
swernli: nitpick: I would call this `submit_qiskit_circuit_to_azure` because the argument is the circuit, while "qiskit" is really the framework/library.

samples/python_interop/submit_qiskit_to_azure.ipynb
```diff
@@ -0,0 +1,255 @@
...
+    "    qc.h(0)\n",
+    "    qc.measure(0, 0)\n",
+    "    return qc\n",
+    "\n",
+    "circuit = get_parameterized_circuit()\n",
+    "circuit.assign_parameters(\n",
+    "    {\"θ\": \"0.5\"},\n",
+    "    inplace=True,\n",
+    ")\n",
+    "circuit.draw(output=\"text\")"
```
swernli: Instead of using `inplace=True` which overwrites the circuit (and then requires getting it again below), I would leave this unspecified to get the default `False` behavior which returns a new instance of the circuit. Then you can just draw the result and submit it directly. Something like this:
```python
circuit.assign_parameters(
    {"θ": "0.5"},
).draw(output="text")
```

samples/python_interop/submit_qiskit_to_azure.ipynb
```diff
@@ -0,0 +1,255 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "circuit = get_parameterized_circuit()\n",
+    "circuit.assign_parameters(\n",
+    "    {\"θ\": \"0.5\"},\n",
+    "    inplace=True,\n",
+    ")\n",
+    "\n",
+    "results = submit_qiskit_to_azure(circuit, \"rigetti.sim.qvm\", \"qiskit-parameterized-job\")\n",
```
swernli: Same thing here, I would skip `inplace=True` and use it like this:
```python
results = submit_qiskit_to_azure(
    circuit.assign_parameters(
        {"θ": "0.5"},
    ),
    "rigetti.sim.qvm",
    "qiskit-parameterized-job",
)
```
Then you don't even need to call `get_parametrized_circuit()` again, since the variable wouldn't be overwritten by the cell above.

samples/python_interop/submit_qiskit_to_azure.ipynb
ScottCarda-MS: done

samples/python_interop/submit_qiskit_to_azure.ipynb
ScottCarda-MS: done, I've updated the description markdown accordingly as well

samples/python_interop/submit_qiskit_to_azure.ipynb
ScottCarda-MS: done

samples/python_interop/submit_qiskit_to_azure.ipynb
ScottCarda-MS: Ah okay, that makes sense. I've removed this line then since it seems it would cause more trouble than help.

minestarks:
I really like committing cell outputs, so people can just pull up and read the notebook (even on GitHub) without having to set up and run stuff. @swernli @ScottCarda-MS any opinions?

minestarks:
Also, even if you are running the notebook locally, it's nice to have reference outputs so you know what to expect, before things go off the rails too much.

PR #2742, opened 2025-10-15
Various models for custom RE Python API

msoeken:
This code provides Python models that can be used with `qsharp.estimate_custom` and provide the behavior of current hard-coded models in the Rust-based system implementation of resource estimation.

_Note:_ The new test case file also implements a custom file system as in `test_project.py`. This code is now redundant and could be merged into a single file.


msoeken:
Closed as now captured by QREv3.

PR #2743, opened 2025-10-17
Circuit diagram snapshot tests (includes Node.js upgrade)

minestarks:
This prepares the codebase for an upcoming PR that will further evolve the circuit renderer. I wanted to make sure we had good tests in place.

- Added snapshot-based circuit rendering tests (`node:test` + `jsdom` + `canvas`) with representative Q# and JSON circuit inputs and stored HTML snapshots.
  - This required a small product fix to remove calls to `getBBox()` in product code since that doesn't work in tests where there's no layout.
  - Also eliminated the unnecessary UUID `id` attribute on the SVG element to make the output deterministic.
- Added a build step for code in `npm/qsharp/ux`. Removed `noEmit` from `tsconfig.json` and tweaked settings so UX tests can consume built artifacts.
  - As a result, the `circuit-vis` code now has to use explicit ESM `.js` import suffixes, like the rest of the codebase.
- Upgraded required Node version to ≥22.14 (root package.json and prereqs.py) to use native `node:test` snapshot features ([docs](https://nodejs.org/api/test.html#contextassertfilesnapshotvalue-path-options))
- Drive-by fixes in `basics.js`: Added an "all katas" test to use nested subtests for clearer output. Also removed an unnecessary console log.

**How to add a new snapshot test case:**

1. Add a `.qs` or `.qsc` file under circuits-cases.
2. Generate/update snapshots:
   - `npm test -- --test-update-snapshots`
3. Commit the new case file and its `.snapshot.html`.

### DIFF
--- a/source/npm/qsharp/package.json
+++ b/source/npm/qsharp/package.json
@@ -37,6 +37,8 @@
   "type": "module",
   "files": [
     "dist",
+    "!dist/ux",
+    "!dist/src",
     "docs",
     "lib",
     "ux"


source/npm/qsharp/test/basics.js
minestarks, line 1: Drive-by fixes in this file. Just want to make the test output look nicer.

image

source/npm/qsharp/ux/circuit-vis/draggable.ts
```diff
@@ -37,15 +36,14 @@ interface Context {
  * @param sqore         Sqore object
  */
 const createDropzones = (container: HTMLElement, sqore: Sqore): void => {
-  const svg = container.querySelector("svg[id]") as SVGElement;
+  const svg = container.querySelector("svg.qviz") as SVGElement;
```
minestarks, line 39: I removed the `id` attribute (since it was a random UUID that wasn't used for anything) and used a more appropriate selector

source/npm/qsharp/ux/circuit-vis/draggable.ts
```diff
@@ -255,18 +248,28 @@ const _addDataWires = (container: HTMLElement) => {
  * i.e. Gate 'Foo' spans on wire 0 (y=40), 1 (y=100), and 2 (y=140)
  *      Function returns [40, 100, 140]
  */
-const _wireYs = (elem: SVGGraphicsElement, wireData: number[]): number[] => {
-  const { y, height } = elem.getBBox();
```
minestarks, line 259: The intent here is to remove the call to `getBBox`, which doesn't work in tests where you don't have an actual layout engine.

Luckily, we only used `getBBox` to re-derive information we should already have when we're generating the circuit. So instead, I'm pre-calculating this information and saving them as `data-` attributes in the elements.

source/npm/qsharp/ux/circuit-vis/formatters/formatUtils.ts
```diff
@@ -98,12 +98,21 @@ export const circle = (
...
-  circle(x, y, radius, "control-dot");
+export const controlDot = (
+  x: number,
+  y: number,
+  wireYs: number[],
+): SVGElement => {
+  const radius = 5;
+  const dot = circle(x, y, radius, "control-dot");
+  dot.setAttribute("data-wire-ys", JSON.stringify(wireYs));
+  dot.setAttribute("data-width", `${2 * radius}`);
```
minestarks, line 113: This is an example of where we set a `data-` attribute _before_ rendering the circuit, instead of using `getBBox` _after_ rendering to compute this information.

source/npm/qsharp/ux/circuit-vis/formatters/formatUtils.ts
```diff
@@ -121,7 +130,7 @@ export const box = (
   y: number,
   width: number,
   height: number,
-  className = "gate-unitary",
+  className: string,
```
minestarks, line 133: I removed this parameter default, because it just looked sus. Indeed, the `dashedBox` elements were getting the default class name of `gate-unitary`, and I'm really not sure if that was intentional. I left that behavior as is, but just made it explicit for now. We'll worry about it when we start using dashed boxes.

(Originally, I was just trying to figure out which elements had the `gate-unitary` class, since we seem to be using the class name as a selector in `draggable.ts` . The implicit defaults made this task much harder.)

source/npm/qsharp/ux/tsconfig.json
```diff
@@ -1,8 +1,11 @@
 {
   "compilerOptions": {
-    "module": "CommonJS",
+    "outDir": "../dist",
+    "declaration": true,
+    "sourceMap": true,
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
     "target": "ES2020",
-    "noEmit": true,
```
billti, line 5: Do we build these just for the tests now? I see they build to the ./dist folder, so it means they will now be a new artifact of the shipping npm package, is that right? Previously they were just source, and didn't have their own build, as they were only consumed and included in the build of other components (such as jupyter widgets or VS Code)

source/npm/qsharp/ux/tsconfig.json
```diff
@@ -1,8 +1,11 @@
 {
   "compilerOptions": {
-    "module": "CommonJS",
+    "outDir": "../dist",
+    "declaration": true,
+    "sourceMap": true,
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
billti, line 7: I spent so many hours fiddling with these settings to fix issues across environment. JS (and TS) modules are such a pain. Are you sure this is working as intended in every place where the UX source is consumed?

source/npm/qsharp/package.json
```diff
@@ -25,11 +25,12 @@
     "./ux": "./ux/index.ts"
   },
   "scripts": {
-    "build": "npm run docs && npm run generate && npm run build:tsc",
+    "build": "npm run docs && npm run generate && npm run build:tsc && npm run build:ux",
```
billti, line 28: The above comment was meant to be here re building this now. (Versus just type checking it)

source/npm/qsharp/ux/tsconfig.json
minestarks: And your efforts are appreciated. Thinking through this again:

- I verified VS Code, playground and Python widgets manually, they work fine.

- Previously, `noEmit: true` was set, so this `tsconfig.json` was only used for checking. So these `tsconfig.json` settings, by themselves, shouldn't cause any issues with external code.

- *BUT* I did have to make source changes to `circuit-vis`, as you probably saw. That code came from an older repo that didn't use current ES import conventions. I had change all the imports to use file extensions, like `import "./foo.js"`  . So I suppose that does pose a hypothetical risk for externals consuming `circuit-vis` directly. I can only think of one (not QCOM, since they don't use the circuit code) and I can follow up with them directly.

Did I miss anything?

source/npm/qsharp/ux/tsconfig.json
minestarks: That's right.

I put them in `dist` because that just seemed to align the best with the current code (all the tests under `npm/qsharp/test` reference product code from `../dist`) . 

Shipping them does feel a bit unnecessary though. 

Perhaps I can exclude `dist/ux` from publishing in `package.json` ? Would that fit with your vision of how this is supposed to work?

billti: Let's exclude it. I assume they might be quite large if parts of them pull in markdown and latex rendering, preact, etc., and the UX pieces aren't intended to be buit and distributed here, just source for other builds to link to.

source/npm/qsharp/ux/tsconfig.json
billti: I started down the rabbit hole and.... I really don't want to have to try and page all that complexity back in. If you tested it and it works... 🤷 . Ship it.

source/npm/qsharp/ux/tsconfig.json
minestarks: Excluded `dist/ux` and `dist/src` (which had crept in via references from `ux`, so... glad you spotted it :/ )

After doing a proper diff of the resulting `.tgz`,  this is exactly what changed from `main`:

(copilot's words, not mine obviously)


### Remaining Differences (20 files - all intentional functional changes)

**1. package.json (3 differences):**
- ✅ Has `build:ux` script (intentional - needed to compile ux code)
- ✅ Build script includes `&& npm run build:ux` (intentional)
- ✅ Has the `!dist/ux` and `!dist/src` exclusions (our fix)

**2. `ux/tsconfig.json` (5 differences):**
- ✅ Compiles to `dist/` with `outDir`, `declaration`, `sourceMap` (intentional)
- ✅ Uses `module: "nodenext"` with `moduleResolution: "nodenext"` (intentional - needed for `.js` extensions)
- ✅ Removed `noEmit: true` (intentional - we want to emit compiled files)

**3. Import statements in 18 `ux/` TypeScript files:**
- ✅ All imports use `.js` extensions (e.g., `from "./circuit.js"` instead of `from "./circuit"`)
- This is required by TypeScript when using `module: "nodenext"`

**4. Functional code changes in circuit visualization:**
- ✅ **`draggable.ts`**: New wire tracking using `data-wire-ys` attributes, different element width calculations
- ✅ **`formatUtils.ts`**: Modified `controlDot()` signature to accept `wireYs` parameter
- ✅ **`gateFormatter.ts`**: Gates now store `data-wire-ys` and `data-width` attributes
- ✅ **`sqore.ts`**: Different SVG selector strategy (`svg.qviz` class vs `svg[id]` attribute)
- ✅ **`utils.ts`**: Removed `createUUID()` function (moved to main branch, not in circuit branch)

PR #2744, opened 2025-10-20
#1114 Array error messages for comma issues

### LINKED ISSUE #1114: Improve error message for arrays with missing commas

joesho112358:
Looking to solve and close #1114. Definitely open to suggestions since this is my first toe dip into Rust at this level.

### DIFF
--- a/source/compiler/qsc_parse/src/expr.rs
+++ b/source/compiler/qsc_parse/src/expr.rs
@@ -359,23 +359,25 @@ fn expr_array_core(s: &mut ParserContext) -> Result<Box<ExprKind>> {
     };
 
     if token(s, TokenKind::Comma).is_err() {
-        let peek = s.peek();
-        let peek_kind = peek.kind;
-        if peek_kind.eq(&TokenKind::Close(Delim::Bracket)) {
-            return Ok(Box::new(ExprKind::Array(vec![first].into_boxed_slice())));
+        match s.peek().kind {
+            TokenKind::Close(Delim::Bracket) => {
+                return Ok(Box::new(ExprKind::Array(vec![first].into_boxed_slice())));
+            }
+            TokenKind::Semi => {
+                return Err(Error::new(ErrorKind::Token(
+                    TokenKind::Close(Delim::Bracket),
+                    s.peek().kind,
+                    s.peek().span,
+                )));
+            }
+            _ => {
+                return Err(Error::new(ErrorKind::Token(
+                    TokenKind::Comma,
+                    s.peek().kind,
+                    s.peek().span,
+                )));
+            }
         }
-        if peek_kind.eq(&TokenKind::Semi) {
-            return Err(Error::new(ErrorKind::Token(
-                TokenKind::Close(Delim::Bracket),
-                peek_kind,
-                peek.span,
-            )));
-        }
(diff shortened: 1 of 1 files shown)


source/compiler/qsc_parse/src/expr.rs
```diff
@@ -349,7 +349,11 @@ fn expr_if(s: &mut ParserContext) -> Result<Box<ExprKind>> {
 fn expr_array(s: &mut ParserContext) -> Result<Box<ExprKind>> {
     token(s, TokenKind::Open(Delim::Bracket))?;
     let kind = expr_array_core(s)?;
-    token(s, TokenKind::Close(Delim::Bracket))?;
+    if Ok::<Box<ExprKind>, Error>(kind.clone()).is_ok() {
+        token(s, TokenKind::Close(Delim::Bracket))?;
+    } else {
+        token(s, TokenKind::Comma)?;
```
swernli: Turns out this change isn't needed. The line `Ok::<Box<ExprKind>>, Error(kind.clone()).is_ok()` always resolves to true (and incurs a copy) so can be just removed. Since that means the only code executing here was the original check for the closing braket, this whole function can be reverted back to how it was before.

source/compiler/qsc_parse/src/expr.rs
```diff
@@ -359,7 +363,23 @@ fn expr_array_core(s: &mut ParserContext) -> Result<Box<ExprKind>> {
...
+                TokenKind::Close(Delim::Bracket),
+                peek_kind,
+                peek.span,
+            )));
+        }
+        return Err(Error::new(ErrorKind::Token(
+            TokenKind::Comma,
+            peek_kind,
+            peek.span,
+        )));
```
swernli: A more Rust-idiomatic way to do this would be with a match statement:

```suggestion
        match s.peek().kind {
            TokenKind::Close(Delim::Bracket) => {
                return Ok(Box::new(ExprKind::Array(vec![first].into_boxed_slice())));
            }
            TokenKind::Semi => {
                return Err(Error::new(ErrorKind::Token(
                    TokenKind::Close(Delim::Bracket),
                    s.peek().kind,
                    s.peek().span,
                )));
            }
            _ => {
                return Err(Error::new(ErrorKind::Token(
                    TokenKind::Comma,
                    s.peek().kind,
                    s.peek().span,
                )));
            }
        }
```

swernli (changes requested):
Thanks for the PR! The logic you have in `expr_array_core` does fix the issue, though I have some feedback on how to make that more idiomatic for Rust. The changes in `expr_array` turn out not to be needed, so can just be reverted. With those updates, I think this would be ready to merge!

source/compiler/qsc_parse/src/expr.rs
joesho112358: Should be back to the way it was. thanks for taking a look!

PR #2748, opened 2025-10-20
Fix Webview and Circuit Editor Left Padding

ScottCarda-MS:
Updates the CSS for the webviews and Circuit Editor in div for left-padding correction.

### DIFF
--- a/source/vscode/src/webview/webview.tsx
+++ b/source/vscode/src/webview/webview.tsx
@@ -176,13 +176,11 @@ function onRowDeleted(rowId: string) {
 function App({ state }: { state: State }) {
   const onFilter = () => undefined;
 
-  let content: preact.JSX.Element;
   switch (state.viewType) {
     case "loading":
-      content = <div>Loading...</div>;
-      break;
+      return <div>Loading...</div>;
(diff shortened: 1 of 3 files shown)


swernli (approved):

source/vscode/src/webview/webview.tsx
```diff
@@ -214,23 +217,30 @@ function App({ state }: { state: State }) {
...
       );
+      break;
     default:
       console.error("Unknown view type in state", state);
-      return <div>Loading error</div>;
+      content = <div>Loading error</div>;
+      break;
   }
+
+  return <div style={{ paddingLeft: "8px" }}>{content}</div>;
```
billti: We already have a CSS file specifically for the webview body. Why not just add the padding in the CSS there for the entire view, rather than in code on the specific component?

https://github.com/microsoft/qdk/blob/main/source/vscode/src/webview/webview.css

ScottCarda-MS: Adding the change to the * block affected way too many things, but adding it to the html block looks like it is working great! An even simpler solution!

PR #2749, opened 2025-10-20
Updated controlled ApproximatelyMultiplexZ, added test

### LINKED ISSUE #1104: State preparation could be better

DmitryVasilevsky:
This PR implements the old suggestion by @msoeken and resolves the last remaining item in the issue https://github.com/microsoft/qdk/issues/1104.

The suggestion was "Instead of performing Controlled X apply an AND chain. I think you added such an operation for the PhaseUp..."

Also introduces a test for controlled version of preparation.

Closes #1104

### DIFF
--- a/library/std/src/Std/StatePreparation.qs
+++ b/library/std/src/Std/StatePreparation.qs
@@ -13,6 +13,7 @@ import
     Std.Arithmetic.ApplyIfGreaterLE,
     Std.Arithmetic.ApplyIfGreaterL,
     Std.Arithmetic.ReflectAboutInteger,
+    Std.ArithmeticUtils.ApplyAsSinglyControlled,
     Std.Convert.IntAsBigInt,
     Std.Math.*,
     Std.Arrays.*;
@@ -351,7 +352,8 @@ operation ApproximatelyMultiplexZ(
         ApproximatelyMultiplexZ(tolerance, coefficients0, control, target);
         if AnyOutsideToleranceD(tolerance, coefficients1) {
             within {
-                Controlled X(controlRegister, target);
+                // Reduce circuit complexity by using aux qubits.
+                Controlled ApplyAsSinglyControlled(controlRegister, (X, target));
             } apply {
                 ApproximatelyMultiplexZ(tolerance, coefficients1, control, target);
             }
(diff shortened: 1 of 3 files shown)


library/std/src/Std/StatePreparation.qs
```diff
@@ -351,7 +352,8 @@ operation ApproximatelyMultiplexZ(
         ApproximatelyMultiplexZ(tolerance, coefficients0, control, target);
         if AnyOutsideToleranceD(tolerance, coefficients1) {
             within {
-                Controlled X(controlRegister, target);
+                // Reduce circuit complexity by using aux qubits.
+                Controlled ApplyAsSinglyControlled(controlRegister, (X, target));
```
msoeken, line 356: One can still save a factor of 2x by only doing the forward direction of the multi-controlled X gate (chain of computing ANDs) and then the backward direction (chain of uncomputing ANDs) afterwards. However, this will require to allocate the helper qubits in this operation.

For a 3-qubit controlRegister the trace would be as follows:

```qsharp
use helper = Qubit[1];

And(controlRegister[0], controlRegister[1], helper[0]);
And(controlRegister[1], helper[0], target);

ApproximatelyMultiplexZ(tolerance, coefficients1, control, target);

Adjoint And(controlRegister[1], helper[0], target);
Adjoint And(controlRegister[0], controlRegister[1], helper[0]);
```

msoeken (changes requested):

msoeken:
> What impact does this have on the resource counts? The two approaches look almost the same in terms of how they apply `AND` under the covers, just one recursively while the other is iterative, so I'm not sure why one would be preferred over the other.

Maybe it's worth to add tests for logical counts to the test suite for some operations like these ones. We can use the `LogicalCounst` backend in Rust or the `logical_counts` function in Python, depending on where the tests live.

PR #2751, opened 2025-10-21
Cirq Sample Notebook for Azure Submission

ScottCarda-MS:
Adding a Cirq sample notebook to our repo to demo how to submit cirq circuits to Azure Quantum.


build.py
```diff
@@ -659,6 +659,8 @@ def run_ci_historic_benchmark():
         for dp, _, filenames in os.walk(samples_src)
         for f in filenames
         if f.endswith(".ipynb")
+        # Skip all Cirq sample notebooks under the cirq_samples folder
```
DmitryVasilevsky: Can you provide a reason why we skip them.

samples/python_interop/cirq_samples/intro.ipynb
```diff
@@ -0,0 +1,235 @@
...
+    "# %pip install cirq --quiet"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c1428ea5",
+   "metadata": {},
+   "source": [
+    "## 2. Imports\n",
+    "We first import Cirq."
```
DmitryVasilevsky: Suggestion: Cirq -> Cirq Python package

samples/python_interop/cirq_samples/intro.ipynb
```diff
@@ -0,0 +1,235 @@
...
+    "    ]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "52049f16",
+   "metadata": {},
+   "source": [
+    "## 4. Building a Circuit\n",
+    "A `cirq.Circuit` organizes moments (time steps). We'll print the raw operations and the formatted circuit."
```
DmitryVasilevsky: Suggest mentioning that automatic scheduling is used (vs. manual creation of moments).

samples/python_interop/cirq_samples/intro.ipynb
```diff
@@ -0,0 +1,235 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fca0117b",
+   "metadata": {},
+   "source": [
+    "## 8. Next Steps\n",
+    "\n",
+    "You now know how to: create a qubit, apply a gate, build a circuit, visualize it, and simulate measurements. Continue with the teleportation notebook to explore multi-qubit circuits and classical control."
```
DmitryVasilevsky: Suggestion: _reference_ or _use_ a qubit instead of _create_. Actually, not sure what's the official terminology.

samples/python_interop/cirq_samples/introduction.ipynb
```diff
@@ -0,0 +1,235 @@
+{
```
DmitryVasilevsky: Suggest naming this file _introduction_ (without shortening).

samples/python_interop/cirq_samples/teleportation.ipynb
```diff
@@ -0,0 +1,188 @@
+{
```
DmitryVasilevsky: Suggest naming this file just _teleportation_.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
```diff
@@ -0,0 +1,188 @@
...
+    "def simple_teleportation() -> Sequence[cirq.Operation]:\n",
+    "    alice = cirq.NamedQubit('alice')\n",
+    "    bob = cirq.NamedQubit('bob')\n",
+    "    program: Sequence[cirq.Operation] = []\n",
+    "    # 1. Create entangled pair shared by Alice and Bob\n",
+    "    program.extend(PrepareBellPair(alice, bob))\n",
+    "    # 2. Prepare a message qubit (state to teleport)\n",
+    "    msg = cirq.NamedQubit('msg')\n",
+    "    program.append(cirq.rx(0.7)(msg))  # Example arbitrary state\n",
+    "    # 3. Entangle message with Alice's half\n",
```
DmitryVasilevsky: _part_ sounds strange. Maybe just _qubit_?

samples/python_interop/cirq_samples/teleportation_basics.ipynb
```diff
@@ -0,0 +1,188 @@
...
+    "    program.extend(PrepareMessage(alice, msg))\n",
+    "    # 4. Measure Alice + msg to obtain classical bits\n",
+    "    program.append(cirq.measure(msg, key='b1'))\n",
+    "    program.append(cirq.measure(alice, key='b2'))\n",
+    "    # 5. Classical feed-forward corrections on Bob\n",
```
DmitryVasilevsky: For those who sees teleportation for the first time, maybe it would help to explicitly mark a place where classical data is communicated to the Bob's site.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
```diff
@@ -0,0 +1,188 @@
...
+    "    program.append(cirq.measure(bob, key='result'))\n",
+    "    return program"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2882e3d0",
+   "metadata": {},
+   "source": [
+    "## 4. Circuit Construction and Display"
```
DmitryVasilevsky: This is the first time moment marks appear on the diagram. Suggest commenting what they are.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
```diff
@@ -0,0 +1,188 @@
...
+   "outputs": [],
+   "source": [
+    "import math\n",
+    "\n",
+    "sim = cirq.Simulator()\n",
+    "result = sim.run(cirq.Circuit(simple_teleportation()), repetitions=1000)\n",
+    "hist = result.histogram(key='result')\n",
+    "print(\"Histogram of Bob's measurements (teleported state):\")\n",
+    "print(hist)\n",
+    "# Expected probabilities for |ψ> = Rx(0.7)|0>\n",
```
DmitryVasilevsky: Can we use LaTeX here?

samples/python_interop/cirq_samples/custom_gates.ipynb
```diff
@@ -0,0 +1,241 @@
...
+    "print(circuit)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8c98e44a",
+   "metadata": {},
+   "source": [
+    "## 5. Full Decomposition\n",
+    "We can ask Cirq to decompose all gates to primitive operations."
```
DmitryVasilevsky: Suggest mentioning where it gest the set of 'primitive operations'.

samples/python_interop/cirq_samples/custom_gates.ipynb
```diff
@@ -0,0 +1,241 @@
...
+   "id": "c9e98407",
+   "metadata": {},
+   "source": [
+    "## 6. Selective Decomposition\n",
+    "\n",
+    "The `keep` predicate controls where recursive decomposition stops:\n",
+    "- If `keep(op)` returns True, the operation is left as-is.\n",
+    "- If it returns False, Cirq tries to expand the operation via its decomposition methods.\n",
+    "\n",
+    "In the example below, we decompose only the custom gates (`BellPairGate`, `MessagePrepGate`) by returning False for them and True for everything else."
```
DmitryVasilevsky: Suggest mentioning that the diagram looks like the diagram from the teleportation notebook without custom gates.

samples/python_interop/cirq_samples/custom_gates.ipynb
```diff
@@ -0,0 +1,241 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6faa0c1e",
+   "metadata": {},
+   "source": [
+    "## 7. Simulation\n",
+    "\n",
+    "We run the teleportation circuit several times to confirm Bob receives (approximately) the same state that was prepared on the message qubit. Because we removed the final inverse rotation, Bob's measurement statistics should reflect |ψ> = Rx(0.7)|0>. Expected probabilities: P(0) ≈ cos^2(0.35) ≈ 0.88, P(1) ≈ sin^2(0.35) ≈ 0.12."
```
DmitryVasilevsky: If you print the expected probabilities maybe there's no need to have them as percentages in the hard-coded string. Also, maybe adding this calculation to the teleportation notebook would help.

samples/python_interop/cirq_samples/custom_gates.ipynb
```diff
@@ -0,0 +1,241 @@
...
+   "execution_count": null,
+   "id": "1bdc4615",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import math\n",
+    "\n",
+    "sim = cirq.Simulator()\n",
+    "result = sim.run(cirq.Circuit(simple_teleportation()), repetitions=1000)\n",
+    "print(\"Histogram of teleported state measurements (expected ~88% zeros, ~12% ones):\")\n",
```
DmitryVasilevsky: If you print the expected probabilities maybe there's no need to have them as percentages in the hard-coded string. Also, maybe adding this calculation to the teleportation notebook would help.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
```diff
@@ -0,0 +1,188 @@
...
+    "print(f\"Expected P(0)≈{p0:.3f}, P(1)≈{p1:.3f}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "73deefb5",
+   "metadata": {},
+   "source": [
+    "## 6. Interpreting Results\n",
+    "If teleportation succeeds, Bob's qubit should be in the same state |ψ> that was originally prepared on the message qubit. Since we prepared |ψ> = Rx(0.7)|0>, the measurement statistics should reflect the amplitudes of that state: P(0) ≈ cos^2(0.35) ≈ 0.88 and P(1) ≈ sin^2(0.35) ≈ 0.12."
```
DmitryVasilevsky: Can we use LaTeX here and other formulas in the notebook? (If even possible)

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,213 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "6f08a1c8",
+   "metadata": {},
+   "source": [
+    "# Moments, Barriers, and Classical Control in Cirq\n",
+    "\n",
+    "This rewritten notebook teaches:\n",
```
DmitryVasilevsky: rewritten?

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,213 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "1f7e477e",
+   "metadata": {},
+   "source": [
+    "## 2. Core Concepts\n",
+    "\n",
+    "A `cirq.Moment` is a parallel layer of operations with no shared qubits. Cirq packs a flat list of operations greedily into moments based only on qubit conflicts (and basic measurement ordering), not on classical-control dependencies.\n",
+    "\n",
+    "Classically controlled operations (e.g. `op.with_classical_controls('key')`) use measurement outcomes at execution time, but they do not automatically create a separate moment in the printed diagram.\n",
```
DmitryVasilevsky: Cirq should work correctly for the variant without custom gates. If you just use standard gates even with classical control, you'll get correct behavior. So, this statement should probably be reworded. This is likely due to the order in which classical dependency tracking, decomposition and scheduling is done in Cirq.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,213 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "699db40e",
+   "metadata": {},
+   "source": [
+    "## 3. Auto-Packed vs. Barrier vs. Explicit Moments\n",
+    "We construct three circuits:\n",
+    "1. Auto-packed (no barrier): may merge feed-forward corrections earlier.\n",
+    "2. Auto-packed with barrier: forces a visible layer boundary.\n",
+    "3. Explicit moments: manually defined phases, independent of greedy packing."
```
DmitryVasilevsky: I believe using explicit moments is the recommended approach. Barrier gate is a hack. Also the first circuit is simply incorrect in all senses.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,213 @@
...
+    "    bob = cirq.NamedQubit('bob')\n",
+    "    msg = cirq.NamedQubit('msg')\n",
+    "    ops: Sequence[cirq.Operation] = []\n",
+    "    ops.append(BP.on(alice, bob))\n",
+    "    ops.append(cirq.rx(0.7)(msg))\n",
+    "    ops.append(MPrep.on(msg, alice))\n",
+    "    ops.append(cirq.measure(msg, key='b1'))\n",
+    "    ops.append(cirq.measure(alice, key='b2'))\n",
+    "    if use_barrier:\n",
+    "        ops.append(BarrierGate([msg, alice, bob]).on(msg, alice, bob))\n",
```
DmitryVasilevsky: Suggest changing barrier gate constructor to take just the number of qubits. This call looks kind of strange as it is now - why are we passing same qubits two times?

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,213 @@
...
+    "\n",
+    "# --- Helper: Bell-basis prep between message and Alice ---\n",
+    "class MessagePrepGate(cirq.Gate):\n",
+    "    def _num_qubits_(self): return 2\n",
+    "    def _decompose_(self, qs):\n",
+    "        m, a = qs\n",
+    "        yield cirq.CNOT(m, a)\n",
+    "        yield cirq.H(m)\n",
+    "    def _circuit_diagram_info_(self, args): return 'ME_M','ME_A'\n",
+    "MPrep = MessagePrepGate()\n",
```
DmitryVasilevsky: That's a lot of code for a single cell. Maybe split the first two out and mention they are already considered in the custom_gates notebook?

samples/python_interop/cirq_samples/submission_to_azure.ipynb
```diff
@@ -0,0 +1,191 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "1478e88a",
+   "metadata": {},
+   "source": [
+    "## Notes\n",
+    "- Ensure all measurement keys appear before any classical condition usage if you introduce classical controls.\n",
+    "- Multi-target or custom gates may need full decomposition before submission if they produce unsupported classical constructs.\n",
+    "- Keep circuits minimal when first validating workspace access; expand complexity gradually.\n",
+    "- Complex classical control expressions created with SymPy symbols are not supported when generating QASM, and therefore can't be submitted to Azure using the method shown in this sample.\n",
```
DmitryVasilevsky: Should we even mention SymPy? We don't have examples for that now.

build.py
ScottCarda-MS: cirq is not installed in the build machines. Missing dependencies seems to be the reason for why the other notebooks were skipped.

DmitryVasilevsky: I mean, add this to the comment :)

samples/python_interop/cirq_samples/custom_gates.ipynb
ScottCarda-MS: Hmm, because this example is doing decomposition, while the teleportation notebook wasn't, this circuit looks significantly different from that one. It decomposes things much more than they were in the teleportation notebook.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: I haven't found this to be true. The cirq doesn't seem to be respecting classical control measurement keys when automatically determining timing. I'm seeing that problematic behavior here in this example. The output of the automatic timing is:
```
Auto-packed (no barrier):
                       ┌────────────┐
alice: ───BP────────────ME_A────────────M('b2')───────
          │             │
bob: ─────BP────────────┼───Z^b1X^b2────M('result')───
                        │
msg: ─────Rx(0.223π)────ME_M────────────M('b1')───────
                       └────────────┘
```

samples/python_interop/cirq_samples/submission_to_azure.ipynb
ScottCarda-MS: Fair enough. Removed.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: removed

samples/python_interop/cirq_samples/custom_gates.ipynb
ScottCarda-MS: They don't need to be rewritten as hard-coded strings in the code. But I do like having the expected state stated in the markdown cell. Hard to describe why, but I feel like it is missing something without it.

samples/python_interop/cirq_samples/custom_gates.ipynb
ScottCarda-MS: I responded in the comment below.

samples/python_interop/cirq_samples/custom_gates.ipynb
ScottCarda-MS: I added a more detailed description of how cirq determines the bases set of gates. Let me know if that reads correctly to you.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
ScottCarda-MS: I added a more detailed description of moments in cirq and how they appear. Let me know if that description makes sense.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
ScottCarda-MS: Good idea! Done.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
ScottCarda-MS: done

samples/python_interop/cirq_samples/intro.ipynb
ScottCarda-MS: went with "reference"

samples/python_interop/cirq_samples/intro.ipynb
ScottCarda-MS: Added a bit of a description for that.

samples/python_interop/cirq_samples/intro.ipynb
ScottCarda-MS: done

build.py
ScottCarda-MS: done

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: I had thought the barrier technique was useful for users to know, especially if they are used to working with barriers, but maybe if it is not the officially supported way of doing things, we shouldn't teach new learners to do it. Do you think we should remove the barrier stuff entirely?

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: Yah this block did get to be a bit too long. I'll split out the first three classes as they are used in the implementation of the teleport algorithm, whereas the rest is about the three different approaches to timing/scheduling.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: Yah that is simpler. Done.

samples/python_interop/cirq_samples/intro.ipynb
DmitryVasilevsky: BTW, Cirq does have a qubit manager that allocates and frees qubits for use in a circuit. I think it would be appropriate to use words like _allocate_ or _create_ when using that qubit manager. If you use NamedQubit or LinearQubit or similar, you kind of just grab existing qubit. I.e. If you say LinearQubit(5) and then LinearQubit(5) again, you'll get the same qubit. That's why I think _reference_ is more appropriate.
(Just close this)

samples/python_interop/cirq_samples/custom_gates.ipynb
DmitryVasilevsky: With only custom gates decomposed it looks exactly the same as far as I can see. I'm talking about section 6.

samples/python_interop/cirq_samples/custom_gates.ipynb
DmitryVasilevsky: Yes, I was talking about the code.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
DmitryVasilevsky: If you look at the diagram with classical control lines present (as in the very first teleport example), the scheduling is correct. Custom gates break scheduling. More advanced classical control also breaks scheduling. So, we probably should say something like in simple cases it works, but more complex cases require manual scheduling. Right now it sounds like scheduling is completely broken in Cirq, which isn't true.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
DmitryVasilevsky: I think we can call barrier gate a "workaround". It may not be recommended officially, because it introduces _quantum_ dependencies, where there's none. But there's no way to not support it.

samples/python_interop/cirq_samples/custom_gates.ipynb
ScottCarda-MS: Ah you are right, I was looking at section 5. I've added a note about it being the same circuit as in teleportation.

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
ScottCarda-MS: Oh I see what you are saying here. I've reworked a lot of the wording around this notebook. Please take a look and see if it makes sense.

samples/python_interop/cirq_samples/teleportation_basics.ipynb
ScottCarda-MS: Sure can! Done.

samples/python_interop/cirq_samples/introduction.ipynb
ScottCarda-MS: done

samples/python_interop/cirq_samples/teleportation.ipynb
ScottCarda-MS: done

samples/python_interop/cirq_samples/classical_control_and_barriers.ipynb
```diff
@@ -0,0 +1,273 @@
...
+    "import cirq\n",
+    "from typing import Sequence"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f8ff5a23",
+   "metadata": {},
+   "source": [
+    "## 2. Core Concepts\n",
```
DmitryVasilevsky: Suggest making it less verbose and restructure. Provided feedback in a personal message.

ScottCarda-MS: Dropping this sample notebook in favor of streamlined Cirq content. Thanks for taking a look at it anyways though. It was educational for me.

PR #2754, opened 2025-10-23
Enable Qiskit 2.0 support

### LINKED ISSUE #2486: Qiskit 2 support
The latest version of Qiskit had some more breaking changes to absorb. We should update to handle this.

idavis:
Closes #2486 

This work introduces a breaking change effectively removing the `QirTarget` class based on issues and doc updates from Qiskit, see https://github.com/Qiskit/qiskit/issues/15142#issuecomment-3422411808 . 

PR Summary: Qiskit 2.x enablement

- Broadened optional qiskit dependency ranges (pyproject.toml, qdk_package/pyproject.toml) so our pip and meta packages install with either Qiskit `[1.22, 2.0)` or `[2.0,3.0)`.
- Refactored QirTarget into a factory (create_target) and updated `BackendBase` to request it, matching the Qiskit 2.0 Target API and allowing callers to inject a qubit count without subclassing.
- Hardened integration and resource-estimator tests for multi-version runs: install requirements once, then execute each suite against both Qiskit major versions, capture version-dependent logical-count baselines, and rely on `skip_transpilation` where needed.
- Adjusted notebook automation so only notebooks with “qiskit” in their filename execute under the version matrix; other samples still run once per build.
- Trimmed shared test requirements (Qiskit now supplied via the build matrix) and touched the Qiskit sample notebook metadata to keep it aligned.

### DIFF
--- a/source/pip/qsharp/interop/qiskit/backends/backend_base.py
+++ b/source/pip/qsharp/interop/qiskit/backends/backend_base.py
@@ -120,7 +120,6 @@ class BackendBase(BackendV2, ABC):
             qiskit_pass_options (Dict): Options for the Qiskit passes.
             transpile_options (Dict): Options for the transpiler.
             qasm_export_options (Dict): Options for the QASM3 exporter.
-            num_qubits: optional number of qubits for the default target if not supplied
             **options: Additional keyword arguments to pass to the
                 execution used by subclasses.
         """


source/pip/tests-integration/interop_qiskit/test_qir.py
```diff
@@ -147,19 +147,15 @@ def test_generating_qir_without_registers_raises():
 def test_custom_qir_intrinsics_generates_qir():
     expected_qir = read_resource_file("custom_intrinsics.ll")
 
-    def my_gate(self: QuantumCircuit, qubit: QubitSpecifier):
+    def my_gate(self, qubit):
```
swernli, line 150: Out of curiosity, why remove the types here?

source/pip/qsharp/interop/qiskit/backends/backend_base.py
```diff
@@ -120,6 +121,7 @@ def __init__(
             qiskit_pass_options (Dict): Options for the Qiskit passes.
             transpile_options (Dict): Options for the transpiler.
             qasm_export_options (Dict): Options for the QASM3 exporter.
+            num_qubits: optional number of qubits for the default target if not supplied
```
swernli: This parameter comment is hard to parse somehow. If I'm understanding the intent, maybe something like this would work?

```suggestion
            num_qubits: optional number of qubits to use when a target is not specified.
```

swernli: Perhaps worth combining this with a tweak to the `target` parameter comment to say that when a target is not specified a default one is constructed.

source/pip/qsharp/interop/qiskit/backends/qirtarget.py
```diff
@@ -52,90 +53,165 @@
...
+        Returns:
+            Target: A Qiskit Target object configured with quantum gates and operations
+                including:
+                - Basic single-qubit gates (X, Y, Z, H, S, T, SX, I)
+                - Rotation gates (RX, RY, RZ) with parameters
+                - Two-qubit gates (CX, CY, CZ, SWAP, controlled rotations)
+                - Three-qubit gates (CCX)
+                - Multi-qubit rotation gates (RXX, RYY, RZZ)
+                - Measurement and reset operations
+                - Control flow operations (if target_profile != Base)
```
swernli: nitpick: my brain parsed the `if` in the parenthetical as if it was supposed to be an example of a "Control flow operation." Maybe just a slight tweak to help readability:

```suggestion
                - Control flow operations (when target_profile != Base)
```

source/pip/tests-integration/interop_qiskit/test_qir.py
idavis: not really valuable for the test and QubitSpecifier isn't really an exported type that is usable. I can't remember if it even exists in 2.0.

source/pip/qsharp/interop/qiskit/backends/backend_base.py
idavis: Removed the param

idavis:

source/pip/tests-integration/interop_qiskit/test_re.py
```diff
@@ -106,23 +137,37 @@ def test_estimate_qiskit_rgqft_multiplier_in_threadpool() -> None:
...
+                "measurementCount": 0,
+            }
+        )
+    elif QISKIT_VERSION.startswith("2."):
+        assert res.logical_counts == LogicalCounts(
+            {
+                "numQubits": 16,
+                "tCount": 140,
+                "rotationCount": 532,
+                "rotationDepth": 369,
```
billti, line 163: Interesting that the RE counts have quite a difference between Qiskit 1 and Qiskit 2. Any idea why? (CC @msoeken )

msoeken: Without having looked at the concrete output circuits from both versions, I suspect that version 2 avoids using `U` gates in favor of more concrete variants such as T gates or R gates. U decomposes in the worst case into 3 rotations.

idavis: Transpiler has different implementations between the two as well. Qiskit 2.0 AFAICT pushes more for optimizing the gate transforms for their HW OOTB.

PR #2755, opened 2025-10-23
Disable circuit tracing in Python by default

minestarks:
The `dump_circuit` API dumps the "circuit" that was simulated in prior calls to `qsharp.eval` or `%%qsharp` notebook cells. It's essentially a trace of what happened so far.

The problem is, in order to make this work, we need to keep a full trace of all the gates that have been applied, in memory, just in case someone calls `dump_circuit`. This should be very rare. We shouldn't pay the perf cost every time we execute Q#.

Note : `dump_circuit` is distinct from `circuit`. The latter is untouched in this PR!

Related: #2099

### DIFF
--- a/source/compiler/qsc/src/interpret.rs
+++ b/source/compiler/qsc/src/interpret.rs
@@ -176,27 +176,13 @@ pub struct TaggedItem {
     pub namespace: Vec<Rc<str>>,
 }
 
-impl Interpreter {
-    pub fn new_with_circuit_trace(
-        sources: SourceMap,
-        package_type: PackageType,
-        capabilities: TargetCapabilityFlags,
-        language_features: LanguageFeatures,
-        store: PackageStore,
-        dependencies: &Dependencies,
-    ) -> std::result::Result<Self, Vec<Error>> {
-        Self::new_with_sources(
-            false,
-            sources,
-            package_type,
-            capabilities,
-            language_features,
-            store,
-            dependencies,
-            true,
-        )
-    }
+#[derive(PartialEq, Eq, Copy, Clone)]
+pub enum TraceCircuitOption {
+    Enabled,
+    Disabled,
+}
 
+impl Interpreter {
     /// Creates a new incremental compiler, compiling the passed in sources.
     /// # Errors
     /// If compiling the sources fails, compiler errors are returned.
@@ -208,7 +194,7 @@ impl Interpreter {
         store: PackageStore,
         dependencies: &Dependencies,
     ) -> std::result::Result<Self, Vec<Error>> {
-        Self::new_with_sources(
+        Self::with_sources(
             false,
             sources,
             package_type,
@@ -216,14 +202,11 @@ impl Interpreter {
             language_features,
             store,
             dependencies,
-            false,
+            TraceCircuitOption::Disabled,
         )
     }
 
-    /// Creates a new incremental compiler with debugging stmts enabled, compiling the passed in sources.
-    /// # Errors
-    /// If compiling the sources fails, compiler errors are returned.
-    pub fn new_with_debug(
+    pub fn with_circuit_trace(
         sources: SourceMap,
         package_type: PackageType,
         capabilities: TargetCapabilityFlags,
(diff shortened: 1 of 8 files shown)


source/compiler/qsc/src/interpret.rs
```diff
@@ -241,56 +264,10 @@ impl Interpreter {
         )
         .map_err(into_errors)?;
 
-        let mut fir_store = fir::PackageStore::new();
```
minestarks, line 244: This code was just duplicated below.

source/compiler/qsc_eval/src/backend.rs
```diff
@@ -665,190 +665,237 @@ fn unwrap_matrix_as_array2(matrix: Value, qubits: &[usize]) -> Array2<Complex<f6
...
+/// so that each operation is executed on both backends.
 /// For any intrinsics that return a value,
-/// the value returned by the chained backend is ignored.
-/// The value returned by the main backend is returned.
-pub struct Chain<T1, T2> {
-    pub main: T1,
-    pub chained: T2,
+/// the value returned by the tracing backend is ignored.
+/// The value returned by the simulator backend is returned.
+pub struct TracingBackend<T1: Backend, T2: Backend> {
```
minestarks, line 673: The renaming of this struct is a teaser for my upcoming PR, where the "tracer" trait will evolve away from `Backend` . I thought I'd make the roles more explicit here, especially since I made the second backend optional, which kind of messes up the "chain" idea.

I don't think my generic "backend chain" idea really panned out. We only use this struct for tracing a circuit while simulating, so let's call it what it is.

source/compiler/qsc/src/interpret.rs
```diff
@@ -1224,12 +1220,16 @@ impl Interpreter {
     }
 }
 
-fn sim_circuit_backend() -> BackendChain<SparseSim, CircuitBuilder> {
-    BackendChain::new(
+fn sim_circuit_backend(trace_circuit: bool) -> TracingBackend<SparseSim, CircuitBuilder> {
```
idavis: Should we consider an enum instead of a bool? Usage of the function in unclear passing boolean flags.

source/compiler/qsc/src/interpret.rs
```diff
@@ -211,25 +231,28 @@ impl Interpreter {
...
             language_features,
             store,
             dependencies,
+            true, // enable circuit tracing in debug mode
         )
     }
 
-    fn new_internal(
+    #[allow(clippy::too_many_arguments)]
+    fn new_with_sources(
```
idavis: `Interpreter::from` and `Compiler::from` were paired in names to indicate their usage together. If renaming `Interpreter::from` to `Interpreter::new_with_package_store` we should rename `Compiler::from` to `Compiler::new_with_package_store`

minestarks: I'm down with renaming `Compiler` as well. I thought `from` was getting a little unclear now that we have so many different options to initialize the interpreter. OK with you?

source/compiler/qsc/src/interpret.rs
minestarks: Good idea, thanks.

source/compiler/qsc/src/interpret.rs
minestarks: Changed.

source/compiler/qsc/src/interpret.rs
minestarks: Changed.

source/compiler/qsc_eval/src/backend.rs
```diff
@@ -665,190 +665,237 @@ fn unwrap_matrix_as_array2(matrix: Value, qubits: &[usize]) -> Array2<Complex<f6
...
+            tracer.custom_intrinsic(name, arg);
+        }
+        res
     }
 
     fn set_seed(&mut self, seed: Option<u64>) {
-        self.chained.set_seed(seed);
-        self.main.set_seed(seed);
+        self.backend.set_seed(seed);
     }
```
idavis, line 900: Are we omitting seed since the tracer won't use it?

minestarks: Yes, that was my intention. Much like we don't use the return value from the tracer in some of the methods (like `m`), this is one where the behavior of the two "backends" aren't exactly parallel.

This method will be dropped altogether in my next PR when `Tracer` evolves away from `Backend`.

idavis:

PR #2756, opened 2025-10-27
Removed References to the QDK Package being "preview"

ScottCarda-MS:


source/qdk_package/README.md
```diff
@@ -1,6 +1,6 @@
 # qdk
 
-Preview package for the Quantum Development Kit (QDK) that bundles the `qsharp` Python package together with optional extras under a single top level package.
+Package for the Quantum Development Kit (QDK) that bundles the `qsharp` Python package together with optional extras under a single top level package.
```
swernli: "Package" alone here reads a little odd. I wonder if rewording it a bit more is better?
```suggestion
The Quantum Development Kit (QDK) bundles the `qsharp` Python package together with optional extras under a single, top-level meta-package.
```

swernli (approved):
One suggestion about phrasing, otherwise looks good.

source/qdk_package/pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "qdk"
 version = "0.0.0"
-description = "Unified Quantum Development Kit Python meta-package (wraps qsharp and optional extras)"
```
billti, line 8: Why call it a `meta-package`? What does that mean to users? Just call it the Quantum Development Kit Python Package. Ideally `qsharp` just becomes thought of as a submodule, just like `openqasm`. They don't need to know (or care) that it is bundling and exposes a separate `qsharp` package under the hood.

source/qdk_package/README.md
ScottCarda-MS: I reworded this to avoid calling it a meta-package and to lean away from thinking of this as a wrapper for the qsharp package, as-per @billti's comment below.

source/qdk_package/pyproject.toml
ScottCarda-MS: Makes sense. I've reworded this to avoid calling it a meta-package.

PR #2759, opened 2025-10-28
Add lint warning for ambiguous if-statement followed by unary operator

swernli:
Because we support unary `+` and `-` a user can mistakenly write an if-statement followed by something like ` - x` and expect that it subtracts that from the result of the if-statement, when in fact parser treats this as two separate statements. This link helps call attention to the ambiguity and encourage the user to use clearer syntax to disambiguate.

### DIFF
--- a/source/vscode/qsharp.schema.json
+++ b/source/vscode/qsharp.schema.json
@@ -65,7 +65,8 @@
                   "needlessOperation",
                   "deprecatedFunctionConstructor",
                   "deprecatedWithOperator",
-                  "deprecatedDoubleColonOperator"
+                  "deprecatedDoubleColonOperator",
+                  "ambiguousUnaryOperatorAfterIf"
                 ]
               },
               "level": {


source/compiler/qsc_linter/src/lints/hir.rs
```diff
@@ -342,3 +343,38 @@ impl HirLintPass for DeprecatedDoubleColonOperator {
...
+            // Not enough statements to have an if-expression followed by a unary operator.
+            return;
+        }
+        for i in 0..(block.stmts.len() - 1) {
+            if let StmtKind::Expr(expr) = &block.stmts[i].kind
+                && matches!(&expr.kind, ExprKind::If(..))
+                && let StmtKind::Expr(next_expr) = &block.stmts[i + 1].kind
+                && matches!(next_expr.kind, ExprKind::UnOp(UnOp::Pos | UnOp::Neg, _))
+                && expr.ty != Ty::UNIT
+                && next_expr.ty != Ty::UNIT
```
ScottCarda-MS: Is that possible? Users can't really have a `-()`.

source/compiler/qsc_linter/src/tests.rs
```diff
@@ -809,6 +809,82 @@ fn deprecated_assign_update_expr_code_action() {
...
+
+#[test]
+fn ambiguous_unary_operator_after_if_fixed_by_parens() {
+    check(
+        &wrap_in_callable("(if true { 42 } else { 0 }) - 1", CallableKind::Function),
+        &expect![[r#"
+            []
+        "#]],
+    );
+}
```
orpuente-MS, line 838: Can we add a test similar to this one checking that the lint doesn't trigger for `(if true { 42 } else { 0 } - 1)`

source/compiler/qsc_linter/src/lints/hir.rs
swernli: That's a good point, this can probably be removed as an extraneous check.

source/compiler/qsc_linter/src/tests.rs
swernli: Good call, added.

PR #2760, opened 2025-10-29
Automatic estimation of overhead in memory/compute architecture

msoeken:
This adds a RE intrinsic called `EnableMemoryComputeArchitecture` that emulates a compute/memory architecture with _compute qubits_ that can be operated on with logical operations and _memory qubits_ that can be turned into compute qubits in order to compute on them. The intrinsic takes as parameter an upper bound `k` on the number of compute qubits. The first `k` qubits are assumed to be compute qubits from the start. However, when an operation is applied to a qubit that is not a compute qubit and there is no more space for compute qubits, one needs to evict a compute qubit and move it into memory. Similarly, if the qubit to be acted on is in memory, it has to be moved into the compute qubit space. The logical resource counter counts the number of times a qubit is read from memory and written to memory. Further, one can specify an eviction strategy (either `LeastRecentlyUsed` that evicts a compute qubit that has been used least recently or `LeastFrequentlyUsed` that evicts a compute qubit that has been used least frequently).

### DIFF
--- a/source/resource_estimator/src/counts/memory_compute.rs
+++ b/source/resource_estimator/src/counts/memory_compute.rs
@@ -6,17 +6,17 @@ use std::hash::Hash;
 mod tests;
 
 pub enum CachingStrategy {
-    LeastRecentlyUsed(LruPQ<usize>),
-    LeastFrequentlyUsed(LfuPQ<usize>),
+    LeastRecentlyUsed(LeastRecentlyUsedPriorityQueue<usize>),
+    LeastFrequentlyUsed(LeastFrequentlyUsedPriorityQueue<usize>),
 }
 
 impl CachingStrategy {
     pub fn least_recently_used(capacity: usize) -> Self {
-        CachingStrategy::LeastRecentlyUsed(LruPQ::new(capacity))
+        CachingStrategy::LeastRecentlyUsed(LeastRecentlyUsedPriorityQueue::new(capacity))
     }
 
     pub fn least_frequently_used(capacity: usize) -> Self {
-        CachingStrategy::LeastFrequentlyUsed(LfuPQ::new(capacity))
+        CachingStrategy::LeastFrequentlyUsed(LeastFrequentlyUsedPriorityQueue::new(capacity))
     }
 }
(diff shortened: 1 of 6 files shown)


source/resource_estimator/src/counts/memory_compute.rs
```diff
@@ -0,0 +1,339 @@
...
+        if self.map.len() > self.max_size {
+            self.max_size = self.map.len();
+        }
+    }
+}
+
+/// LFU priority queue / set. Maintains up to `capacity` distinct keys; eviction
+/// removes the key with lowest frequency (ties broken by oldest insertion among
+/// that frequency bucket).
+pub struct LfuPQ<K> {
```
ivanbasov: nit: maybe the full name would be better. 
But it is a personal preference not a strong suggestion.

ivanbasov (approved):

source/resource_estimator/src/counts.rs
```diff
@@ -567,6 +668,14 @@ impl Backend for LogicalCounter {
...
                 )
             }
+            "EnableMemoryComputeArchitecture" => {
+                let values = arg.unwrap_tuple();
+                let [compute_capacity, strategy] = array::from_fn(|i| values[i].clone());
+                let compute_capacity = compute_capacity.unwrap_int();
+                let strategy = strategy.unwrap_int();
+                self.enable_memory_compute(compute_capacity, strategy);
+                Some(Ok(Value::unit()))
+            }
```
swernli, line 678: I think you need to add a no-op version of this to the implementation of the `Backend` trait in source/compiler/qsc_eval/src/backend.rs. Otherwise a program that uses this call won't be able to run through the simulator.

source/resource_estimator/src/counts/memory_compute.rs
msoeken: Changed to long names.

source/resource_estimator/src/counts.rs
msoeken: Added to other backends.

PR #2761, opened 2025-10-29
Show source code locations on circuit diagrams

minestarks:
This PR adds a feature to circuit diagrams to display Q#/QASM source code locations for each qubit and gate as clickable links in VS Code. They can also be displayed as hover text in the Python circuit widget.

https://github.com/user-attachments/assets/fd8203a6-e488-439a-ab65-7e5e18c0928d

### Operation call sites

Operation boxes are clickable and show hover text with the location of the call site.

image

We always show the nearest user code location to the actual call site.

For example, if the user code shows `ForEach(MResetZ, qs)`, even though the actual `MResetZ` gate is applied in the Std library, the location in the circuit diagram will point to the `ForEach` expression.

In Python, "user code" includes both cell contents and any real files pulled via `qsharp.init(project_root=...)`

### Qubit declaration sites

Qubit labels are clickable and show hover text with a location.

image

This location corresponds to the `use` statement that allocated the qubit. Example:

    use q = Qubit();

There is an exception to this when the circuit diagram is generated for an operation which takes qubit input arguments.
In this case, the declaration locations are the input arguments to the operation. Example:

    operation Test(q1: Qubit, q2: Qubit) : Result[] {}

Here, `q0` is declared at 1:16 and `q1` is declared at 1:27.

Qubits can have multiple declaration sites. Example:

    operation Main() : Unit {
        {
            use q1 = Qubit();
            X(q1);
            Reset(q1);
        }
        {
            use q2 = Qubit();
            Y(q2);
            Reset(q2);
        }
    }

The circuit diagram for this code only shows one qubit wire, but it maps to both `q1` and `q2`. So clicking on that qubit label will show both locations:

image

### Circuit JSON schema

Components now have a `source` field.

    {
        "kind": "unitary",
        "gate": "H",
        "targets": [{ "qubit": 0 }],
        "source": { "file": "line_0", "line": 0, "column": 43 }
    }

Qubits have a `declarations` field.

    {
        "id": 0,
        "numResults": 0,
        "declarations": [{ "file": "foo.ts", "line": 0, "column": 26 }]
    }

These fields don't exist in the .qsc file schema. Even though we don't have a user scenario that writes circuits out to editable .qsc files yet, if we ever do, these fields should be excluded as they lose their meaning when the generated circuit is divorced from its originating source code.

### VS Code

In VS code, clicking on an element with a source code location will navigate to the source code.

image

### Python

Source code locations can be enabled in Python by passing in `source_locations=True` into `qsharp.circuit()`.

image

While the hover text shows the location, clicking on the gates does nothing.

Source code locations are not enabled by default in Python. This is because today, the locations for Q#/QASM code evaluated via `qsharp.eval` don't look very meaningful (`line_1:4:2` etc). If a `qsharp.json` is used, then the files will have proper disk locations, but this is a relatively uncommon scenario.

### ASCII art update to show source code location labels, e.g.

ASCII art update to show source code location labels, e.g.

    q_0@test.qs:4:20 ──────── ● ────────
    q_1@test.qs:4:20 ──────── ● ────────
    q_2@test.qs:4:20 ─ X@test.qs:5:20 ──

These labels are disabled in the Python string rendering (the only place this ASCII art would be shown to the user) . It's useful in test snapshots (see `circuit_tests.rs` ), but at the moment they look too ugly for public consumption.

### Configuration and enablement

Since there is a perf cost to capturing call stacks during circuit tracing, we have to be conservative about when to enable circuit tracing. 

At this point circuit tracing is only enabled when:

 - Explicitly generating a circuit, via the "show circuit" command in VS code, or the `circuit()` API in Python
 - Debugging in VS Code  PR forthcoming to decouple the "Run" command from the debugger to avoid the perf hit)
 - When the Python library is initialized with `trace_circuit=True`

Additionally source location capturing is configurable via a user setting. When it's disabled, the Q#/QASM evaluator skips call stack captures.

In VS Code, new `settings.json` configurations have been added to tweak circuit generation options.

    "Q#.circuits.config": {
        "maxOperations": 10001,
        "generationMethod": "classicalEval",
        "sourceLocations": true
    }  

While `maxOperations` and `generationMethod` are not new features, they weren't configurable by the user. 

In Python, these configuration options are exposed in the `circuit()` function:

    generation_method: Optional[CircuitGenerationMethod] = None,
    max_operations: Optional[int] = None,
    source_locations: bool = False,

#### Evaluator and circuit tracer changes

Introduced a `Tracer` trait, now implemented by the circuit tracer (a.k.a. the circuit builder). This trait diverges from `Backend` mainly in that it takes a call stack parameter (`stack: &[Frame]`), which allows the circuit tracer implementation to capture source code metadata in the circuit. The `TracingBackend` composes a regular `Backend`, a `Tracer`, or both, so backends that do not care about tracing remain unchanged.

Every time the evaluator makes an intrinsic call, it captures the call stack and passes it to tracer (unless source location capturing is disabled).

Also removed `Result` associated type from `Backend` as it made passing this type around needlessly complicated. Implementors can simply return `val::Result`, as the `impl Into<Result>` was enforced everywhere anyway.

The mechanics of the circuit tracer is largely unchanged, but it is refactored to make way for future changes. `WireMapBuilder` and `OperationBuilder` are separated out for manageability; these will continue evolving as we implement further circuit diagram features.

#### Entry expressions for operation circuits    

Note that when we generate an circuit for an operation that takes qubits (as opposed to an Entrypoint, which has no input parameters), we have to do some additional processing to figure out qubit declaration locations. 

Since in this scenario we use a generated entry expression, the real qubit declarations would live there. Example:

    {
        use qs = Qubit[2];
        (Test.Test)(qs[0], qs[1]);
        let r: Result[] = [];
        r
    }

We can't use these as the qubit declaration locations since they don't map to user code, and don't provide any meaningful insight to the user.

#### VS Code command

A new generic `qsharp-vscode.gotoLocations` command was added to jump to a specific location in source. The built-in VS Code commands didn't play nicely with our UX.

#### Circuit rendering

If operations or gate labels contain source code locations, they are rendered as clickable SVG elements. The hover text should show the source code location as well.

Minor fixes in rendering and CSS (dashed lines were invisible, SWAP gates had an error in the bounding box calculation.)

Changed the signature of `draw` to better organize the increasing number of options.

### DIFF
--- a/source/vscode/package.json
+++ b/source/vscode/package.json
@@ -383,8 +383,10 @@
       },
       {
         "command": "qsharp-vscode.runProgram",
-        "title": "Run program",
-        "category": "QDK"
+        "title": "Run",
+        "category": "Debug",
+        "enablement": "!inDebugMode",
+        "icon": "$(play)"
       },
       {
         "command": "qsharp-vscode.runEditorContentsWithCircuit",


source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -210,22 +226,22 @@ fn m_base_profile() {
...
+        CircuitGenerationMethod::ClassicalEval,
+        Default::default(),
+    )
+    .expect("circuit generation should succeed");
 
     expect![[r#"
-        q_0    ── H ──── M ──
-                         ╘═══
+        q_0@test.qs:5:20 ─ H@test.qs:6:20 ─── M@test.qs:7:21 ──
+                                           ╘═════════
```
ScottCarda-MS: It looks like the measurement line is unaligned here.

source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -294,17 +302,105 @@ fn mresetz_base_profile() {
...
+    ",
+        CircuitEntryPoint::EntryPoint,
     );
 
-    let circ = interpreter
-        .circuit(CircuitEntryPoint::EntryPoint, false)
-        .expect("circuit generation should succeed");
+    expect![[r#"
+        q_0@test.qs:3:32 ─ H@test.qs:4:16 ────────── ● ──────────────────────────── X@test.qs:8:16 ─── M@test.qs:10:16 ─── |0〉@test.qs:10:16 ──
+                                           │                                     │                  ╘════════════════════════════════
```
ScottCarda-MS: These control connection lines also seem to be misaligned.

source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -294,17 +302,105 @@ fn mresetz_base_profile() {
...
+        q_0@test.qs:4:20, test.qs:9:20 ─ X@test.qs:5:20 ─── M@test.qs:6:20 ──── |0〉@test.qs:6:20 ──── Y@test.qs:10:20 ── M@test.qs:11:20 ─── |0〉@test.qs:11:20 ──
+                                           ╘════════════════════════════════════════════════════════════╪════════════════════════════════
+                                                                                                        ╘════════════════════════════════
     "#]]
-    .assert_eq(&circ.to_string());
+    .assert_eq(&circ);
+}
+
+#[test]
+fn qubit_reuse_no_measurements() {
```
ScottCarda-MS, line 379: I did not know that circuit generation would reuse qubits like that.

minestarks: Sure does! (Just checked to confirm this is not new behavior). This is why qubit labels can map to multiple source locations.

We _could_ change this behavior, but as soon as we start generating circuits from RIR/QIR, we'll be back to the way it is today. This is because qubit reuse is also baked into RIR/QIR generation. Qubits are released at the end of each block, and we reuse hardware qubits whenever we can.

source/compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Looks like I messed up the column width calculation with the new labels. Looking!

source/compiler/qsc_circuit/src/json_to_circuit.rs
```diff
@@ -101,6 +101,7 @@ fn to_circuit_group(mut json: Map<String, Value>) -> Result<CircuitGroup, Error>
                                 .unwrap_or(0),
                         )
                         .map_err(|_| Error::custom("Value of 'numChildren' is out of range"))?,
+                        declarations: vec![],
```
ScottCarda-MS, line 104: Does this mean that if the json persisted declaration information, those declarations will be ignored upon reading this in as a circuit?

source/compiler/qsc_circuit/src/operations.rs
```diff
@@ -27,35 +27,47 @@ pub enum Error {
     ControlledUnsupported,
 }
 
+pub struct QubitParamInfo {
+    /// The qubit parameters this operation takes.
+    pub(crate) qubit_params: Vec<QubitParam>,
+    /// The total number of qubits that would
+    /// need be allocated to run this operation for the purposes of circuit generation.
+    pub total_num_qubits: u32,
```
ScottCarda-MS: Is this specifically the qubits that are allocated _internally_ to the operation, or does it also include the qubits that are passed in from the calling context?

minestarks: Just passed in from the caller, so maybe `total_input_qubits` would be more accurate.

source/compiler/qsc_circuit/src/json_to_circuit.rs
minestarks: Yes. I *think* this is the right thing to do in theory. But it's also kind of hard to reason about the real-life implications of doing this, since it's not an actual a real user scenario. You'd have to manually create a JSON file with a `declarations` field in it to be able to hit this code path.

source/compiler/qsc_circuit/src/operations.rs
```diff
@@ -27,35 +27,47 @@ pub enum Error {
     ControlledUnsupported,
 }
 
+pub struct QubitParamInfo {
```
ScottCarda-MS: I'm not sure I agree with having this struct. The name is a little confusing because it sounds like it is info about a qubit param, rather than info about an operation's full list of qubit parameters.
It is also a wrapper around a vec, with an added shortcut field that sums up over an aspect of the elements in that vec. I don't think the `total_num_qubits` field adds anything that couldn't be found by looping over the vec, and it is awkward to maintain it manually. A param could theoretically be added to the `qubit_params` without updating the `total_num_qubits`, invalidating the information in the struct.
I guess I would just ask, is it really worth the risky maintenance cost and the difficult-to-name name of the struct to avoid looping over a vec when you need to know the total number of qubits?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -184,28 +204,28 @@ const _gatePosition = (
...
   switch (type) {
     case GateType.Group: {
       const padding = groupBoxPadding - nestedDepth * nestedGroupPadding;
 
       x1 = x - 2 * padding;
       y1 = minY - gateHeight / 2 - padding;
-      x2 = width + 2 * padding;
-      y2 = maxY + +gateHeight / 2 + padding - (minY - gateHeight / 2 - padding);
+      w = width + 2 * padding;
+      h = maxY - minY + gateHeight + 2 * padding;
```
ScottCarda-MS: I can see the equation from before was extra confusing, I'm still confused why `maxY - minY` is needed. Shouldn't height just be `gateHeight + 2 * padding`?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -184,28 +204,28 @@ const _gatePosition = (
   const maxY = Math.max(...ys);
   const minY = Math.min(...ys);
 
-  let x1: number, y1: number, x2: number, y2: number;
+  let x1: number, y1: number, w: number, h: number;
```
ScottCarda-MS: nit: now that there's not `x2, y2`, `x1, y1` seems a bit strange compared to just `x, y`

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -391,10 +413,9 @@ const _swap = (renderData: GateRenderData, nestedDepth: number): SVGElement => {
   const { x, targetsY } = renderData;
 
   // Get SVGs of crosses
-  const [x1, y1, x2, y2] = _gatePosition(renderData, nestedDepth);
+  const [x1, y1, w, h] = _gatePosition(renderData, nestedDepth);
```
ScottCarda-MS: nit: same here about `x1, y1`

source/npm/qsharp/ux/circuit-vis/index.ts
```diff
@@ -17,13 +17,10 @@ import { CircuitGroup } from "./circuit.js";
 export const draw = (
   circuitGroup: CircuitGroup,
   container: HTMLElement,
-  renderDepth = 0,
-  isEditable = false,
-  editCallback?: (circuitGroup: CircuitGroup) => void,
-  runCallback?: () => void,
+  options: DrawOptions = {},
```
ScottCarda-MS, line 20: nice ❤️

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -1250,19 +1273,21 @@ impl State {
         callee_id: StoreItemId,
         functor: FunctorApp,
         callee: &fir::CallableDecl,
-        sim: &mut impl Backend<ResultType = impl Into<val::Result>>,
+        sim: &mut TracingBackend,
```
swernli: nitpick: In some places you renamed this parameter from `sim` to `tracing_backend` and in others like this one you left it as `sim`. I'd personally prefer leaving it as `sim` since the type (which shows nicely with the editor extension) describes the tracing part well and then thew variable conveys what it is used for. Not critical though.

source/npm/qsharp/ux/circuit-vis/sqore.ts
```diff
@@ -203,10 +206,14 @@ export class Sqore {
     };
 
     const { qubits, componentGrid } = circuit;
-    const { qubitWires, registers, svgHeight } = formatInputs(qubits);
+    const { qubitWires, registers, svgHeight } = formatInputs(
+      qubits,
+      !this.options.isEditable ? this.options.renderLocations : undefined,
```
ScottCarda-MS: nit: Can this condition be flipped to avoid the negation?
`this.options.isEditable ? undefined : this.options.renderLocations`

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: Haha, some of this stuff really made me question my sanity (and my 5th grade math skills). I think it's because `gateHeight` only represents the height of a single box (it's hardcoded to 40) but this function is supposed to compute the _full_ height of a gate, even when it spans multiple wires.

Like if you have a CNOT gate like this:

```
--*---
  |
--X---
```

`gateHeight` would be 1 unit, but  the `height` returned from this function should be 3 units (plus padding).

source/npm/qsharp/ux/circuit.tsx
```diff
@@ -329,6 +331,37 @@ export function CircuitPanel(props: CircuitProps) {
           isEditable={props.isEditable}
           editCallback={props.editCallback}
           runCallback={props.runCallback}
+          renderLocations={(locations) => {
```
ScottCarda-MS: This lambda is non-trivial; can it be factored out into a named function?

source/npm/qsharp/ux/qsharp-circuit.css
```diff
@@ -72,6 +72,34 @@
     fill: var(--vscode-button-secondaryForeground, #ffffff);
   }
 
+  a.qs-circuit-source-link .qs-qubit-label {
+    pointer-events: auto;
+  }
+
+  /* Hover effect for clickable gates */
+  a.qs-circuit-source-link:hover .gate-unitary {
+    fill: var(--vscode-button-secondaryHoverBackground, #4d4d4d);
```
ScottCarda-MS, line 81: When picking default values for these, the thing to keep in mind is that (I believe) all the out-of-the-box VS Code themes defined all these variables _except_ the two high contrast themes (which happen to be the ones I use 😊). We don't really need to go out of our way to make sure the defaults look good with custom themes, because people do wacky things, but we should make sure the defaults work well with the out-of-the-box high contrast themes.
Checking those, under the light high contrast theme, the unitary gates are not readable with these colors, and the rest of the elements don't change. Under the dark high contrast theme, most things are fine, but measurements and kets don't change any color.

minestarks: That's an excellent point, thank you. I'll test with those themes

source/compiler/qsc_circuit/src/operations.rs
minestarks: I think you hit the nail on the head. `total_num_qubits` did feel off somehow, but I couldn't figure out how to improve it. It could be a method/function rather than a field.

source/npm/qsharp/ux/qsharp-circuit.css
```diff
@@ -72,6 +72,34 @@
     fill: var(--vscode-button-secondaryForeground, #ffffff);
   }
 
+  a.qs-circuit-source-link .qs-qubit-label {
+    pointer-events: auto;
+  }
+
+  /* Hover effect for clickable gates */
```
ScottCarda-MS, line 79: I'm not entirely sure what code change is causing this, but this PR is breaking circuit visualization for dark theme high-contrast: the measurements and kets (as well as the Run button in the editor) are supposed to be blue, but they now appear black.

ScottCarda-MS: But only in the browser, for some reason!?

source/compiler/qsc_eval/src/lib.rs
minestarks: No, this is exactly the level of pedantry I was hoping for ;) Changed to `sim`

source/npm/qsharp/ux/qsharp-circuit.css
minestarks: I can repro this on `main` when I use the `vscode start` command to start the local vs code server. However it doesn't repro on actual `vscode.dev` (https://insiders.vscode.dev/+quantum.qsharp-lang-vscode-dev/playground/)  . So I don't think it's because of anything in my PR.

source/compiler/qsc/src/interpret/circuit_tests.rs
minestarks: Fixed. I wasn't taking qubit label length into account.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: Went ahead and renamed a whole bunch of variables around here to eliminate confusion.

source/npm/qsharp/ux/circuit.tsx
minestarks: Refactored.

source/npm/qsharp/ux/circuit-vis/sqore.ts
minestarks: Fixed

source/npm/qsharp/ux/qsharp-circuit.css
minestarks: In an amazing turn of events, VS Code pushed an update that changed these variables for the High Contrast themes (https://github.com/microsoft/vscode/pull/270768).  So this repros differently now. 

These definitions make it very challenging to work with high contrast themes. There's no distinct hover color for `button.background` , which is intentional for high contrast themes I think.  

`button.secondaryBackground` and `button.secondaryHoverBackground` are mostly null... except when they randomly aren't. 

We are falling back to our default colors when these values aren't defined, but in some cases like HC light, `secondaryBackground` IS defined but its hover color isn't. So we only fall back to our default on hover, which looks incongruent with the rest of the HC theme. We can't reasonably detect and work around these cases.

image

source/npm/qsharp/ux/qsharp-circuit.css
minestarks: Tracked this down to a theme update in VS Code itself (mentioned in my other comment)

source/npm/qsharp/ux/qsharp-circuit.css
minestarks: Opened https://github.com/microsoft/qdk/issues/2787

ScottCarda-MS: Dang, HC themes can be tough to work with. Maybe the extra requirements placed on them for accessibility means there are more undefined fields? Not sure, but in any case we can fiddle with the CSS later if we decide things look bad enough to change it. Thanks for looking into this!

ScottCarda-MS (approved):

PR #2762, opened 2025-10-30
Adjust over-large address register in Unlookup

msoeken:
We check in PhaseLookup whether the number of data bit-strings matches the address length. However, the address register may be larger than necessary (which is not a problem), in which case we need to shrink it to the required size. The computation of the required size and its validation is already present in the `Unlookup` operation.

### DIFF
--- a/library/std/src/Std/TableLookup.qs
+++ b/library/std/src/Std/TableLookup.qs
@@ -251,7 +251,7 @@ operation Unlookup(
         let phaseData = Padded(-2^addressBitsNeeded, false, phaseData);
 
         // Apply phase lookup to correct phases in the address register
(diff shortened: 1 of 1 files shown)


library/std/src/Std/TableLookup.qs
```diff
@@ -251,7 +251,7 @@ operation Unlookup(
         let phaseData = Padded(-2^addressBitsNeeded, false, phaseData);
 
         // Apply phase lookup to correct phases in the address register
-        PhaseLookup(address, phaseData);
+        PhaseLookup(address[...addressBitsNeeded - 1], phaseData);
```
DmitryVasilevsky, line 254: We probably need a test case for this.

PR #2770, opened 2025-10-31
Update Generated Sample Notebook to use `qdk` Instead of Separate Packages

ScottCarda-MS:


source/vscode/src/notebookTemplate.ts
```diff
@@ -13,23 +13,23 @@ export const notebookTemplate: vscode.NotebookData = {
...
 To render charts as shown in this notebook, Matplotlib with notebook widgets support should be installed via "ipympl".
 
-To submit to Azure Quantum as shown in this notebook, the "azure-quantum" package needs to be installed.
+To submit to Azure Quantum as shown in this notebook, the "azure" extra of the "qdk" package needs to be installed.
 
 You can install all the above packages in your Python environment by running the below in your terminal:
 
 \`\`\`bash
-pip install jupyterlab qsharp ipympl azure-quantum
+pip install qdk[jupyter,azure] ipympl
```
billti: You need double quotes around the package when including extras

ScottCarda-MS: done

billti (approved):

Issue #2771, opened 2025-10-31
Creating an explicit namespace with the same name as a callable breaks Python interop


swernli:
From the Q# side, defining an explicit namespace in a cell with the same name as an earlier callable works because the two different entities are differentiable by type. But in Python interop, they try to occupy the same entity under `qsharp.code` and end up breaking the interop mechanisms:

<img width="885" height="1132" alt="Image" src="https://github.com/user-attachments/assets/e2a9d8d6-9305-4e0b-831b-2acc5f355d08" />

We should decide what the Q# behavior and the Python interop expectation is in this situation. Does the later defined instance shadow the first? Should Q# disallow this?

PR #2772, opened 2025-10-31
Avoid name collision in sample

swernli:
This avoids a name collision in the sample that can cause confusion (see #2771)

### DIFF
--- a/samples/notebooks/sample.ipynb
+++ b/samples/notebooks/sample.ipynb
@@ -480,8 +480,8 @@
     "%%qsharp\n",
     "\n",
     "import Std.Diagnostics.DumpMachine;\n",
-    "namespace Bar {\n",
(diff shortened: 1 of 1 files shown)


samples/notebooks/sample.ipynb
```diff
@@ -480,8 +480,8 @@
     "%%qsharp\n",
     "\n",
     "import Std.Diagnostics.DumpMachine;\n",
-    "namespace Foo {\n",
-    "    operation Bar() : Unit {\n",
+    "namespace Bar {\n",
```
billti: Bar is already an operation earlier in this notebook.

swernli: fixed again, but better this time...

billti (approved):

PR #2774, opened 2025-11-04
Update rust version to 1.90.0

orpuente-MS:
This PR updates the rust version to 1.90.0 and fixes new clippy lints that are part of the new version.

### DIFF
--- a/source/pip/src/interpreter.rs
+++ b/source/pip/src/interpreter.rs
@@ -1,6 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(
+    clippy::doc_markdown,
+    reason = "docstrings in this module conform to the python docstring format."
+)]
+
 pub(crate) mod data_interop;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_qasm/src/lib.rs
```diff
@@ -192,11 +192,11 @@ impl OperationSignature {
...
         if let Some(ns) = &self.ns {
-            write!(expr, "{ns}.").unwrap();
+            write!(expr, "{ns}.").expect("write should succeed");
         }
-        write!(expr, "{}(", self.name).unwrap();
-        write!(expr, "{}", params.as_ref()).unwrap();
-        write!(expr, ")").unwrap();
+        write!(expr, "{}(", self.name).expect("write should succeed");
+        write!(expr, "{}", params.as_ref()).expect("write should succeed");
+        write!(expr, ")").expect("write should succeed");
```
minestarks, line 199: Curious if you know why these weren't flagged before (since clippy has always yelled at me when I tried to use `unwrap()`)

source/pip/src/interpreter.rs
```diff
@@ -455,11 +455,11 @@ impl Interpreter {
     /// Interprets Q# source code.
     ///
     /// :param input: The Q# source code to interpret.
-    /// :param output_fn: A callback function that will be called with each output.
+    /// :param `output_fn`: A callback function that will be called with each output.
```
minestarks: I think these are supposed to conform to a Python doc comment format, so the lack of backticks is intentional. (Same comment for other instances in this crate)

minestarks (approved):

source/compiler/qsc_qasm/src/lib.rs
orpuente-MS: I was asking myself the same question. But haven't been able to find why

source/pip/src/interpreter.rs
orpuente-MS: Nice call.

PR #2786, opened 2025-11-17
Migrate "Cancel Job" Command and Add "Delete Job" Command for testing

ScottCarda-MS:
This PR moves the "Cancel job" command off of the deprecated DELETE action and onto the new POST action supported by the service. It also adds the "Delete job" command that will make use of the DELETE action once the service supports it. The "Delete job" is gated behind an experimental setting until that action is fully supported by the service. The setting is not shown in the settings UI, so it will need to be added manually to the settings JSON to use the feature like so:
```JSON
{
    "Q#.azure.experimental.enableJobDelete": true
}
```

The Delete Job command can be used via a trash icon button on the tree view for jobs:
image

Or it can be used via the command palette when a job is selected in the tree view:
image

And it always checks with a dialog before committing to the deletion:
image

### DIFF
--- a/source/vscode/src/azure/commands.ts
+++ b/source/vscode/src/azure/commands.ts
@@ -138,8 +138,6 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
         `${qsharpExtensionId}.treeItemIsCancelable`,
         isCancelable,
       );
-      // Also (re)apply the delete jobs feature flag context so palette menu conditions re-evaluate.
-      await setDeleteJobsFeatureFlagContext();
     }),
   );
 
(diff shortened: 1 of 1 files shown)


source/vscode/src/azure/commands.ts
```diff
@@ -125,14 +128,43 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
...
+        `${qsharpExtensionId}.treeItemIsDeletable`,
+        isDeletable,
+      );
       await vscode.commands.executeCommand(
         "setContext",
         `${qsharpExtensionId}.treeItemIsCancelable`,
         isCancelable,
       );
+      // Also (re)apply the delete jobs feature flag context so palette menu conditions re-evaluate.
+      await setDeleteJobsFeatureFlagContext();
```
billti: If this is going to run on initialization (line 147) and then any time the configuration changes (line 150), then why does it also need to run every time the treeview selection is changed?

ScottCarda-MS: You are right, that is overkill. It doesn't need to be updated every time the treeview selection changes.

PR #2788, opened 2025-11-17
Run program in terminal instead of the debugger

### LINKED ISSUE #149: Improve WASM layer error handling

minestarks:
Before this change, the "run" command (accessible via the "Run" code lens, the play button in the corner of the editor or the command palette) was simply invoking the debugger without any breakpoints.

The problem with this approach is that the debugger is slow relative to just running a program using the `run()` method in the WASM module.  With the recent circuit tracing changes, the debugger is getting more heavyweight.

This PR decouples the "Run" command from the debugger. Instead, we create a VS Code terminal and direct the program output to that terminal.

As part of this change, I had to make some changes to error reporting, which happens to fix #149 .

The changes in error reporting and console output handling means that there are subtle updates to how Test Explorer and Copilot reports failed programs as well. 

See below for a full UX comparison.

| | Before | After |
|---|---|---|
| Run | image | image |
| Debug | image | image |
| Run (compile error) | image | image |
| Debug (compile error) | image | image |
| Run (runtime error) | image | image |
| Debug (runtime error) | image | image |
| Run (timeout) |  N/A | image |
| Run (user cancellation) | N/A | image |
| Run (OOM/stack overflow) | image | image |
| Debug (OOM/stack overflow) | image | image |
| Copilot run | image | image |
| Copilot run (compile error) | image | image |
| Copilot run (runtime error) | image | image |
| Test Explorer (runtime error) | image | image |

### DIFF
--- a/source/vscode/src/debugger/output.ts
+++ b/source/vscode/src/debugger/output.ts
@@ -108,7 +108,7 @@ function formatMatrix(
 }
 
 /**
- * Formats a Q# error into a human-readable string with file path, line/column, and message.
+ * Formats a QDK error into a human-readable string with file path, line/column, and message.
  * If a stack trace is available, it formats the full trace; otherwise formats a single-line error.
  */
 function formatErrorMessage(error: IQSharpError): string {


source/vscode/src/language-service/testExplorer.ts
```diff
@@ -87,7 +87,9 @@ export function startTestDiscovery(
     }
     const run = ctrl.createTestRun(request);
     const evtTarget = createDebugConsoleEventTarget((msg) => {
-      run.appendOutput(`${msg}\n`);
+      // replace \n with \r\n for proper terminal display
```
ScottCarda-MS, line 90: Is this environment-dependent, or does the VS Code terminal _always_ use \r\n?

minestarks: That's explicitly called out in the VS code API docs. `\r\n` is necessary since these characters are used in the "literal" sense in the terminal implementation (LF just moves the cursor one line down, CR moves it to the beginning of the line)

source/vscode/src/run.ts
```diff
@@ -0,0 +1,268 @@
...
+        buckets: Array.from(buckets.entries()) as [string, number][],
+        shotCount: resultCount,
+      };
+      options.onResultsUpdate?.(histogram, failures);
+      if (
+        options.shots === resultCount ||
+        failures.length > 0 ||
+        options.cancellationToken?.isCancellationRequested
+      ) {
+        // TODO: ugh
```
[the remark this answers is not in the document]
minestarks: Fixed

source/npm/qsharp/src/compiler/common.ts
```diff
@@ -103,6 +111,10 @@ export function eventStringToMsg(msg: string): EventMsg | null {
 
 export type ShotResult = {
   success: boolean;
-  result: string | VSDiagnostic;
+  /**
+   * `VSDiagnostic` is kept for backward compatibility, check runtime type
+   * for `IQSharpError[]` and prefer that when possible.
+   */
+  result: string | (VSDiagnostic & { errors: IQSharpError[] });
```
idavis, line 118: Could we replace these two fields with `Result` defined on line 12 above?

source/vscode/src/debugger/output.ts
```diff
@@ -23,65 +27,125 @@ export function createDebugConsoleEventTarget(
...
+) {
+  return evt.detail.matrix
+    .map((row) =>
+      row.map((entry) => formatComplex(entry[0], entry[1])).join(", "),
+    )
+    .join("\n");
+}
+
+/**
+ * Formats a Q# error into a human-readable string with file path, line/column, and message.
```
idavis: Is this a Q# or a QDK error? The type says IQSharpError, but I think we overload that type

source/npm/qsharp/src/compiler/common.ts
minestarks: Yes, aligning these types would have been the sane thing to do. 

However... these types are ever so slightly different. The field names are different (`value` vs `result`). Also I found a place in the code where a `ShotResult` is set to `{ success: false, result: "", events: [] }`  (a failure value with a `string` result). Even though there's seemingly no good reason for this, I'm very nervous to touch this code now after carefully testing all the edge cases. Maybe in a smaller future PR?

source/vscode/src/debugger/output.ts
minestarks: I'll call it a QDK error in the comment. The type `IQSharpError` was unfortunately named before we generalized everything to QDK.

PR #2792, opened 2025-11-24
Revert Histogram Colors for Dark High Contrast Theme CSS

ScottCarda-MS:
A recent change to the VS Code Dark High Contrast theme caused our histogram widget to be unreadable under that theme. The color for the histogram bars were the same black as the background. This fix addresses that by coloring the histogram bars as they were before the change to the theme, targeting CSS specific to the Dark High Contrast theme.

image

### DIFF
--- a/source/npm/qsharp/ux/qsharp-ux.css
+++ b/source/npm/qsharp/ux/qsharp-ux.css
@@ -357,6 +357,14 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for 
   fill: var(--vscode-button-hoverBackground, var(--nav-hover-background));
 }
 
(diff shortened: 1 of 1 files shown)


source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -357,6 +357,14 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
   fill: var(--vscode-button-hoverBackground, var(--nav-hover-background));
 }
 
+[data-vscode-theme-kind="vscode-high-contrast"] .bar {
```
ScottCarda-MS, line 360: Note that "vscode-high-contrast" is for the dark HC theme. The light HC theme uses "vscode-high-contrast-light".

PR #2796, opened 2025-11-25
Circuit diagrams: group operations by containing callable

minestarks:
This enables a circuit diagram feature where operations are grouped in the visualization based on the containing callable.

Example:

image

### DIFF
--- a/source/compiler/qsc_circuit/src/builder.rs
+++ b/source/compiler/qsc_circuit/src/builder.rs
@@ -379,36 +379,41 @@ impl SourceLookup for PackageStore {
     }
 
     fn resolve_scope(&self, scope_id: ScopeId) -> LexicalScope {
+        let ScopeId(item_id, functor_app) = scope_id;
         let package = self
-            .get(map_fir_package_to_hir(scope_id.0.package))
+            .get(map_fir_package_to_hir(item_id.package))
             .expect("package id must exist in store");
 
         let item = package
             .package
             .items
-            .get(map_fir_local_item_to_hir(scope_id.0.item))
+            .get(map_fir_local_item_to_hir(item_id.item))
             .expect("item id must exist in package");
 
-        let (scope_offset, scope_name) = match &item.kind {
-            hir::ItemKind::Callable(callable_decl) => {
-                let spec_decl = if scope_id.1.adjoint {
-                    callable_decl.adj.as_ref().unwrap_or(&callable_decl.body)
-                } else {
-                    &callable_decl.body
-                };
-
-                (spec_decl.span.lo, callable_decl.name.name.clone())
-            }
-            _ => panic!("only callables should be in the stack"),
+        let hir::ItemKind::Callable(callable_decl) = &item.kind else {
+            panic!("only callables should be in the stack");
         };
+
+        // Use the proper spec declaration from the source code
+        // depending on which functor application was used.
+        let spec_decl = match (functor_app.adjoint, functor_app.controlled) {
+            (false, 0) => Some(&callable_decl.body),
+            (false, _) => callable_decl.ctl.as_ref(),
+            (true, 0) => callable_decl.adj.as_ref(),
+            (true, _) => callable_decl.ctl_adj.as_ref(),
+        };
+
+        let spec_decl = spec_decl.unwrap_or(&callable_decl.body);
+        let scope_start_offset = spec_decl.span.lo;
+        let scope_name = callable_decl.name.name.clone();
 
         LexicalScope::Callable {
             location: PackageOffset {
-                package_id: scope_id.0.package,
-                offset: scope_offset,
+                package_id: item_id.package,
+                offset: scope_start_offset,
             },
             name: scope_name,
-            functor_app: scope_id.1,
+            functor_app,
         }
     }
 }


source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
```diff
@@ -0,0 +1,272 @@
...
+#[test]
+fn two_ops_same_grandparent() {
+    let mut c = FakeCompilation::default();
+    let program = vec![
+        (
+            vec![c.user_code_frame("Main", 1), c.user_code_frame("Foo", 2)],
+            "H",
+        ),
+        (
+            vec![c.user_code_frame("Main", 1), c.user_code_frame("Bar", 3)],
```
ScottCarda-MS, line 101: I guess I expected this to be "Main", 2 because it is the second thing in the "Main" scope.

ScottCarda-MS: Actually, I'm realizing I just don't know what that number is referring to.

source/wasm/src/lib.rs
```diff
@@ -145,15 +143,18 @@ pub(crate) fn get_estimates_from_openqasm(
 
 serializable_type! {
     CircuitConfig,
-    {
+    {generation_method: String,
```
ScottCarda-MS: nit: the formatting on this seems to have gotten a bit off. You just meant to add the `group_scopes: bool,` line, right?

source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
minestarks: That's the source location in the stack entry, so like you can read this as:

"X" at
function `Bar()` on line 3 at
function `Main()` on line 1

minestarks: Source locations are used as differentiators when you have multiple calls to a function, for example

```qsharp
1 operation Foo() : Unit {
2     Bar();
3     Bar();
4 }
5 operation Bar() : Unit {
6     X();
7     Y();
8 }
```

the first two stack traces will be `(Foo, 2) -> (Bar, 6) -> X`, `(Foo, 2) -> (Bar, 7) -> Y`  , and then for the second call to `Bar()`, the stack traces will be `(Foo, 3) -> (Bar, 6) -> X` and  `(Foo, 3) -> (Bar, 7) -> Y` . The differing line numbers in the first frame imply that these are separate calls to `Bar`, so two separate boxes for `Bar` will be created that encapsulate the two sets of `X` and `Y` operations.

ScottCarda-MS: Should your example above be `(Foo, 3) -> (Bar, 7) -> Y` ? Or am I still missing something?

minestarks: 🤦‍♀️  yes, you're right, the `Y`s are on line 7... I'll update my comment to reduce confusion.

source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
```diff
@@ -0,0 +1,272 @@
...
+                c.user_code_frame("F", 10),
+            ],
+            "Y",
+        ),
+    ];
+    check_groups(
+        &c,
+        &program,
+        &expect![[r#"
+            q_0    ─ [ [A] ── [ [B] ── [ [F] ─── Z ──── ] ─── [ [F] ─── Y ──── ] ──── ] ──── ] ──
```
ScottCarda-MS, line 191: So there are two different definitions for `F` being called here?

source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
```diff
@@ -0,0 +1,272 @@
...
+            ],
+            "X",
+        ),
+        (vec![c.user_code_frame("A", 1)], "Y"),
+    ];
+    check_groups(
+        &c,
+        &program,
+        &expect![[r#"
+            q_0    ─ [ [A] ── [ [B] ── [ [C] ─── X ──── ] ──── ] ──── Y ──── ] ──
```
ScottCarda-MS, line 214: Shouldn't `Y` get called in `A` first since it has an earlier line number at 1 vs. the call to `B` which happens at line 2?

source/vscode/src/webviewPanel.ts
```diff
@@ -187,6 +187,7 @@ export function registerWebViewCommands(context: ExtensionContext) {
       async (resource?: vscode.Uri, operation?: IOperationInfo) => {
         await showCircuitCommand(
           context.extensionUri,
+          context.extension.id.includes("-dev"),
```
ScottCarda-MS, line 190: The feature is already gated on if the hidden option is set to "true" in the VSCode settings. Does it need to also be gated on the version number of the QDK being a dev version?

source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
minestarks: No, it's just that the first call to `F` happens to execute a `Z` operation at location 11, and the second call to `F` executes a `Y` operation at location 10.

So for example, the following program could have produced this trace:

```qsharp
@EntryPoint()
operation A() : Unit {
    use q = Qubit();    
    B(q); // LOCATION = 2
}

operation B(q : Qubit) : Unit {
    F(q, false); // LOCATION = 5
    F(q, true) // LOCATION = 6
}

operation F(q : Qubit, applyY : Bool) : Unit {
    if (applyY) {
        Y(q); // LOCATION = 10
    } else {
        Z(q); // LOCATION = 11
    }
}
```

source/compiler/qsc_circuit/src/builder/tests/group_scopes.rs
minestarks: No, source locations are used for differentiation only. Source order has no bearing on execution order.

For example, this trace could have come from the following program:

```qsharp
@EntryPoint()
operation A() : Unit {
    use q = Qubit();   
    for i in 0..1 {
        if i % 2 == 1 {
            Y(q) // LOCATION = 1
        } else {
            B(q) // LOCATION = 2
        }
    }
}

operation B(q : Qubit) : Unit {
    C(q); // LOCATION = 6
}

operation C(q : Qubit) : Unit {
    X(q); // LOCATION = 11
}
```

The `if`/`else` could have easily been flipped so that the call to `Y` comes before the call to `B` in the source.

source/vscode/src/webviewPanel.ts
minestarks: This parameter ends up controlling the default value for the `groupScopes` setting (if the user hasn't explicitly specified it). When in a `"-dev"` build, we want to enable the feature by default. Did that answer your question?

source/wasm/src/lib.rs
minestarks: Oops, fixed!

source/vscode/src/webviewPanel.ts
ScottCarda-MS: yes, thanks for explaining

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -347,6 +377,40 @@ impl SourceLookup for PackageStore {
...
+            .get(map_fir_package_to_hir(scope_id.0.package))
+            .expect("package id must exist in store");
+
+        let item = package
+            .package
+            .items
+            .get(map_fir_local_item_to_hir(scope_id.0.item))
+            .expect("item id must exist in package");
+
+        let (scope_offset, scope_name) = match &item.kind {
```
idavis: Can you please add a comment for why this is necessary?

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -375,13 +443,16 @@ fn resolve_source_location_if_unresolved(
...
 }
 
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Clone, Debug, Copy)]
 pub struct TracerConfig {
     /// Maximum number of operations the builder will add to the circuit
     pub max_operations: usize,
     /// Capture the source code locations of operations and qubit declarations
     /// in the circuit diagram
     pub source_locations: bool,
```
idavis, line 458: nit: `capture_source_locations` or some other name that gives a clue that it is a boolean option. As it stands I'd expect a vec of str or some struct

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -538,6 +610,306 @@ impl WireMapBuilder {
...
+                            result: Some(r.1),
+                        }));
+                    unitary
+                        .targets
+                        .sort_unstable_by_key(|r| (r.qubit, r.result));
+                    unitary.targets.dedup();
+                }
+                Operation::Ket(_) => {}
+            }
+        }
```
idavis: I don't think we need the extra scope

idavis (approved):

source/compiler/qsc_circuit/src/builder.rs
minestarks: Bah, I agree, but now I'm not sure what to do 😞 . These map directly to user-facing configuration options in both VS Code and Python called `source_locations`/`sourceLocations`. We already shipped it with this name and I don't want the internal and user-facing names to diverge :(

image

source/compiler/qsc_circuit/src/builder.rs
minestarks: Why which part is necessary? The function, or this line?

source/compiler/qsc_circuit/src/builder.rs
minestarks: At least, I went ahead and renamed `group_scopes` to `group_by_scope` to avoid a similar confusion.

source/compiler/qsc_circuit/src/builder.rs
minestarks: I added a comment and reworked this function slightly to be clearer (and more correct!)

source/compiler/qsc_circuit/src/builder.rs
minestarks: (Resolving to get this PR done, but happy to discuss further if we want to update these)

PR #2797, opened 2025-11-25
Fixes issue with dragging a control off the last qubit line

ScottCarda-MS:
Dragging the the last element of the last qubit off the grid should delete the element and remove the last qubit line, which is now empty. This PR fixes a bug where the qubit line is not removed if the element being dragged/removed is a control.

### DIFF
--- a/source/npm/qsharp/ux/circuit-vis/events.ts
+++ b/source/npm/qsharp/ux/circuit-vis/events.ts
@@ -170,11 +170,7 @@ class CircuitEvents {
           this.selectedOperation.controls != null &&
           this.selectedWire != null
         ) {
-          const controlIndex = this.selectedOperation.controls.findIndex(
-            (control) => control.qubit === this.selectedWire,
-          );
-          if (controlIndex !== -1)
-            this.selectedOperation.controls.splice(controlIndex, 1);
+          removeControl(this, this.selectedOperation, this.selectedWire);
         } else {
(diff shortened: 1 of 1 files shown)


source/npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -170,11 +170,7 @@ class CircuitEvents {
           this.selectedOperation.controls != null &&
           this.selectedWire != null
         ) {
-          const controlIndex = this.selectedOperation.controls.findIndex(
-            (control) => control.qubit === this.selectedWire,
-          );
-          if (controlIndex !== -1)
-            this.selectedOperation.controls.splice(controlIndex, 1);
+          removeControl(this, this.selectedOperation, this.selectedWire);
```
billti, line 173: If the `removeControl` helper was already present, why was it not used previously? (i.e. what more complex logic did the prior implementation think it was needed for?)

ScottCarda-MS: This was likely just a missed location where the `removeControl` logic should have been called. The replaced logic removes the control in the same way as the helper does, but the helper goes a step further and manages the `qubitUseCounts` field of the circuit, which is used in deciding if qubit lines should get removed when redrawing. Since that management logic was missing here before, the qubit lines weren't getting properly removed.

billti (approved):

PR #2802, opened 2025-12-04
Add optional `prune_classical_qubits` setting for circuit generation

swernli:
This change adds a parameter to Q# and OpenQASM circuit generation called `prune_classical_qubits` that will generate a simplified circuit diagram where qubits that are unused or purely classical (never enter superposition) are removed. This can be handy for places where the program comes from generated logic that may use more qubits than strictly needed. This is purely for visualizing the simplified circuit and does not perform any kind of execution optimization.

So a program like this that has nine qubits but doesn't really use all of them:
image

will by default show the full circuit as it does today:
image

but with `prune_classical_qubits = True` will remove the purely classical q1, q2, q4, q5, and q8:
image

### DIFF
--- a/source/pip/qsharp/openqasm/_circuit.py
+++ b/source/pip/qsharp/openqasm/_circuit.py
@@ -47,8 +47,8 @@ def circuit(
 
     max_operations = kwargs.pop("max_operations", None)
     generation_method = kwargs.pop("generation_method", None)
-    source_locations = kwargs.pop("source_locations", None)
-    group_by_scope = kwargs.pop("group_by_scope", None)
+    source_locations = kwargs.pop("source_locations", False)
+    group_by_scope = kwargs.pop("group_by_scope", False)
     prune_classical_qubits = kwargs.pop("prune_classical_qubits", False)
     config = CircuitConfig(
         max_operations=max_operations,


source/pip/qsharp/_qsharp.py
```diff
@@ -838,6 +838,7 @@ def circuit(
     max_operations: Optional[int] = None,
     source_locations: bool = False,
     group_by_scope: bool = False,
+    trim: bool = False,
```
minestarks: Is `trim` established technical jargon? It sounds very general to me., and it may get muddy as these config switches proliferate (which they definitely will). Spitballing a couple more descriptive names:

`trim_unnecessary_qubits`?
`trim_basis_state_qubits`?

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -42,6 +42,7 @@ pub struct CircuitTracer {
     circuit_builder: OperationListBuilder,
     next_result_id: usize,
     user_package_ids: Vec<PackageId>,
+    nontrimmable_qubits: FxHashSet<usize>,
```
minestarks: A more descriptive name, such as `used_qubits`, `necessary_qubits`, `qubits_in_superposition` or `qubits_ever_in_superposition`?

(If you have to stick with `nontrimmable` - at least `untrimmable` sounds more natural to my ears)

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -220,16 +227,36 @@ impl CircuitTracer {
         operations: &[OperationOrGroup],
         source_lookup: &impl SourceLookup,
     ) -> Circuit {
-        let operations = operations
+        let mut operations: Vec<Operation> = operations
```
minestarks: Because `operations` can contain grouped operations, I think you'll need to handle the trimming at an earlier step before the grouping occurs. `push_op` maybe?  This should save you from having to recurse here.

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -311,6 +338,60 @@ impl CircuitTracer {
...
+    fn update_qubit_status(&mut self, name: &str, targets: &[usize], controls: &[usize]) {
+        match name {
+            "H" | "Rx" | "Ry" | "SX" | "Rxx" | "Ryy" => {
+                // These gates create superpositions, so mark the qubits as non-trimmable
+                for &q in targets {
+                    let mapped_q = self.wire_map_builder.wire_map.qubit_wire(q);
+                    self.nontrimmable_qubits.insert(mapped_q.into());
+                }
+            }
+            "Rzz"
```
DmitryVasilevsky: Not sure I understand. If I'm not mistaken [Rzz gate is symmetric](https://microsoft.github.io/qdk/?code=H4sIAAAAAAAACssvSC1KLMnMz1PwTczM09BUsFJwys%252FPUajmUlBQUAguSdFzyUxMz8svLslMLtZzzkhNzvaHaSl2LEp1LSxNzNEAKwYBIx04M1HB1k4hqKpKw0DPXEchMdogFkQaxmoilCShKEmKNowFkQaxmmAVmly1XABsoCnRnwAAAA%253D%253D) - you can swap qubits and it's going to be the same. Why this asymmetry then?

source/compiler/qsc_circuit/src/builder/tests.rs
```diff
@@ -405,3 +416,238 @@ fn qubit_labels_for_preallocated_qubits() {
...
+    .assert_eq(&circuit.to_string());
+}
+
+#[test]
+fn bell_circuit_keeps_classical_qubit_when_second_qubit_in_rzz_is_not_classical() {
+    let mut builder = CircuitTracer::new(
+        TracerConfig {
+            max_operations: 100,
+            source_locations: false,
+            group_by_scope: false,
```
minestarks: Suggest having a couple cases where `group_by_scope` is `true` to check the potential issue I mentioned above

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -311,6 +338,60 @@ impl CircuitTracer {
         }
         first_user_code_location(&self.user_package_ids, stack)
     }
+
+    fn update_qubit_status(&mut self, name: &str, targets: &[usize], controls: &[usize]) {
+        match name {
+            "H" | "Rx" | "Ry" | "SX" | "Rxx" | "Ryy" => {
```
DmitryVasilevsky, line 401: Rx and Ry are the same as X, Y and I for some angles so they could be trimmable in those cases. But I assume if someone used Rx instead of X, that makes a qubit interesting enough not to trim.
BTW, this "interesting" status may give ideas for a name of non-trimmable qubits.

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -311,6 +338,61 @@ impl CircuitTracer {
...
+                    .insert(self.wire_map_builder.wire_map.qubit_wire(targets[0]).into());
+            }
+            "X" | "Y" | "Z"
+                if !controls.is_empty()
+                    && controls.iter().all(|q| {
+                        self.nontrimmable_qubits
+                            .contains(&self.wire_map_builder.wire_map.qubit_wire(*q).into())
+                    }) =>
+            {
+                // Controlled Pauli gates with controls that are non-trimmable make the targets non-trimmable
```
DmitryVasilevsky: I cannot deduce what you require here from the comment - all controls or some controls. Consider making comment clearer. But from the code it looks like target becomes non-trimmable only if all controls are not trimmable. This is suspicious. If I have two controls, one of which is in |1> state, say after X gate - it is trimmable. So, the target will become trimmable. However, isn't this equivalent to a singly controlled gate on a second control? If that second control is non-trimmable, the target should become non-trimmable...

swernli: You're right... I updated this logic from `any` to `all` to avoid a different issue I found, but that introduces exactly the problem you describe. I've got to think this through; the reference code I'm using doesn't handle CCX at all, so that doesn't help me here!

source/compiler/qsc_circuit/src/builder.rs
swernli: You're right, this doesn't do grouping right. Unfortunately, I don't think I can avoid having the recursion here, because I don't want to drop gates on a qubit that is trimmable at that point but becomes non-trimmable later. The intent is not to prune gates but to prune wires; all gates that operate on a wire that is kept should be kept and all gates that involve a pruned wire should be pruned. I don't know which wires are going to be pruned until the end of execution.

source/pip/qsharp/_qsharp.py
DmitryVasilevsky: Made my own comment outside of the code to explain why I don't like the 'trim' name.

source/compiler/qsc_circuit/src/builder.rs
swernli: Yes... I think I was overthinking this by imagining replacing Rzz with its decomposition. But taken at it's face, Rzz doesn't introduce any new superposition, so I can move it up to the other case so it gets treated like a Pauli gate or Rz. For that matter, neither does CZ, so I think I can also move that...

source/compiler/qsc_circuit/src/builder.rs
swernli: Yes, agreed. Maybe in the future we could update this to check angle paramters for rotations, but for now I think it's ok to treat any Rx or Ry as "interesting."

source/compiler/qsc_circuit/src/builder.rs
swernli: I think I've got the right logic now. The code now tracks which qubits can be assumed to be classical 1 and updates the controls based on this assumption. This may result in CCX becoming CX or even just X (same with Y).

source/compiler/qsc_circuit/src/builder.rs
swernli: I went with `superposition_qubits`

source/compiler/qsc_circuit/src/builder.rs
swernli: I did end up going a recursive route with this, which allows me to maintain the gates I don't want to lose along the way. But now the logic is definitely much more complex.

source/compiler/qsc_circuit/src/builder/tests.rs
swernli: I'm going to need a lot more tests now the that the recusive logic handles this...

source/pip/qsharp/_qsharp.py
swernli: I renamed it to `prune_classical_qubits`.

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -220,16 +237,55 @@ impl CircuitTracer {
...
+            qubits.retain(|q| self.superposition_qubits.contains(&q.id));
+
+            // Remove operations that don't use any non-classical qubits.
+            operations.retain_mut(|op| self.should_keep_operation_mut(op));
+        }
+
+        finish_circuit(qubits, operations, num_qubits, source_lookup)
+    }
+
+    fn should_keep_operation_mut(&self, op: &mut Operation) -> bool {
```
minestarks: Can this operate on `OperationOrGroup` _before_ we transform it into `Operation`? `OperationOrGroup` is the "working" data structure which is more ergonomic to manipulate, and already has methods like `all_qubits()` . `Operation` is essentially the final "wire format" that we serialize to JSON, so we have to keep it stable but is kind of annoying to manipulate. You shouldn't have to change this function too much.

source/compiler/qsc_circuit/src/circuit.rs
```diff
@@ -214,6 +214,28 @@ impl Operation {
...
+            Operation::Measurement(m) => m.qubits.clone(),
+            Operation::Unitary(u) => {
+                let mut qubits = u.targets.clone();
+                qubits.extend_from_slice(&u.controls);
+                qubits
+            }
+            Operation::Ket(k) => k.targets.clone(),
+        }
+    }
+
```
minestarks: Please see above comment - if you use `OperationOrGroup` when manipulating the circuit, you shouldn't have to add these methods

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -42,6 +42,8 @@ pub struct CircuitTracer {
     circuit_builder: OperationListBuilder,
     next_result_id: usize,
     user_package_ids: Vec<PackageId>,
+    superposition_qubits: FxHashSet<usize>,
```
minestarks: These are "wire ID"s right? Can we leave the type as `QubitWire` for clarity?

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -311,6 +367,105 @@ impl CircuitTracer {
         }
         first_user_code_location(&self.user_package_ids, stack)
     }
+
+    fn mark_qubit_in_superposition(&mut self, wire: QubitWire) {
+        self.superposition_qubits.insert(wire.into());
```
minestarks: Perhaps we can have an assert here (and in `flip_classical_qubit` below) that the `prune_classical_qubits` config is set to true. Reading this, I get a bit nervous that we might be doing this work regardless of config flag. I can see you obviously took it into consideration this in this PR, but it feels like it could be accidentally regressed with a future change.

source/compiler/qsc_circuit/src/builder.rs
minestarks: Since this is expected to be very dense with a range of `0..max_qubits` ... is `FxHashSet<usize>` still the right choice or should we use `Vec<bool>`? 

(I can totally buy "it doesn't matter for circuits this small" though)

source/pip/src/interpreter.rs
```diff
@@ -1303,23 +1304,27 @@ pub(crate) struct CircuitConfig {
...
 impl CircuitConfig {
     #[new]
-    #[pyo3(signature=(*,max_operations=None, generation_method=None, source_locations=None, group_by_scope=None))]
+    #[pyo3(signature=(*,max_operations=None, generation_method=None, source_locations=None, group_by_scope=None, prune_classical_qubits=false))]
     fn new(
         max_operations: Option<usize>,
         generation_method: Option<CircuitGenerationMethod>,
         source_locations: Option<bool>,
         group_by_scope: Option<bool>,
+        prune_classical_qubits: bool,
```
minestarks, line 1327: I can't process why this flag looks different than the others. Why not `Option<bool>`? Should the others have been `bool`?

source/wasm/src/lib.rs
```diff
@@ -181,6 +181,7 @@ pub fn get_circuit(
         source_locations: config.source_locations,
         max_operations: config.max_operations,
         group_by_scope: config.group_by_scope,
+        prune_classical_qubits: false,
```
minestarks, line 184: Don't we want to plumb this up to VS Code?

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -220,16 +237,55 @@ impl CircuitTracer {
...
         source_lookup: &impl SourceLookup,
     ) -> Circuit {
-        let operations = operations
+        let mut operations: Vec<Operation> = operations
             .iter()
             .map(|o| o.clone().into_operation(source_lookup))
             .collect();
+        let mut qubits = self.wire_map_builder.wire_map.to_qubits();
+        // We need to pass the original number of qubits, before any trimming, to finish the circuit below.
+        let num_qubits = qubits.len();
```
minestarks, line 243: Just curious, why do we need the original `num_qubits`? _Should_ we need it? I'm not familiar with the code in `operation_list_to_grid` so I'm genuinely asking to see if there's something that can be improved there.

source/wasm/src/lib.rs
swernli: No, I don't think we want that yet (see the longer response to Mathias in the top-level PR chat). I think if we do circle back to exposing this in VS Code, we'd want more strategies for visualizing what changed in the circuit. For now, the intent is just to provide an optional visualization technique from Python for chemistry scenarios. With the infra present after this PR, we could investigate and design some ways we'd want to more fully integrate this feature in the extension.

source/pip/src/interpreter.rs
swernli: The others never needed to be `Option<bool>` since `bool` passes natively back and forth through the Python interop. So rather that `Option<bool>` with default `None` they should all be `bool` with default `false`. But I didn't want to change the existing flags as part of this PR. Happy to do it if you feel it makes sense to include here though!

source/compiler/qsc_circuit/src/builder.rs
swernli: That's a tricky one that I debated. On the one hand, if every qubit ends up in there then it will hold `0..max_qubits` in a dense arrangement, as you say. But for the circuits someone cares about trimming, how many qubits will never be in superposition and trimmed out? A third? Half? How big even is the circuit, since we are limiting this to things someone would reasonably try to visualize? In the end, I picked `FxHashSet` for pragmatism and ease of use, since I don't know which order qubits will be addressed and `Vec<bool>` would require allocating up to the current ID and filling with `false`. Our `IndexMap` type helps some with this, but also isn't quite the right fit since I just care about membership and not the mapped value. We could introduce `IndexSet` with similar internal helpers... but by this point in the consideration I figured the scale/scope is small enough that I'm probably overthinking it and just picked `FxHashSet`.

source/compiler/qsc_circuit/src/builder.rs
swernli: Good point, that would definitely simiplify things and let me reuse those helpers. I'll try to make that shift.

source/compiler/qsc_circuit/src/builder.rs
swernli: Sure, seems reasonable.

swernli: Updated!

source/compiler/qsc_circuit/src/builder.rs
swernli: Added asserts for safety.

source/compiler/qsc_circuit/src/builder.rs
swernli: There is a spot deeper in `operation_list_to_grid` that has an assumption built into it that the highest qubit ID is `num_qubits - 1`. I tried to see if there was a way to adjust this assumption, but it ended up getting a bit messy and felt off-topic, so it was easier to pass the original number of qubits so that the internal assumptions held. Definitely seems like a place where refactoring the grid conversion logic could help, as a potential follow up item.

source/compiler/qsc_circuit/src/circuit.rs
swernli: I did end up having to keep `targets_mut` as I needed a way to prune the targets of the operation representing a group.

source/compiler/qsc_circuit/src/builder.rs
swernli: Made the adjustments, and it's definitely cleaner, thanks!

source/pip/src/interpreter.rs
minestarks: If you don't mind doing it, I'd really appreciate it if you fixed all of them!

PR #2804, opened 2025-12-08
Split OpenQASM into parser and compiler crates

idavis:


source/compiler/qsc_data_structures/src/error.rs
orpuente-MS, line 1: Should this file be called `with_source.rs`? It doesn't really implement any errors, just the `WithSource` wrapper.

source/compiler/qsc_openqasm_compiler/src/tests/output.rs
```diff
@@ -464,3 +464,65 @@ fn qir_generation_for_box_with_qdk_qir_intrinsic() -> miette::Result<(), Vec<Rep
 
     Ok(())
 }
+
+#[test]
+fn bigint_output_is_initially_zero_l() {
```
orpuente-MS, line 469: What does the `_l` suffix on this test's name mean?

source/compiler/qsc_openqasm_compiler/src/compiler.rs
```diff
@@ -1190,7 +1200,11 @@ impl QasmCompiler {
 
     fn compile_pragma_stmt(&mut self, stmt: &semast::Pragma) {
         fn is_parameterless_and_returns_void(args: &Arc<[Type]>, return_ty: &Arc<Type>) -> bool {
-            args.is_empty() && matches!(&**return_ty, crate::semantic::types::Type::Void)
+            args.is_empty()
+                && matches!(
+                    &**return_ty,
+                    qsc_openqasm_parser::semantic::types::Type::Void
```
orpuente-MS, line 1206: These paths to `Type` are getting quite long. Maybe we should `use qsc_openqasm_parser::semantic` to make them a bit shorter. Or even `use qsc_openqasm_parser::semantic::types::Type as SemanticType`.

source/compiler/qsc_openqasm_compiler/src/lib.rs
```diff
@@ -44,16 +37,25 @@ pub struct Error(pub ErrorKind);
...
-    pub fn is_semantic_error(&self) -> bool {
-        matches!(self.0, ErrorKind::Semantic(..))
+    pub fn is_parse_error(&self) -> bool {
+        matches!(self.0, ErrorKind::Parser(..))
     }
 }
 
+pub(crate) fn get_compile_errors_from_parse_result(
+    res: &QasmSemanticParseResult,
+) -> Vec<WithSource<crate::Error>> {
```
orpuente-MS: Something feels odd about this function's name. Do we get compilation errors from parsing? Should it be get_semantic_errors_from_lowering_result instead?
Also, should that struct be called QasmSemanticLoweringResult?

source/compiler/qsc_openqasm_parser/src/semantic/types/tests.rs
```diff
@@ -0,0 +1,39 @@
...
+
+//     check_qasm_to_qsharp(
+//         source,
+//         &expect![[r#"
+//         import Std.OpenQASM.Intrinsic.*;
+//         mutable arr_1 = [[[false, false]], [[false, false]], [[false, false]], [[false, false]], [[false, false]]];
+//         mutable arr_2 = arr_1[1..3];
+//     "#]],
+//     );
+// }
```
orpuente-MS: Why is this file commented out?

source/compiler/qsc_data_structures/src/error.rs
idavis: This is the name of the file as it was in the original and contents are 99% the same. Just trying to keep the nename of file clean.

source/compiler/qsc_openqasm_compiler/src/tests/output.rs
idavis: bigint

source/compiler/qsc_openqasm_compiler/src/compiler.rs
idavis: We can consider it, but I don't want refactoring in the middle of this.

source/compiler/qsc_openqasm_parser/src/semantic/types/tests.rs
idavis: Forgot to move them, thanks!

source/compiler/qsc_openqasm_compiler/src/lib.rs
idavis: Renamed the function. As metioned in other comments, I'm not renaming a bunch of types as part of the PR to avoid issues.

PR #2808, opened 2025-12-09
Bump version to 1.23

billti:


source/vscode/changelog.md
```diff
@@ -1,5 +1,48 @@
...
+```python
+provider = AzureQuantumProvider(workspace)
+backend = provider.get_backend(target_name)
+job = backend.run(circuit, shots, job_name)
+counts = job.result().get_counts(circuit)
+```
+
+See [the updated Qiskit submission sample notebook](https://github.com/microsoft/qdk/blob/main/samples/python_interop/submit_qiskit_circuit_to_azure.ipynb) for the new supported method of Azure submission with Qiskit 2.
+
+To make sure you get the updated `qdk` package with this support, please use the command `pip install "qdk[azure]" --upgrade`
```
swernli: Does "qdk[azure]" automatically get Qiskit, or is a different option?

source/vscode/changelog.md
```diff
@@ -1,5 +1,48 @@
...
+
+### Interactive Circuit Diagrams with Source Code Navigation
+
+Circuit diagrams now display clickable source code locations for gates and qubits in VS Code. Click on any operation box to jump directly to where it was called in your Q# or QASM code, or on a qubit label to jump to its declaration site. In Python Jupyter notebooks, source locations can be enabled via `qsharp.circuit(source_locations=True)` to display hover text with code locations.
```
swernli, line 24: Is operation grouping also enabled by default, or is that just in the dev build?

swernli (approved):

source/vscode/changelog.md
billti: Qiskit is a different "extra", but you don't need to update it for this to work. You do need to update the 'azure' extra though.

source/vscode/changelog.md
billti: Only in dev builds by default for now.

source/vscode/changelog.md
billti: Updated to add it.

PR #2809, opened 2025-12-09
Add ability to switch between debug and no-debug execution graph

minestarks:
Currently the FIR lowerer can generate two flavors of `ExecGraph`. When the `dbg` flag is passed to the lowerer, the `ExecGraph` contains instructions to track block scopes. Without the `dbg` flag, these instructions are elided for performance.

Currently the `dbg` flag is set at the initialization of the `Interpreter`, and cannot be changed over its lifetime.

This change makes it so that we can dynamically switch between the debug and no-debug `ExecGraph` configurations depending on what kind of evaluation we're performing.

The motivation for this change is the following scenario:

```python
from qsharp import eval, circuit, run
eval("operation Foo() : Unit { use q = Qubit(); M(q); }")
run("Foo()")
circuit("Foo()")
```

Here, evaluating `Foo()` via the call to `circuit()` will need to use the debug `ExecGraph` . This is because in the future, the circuit diagrams generated with `circuit()` will show grouping of operations based on loop block scopes. So we'll need the `ExecGraph` that contains these block scope tracking instructions.

On the other hand, `run`, `eval` etc. need to use the no-debug `ExecGraph` for the best performance.

### DIFF
--- a/source/compiler/qsc/src/interpret.rs
+++ b/source/compiler/qsc/src/interpret.rs
@@ -53,7 +53,7 @@ pub use qsc_eval::{
 };
 use qsc_fir::{
     fir::{
-        self, Block, BlockId, ExecGraphConfig, ExecGraph, Expr, ExprId, Global, Package, PackageId,
+        self, Block, BlockId, ExecGraph, ExecGraphConfig, Expr, ExprId, Global, Package, PackageId,
         PackageStoreLookup, Pat, PatId, Stmt, StmtId,
     },
     visit::{self, Visitor},


source/compiler/qsc/src/interpret.rs
```diff
@@ -358,6 +360,11 @@ impl Interpreter {
             classical_seed: None,
             package,
             source_package: map_hir_package_to_fir(source_package_id),
+            eval_config: if dbg {
+                ExecGraphConfig::Debug
+            } else {
+                ExecGraphConfig::NoDebug
+            },
```
idavis: nit: should this be pushed up to avoid passing the boolean?

minestarks: I can do that. It will likely propagate to a bunch of files - wasn't sure if that was going to be a pain to review.

source/compiler/qsc_fir/src/fir.rs
```diff
@@ -907,8 +907,116 @@ impl Display for SpecDecl {
     }
 }
 
+#[derive(Clone, PartialEq, Debug, Default)]
```
idavis, line 910: Do we need every derive listed for these types?

source/compiler/qsc/src/interpret.rs
swernli: I think it's fine either way, but if you are willing, it would be nice to have the enum propagated up for ease of reading.

swernli (approved):

source/compiler/qsc/src/interpret.rs
minestarks: Actually, I realized I probably want to _keep_ the `dbg` parameter in the [one _public_ constructor that uses it](https://github.com/microsoft/qdk/blob/46b6d1ad4c66d4936fa40e3e5e89ee7d95b30963/source/compiler/qsc/src/interpret.rs#L282-L303), since that parameter controls more than one thing. `dbg` controls both `TracerConfig` (debugging enables circuit tracing) and `ExecGraphConfig` (execute debug nodes). So it makes sense to keep it as a differently named parameter.

I'll still plumb the `eval_config` flag explicitly in the private methods for ease of reading.

source/compiler/qsc_fir/src/fir.rs
minestarks: Kind of. `Clone` and `Default` are explicitly used. `Debug` is imposed by `PackageStore`, `PartialEq` by `Item`, but these are ultimately unused. Do you want me to prune them out while I'm at it? It's a lot of changes.

minestarks: Actually, it's a string that keeps unraveling the more I pull on it. I may make a separate PR for everything in the FIR.

PR #2821, opened 2025-12-18
Infallible `PackageId` in `hir::ItemId`

### LINKED ISSUE #2815: Q# structs from Q# projects cannot be passed as arguments in Python
**Describe the bug**

Evaluating a Q# operation with a Q# struct argument loaded from a Q# project fails in Python, but not when evaluated directly not loaded from a project.

**To Reproduce**

This works:

```python
# test.py

import qsharp


qsharp.eval("""

operation Test1(value: Int) : Unit {
}

struct Foo { value: Int }

operation Test2(value: Foo) : Unit {
}

""")

counts = qsharp.logical_counts(qsharp.code.Test1, 42)
print(counts)

arg = qsharp.code.Foo(42)
counts = qsharp.logical_counts(qsharp.code.Test2, arg)
print(counts)
```

This does not work:

```json
// project/qsharp.json

{}
```

```qsharp
// project/src/Main.qs

operation Test1(value: Int) : Unit {
}

struct Foo { value: Int }

operation Test2(value: Foo) : Unit {
}
```

```python
# test.py

from pathlib import Path
import qsharp

qsharp.init(project_root=str(Path(__file__).parent / "project"))

counts = qsharp.logical_counts(qsharp.code.Main.Test1, 42)
print(counts)

arg = qsharp.code.Main.Foo(42)
counts = qsharp.logical_counts(qsharp.code.Main.Test2, arg) # FAILS HERE
print(counts)
```

It fails with `pyo3_runtime.PanicException: item should be in this package`

**Expected behavior**

There should not be a difference whether the code is loaded via `qsharp.eval` or `qsharp.init(project_root=...)`

**System information**

- qdk==1.23.0

orpuente-MS:
Fixes #2815

The cause of #2815 is a recurrent issue rooted in that we don't know the `PackageId` of packages until we insert them in the `PackageStore`. So, we have to jump through hoops in multiple places in our codebase to "resolve" the package ID of items, some of those resolutions are non-trivial. We've had to fix other bugs like this in the past; this is another instance where we didn't resolve a `PackageId` correctly.
 
All this pain is unnecessary because whenever we compile, we already have the `PackageStore` available, so, we can ask it to give us a `PackageId` ahead of time and use that in the compiler, so that all items are compiled with an infallible package ID and not an `Option::None` that we need to resolve later. That is what this PR does.

As an extra, `ItemId` is now 16 bytes instead of 24. This saves ~0.5 MB of memory when compiling 'core + stdlib', as shown in the Memory Report.

### DIFF
--- a/source/fuzz/fuzz_targets/qasm.rs
+++ b/source/fuzz/fuzz_targets/qasm.rs
@@ -52,7 +52,6 @@ fn compile(data: &[u8]) {
                 package,
                 sources,
                 PackageType::Lib,
-                store.peek_next_package_id(),
                 profile.into(),
             );
         });
--- a/source/fuzz/fuzz_targets/qsharp.rs
+++ b/source/fuzz/fuzz_targets/qsharp.rs
@@ -14,9 +14,8 @@ fn compile(data: &[u8]) {
         thread_local! {
             static STORE_STD: (PackageStore, PackageId) = {
                 let mut store = PackageStore::new(qsc::compile::core());
-                let std_id = store.new_package_id();
-                store.insert(std_id, qsc::compile::std(std_id, &store, Profile::Unrestricted.into()));
-                (store, std_id)
+                let std = store.insert(qsc::compile::std(&store, Profile::Unrestricted.into()));
+                (store, std)
             };
         }
         let sources = SourceMap::new([("fuzzed_code".into(), fuzzed_code.into())], None);
@@ -26,7 +25,6 @@ fn compile(data: &[u8]) {
                 &[(*std, None)],
                 sources,
                 qsc::PackageType::Lib,
-                store.peek_next_package_id(),
                 Profile::Unrestricted.into(),
                 LanguageFeatures::default(),
             );
--- a/source/pip/src/interop.rs
+++ b/source/pip/src/interop.rs
@@ -681,14 +681,12 @@ fn create_interpreter_from_ast(
     let (stdid, mut store) = qsc::compile::package_store_with_stdlib(capabilities);
     let dependencies = vec![(PackageId::CORE, None), (stdid, None)];
 
-    let source_package_id = store.new_package_id();
     let (mut unit, errors) = qsc::compile::compile_ast(
         &store,
         &dependencies,
         ast_package,
         source_map,
         package_type,
-        source_package_id,
         capabilities,
     );
 
@@ -697,7 +695,7 @@ fn create_interpreter_from_ast(
     }
 
     unit.expose();
-    store.insert(source_package_id, unit);
+    let source_package_id = store.insert(unit);
 
     interpret::Interpreter::with_package_store(
         false,
--- a/source/samples_test/src/tests.rs
+++ b/source/samples_test/src/tests.rs
@@ -142,14 +142,12 @@ fn compile_and_run_qasm_internal(source: &str, debug: bool) -> String {
     let (stdid, mut store) = qsc::compile::package_store_with_stdlib(capabilities);
     let dependencies = vec![(PackageId::CORE, None), (stdid, None)];
 
-    let source_package_id = store.new_package_id();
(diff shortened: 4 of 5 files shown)


source/compiler/qsc_frontend/src/compile.rs
```diff
@@ -257,6 +273,7 @@ pub fn compile_ast(
     store: &PackageStore,
     dependencies: &Dependencies,
     mut ast_package: ast::Package,
+    package_id: PackageId,
```
swernli: The addition of `peek_next_package_id` made me think this signature change might be unnecessary. The `store` is already an argument here, so we could have a pattern like this:

 - `compile_ast` uses `peek_next_package_id` to get the expected id, uses that to compile the given sources
 - `CompileUnit` has a field or accessor added to it that says what package id that unit was compiled with
 - `PackageStore::insert` still performs the increment of the package id tracking, but first confirms/asserts the current value of package id that will be used for insertion matches what is in the given `CompileUnit`

 I think with that adjustment, most of the changes in the PR go away, because the calling pattern remains the same.

source/compiler/qsc_frontend/src/compile.rs
```diff
@@ -117,13 +117,27 @@ impl PackageStore {
         &self.core
     }
 
-    pub fn insert(&mut self, unit: CompileUnit) -> PackageId {
+    /// Returns the next package ID and advances the inner `next_id`.
+    ///
+    /// If you need an ID for a throw-away compilation unit,
+    /// look at [`Self::peek_next_package_id`].
+    #[must_use]
+    pub fn new_package_id(&mut self) -> PackageId {
```
minestarks: Naming: `next_package_id()` and `peek_package_id()` sound more familiar to my ears than the ones you chose. These would echo the names of `iter` methods, and communicate behavior clearly.

source/compiler/qsc/src/codegen.rs
```diff
@@ -46,7 +48,7 @@ pub mod qir {
             return Err(errors.iter().map(|e| Error::Compile(e.clone())).collect());
         }
 
-        let package_id = store.insert(unit);
+        store.insert(package_id, unit);
```
minestarks: I was wondering - to avoid disagreement between the actual `package_id` and what's in the `CompileUnit`, would it be a good change to just add `package_id` as a field to `CompileUnit`? I notice these two types also get passed around together a lot as arguments, so maybe coupling them would reduce verboseness?

minestarks: Oh! This suggestion was already made by @swernli , never mind!

swernli: great minds think alike ;)

source/compiler/qsc_frontend/src/compile.rs
```diff
@@ -117,8 +133,18 @@ impl PackageStore {
...
+        self.next_id
+    }
+
     pub fn insert(&mut self, unit: CompileUnit) -> PackageId {
         let id = self.next_id;
+        assert_eq!(
+            id,
+            unit.package_id(),
+            "The id of the inserted unit should match the next_id of the store."
+        );
```
orpuente-MS, line 147: Here is that assert we talked about.

source/compiler/qsc/src/codegen.rs
orpuente-MS: I actually added the `package_id` to the `Package` struct instead of adding it to `CompileUnit`, since in a couple of places we need to pass the package along with its `package_id`. I also added an accessor `CompileUnit::package_id()` that returns `self.package.package_id`.

source/compiler/qsc_frontend/src/compile.rs
orpuente-MS: Good idea. I applied this suggestion. Thank you!

PR #2822, opened 2025-12-22
Use 'build' package to build wheels

idavis:
Switches from `pip` wheel command to `build` package. Updates maturin. Simplifies `manylinux` handling offloading it to `maturin` which will handle tagging to the lowest compatability level.


idavis:
@minestarks It should, The only change to the pyproject.toml is the version of `maturin`

PR #2824, opened 2025-12-29
Add Cirq and Rigetti content from Azure-Quantum to QDK package

ScottCarda-MS:
Adds the cirq extra, `qdk.azure.cirq`, and `qdk.azure.target.rigetti`.

### DIFF
--- a/samples/python_interop/cirq_submission_to_azure.ipynb
+++ b/samples/python_interop/cirq_submission_to_azure.ipynb
@@ -32,7 +32,7 @@
    "source": [
     "## Prerequisites\n",
     "\n",
-    "Ensure the `qdk` package (with Azure extra) and `cirq` are installed. If not, install dependencies below."
+    "Ensure the `qdk` package is installed with `azure` and `cirq` extras. If not, install dependencies below."
    ]
   },
   {
@@ -42,7 +42,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install \"qdk[azure]\" cirq"
+    "%pip install \"qdk[azure,cirq]\""
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


source/qdk_package/pyproject.toml
```diff
@@ -16,10 +16,13 @@ dependencies = ["qsharp==0.0.0"]
 jupyter = ["qsharp-widgets==0.0.0", "qsharp-jupyterlab==0.0.0"]
 azure = ["azure-quantum>=3.5.0"]
 qiskit = ["qiskit>=1.2.2,<3.0.0"]
+cirq = ["cirq-core>=1.3.0,<=1.4.1", "cirq-ionq>=1.3.0,<=1.4.1"]
 all = [
   "qsharp-widgets==0.0.0",
   "azure-quantum>=3.5.0",
   "qiskit>=1.2.2,<3.0.0",
+  "cirq-core>=1.3.0,<=1.4.1",
+  "cirq-ionq>=1.3.0,<=1.4.1",
```
billti, line 25: Why do we need a separate `cirq-ionq` package? Is this similar to the `qiskit-ionq` dependency @idavis removed (https://github.com/microsoft/azure-quantum-python/pull/694/changes#diff-7be6800e2a7d73f55566f4819bbc22349cf365d410a0d50d3efe746b9c3c6e64)? If so, should we do something similar here?

billti: (Actually, I see this is a package by cirq itself (https://github.com/quantumlib/Cirq/tree/main/cirq-ionq), so possibly unrelated to the pattern that was used by qiskit-ionq)

ScottCarda-MS: These are the requirements pulled from azure quantum: https://github.com/microsoft/azure-quantum-python/blob/main/azure-quantum/requirements-cirq.txt. @idavis suggested that we reference this dependency directly, though, by having the azure extra depend on azure-quantum[cirq], which is an idea I like as it removes the need for an additional "cirq" specific extra.

ScottCarda-MS: We decided in team chat to go ahead with a cirq extra and to use direct references to cirq dependencies.

PR #2825, opened 2025-12-30
Updated comment to trig functions to mention radians

DmitryVasilevsky:
As discussed elsewhere,

there's only one way to specify argument for trig functions and we already use it. We provide PI() so it should be convenient. Having said that, degrees are taught in school, so the only reason we may consider doing something about it is to facilitate education somehow.
I probably mentioned this before, but "radians" are just numbers. They are ratios of two lengths. So, no matter the unit for the length, the ration is just a number.
As for the issue, I don't see much harm in mentioning that we are using radians. They are part of the International System of Units after all. Although as a "derived unit". Degrees don't have the honor. They are just "acceptable". (https://www.bipm.org/documents/20126/41483022/SI-Brochure-9.pdf, page 145)

Resolves https://github.com/microsoft/qdk/issues/2773

### DIFF
--- a/library/std/src/Std/Math.qs
+++ b/library/std/src/Std/Math.qs
@@ -250,79 +250,85 @@ function Min(values : Int[]) : Int {
 
 /// # Summary
 /// Returns the angle whose cosine is the specified number.
+/// Returned value is expressed in radians.
 function ArcCos(x : Double) : Double {
     body intrinsic;
 }
 
 /// # Summary
 /// Returns the angle whose sine is the specified number.
+/// Returned value is expressed in radians.
 function ArcSin(y : Double) : Double {
     body intrinsic;
 }
 
 /// # Summary
 /// Returns the angle whose tangent is the specified number.
+/// Returned value is expressed in radians.
 function ArcTan(d : Double) : Double {
     body intrinsic;
 }
 
 /// # Summary
 /// Returns the angle whose tangent is the quotient of two specified numbers.
+/// Returned value is expressed in radians.
 function ArcTan2(y : Double, x : Double) : Double {
     body intrinsic;
 }
 
 /// # Summary
 /// Returns the cosine of the specified angle.
+/// Angle is specified in radians.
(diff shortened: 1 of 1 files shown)


library/std/src/Std/Math.qs
```diff
@@ -250,79 +250,85 @@ function Min(values : Int[]) : Int {
...
 
 /// # Summary
 /// Returns the cosine of the specified angle.
+/// Angle is specified in radians.
 function Cos(theta : Double) : Double {
     body intrinsic;
 }
 
 /// # Summary
-/// Returns the hyperbolic cosine of the specified angle.
```
ScottCarda-MS, line 282: Is "angle" inaccurate? I know this PR is about specifying that we use radians, but radians are an expression of angle, so I would guess that "angle" is still appropriate.

ScottCarda-MS: Could be that I'm missing some context regarding the hyperbolic trig functions though. It's been a while since I've learned about those.

ScottCarda-MS (approved):

library/std/src/Std/Math.qs
DmitryVasilevsky: Yep. The angle is a "hyperbolic angle" in this case. So, it's not really an angle. Calling it a number is more appropriate: https://en.wikipedia.org/wiki/Hyperbolic_angle

jlehnersd: We could say "specified hyperbolic angle"

ScottCarda-MS: Oh interesting, they don't have any unit like "radians". In that case "number" makes the most sense, as you had it.

DmitryVasilevsky: > We could say "specified hyperbolic angle"

People treat this as just a function from numbers to numbers. Even sin/cos/tg are treated as functions from numbers to numbers outside of geometry. So, I don't think we should. No one recalls it as a "hyperbolic angle".

PR #2826, opened 2025-12-30
Circuit diagrams: Show source code links for grouped operations

minestarks:
Enables source code links for grouped operations, just as with simple gates.

image

### DIFF
--- a/source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
+++ b/source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
@@ -472,6 +472,7 @@
                 x2="160"
                 y1="40"
                 y2="120"
+                class="control-line"
                 style="pointer-events: none"
               />
               <circle
--- a/source/npm/qsharp/test/circuits-cases/two-qubit-gates.qs.snapshot.html
+++ b/source/npm/qsharp/test/circuits-cases/two-qubit-gates.qs.snapshot.html
@@ -245,6 +245,7 @@
                     x2="246.703125"
                     y1="100"
                     y2="160"
+                    class="control-line"
                     style="pointer-events: none"
                   />
                   <circle
@@ -590,6 +591,7 @@
                 x2="236.703125"
                 y1="100"
                 y2="160"
+                class="control-line"
                 style="pointer-events: none"
               />
               <circle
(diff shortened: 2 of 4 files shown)


source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
minestarks: Enable "hide whitespace" for this diff to save your sanity

ScottCarda-MS:
I noticed two things while playing around with this:
1. If I collapse all of a circuit to the Main operation, that Main box is not clickable despite the fact that there is a Main operation that it could navigate the user to.
2. The expand/collapse state of a group is not preserved when collapsing a surrounding group. If A calls B, there will be a group A that, when expanded will show a collapsed group B. If I then expand the group B and collapse and re-expand the surrounding group A, B does not remember that it was expanded and shows as collapsed instead.

source/compiler/qsc_circuit/src/builder/tests.rs
```diff
@@ -499,3 +499,35 @@ fn measurement_target_propagated_to_group() {
...
+        false,
+        &[0],
+        &[],
+        None,
+    );
+
+    let circuit = builder.finish(&c);
+
+    expect![[r#"
+        q_0    ─ [ [Main] ─── [ [Foo@user_code.qs:0:10] ── X@user_code.qs:0:10 ─── ] ──── ] ──
```
ScottCarda-MS, line 530: I think this also demoes the issue I was seeing with Main. Shouldn't the `[Main]` have a user_code.qs location attached to it?

minestarks: Hmmm, no, because there's no _caller_ of main in user code. Clicking on a gate navigates to its call site. I do see how your expectation would intuitively kind of make sense though. I wonder if we should special case the entry point?

ScottCarda-MS: Oh I see, yes you are right, it goes to the call of the operations. I think my intuition was based off of the go-to-def behavior from the language service in code. Ctrl-clicking a reference takes you to the definition in those cases.
I can see your approach is consistent, I think it just bucks against my intuition.

minestarks:
@ScottCarda-MS I have some expand/collapse behavior refinement in the works - stay tuned!

PR #2827, opened 2025-12-31
Circuit diagrams: Group loops

minestarks:
## Feature changes

Loops from user source are now grouped in circuit diagrams.

**BEFORE:**

image

```plaintext
q_0@test.qs:4:20 ─ X@test.qs:6:24 ─── X@test.qs:6:24 ─── X@test.qs:6:24 ──
```

**AFTER:**

image

image

```plaintext
q_0@test.qs:4:20 ─ [ [Main] ─── [ [loop: 0..2@test.qs:5:20] ── [ [(1)@test.qs:5:34] ─── [ [Foo@test.qs:6:24] ─── X@test.qs:11:20 ── Y@test.qs:12:20 ─── ] ──── ] ─── [ [(2)@test.qs:5:34] ─── [ [Foo@test.qs:6:24] ─── X@test.qs:11:20 ── Y@test.qs:12:20 ─── ] ──── ] ─── [ [(3)@test.qs:5:34] ─── [ [Foo@test.qs:6:24] ─── X@test.qs:11:20 ── Y@test.qs:12:20 ─── ] ──── ] ──── ] ──── ] ──
```

- All loops (`for`, `while`, `repeat`/`until`) are supported. The label is the expression in the loop header, which can be, depending on the loop type, an iterable, a "while" condition, or an "until" condition. e.g.:
  - `loop: 0..2`
  - `loop: i < 2`
  - `loop: i==2`

- Exception 1: If a loop only has a single iteration, we don't group.
- Exception 2: If a loop is "vertical", meaning each iteration of the loop only interacts with a distinct set of qubits, then there is no grouping, since grouping in this case tends to look more confusing than helpful.

### High level circuit diagram pipeline

To recap, this is how circuit diagrams get currently generated.

**Q#/OpenQASM** gets compiled to (via the compilation pipeline --`qsc_frontend`, `qsc_lowererer` FIR lowerer etc)...

...**FIR** gets evaluated and traced (via `qsc_eval` Evaluator)...

...**Traces with raw stacks** get captured and transformed to (via Circuit Tracer)...

...**Traces with logical stacks** get transformed to (via Circuit Tracer)...

...**`Circuit` representation** gets serialized to (via `serde-json`)...

...**Circuit JSON object** gets rendered as (via `circuit-vis`)

...**SVG & HTML circuit diagram**

### Evaluator changes

- Tracing in the Evaluator: The `Tracer` now captures the stack of `Scope`s in addition to the usual call stack (`Frame`s) in the tracing calls. The scope stack includes the stack of current lexical scopes and loops, including iteration count. This allows the circuit tracer to build a comprehensive "logical" stack that includes both call frames and lexical scopes in the source.

- The evaluator, when in DEBUG configuration, pushes LOOPs as a scope into the `Scope` stack. Before this change, we did track each Block as a scope, but not specifically loops. So now when we're in a loop, there is an extra `Scope` in the stack:

```plaintext

>  BEGIN LOOP SCOPE
>  for i in 0..4 
> 
>  BEGIN BLOCK SCOPE
>  {
>     H(qs[i]);
>  }
>  END BLOCK SCOPE
> 
>  END LOOP SCOPE

```

- The loop scope is also associated with an `iteration_count` which the evaluator is instructed to increment every time it enters the body block of the loop.

### FIR lowerer changes

- FIR: In the execution graph, debug-only nodes are consolidated into a separate `ExecGraphDebugNode` enum, just for legibility.

- FIR: during lowering, we add some extra data to the execution graph for loops: The a new `PushLoopScope` instruction to push loop scope, with an attached `ExprId` (used later by the circuit tracer to look up the loop source code and display the loop label) , and a `LoopIteration` instruction that is used to indicate a new iteration of the loop has started.

### Circuit tracer changes

- `builder.rs` / `CircuitTracer`: Introduces a `LogicalStack` which is a blended stack of call frames and lexical scopes. This struct is produced by transforming the stack traces passed down from the evaluator into a more "friendly" shape, and it corresponds to the structure that will ultimately be seen in the circuit diagram.

- `circuit.rs` / `Circuit` : The `SourceLocation::Unresolved` enum is a now-unnecessary abstraction that is removed in this change.

- in `CircuitTracer`, `finish` now takes both FIR and HIR store to resolve sources and scopes

### `circuit-vis` HTML renderer

- In the renderer, we now automatically expand any groups that only contain a single operation, which saves the user having to manually expand multiple levels of operations.

## Test coverage

- Native interpreter tests in `compiler/qsc/src/interpret/circuit_tests.rs` test direct all the way from Q# to the `Circuit` representation
- Snapshot tests in `npm/qsharp/test/circuits-cases` test all the way from Q# to rendered SVG & HTML circuit diagram using the JS/WASM component
- `compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs` - this exercises the Q# -> traces pipeline, validating that the evaluator returns expected "logical" stack traces

### DIFF
--- a/library/src/tests.rs
+++ b/library/src/tests.rs
@@ -11,6 +11,7 @@ mod intrinsic;
 mod logical;
 mod math;
 mod measurement;
+mod openqasm;
 mod state_preparation;
 mod table_lookup;
 
--- a/library/src/tests/openqasm.rs
+++ b/library/src/tests/openqasm.rs
@@ -1,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+mod angle;
--- a/library/src/tests/openqasm/angle.rs
+++ b/library/src/tests/openqasm/angle.rs
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::tests::test_expression;
+use indoc::indoc;
+use qsc::interpret::Value;
+
+#[test]
+fn check_adjust_angle_size_no_truncation_increases_size() {
+    test_expression(
+        indoc! {r#"{
+            import Std.OpenQASM.Angle.*;
+            let angle = IntAsAngle(100, 16);
+            let adjusted_angle = AdjustAngleSizeNoTruncation(angle, 32);
+            adjusted_angle.Size
+        }"#},
+        &Value::Int(32),
+    );
+}
--- a/library/std/src/Std/OpenQASM/Angle.qs
+++ b/library/std/src/Std/OpenQASM/Angle.qs
@@ -126,14 +126,14 @@ function AdjustAngleSize(angle : Angle, new_size : Int, truncate : Bool) : Angle
                 upper_bits
             }
         };
-        new Angle { Value = value, Size = size }
+        new Angle { Value = value, Size = new_size }
     } elif new_size == size {
         // Same size, no change
         angle
     } else {
         // Padding with zeros
         let value = value <<< (new_size - size);
-        new Angle { Value = value, Size = size }
+        new Angle { Value = value, Size = new_size }
     }
 }
 
--- a/source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
+++ b/source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
@@ -327,10 +327,10 @@
       <svg
         xmlns="http://www.w3.org/2000/svg"
         class="qviz"
-        width="330"
+        width="390"
         height="200"
         style="max-width: fit-content"
-        viewBox="0 0 330 200"
+        viewBox="0 0 390 200"
       >
         <g class="qubit-input-states">
           <a href="#" class="qs-circuit-source-link">
@@ -369,33 +369,33 @@
           </a>
         </g>
         <g class="wires">
-          <line x1="40" x2="330" y1="40" y2="40" class="qubit-wire" />
-          <line x1="40" x2="330" y1="120" y2="120" class="qubit-wire" />
+          <line x1="40" x2="390" y1="40" y2="40" class="qubit-wire" />
+          <line x1="40" x2="390" y1="120" y2="120" class="qubit-wire" />
           <g>
             <line
-              x1="221"
-              x2="221"
+              x1="261"
+              x2="261"
               y1="40"
               y2="79"
               class="register-classical"
             />
             <line
-              x1="219"
-              x2="219"
+              x1="259"
+              x2="259"
               y1="40"
               y2="81"
               class="register-classical"
             />
             <line
-              x1="221"
-              x2="330"
+              x1="261"
+              x2="390"
               y1="79"
               y2="79"
               class="register-classical"
             />
             <line
-              x1="219"
-              x2="330"
+              x1="259"
+              x2="390"
               y1="81"
               y2="81"
               class="register-classical"
@@ -403,29 +403,29 @@
           </g>
           <g>
             <line
-              x1="221"
-              x2="221"
+              x1="261"
+              x2="261"
               y1="120"
               y2="159"
               class="register-classical"
             />
             <line
-              x1="219"
-              x2="219"
+              x1="259"
+              x2="259"
               y1="120"
               y2="161"
               class="register-classical"
             />
             <line
-              x1="221"
-              x2="330"
+              x1="261"
+              x2="390"
               y1="159"
               y2="159"
               class="register-classical"
             />
             <line
-              x1="219"
-              x2="330"
+              x1="259"
+              x2="390"
               y1="161"
               y2="161"
               class="register-classical"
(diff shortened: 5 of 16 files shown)


source/compiler/qsc_lowerer/src/lib.rs
```diff
@@ -727,16 +738,22 @@ impl Lowerer {
...
+                self.exec_graph
+                    .debug_push(ExecGraphDebugNode::PushLoopScope(id));
+
                 let cond_idx = self.exec_graph.len();
                 let cond = self.lower_expr(cond);
                 let idx = self.exec_graph.len();
                 // Put a placeholder in the execution graph for the jump past the loop
                 self.exec_graph.push(ExecGraphNode::Jump(0));
-                let body = self.lower_block(body);
+                let body = self.lower_block(body, true);
```
swernli: since this is the only place that `is_loop_body` is passed as true, and that only causes the extra node to be pushed, it might make sense to put the node pushing here and then leave `lower_block` signature unchanged.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -461,6 +461,19 @@ impl Env {
         self.scopes.push(scope);
     }
 
+    pub fn push_loop_scope(&mut self, frame_id: usize, loop_expr: ExprId) {
+        let iteration_count = 0;
+        let scope = Scope {
+            frame_id,
+            loop_scope: Some(LoopScope {
+                loop_expr,
+                iteration_count,
```
swernli: nitpick: this is very "code golf" of me, but the extra local to define iteration_count feels unnecessary when this line could just be `interation_count: 0,` instead.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    fn qubit_release(&mut self, stack: &qsc_eval::StackTrace, q: usize) {
+        self.write_stack(stack);
+        let _ = writeln!(self.trace, "qubit_release(q_{q})");
+    }
+
+    fn qubit_swap_id(&mut self, stack: &qsc_eval::StackTrace, q0: usize, q1: usize) {
+        self.write_stack(stack);
+        let _ = writeln!(self.trace, "qubit_swap_id(q_{q0} q_{q1})");
```
ScottCarda-MS: Super nit, but to be consistent with other trace strings, I think you should have a comma separating the two args, right?

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+        operation Main() : Unit {
+            use q = Qubit[3];
+        }
+        "},
+        "A.Main()",
+        ExecGraphConfig::Debug,
+        &expect![[r#"
+            Main@A.qs:1:4 -> AllocateQubitArray@qsharp-library-source:core/qir.qs:17:8 -> loop: 0..size - 1@qsharp-library-source:core/qir.qs:17:29[1] -> (1)@qsharp-library-source:core/qir.qs:18:23 -> qubit_allocate(q_0)
+            Main@A.qs:1:4 -> AllocateQubitArray@qsharp-library-source:core/qir.qs:17:8 -> loop: 0..size - 1@qsharp-library-source:core/qir.qs:17:29[2] -> (2)@qsharp-library-source:core/qir.qs:18:23 -> qubit_allocate(q_1)
+            Main@A.qs:1:4 -> AllocateQubitArray@qsharp-library-source:core/qir.qs:17:8 -> loop: 0..size - 1@qsharp-library-source:core/qir.qs:17:29[3] -> (3)@qsharp-library-source:core/qir.qs:18:23 -> qubit_allocate(q_2)
```
ScottCarda-MS, line 247: I was confused for quite a while looking at this because I thought it was `loop: 0..size` -> `1@...` -> `(1)@...`. I see now that it is actually `loop: 0..size - 1@...` -> `(1)@...`. The spacing in the loop expression confused me.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+            Main@A.qs:2:4 -> loop: 0..2@A.qs:2:18[2] -> (2)@A.qs:3:8 -> Foo@A.qs:9:4 -> H@qsharp-library-source:Std/Intrinsic.qs:205:8 -> gate(H, targets=(q_0), controls=())
+            Main@A.qs:2:4 -> loop: 0..2@A.qs:2:18[3] -> (3)@A.qs:3:8 -> Foo@A.qs:9:4 -> H@qsharp-library-source:Std/Intrinsic.qs:205:8 -> gate(H, targets=(q_0), controls=())
+            Main@A.qs:5:4 -> MResetZ@qsharp-library-source:Std/Measurement.qs:135:4 -> measure(MResetZ, q_0, Id(0))
+            Main@A.qs:1:4 -> qubit_release(q_0)
+        "#]],
+    );
+}
+
+#[test]
+fn repeat_until_loop() {
```
ScottCarda-MS: This test looks like a copy-paste from the previous, and doesn't have a repeat-until loop in it. Maybe a copy error?

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+            Test@A.qs:1:5 -> M@qsharp-library-source:Std/Intrinsic.qs:268:4 -> Measure@qsharp-library-source:Std/Intrinsic.qs:304:12 -> measure(M, q_0, Id(0))
+            Test@A.qs:1:12 -> M@qsharp-library-source:Std/Intrinsic.qs:268:4 -> Measure@qsharp-library-source:Std/Intrinsic.qs:304:12 -> measure(M, q_1, Id(1))
+            ReleaseQubitArray@qsharp-library-source:core/qir.qs:24:8 -> loop: qs@qsharp-library-source:core/qir.qs:24:20[1] -> (1)@qsharp-library-source:core/qir.qs:25:12 -> qubit_release(q_0)
+            ReleaseQubitArray@qsharp-library-source:core/qir.qs:24:8 -> loop: qs@qsharp-library-source:core/qir.qs:24:20[2] -> (2)@qsharp-library-source:core/qir.qs:25:12 -> qubit_release(q_1)
+        "#]],
+    );
+}
+
+#[test]
+fn adjoint_operation_in_entry_expr() {
```
ScottCarda-MS, line 686: I'd also like to see a test with an auto-generated adjoint call, it if isn't too much trouble. All the calls in the auto-adjoint should be adjointed themselves and be in reverse order. This also brings up the need to test loops in auto-adjoint to ensure that they are processed backwards when adjointed. There could also be a test using a within-apply.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+            r
+        }"},
+        ExecGraphConfig::Debug,
+        &expect![[r#"
+            AllocateQubitArray@qsharp-library-source:core/qir.qs:17:8 -> loop: 0..size - 1@qsharp-library-source:core/qir.qs:17:29[1] -> (1)@qsharp-library-source:core/qir.qs:18:23 -> qubit_allocate(q_0)
+            <lambda>@<entry>:2:10 -> H@qsharp-library-source:Std/Intrinsic.qs:205:8 -> gate(H, targets=(q_0), controls=())
+            ReleaseQubitArray@qsharp-library-source:core/qir.qs:24:8 -> loop: qs@qsharp-library-source:core/qir.qs:24:20[1] -> (1)@qsharp-library-source:core/qir.qs:25:12 -> qubit_release(q_0)
+        "#]],
+    );
+}
```
ScottCarda-MS, line 739: Would be good to have a test or two with gates/operations using control qubits, to ensure that those are handled correctly.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
```diff
@@ -0,0 +1,611 @@
...
+            Main@A.qs:2:4 -> loop: 0..2@A.qs:2:18[2] -> (2)@A.qs:3:8 -> Foo@A.qs:9:4 -> H@qsharp-library-source:Std/Intrinsic.qs:205:8 -> gate(H, targets=(q_0), controls=())
+            Main@A.qs:2:4 -> loop: 0..2@A.qs:2:18[3] -> (3)@A.qs:3:8 -> Foo@A.qs:9:4 -> H@qsharp-library-source:Std/Intrinsic.qs:205:8 -> gate(H, targets=(q_0), controls=())
+            Main@A.qs:5:4 -> MResetZ@qsharp-library-source:Std/Measurement.qs:135:4 -> measure(MResetZ, q_0, Id(0))
+            Main@A.qs:1:4 -> qubit_release(q_0)
+        "#]],
+    );
+}
+
+#[test]
+fn while_loop() {
```
ScottCarda-MS: Could you add a test or two checking non-standard while loops? Some while-loops that don't increment their iterator as the last statement, increment their iterator by more than 1 at a time (or better yet a different amount depending on the loop iteration), and do different gates depending on their iteration.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -583,6 +603,37 @@ pub struct State {
     exec_graph_config: ExecGraphConfig,
 }
 
+#[derive(Default)]
+pub struct StackTrace {
+    call_stack: Vec<Frame>,
+    scope_stack: Vec<Scope>,
```
minestarks: @swernli suggested a way to eliminate `scope_stack` from this struct, and just use `Vec<Frame>` as before. Will try it out.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
minestarks: Sure, though in practice all we're counting is how many times we enter and exit the body block as we're evaluating, so the properties you called out shouldn't interfere with tracing. I guess a unit test would make that clearer, so I'll add one.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
minestarks: Added, thanks. Turns out we don't label controlled specializations properly (e.g. calls to  `Controlled Foo` and `Foo` are both labeled `Foo` ), which may not be ideal? I'll address that separately.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
minestarks: Aren't we just testing the Q# evaluator at this point? There are a multitude of interesting adjoint/uncomputation behavior, but this seems outside the scope of circuit diagrams / stack tracing, and unaffected by any changes in this PR. Surely these are already covered by existing Q# tests?

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
minestarks: Oops. Thanks for catching. Looks like a dupe of `for_loop_debug`.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
minestarks: :cry: all I can say is, I tried so many different combinations of spacing / newlines / parentheticals until I arrived at this. If I add too many parantheses, the lines become too long and _less_ readable. If I use newlines instead of `->`, the trace looks very verbose and the "shape" of the pushes/pops becomes hard to follow. At some point, spending so much time tweaking test output felt excessive, and I'm out of ideas :/

source/compiler/qsc_lowerer/src/lib.rs
minestarks: Ha, that's so obviously the right thing to do, thank you. I hated this `bool` but just couldn't figure out how to get rid of it 😆

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
ScottCarda-MS: I thought that was what this test was trying to test for. Maybe this was meant to just be a test that the dagger symbol gets printed with adjoint calls?

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
ScottCarda-MS: Fait enough. I think the `@` sign cannot appear in the printed expression, so it should make for a definitive delimiter. If I go into reading these looking for the ampersat instead of the arrows, it helps.

source/compiler/qsc_circuit/src/builder/tests/logical_stack_trace.rs
ScottCarda-MS: Thanks for adding some more tests for this.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -549,10 +563,11 @@ impl Env {
...
 
-#[derive(Default)]
-struct Scope {
+#[derive(Default, Clone)]
+pub struct Scope {
     bindings: IndexMap<LocalVarId, Variable>,
-    frame_id: usize,
+    pub frame_id: usize,
+    pub is_loop: bool,
 }
```
swernli: As it turns out, the new visibility and `Clone` derive are no longer needed.

```suggestion
#[derive(Default)]
struct Scope {
    bindings: IndexMap<LocalVarId, Variable>,
    frame_id: usize,
    is_loop: bool,
}
```

source/compiler/qsc_eval/src/lib.rs
minestarks: pushed his commit.

PR #2831, opened 2026-01-07
Panic in QIR generation: "only some primitive types are supported"

### LINKED ISSUE #2830: Panic in QIR generation: "only some primitive types are supported"
QIR generation for this code

```qsharp
operation Main() : Unit {
    use qs = Qubit[5];
    CustomIntrinsic(qs);
}

operation CustomIntrinsic(qs : Qubit[]) : Unit {
    body intrinsic;
}

```

panics here:

https://github.com/microsoft/qdk/blob/20c2fb1c118e553b69d042c5289d61dda9c55e76/source/compiler/qsc_partial_eval/src/lib.rs#L3671

[Playground link](https://microsoft.github.io/qdk/?code=H4sIAAAAAAAACssvSC1KLMnMz1PwTczM09BUsFIIzcssUajmUlBQUCgtTlUoLFawVQgsTcosiTaNtQYLO5cWl%252BTneuaVFGXmFWcmaxQWa1pz1XJx5cMNw1ShYAU1JBbNjqT8lEqFTJhKkDkASeoQq5QAAAA%253D)

Saw repro in 20c2fb1c118e553b69d042c5289d61dda9c55e76

swernli:
Fixes #2830

### DIFF
--- a/source/compiler/qsc_partial_eval/src/lib.rs
+++ b/source/compiler/qsc_partial_eval/src/lib.rs
@@ -98,9 +98,12 @@ pub enum Error {
     #[diagnostic(help("try invoking the desired callable directly"))]
     UnexpectedDynamicValue(#[label] PackageSpan),
 
-    #[error("cannot use a dynamic value of type `{0}` used in intrinsic callable signature")]
-    #[diagnostic(code("Qsc.PartialEval.UnexpectedDynamicIntrinsicType"))]
-    UnexpectedDynamicIntrinsicType(String, #[label] PackageSpan),
+    #[error("unsupported type `{0}` in custom intrinsic callable")]
+    #[diagnostic(help(
+        "variables of type `{0}` cannot be emitted into QIR and should not appear in custom intrinsic callable signatures"
+    ))]
+    #[diagnostic(code("Qsc.PartialEval.UnsupportedType"))]
+    UnsupportedCustomIntrinsicType(String, #[label] PackageSpan),
 
     #[error("partial evaluation failed with error: {0}")]
     #[diagnostic(code("Qsc.PartialEval.EvaluationFailed"))]
@@ -148,7 +151,7 @@ impl Error {
         match self {
             Self::CapabilityError(_) => None,
             Self::UnexpectedDynamicValue(span)
-            | Self::UnexpectedDynamicIntrinsicType(_, span)
+            | Self::UnsupportedCustomIntrinsicType(_, span)
             | Self::EvaluationFailed(_, span)
             | Self::OutputResultLiteral(span)
             | Self::Unexpected(_, span)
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -98,9 +98,9 @@ pub enum Error {
     #[diagnostic(help("try invoking the desired callable directly"))]
     UnexpectedDynamicValue(#[label] PackageSpan),
 
-    #[error("cannot use a dynamic value of type `{0}` returned from intrinsic callable")]
-    #[diagnostic(code("Qsc.PartialEval.UnexpectedDynamicIntrsinicReturnType"))]
-    UnexpectedDynamicIntrinsicReturnType(String, #[label] PackageSpan),
+    #[error("cannot use a dynamic value of type `{0}` used in intrinsic callable signature")]
+    #[diagnostic(code("Qsc.PartialEval.UnexpectedDynamicIntrsinicType"))]
```
minestarks:
```suggestion
    #[diagnostic(code("Qsc.PartialEval.UnexpectedDynamicIntrinsicType"))]
```

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -98,9 +98,9 @@ pub enum Error {
     #[diagnostic(help("try invoking the desired callable directly"))]
     UnexpectedDynamicValue(#[label] PackageSpan),
 
-    #[error("cannot use a dynamic value of type `{0}` returned from intrinsic callable")]
-    #[diagnostic(code("Qsc.PartialEval.UnexpectedDynamicIntrsinicReturnType"))]
-    UnexpectedDynamicIntrinsicReturnType(String, #[label] PackageSpan),
+    #[error("cannot use a dynamic value of type `{0}` used in intrinsic callable signature")]
```
minestarks: This error message isn't quite making sense to me, at least for the repro I saw. There are no dynamic values in the repro. Also "value" is a runtime concept while "callable signature" is a type, so the phrasing is a bit confusing.

Maybe more to the point, shouldn't custom intrinsics with an invalid signature just be a compiler error, then? We can't run them. We can't estimate them. We can't generate QIR for them. What are they good for?

swernli: I debated on this, and I'm not sure. We have intrinsic callables with arrays in the signature as part of the stdlib, we just also have runtime/codegen support for them. To make this a compile time check, we'd have to not just check the signature of intrinsic callables but also check the names against a whitelist that we keep up to date with the runtime support. That's certainly doable, just felt weird to me. If that's an acceptable implementation quirk, then this can indeed become a compile time check.

In fact, this is a compile time check that becomes easier with some of the updates I'd like to make to RCA...

swernli:
Rebased on main to remove the update of package-lock.json

source/compiler/qsc_partial_eval/src/lib.rs
swernli: After some discussion, we settled for now on an improved error message for clarity. We can circle back to what it would take to pull this into a design-time check later.

minestarks (approved):

PR #2834, opened 2026-01-07
First version of table lookup library

DmitryVasilevsky:
The `table_lookup` library defines various primitives useful to perform computation and uncomputation of table lookup. It also defines wrapper function which uses one of the approaches depending on options.
More in readme.md.

### DIFF
--- a/library/table_lookup/README.md
+++ b/library/table_lookup/README.md
@@ -19,7 +19,7 @@ which uses one of the approaches depending on options.
 
 * `DoStdUnlookup` - Use unlookup algorithm defined in the Q# standard library.
 * `DoUnlookupViaLookup` - Perform unlookup via the same algorithm as lookup as it is self-adjoint.
-* `DoMCXUnlookup` - Perform measurement-based unlookup with corrections via multicontrolled X gates. See [arXiv:2211.01133](https://arxiv.org/abs/2211.01133), Section 2
+* `DoMCXUnlookup` - Perform measurement-based unlookup with corrections via multicontrolled X gates. See [arXiv:2211.01133](https://arxiv.org/abs/2211.01133), Section 2.
 * `DoPPUnlookup` - Perform measurement-based unlookup with corrections via power products without address split (Phase lookup). See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917), Section A.3.
 * `DoSplitPPUnlookup` - Perform measurement-based unlookup with corrections via power products with address split (Phase lookup). See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917), Section A.3.
 


library/table_lookup/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Options for lookup:
+
+* `DoStdLookup` - Use lookup algorithm defined in the Q# standard library.
+* `DoMCXLookup` - Use naive lookup algorithm via multicontrolled X gates.
+* `DoRecursiveSelectLookup` - Use select network algorithm via recursion. See [arXiv:2211.01133](https://arxiv.org/abs/2211.01133)
+* `DoPPLookup` - Use lookup algorithm via power products without address split. See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917)
+* `DoSplitPPLookup` - Use lookup algorithm via power products with address split. See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917)
+
+### Options for unlookup:
```
swernli: Markdown guidance is to avoid trailing punctuation in headings.

```suggestion
### Options for lookup

* `DoStdLookup` - Use lookup algorithm defined in the Q# standard library.
* `DoMCXLookup` - Use naive lookup algorithm via multicontrolled X gates.
* `DoRecursiveSelectLookup` - Use select network algorithm via recursion. See [arXiv:2211.01133](https://arxiv.org/abs/2211.01133)
* `DoPPLookup` - Use lookup algorithm via power products without address split. See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917)
* `DoSplitPPLookup` - Use lookup algorithm via power products with address split. See [arXiv:2505.15917](https://arxiv.org/abs/2505.15917)

### Options for unlookup
```

library/table_lookup/qsharp.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "MIT",
+  "dependencies": {
+    "Qtest": {
+      "github": {
+        "owner": "Microsoft",
+        "repo": "qsharp",
+        "ref": "v1.21.0",
+        "path": "library/qtest"
+      }
+    }
```
swernli: I don't think you ended up using this, so it can be removed.

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+
+    mutable result = qs;
+    // For each bit position (from least to most significant)
+    for i in 0..n-1 {
+        let step = 2^i;
+        // For each pair of positions that differ only in that bit
+        for j in 0..(step * 2)..len-1 {
+            for k in 0..step-1 {
+                // XOR the "upper" position with the "lower" position
+                result[j + k + step] = Std.Logical.Xor(result[j + k + step], result[j + k]);
```
swernli: This is a nitpick/personal preference, so feel free to ignore, but I tend to lean toward importing the function that is used (via `import Std.Logical.Xor;`) so this spot can just use `Xor` or removing the library use here entirely since logical xor on Boolean is just `!=`.

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+    return result;
+}
+
+/// # Summary
+/// Measures all qubits in the given register in the X basis. And resets them to |0>.
+operation MeasureAndComputePhaseData(target : Qubit[], data : Bool[][], size : Int) : Bool[] {
+    // Measure target register in X basis
+    let measurements = ResultArrayAsBoolArray(ForEach(MResetX, target));
```
swernli: Have you confirmed this works well with QIR codegen? I recall having trouble in the past with using `ForEach` with a measurement function. If it doesn't, then a combination of `ApplyToEach(H, target)` and `MResetEachZ` should work.

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+    // Get phasing data via parity checks
+    let phaseData = Mapped(BinaryInnerProduct(measurements, _), data);
```
swernli: Similar concern here regarding QIR codegen and `Mapped`; if it does work that is good news, but I'm worried that it may cause issues with partial evaluation.

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+/// # Summary
+/// Computes dot (inner) product of two vectors over GF(2) field.
+/// This isn't a proper inner product as it is not positive-definite.
+///
+/// It is used to see if a phase correction is needed for a bit string `data`
+/// after obtaining a measurement result `measurements`.
+function BinaryInnerProduct(data : Bool[], measurements : Bool[]) : Bool {
+    mutable sum = false;
+    for i in IndexRange(measurements) {
+        set sum = Std.Logical.Xor(sum, (data[i] and measurements[i]));
```
swernli: Same nitpick here regarding `Xor`, which could be replaced with a comment and `!=`

library/table_lookup/src/PhaseLookup.qs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+operation ApplyPhasingViaZ(qs : Qubit[], mask : Bool[]) : Unit {
+    // Ignore the first element of mask, it affects the global phase.
+    ApplyPauliFromBitString(PauliZ, true, Std.Arrays.Rest(mask), qs);
+
+    // Add global phase if requested.
+    // if (mask[0]) {
+    //     Exp([], PI(), []);
+    // }
```
swernli: Is this still needed or can it be removed?

library/table_lookup/src/PowerProducts.qs
```diff
@@ -0,0 +1,181 @@
...
+    for qubit_index in 0..Length(qubits)-1 {
+        // First, add the set that consists of only one qubit at index qubit_index.
+        power_products += qubits[qubit_index..qubit_index];
+        // Then, construct and add sets that include this new qubit as the last one.
+        for existing_set_index in 0..Length(power_products)-2 {
+            // Take the next qubit for the new set
+            let next_power_product = aux_qubits[next_available];
+            next_available += 1;
+            // Create appropriate set and add it to the result
+            AND(power_products[existing_set_index], qubits[qubit_index], next_power_product);
```
swernli, line 41: For this and other places where `AND` is used in a spot that is not adjointable, is there any benefit over just calling `CCNOT` directly?

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+///
+/// # Output
+/// Boolean array of the same length as input containing monomial coefficients.
+///
+/// # Remarks
+/// This function is the classical preprocessing step for quantum phase lookup operations,
+/// converting phase data from standard basis coefficients to power product coefficients.
+/// The transformation is its own inverse when applied twice.
+function FastMobiusTransform(qs : Bool[]) : Bool[] {
+    let len = Length(qs);
```
swernli: It feels a little weird that this boolean array parameter is named `qs` since that is often what we call qubit arrays.

library/table_lookup/src/Utils.qs
```diff
@@ -0,0 +1,242 @@
...
+
+/// # Summary
+/// Computes dot (inner) product of two vectors over GF(2) field.
+/// This isn't a proper inner product as it is not positive-definite.
+///
+/// It is used to see if a phase correction is needed for a bit string `data`
+/// after obtaining a measurement result `measurements`.
+function BinaryInnerProduct(data : Bool[], measurements : Bool[]) : Bool {
+    mutable sum = false;
+    for i in IndexRange(measurements) {
```
swernli, line 158: nitpick: in most of the rest of the library you use the pattern `0..Length(arr)-1` but here you use the `IndexRange` helper, which is equivalent. It doesn't matter functionally, just wondering if there is value in keeping to the same pattern throughout.

library/table_lookup/src/Tests.qs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import Std.Diagnostics.*;
+
+import Main.*;
+
+internal operation MatchLookupToStd(
```
swernli, line 8: nitpick: since operations are internal by default until exported, this keyword doesn't really do anything anymore (it doesn't actually prevent it from later being exported).

library/table_lookup/src/LookupViaPP.qs
```diff
@@ -0,0 +1,311 @@
...
+    let products2 = ConstructPowerProducts(h2, aux_qubits2);
+
+    ApplyFlips(data, products1, products2, target);
+
+    // Undo power products of both halves
+    DestructPowerProducts(products1);
+    DestructPowerProducts(products2);
+
+    Fact(Std.Diagnostics.CheckAllZero(aux_qubits1), "Auxiliary1 qubits should be reset to zero after SelectViaPowerProducts");
+    Fact(Std.Diagnostics.CheckAllZero(aux_qubits2), "Auxiliary2 qubits should be reset to zero after SelectViaPowerProducts");
```
swernli: use of `CheckAllZero` here will prevent this operation from generating to QIR.

swernli (commented):
Did an initial review with some feedback.

Also, is this library one that you would want to add to the registry in source/vscode/src/registry.json?

library/table_lookup/qsharp.json
DmitryVasilevsky: Yes! I ended up not using it. Removed.

library/table_lookup/src/Utils.qs
DmitryVasilevsky: These functions were moving from file to file as I was developing this library, so it was easier to just fully reference Xor inline. Now that they ended up in the separate Utils file, it does make sense to remove full inline reference.

library/table_lookup/src/Utils.qs
DmitryVasilevsky: Used import.

library/table_lookup/src/PhaseLookup.qs
DmitryVasilevsky: Used this when I was still experimenting and verifying exact correctness. This function doesn't preserve global phase without the commented-out part. We don't really preserve global phase so this is not needed. Removed.

library/table_lookup/src/PowerProducts.qs
DmitryVasilevsky: I actually prefer using AND instead of CCNOT if I know that the target qubit is in zero state. That's the semantics of AND. I think this provides additional information to the system (although it may not be used at this time). I'd prefer to keep AND.

library/table_lookup/src/Utils.qs
DmitryVasilevsky: Thanks, renamed!

library/table_lookup/src/PowerProducts.qs
swernli: That's fair. Just worth remembering that `AND` is implemented differently for Base Profile, so your programs will look different depending on the profile chosen:
image

library/table_lookup/src/Utils.qs
DmitryVasilevsky: I prefer IndexRange for cases where the main goal of the loop is to iterate over all elements of one array and index is also needed for some other computation. I looked through the code and used IndexRange in these cases.
Sometimes I still need to use the explicit range if index means something special or iteration isn't over the entire range.

library/table_lookup/src/LookupViaPP.qs
DmitryVasilevsky: The way this operation ended up I don't need these checks anymore. Now aux qubits are allocated here and the check happens automatically when they go out of scope. Removed.

library/table_lookup/src/Tests.qs
DmitryVasilevsky: Yes... I still like this internal tag somehow. I can see easily that this isn't intended for external use. I'd prefer to keep it.

library/table_lookup/src/PowerProducts.qs
DmitryVasilevsky: These consequences are both intended and unintended! :) It's interesting that AND is decomposed this way for base profile while CCNOT is still available. However, this is the goal I want to achieve... I guess. If this is the best decomposition of AND for the base profile, then so be it.
This also intersects a bit with 'prefer measurement based uncomputation' option, but not really. That option is also about AND vs. CCNOT mostly, but it affects the uncomputation part.

library/table_lookup/src/Lookup.qs
```diff
@@ -0,0 +1,292 @@
...
+                        input.fitData,
+                        input.fitAddress,
+                        target
+                    );
+                }
+            } else {
+                // To use control qubit as an extra address qubit we need to respect entire address.
+                // Power products implementation already does that.
+                within {
+                    // Invert control so that data is selected when control is |1>
```
msoeken:
```suggestion
                    // Invert control so that data is selected when control is |1>.
```

Consistency in comments.

msoeken: There are other occurrences.

library/table_lookup/src/PhaseLookup.qs
```diff
@@ -0,0 +1,193 @@
...
+    for i in 0..data_value_count-1 {
+        let data = Std.Convert.IntAsBoolArray(i, data_length);
+        let same = CheckOperationsAreEqual(
+            address_size,
+            PhaseLookupViaSplitPP(_, data),
+            Multicontrolled.PhaseLookupViaMCX(data, _)
+        );
+        Fact(same, $"PhaseLookupViaPPandCZ must be the same as PhaseLookupViaMCX for {data}.");
+    }
+}
```
msoeken:
```suggestion
}

```

library/table_lookup/src/PhaseLookup.qs
```diff
@@ -0,0 +1,193 @@
...
+///
+/// # Input
+/// ## qs
+/// Qubit register whose basis states will have their phases inverted.
+///
+/// ## data
+/// Boolean array indicating which basis states to invert. If `data[i]` is `true`,
+/// the phase of |i⟩ gets inverted (multiplied by -1).
+///
+/// # Reference
```
msoeken, line 59: In addition to the docstring references it would be great to also note the location in the paper where some of the operations are defined (e.g., figure or section number)

msoeken (approved):

library/table_lookup/src/Utils.qs
DmitryVasilevsky: Updated code to use explicit loops so the QIR can be generated. Thanks! I think it's important here because we want this for experiments with different environments and our interoperability is via QIR.

library/table_lookup/src/Utils.qs
DmitryVasilevsky: Updated.

library/table_lookup/src/PhaseLookup.qs
DmitryVasilevsky: Updated as part of another change.

library/table_lookup/src/Lookup.qs
DmitryVasilevsky: Went through comments and made them more consistent.

library/table_lookup/src/PhaseLookup.qs
DmitryVasilevsky: Added section numbers to references.

swernli (approved):

PR #2835, opened 2026-01-07
Refactor RCA, part 1

swernli:
This is the first part of a refactoring for Runtime Capabailities Analysis (RCA) internals. This is meant to simplify the internals by removing some elements that are no longer used, simplifying some parts, and preparing for later renaming and feature work. In particular, this includes:

- Removal of the `ValueKind` struct that differentiated between elements and arrays, rename the exisiting `RuntimeKind` to `ValueKind`
- Use `RuntimeFeatureFlags::UseOfDynamicallySizedArray` more consistently to identify places where arrays are dynamic.
- Propagate `RuntimeFeatureFlags` via increased use of `ComputeKind` struct.
- Simplify array related `ApplicationInstance` entries to drop distinction between dynamic and static content.
- Update RCA tests for changes in structure.

Note that for the test updates, the actual content of the field (static vs dynamic) and importantly the `runtime_features` do NOT change, showing that the refactor did not affect the behavior of these tests. The one exception is a new test case in arrays that shows an indirect call to the `Length` intrinsic with a dynamically sized array now correctly reflects the output as a dynamic integer, which was a pre-existing subtle propagation bug fixed during refactor.

### DIFF
--- a/source/compiler/qsc_rca/src/lib.rs
+++ b/source/compiler/qsc_rca/src/lib.rs
@@ -597,17 +597,6 @@ impl ValueKind {
             Self::Dynamic
         }
     }
-
-    pub(crate) fn project_onto_variant(self, variant: &mut ValueKind) {
-        match self {
-            ValueKind::Static => {
-                // No changes needed.
-            }
-            ValueKind::Dynamic => {
-                *variant = ValueKind::Dynamic;
-            }
-        }
-    }
 }
 
 bitflags! {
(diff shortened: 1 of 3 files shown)


source/compiler/qsc_rca/src/lib.rs
```diff
@@ -474,24 +443,8 @@ impl Display for ComputeKind {
...
-                    runtime_features,
-                    value_kind,
-                })
-            }
-        }
-    }
-
-    pub(crate) fn new_with_runtime_features(
+    #[must_use]
+    pub fn new_with_runtime_features(
```
minestarks, line 447: Ahhh, this name had me confused for a good while, because I did not see that it initialize the kind to `Self::Quantum` . I'd simply call it `quantum()` so that it would look like this at the call site:

`ComputeKind::quantum(runtime_features, value_kind)`

makes me go "yep, we're initialing a quantum computekind"

minestarks: Or I mean... maybe we don't even need a constructor for such a trivial initialization?

swernli: I think a rename here makes sense, or possibly even removal. It makes even more sense in part 2, when this enum is renamed to `Static` and `Dynamic` and the intervening features are deleted. I'll make this change there where the names and ergonomics are even nicer!

PR #2845, opened 2026-01-13
Fixed angle adjustment size bug

filipw:
There is a subtle bug in `AdjustAngleSize`.

When creating the new `Angle` struct in both the truncation and padding branches, the codeuses the original size instead of `new_size`. This means the `Value` is scaled to the new precision, but the `Size` metadata remains incorrect.

This then cascades and impacts other functions that make use of `Size`, e.g. `AngleAsResultArrayBE`.

I couldn't find any tests for that logic so I added a test with a new mod. The test is also a repro and fails without the change.


idavis (approved):

idavis:
@filipw - there is a clippy issue with the use of mod.rs, but it looks good other than that

PR #2848, opened 2026-01-13
Circuit diagrams: Enable group_by_scope by default

minestarks:
VS Code:

- Remove the "experimental" tag
- Remove the prerelease check

Python:
- Default to `group_by_scope=True` for Q# and OpenQASM versions of `.circuit`

- Remove the native `TracerConfig::Default()` since it was hiding a lot of scenario-specific defaults we may want to set explicitly. 
- Set those other scenario-specific defaults

### DIFF
--- a/source/compiler/qsc/src/interpret.rs
+++ b/source/compiler/qsc/src/interpret.rs
@@ -1477,7 +1477,7 @@ impl Debugger {
         TracerConfig {
             max_operations: TracerConfig::DEFAULT_MAX_OPERATIONS,
             source_locations: true,
-            group_by_scope: true,
+            group_by_scope: false,
             prune_classical_qubits: false,
         }
     }


source/compiler/qsc/src/interpret.rs
```diff
@@ -1467,6 +1471,16 @@ impl Debugger {
             .get(map_fir_package_to_hir(self.interpreter.source_package))
             .expect("Could not load package")
     }
+
+    /// Configuration used to trace the circuit while debugging.
+    fn circuit_config() -> TracerConfig {
+        TracerConfig {
+            max_operations: TracerConfig::DEFAULT_MAX_OPERATIONS,
+            source_locations: true,
+            group_by_scope: true,
```
swernli: I thought the circuit that gets accumulated during debuging was expected to not group by default?

minestarks: Yeah, maybe it's best to make this `false` given the current behavior is not perfect. Ideally, we _would_ show groups, but they'd be expanded. But we can't do that yet.

swernli (approved):

PR #2850, opened 2026-01-15
Sccarda/histogram style

ScottCarda-MS:
To avoid difficult to read graphics, this PR removes the variability in the histogram colors as VS Code theme changes, and replaces it with two themes, a light theme histogram for light VS Code themes and a dark theme histogram for dark VS Code themes. Individual components should no longer change unless changing the VS Code theme between these two categories.

Light Themes:
Histogram Light Theme

Dark Themes:
Histogram Dark Theme

Also improves the ket label font to be more legible.

Before:
Worse Ket Font

After:
Better Ket Font


source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -48,6 +48,69 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
...
+  --qs-histogram-menu-box-stroke: #3b3b3b;
+  --qs-histogram-menu-item-fill: #b7d5ff;
+  --qs-histogram-menu-item-hover-fill: #9fc7ff;
+  --qs-histogram-menu-item-stroke: #6b6b6b;
+  --qs-histogram-menu-item-selected-fill: #6aa9ff;
+  --qs-histogram-menu-text-fill: #202020;
+  --qs-histogram-menu-separator-stroke: #6b6b6b;
+  --qs-histogram-help-fill: #ffffff;
+  --qs-histogram-help-stroke: #6b6b6b;
+  --qs-histogram-help-text-fill: #202020;
```
billti, line 74: Where did these color codes come from? Did you take them from any existing UX? We already are a little inconsistent across widgets, so I'd rather try and reuse colors from existing VS Code or QDK UX than add yet more divergence to our palette if possible.

source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -48,6 +48,69 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
...
+
+Important: VS Code adds .vscode-high-contrast for backwards compatibility even
+when the theme is actually vscode-high-contrast-light, so we must explicitly
+exclude that case.
+*/
+body.vscode-dark,
+body.vscode-high-contrast:not(.vscode-high-contrast-light),
+body[data-vscode-theme-kind="vscode-dark"],
+body[data-vscode-theme-kind="vscode-high-contrast"],
+body[vscode-color-theme="dark"] {
```
billti: How does this work for Jupyter in the browser where we don't have the VS Code specific CSS selectors set? I think the histogram may have some issues already there, which would be great to fix at the same time rather than revisit this yet again for those issues.

Jupyter has its own selectors. You can see for example on lines 33, 37, 41... above where the `--jp-` selectors are used as fallbacks if the VS Code variables aren't set. This is a way to match the theme in either environment.

source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -48,6 +48,69 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
...
+ - Attribute: data-vscode-theme-kind="vscode-..." (preferred)
+
+Important: VS Code adds .vscode-high-contrast for backwards compatibility even
+when the theme is actually vscode-high-contrast-light, so we must explicitly
+exclude that case.
+*/
+body.vscode-dark,
+body.vscode-high-contrast:not(.vscode-high-contrast-light),
+body[data-vscode-theme-kind="vscode-dark"],
+body[data-vscode-theme-kind="vscode-high-contrast"],
```
billti: For VS Code specifically, I think you only need lines 89 and 90 here (the `body[data-vscode-theme-kind=...` selectors). What are the other 3 selectors needed for that these 2 don't cover?

source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -345,102 +408,112 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
...
   font-size: 3pt;
-  fill: var(--vscode-icon-foreground, var(--main-color));
+  fill: var(--qs-histogram-bar-label-fill);
   text-anchor: end;
   pointer-events: none;
 }
 
 .bar-label-ket {
-  font-family: "Courier", monospace;
+  font-family: var(
```
billti, line 425: Where did this list come from? How does Jupyter in different browsers look with it? (The first one `--vscode-editor-font-family` obviously won't resolve outside of VS Code, and `ui-monospace` is not implemented in most browsers - see https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-family ).

billti: Also, as per above MDN reference, `Consolas`, `Menlo`, and `Monaco` aren't "generic font families", and thus need to be quoted. (I've been bitten by this before).

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: I was trying to cover all the bases in case some of these are supported in some environments. The two environments I checked against were Chrome and Edge with an up-to-date VS Code, as well as local VS Code in windows, but it is difficult to test against all the possible environments. I'm happy to lean on your experience in this, though, and streamline this to just use the data-vscode-theme-kind selectors if you think that is sufficient.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: This is the list that copilot suggests to cover a wide range of environments, more environments then I could test on my machine. The environments I've tried are Chrome and Edge and local VS Code app, with VS Code Q# Extension and Jupyter notebook. I've removed from the list some of the less common fonts that you pointed out.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: Getting a CSS selector to detect if jupyter was in dark mode or not was a bit more complex than initially thought. The widget gets registered in a shadow DOM so it required special logic in index.tsx to get the right DOM, check that DOM for the correct class on body, and set whether the theme is dark/light in the host with a special attribute. But after doing all of that, then it becomes easy in the CSS to select against that special attribute to see if we should be in light or dark mode.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: I've change some of the selectors during the changes to make jupyter-only environments be responsive to dark mode. While doing that, I removed the extra selectors you pointed out.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: This was the color scheme that copilot suggested and I tweaked them a bit after until I thought they looked good. It is sort of difficult to find color combinations with three different colors that are all mutually visible against one another, which is needed here as the histogram bars have to be visible against the background, but the labels also need to be visible against both the bars and the background (for when they flow over the top of the bar graphic). My thinking is that we can play around with the color scheme a bit more to make things consistent across the extension if we really want to, but this color scheme makes everything visible and easily legible, which seems most important and is the motivation for the PR.

minestarks:
tested on vs code & jupyterlab on my machine, looks good to me. I'd prefer the Ket labels to be regular weight (instead of bold) though.

source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -345,102 +406,110 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
...
+      --vscode-editor-font-family,
+      "Consolas",
+      "Menlo",
+      "Monaco",
+      "Liberation Mono",
+      "Courier New",
+      monospace
+    ),
+    "Segoe UI Symbol";
```
billti: Why is `Segeo UI Symbol` here? If you want it to be a fallback, why is it not with the other fallbacks above? But do you even want it to be a fallback, being it is not a monospace font, and the generic family `monospace` above should always resolve, so should never get to it anyway.

source/widgets/js/index.tsx
```diff
@@ -38,9 +38,126 @@ type RenderArgs = {
...
+  }
+}
+
 function render({ model, el }: RenderArgs) {
   const componentType = model.get("comp");
 
+  // Set a stable color-scheme marker for CSS (works in Jupyter-only and VS Code).
+  // This is especially important when AnyWidget uses Shadow DOM, where selectors
+  // like body.jp-mod-theme-dark won't match inside the widget stylesheet.
+  applyAndObserveTheme(el);
```
billti: I'm a little worried about this, and don't think it should go in this close to shipping without a lot more diligence and testing.

For example, the `render` method this call is in can be called at a very high frequence for the same widget (e.g. when the histogram animates such as when dragging a slider to change the noise and re-run, or when live updating with incoming shot results - basically any time Python wants to send new values to the widget to display).

Yet looking at this code it seems on every call to render you are doing quite a bit of work at the document level that could be quite expensive or cause a bit of document wide recalculating (e.g. if changing attributes of classes on the body node, as calling applyAndObserveTheme always calls apply, which always calls computeColorScheme and always calls setAttribute on the host. Bare in mind the containing document may be more than just a VS Code webview or Jupyter cell (e.g. we put the Histogram on the playground page, and there the 'document' is the entire playground).

As this is now global code that runs for every widget (not just histograms), I'd also want to ensure we do sufficient testing when multiple widgets are on the same "host", or even nested (as they are with some of the Resource Estimation views, which include the space chart, scatter chart, details table widgets all as children of the another container widget).

Bottom line: Maybe it's fine, but there's a lot of subtlety and hazzards here for some scenarios, so I think it should wait until after the Jan release. Thanks.

ScottCarda-MS: Would it make sense to separate this logic, which is meant to be a fix for an existing bug for widgets in Jupyter, into a separate PR from the rest of the changes in this PR? That way we could have this PR focused on making stylistic changes for readability.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: Removed Segeo UI Symbol as a fallback.

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: Since this is a more involved fix for a pre-existing bug, I've moved the logic to fix that to a separate branch and kept this PR about updating the style. See https://github.com/microsoft/qdk/blob/sccarda/JupyterDarkThemeFix/source/widgets/js/index.tsx#L41 for the fix in the separate branch.

source/widgets/js/index.tsx
ScottCarda-MS: Since this is a more involved fix for a pre-existing bug, I've moved the logic to fix that to a separate branch and kept this PR about updating the style. See https://github.com/microsoft/qdk/blob/sccarda/JupyterDarkThemeFix/source/widgets/js/index.tsx#L41 for the fix in the separate branch.

PR #2856, opened 2026-01-16
Types to define ISAs and ISA requirements

msoeken:
This pull request sets up the integration of QRE into the code base. In the first PR it defines classes and functions to create ISAs and ISA requirements.

It seems that the changes in Github code owners submitted to the feature branch was not sufficient.

### DIFF
--- a/source/pip/qsharp/qre/_instruction.py
+++ b/source/pip/qsharp/qre/_instruction.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from typing import Optional, overload
+from typing import Optional, overload, cast
 from enum import IntEnum
 
 from ._qre import (
@@ -55,18 +55,18 @@ def instruction(
     arity: int = 1,
     space: Optional[int] = None,
     length: Optional[int] = None,
-    error_rate: Optional[float] = None
+    error_rate: float
 ) -> Instruction: ...
 @overload
 def instruction(
     id: int,
     encoding: Encoding = PHYSICAL,
     *,
-    time: IntFunction,
+    time: int | IntFunction,
     arity: None = ...,
     space: Optional[IntFunction] = None,
     length: Optional[IntFunction] = None,
-    error_rate: Optional[FloatFunction] = None
+    error_rate: FloatFunction
 ) -> Instruction: ...
 def instruction(
     id: int,
@@ -76,7 +76,7 @@ def instruction(
     arity: Optional[int] = 1,
     space: Optional[int] | IntFunction = None,
     length: Optional[int | IntFunction] = None,
-    error_rate: float | FloatFunction = None
+    error_rate: float | FloatFunction
 ) -> Instruction:
     """
     Creates an instruction.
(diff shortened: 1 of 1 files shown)


source/pip/qsharp/qre/_instruction.py
```diff
@@ -0,0 +1,92 @@
...
+            qubits. If None, arity is used.
+        error_rate (float | FloatFunction): The instruction error rate.
+
+    Returns:
+        Instruction: The instruction.
+    """
+    if arity is not None:
+        return Instruction.fixed_arity(
+            id, encoding, arity, time, space, length, error_rate
+        )
```
orpuente-MS, line 109: `Instruction.fixed_arity` assumes that `time`, `space`, and `length` are ints, but from the signature of `instruction` they could be `IntFunction` as well.

msoeken: This is intended, but I can try to make this more precise with an `@overload` signature. There are two types of instructions and depending on whether `arity` is `None` or not, I decide which type to take. Each type takes different types for `time`, `space`, and `length`.

source/pip/qsharp/qre/instruction_ids.py
```diff
@@ -0,0 +1,88 @@
...
+PREP_Y = 0x31
+PREP_Z = 0x32
+
+# Generic Cliffords
+ONE_QUBIT_CLIFFORD = 0x50
+TWO_QUBIT_CLIFFORD = 0x51
+N_QUBIT_CLIFFORD = 0x52
+
+# Measurements
+MEAS_X = 0x100
```
brad-lackey, line 41: Do we want to distinguish between destructive X-measurement versus nondemolition X-measurement? This would have different signatures. Similarly for Y- and Z-measurement.

source/pip/qsharp/qre/instruction_ids.py
```diff
@@ -0,0 +1,88 @@
...
+RYY = 0x413
+RZZ = 0x414
+
+# Multi-qubit Pauli measurement
+MULTI_PAULI_MEAS = 0x1000
+
+# Some generic logical instructions
+LATTICE_SURGERY = 0x1100
+
+# Memory/compute operations (used in compute parts of memory-compute layouts)
```
brad-lackey, line 83: Should we move higher-order primitives futher up the stack? Namely, the gate-injection circuit seems to be appropriate here, and read/write from memory goes further down, say 0x80??.

source/pip/qsharp/qre/_instruction.py
msoeken: Added `@overload` signatures to distinguish the two cases.

source/pip/qsharp/qre/instruction_ids.py
msoeken: Yes, I added `MEAS_RESET_*` for destructive *-measurement and assume the other variant for nondemolition.

source/pip/qsharp/qre/instruction_ids.py
msoeken: Let's reorganize this at a later time. I might want to encode other properties in the instruction ID when building compact traces (e.g., arity or parameter types).

PR #2857, opened 2026-01-16
Update branding to Microsoft Quantum

minestarks:


samples/notebooks/iterative_phase_estimation.ipynb
```diff
@@ -14,7 +14,7 @@
     "\n",
     "We also configure the target profile as `Adaptive_RI` to indicate we use the QIR profile with support for mid-circuit measurement, measurement-based control flow, and classical integer computation as part of compilation.\n",
     "\n",
-    "Replace the `subscription_id`, `resource_group`, `name`, and `location` connection parameters with the values for your configured Azure Quantum Workspace, or try copying the Python code for connecting to your workspace from the VS Code Azure Quantum Extension."
+    "Replace the `subscription_id`, `resource_group`, `name`, and `location` connection parameters with the values for your configured Azure Quantum Workspace, or try copying the Python code for connecting to your workspace from the VS Code Microsoft Quantum Extension."
```
ScottCarda-MS:
```suggestion
    "Replace the `subscription_id`, `resource_group`, `name`, and `location` connection parameters with the values for your configured Azure Quantum Workspace, or try copying the Python code for connecting to your workspace from the VS Code Microsoft Quantum Development Kit Extension."
```

source/vscode/package.json
```diff
@@ -908,7 +908,7 @@
       {
         "name": "qdk-run-program",
         "tags": [
-          "azure-quantum",
+          "quantum",
```
ScottCarda-MS: I'm unsure if these should be changed away from "azure-quantum". Isn't this referring to the azure-quantum python package?

minestarks: These are tags for the GitHub Copilot tools we provide. The ones I'm changing are the ones that don't actually have anything to do with Azure Quantum (run code locally, etc). So I think this is right, but on the other hand, since I have no idea where (if anywhere) these tags are even used, who knows? 🤔 

> A set of tags that roughly describe the tool's capabilities. A tool user may use these to filter the set of tools to just ones that are relevant for the task at hand, or they may want to pick a tag that can be used to identify just the tools contributed by this extension.

(That's what the VS Code documentation says, but I don't see any actual filtering functionality anywhere).

Anyway, I guess it's more important that we're just consistent. Maybe I'll just apply ALL the tags to ALL the tools.

PR #2858, opened 2026-01-16
Support QDK_PYTHON_TELEMETRY environment variable

minestarks:


source/pip/qsharp/telemetry.py
```diff
@@ -39,16 +39,18 @@
...
+# If explicitly enabled via either environment variable, enable telemetry.
+# Otherwise, enable telemetry only in release builds.
+_disable_values = {"0", "false", "disabled", "none"}
+_enable_values = {"1", "true", "enabled"}
+_env_values = {
+    (os.environ.get("QSHARP_PYTHON_TELEMETRY") or "").lower(),
+    (os.environ.get("QDK_PYTHON_TELEMETRY") or "").lower(),
+}
+
+TELEMETRY_ENABLED = not _env_values & _disable_values and (
```
billti, line 55: What is this sourcery! Does Python let you 'bitwise' AND between dict entries or something :-/ Maybe a comment on the "cleverness" used here for future readers.

minestarks: You probably already know, but we can thank Copilot for this.

PR #2863, opened 2026-01-20
Add sims and widgets

billti:


source/paulimer/Cargo.toml
minestarks, line 1: should this be added to the workspace Cargo.toml?

source/vscode/test/runTests.mjs
minestarks, line 1: So this PR is supposed to disable tests again?

source/qdk_package/pyproject.toml
```diff
@@ -10,7 +10,7 @@ readme = "README.md"
 authors = [ { name = "Microsoft" } ]
 license = { file = "LICENSE.txt" }
 requires-python = ">=3.10"
-dependencies = ["qsharp==0.0.0"]
+dependencies = ["qsharp==0.0.0", "pyqir<0.12"]
```
minestarks, line 13: Just checking that this is intentional? to have an unconditional dependency on pyqir?

source/simulators/Cargo.toml
minestarks, line 1: does this need to be added to the workspace cargo.toml?

minestarks (approved):
left a few comments

source/qdk_package/pyproject.toml
ScottCarda-MS: I do remember discussing this, and I believe that this was the decision we came to, that pyqir was going to be a default dependency.

source/paulimer/Cargo.toml
orpuente-MS: Adding it to the members field in the workspace Cargo.toml doesn't seem necessary according to the documentation, since cargo can infer this info. But it won't hurt making it explicit to match the pattern we've been following until now.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

source/vscode/test/runTests.mjs
swernli: I don't think it should... those tests have been running fine here lately.

billti: Let's get it in then re-enable if stuff is passing reliably. I didn't want to re-enable again and then we're blocked getting a PR in or a build out.

source/paulimer/Cargo.toml
billti: Stefan is going to check some stuff and do a follow up (if needed)

source/simulators/Cargo.toml
billti: As above.

PR #2870, opened 2026-01-21
Circuit Editor State Visualization Panel

ScottCarda-MS:
This adds a collapsible State Visualization Panel to the Circuit Editor for seeing the quantum state resulting from running the circuit on qubits assumed to be in |0> state. The panel starts collapse on the right side of the editor window and expands when clicked:
State Viz 1

The state is show with a probability density bar graph for each base state in superposition, and a phase diagram to indicate the phase associated with that base state as well. The color of the bar in the bar graph as well as the phase diagram also indicate phase:
State Viz 2

The default maximum number of columns is 16. If there are too many states to represent, an "Others" column will be created. The following shows this as well as that state labels will go vertical if too long:
State Viz Other

If the circuit is empty, the visualizer shows that it can't represent the state of an empty circuit:
State Viz Empty

Currently the visualizer can't support Measurement or Reset gates, as they create mixed states, and the visualizer is currently only able to show one superposition state:
image

Lastly I want to point out some things that are intentionally left out of this PR but may be implemented in future PRs:
- While some snapshot tests where added for the state visualizer, I didn't add snapshot tests showing the whole editor with an opened visualizer panel. I had difficulties getting snapshots to be consistent (exactly the same) across different environments. This is something I can look into in the future.
- Setting the panel to be expanded by default instead of collapsed. Currently the panel starts collapsed, and this helps obfuscate the initialization process of the panel. When starting expanded the initialization causes some resizing that may look strange to the user. Having the panel expanded by default also causes some of the snapshots tests to try to capture the panel mid-initialization which looks strange and may appear unexpected.
- Removing the "Run" button from the toolbox. The "Run" button is arguably superseded by the state panel as they were both meant to help circuit-writers be able to see what state the circuits they were writing would produce. The state visualizer, I believe, does a much better job of this as it is much faster, doesn't require clicking a button, shows the state in a much more ideal way, and doesn't require using (maybe considered abusing) the entry point expression of Q# programs which was controversial.
- Removing the frontend simulator in stateCompute.ts and use the rust simulator. The frontend simulator is very simple and works well (and fast) with the limited circuits that can appear in the circuit editor. There are concerns that this simulator will not scale nicely as we add more complex capabilities to the circuit editor, however, and we don't want to maintain more simulators than are needed, so there is a desire to replace the usage of the frontend simulator with the use of the rust simulator, which is currently used by the "Run" button. This is another reason to keep the "Run" button until we are able to make use of the rust simulator in the state visualizer panel. This is something that I want to do in a follow up PR.

### DIFF
--- a/source/npm/qsharp/test/stateViz.js
+++ b/source/npm/qsharp/test/stateViz.js
@@ -19,7 +19,7 @@ import {
   createStatePanel,
   updateStatePanelFromColumns,
 } from "../dist/ux/circuit-vis/state-viz/stateViz.js";
-import { prepareStateVizColumnsFromAmpMap } from "../dist/ux/circuit-vis/state-viz/stateVizPrep.js";
+import { prepareStateVizColumnsFromAmpMap } from "../dist/ux/circuit-vis/state-viz/worker/stateVizPrep.js";
 
 const documentTemplate = `<!doctype html><html>
   <head>


source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
```diff
@@ -0,0 +1,98 @@
...
+  computeAmpMapForCircuit,
+  type Endianness,
+} from "./stateComputeCore.js";
+import {
+  prepareStateVizColumnsFromAmpMap,
+  type PrepareStateVizOptions,
+} from "./stateVizPrep.js";
+import type { StateColumn } from "./stateViz.js";
+
+const STATE_COMPUTE_LOG_PREFIX = "[qsharp][state-compute]";
```
billti: We should use `qdk`, not `qsharp`.

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
```diff
@@ -0,0 +1,98 @@
...
+  const nodeProcess = (globalThis as any)?.process as any;
+  const isNode =
+    typeof nodeProcess !== "undefined" && !!nodeProcess?.versions?.node;
+  if (isNode) return;
+
+  if (typeof console === "undefined" || typeof console.debug !== "function") {
+    return;
+  }
+
+  console.debug(
```
billti: Wny are you logging to console at all?

source/npm/qsharp/ux/qsharp-circuit.css
```diff
@@ -494,3 +501,306 @@
...
+}
+
+/* Full-height clickable edge with vertical text */
+.state-edge {
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  width: 36px;
+  background: var(--vscode-editorWidget-background, #e6e6e6);
```
billti: We should only use color vars that come from the `qdk-theme.css` to ensure consistency and avoid the theming issues we hit with histogram and atom visualizer.

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
ScottCarda-MS: Sorry, are you saying that there is a different way I should be logging, or that I shouldn't be logging at all? The logging is very helpful in debugging to make sure the web worker is starting and running as expected. I see that there is a log.debug method used in the src/ folder, but importing that into content in the ux/ folder not safe.

source/npm/qsharp/ux/qsharp-circuit.css
ScottCarda-MS: Updated! All the state-viz related classes should now use --qdk variables instead of --vscode variables.

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
ScottCarda-MS: Done

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
ScottCarda-MS: I removed the logging to try and simplify this PR a bit.

source/vscode/src/webview/stateComputeWorker.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { computeAmpMapForCircuit } from "../../../npm/qsharp/ux/circuit-vis/state-viz/stateComputeCore";
+import { prepareStateVizColumnsFromAmpMap } from "../../../npm/qsharp/ux/circuit-vis/state-viz/stateVizPrep";
```
billti: This is not how you should be importing from the npm folder. Import from the package like all the other code we have.

ScottCarda-MS: We can't import from qsharp-lang/ux at this location in the worker code because that attempts to pull in the top-level CSS content. Because of the way this code gets built and inlined, there is nowhere for that built CSS file to be emitted.
However, I can still import from the npm and avoid relative imports like this if I make my own state-viz entrypoint that avoids any DOM or CSS content, so that's what I've done. Added an index.ts to the state-viz folder that acts as the entrypoint for the work to access the state-compute code without touching any DOM or CSS content.

source/npm/qsharp/package.json
```diff
@@ -22,6 +22,7 @@
     "./debug-service-worker": "./dist/debug-service/worker-browser.js",
     "./katas": "./dist/katas.js",
     "./katas-md": "./dist/katas-md.js",
+    "./state-viz": "./ux/circuit-vis/state-viz/index.ts",
     "./ux": "./ux/index.ts",
```
minestarks: Does this have to be its own entrypoint? Everything else is (re)exported from `./ux` , can we follow that pattern?

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
```diff
@@ -0,0 +1,65 @@
...
+  computeAmpMapForCircuit,
+  UnsupportedStateComputeError,
+} from "./stateComputeCore.js";
+import {
+  prepareStateVizColumnsFromAmpMap,
+  type PrepareStateVizOptions,
+} from "./stateVizPrep.js";
+import type { StateColumn } from "./stateViz.js";
+
+const MAX_QUBITS_FOR_STATE_VIZ = 20;
```
minestarks: Suggestion: would be nice to plumb this into the circuit configuration in VS Code. (not really asking to be done in this PR)

source/npm/qsharp/ux/circuit-vis/sqore.ts
```diff
@@ -52,6 +52,7 @@ export type DrawOptions = {
   editCallback?: (circuitGroup: CircuitGroup) => void;
   runCallback?: () => void;
   renderLocations?: (l: SourceLocation[]) => { title: string; href: string };
+  statePanelInitiallyExpanded?: boolean;
```
minestarks: Is this used? I don't see it being set anywhere.

Also I wanted to suggest having a config to enable/disable the State Panel entirely. It would help not to clutter the `.qsc` file HTML snapshots (which are supposed to validate the circuit rendering itself)

source/npm/qsharp/ux/circuit-vis/state-viz/worker/stateCompute.ts
```diff
@@ -0,0 +1,261 @@
...
+// Core state computation for circuit-vis.
+// Implements a small statevector simulator that evaluates the circuit model and
+// produces an amplitude map. Intentionally avoids DOM/visualization concerns so
+// it can run on the main thread or in a Web Worker.
+
+import type { ComponentGrid, Operation, Qubit } from "../circuit.js";
+import { evaluateAngleExpression } from "../angleExpression.js";
+
+// This is intentionally a minimal structural type so it can be used from both the
+// main thread and a Web Worker without importing DOM-heavy visualization modules.
```
minestarks: This comment sounds a bit vague/inaccurate to me... Maybe we can rephrase it this way (with Copilot's help) to be more precise?

```
// Only using primitives in this type to enable serialization for cross-thread communication.
// This module itself avoids DOM dependencies so it can run in Web Workers.
```

Though now that I asked you, it sounds like we don't use it for data transfer between threads. Do we just need to update the comment?

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// State compute bridge for circuit-vis.
```
minestarks: It would be helpful to be able to easily tell which file gets called in which context (main thread vs web worker) . Can we add that information to the top comment? (Or separate them into different folders?)

source/npm/qsharp/ux/circuit-vis/state-viz/worker/stateCompute.ts
```diff
@@ -0,0 +1,261 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// Core state computation for circuit-vis.
```
minestarks, line 4: Like the other file, please disambiguate where this is run (main/worker/both)

source/vscode/src/webview/editor.tsx
```diff
@@ -26,9 +28,196 @@ type State = { viewType: "loading" } | CircuitState;
...
+    ) => computeStateVizColumnsInWorker(model, opts),
+
+    dispose: (reason?: string) => {
+      cancelActiveStateComputeWorker(reason ?? "disposed");
+    },
+  };
+
+  (globalThis as any).qsharpStateComputeApi = api;
+}
+
```
minestarks: I wonder if we can pass this api into sqore.draw() just as we're passing the `runCallback` and `editCallback` etc.

source/npm/qsharp/ux/circuit-vis/events.ts
```diff
@@ -55,6 +56,26 @@ const enableEvents = (
     events.dispose();
   }
   events = new CircuitEvents(container, sqore, useRefresh);
+
+  // Track which rendered SVG the current `events` instance is associated with.
+  // This lets other modules avoid reading a stale model during a re-render where
+  // the SVG has been replaced but `enableEvents` hasn't run yet.
+  currentCircuitSvg = container.querySelector("svg.qviz") as SVGElement | null;
```
minestarks, line 69: I wonder if you can just keep track of the current `Circuit` instead of the rendered SVG . Since that's ultimately the data you'll be using.

ScottCarda-MS: I've tried a couple of different way to make this work with `Circuit` instead of svgs, but I always end up with much more complexity and the circuit editor becomes unstable with flakiness and timing errors. I'm going to leave this as-is for now, and if I come to a good solution in the future, I can make a new PR for this specifically.

source/vscode/src/webview/editor.tsx
ScottCarda-MS: Done. This provided a good opportunity for me to clean up the DrawOptions type as well.

source/npm/qsharp/ux/circuit-vis/state-viz/worker/stateCompute.ts
ScottCarda-MS: The files that are associated with the state-compute worker are now in their own subfolder `worker/` to distinguish them from the content that is meant to run on the main thread.

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
ScottCarda-MS: The files that are associated with the state-compute worker are now in their own subfolder `worker/` to distinguish them from the content that is meant to run on the main thread.

source/npm/qsharp/ux/circuit-vis/sqore.ts
ScottCarda-MS: This was used in snapshot tests to keep the panel open, but those tests got factored out, so this is no longer used. Removed!

source/npm/qsharp/package.json
ScottCarda-MS: It does need to be separate. The ./ux entrypoint pulls in a lot of css content that can't be build with the worker content. This new entrypoint is specifically small and free of css so that it can be built with the web worker content.

source/npm/qsharp/ux/circuit-vis/state-viz/worker/stateCompute.ts
ScottCarda-MS: Updated the comment. The type is intended to be used in the web worker, but could also be used by the main thread in the fallback scenario. However, it is no longer something that is communicated across the thread boundary.

source/npm/qsharp/ux/circuit-vis/state-viz/stateCompute.ts
ScottCarda-MS: It is something to keep in mind for the future. There are some ideas for efficiency gains that could be implemented in a future PR and this efficiency gain here, limiting circuit size, could be reworked a bit during that.

PR #2872, opened 2026-01-21
[Language Service] Fix project path on notebooks when running on Windows

### LINKED ISSUE #2866: [Language Service] Project paths when calling `qsharp.init()` in notebooks are invalid on Windows
**Describe the bug**

When opening a notebook that loads a project, red squiggles are shown when using members of that project.

**To Reproduce**

1. Open the carbon notebook in `samples/notebooks/carbon_error_correction/carbon.ipynb`
2. Click Run All
3. See error

**Expected behavior**

There shouldn't be any red squiggles

**Screenshots**

Image

If click on the path URL in the output of the first cell, I get this error:

Image

**System information**

- Version of the package or extension for which the problem occurs: v1.23.4
- Your operating system: Windows

orpuente-MS:
Fixes #2866

### DIFF
--- a/source/pip/qsharp/_qsharp.py
+++ b/source/pip/qsharp/_qsharp.py
@@ -38,9 +38,9 @@ from .estimator._estimator import (
 )
 import json
 import os
-import pathlib
 import sys
 import types
+from pathlib import Path
 from time import monotonic
 from dataclasses import make_dataclass
 
@@ -161,8 +161,7 @@ class Config:
             # For now, we only support local project roots, so use a file schema in the URI.
             # In the future, we may support other schemes, such as github, if/when
             # we have VS Code Web + Jupyter support.
-            abs_project_root = pathlib.Path(os.getcwd(), project_root)
-            self._config["projectRoot"] = abs_project_root.as_uri()
+            self._config["projectRoot"] = Path(os.getcwd(), project_root).as_uri()
 
     def __repr__(self) -> str:
         return "Q# initialized with configuration: " + str(self._config)


source/pip/qsharp/_qsharp.py
```diff
@@ -161,7 +161,7 @@ def __init__(
             # In the future, we may support other schemes, such as github, if/when
             # we have VS Code Web + Jupyter support.
             normalized_root = os.path.normpath(os.path.join(os.getcwd(), project_root))
-            self._config["projectRoot"] = "file://" + normalized_root
+            self._config["projectRoot"] = "file:///" + normalized_root
```
orpuente-MS: My VS-Code instance was confusing the drive `c:` with the authority of the URI. The URI spec says that the way to specify an empty authority is by doing `"file:///path"`

https://datatracker.ietf.org/doc/html/rfc8089#appendix-B

source/vscode/src/projectSystem.ts
```diff
@@ -105,7 +105,14 @@ export async function listDirectory(
 
   const fileSearchResult = await vscode.workspace.fs.readDirectory(uriToQuery);
   const mappedFiles: [string, vscode.FileType][] = fileSearchResult.map(
-    ([name, type]) => [Utils.joinPath(uriToQuery, name).toString(), type],
+    ([name, type]) => [
+      // Rust expects paths, not encoded URIs, so we need to decode here.
+      // We need to decode the components and then the URI.
+      decodeURI(
+        decodeURIComponent(Utils.joinPath(uriToQuery, name).toString()),
```
billti: Are there risks here when not working with file system paths (e.g. when we load files from github, or in the web experience)?

source/pip/qsharp/_qsharp.py
billti: This seems odd in that on Mac and Linux the path will already start with a forward-slash. Having the extra slash may be benign, but also isn't correct.

source/vscode/src/projectSystem.ts
billti: Is this just solving the same problem you already had to do recently for the `#` char @minestarks ? This look safe to you?

source/vscode/src/projectSystem.ts
```diff
@@ -105,7 +105,14 @@ export async function listDirectory(
 
   const fileSearchResult = await vscode.workspace.fs.readDirectory(uriToQuery);
   const mappedFiles: [string, vscode.FileType][] = fileSearchResult.map(
-    ([name, type]) => [Utils.joinPath(uriToQuery, name).toString(), type],
+    ([name, type]) => [
+      // Rust expects paths, not encoded URIs, so we need to decode here.
```
minestarks: > Rust expects paths

I don't think this is right - in general, all the paths in the language service should be URIs coming directly from VS Code. Round tripping VS Code-generated Uris with with `vscode.Uri.toString()` and `vscode.Uri.parse()` should work fine without any need for decoding

source/pip/qsharp/_qsharp.py
orpuente-MS: I am using `pathlib.Path.as_uri()` now. So, we are not doing manual URIs anymore.

source/vscode/src/projectSystem.ts
orpuente-MS: I removed this change

source/vscode/src/projectSystem.ts
orpuente-MS: I removed this change

minestarks:
There are still issues in remote VS Code scenarios such as WSL, Codespaces,  VS Code Web . But that's not new with this PR, nor does this PR make that issue worse. @orpuente-MS  and I discussed a possible fix for the remote scenarios. (Using the scheme from the .ipynb file, and "mashing it up" with the local OS path retrieved from Python). I'm still approving this PR since it fixes the intended case (local Windows)

PR #2880, opened 2026-01-22
Minor samples cleanup

swernli:
This removes the inadvertantly checked cell outputs for carbon.ipynb, removes a todo from benzene.ipynb, and updates how the chemistry sample loads the chemistry library so that it's more stand-alone.

### DIFF
--- a/library/fixed_point/qsharp.json
+++ b/library/fixed_point/qsharp.json
@@ -5,7 +5,7 @@
     "Signed": {
       "github": {
         "owner": "Microsoft",
-        "repo": "qsharp",
+        "repo": "qdk",
         "ref": "v1.25.0",
         "path": "library/signed"
       }
--- a/library/signed/qsharp.json
+++ b/library/signed/qsharp.json
@@ -5,7 +5,7 @@
     "Qtest": {
       "github": {
         "owner": "Microsoft",
-        "repo": "qsharp",
+        "repo": "qdk",
         "ref": "v1.25.0",
         "path": "library/qtest"
       }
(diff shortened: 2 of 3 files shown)


source/samples_test/build.rs
```diff
@@ -226,6 +226,15 @@ fn collect_qsharp_project_folders(path: &Path) -> Vec<PathBuf> {
         if entry.is_dir() {
             projects.append(&mut collect_qsharp_project_folders(entry));
         } else if Some("qsharp.json") == entry.file_name().and_then(OsStr::to_str) {
+            if entry
+                .parent()
+                .expect("file should have parent dir")
+                .ends_with("SPSA")
+            {
+                // Skip the SPSA project since it has GitHub-based dependencies that
```
DmitryVasilevsky, line 234: Can you mention in the comment that this is for running tests?

samples/chemistry/SPSA/qsharp.json
```diff
@@ -1,7 +1,12 @@
 {
-  "dependencies": {
-    "Chemistry": {
-      "path": "../../../library/chemistry"
+    "dependencies": {
+        "Chemistry": {
+            "github": {
+                "ref": "v1.25.0",
+                "owner": "microsoft",
+                "repo": "qsharp",
```
billti: We should be refering to 'qdk' repo now.

swernli: Oh, that points to another issue. This was auto-inserted by the extension. That means the code that handles that insertion needs to be updated.

source/samples_test/build.rs
swernli: The comment does already say "can't be resolved with the test filesystem" and this is in `source/samples_test/build.rs` so I think there is enough context for it here.

samples/chemistry/SPSA/qsharp.json
swernli: never mind, I'm just on an older extension. The fix to use "qdk" instead of "qsharp" went in yesterday and isn't in a dev extension yet.

PR #2883, opened 2026-01-23
Update extension readme

swernli:


source/vscode/README.md
minestarks: Unrelated to your change, but I wonder if this part should say something like 


```suggestion
- Editing features such as go-to-definition, suggestions and signature help for Q# and OpenQASM
- Code samples for Q# and OpenQASM demonstrating well known algorithms
```

swernli: I do think that reads better. Could even be combined with the first point about syntax highlighting.

PR #2891, opened 2026-01-27
Show doc comments on hover of struct fields

### LINKED ISSUE #2841: Allow doc comments on struct fields

Sometimes struct fields have meaning that's hard to derive from the description of the struct. Doc comments would help in such cases. Unfortunately, we don't support doc comments on fields.

Allow doc comments on struct fields and see them in mouse over hints.

Alternatively, we can do nothing.

swernli:
This adds support for parsing doc comments on the new struct-style syntax for UDTs. In addition, this captures the doc comment contents for both struct fields and named `newtype` fields (the older syntax) such that they can be included in hover on uses of the field. This fixes #2841.

Note that this does NOT add them to generated API docs, as we need some more thinking to find the right way to merge that with any existing doc comments on the UDT itself, where things like markdown are supported. A follow up issue will be filed once this merges to continue to generated API docs support.

Before:
image

After:
image

### DIFF
--- a/source/compiler/qsc_parse/src/item/tests.rs
+++ b/source/compiler/qsc_parse/src/item/tests.rs
@@ -203,9 +203,9 @@ fn allow_docs_on_struct_fields() {
             Item _id_ [0-118]:
                 Struct _id_ [0-118] (Ident _id_ [7-10] "Foo"):
                     FieldDef _id_ [51-58] (Ident _id_ [51-52] "x"): Type _id_ [55-58]: Path: Path _id_ [55-58] (Ident _id_ [55-58] "Int")
-                        doc: doc for x
+                        Doc: doc for x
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_parse/src/item/tests.rs
```diff
@@ -247,9 +267,11 @@ fn udt_item_doc() {
                     TyDef _id_ [56-66]: Field:
                         Ident _id_ [56-60] "arg1"
                         Type _id_ [63-66]: Path: Path _id_ [63-66] (Ident _id_ [63-66] "Int")
+                        doc-string for arg1
```
ScottCarda-MS: I think we should have some standardized indication that what we are printing out here is a doc string, like we do above with `doc:`. Though in both cases I would capitalize `Doc:` to match with other fields of objects that get printed.

swernli: Updated!

PR #2894, opened 2026-01-27
Enumerate ISAs with transforms and queries

msoeken:
This implements ISA transforms that can provide new ISAs from existing ones. It also builds data structures to define ISA queries that can combine transformations and their parameters to define the ISA exploration space.

### DIFF
--- a/source/pip/qsharp/qre/_isa_enumeration.py
+++ b/source/pip/qsharp/qre/_isa_enumeration.py
@@ -39,16 +39,25 @@ class Node(ABC):
 
     def __add__(self, other: Node) -> SumNode:
         """
-        Combines two enumeration nodes in an additive manner.
+        Performs a union of two enumeration nodes.
 
-        Enumerating the sum node yields all ISAs from this node, followed by
-        all ISAs from the other node.
+        Enumerating the sum node yields all ISAs from this node, followed by all
+        ISAs from the other node.  Duplicate ISAs may be produced if both nodes
+        yield the same ISA.
 
         Args:
             other (Node): The other enumeration node.
 
         Returns:
             SumNode: A node representing the union of both enumerations.
+
+        Example:
+
+            The following enumerates ISAs from both SurfaceCode and ColorCode:
+
+        .. code-block:: python
+            for isa in SurfaceCode.q() + ColorCode.q():
+                ...
         """
         if isinstance(self, SumNode) and isinstance(other, SumNode):
             sources = self.sources + other.sources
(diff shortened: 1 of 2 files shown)


source/pip/qsharp/qre/_enumeration.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Generator, Type, TypeVar, Literal, get_args, get_origin
+from dataclasses import MISSING
+from itertools import product
+from enum import Enum
+
+
+T = TypeVar("T")
+
+
+def _enumerate_instances(cls: Type[T], **kwargs) -> Generator[T, None, None]:
```
brad-lackey, line 13: Does this really belong in _architecture.py? It seems like something that would be more suited to a utilities module.

brad-lackey: Nevermind--lost track of what file I was looking at. Okay, _enumeration.py is that module. Got it.

source/pip/qsharp/qre/_isa_enumeration.py
```diff
@@ -0,0 +1,329 @@
...
+        Args:
+            ctx (Context): The enumeration context containing shared state,
+            e.g., access to the underlying architecture.
+
+        Yields:
+            ISA: A possible ISA that can be generated from this node.
+        """
+        pass
+
+    def __add__(self, other: Node) -> SumNode:
```
brad-lackey, line 40: I think we may want to be careful here when we document ISAs. Implicitly, a generator will kick out ISAs in a specific order. But I don't think we want to claim that functionality. In fact, the opposite: we should indicate that a generator could returns ISAs in an arbitrary (even nondeterministic) ordering. In this case the "sum" is the union, and we should not guarantee that the first ISA is enumerated first, followed by the second (even though that might happen in practice).

brad-lackey: I was just thinking about whether this should be the disjoint union (i.e. the coproduct--naturally dual to the product below). Namely, should an ISA implement a notion of equivalence? If not, then we might have two functionally equivalent ISAs after the sum. Is that desirable?

source/pip/qsharp/qre/_isa_enumeration.py
```diff
@@ -0,0 +1,329 @@
...
+        else:
+            return SumNode([self, other])
+
+    def __mul__(self, other: Node) -> ProductNode:
+        """
+        Combines two enumeration nodes in a multiplicative manner.
+
+        Enumerating the product node yields ISAs resulting from the Cartesian
+        product of ISAs from both nodes. The ISAs are combined using
+        concatenation (logical union).
```
brad-lackey: See above about logical union versus disjoint union.

source/pip/qsharp/qre/_qre.pyi
```diff
@@ -18,6 +18,13 @@ class ISA:
         """
         ...
 
+    def __add__(self, other: ISA) -> ISA:
+        """
+        Concatenates two ISAs. Instructions in the second operand overwrite
+        instructions in the first operand if they have the same ID.
```
brad-lackey: So this implements the logical union. See notes in _isa_enumeration.py about logical union versus disjoint union.

msoeken: Yes, this is a different `__add__` operator, the one that combines two ISAs into a single one. This is important when combining ISA generators with the cross product operator (`__mul__`). Let's say we have two codes, `SurfaceCode` and `ColorCode` that both produce ISAs with a `LATTICE_SURGERY` instruction. Then

```
SurfaceCode.q() * ColorCode.q()
```

would keep all instructions from the color code, whereas

```
ColorCode.q() * SurfaceCode.q()
```

would keep all instructions from the surface code.

This also means that we need to consider a larger enumeration space to make sure to obtain all ISAs. Worth to be very precise about the semantics in all these cases.

source/pip/qsharp/qre/_isa_enumeration.py
msoeken: It's the logical union for now (maybe we can change this later), but it is documented more clearly.

source/pip/qsharp/qre/_qre.pyi
msoeken: It's the logical union for now (maybe we can change this later), but it is documented more clearly.

source/pip/qsharp/qre/_isa_enumeration.py
msoeken: It's the logical union for now (maybe we can change this later), but it is documented more clearly.

PR #2899, opened 2026-01-28
CSS updates

billti:
This fixes the molecule viewer to react property in notebooks by using the observe/update mechansim I recently added.

I updated the neutral atom visualizer to use the new QDK theme variables recently added, which cleans that up a bit and fixes it on notebooks in the browser (it was always dark mode there before).

I also remove the playground specific styles out of the general npm/widgets CSS and into the playground.


source/widgets/js/index.tsx
```diff
@@ -287,7 +287,7 @@ function renderCircuit({ model, el }: RenderArgs) {
 
 function renderAtoms({ model, el }: RenderArgs) {
   const onChange = () => {
-    const machineLayout = model.get("machine_layout") as MachineLayout;
+    const machineLayout = model.get("machine_layout") as ZoneLayout;
```
ScottCarda-MS, line 290: nit: if this type is renamed to ZoneLayout, can other instances of "machine layout" be renamed to "zone layout", like this variable name, the hash map key string here, or the instances in qsharp_widgets/\_\_init\_\_.py?

billti: > the instances in qsharp_widgets/__init__.py

I can look into making more consistent, but being we did ship already, I can't rename anything that is publicly use (e.g. named args)

billti: Oh, forgot to mention - there actually was no MachineLayout export before, so the import was an error. Not sure why it only showed up in the editor and not the build. But yeah, the Python API already has the `machine_layout` arg and shipped with that, so will have to live with it now.

PR #2905, opened 2026-01-29
Add CPU simulators test suite with 100+ initial tests

orpuente-MS:
The main purpose of this PR is to build some infrastructure for writing tests for the CPU simulators. I also added tests for each gate of every simulator verifying known properties of those gates, but I might be missing something.

The easiest way to review this PR is to go to these four new files:
- `full_state_noiseless.rs`
- `full_state_noisy.rs`
- `clifford_noiseless.rs`
- `clifford_noisy.rs`

and read the module level docstring at the top of the file. That should have a summary with all the properties being tested for each gate in that file. And if you are interested on how a property is being tested you can jump to the corresponding test.

Missing in this PR: tests for CY and CCX gates, since those are still unimplemented.

### DIFF
--- a/source/pip/src/qir_simulation/cpu_simulators/tests/clifford_noiseless.rs
+++ b/source/pip/src/qir_simulation/cpu_simulators/tests/clifford_noiseless.rs
@@ -140,6 +140,11 @@ fn two_qubit_gate_truth_tables() {
             (qir! { cx(0, 1) }, 0b01 => 0b11),  // q0=1 → flip q1
             (qir! { cx(0, 1) }, 0b10 => 0b10),  // q0=0 → identity
             (qir! { cx(0, 1) }, 0b11 => 0b01),  // q0=1 → flip q1
+            // CY gate: phase only, flips q1 when q0=|1⟩
+            (qir! { cy(0, 1) }, 0b00 => 0b00),
+            (qir! { cy(0, 1) }, 0b01 => 0b11),  // q0=1 → flip q1
+            (qir! { cy(0, 1) }, 0b10 => 0b10),  // q0=0 → identity
+            (qir! { cy(0, 1) }, 0b11 => 0b01),  // q0=1 → flip q1
             // CZ gate: phase only, no bit changes
             (qir! { cz(0, 1) }, 0b00 => 0b00),
             (qir! { cz(0, 1) }, 0b01 => 0b01),
--- a/source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noiseless.rs
+++ b/source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noiseless.rs
@@ -150,6 +150,11 @@ fn two_qubit_gate_truth_tables() {
             (qir! { cx(0, 1) }, 0b01 => 0b11),  // q0=1 → flip q1
             (qir! { cx(0, 1) }, 0b10 => 0b10),  // q0=0 → identity
             (qir! { cx(0, 1) }, 0b11 => 0b01),  // q0=1 → flip q1
+            // CY gate: phase only, flips q1 when q0=|1⟩
+            (qir! { cy(0, 1) }, 0b00 => 0b00),
+            (qir! { cy(0, 1) }, 0b01 => 0b11),  // q0=1 → flip q1
+            (qir! { cy(0, 1) }, 0b10 => 0b10),  // q0=0 → identity
+            (qir! { cy(0, 1) }, 0b11 => 0b01),  // q0=1 → flip q1
             // CZ gate: phase only, no bit changes
             (qir! { cz(0, 1) }, 0b00 => 0b00),
             (qir! { cz(0, 1) }, 0b01 => 0b01),
--- a/source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
+++ b/source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
@@ -702,8 +702,9 @@ where
 
         assert!(
             actual_output == expected_output,
-            "Basis table test failed: gate={program:?} on input |{input_bits:0width$b}⟩ \
+            "Basis table test failed: gate={program:?} on input |{:0width$b}⟩ \
                  Actual output:\n{:#?}\nExpected output:\n{:#?}",
+            input_bits.reverse_bits() >> (32 - num_qubits),
             actual_output,
             expected_output,
             width = num_qubits as usize,
--- a/source/simulators/src/cpu_full_state_simulator.rs
+++ b/source/simulators/src/cpu_full_state_simulator.rs
@@ -94,10 +94,10 @@ static CX: LazyLock<Operation> = LazyLock::new(|| {
 
 static CY: LazyLock<Operation> = LazyLock::new(|| {
     let i = Complex::I;
-    operation!([1., 0., 0., 0.;
-                0., 1., 0., 0.;
+    operation!([1., 0., 0.,  0.;
                 0., 0., 0., -i;
-                0., 0., i,  0.;])
+                0., 0., 1.,  0.;
+                0., i,  0.,  0.;])
     .expect("operation should be valid")
 });
 


source/noisy_simulator/src/state_vector_simulator.rs
```diff
@@ -26,6 +27,58 @@ pub struct StateVector {
...
+                }
+                // Check that all elements match after applying the phase
+                self.data
+                    .iter()
+                    .zip(other.data.iter())
+                    .all(|(a, b)| eq_with_tolerance((a * phase - b).norm(), 0.0, TOLERANCE))
+            }
+            // The first vector is the zero vector.
+            // We return `true` iff the second vector is also the zero vector.
+            None => other.data.iter().all(|b| b.norm() <= TOLERANCE),
```
DmitryVasilevsky: Do you normalize? If so, is this even possible to have all values so small?

source/noisy_simulator/src/state_vector_simulator.rs
```diff
@@ -26,6 +27,58 @@ pub struct StateVector {
...
+        if !eq_with_tolerance(self.trace_change, other.trace_change, TOLERANCE) {
+            return false;
+        }
+
+        // Find the first non-zero element in self to determine the global phase
+        let phase = self
+            .data
+            .iter()
+            .zip(other.data.iter())
+            .find(|(a, _)| a.norm() > TOLERANCE)
```
DmitryVasilevsky: Consider finding the largest by norm. May improve precision.

source/pip/src/qir_simulation/cpu_simulators/tests/clifford_noiseless.rs
```diff
@@ -0,0 +1,581 @@
...
+//! | H       | H^2 ~ I, H X H ~ Z, creates superposition         |
+//! | S       | S^2 ~ Z, S S_ADJ ~ I                              |
+//! | S_ADJ   | S_ADJ^2 ~ Z                                       |
+//! | SX      | SX^2 ~ X, SX SX_ADJ ~ I                           |
+//! | SX_ADJ  | SX_ADJ^2 ~ X                                      |
+//! | CX      | CX|00⟩ = |00⟩, CX|10⟩ = |11⟩                      |
+//! | CZ      | CZ|x0⟩ = |x0⟩, CZ(a,b) = CZ(b,a)                  |
+//! | SWAP    | Exchanges states, SWAP^2 ~ I                      |
+//! | MZ      | MZ ~ MZ MZ (idempotent, does not reset)           |
+//! | RESET   | OP RESET ~ I (resets to |0⟩)                      |
```
DmitryVasilevsky: RESET and RESETZ aren't equivalent to I. They change the state of anything other than 0. However, check_programs_are_eq only check for starting in 0 state. This feels a bit misleading: on one hand, "A equivalent to B" sounds like the two should be the **same operators** (for example, XX ~ I). One the other hand, the code only checks if they are the same starting from 0 input. One needs to use something like Choi–Jamiołkowski isomorphism to test equivalence.
Having said that, I'm not sure you need to test the full equivalence for these basic tests. You probably need to clarify the definition of equivalence. Or, better yet, don't call it equivalence (even though it is an equivalence relation, just not the one people expect), and say something like "producing the same state starting from zero".

source/pip/src/qir_simulation/cpu_simulators/tests/clifford_noiseless.rs
```diff
@@ -0,0 +1,581 @@
...
+//! | Two-qubit       | CX, CZ, SWAP                               |
+//! | Measurement     | MZ, MRESETZ, RESET                         |
+//! | Other           | MOV                                        |
+//! ```
+//!
+//! # Not Supported (Panics)
+//!
+//! `T`, `T_ADJ`, `Rx`, `Ry`, `Rz`, `Rxx`, `Ryy`, `Rzz` (non-Clifford gates)
+//!
+//! # Gate Properties
```
DmitryVasilevsky, line 33: Testing various equivalences starting from 0 state is a very useful addition to test!
One thing feels missing - more basic test. I think for primitive gates we should have tests that check them on all basis vectors. I.e. If we have a one-qubit gate, we check it on |0> and |1>. If we have two-qubit gate, we check it on |00>, |01>, |01>, |11>. If we had a way to set the state before the gate and check the state after the gate, these tests could be as simple as going over a table of each state for each gate.
I'm not sure if we can access or even keep state in different simulators. And this doesn't have to be part of this PR - something to think about when checking basic functionality.

orpuente-MS: This is a really good idea. What if we add logic to `check_programs_are_eq!` to make it so that it runs tests starting from all possible basis vectors instead of just starting from zero? This is not perfect, but it's better than just checking from zero. For example, it would make the test checking that `I` is equivalent to `MResetZ` fail, which is good, since that test should be written in another way.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noiseless.rs
```diff
@@ -0,0 +1,929 @@
...
+//! | Two-qubit         | CX, CY, CZ, SWAP                           |
+//! | Three-qubit       | CCX                                        |
+//! | Rotation          | Rx, Ry, Rz, Rxx, Ryy, Rzz                  |
+//! | Measurement       | M, MZ, MRESETZ, RESET                      |
+//! | Other             | MOV                                        |
+//! ```
+//!
+//! # Gate Properties
+//!
+//! The `~` symbol denotes equivalence up to global phase.
```
DmitryVasilevsky, line 34: Same note on equivalence as in the other file.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+        },
+        format: histogram,
+        output: expect![[r#"
+                    0: 97
+                    1: 903"#]],
+    }
+}
+
+#[test]
+fn x_noise_on_h_gate_affects_superposition() {
```
DmitryVasilevsky: X noise on H gate does affect the state but does not affect the outcome of this program. No matter what X noise you apply, the probability distribution should be 50/50. Test is fine, rename maybe?

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+            cz(0, 1);
+            mresetz(0, 0);
+            mresetz(1, 1);
+        },
+        num_qubits: 2,
+        num_results: 2,
+        shots: 1000,
+        seed: SEED,
+        noise: noise_config! {
+            cz: { xi: 0.1 },
```
DmitryVasilevsky, line 319: Again, this noise doesn't affect outcome.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+            h(0);
+            x(0);
+            mresetz(0, 0);
+        },
+        num_qubits: 1,
+        num_results: 1,
+        shots: 1000,
+        seed: SEED,
+        noise: noise_config! {
+            x: { x: 0.2 },
```
DmitryVasilevsky: Again, noise doesn't affect outcome.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+                    0: 506
+                    1: 494"#]],
+    }
+}
+
+// ==================== Multiple Gates / Accumulated Noise Tests ====================
+
+#[test]
+fn noise_accumulates_across_multiple_gates() {
+    // Two X gates, each with noise - errors compound
```
DmitryVasilevsky, line 383: The interesting part here is to check the probabilities more precisely. The probability of 0 should be 0.9^2+0.1^2=0.82 and the probability of 1 should be 0.9*0.1*2=0.18. The only way to test this reliably is to run this test manually without a seed and with very large number of shots. We cannot really have automatic seedless tests with large number of shots so we can only check that results are reproducible with the seed and small number of shots.
I think this should have a comment on what to do if the test fails due to some change in the simulator. Something like - if updating expected value ensure that the probability of 0 is ... and 1 is ...

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+            cx(0, 1);
+            mresetz(0, 0);
+            mresetz(1, 1);
+        },
+        num_qubits: 2,
+        num_results: 2,
+        shots: 1000,
+        seed: SEED,
+        noise: noise_config! {
+            h: { x: 0.02 },
```
DmitryVasilevsky: X noise on H doesn't affect this program.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+            cx: { xi: 0.02, ix: 0.02 },
+        },
+        format: top_n(4),
```
DmitryVasilevsky: There could be only 4 outcomes, right?

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+        output: expect![[r#"
+                    00: 491
```
DmitryVasilevsky: 00: 0.5 * 0.96
11: 0.5 * 0.96
01: 0.5 * 0.02 * 2
10: 0.5 * 0.02 * 2

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
```diff
@@ -0,0 +1,687 @@
...
+            mresetz(0, 0);
+            mresetz(1, 1);
+        },
+        num_qubits: 2,
+        num_results: 2,
+        shots: 1000,
+        seed: SEED,
+        noise: noise_config! {
+            intrinsics: {
+                0: { x: 0.2 },
```
DmitryVasilevsky: I suggest significantly different noise probability. Like 0.1 and 0.5.

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
```diff
@@ -0,0 +1,703 @@
...
+
+    // Match within { } apply { } at the end (no trailing semicolon or more instructions)
+    (@accum [$($acc:expr),*] within { $($within_tt:tt)* } apply { $($apply_tt:tt)* } $($rest:tt)*) => {{
+        let mut result: Vec<QirInstruction> = vec![$($acc),*];
+        result.extend(qir!($($within_tt)*));  // forward within
+        result.extend(qir!($($apply_tt)*));   // apply
+        let within_rev: Vec<QirInstruction> = {
+            let mut v = qir!($($within_tt)*); // expand tokens again for reverse
+            v.reverse();
+            v
```
DmitryVasilevsky, line 406: This just reverses the gate order? This looks awfully dangerous. The resulting language looks similar to Q# and one would expect inverse of gates. Even if now all the gates are self-inverse (I haven't checked) I suggest you don't call it "within-apply". Or add a hard explicit check that all gates here are self-inverse.

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
```diff
@@ -0,0 +1,703 @@
...
+///     num_qubits: 8,
+///     num_results: 2,
+///     shots: 100,
+///     seed: 42,
+///     noise: noise_config! { ... },
+///     format: histogram,
+///     output: expect![[r#"..."#]],
+/// }
+/// ```
+macro_rules! check_sim {
```
DmitryVasilevsky, line 477: Do you really need a macro for this? I'm a bit concerned that macro here is just and additional complexity. Could it be a function? What would be less convenient if it becomes a function?

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
```diff
@@ -0,0 +1,703 @@
...
+/// check_programs_are_eq! {
+///     simulator: NoiselessSimulator,
+///     programs: [
+///         qir! { i(0) },
+///         qir! { x(0); x(0); }
+///     ],
+///     num_qubits: 1,
+/// }
+/// ```
+macro_rules! check_programs_are_eq {
```
DmitryVasilevsky, line 631: Also this one. What improves when this is a macro?

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
orpuente-MS: Thanks for the feedback! I totally missed this detail. I can change this to verify that the instructions are indeed adjointable and use their adjoint. This should be simple to add.

source/pip/src/qir_simulation/cpu_simulators/tests/clifford_noiseless.rs
DmitryVasilevsky: It feels like existing way of writing tests would be still heavy for these basic tests. I think just something like a table should be more appropriate. Gate, Input state -> result state. That way it will be all on one page easy to grasp. And a function that will go over this table and execute checks.

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
orpuente-MS: That is a first concern. In this case, the macro is doing two things:
1. First, they buy us named, optional parameters. Rust doesn't have default arguments or named parameters for functions. If this were a function, every test would have multiple arguments set to `None`, and it would be harder to tell what arguments are.
 
In tests that quickly hurts readability. For example:
```rust
check_sim<StabilizerSimulator>(
    qir! {
        x(0);
        mresetz(0, 0);
    },
    1,
    1,
    1,
    None,
    None,
    raw,
    expect![[r#"1"#]],
)
```

Compare that to the macro version:
```rust
check_sim! {
    simulator: StabilizerSimulator,
    program: qir! {
        x(0);
        mresetz(0, 0);
    },
    num_qubits: 1,
    num_results: 1,
    output: expect![[r#"1"#]],
}
```

With the macro the test is shorter and easier to read / understand, specially in platforms like GitHub without lang-server support.
This matters because we will have hundreds of tests.

2. Hiding syntax, semantics, and machinery that isn't the point of the test. I feel this will become more evident when I try to integrate the GPU simulator in the test-suite. The GPU simulator is quite different from the CPU ones. Trying to work around the type system to accomodate an API for the test suite that works for both could be possible (I am not sure yet), but if it is, it will require more work than just having a macro. And if we pulled it off, the test suite will likely end with syntax and semantics that are concerned with Rust's generic type system, instead of minimal syntax and semantics that are concerned with what is being tested.

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
orpuente-MS: See my answer to the previous question.

source/noisy_simulator/src/state_vector_simulator.rs
orpuente-MS: You are right, I renormilize and even return a numerical error if this happens. So this code is actually unreachable.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
orpuente-MS: I changed the H gate from `X` to `loss`. So, these numbers changed. I also incremented the number of shots to 100k to get more accurate numbers in this case.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
orpuente-MS: I increased the shots to 100k here; that should give us a good approximation. I would need 10M shots for the result to be stable on 4 significant digits (like is needed to show a stable percent with 2 digits after the decimal point), but that would make the test take forever to run.

source/pip/src/qir_simulation/cpu_simulators/tests/test_utils.rs
DmitryVasilevsky: Personally, I think that macros have been added to the language to cover the useful functionality that happened to be too ugly to be in the main language. So, I prefer to avoid them completely, if possible. And if they are unavoidable, I would use them strictly for the part that's not covered by the main language. In this case, if default values are what we are after, we just use macros strictly for default values and call the main function where all other logic is implemented.
Anyway, I realize other people don't agree with me so it's up to you. If you prefer to have it as a macro, that's fine. It does make tests look better.

source/pip/src/qir_simulation/cpu_simulators/tests/full_state_noisy.rs
DmitryVasilevsky: I didn't mean to increase number of shots to super high number. Another way would be to keep number of shots small (say 1000) but check the range on the output probabilities. I.e. 0.82 +- 0.05.
I'm not sure if this is easy with the 'expect' approach though. Anyway, if 100k shot test is still fast and gives enough precision, maybe it's fine.

DmitryVasilevsky (approved):

PR #2915, opened 2026-02-03
Add QIR noise intrinsic

orpuente-MS:
This PR:
1. Adds a "qdk_noise" attribute to QIR.
2. Adds a `@NoiseIntrinsic()` attribute to Q# and a `@qdk.qir.noise_intrinsic` attribute to the OpenQASM compiler, which map to the new `"qdk_noise"` attribute during codegen.
3. Patches the experimental `run_qir` API so that it can run noiseless simulations on QIR that has noise intrinsics.

### DIFF
--- a/source/compiler/qsc_passes/src/noise_intrinsic/tests.rs
+++ b/source/compiler/qsc_passes/src/noise_intrinsic/tests.rs
@@ -24,6 +24,41 @@ fn check(file: &str, expect: &Expect) {
 }
 
 #[test]
+fn test_noise_intrinsic_with_body_intrinsic_works() {
+    check(
+        indoc! {r#"
+        namespace Test {
+            @NoiseIntrinsic()
+            operation Foo(q: Qubit) : Unit {
+                body intrinsic;
+            }
+        }
+    "#},
+        &expect![[r#"
+            []
+        "#]],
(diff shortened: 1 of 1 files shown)


source/compiler/qsc_passes/src/noise_intrinsic/tests.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::noise_intrinsic::validate_noise_intrinsic_declarations;
+use expect_test::{Expect, expect};
+use indoc::indoc;
+use qsc_data_structures::{
+    language_features::LanguageFeatures, source::SourceMap, target::TargetCapabilityFlags,
+};
+use qsc_frontend::compile::{self, PackageStore, compile};
+
+fn check(file: &str, expect: &Expect) {
```
swernli, line 12: This file validates the two error cases, but might be worth ensuring the positive cases work too (`body intrinsic;` and `@SimulatableIntrinsic()`).

orpuente-MS: Oh, those positive tests are already in the codegen code. But it doesn't hurt to add them here as well. Added in the last commit.

PR #2916, opened 2026-02-03
ISA queries, trace queries, and enumeration.

msoeken:
This introduces ISA queries and trace queries in the Python API. With that one can implement a preliminary implementation for the central `estimate` function which can enumerate traces and ISA using these queries and perform estimation. It also moves some of the preliminary models from the test code to proper places in the API (`qsharp.qre.models`) with `AQREGateBased` as first architecture and `SurfaceCode` as first QEC model. Further it defines the application base class with a `QSharpApplication` as first implementation of it.


source/pip/qsharp/qre/models/qubits/_aqre.py
brad-lackey, line 1: This would be a good candidate for a ONE_QUBIT_UNITARY instruction, from which we can coercise, H, S, T, etc. Would PAULI_I coerce to a ONE_QUBIT_UNITARY instruction? In other words, should NOOP (or IDLE) be an instruction?

source/pip/qsharp/qre/_application.py
```diff
@@ -0,0 +1,139 @@
...
+    An application defines a class of quantum computation problems along with a
+    method to generate traces for specific problem instances.
+
+    We distinguish between application and trace parameters.  The application
+    parameters define which particular instance of the application we want to
+    consider.  The trace parameters define how to generate a trace.  They
+    change the specific way in which we solve the problem, but not the problem
+    itself.
+
+    For example, in quantum cryptography, the application parameters could
```
brad-lackey, line 46: "...quantum cryptanalysis..." (quantum cryptography is a different subject altogether).

source/pip/qsharp/qre/_trace.py
```diff
@@ -0,0 +1,89 @@
...
+    @classmethod
+    def q(cls, **kwargs) -> TraceQuery:
+        return TraceQuery(cls, **kwargs)
+
+
+@dataclass
+class PSSPC(TraceTransform):
+    _: KW_ONLY
+    num_ts_per_rotation: int = field(
+        default=10, metadata={"domain": list(range(1, 21))}
```
brad-lackey, line 27: Perhaps we should make default=20. Rationale: 10 T's per rotation doesn't even reach the fidelity needed for toy problems. So to be conservative, if the user does not fix this it should work for any full scale problem.

source/pip/qsharp/qre/instruction_ids.pyi
```diff
@@ -0,0 +1,92 @@
...
+RX: int
+RY: int
+RZ: int
+CRX: int
+CRY: int
+CRZ: int
+RXX: int
+RYY: int
+RZZ: int
+
```
brad-lackey, line 77: Generic ONE_QUBIT_UNITARY and TWO_QUBIT_UNITARY that live in between ONE_QUBIT_CLIFFORD and GENERIC. Namely some architectures don't care which one qubit gate is done, they are all basically the same.

source/pip/qsharp/qre/_trace.py
msoeken: Sounds good. In fact, the default values do not play a role when we enumerate over transforms in trace queries. Then the `domain` is used as the default value. But the default value matters when we construct a `PSSPC` transform manually to apply it to some trace.

PR #2917, opened 2026-02-03
Ket Labels on Results from Azure

ScottCarda-MS:
Job results from some hardware targets used a different representation for results than our simulators. This PR makes the ket-label-building logic on histograms more general to account for integer result representations.

### DIFF
--- a/source/npm/qsharp/ux/histogram.tsx
+++ b/source/npm/qsharp/ux/histogram.tsx
@@ -67,14 +67,14 @@ function getDefaultMenuSelection(
   return selection;
 }
 
-const reKetResult = /^\[(?:(Zero|One|Loss|0|1|-), *)*(Zero|One|Loss|0|1|-)\]$/;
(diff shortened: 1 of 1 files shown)


source/npm/qsharp/ux/histogram.tsx
```diff
@@ -67,17 +67,23 @@ function getDefaultMenuSelection(
   return selection;
 }
 
-const reKetResult = /^\[(?:(Zero|One|Loss), *)*(Zero|One|Loss)\]$/;
+const reKetResult = /^\[(?:(Zero|One|Loss|0|1|-), *)*(Zero|One|Loss|0|1|-)\]$/;
```
billti: You need to handle the digit 2 also. That's what the JSON results currently contain from Atom runs.

ScottCarda-MS: Done

PR #2920, opened 2026-02-05
Change `import_openqasm` to hoist qubits into the arguments of the generated Q# operation

swernli:
The purpose of the `import_openqasm` function in Python is to take an OpenQASM program and generate the corresponding Q# operation. However, the existing implementation only takes the classical parameters and handles all qubit allocation internally. Instead, this changes the behavior to hoist/lift qubits declared in the OpenQASM into parameters of the generated operation. This makes it easier to use the resulting operation as a subset of a larger Q# program. Note: this is a breaking change, technically, but does seem to better match the intent of the API. Functionality similar to the old one can still be achieved by using `ProgramType.File` to create an operation that represents the entire program.

Additionally, this updates the `ProgramType.File` import to avoid an issue where the code was run once as a side-effect of the import. Critically, this does NOT change the behavior of APIs other than `import_openqasm` that use `ProgramType.File` so that their behavior remains unchanged.

This change also includes an update to how qubit borrowing works to enable Qiskit-style qubit semantics without the side effect of leaking qubits. With this update, qubits that are allocated with `borrow` statements are not checked in simulation to be in the zero state on release (the simulator still resets them as part of releasing to avoid issues on later allocation). This allows the `QubitSematics::Qiskit` option to use `borrow` statements rather than manual qubit allocation, still get the desired behavior of ignoring qubit state on release, and avoid qubit leaks when invoked from Q#. This has no effect on other features like circuit display, resource estimation, or QIR codegen as the `borrow` still triggers qubit ID allocation as normal.

### DIFF
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -1346,7 +1346,6 @@ impl State {
                 let is_zero = sim.qubit_release(qubit.0, &call_stack);
                 let is_borrowed = self.dirty_qubits.remove(&qubit.0);
                 if is_zero || is_borrowed {
-                    self.dirty_qubits.remove(&qubit.0);
                     Value::unit()
                 } else {
                     return Err(Error::ReleasedQubitNotZero(qubit.0, arg_span));


swernli:
Converting to draft while I rething the `ProgramType.File` approach.

source/vscode/resources/chat-instructions/openqasm.instructions.md
```diff
@@ -35,7 +35,7 @@ When generating OpenQASM 2.0 file, start with the version declaration: `OPENQASM
 
 The `qsharp` package provides a way to compile, run, and resource estimate OpenQASM 3.0 programs in Python. The functions can be imported from the `qsharp.openqasm` module and are named `compile`, `run`, and `estimate`.
 
-OpenQASM 3.0 programs can be imported into Python using the `qsharp.openqasm.import_openqasm` function and accessed via the `qsharp.code` module: `qsharp.openqasm.import_openqasm("qubit q; bit c; c = measure q;", name="Foo"); qsharp.code.Foo()`.
+OpenQASM 3.0 programs can be imported into Python using the `qsharp.openqasm.import_openqasm` function and accessed via the `qsharp.code` module: `qsharp.openqasm.import_openqasm("qubit q; bit c; c = measure q;", name="Foo", program_type=qsharp.openqasm.ProgramType.File); qsharp.code.Foo()`.
```
idavis: Should we have anther example that uses ProgramType.Operation?

source/pip/tests/test_qasm.py
```diff
@@ -654,7 +660,7 @@ def test_circuit_from_callable_with_args() -> None:
         """,
         name="Foo",
     )
-    c = circuit(code.Foo, 2)
+    c = qsharp_circuit("{ use qs = Qubit[2]; Foo(2, qs); }")
```
idavis, line 714: I think we could use a few more test cases where we define qubits and qubit registers multiple times.

source/vscode/resources/chat-instructions/openqasm.instructions.md
swernli: Makes sense, I'll add one.

source/pip/tests/test_qasm.py
swernli: Added a couple more.

source/compiler/qsc_eval/src/lib.rs
```diff
@@ -1344,12 +1344,8 @@ impl State {
                     .ok_or(Error::QubitDoubleRelease(arg_span))?;
                 env.release_qubit(&qubit);
                 let is_zero = sim.qubit_release(qubit.0, &call_stack);
-                if is_zero {
-                    self.dirty_qubits.remove(&qubit.0);
-                    Value::unit()
-                } else if self.dirty_qubits.contains(&qubit.0) {
-                    // If the qubit is dirty, it means it was borrowed rather than allocated.
-                    // In this case, we consider it released regardless of state and don't error.
+                let is_borrowed = self.dirty_qubits.remove(&qubit.0);
```
idavis, line 1347: calling remove twice

swernli: whoops, removed the remove.

PR #2923, opened 2026-02-07
New ISA and ISARequirements function

msoeken:
This PR provides some support for upcoming models by providing new features to ISAs and the Rust bindings in general
- ISA instructions can have properties (e.g., the `LATTICE_SURGERY` instruction can have a distance property if it was generated by a surface code)
- ISA requirement constraints can check if some properties are set, e.g., one can require that the `LATTICE_SURGERY` instruction also provides a `distance` property.
- Variable arity instructions can provide space, time, and error_rate functions as generic Python functions. This can be slower than the built-in functions `const`, `linear`, and `block_linear`, but useful for cases that do not have an easy structure. The new function is called `generic` (`generic_function` in the API)
- We expose a `binom_ppf` functions that works like `scipy.binom.ppf` and uses the `probability` crate that is already a dependency. This is not only faster but also does not require an additional dependency. It will be used to model round-based distillation factories.

I missed to address some comments in the last PR. This is now done here.

### DIFF
--- a/source/pip/benchmarks/bench_qre.py
+++ b/source/pip/benchmarks/bench_qre.py
@@ -3,6 +3,7 @@
 
 import timeit
 from dataclasses import dataclass, KW_ONLY, field
+from qsharp.qre import linear_function, generic_function, instruction
 from qsharp.qre.models import AQREGateBased, SurfaceCode
 from qsharp.qre._enumeration import _enumerate_instances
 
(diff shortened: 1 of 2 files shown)


source/pip/qsharp/qre/_instruction.py
```diff
@@ -82,6 +106,7 @@ def instruction(
     space: Optional[int] | IntFunction = None,
     length: Optional[int | IntFunction] = None,
     error_rate: float | FloatFunction,
+    **kwargs: int,
```
brad-lackey, line 109: Should we restrict to int-valued pairs? Even for distance, I would think we would want Optional[int]. Namely, we want to use DISTANCE, but it might not yet be defined (or is irrelevant) and so is None-valued.

source/qre/src/utils.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use probability::prelude::{Binomial, Inverse};
+
+#[allow(clippy::doc_markdown)]
+/// Faster implementation of SciPy's binom.ppf
+#[must_use]
+pub fn binom_ppf(q: f64, n: usize, p: f64) -> usize {
```
brad-lackey, line 9: Hmmm.... No issues here. I'd just never heard of a ppf. I had to look it up. I've always called the inverse cumulative distribution the "inverse cumulative distribution." Guess all those years of probability in grad school failed me.

source/pip/qsharp/qre/_instruction.py
msoeken: I think in this case we just do not specify it. The purpose is that an `ISARequirement` of an ISA transform can ask for an instruction to provide some property, e.g., distance. That will guarantee that if we call `inst.get_property(DISTANCE)` on an instruction we are provided with that distance value. If the instruction cannot provide that information, the ISA is discarded from being a potential implementation ISA for that ISA transform.

source/qre/src/utils.rs
msoeken: I also see ICD a lot more, just thought of making it look more like binom.ppf which I used as reference implementation and I know that it produces the same numbers as that one.

PR #2928, opened 2026-02-09
Add `load_csv_dir` method to `NoiseConfig` class

orpuente-MS:
This PR adds a `load_csv_dir` method to the Python `NoiseConfig` class, used to load correlated noise tables into a `NoiseConfig` object. It also changes all of our simulator layers to store Pauli strings as `u64` instead of `String`, to avoid allocating heap memory. This resulted in a significant speed up of both loading noise and running the simulations. (I saw an 8x speed up for both scenarios on my laptop).

### DIFF
--- a/source/pip/tests/test_correlated_noise.py
+++ b/source/pip/tests/test_correlated_noise.py
@@ -65,6 +65,15 @@ def test_noisy_simulation():
         assert output == [[Result.One, Result.Zero, Result.One]]
 
 
+@pytest.mark.skipif(not GPU_AVAILABLE, reason=SKIP_REASON)
+def test_noisy_simulation_gpu():
+    noise = NoiseConfig()
+    table = noise.intrinsic("test_noise_intrinsic", 3)
+    table.yyy = 1.0
+    output = run_qir(QIR_WITH_CORRELATED_NOISE, shots=1, noise=noise, type="gpu")
+    assert output == [[Result.One, Result.Zero, Result.One]]
+
+
 def test_load_csv_dir():
     noise = NoiseConfig()
     noise.load_csv_dir("./csv_dir_test")
@@ -74,7 +83,7 @@ def test_load_csv_dir():
 
 
 @pytest.mark.skipif(not GPU_AVAILABLE, reason=SKIP_REASON)
-def test_noisy_simulation_gpu():
+def test_load_csv_dir_gpu():
     noise = NoiseConfig()
     noise.load_csv_dir("./csv_dir_test")
     output = run_qir(QIR_WITH_CORRELATED_NOISE, shots=1, noise=noise, type="gpu")
@@ -82,7 +91,11 @@ def test_noisy_simulation_gpu():
 
 
 def test_noisy_simulation_with_missing_gates_fails():
+    """
+    This failure happens before the list of QIR instructions makes it to
+    any specific simulator. So, we don't need separate tests for all of them.
+    """
     noise = NoiseConfig()
     with pytest.raises(ValueError) as excinfo:
-        run_qir(QIR_WITH_CORRELATED_NOISE, shots=1, noise=noise)
+        run_qir(QIR_WITH_CORRELATED_NOISE, shots=1, noise=noise, type="cpu")
     assert "Missing noise intrinsic: test_noise_intrinsic" in str(excinfo.value)


source/pip/src/qir_simulation/correlated_noise.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        // Validate characters, and encode to u64 in one pass.
+        let pauli_bytes = pauli.as_bytes();
+        let mut key: u64 = 0;
+        for &b in pauli_bytes {
+            let bits = match b {
+                b'I' => 0u64,
+                b'X' => 1u64,
+                b'Y' => 2u64,
+                b'Z' => 3u64,
```
billti: This doesn't look right. If you look in simulator.wgsl around line 183 you'll see it expect a bit for bit-flip, a bit for phase-flip, or both for bit-flip and phase-flip. Thus 'Y' is 3u64, not 'Z'.

source/simulators/src/noise_config.rs
```diff
@@ -335,6 +343,39 @@ impl<F: Fault + Clone> CorrelatedNoiseSampler<F> {
...
 /// Checks if a pauli string is the identity.
 #[must_use]
-pub fn is_pauli_identity(pauli_string: &str) -> bool {
-    pauli_string.chars().all(|c| c == 'I')
+pub fn is_pauli_identity(pauli_string: u64) -> bool {
+    pauli_string == 0
+}
+
+/// Encode a validated Pauli string as a `u64` using 2 bits per character
+/// (I=0, X=1, Y=2, Z=3). Supports up to 32 qubits.
```
billti: See prior comment. We should ensure we're consistent across the codebase on which bits mean what (bit-flip, phase-flip, etc.). I believe I wrote code under the assumption bit0 = bitflip and bit1 = phaseflip, thus X=1, Z=2, Y=3.

source/pip/src/qir_simulation/correlated_noise.rs
orpuente-MS: I actually had it how you are suggesting initially. But I can change it back for consistency.

Longer explanation:
I'm pretty sure these are completely disconnected from the numbers you use internally in the GPU. Because before my change, these values made it to the GPU as strings. These values always get decoded into something else before we use them in computation, see the change in `gpu_full_state_simulator::noise_mapping.rs` for the code where these values get used. Also, since all the tests passed after this, I assumed it was right, unless I'm missing something.

source/pip/src/qir_simulation/correlated_noise.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let bytes = contents.as_bytes();
+    let num_threads = rayon::current_num_threads().max(1);
+
+    // For small inputs, avoid parallelism overhead.
+    if contents.len() < 128 * 1024 || num_threads <= 1 {
+        let (entries, qubits) = parse_noise_chunk(contents, 0)?;
+        let pauli_noise = FxHashMap::from_iter(entries);
+        return Ok(NoiseTable {
+            qubits: qubits.unwrap_or(0),
```
billti, line 85: Why default to 0 here? Shouldn't it just panic (or report an error) if qubits isn't a valid value?

source/pip/src/qir_simulation/correlated_noise.rs
```diff
@@ -0,0 +1,255 @@
...
+            pauli_noise,
+            loss: 0.0,
+        });
+    }
+
+    // Split the buffer into roughly equal chunks at line boundaries.
+    let chunk_size = contents.len() / num_threads;
+    let mut boundaries = vec![0usize];
+    for i in 1..num_threads {
```
billti, line 94: Shouldn't this be an inclusive range (`1..=num_threads`) if you want a group for each range. (Or is that why you have the extra push on line 103 to ensure the final range is added? If so, add a comment to explain that to readers).

How much of a perf win was this? I appreciate trying to make it parallel, but it also adds a bit of code which would have some subtle bugs in it.

source/pip/src/qir_simulation/correlated_noise.rs
```diff
@@ -0,0 +1,255 @@
...
+                    return Err(ParseError::InconsistentLength {
+                        expected: exp,
+                        found: q,
+                    });
+                }
+                _ => (),
+            }
+        }
+        for (key, prob) in entries {
+            pauli_noise.insert(key, prob);
```
billti, line 151: So you still end up serially inserting all the entries into a hash_map anyway? I'd be interested then if doing all the work to parse chunks in parallel buys much.

orpuente-MS: Yes, I tried merging them in parallel with rayon's `flat_map_iter` but that was actually slower.

billti (approved):

source/pip/src/qir_simulation/correlated_noise.rs
billti: Let's follow up on this separately.

source/pip/src/qir_simulation/correlated_noise.rs
billti: As discussed, lets add tests in a follow up PR, or remove the parallelism if benchmarking shows it doesn't buy much.

source/pip/src/qir_simulation/correlated_noise.rs
orpuente-MS: No, if qubits=0 it means the file was empty, the NoiseTable will be empty, and it will be treated as noiseless. This is specially, useful if you are incrementally filling in the .csv files. This in the same way that you can just have a `NoiseTable(qubits=0)`

PR #2929, opened 2026-02-09
Deprecate `borrow` keyword with Lint and Code-Action

ScottCarda-MS:
Add lint for `borrow` keyword. Also add titles to lint-based code-actions.

### DIFF
--- a/source/compiler/qsc_linter/src/linter.rs
+++ b/source/compiler/qsc_linter/src/linter.rs
@@ -157,6 +157,15 @@ impl Compilation<'_> {
     }
 }
 
+/// A code action that can be applied to fix a lint.
+#[derive(Debug, Clone)]
+pub struct CodeAction {
+    /// The title for the code action.
+    pub title: String,
+    /// The suggested edits to fix the lint.
+    pub edits: Vec<(String, Span)>,
+}
+
 /// A lint emitted by the linter.
 #[derive(Debug, Clone, thiserror::Error)]
 pub struct Lint {
@@ -170,10 +179,8 @@ pub struct Lint {
     pub help: &'static str,
     /// An enum identifying this lint.
     pub kind: LintKind,
-    /// The suggested edits to fix the lint.
-    pub code_action_edits: Vec<(String, Span)>,
-    /// An optional title for the code action. If not provided, the lint message will be used as the title.
-    pub code_action_title: Option<String>,
+    /// An optional code action that the user can apply to fix the lint.
+    pub code_action: Option<CodeAction>,
 }
 
 impl std::hash::Hash for Lint {
--- a/source/compiler/qsc_linter/src/lints/ast.rs
+++ b/source/compiler/qsc_linter/src/lints/ast.rs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 use super::lint;
-use crate::linter::{Compilation, ast::declare_ast_lints};
+use crate::linter::{CodeAction, Compilation, ast::declare_ast_lints};
 use qsc_ast::ast::{
     BinOp, Block, Expr, ExprKind, Item, ItemKind, Lit, Namespace, NodeId, QubitSource, Stmt,
     StmtKind, TernOp,
@@ -74,8 +74,10 @@ impl NeedlessParens {
             buffer.push(lint!(
                 self,
                 child.span,
-                Self::get_code_action_edits(child.span),
-                Some("Remove unnecessary parentheses".to_string())
+                CodeAction {
+                    title: "Remove unnecessary parentheses".to_string(),
+                    edits: Self::get_code_action_edits(child.span)
+                }
             ));
         }
     }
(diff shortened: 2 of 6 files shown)


samples/language/QuantumMemory.qs
```diff
@@ -3,23 +3,27 @@
...
 operation Main() : Unit {
-    // Clean qubits are allocated with `use` statements.
-    // Clean qubits are guaranteed to be in a |0⟩ state upon allocation.
-    use qubit = Qubit();
-    use threeQubits = Qubit[3];
+    // Allocate a single qubit and an array of qubits.
+    use q = Qubit();
+    use qs = Qubit[3];
+
+    // If you change a qubit's state, reset it before leaving scope.
```
swernli: I don't feel like calling out reset specifically is helpful here... it may get folks in the bad habit of only using `Reset` rather than relying on other mechanisms, such as reliable uncomputation, to get the qubit back into the ground state (especially LLMs, which may take this guidance as reason to blindly add resets to every qubit). Personally I'd be more in favor of keeping the example specific to allocation syntax by showing a few more examples of that, such as tuple allocation with a mix of arrays and individual qubits, ie:
```qsharp
use (q0, qubitArray, q1) = (Qubit(), Qubit[], Qubit());
```

msoeken: Agree, we should either uncompute or measure the qubit and do something with the measurement result. This also affects the comment above `Main`.

msoeken (approved):

source/compiler/qsc_linter/src/lints/ast.rs
```diff
@@ -36,6 +37,7 @@ declare_ast_lints! {
     (DeprecatedAssignUpdateExpr, LintLevel::Allow, "deprecated use of update assignment expressions", "update assignment expressions \"a w/= b <- c\" are deprecated; consider using explicit assignment instead \"a[b] = c\""),
     (DeprecatedUpdateExpr, LintLevel::Allow, "deprecated use of update expressions", "update expressions \"a w/ b <- c\" are deprecated; consider using explicit assignment instead"),
     (AvoidNamespaceBlock, LintLevel::Allow, "avoid using explicit namespace blocks", "Q# best practice is to not use namespace blocks to enclose code; the namespace is inferred from the file path"),
+    (DeprecatedBorrow, LintLevel::Warn, "deprecated `borrow` qubit allocation", "the `borrow` keyword for qubit allocation is deprecated, use `use` instead"),
```
orpuente-MS, line 40: This is the first instance where we set the default level of a deprecation lint to `warn` instead of `allow`:

1. Should we set it to `allow` following the same logic of the other deprecation lints? That is, that we will schedule a plan at some point to do another major release, and we will turn on deprecations N months before that.

2. Or, if setting it to `warn` is acceptable right now, should we also set the other deprecation lints to `warn` as well?

swernli: I think this particular language feature is so infrequently used (most folks aren't even aware it exists) that it's worth leaving it as warning without changing any other existing lints.

source/compiler/qsc_linter/src/linter.rs
```diff
@@ -172,6 +172,8 @@ pub struct Lint {
     pub kind: LintKind,
     /// The suggested edits to fix the lint.
     pub code_action_edits: Vec<(String, Span)>,
+    /// An optional title for the code action. If not provided, the lint message will be used as the title.
+    pub code_action_title: Option<String>,
```
orpuente-MS: I was just thinking that it doesn't make sense to have a `code_action_title` without `code_action_edits`. Could we enforce this with a new type so that now the Lint struct looks like:
```rust
pub struct Lint {
    pub span: Span,
    pub level: LintLevel,
    pub message: &'static str,
    pub help: &'static str,
    pub kind: LintKind,
    /// A code action to fix the lint.
    pub code_action: Option<CodeAction>,
}
```

and a `CodeAction` looks like

```rust
struct CodeAction {
    /// A title for the code action.
    title: String,
    /// The suggested edits to fix the lint.
    edits: Vec<(String, Span)>,
}
```

In that way we enforce that if we have a `CodeAction`, we must provide a good title for it.

samples/language/QuantumMemory.qs
ScottCarda-MS: That makes sense. @swernli Any other examples of allocation to include? I'm struggling to come up with some more.

ScottCarda-MS: Was thinking of nested tuple destructuring, but I think that leans more towards a demo of tuples than it does qubit allocation at that point.

source/compiler/qsc_linter/src/linter.rs
ScottCarda-MS: Great idea! Done!

samples/language/QuantumMemory.qs
ScottCarda-MS: Refactored to focus on allocation syntax.

PR #2871, opened 2026-01-21
adjust classical wire spacing

minestarks:
**AFTER**

image

**BEFORE**

image

PR #2931, opened 2026-02-10
Adjust Line Spacing in Circuit Visualizations

ScottCarda-MS:
This addresses two issues:
1. A bug had caused circuits drawn in the circuit editor to assume spacing for a classical line on qubit line with no measurements. This was caused by an unresolved `undefined` value being used to specify the length of an array.
2. The spacing changes made in #2871 caused circuits to have more space than necessary between lines. There should, by default, only be 12px (two gate paddings worth) between each gate both vertically between lines and horizontally between adjacent gates on the same line. The spacing change to accommodate boxes being drawn on classical lines looked like it wasn't quite enough visually. This makes the spacing rules for classical lines consistent with qubit lines now, because classical lines can now also have boxes drawn on them.

**Before**
image

**After**
image


source/npm/qsharp/ux/circuit-vis/formatters/inputFormatter.ts
```diff
@@ -111,15 +115,17 @@ const formatInputs = (
     // Increment current height by classical register height for attached classical registers
 
     // Add classical wires
-    registers[id].children = Array.from(Array(numResults), () => {
-      currY += classicalRegHeight;
+    registers[id].children = Array.from(Array(numResults ?? 0), () => {
+      // currY += classicalRegHeight;
```
minestarks:
```suggestion
```

PR #2937, opened 2026-02-11
Elide lambdas in circuit builder when the wrap a single item

swernli:
This change updates the way we build circuits such that, like some loops, lambdas that wrap a single "box" are elided in favor of just showing that box. This makes the circuit have less "trivial" nesting and makes display of circuits for certain Q# patterns nicer. 

The change also includes a minor update to the build script so that the notebook integration tests work correctly when run locally without a Python virtual environment.

Before:
image

After:
image

New circuit snapshot test case shows that lambdas with multiple operations still produce a "box:"
image


source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -519,41 +523,56 @@ fn collapse_unnecessary_loop_scopes(operations: &mut Vec<OperationOrGroup>) {
...
+            && children.len() == 1
+            && source_lookup
+                .resolve_scope(scope_stack.current_lexical_scope())
+                .name
+                .as_ref()
+                == "<lambda>"
+        {
+            // remove the lambda scope
+            let only_child = children.remove(0);
+            return Some(vec![only_child]);
```
minestarks: Can you return `Some(take(children))` maybe? Just for aesthetic reasons.

source/npm/qsharp/test/circuits-cases/lambda_multi_op.qs
```diff
@@ -0,0 +1,6 @@
+operation Main() : Result[] {
+    use q = Qubit();
+    let lambda = (q => { H(q); S(q); });
+    lambda(q);
+    [MResetZ(q)]
```
minestarks: Maybe this can be added to the existing lambda.qs? I slightly prefer if we have multiple similar operation in one test file, rather than have a separate test file for each operation. Within reason. Don't quote me on this. I may change my mind.

build.py
minestarks, line 1: Intentional?

source/npm/qsharp/test/circuits-cases/lambda_multi_op.qs
swernli: I'll merge them into a single test case :)

source/compiler/qsc_circuit/src/builder.rs
swernli: Updated. Also, something hilarious/weird about code that says, "some take children."

build.py
swernli: Yes. I tried to add it to the PR description, but the newlines got eaten and the callout was lost. Local integration tests have been failing for me for a while due to this, which seems to be because I don't use Python virtual environments.

PR #2940, opened 2026-02-12
Allow Passing of Q# callables and closures in Python

swernli:
This change updates the Python interop to support passing of Q# values representing callables and closures as arguments to invocations of Q#. The change also allows these values to be passed directly as "entry_expr" to QIR compilation, circuit generation, and resource estimation related APIs.

### DIFF
--- a/source/pip/tests/test_callable_passing.py
+++ b/source/pip/tests/test_callable_passing.py
@@ -67,6 +67,24 @@ def test_run_qsharp_callable_passed_to_python_callable() -> None:
     assert res == 6
 
 
+def test_python_callable_with_unsupported_types_passed_to_python_callable() -> None:
+    qsharp.init()
+    qsharp.eval("""
+        function MakeRange() : Range {
+            1..10
+        }
+        function SumRangeFromMaker(maker : Unit -> Range) : Int {
+            mutable sum = 0;
+            for v in maker() {
+                sum += v;
+            }
+            sum
+        }
+    """)
+    from qsharp.code import MakeRange, SumRangeFromMaker
+    assert SumRangeFromMaker(MakeRange) == 55
+
+
 def test_qsharp_closure_from_python_callable_passed_to_python_callable() -> None:
     qsharp.init()
     qsharp.eval("""


source/pip/tests/test_callable_passing.py
```diff
@@ -0,0 +1,269 @@
...
+        }
+        function AddOne(x : Int) : Int {
+            x + 1
+        }
+    """)
+    from qsharp.code import InvokeWithFive, AddOne
+    assert InvokeWithFive(AddOne) == 6
+
+
+def test_python_callable_passed_to_qsharp_callable() -> None:
```
orpuente-MS, line 23: I think we are missing a test here for completeness. You have:
```
test_python_callable_passed_to_python_callable
test_python_callable_passed_to_qsharp_callable
test_run_qsharp_callable_passed_to_qsharp_callable
```

Would be good to also have `test_run_qsharp_callable_passed_to_python_callable`.

source/pip/qsharp/_qsharp.py
```diff
@@ -66,10 +67,16 @@ def lower_python_obj(obj: object, visited: Optional[Set[object]] = None) -> Any:
     if isinstance(obj, dict):
         return {name: lower_python_obj(val, visited) for name, val in obj.items()}
 
+    # Recursive case: Callable or Closure
+    if hasattr(obj, "__global_callable"):
+        return obj.__getattribute__("__global_callable")
+    if isinstance(obj, (GlobalCallable, Closure)):
+        return obj
+
```
orpuente-MS: Nitpick on the comment: I think this would be one of the base cases, because we are not recursively calling `lower_python_obj` here.

source/pip/qsharp/_qsharp.py
```diff
@@ -66,10 +67,16 @@ def lower_python_obj(obj: object, visited: Optional[Set[object]] = None) -> Any:
...
     # Recursive case: Class with slots
     if hasattr(obj, "__slots__"):
         fields = {}
-        for name in obj.__slots__:
+        for name in obj.__getattribute__("__slots__"):
```
orpuente-MS: Do we need this change? The distinction between dot access, `__getattr__`, and `__getattribute__` is non-trivial, so it would be good to have a comment here if we do need one of the other variants.

source/pip/qsharp/_qsharp.py
swernli: Good point!

source/pip/qsharp/_qsharp.py
swernli: Changed it to `getattr` to match the check above. The reason for the change was to make Pylance checks happy. Since I was already editting this file and updating some types and handling, I worked on making it warning clean.

source/pip/tests/test_callable_passing.py
swernli: Added!

source/pip/qsharp/_qsharp.py
orpuente-MS: Sounds good, I have tests for `__slots__` usages, so if they pass it should be fine. Just for reference, here the link to the Python docs page that describes all the differences between all those forms of accessing attributes:

https://docs.python.org/3/howto/descriptor.html#invocation-from-an-instance

PR #2942, opened 2026-02-12
[Circuit Diagrams] 1 - RIR debug metadata

minestarks:
This is part of the groundwork for supporting the visualization of classical control in circuit diagrams.

To support classical control flow (branches) in circuit diagrams, we will use the RIR representation which we will transform into a circuit. To be able to preserve source information in the circuit diagram, we first need to _emit_ source metadata into RIR. This also sets us up nicely to potentially emit LLVM debug metadata to QIR, if ever the need arises.

- Add a `debug` module to `qsc_rir` with structures modeled after LLVM's source-level debugging metadata (`DbgLocation`, `DbgScope`, `DbgInfo`). These track source offsets, callable scopes, loop body scopes (with per-iteration discriminators), and inline-at chains.
- Rename the `Instruction` enum to `InstructionKind` and introduce a new `Instruction` struct that pairs an `InstructionKind` with optional `InstructionDbgMetadata`. All RIR passes, codegen, builder helpers, and tests updated accordingly.
- During partial evaluation, populate debug metadata by tracking call site locations, callable scopes (`SubProgram`), and loop iteration scopes (`LexicalBlockFile`). Unused entries are pruned before returning the program.
- Add `PartialEvalConfig` with a `generate_debug_metadata` flag. Enabled for `fir_to_rir` (used by the VS Code extension for circuit/RIR views), disabled for `fir_to_qir` (text QIR output).
- Fix debug location assignment for short-circuit binary ops (`&&`, `||`) — the continuation block now gets the correct source location.
- New test suite in debug_metadata.rs covering gates, measurements, loops, callable inlining, and short-circuit binary ops.

### DIFF
--- a/source/compiler/qsc_frontend/src/compile.rs
+++ b/source/compiler/qsc_frontend/src/compile.rs
@@ -174,11 +174,6 @@ impl PackageStore {
             open: id,
         }
     }
-
-    #[must_use]
-    pub fn is_empty(&self) -> bool {
-        self.units.is_empty()
-    }
 }
 
 impl<'a> IntoIterator for &'a PackageStore {
--- a/source/compiler/qsc_partial_eval/src/lib.rs
+++ b/source/compiler/qsc_partial_eval/src/lib.rs
@@ -3672,16 +3672,15 @@ impl<'a> PartialEvaluator<'a> {
             if let Some(s) = s {
                 Some(s)
             } else {
-                let (name, span) = match &self.package_store.get_item(item_id).kind {
-                    fir::ItemKind::Callable(callable_decl) => {
-                        let name = if functor_app.adjoint {
-                            format!("{}'", callable_decl.name.name).into()
-                        } else {
-                            callable_decl.name.name.clone()
-                        };
-                        (name, callable_decl.span)
-                    }
-                    _ => panic!("expected callable"),
+                let fir::ItemKind::Callable(callable_decl) =
+                    &self.package_store.get_item(item_id).kind
+                else {
+                    panic!("expected callable");
+                };
+                let name = if functor_app.adjoint {
+                    format!("{}'", callable_decl.name.name).into()
+                } else {
+                    callable_decl.name.name.clone()
                 };
                 let current_package_id = self.get_current_package_id();
                 let package_id = current_package_id.into();
@@ -3689,7 +3688,7 @@ impl<'a> PartialEvaluator<'a> {
                     name,
                     location: DbgPackageOffset {
                         package_id,
-                        offset: span.lo,
+                        offset: callable_decl.span.lo,
                     },
                 };
                 let i = self.program.dbg_info.add_scope(scope);
--- a/source/compiler/qsc_partial_eval/src/tests.rs
+++ b/source/compiler/qsc_partial_eval/src/tests.rs
@@ -128,10 +128,62 @@ pub fn get_rir_program_with_dbg_metadata(source: &str) -> Program {
         Ok(program) => {
             // Verify the program can go through transformations.
             check_and_transform(&mut program.clone());
+            validate(&program);
             program
         }
         Err(error) => panic!("partial evaluation failed: {error:?}"),
     }
+}
+
+fn validate(program: &Program) {
+    let mut dbg_scopes = program.dbg_info.dbg_scopes.clone();
+    let mut dbg_locations = program.dbg_info.dbg_locations.clone();
+
+    // All scope and inlined_at references should be to existing dbg scopes and locations.
+    for (dbg_location, _) in dbg_locations.values() {
+        assert!(dbg_scopes.contains_key(dbg_location.scope));
+        if let Some(inlined_at) = dbg_location.inlined_at {
+            assert!(dbg_locations.contains_key(inlined_at));
+        }
+    }
+
+    // All dbg location references in instructions should be to existing dbg locations.
+    for instruction in program.blocks.iter().flat_map(|(_, block)| &block.0) {
+        if let Some(dbg_location) = instruction.metadata().map(|metadata| metadata.dbg_location) {
+            assert!(dbg_locations.contains_key(dbg_location));
+        }
+    }
+
+    // Ensure all entries are referenced by removing referenced scopes/locations from the lists and then checking if any remain at the end.
+    for instruction in program.blocks.iter().flat_map(|(_, block)| &block.0) {
+        if let Some(dbg_location) = instruction.metadata().map(|metadata| metadata.dbg_location) {
+            let mut to_remove = vec![dbg_location];
+            let mut next = dbg_locations.get(dbg_location);
+
+            while let Some(entry) = next {
+                // remove referenced scope
+                dbg_scopes.remove(entry.0.scope);
+
+                if let Some(inlined_at) = entry.0.inlined_at {
+                    // collect referenced dbg locations
+                    next = dbg_locations.get(inlined_at);
+                    to_remove.push(inlined_at);
+                } else {
+                    break;
+                }
+            }
+            for id in to_remove {
+                dbg_locations.remove(id);
+            }
+        }
+    }
+
+    assert!(
+        dbg_locations.is_empty(),
+        "unreferenced entry in dbg locations"
+    );
+
+    assert!(dbg_scopes.is_empty(), "unreferenced entry in dbg scopes");
 }
 
 #[must_use]
--- a/source/compiler/qsc_data_structures/src/index_map.rs
+++ b/source/compiler/qsc_data_structures/src/index_map.rs
@@ -39,7 +39,7 @@ impl<K, V> IndexMap<K, V> {
 
     #[must_use]
     pub fn is_empty(&self) -> bool {
-        self.values.is_empty()
+        self.values.iter().all(Option::is_none)
     }
 
     // `Iter` does implement `Iterator`, but it has an additional bound on `K`.
--- a/source/compiler/qsc_rir/src/rir.rs
+++ b/source/compiler/qsc_rir/src/rir.rs
@@ -349,6 +349,16 @@ pub enum Instruction {
     Return,
 }
 
+impl Instruction {
+    #[must_use]
+    pub fn metadata(&self) -> Option<&InstructionDbgMetadata> {
+        match self {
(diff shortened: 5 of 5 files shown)


source/compiler/qsc_partial_eval/src/tests/debug_metadata.rs
```diff
@@ -0,0 +1,525 @@
...
+                Call id(1), args( Pointer, )
+                Call id(2), args( Qubit(0), ) !dbg dbg_location=2
+                Call id(3), args( Integer(0), EmptyTag, )
+                Return
+
+            dbg_metadata_scopes:
+                0 = SubProgram name=Main location=(2-40)
+                1 = SubProgram name=H location=(1-110222)
+            dbg_locations:
+                [0]: scope=0 location=(2-74)
```
minestarks: Consider trimming unused locations?

source/compiler/qsc_partial_eval/src/tests/debug_metadata.rs
```diff
@@ -0,0 +1,525 @@
...
+                [36]: scope=0 location=(2-74)
+                [37]: scope=13 location=(0-2937) inlined_at=36
+                [38]: scope=14 location=(0-2963) inlined_at=37
+                [39]: scope=15 location=(0-2963) inlined_at=37
+                [40]: scope=16 location=(0-2963) inlined_at=37"#]],
+    );
+}
+
+#[test]
+fn lambda() {
```
minestarks, line 324: What eactly am I testing here?

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -187,13 +201,15 @@ impl<'a> PartialEvaluator<'a> {
         compute_properties: &'a PackageStoreComputeProperties,
         entry: &'a ProgramEntry,
         capabilities: TargetCapabilityFlags,
+        generate_debug_metadata: bool,
```
minestarks: Can we remove this new parameter from most of the constructors (and just add a new constructor)?

source/compiler/qsc_partial_eval/src/evaluation_context.rs
```diff
@@ -85,7 +92,7 @@ impl EvaluationContext {
     }
 }
 
-/// Struct that represents a block node when we interpret an RIR program as a graph.
+/// Struct that represents a block node when we intepret an RIR program as a graph.
```
ScottCarda-MS: "intepret"?

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1772,9 +1822,13 @@ impl<'a> PartialEvaluator<'a> {
...
             .collect();
 
-        let instruction = Instruction::Call(callable_id, args_operands, output_var);
+        let instruction = InstructionKind::Call(callable_id, args_operands, output_var);
+        let metadata = self.metadata_from_current_debug_location();
         let current_block = self.get_current_rir_block_mut();
-        current_block.0.push(instruction);
+        current_block.0.push(Instruction {
+            kind: instruction,
+            metadata,
```
minestarks: This doesn't seem to be set correctly for short circuiting binary operations

```
operation Main() : Unit {

    use qm = Qubit();
    use q = Qubit();
    H(q);
    let a = {
        X(q);
        M(q) == One
    } or {
        Y(q);
        M(q) == One
    };
}
```

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2341,57 +2424,61 @@ impl<'a> PartialEvaluator<'a> {
...
                 rhs_operand,
                 bin_op_rir_variable,
             ),
             _ => panic!("unsupported binary operation for double: {bin_op:?}"),
         };
-        self.get_current_rir_block_mut().0.push(bin_op_rir_ins);
+        let metadata = self.metadata_from_current_debug_location();
+        self.get_current_rir_block_mut().0.push(Instruction {
+            kind: bin_op_rir_ins,
+            metadata,
```
minestarks: Do we really need to attach metadata to binops ? Is this accurate?

source/compiler/qsc/src/codegen.rs
```diff
@@ -103,20 +103,26 @@ pub mod qir {
...
-                    &source_package.sources,
-                    e,
-                ))]
-            })?;
+        let (raw, ssa) = fir_to_rir(
+            &fir_store,
+            capabilities,
+            Some(compute_properties),
+            &entry,
+            true,
```
minestarks: I'm making a change turning this boolean parameter (`generate_debug_metadata`) to a struct to make the callsites more readable

source/compiler/qsc_partial_eval/src/tests/debug_metadata.rs
minestarks: Making this change...

source/compiler/qsc_partial_eval/src/tests/debug_metadata.rs
minestarks: updating test

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: I decided to keep it - the change is already made and I actually prefer it when the parameters are fairly consistent across the constructors, at least when they have a generic name like `new()`

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: For our current purposes, we don't need to attach dbg metadata to any instruction except for Call and Branch. So I'll remove these.

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: making change to set debug location for the branch to be the RHs expression.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -469,6 +498,8 @@ impl<'a> PartialEvaluator<'a> {
             .try_into()
             .expect("results count should fit into a u32");
 
+        self.program.dbg_info.remove_unused_dbg_metadata();
```
swernli, line 498: I like that this filters to just the debug entries that are used... do we have anything in the tests (besides the snapshots, I guess) that verifies that every id referenced by an instruction or an "inlined-at" exists in the dbg_info? Shouldn't need it in production, but might be nice to have a utility that verifies dbg_info is well formed.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -3519,6 +3697,234 @@ impl<'a> PartialEvaluator<'a> {
...
+                .dbg_context
+                .dbg_callable_to_scope
+                .get(&(item_id, functor_app.adjoint))
+                .copied();
+
+            if let Some(s) = s {
+                Some(s)
+            } else {
+                let (name, span) = match &self.package_store.get_item(item_id).kind {
+                    fir::ItemKind::Callable(callable_decl) => {
```
swernli: nitpick: since this match just panics for all other cases, it could be a let-else, but I'm not sure that makes the code much better.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -3519,6 +3697,234 @@ impl<'a> PartialEvaluator<'a> {
...
+                        let name = if functor_app.adjoint {
+                            format!("{}'", callable_decl.name.name).into()
+                        } else {
```
swernli: This does something special for adjoint but not for controlled. While I think that makes sense for circuit generation, I wonder if it gets weird when we think of these entries more generically as debug metadata?

source/compiler/qsc_partial_eval/src/tests.rs
```diff
@@ -79,7 +98,32 @@ pub fn get_partial_evaluation_error_with_capabilities(
...
+            // Verify the program can go through transformations.
+            check_and_transform(&mut program.clone());
+            program
+        }
+        Err(error) => panic!("partial evaluation failed: {error:?}"),
+    }
+}
+
+#[must_use]
+pub fn get_rir_program_with_dbg_metadata(source: &str) -> Program {
```
swernli, line 119: Relative to my other comment about tests verifying debug info is well formed, this could be a place to call a utility (ahead of the expect) that checks the debug info doesn't have any "dangling pointers" and that all entries are used. I from the spot checks I've done in the existing tests it looks like they are all correct, just thinking about future changes that may subtly break things in a way that is hard to notice with just the snapshot.

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: Adding simple `validate()` to be used in tests.

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: That's much better, it gets rid of the weird nested expressions too.

source/compiler/qsc_partial_eval/src/lib.rs
minestarks: No I don't think there's anything about this use case that makes controlled calls particularly weird.

BUT, it would be helpful to think of a unique label for controlled operations just so we can show them in circuit diagrams. The only blocker is that I don't know _call_ them. I went down a literature rabbit hole recently and found a couple of options. `controlled-Foo` is an option, thought that doesn't tell you how many qubits. `controlled(2)-Foo`, maybe. I've also seen `∧₂Foo` , which makes me happy as a symbol lover, but I'm sure many will hate it.

(I think this is a separate discussion to this PR)

source/compiler/qsc_partial_eval/src/tests.rs
minestarks: added a validator

source/compiler/qsc_frontend/src/compile.rs
```diff
@@ -174,11 +174,6 @@ impl PackageStore {
             open: id,
         }
     }
-
-    #[must_use]
-    pub fn is_empty(&self) -> bool {
-        self.units.is_empty()
```
minestarks, line 180: This is dead code that happened to be referencing a buggy `is_empty()` implementation. I deleted it because why not.

PR #2943, opened 2026-02-12
[Circuit diagrams] 4 - Show conditionals in circuits based on RIR debug metadata

minestarks:
This PR completes the feature for rendering classical control flow and conditionals in quantum circuit diagrams. The changes span the compiler, code generation, circuit visualization, and VS Code integration.

**Circuit Generation from RIR**
   - Implemented new RIR-to-circuit conversion (rir_to_circuit.rs module) that reconstructs control flow structure from RIR blocks
   - Added structured control flow analysis to properly traverse branches in execution order
   - Built logic to handle complex scenarios:
     - Measurement-based conditionals (if-else based on qubit measurement results)
     - Nested conditionals and loops with quantum operations
     - Binary operation short-circuit evaluation
     - Multiple possible values in function arguments based on control flow branches

**VS Code Extension Fallback Logic**
   - Added fallback mechanism: if circuit generation fails due to adaptive profile compliance issues when using "static" method, the extension automatically retries with "classicalEval" method and "Unrestricted" profile
   - Preserves existing fallback for result comparison errors in "classicalEval" mode

### Usage Example

For Q# code with measurement-based conditionals:
```qsharp
operation TestConditional(q : Qubit) : Unit {
    H(q);
    let r = M(q);
    if (r == One) {
        X(q);
    }
}
```

The circuit diagram now displays the conditional control flow structure, showing which gates execute based on measurement outcomes, alongside the quantum operations.


source/compiler/qsc_circuit/src/rir_to_circuit.rs
```diff
@@ -0,0 +1,1981 @@
...
+    let structured_control_flow = reconstruct_control_flow(&program_rir.blocks, entry_block_id);
+
+    let num_qubits = program_rir
+        .num_qubits
+        .try_into()
+        .expect("number of qubits should fit into usize");
+    let mut wire_map_builder = FixedQubitRegisterMapBuilder::new(num_qubits);
+
+    let mut builder = OperationListBuilder::new(
+        100,
```
minestarks: Why is this 100??!

source/compiler/qsc/src/interpret.rs
```diff
@@ -1041,11 +1051,117 @@ impl Interpreter {
...
+                    // Fall back to classical eval.
+                    self.invoke_with_tracing_backend(
+                        &mut TracingBackend::<SparseSim>::no_backend(&mut tracer),
+                        &mut out,
+                        callable,
+                        args,
+                        eval_config,
+                    )?;
+                }
+                return self.static_circuit(entry_expr.as_deref(), tracer_config);
```
swernli: is this supposed to be in an `else`?

minestarks: 🫨  you're right!

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,346 @@
...
+    let topo_index: FxHashMap<BlockId, usize> =
+        ordered.iter().enumerate().map(|(i, id)| (*id, i)).collect();
+    let must_reach = compute_must_reach_sets(blocks, return_block, &ordered);
+
+    build_structured(blocks, &must_reach, &topo_index, entry, None)
+}
+
+pub(super) enum StructuredControlFlow {
+    Seq(Vec<StructuredControlFlow>),
+    /// A single basic block's "payload" (you can expand to instructions later).
```
minestarks:
```suggestion
```

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,345 @@
...
+        .iter()
+        .filter_map(|(id, b)| matches!(terminator(b), Terminator::Return).then_some(id))
+        .collect::<Vec<_>>();
+
+    assert_eq!(returns.len(), 1, "expected exactly 1 Return block");
+    returns.pop().expect("just checked non-empty")
+}
+
+/// Produce an order where every block appears before anything it can jump to.
+/// (This works because you said there are no cycles.)
```
minestarks:
```suggestion
/// This assumes there are no cycles, which should be true for Adaptive-compliant RIR we generate.
```

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,345 @@
...
+            let mut s = FxHashSet::default();
+            s.insert(return_block);
+            must_reach.insert(b, s);
+            continue;
+        }
+
+        let succs = next_blocks(blocks.get(b).expect("block should exist"));
+        assert!(
+            !succs.is_empty(),
+            "non-return block must have a next step under your assumptions"
```
minestarks:
```suggestion
            "non-return block must have a next step"
```

sorry, I left in too many comments where ChatGPT was addressing me directly... 😬

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,345 @@
...
+    a: BlockId,
+    b: BlockId,
+) -> BlockId {
+    let sa = must_reach.get(&a).expect("must reach set should exist");
+    let sb = must_reach.get(&b).expect("must reach set should exist");
+
+    let mut shared: Vec<BlockId> = sa.intersection(sb).copied().collect();
+    assert!(
+        !shared.is_empty(),
+        "paths should reconverge under your assumptions"
```
minestarks:
```suggestion
        "paths should reconverge"
```

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,345 @@
...
+    );
+
+    shared.sort_by_key(|id| order_index[id]);
+    shared[0]
+}
+
+/// Collect blocks reachable from `start` without stepping through `stop`.
+/// Useful if you want to validate that a branch arm is a clean, contained region.
```
minestarks:
```suggestion
```

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -1103,32 +1291,34 @@ impl OperationOrGroup {
...
-                    .targets
-                    .extend(target_results.iter().map(|r| Register {
-                        qubit: r.0,
-                        result: Some(r.1),
-                    }));
-                unitary
-                    .targets
-                    .sort_unstable_by_key(|r| (r.qubit, r.result));
-                unitary.targets.dedup();
+        {
```
swernli: This extra block is harmless, but doesn't do anything...

source/compiler/qsc_circuit/src/builder.rs
```diff
@@ -1335,48 +1603,56 @@ fn add_scoped_op(
...
                 && strip_scope_stack_prefix(op_call_stack, last_scope_stack).is_some()
             {
                 // The last scope matched, add to it
                 let last_scope_stack = last_scope_stack.clone();
 
-                last_op.extend_target_qubits(&op.all_qubits());
-                last_op.extend_target_results(&op.all_results());
-                let last_op_children = last_op.children_mut().expect("operation should be a group");
+                // last_op.merge_inputs(&op);
+                // let last_op_children = last_op.children_mut().expect("operation should be a group");
```
swernli: Should these lines be deleted or might they still be needed?

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,345 @@
...
+        .filter_map(|(id, b)| matches!(terminator(b), Terminator::Return).then_some(id))
+        .collect::<Vec<_>>();
+
+    assert_eq!(returns.len(), 1, "expected exactly 1 Return block");
+    returns.pop().expect("just checked non-empty")
+}
+
+/// Produce an order where every block appears before anything it can jump to.
+/// (This works because you said there are no cycles.)
+fn execution_order(blocks: &IndexMap<BlockId, Block>) -> Vec<BlockId> {
```
swernli: Not important for this PR, but food for thought for the future: the RIR SSA passes do very similar analysis on the control flow, computing the Directed Acyclic Graph for the program in order to transform or verify the transformation to SSA. I wonder if there could be an opportunity to expose some of the that from within qsc_rir such that this computation can rely on those. In a sense, this code is doing RIR analysis, so a generic suite of RIR analysis code would be useful.

minestarks: Huh! I had no idea that we already had that available. Yes, I would love to look into reusing that code.

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,338 @@
...
+        .iter()
+        .filter_map(|(id, n)| (*n == 0).then_some(*id))
+        .collect();
+
+    // Optional: keep deterministic ordering.
+    {
+        let mut v: Vec<_> = ready.drain(..).collect();
+        v.sort();
+        ready.extend(v);
+    }
```
swernli: The comment here mentions the sort is optional, but performs it unconditionaly inside an extraneous block by draining out of the `VecDeque` into a `Vec`, sorting, and then filling the first one again. This could more effeciently collect into a `Vec` upfront that it can sort and then converting it into a `VecDeque`
```suggestion
    // Start with blocks that have no incoming edges.
    let mut ready: Vec<_> = incoming_count
        .iter()
        .filter_map(|(id, n)| (*n == 0).then_some(*id))
        .collect();
    ready.sort();
    let mut ready: VecDeque<_> = ready.into();

```

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,338 @@
...
+use std::collections::VecDeque;
+use std::vec;
+
+/// RIR blocks -> Structured Control Flow
+pub(super) fn reconstruct_control_flow(
+    blocks: &IndexMap<BlockId, Block>,
+    entry: BlockId,
+) -> StructuredControlFlow {
+    let return_block = find_return_block(blocks);
+    let ordered = execution_order(blocks);
```
swernli: It turns out that RIR post-transformation is well-ordered, so this can be simplified to:
```suggestion
    // The RIR that comes back from passes uses block IDs that already ordered matching the control flow
    // as long as the program is a Directec Acyclic Graph (see source/compiler/qsc_rir/src/passes/remap_block_ids.rs).
    // Further, the `IndexMap` data structure always returns keys in ascending order, matching that DAG.
    // We rely on both assumptions below so that later code can use the ordering to recreate that
    // structure control flow.
    let ordered = blocks.iter().map(|(id, _)| id).collect::<Vec<_>>();
    assert!(ordered.is_sorted(), "IndexMap iteration order should be deterministic and sorted");
```
Though it's probably worth a nice, juicy comment about all what is happening here in addition to the assert. EDIT: added the comment!

swernli: That will also allow you to delete `execution_order` entirely, as it wouldn't be called anymore.

source/compiler/qsc_circuit/src/rir_to_circuit/control_flow.rs
```diff
@@ -0,0 +1,338 @@
...
+) -> BlockId {
+    let sa = must_reach.get(&a).expect("must reach set should exist");
+    let sb = must_reach.get(&b).expect("must reach set should exist");
+
+    let mut shared: Vec<BlockId> = sa.intersection(sb).copied().collect();
+    assert!(!shared.is_empty(), "paths should reconverge");
+
+    shared.sort_by_key(|id| order_index[id]);
+    shared[0]
+}
```
swernli, line 175: because the block ids are already in a topological ordering, you can avoid `order_index` and the sort here and just return the lowest id from the intersection:
```suggestion
fn earliest_merge_point(
    must_reach: &FxHashMap<BlockId, FxHashSet<BlockId>>,
    a: BlockId,
    b: BlockId,
) -> BlockId {
    let sa = must_reach.get(&a).expect("must reach set should exist");
    let sb = must_reach.get(&b).expect("must reach set should exist");

    *sa.intersection(sb)
        .min()
        .expect("there should be at least the return block in common")
}
```
Since this is the only place the `order_index` is used, it can be removed from the signature and caller, eventually bubbling up to removing `topo_index` from `reconstruct_control_flow`

source/compiler/qsc_circuit/src/rir_to_circuit.rs
```diff
@@ -0,0 +1,1372 @@
...
+    let operations = builder.into_operations();
+    let circuit = finish_circuit(source_lookup, operations, qubits, config.group_by_scope);
+
+    Ok(circuit)
+}
+
+/// Recursively traverses the structured control flow, pushing operations and measurement results
+/// to the builder as it goes.
+fn build_operation_list(
+    variable_tracker: &mut VariableTracker,
```
swernli, line 88: not important for right now, but if we ever run into issues with stack exhaustion I think this function could be rewritting to use an internal work queue and loop, avoiding any recursion.

PR #2945, opened 2026-02-12
[Circuit Diagrams] 3 - SVG rendering changes for classically controlled circuits

minestarks:
Reusing the existing `isConditional` field in the Circuit/.qsc schema , but changing how we render them. The old rendering code was dead code considering we never emitted this field until now.

Rendering of classical control:

image

### DIFF
--- a/source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
+++ b/source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
@@ -7,13 +7,13 @@ import {
   gateHeight,
   labelFontSize,
   argsFontSize,
-  controlBtnRadius,
-  controlBtnOffset,
+  controlCircleRadius,
+  controlCircleOffset,
   groupPaddingX,
-  classicalRegHeight,
   groupTopPadding,
   labelPaddingX,
   groupLabelPaddingY,
+  groupBottomPadding,
 } from "../constants.js";
 import {
   createSvgElement,
@@ -181,7 +181,7 @@ const _zoomButton = (renderData: GateRenderData): SVGElement | null => {
  *
  * @param renderData Operation render data.
  *
- * @returns Bounding box of the gate, not including any control dots or circles.
+ * @returns Bounding box of the gate, including any control dots and circles.
  */
 const _gateBoundingBox = (
   renderData: GateRenderData,
@@ -210,10 +210,7 @@ const _gateBoundingBox = (
   //   ╚══╡ ╞═╪═════╪═
   //      │c│┄╎     ╎
   //      ╰─╯ └╌╌╌╌╌┘
-  let ys = targetsY.flatMap((y) => y as number[]);
-  if (renderData.type === GateType.ClassicalControlled) {
-    ys = ys.concat(controlsY);
-  }
+  const ys = targetsY.flatMap((y) => y as number[]).concat(controlsY);
 
   const maxY = Math.max(...ys);
   const minY = Math.min(...ys);
@@ -639,23 +636,24 @@ const _classicalControlled = (
     const controlY = controlsY[i];
     const label = classicalControlIds?.[i] ?? null;
     // Draw control button and attached dashed line to dashed box
-    const controlCircleX: number = x + controlBtnRadius;
+    const controlCircleX: number = x + controlCircleRadius;
     const controlCircle: SVGElement = _controlCircle(
       controlCircleX,
       controlY,
       label,
     );
-    const lineY1: number = controlY + controlBtnRadius,
-      lineY2: number = controlY + classicalRegHeight / 2;
-    const vertLine: SVGElement = line(
+    const lineY1: number = controlY + controlCircleRadius;
+    const lineY2: number =
+      controlY + controlCircleRadius + groupBottomPadding / 2;
+    const vertLine: SVGElement = dashedLine(
       controlCircleX,
       lineY1,
       controlCircleX,
       lineY2,
       "classical-line",
     );
-    const lineEndX = x + controlBtnOffset;
-    const horLine: SVGElement = line(
+    const lineEndX = x + controlCircleOffset;
+    const horLine: SVGElement = dashedLine(
       controlCircleX,
       lineY2,
       lineEndX,
@@ -666,9 +664,9 @@ const _classicalControlled = (
     elems.push(horLine, vertLine, controlCircle);
   }
 
-  x += controlBtnOffset;
+  x += controlCircleOffset;
 
-  width = width - controlBtnOffset + (padding - groupPaddingX) * 2;
+  width = width - controlCircleOffset + (padding - groupPaddingX) * 2;
   x += groupPaddingX - padding;
 
   // Draw dashed box around children gates
@@ -717,7 +715,7 @@ const _controlCircle = (
   y: number,
   controlId: number | null,
 ): SVGElement =>
-  group([circle(x, y, controlBtnRadius), controlLabel(x, y, controlId)], {
+  group([circle(x, y, controlCircleRadius), controlLabel(x, y, controlId)], {
     class: "classically-controlled-btn",
   });
 
--- a/source/npm/qsharp/ux/circuit-vis/utils.ts
+++ b/source/npm/qsharp/ux/circuit-vis/utils.ts
@@ -7,7 +7,7 @@ import {
   labelPaddingX,
   labelFontSize,
   argsFontSize,
-  controlBtnOffset,
+  controlCircleOffset,
 } from "./constants.js";
 import { ComponentGrid, Operation } from "./circuit.js";
 import { Register } from "./register.js";
@@ -70,7 +70,7 @@ const getMinGateWidth = ({
     default: {
       // Classically controlled gates are wider because of the control button on the left
       const controlButtonWidth =
-        type === GateType.ClassicalControlled ? controlBtnOffset : 0;
+        type === GateType.ClassicalControlled ? controlCircleOffset : 0;
       const labelWidth = _getStringWidth(label);
       const argsWidth =
         displayArgs != null ? _getStringWidth(displayArgs, argsFontSize) : 0;
--- a/source/npm/qsharp/test/circuits-cases/conditionals.qsc.snapshot.html
+++ b/source/npm/qsharp/test/circuits-cases/conditionals.qsc.snapshot.html
@@ -1819,16 +1819,18 @@
                       <line
                         x1="219"
                         x2="244"
-                        y1="208"
-                        y2="208"
+                        y1="216"
+                        y2="216"
                         class="classical-line"
+                        stroke-dasharray="8, 8"
                       />
                       <line
                         x1="219"
                         x2="219"
                         y1="211"
-                        y2="208"
+                        y2="216"
                         class="classical-line"
+                        stroke-dasharray="8, 8"
                       />
                       <g class="classically-controlled-btn">
                         <circle cx="219" cy="196" r="15" />
@@ -1991,16 +1993,18 @@
                       <line
                         x1="219"
                         x2="244"
-                        y1="488"
-                        y2="488"
+                        y1="496"
+                        y2="496"
                         class="classical-line"
+                        stroke-dasharray="8, 8"
                       />
                       <line
                         x1="219"
                         x2="219"
                         y1="491"
-                        y2="488"
+                        y2="496"
                         class="classical-line"
+                        stroke-dasharray="8, 8"
                       />
                       <g class="classically-controlled-btn">
                         <circle cx="219" cy="476" r="15" />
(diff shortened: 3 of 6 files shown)


source/npm/qsharp/ux/circuit-vis/gateRenderData.ts
```diff
@@ -61,4 +61,6 @@ export interface GateRenderData {
   dataAttributes?: DataAttributes;
   /** Link href and title for clickable gate. */
   link?: { href: string; title: string };
+  /** Labels for the classical control registers (applicable only for GateType.ClassicalControlled). */
+  classicalControlIds?: (number | null)[];
```
ScottCarda-MS, line 65: When would this have a null entry?

source/npm/qsharp/test/circuits.js
```diff
@@ -198,6 +198,7 @@ test("circuit snapshot tests - .qsc files", async (t) => {
           editCallback: () => {},
         },
         renderLocations,
+        renderDepth: 999999,
```
ScottCarda-MS, line 201: Is renderDepth set here so that the circuits are draw fully expanded?

source/npm/qsharp/ux/circuit-vis/gateRenderData.ts
```diff
@@ -52,7 +52,7 @@ export interface GateRenderData {
   /** Gate width. */
   width: number;
   /** Children operations as part of group. */
-  children?: GateRenderData[][] | GateRenderData[][][];
```
ScottCarda-MS, line 55: I seem to recall running into problems with this data type before, where sometimes it would be a tripply-nested array and other times not. Good to see it is simplified here. What was the change that allowed this?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -180,15 +181,40 @@ const _zoomButton = (renderData: GateRenderData): SVGElement | null => {
...
 const _gateBoundingBox = (
   renderData: GateRenderData,
-): [number, number, number, number] => {
-  const { x: centerX, width, targetsY, topPadding, bottomPadding } = renderData;
+): { x: number; y: number; width: number; height: number } => {
+  const {
+    x: centerX,
+    width,
+    targetsY,
+    controlsY,
```
ScottCarda-MS, line 193: Do controlsY need to also be grouped based on what box of the gate they are interacting with? Similar to targetsY?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -590,67 +626,73 @@ const _classicalControlled = (
...
-  if (children != null) {
-    if (children.length !== 2)
-      throw new Error(
-        `Invalid number of children found for classically-controlled gate: ${children.length}`,
-      );
+  for (let i = 0; i < controlsY.length; i++) {
+    const controlY = controlsY[i];
+    const label = classicalControlIds?.[i] ?? null;
+    // Draw control button and attached dashed line to dashed box
+    const controlCircleX: number = x + controlBtnRadius;
```
ScottCarda-MS: Isn't the `x` here the x value of the classically controlled group (the box)? I thought the control label circle was draw to the left of the boxes?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -590,67 +626,73 @@ const _classicalControlled = (
...
+    const controlCircle: SVGElement = _controlCircle(
+      controlCircleX,
+      controlY,
+      label,
+    );
+    const lineY1: number = controlY + controlBtnRadius,
+      lineY2: number = controlY + classicalRegHeight / 2;
```
ScottCarda-MS: nit: Syntax here feels a bit strange. I'd prefer to just have two `const` declarations over two lines. That would feel more "normal" to me.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -590,67 +626,73 @@ const _classicalControlled = (
...
+    const controlY = controlsY[i];
+    const label = classicalControlIds?.[i] ?? null;
+    // Draw control button and attached dashed line to dashed box
+    const controlCircleX: number = x + controlBtnRadius;
+    const controlCircle: SVGElement = _controlCircle(
+      controlCircleX,
+      controlY,
+      label,
+    );
+    const lineY1: number = controlY + controlBtnRadius,
```
ScottCarda-MS: nit: Might consider changing the name of the constant `controlBtnRadius` now that it isn't actually a clickable button.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
```diff
@@ -590,67 +626,73 @@ const _classicalControlled = (
...
+      lineY2: number = controlY + classicalRegHeight / 2;
+    const vertLine: SVGElement = line(
```
ScottCarda-MS: This vertical line doesn't show up because it is based on classicalRegHeight rather than the radius of the control circle. Currently reg height / 2 is less than the radius.

source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
```diff
@@ -38,17 +38,22 @@ const formatRegisters = (
 
     // Render classical wires
     for (const classical of registers[qId].children || []) {
+      const ys: number[] = [];
       for (const gate of allGates.flat()) {
         if (
-          gate.type === GateType.Group &&
-          gate.dataAttributes?.["expanded"] === "true"
+          gate.type === GateType.Group ||
+          gate.type === GateType.ClassicalControlled
```
ScottCarda-MS, line 45: This seems like a strange change. Previously we weren't rendering classical wires for groups only when they were expanded, but now we aren't rendering classical wires for groups ever.

source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
```diff
@@ -38,17 +38,22 @@ const formatRegisters = (
...
         ) {
           // Don't render classical wires for a group that is expanded - the wires
           // will be coming out of the measurement operations *inside* the group.
           continue;
         }
 
         for (const y of gate.targetsY.flat().filter((y) => y === classical.y)) {
+          if (ys.includes(y)) {
```
ScottCarda-MS, line 53: Is this just for efficiency, or were you seeing this cause errors when we weren't skipping over already seen registers?

source/npm/qsharp/ux/circuit-vis/utils.ts
```diff
@@ -260,20 +261,9 @@ function getMinMaxRegIdx(
...
-  );
-  const isClassicallyControlled: boolean = clsControls.length > 0;
-  if (!isClassicallyControlled && qRegs.length === 0) return [-1, -1];
-  // If operation is classically-controlled, pad all qubit registers. Otherwise, only pad
-  // the contiguous range of registers that it covers.
-  const minRegIdx: number = isClassicallyControlled ? 0 : Math.min(...qRegs);
-  const maxRegIdx: number = isClassicallyControlled
-    ? numQubits - 1
-    : Math.max(...qRegs);
+  const qRegs = [...controls, ...targets].map(({ qubit }) => qubit);
```
ScottCarda-MS, line 264: It is concerning to me that we aren't filtering out the classical registers from this list anymore. That means the `qRegs` doesn't mean the qubit registers anymore.

ScottCarda-MS: I think that could have significant impact to the behavior of the editor.

ScottCarda-MS: Discussed offline: the only effect (for CE circuits) is that qubit registers might be listed multiple times, but that won't effect the ultimate min and max qubit indexes returned by the function.

source/npm/qsharp/test/circuits.js
minestarks: Yes. Though this will become irrelevant when I delete the .qsc files that contain nested/conditional operations in the next PR.

source/npm/qsharp/ux/circuit-vis/gateRenderData.ts
minestarks: Deleting the old rendering logic for `ClassicallyControlled` gates. They used to have two sets of children "Ones" and "Zeros" that you could toggle between. The additional array layer was to contain these two sets. 

With my change, these gates only have one set of children, and the toggle functionality is gone.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: No, the additional `targetsY` grouping is when an operation box gets split up into two boxes vertically (because it's crossing unused qubit wires). we don't need that info for `controlsY`

```
[foo]
  |
  |
  |
[foo]
```

Whoever wrote this code really loved arrays, unfortunately.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: I agree. The original author also loved comma-separated declarations, which always confuse my eyes too.

(I tried to eliminate them where I touched the code, but looks like I missed this).

source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
minestarks: This was a real issue. We drew the same `<line>` for the classical wire over and over again, resulting in a weirdly thick line

source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
minestarks: Ah, sorry yes - there are two changes here:

1. Turns out that  `gate.dataAttributes?.["expanded"]` was just redundant. If type is set to `GateType.Group`, that means it IS an expanded group by definition. (unexpanded groups become `GateType.Unitary`)
2. The `GateType.ClassicalControlled` case is newly included in this check.

minestarks: Essentially this check is asking "is this a dashed box with visible children"?

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
ScottCarda-MS: Do classical-control boxes also get split up like that? If they do, they might need to follow the same pattern as the targetY groupings

source/npm/qsharp/ux/circuit-vis/gateRenderData.ts
minestarks: This should never happen in practice (when we are generating the circuit), but hypothetically, you could receive a `Circuit` (.qsc) object that is missing a label for a particular classical register. In that case, the corresponding ID in this array would be `null`.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: They don't. Only `GateType.Unitary`ies get split up like that. For dashed-box type gates (`GateType.Group` and `GateType.Controlled`), there's one big dashed box that simply goes around everything.

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: Hmm... apparently not. The box returned by `_gateBoundingBox` does include the control circle, so `x` would be the beginning of the control circle here. 

```
 (c0) ---- [box]
^
```

The comment in `_gateBoundingBox` was wrong, I'll update it. For simplicity's sake, I will also adjust the behavior `_gateBoundingBox` to always include control dots/circles (regardless of gate type). (This won't change any behavior in practice, since today `_gateBoundingBox` is only called in for gates that can't have any qubit controls.)

source/npm/qsharp/ux/circuit-vis/formatters/gateFormatter.ts
minestarks: I changed it to be based on `controlBtnRadius` so that it juts out a little below the circle.

I also made a last-minute aesthetic decision to change this line back to dashed since I (subjectively) think it looks more cohesive

image

minestarks:
> The "if" in the conditional label looks a little off

@ScottCarda-MS  oh that's a weird one... looks like that's just the math font we use. Here is an example (this is `main`, without any of my changes) . I wonder if `f` is special because it usually denotes a function in math. 🤷‍♀️ 

image

PR #2949, opened 2026-02-12
Introduce separate multi-platform CI on merge to Main

swernli:
This change updates how our CI is structured, using single platform for PR and merge queue CI runs and a new multi-platform task to execute after merges to main, filing issues when failures occur to prompt investigation.


swernli:
Here's an example of the multi-platform run:

https://github.com/microsoft/qdk/actions/runs/21969059537

PR #2952, opened 2026-02-14
Refactor trace and result properties

msoeken:
This PR cleans up the way properties are handled for traces and results. Trace properties can be assigned by application generators and trace transforms and are stored in optimal resource estimation results. The Python API automatically detects whether the property value is `bool`, `int`, `float`, or `str` and does not require a dedicated property class exposed from Rust.

Some other changes include
- Fixes a deadlock due to the Python GIL
- Allows parallel trace generation from application generators
- Reorganizes code around Q# application generators
- Sets up resource estimation for memory qubits

### DIFF
--- a/source/pip/qsharp/qre/_qre.pyi
+++ b/source/pip/qsharp/qre/_qre.pyi
@@ -958,6 +958,18 @@ class Trace:
         """
         ...
 
+    def has_property(self, key: str) -> bool:
+        """
+        Checks if a property with the given key exists.
+
+        Args:
+            key (str): The property key.
+
+        Returns:
+            bool: True if the property exists, False otherwise.
+        """
+        ...
+
     @property
     def depth(self) -> int:
         """
--- a/source/pip/src/qre.rs
+++ b/source/pip/src/qre.rs
@@ -724,6 +724,10 @@ impl Trace {
         }
     }
 
+    pub fn has_property(&self, key: &str) -> bool {
+        self.0.has_property(key)
+    }
+
     #[allow(clippy::needless_pass_by_value)]
     #[getter]
     pub fn resource_states(self_: PyRef<'_, Self>) -> PyResult<Bound<'_, PyDict>> {
--- a/source/qre/src/result.rs
+++ b/source/qre/src/result.rs
@@ -99,6 +99,11 @@ impl EstimationResult {
     }
 
     #[must_use]
+    pub fn has_property(&self, key: &str) -> bool {
+        self.properties.contains_key(key)
+    }
+
+    #[must_use]
     pub fn properties(&self) -> &FxHashMap<String, Property> {
         &self.properties
     }
--- a/source/qre/src/trace.rs
+++ b/source/qre/src/trace.rs
@@ -128,6 +128,11 @@ impl Trace {
     }
 
     #[must_use]
+    pub fn has_property(&self, key: &str) -> bool {
+        self.properties.contains_key(key)
+    }
+
+    #[must_use]
     pub fn deep_iter(&self) -> TraceIterator<'_> {
         TraceIterator::new(&self.block)
     }


source/pip/qsharp/qre/_estimation.py
```diff
@@ -50,6 +51,9 @@ def estimate(
     app_ctx = application.context()
     arch_ctx = architecture.context()
 
+    if trace_query is None:
```
brad-lackey, line 54: If I understand this correctly, this sets PSSPC over lattice surgery as the default layout. Suppose this matches what happens in QREv2 (by necessity). This is probably best for testing and getting out the first beta QRE. But maybe in the long run we don't want a default, and really throw and error if no layout is specified (namely we want a layout better tied to the selection of codes and/or compute-memory model.

source/pip/qsharp/qre/_qre.pyi
```diff
@@ -967,6 +886,44 @@ class Trace:
         """
         ...
 
+    @property
+    def memory_qubits(self) -> Optional[int]:
```
brad-lackey, line 890: This is a bit inconsistant with the general form of set_property/get_property below. It looks like this should be named get_memory_qubits, which is syntactic sugar for composin a conditional based on has_memory_qubits() followed by returning the number.

source/pip/qsharp/qre/_qre.pyi
```diff
@@ -977,25 +934,27 @@ class Trace:
         """
         ...
```
brad-lackey, line 936: Should we expose has_property(), so as to match the API for memory qubits?

source/pip/src/qre.rs
```diff
@@ -725,12 +686,42 @@ impl Trace {
...
+        } else if let Ok(f) = value.extract::<f64>() {
+            qre::Property::new_float(f)
+        } else {
+            qre::Property::new_str(value.to_string())
+        };
+
+        self.0.set_property(key, property);
+
+        Ok(())
     }
```
brad-lackey, line 704: See comment in _qre.pyi.

source/pip/src/qre.rs
```diff
@@ -775,6 +766,23 @@ impl Trace {
         })
     }
 
+    #[getter]
+    pub fn memory_qubits(&self) -> Option<u64> {
```
brad-lackey, line 774: See comment in _qre.pyi.

source/qre/src/result.rs
```diff
@@ -87,6 +88,20 @@ impl EstimationResult {
     pub fn isa(&self) -> &ISA {
         &self.isa
     }
+
```
brad-lackey, line 91: Similarly, expose has_property() for testing if a property key exists?

source/qre/src/trace.rs
```diff
@@ -68,6 +68,33 @@ impl Trace {
         self.base_error += amount;
     }
 
+    #[must_use]
+    pub fn memory_qubits(&self) -> Option<u64> {
```
brad-lackey, line 72: As above should be named get_memory_qubits()?

source/pip/qsharp/qre/_estimation.py
msoeken: I agree. It would be great if the signature allows to omit the trace query (default) but there is no default layout. It is derived from the architecture and application.

source/pip/qsharp/qre/_qre.pyi
msoeken: It is consistent with all the other getters, like `compute_qubits`, `resource_states`, `depth`. In contrast, `get_property` is a function because we must pass a property key. In a code example it's as follows:

```python
trace = get_a_trace_from_somewhere()
c = trace.compute_qubits
m = trace.memory_qubits
r = trace.resource_states
name = trace.get_property("name")
```

What we could do to enhance consistency is returning all properties and then querying them.

```python
name = trace.properties["name"]
has_name = "name" in trace.properties
```

source/pip/qsharp/qre/_qre.pyi
msoeken: Fixed in next commit.

PR #2963, opened 2026-02-19
Various RE models and some API enhancements

msoeken:
This PR adds various models:
- Qubits
  - Majorana
- Error correcting codes
  - ThreeAux
  - Yoked surface code
- Factories
  - Round-based factory
  - Litinski19 factory

It also makes some updates to the surface code model and improves documentation. Further, it adds a large test suite for all the models. Some of these tests are very detailed but they will help us to spot if we introduce inconsistencies with future code changes.

Besides that there are some fixes to the Python API:
- Adds caching to Q# to Trace transformation
- InstructionFrontier can be 2D or 3D (without and with error rates)
- Instance enumeration can also consider nested types and union types now
- EstimationResult can be used from the Python API to create new results

### DIFF
--- a/source/pip/test_requirements.txt
+++ b/source/pip/test_requirements.txt
@@ -1,2 +1,3 @@
 pytest
 pyqir<0.12
+cirq==1.6.1


source/pip/qsharp/qre/models/factories/_litinski.py
```diff
@@ -0,0 +1,359 @@
...
+                    yield ISA(
+                        make_instruction(t_entry),
+                        make_instruction(ccz_entry),
+                    )
+        else:
+            # Table 2 scenarios: only T gates available
+            for t_entry in t_entries:
+                yield ISA(make_instruction(t_entry))
+
+    def _initialize_entries(self):
```
brad-lackey, line 125: Yep, all the numbers from Tables 1 and 2 validated. We will build some more concrete models starting next week.

source/pip/qsharp/qre/models/factories/_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    SQRT_SQRT_Y_DAG,
+    SQRT_SQRT_Z,
+    SQRT_SQRT_Z_DAG,
+    CCX,
+    CCY,
+    CCZ,
+)
+
+
+class MagicUpToClifford(ISATransform):
```
brad-lackey, line 22: Hmmm.... This is essentially type coersion. It's not clear that we want push type coersion out to each component. Since we're only looking at T and CCZ at this point, this is not too cumbersome. But I can see things getting out of control quickly. An alternative would be do define a type for T_LIKE and CCZ_LIKE, and set the fectories for providing these types, and then do coersion as we have done with (for instance) ONE_QUBIT_UNITARY.

source/pip/qsharp/qre/models/qec/_three_aux.py
```diff
@@ -0,0 +1,118 @@
...
+    LATTICE_SURGERY,
+    MEAS_X,
+    MEAS_XX,
+    MEAS_Z,
+    MEAS_ZZ,
+)
+
+
+@dataclass
+class ThreeAux(ISATransform):
```
brad-lackey, line 31: Looks good. Numbers validated against the paper. We should consider additional analysis for this as it is important for other programmes too.

source/pip/qsharp/qre/models/qec/_yoked.py
```diff
@@ -0,0 +1,158 @@
...
+        SQUARE: The shape that minimizes the difference between the number of rows
+            and columns.
+    """
+
+    MIN_AREA = 0
+    SQUARE = 1
+
+
+@dataclass
+class YokedSurfaceCode(ISATransform):
```
brad-lackey, line 31: This is just one instance of the yoked surface code. I'll add a task to my board open a new pull request after this one is merged to add to the docs about this version of the code, and add a number of TODOs to implement other versions.

source/pip/qsharp/qre/models/qubits/_msft.py
```diff
@@ -0,0 +1,97 @@
...
+    MEAS_Z,
+)
+from ..._instruction import ISA, instruction
+
+
+@dataclass
+class Majorana(Architecture):
+    """
+    This class models physical instructions that may be relevant for future
+    Majorana qubits [71, 74, 117]. For these qubits, we assume that measurements
```
brad-lackey: Can remove numerical references.

source/pip/qsharp/qre/models/factories/_utils.py
msoeken: I think there is a difference between coercion, where we can say that T is a ONE_QUBIT_UNITARY, and this case. Here we say, e.g., that we want to use CCZ magic state factories to produce CCX magic states required by our algorithm, and explicitly mark that we assume that turning the CCZ magic state factory into a CCX factory would have the same cost.

source/pip/qsharp/qre/models/qubits/_msft.py
msoeken:
```suggestion
    Majorana qubits. For these qubits, we assume that measurements
```

PR #2964, opened 2026-02-19
Use tagged aggregates in QIR

swernli:
This change updates the QIR generation to use non-empty tags for aggregate types (arrays and tuples) for compatiblity with the latest service translation changes.

NOTE: The vast majority of this change is updated the expected output for tests, including both QIR snapshots and execution snapshots. The only production code changes are in:

- source/compiler/qsc_codegen/src/qir.rs
- source/compiler/qsc_codegen/src/qir/template.ll
- source/compiler/qsc_partial_eval/src/lib.rs
- source/compiler/qsc_rir/src/rir.rs

### DIFF
--- a/source/pip/test_requirements.txt
+++ b/source/pip/test_requirements.txt
@@ -1,4 +1,3 @@
 pytest
 expecttest==0.3.0
 pyqir<0.12
-expecttest==0.3.0


source/pip/test_requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest
 expecttest==0.3.0
 pyqir<0.12
+expecttest==0.3.0
```
orpuente-MS: Looks like `expecttest==0.3.0` was already in this file

swernli: Ah, I added it in a parallel change and git merge the two. Good catch!

swernli:
Tested with submissions to all simulator targets

PR #2973, opened 2026-02-26
Resource estimation results table

msoeken:
The main feature from this PR is a resource estimation result table that is returned from the `estimate` call. It can be configured to have additional columns which are considered in the creation of a pandas DataFrame.

Besides that, the PR
- refactors how instructions are build: they are now all build inside the provenance graph that is part of the architecture context
- enhances enumeration capabilities: one can now enumerate over nested attributes and union attributes, and also restrict domains, further one can restrict the domain of the application inside the trace query
- adds post processing to applications: applications can post-process estimation results _before_ they are inserted in the estimation table. This post processing step comes currently at a cost in runtime, because estimation is parallelized in Python and not in Rust

### DIFF
--- a/source/pip/qsharp/magnets/geometry/complete.py
+++ b/source/pip/qsharp/magnets/geometry/complete.py
@@ -11,7 +11,7 @@ systems with all-to-all or bipartite all-to-all interactions.
 from qsharp.magnets.utilities import (
     Hyperedge,
     Hypergraph,
-    greedy_edge_coloring,
+    HypergraphEdgeColoring,
 )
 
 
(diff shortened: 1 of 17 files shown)


source/pip/test_requirements.txt
```diff
@@ -2,3 +2,4 @@ pytest
 expecttest==0.3.0
 pyqir<0.12
 cirq==1.6.1
+pandas==2.3.3
```
billti: I see v3 shipped recently (https://pandas.pydata.org/community/blog/pandas-3.0.html). Does it work fine with that release also?

msoeken: I think it does, I will verify. I used this version because it's the one that cirq already depends on. But I can change the operator.

billti: Will QRE require a specific version of Cirq then also? Or is the goal that it should be broadly compatible with whatever (recent) version of Cirq you might be already using?

msoeken: It doesn't require a specific version. I made a change to relax the pandas requirement such that it still works. I double checked that it works with pandas 3.0.1, but cirq pins the pandas version to pandas~=2.1, so there might be pip resolver warnings if a user has pandas 3 when also cirq installed.

PR #2974, opened 2026-02-26
Add orbital entanglement diagram widget

nabbelbabbel:
**Note:** This PR is a lot of AI generated infra to achieve an outcome for testing purposes. While the features work, the way they are done needs some serious work and rethinking.

Adds:
- orbital entanglement diagram widget
- option to generate static SVG from orbital entanglement diagram widget for publications/presentations with python code
- option to generate static SVG from histogram
- option to generate static SVG from circuit

Widget with active space selection (gold):
image

Widget while hovering individual orbital/entangled node:
image


source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,220 @@
...
+    );
+
+    output = html;
+    break;
+  }
+
+  // ---- Circuit (imperative DOM via qviz) ----
+  case "Circuit": {
+    // qviz.draw() needs a real DOM.  Use jsdom.
+    const { JSDOM } = await import("jsdom");
```
billti: This feature assume a) Node is on the users PATH (and at least it gives an error message if not), and b) That `jsdom` is available as a package to import in node (which it isn't by default, the user would need to install it into the global (or local) node environment. This seems fragile and likely to fail unless users have followed some specific setup instructions first. (We also don't want to bundle JSDOM into our package as it's huge).

Not sure what to do about this, but worth thinking about.

source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,220 @@
...
+import { OrbitalEntanglement } from "../../npm/qsharp/ux/orbitalEntanglement.tsx";
+import { Histogram } from "../../npm/qsharp/ux/histogram.tsx";
+import { draw as drawCircuit } from "../../npm/qsharp/ux/circuit-vis/index.ts";
+import { toCircuitGroup } from "../../npm/qsharp/src/data-structures/legacyCircuitUpdate.ts";
+
+// ---------- Embedded CSS for standalone SVGs ----------
+// The interactive widgets inherit CSS from the page's theme variables.
+// For standalone SVG export we resolve those variables to concrete values.
+
+const HISTOGRAM_CSS_LIGHT = `
```
billti: Are the next 70 or so lines extracted from our CSS somehow? What's the plan to keep this up to date as we iterate on our CSS (which we tend to a fair bit lately). Manually keeping in sync seems fragile and error prone.

source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,220 @@
...
+  .bar-label-ket { font-family: Consolas, "Menlo", monospace; font-variant-ligatures: none; }
+  .histo-label { font-size: 3.5pt; fill: #eee; }
+  .hover-text { font-size: 3.5pt; fill: #eee; text-anchor: middle; }
+`;
+
+const CIRCUIT_CSS_LIGHT = `
+  line, circle, rect { stroke: #202020; stroke-width: 1; }
+  text { fill: #202020; dominant-baseline: middle; text-anchor: middle;
+         user-select: none; pointer-events: none; }
+  .qs-maintext { font-family: "KaTeX_Main", sans-serif; font-style: normal; }
```
billti: Unless the user has installed this font-family on their machine somehow, then loading these fonts will likely fail (in a web context, the font files are referenced and fetched). Do the SVG render fine and look correct without these files?

package.json
```diff
@@ -49,6 +49,7 @@
     "monaco-editor": "^0.44.0",
     "openai": "^4.83.0",
     "preact": "^10.20.0",
+    "preact-render-to-string": "^6.6.6",
```
billti: The version of the beast! 😡

source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,220 @@
...
+      /(<svg\s[^>]*>)/,
+      `$1<defs><style>${css}</style></defs>`,
+    );
+
+    output = html;
+    break;
+  }
+
+  // ---- Circuit (imperative DOM via qviz) ----
+  case "Circuit": {
```
billti: Did you try this at all? Thinking through how this works, you wouldn't get what's displayed on screen (unless you've not interacted with it). You'll just get the default new rendering with a circuit, which now has all blocks/calls/operations collapsed. If the user has drilled into operations to expand them and show the gates etc., that's not going to show up here, as it's re-rendering the circuit from a clean state, not the current UX state.

nabbelbabbel: Yes that was the intended behavior (for now).

The point of these functions was not so save the user-altered state of the widget into an SVG. (Though that would be even better!)

The idea was to be able to make the widget class, call the SVG dump and never need to look at it to extract an SVG to later view/use.

Will see if I can get it to work using the actual state of the widget.

source/widgets/js/render_svg.mjs
nabbelbabbel: Agree, an initial version had JSDOM as dependency but I removed it for the reason you stated. I dont think I can get rid of the Node requirement if I want to allow rendering without actually displaying the widget. But I can shim/mock the few parts of JSDOM we need.

source/widgets/js/render_svg.mjs
nabbelbabbel: I changed this, it is now removed.

source/widgets/js/render_svg.mjs
nabbelbabbel: This too is gone with the removal of the custom CSS

source/widgets/js/render_svg.mjs
nabbelbabbel: I added the shim, but its 500 lines of extra stuff, thoughts?

source/widgets/js/render_svg.mjs
nabbelbabbel: I have now added functionality to use the actual state of the widgets but still allowing a different path to render without the widget being displayed to the user.

source/npm/qsharp/ux/orbitalEntanglement.tsx
```diff
@@ -0,0 +1,844 @@
...
+const DEFAULT_EDGE_CMAP: [string, string, string] = [
+  "#d8d8d8",
+  "#2060b0",
+  "#1a1a1a",
+];
+
+/**
+ * Detect whether the host background is dark or light by sampling the
+ * computed background-color of the nearest ancestor with one.
+ * Returns a high-contrast colour for selection outlines.
```
billti, line 218: Why not use CSS values to match the current theme like the other widgets?

source/npm/qsharp/ux/orbitalEntanglement.tsx
```diff
@@ -0,0 +1,844 @@
...
+    edgeColorbarLabel = null,
+    nodeHoverPrefix = "",
+    edgeHoverPrefix = "",
+    groupSelected = false,
+    darkMode,
+    static: isStatic = false,
+  } = props;
+
+  // --- theme-resolved colours ---
+  const FONT_FAMILY = '"Segoe UI", Roboto, Helvetica, Arial, sans-serif';
```
billti: Again, we should use CSS for appearance properties like font selection and theme colors. Baking it into the code is hard to manage and get consistent behavior across widgets.

source/npm/qsharp/ux/orbitalEntanglement.tsx
```diff
@@ -0,0 +1,844 @@
...
+  }, []);
+  const selectionColor = selectionColorProp ?? autoSelectionColor;
+
+  // --- labels ---
+  const labels: string[] =
+    labelsProp && labelsProp.length === n
+      ? labelsProp
+      : Array.from({ length: n }, (_, i) => String(i));
+
+  // --- colour scales ---
```
billti, line 494: Nit: the spelling of 'color' or 'colour' is inconsistent every few lines (e.g. line 287, 289, 303, etc.). Being most code authors are in the U.S., we should probably pick the U.S. spelling and stick with it.

source/widgets/js/index.tsx
```diff
@@ -61,6 +63,13 @@ function render({ model, el }: RenderArgs) {
     el.ownerDocument.head.appendChild(forceStyle);
   }
 
+  // Belt-and-suspenders: also set the background inline on the nearest
```
billti: What does "Belt-and-suspenders" mean? Also, is the class you are looking for below consistent across notebooks in VS Code, notebooks in the browser, and notebooks in Google Colab? Again, prefer the detection and theme selection we already do in other widgets via CSS for consistency in behavior across widgets.

source/widgets/js/index.tsx
```diff
@@ -301,6 +361,282 @@ function renderAtoms({ model, el }: RenderArgs) {
...
+  const svgEl = el.querySelector("svg");
+  if (!svgEl) return null;
+
+  // Clone so we don't mutate the live DOM
+  const clone = svgEl.cloneNode(true) as SVGSVGElement;
+  if (stripSelector) {
+    clone.querySelectorAll(stripSelector).forEach((n) => n.remove());
+  }
+  // Remove interactive-only elements common across widgets
+  clone
```
billti: I think this is an interesting approach (cloning and serializing the SVG), but this implementation seems hard to maintain and fragile. Below you are baking knowledge of the internals of various widgets into this one central location outside of the widgets. This would mean if we ever touch those widgets (e.g. rename the `.menu-icon` class or `#menu` id or some of the other classes or styles hard-coded below, we'd need to remember to come here and change it also. That would be easy to miss and introduce bugs we wouldn't find for a while (being automated testing for this is hard)

I'm also not a big fan of automatically watching for changes and cloning the svg on mutation, being the SVG can be really huge (esp. for circuits, and likely for chord diagrams), and the vast majority of time people aren't going to use the 'save the svg feature'. 

Is there a way to push the 'cleaning' of the SVG/DOM to the owning widget (so it's easy to see when changing the widget itself), and only clone & serialize the SVG when the user actually wants to save the SVG?

source/widgets/js/svgDomShim.mjs
```diff
@@ -0,0 +1,596 @@
...
+  return function restoreGlobals() {
+    for (const [k, v] of Object.entries(saved)) {
+      if (v === undefined) {
+        delete globalThis[k];
+      } else {
+        globalThis[k] = v;
+      }
+    }
+  };
+}
```
billti: This is a lot of _very_ subtle code (the DOM has a lot of nuance) to turn around in a few hours, and will take careful review. Did you personally write this? Or is it taken from another library or AI generated?

source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,159 @@
...
+// standalone SVG the :root selector matches the <svg> element, so
+// CSS var() fallback chains resolve to their concrete light-mode values
+// (e.g. var(--vscode-editor-foreground, var(--jp-widgets-color, #202020))
+// → #202020) because none of the host-specific custom properties exist.
+import themeCss from "../../npm/qsharp/ux/qdk-theme.css";
+import uxCss from "../../npm/qsharp/ux/qsharp-ux.css";
+import circuitCss from "../../npm/qsharp/ux/qsharp-circuit.css";
+
+const input = readFileSync(0, "utf-8"); // stdin
+const { component, props } = JSON.parse(input);
```
billti: If you're cloning and serializing the SVG above now, why do you need to shell out to node and render the component via the prop at all? What's the use-case for this file now?

source/widgets/js/svgDomShim.mjs
nabbelbabbel: This was AI generated to meet the needs for now. I consider this still in the ideation stage to see what could be the best way of getting the result we are looking for.

I will draft the PR to make this a bit clearer.

source/widgets/js/render_svg.mjs
nabbelbabbel: The use case I have in mind is not having a browser/notebook/vscode instance at all.

Essentially thinking about someone running qdk-chemistry or any other component on a remote HPC system and wanting to generate a picture to quickly see if a calculation did what it was supposed to.

If I understand correctly, anywidget uses the DOM of whatever the widget is shown in, meaning that if there is not VSCode/jupyterlab/etc, there also is no DOM to render the SVG with.

source/widgets/js/index.tsx
nabbelbabbel: Fully agree, this is all an artifact of the idea that I outlined below with your question about the second svg export route.

source/widgets/js/render_svg.mjs
```diff
@@ -0,0 +1,159 @@
...
+    // match from the last <g style="display: none;"> to the closing </svg>
+    html = html.replace(
+      /<g\s+style="display:\s*none;"[^>]*>[\s\S]*?<\/g>\s*(?=<\/svg>)/,
+      "",
+    );
+
+    // Add xmlns for standalone SVG and set a reasonable default render size
+    html = html.replace(
+      /<svg\s+class="histogram"/,
+      '<svg xmlns="http://www.w3.org/2000/svg" class="histogram" width="600"',
```
billti: This is just way too fragile. You're looking for specific CSS classes and ids from various widgets to do fixups in multiple places in this code (line 76, 77, 88, 124, 130). This will break as soon as anyone touches those style/class names (which we've done a bit), or adds new styles/classes that this code would also need to handle. I think we need to rethink how to handle this special cases, maybe with something consistent and foolproof across widgets.

source/npm/qsharp/ux/orbitalEntanglement.tsx
nabbelbabbel: Fixed that

source/npm/qsharp/ux/orbitalEntanglement.tsx
nabbelbabbel: The current version should fix this but have generic fall backs in case there is no theme.

source/widgets/js/index.tsx
```diff
@@ -301,6 +305,60 @@ function renderAtoms({ model, el }: RenderArgs) {
...
+        arcWidth={options.arc_width as number | undefined}
+        lineScale={options.line_scale as number | null | undefined}
+        miThreshold={options.mi_threshold as number | undefined}
+        s1Vmax={options.s1_vmax as number | null | undefined}
+        miVmax={options.mi_vmax as number | null | undefined}
+        title={options.title as string | null | undefined}
+        width={options.width as number | undefined}
+        height={options.height as number | undefined}
+        selectionColor={options.selection_color as string | undefined}
+        selectionLinewidth={options.selection_linewidth as number | undefined}
```
billti: I don't think you should break out every option and pass it individually. Just pass the `options` object to the widget as a prop member.

source/widgets/src/qsharp_widgets/__init__.py
```diff
@@ -224,6 +231,102 @@ def __init__(self, machine_layout, trace_data):
...
+            ).tolist()
+            mutual_information = np.asarray(
+                wavefunction.get_mutual_information()
+            ).tolist()
+            n = len(s1_entropies)
+            if labels is None:
+                try:
+                    orbitals = wavefunction.get_orbitals()
+                    if orbitals.has_active_space():
+                        active_indices = orbitals.get_active_space_indices()[0]
```
billti, line 350: Just going over this in more detail, and I see `wavefunction` is expected to be of a very specific type (that supports methods such as `get_single_orbital_entropies()`, `get_mutual_information()`, `get_active_space_indices()`, etc.)

What is this type? The API contains little in the way of information (or type checking) as to what it is. Is it from qdk-chemistry? If so, are these _super_ stable APIs? Seems like we'd want to tests to cover regressions if these APIs might change across versions. (So we know to update the widget when it breaks)

I see this code also has a runtime dependency on `numpy` on line 292 (which the QDK doesn't have a dependency on). That might be fine, as if you're using Jupyter then you almost certainly have it. That said, if dynamically imported packages at runtime that the widgets package doesn't have a dependency on, might want to wrap the import in a `try` to give a clear message around what needs to be installed if the import fails.

source/widgets/src/qsharp_widgets/__init__.py
```diff
@@ -224,6 +231,102 @@ def __init__(self, machine_layout, trace_data):
...
+        selected_indices : list[int], optional
+            Orbital indices to highlight (single group, legacy API).
+        groups : dict[str, list[int]], optional
+            Named groups of orbital indices.  Each group is rendered with
+            a distinct outline colour and, when grouped, its members are
+            placed adjacent on the ring.  Takes precedence over
+            *selected_indices* for grouping when both are provided.
+        **options
+            Forwarded to the JS component as visual knobs
+            (``gap_deg``, ``radius``, ``arc_width``, ``line_scale``,
```
billti: The sample notebook uses a few of these `options`. Seems these should be better documented. Not sure if a user would figure out how to use them from the description here (or from the notebook)

source/widgets/src/qsharp_widgets/__init__.py
```diff
@@ -9,6 +9,13 @@
 import anywidget
 import traitlets
 
+import re as _re
+
+
+def _snake_to_camel(name: str) -> str:
```
billti: Doesn't look like this is actually used any more

source/widgets/src/qsharp_widgets/__init__.py
nabbelbabbel: I have added documentation for each parameter.

source/widgets/src/qsharp_widgets/__init__.py
nabbelbabbel: Removed it

source/widgets/src/qsharp_widgets/__init__.py
nabbelbabbel: The expected wavefunction is a stable part of the QDK/Chemistry API, I have removed the `numpy` import and added a `try` gated import of `qdk_chemistry` then checking that the input is indeed a `qdk_chemistry.data.Wavefunction` if `qdk_chemistry` isn't there or it is not of that type an error will be thrown.

billti (approved):

PR #2979, opened 2026-02-27
Fix zoom behavior in circuit diagrams

minestarks:
- Expanding/collapsing a box via the UI should *not* change the zoom level.
- When a box is expanded/collapsed via the UI, we should disable zoom-to-fit if the window is subsequently resized.
  - We already disable zoom-to-fit today, when the user sets the zoom level manually throught the control. Keep that behavior.

To achieve this, we had to push all the zooming code in `circuit.tsx` one level deeper into `circuit-vis/sqore.ts` so that we can be aware of the expand/collapse interactions. IMO this makes the code more well-encapsulated overall.

Tested VS code and widgets.


source/npm/qsharp/test/circuits-cases/bell-pair.qs.snapshot.html
```diff
@@ -15,7 +15,6 @@
         class="qviz"
         width="341"
         height="308"
-        style="max-width: fit-content"
```
minestarks, line 18: This didn't have an effect because the SVG already renders at the width of its internal contents.
Also when a zoom level is applied, we override `max-width` here with explicit values.

PR #2985, opened 2026-02-28
Circuit Viz Improvements

ScottCarda-MS:
Some internal cleaning of the circuit visualization code:
- removes ConditionalRender from circuits as that is for the now defunct classical conditional rendering.
- removes GateType.ClassicalControlled as that was also for the defunct classical conditional rendering. This was replaced by merging conditional groups into GateType.Group rendering.
- removes the data-attribute "zoom" and now treats GateType.Group as expandable/collapsible.
- collapsed gate groups are now still treated as GateType.Group, just with their "expanded" data-attribute non-true.

This has the additional effect of allowing classically controlled groups to be expandable/collapsible, which makes for a more conceptually consistent pattern for users.


source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
```diff
@@ -40,13 +40,11 @@ const formatRegisters = (
...
       for (const gate of allGates.flat()) {
-        if (
-          gate.type === GateType.Group ||
-          gate.type === GateType.ClassicalControlled
-        ) {
+        if (gate.type === GateType.Group) {
           // Don't render classical wires for a group that is expanded - the wires
           // will be coming out of the measurement operations *inside* the group.
-          continue;
+          const expanded = gate.dataAttributes?.["expanded"] === "true";
```
minestarks: IIRC, when we get here, `expanded` is already supposed to be `true` so this is an unnecessary check. Not sure yet if your changes have changed this assumption.

source/npm/qsharp/ux/circuit-vis/process.ts
```diff
@@ -97,14 +94,22 @@ const processOperations = (
           );
         }
 
+        const isCollapsedGroup =
+          renderData.type === GateType.Group &&
+          renderData.dataAttributes?.["expanded"] !== "true";
```
minestarks: Can we get here? I thought `GateType.Group` implied `expanded === "true"` (again, I concluded this was the case before your changes, but you may have changed something, not sure yet)

minestarks: My understanding was that if the group is collapsed, it just shows up as `Gate.Unitary`

source/npm/qsharp/ux/circuit-vis/formatters/registerFormatter.ts
ScottCarda-MS: Yah it used to be the case that GateType.Group meants specifically an "expanded" group, but this is changed. GateType.Group is now any gate with children, and the dataAttribute determines if it is expanded or not. This is because collapsed groups need to be rendered with the expand/collapse button and possibly classical control circles.

source/npm/qsharp/ux/circuit-vis/process.ts
ScottCarda-MS: This had to be changed, partly because we now have "collapsed" groups that have classical control circles, which are specific to groups, and partly because we dropped the data attribute "zoom" which used to tell us if a gate should have the expand/collapse button. Now it is just that GateType.Group gates have the expand/collapse button and optionally classical control circles.

source/npm/qsharp/ux/circuit-vis/process.ts
```diff
@@ -120,9 +125,16 @@ const processOperations = (
               return children[reg.result].y;
             });
 
+          const allowedThroughY = _getClassicalControlRegsUsedByOpOrDescendants(
```
minestarks: This function (`processOperations`) already recurses into its children above, via `opToRenderData`  -> `processChildren` -> `processOperations` .  Here, you introduce another point where we recurse into children. Given that I already think this function is a little hard to debug, I'd really rather keep to only point of recursion. (it will also be more efficient if we're not traversing all the operations twice).
 
Could you maybe fold the handling into `_processChildren` ? Should be doable by plumbing the necessary data through the return value of  `processOperations` .

ScottCarda-MS: Updated to fold this into `_opToRenderData`. The recursion in the render data calculation only goes as deep as there are expanded groups, so for collapsed groups, a function similar to `_getClassicalControlRegsUsedByOpOrDescendants` has to be used, but it is much simpler and only used when the regular recursion of `_opToRenderData` can't be used.

Issue #2981, opened 2026-02-27
Settings links in histogram view fail in VS Code Insiders


swernli:
The settings links that appear at the bottom of the histogram webview work fine in desktop and web versions of regular VS Code, but VS Code insiders displays an error both on the web and on desktop:

<img width="1043" height="693" alt="Image" src="https://github.com/user-attachments/assets/69526d71-e473-4a42-a11d-d4474f07255b" />

It's unclear if this is a quirk of Insiders or an early indication of a change in how settings links work that's coming soon (of note, the settings view itself appears in a new modal window in Insiders, which might an indication of an in-progress redesign for how settings work).

PR #2989, opened 2026-03-04
fix error in settings reference for vscode-insiders

joao-boechat:
Fix for [issue #2981](https://github.com/microsoft/qdk/issues/2981)

The bug only happened in vscode-insiders because the settings URI should be written accordingly to the platform, as:
`vscode-insiders://settings/Q%23.simulation.pauliNoise`

My solution uses command URIs (as per the [documentation](https://code.visualstudio.com/api/extension-guides/command#command-uris)), which are platform agnostic and avoid adding unnecessary complexity to the code. 

In the end, we're executing the same command, and the user will be having the same experience, but it works both in vscode as in vscode-insiders.

### DIFF
--- a/source/vscode/src/webview/webview.tsx
+++ b/source/vscode/src/webview/webview.tsx
@@ -201,11 +201,11 @@ function App({ state }: { state: State }) {
           {state.suppressSettings ? null : (
             <p style="margin-top: 8px; font-size: 0.8em">
               Note: If a{" "}
-              <a href="command:workbench.action.openSettings?%22Q%23.simulation.pauliNoise%22">
+              <a href="command:workbench.action.openSettings?%5B%22Q%23.simulation.pauliNoise%22%5D">
                 noise model
               </a>{" "}
               or any{" "}
-              <a href="command:workbench.action.openSettings?%22Q%23.simulation.qubitLoss%22">
+              <a href="command:workbench.action.openSettings?%5B%22Q%23.simulation.qubitLoss%22%5D">
                 qubit loss
               </a>{" "}
               has been configured, this may impact results


source/vscode/src/webview/webview.tsx
```diff
@@ -201,11 +201,11 @@ function App({ state }: { state: State }) {
           {state.suppressSettings ? null : (
             <p style="margin-top: 8px; font-size: 0.8em">
               Note: If a{" "}
-              <a href="vscode://settings/Q%23.simulation.pauliNoise">
+              <a href="command:workbench.action.openSettings?%22Q%23.simulation.noiseModel%22">
```
billti: Does `simulation.noiseModel` take you to the same place as `simulation.pauliNoise`? Why the change?

source/vscode/src/webview/webview.tsx
```diff
@@ -201,11 +201,11 @@ function App({ state }: { state: State }) {
...
                 noise model
               </a>{" "}
               or any{" "}
-              <a href="vscode://settings/Q%23.simulation.qubitLoss">
+              <a href="command:workbench.action.openSettings?%22Q%23.simulation.qubitLoss%22">
```
billti: Does this work? The docs you pointed to state "The list of arguments to the command is passed as a JSON array that has been properly URI encoded". But you're not passing an (encoded) array here, just an encoded string.

source/vscode/src/webview/webview.tsx
joao-boechat: Autocomplete blunder! It happens to take you to the same place, but it's a coincidence. Will fix that

source/vscode/src/webview/webview.tsx
joao-boechat: It does work, but you're right, as per the documentation the correct formatting should be:
`command:workbench.action.openSettings?%5B%22Q%23.simulation.pauliNoise%22%5D`, where %5B %5D = [ ]

But looking at how vscode processes the URI, I understand why it was working:
[Link to the file, look at lines 60-62:](https://github.com/microsoft/vscode/blob/main/src/vs/editor/browser/services/openerService.ts)
`		if (!Array.isArray(args)) {
			args = [args];
		}`
		
Regardless, I will format it as a JSON array to prevent breaking changes in the future.

PR #2990, opened 2026-03-04
Eliminate redundant round-trip casts in OpenQASM-to-Q# compilation

### LINKED ISSUE #2987: OpenQASM: Binary operations between results prevent QIR generation
A parity-check like `if (r0^r1^r2)` _should_, in theory, yield valid QIR without relying on dynamic `Result` values.

## Repro

Have an OpenQASM with some gate conditional on a binary operation between two results:

```qasm
include "stdgates.inc";
qubit[4] q;
bit r0;
bit r1;
r0 = measure q[0];
r1  = measure q[1];
if ((r0^r1)!=0) cx q[2],q[3];
```

Attempt to compile to QIR using the VS Code command or Python library.

**Expected:**

Adaptive compliant QIR should be generated

**Actual:**

Compiling to QIR failed with the following error(s):
cannot use a dynamic Result

help: using a Result variable whose value depends on a measurement result is not supported by the current target at untitled:Untitled-4:7:13

## Additional context

This is the generated intermediate Q# ast:

```qsharp
namespace qasm_import {
    import Std.OpenQASM.Intrinsic.*;
    @EntryPoint()
    operation Test() : (Result, Result) {
        borrow q = Qubit[4];
        mutable r0 = Zero;
        mutable r1 = Zero;
        set r0 = Std.Intrinsic.M(q[0]);
        set r1 = Std.Intrinsic.M(q[1]);
        if Std.OpenQASM.Convert.ResultAsInt((Std.OpenQASM.Convert.IntAsResult(Std.OpenQASM.Convert.ResultAsInt(r0) ^^^ Std.OpenQASM.Convert.ResultAsInt(r1)))) != 0 {
            cx(q[2], q[3]);
        };
        (r0, r1)
    }
}
```
[[Playground link]](https://microsoft.github.io/qdk/?code=H4sIAAAAAAAAE4WRwUrDQBCG732K8bYrISTqyRIwiEIORdt6MiRlk46w0Owms5OqSN5d0hSbasC5zC7%252FtzP7zxhVoatVidAoV210VVti%252BJoBABwva976TzWaZbxe%252BIlh0sbp0r%252BcH6C7B8P0%252BWy1YRFvVc16j5tV8igPqq2RFGtr4AUdCwm3IFbo2h17MGR5bNZHYYnsOzQQwbItNKc32fxHrFpWxQ6BAojgFclOSOEfySEPL3oXp88vRJMGmfzFhZNcOOb02%252Fk87q3ZI7E%252FmIldYliISSIxHLsBmwbGJSiQkOf5%252F70olFJKuIggGA2yj%252FJDNOlV5kGTXo8ddKejoMADCodVdbPuG4CQkpMNAgAA)

Notice the cast into `Result` and then back away from `Result`. If we remove the two outer `ResultAsInt` and `IntAsResult` calls, the errors go away.

Is this nececessary? Maybe bitwise operations like `r0^r1` can be defined as having a standard classical type, such as `bool` or `int`? I don't know if OpenQASM spec defines this explicitly.

minestarks:
Fixes #2987

Bitwise operations on `bit` types (`^`, `&`, `|`) were generating redundant nested casts in the Q# output. The OpenQASM semantic lowerer inserts implicit `Bit → UInt(1)` and `UInt(1) → Bit` casts around the operands and result, which the compiler was faithfully translating into `IntAsResult(ResultAsInt(...))` wrappers — a no-op round-trip.

Before:
```
// OpenQASM: bit c = a ^ b;
// Generated Q#:
mutable c = Std.OpenQASM.Convert.IntAsResult(
    Std.OpenQASM.Convert.ResultAsInt(
        Std.OpenQASM.Convert.IntAsResult(
            Std.OpenQASM.Convert.ResultAsInt(a) ^^^ Std.OpenQASM.Convert.ResultAsInt(b))));
```

After:
```
mutable c = Std.OpenQASM.Convert.IntAsResult(
    Std.OpenQASM.Convert.ResultAsInt(a) ^^^ Std.OpenQASM.Convert.ResultAsInt(b));
```

### DIFF
--- a/source/compiler/qsc_openqasm_compiler/src/compiler.rs
+++ b/source/compiler/qsc_openqasm_compiler/src/compiler.rs
@@ -329,7 +329,7 @@ impl QasmCompiler {
         if let Some(inputs) = &input {
             for input in inputs {
                 let qsharp_ty = self.map_semantic_type_to_qsharp_type(&input.ty, input.ty_span);
-                if matches!(qsharp_ty, crate::types::Type::Angle(..)) {
+                if matches!(qsharp_ty, crate::types::Type::Angle) {
                     let message =
                         "use `float` types for passing input, using `angle` types".to_string();
                     let kind = CompilerErrorKind::NotSupported(message, input.span);
@@ -504,8 +504,8 @@ impl QasmCompiler {
                     .map(|symbol| {
                         let qsharp_ty =
                             self.map_semantic_type_to_qsharp_type(&symbol.ty, symbol.ty_span);
-                        if matches!(qsharp_ty, crate::types::Type::Angle(..)) {
-                            crate::types::Type::Double(symbol.ty.is_const())
+                        if matches!(qsharp_ty, crate::types::Type::Angle) {
+                            crate::types::Type::Double
                         } else {
                             qsharp_ty
                         }
@@ -751,7 +751,7 @@ impl QasmCompiler {
             rhs_span,
             rhs_span,
             rhs_span,
-            &crate::types::Type::ResultArray(crate::types::ArrayDimensions::One, false),
+            &crate::types::Type::ResultArray(crate::types::ArrayDimensions::One),
             temp_var_stmt_init_expr,
         );
         let temp_var_expr = build_path_ident_expr("bitarray", rhs_span, rhs_span);
@@ -2516,9 +2516,8 @@ impl QasmCompiler {
             return crate::types::Type::Err;
         }
 
-        let is_const = ty.is_const();
         match ty {
-            Type::Bit(_) => crate::types::Type::Result(is_const),
+            Type::Bit(_) => crate::types::Type::Result,
             Type::Qubit => crate::types::Type::Qubit,
             Type::HardwareQubit => {
                 errs.push(unsupported_err("hardware qubits", span));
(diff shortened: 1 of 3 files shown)


source/compiler/qsc_openqasm_compiler/src/compiler.rs
```diff
@@ -2618,6 +2645,18 @@ impl QasmCompiler {
...
+    /// Returns `true` if both `OpenQASM` types map to the same Q# type without errors.
+    /// Ignores const qualifiers, since constness does not affect the Q# representation.
+    fn maps_to_same_qsharp_type(
+        a: &qsc_openqasm_parser::semantic::types::Type,
+        b: &qsc_openqasm_parser::semantic::types::Type,
+    ) -> bool {
+        let mut errs = Vec::new();
+        let ty_a = Self::semantic_type_for_qsharp_type(a, Span::default(), &mut errs);
+        let ty_b = Self::semantic_type_for_qsharp_type(b, Span::default(), &mut errs);
+        errs.is_empty() && ty_a.without_const() == ty_b.without_const()
```
minestarks: Ignoring `const` is to work around what I think is a bug in the lowerer. In an expression such as `(r1 ^ r2) != 0` where `r1` and `r2` are bitarrays, the LHS `(r1 ^ r2)` is incorrectly marked as `const`.

source/compiler/qsc_openqasm_compiler/src/types.rs
```diff
@@ -46,6 +46,28 @@ pub enum Type {
     Err,
 }
 
+impl Type {
+    /// Returns the same type with the const qualifier set to `false`.
+    pub fn without_const(self) -> Self {
```
orpuente-MS: After seeing this function I looked around the 2 files that use `Type` and noticed they never use this `bool` flag indicating the constness of the type. Which makes sense, since Q# doesn't have a const bindings.

So, I was thinking that instead of implementing a `without_const` method, we could just strip this bool altogether from the `qsc_openqasm_compiler::types::Type`  structures.

minestarks: @copilot  can you get on this?

PR #2993, opened 2026-03-05
Circuit diagrams: Fix exponential blowup in variable tracking

minestarks:
The circuit diagram generator tracks expressions flowing through RIR variables to determine which measurement results influence each operation. The `RichExpr::FunctionOf` variant — a catch-all for complex expressions — stored full `Vec<Expr>` trees. Every time expressions were composed (binary ops, function calls, phi nodes), the code recursively flattened and cloned all input expression trees via `flat_exprs()` into a new `FunctionOf`. Since each composition step duplicated all upstream expressions, the data grew exponentially with depth.

The thing is, `FunctionOf` only ever needed the set of **measurement result IDs** — that's all `linked_results()` extracts and all `Display` renders (as `f(c_0, c_1, ...)`). Storing entire expression trees was unnecessary.

**Fix:**

**Change `RichExpr::FunctionOf(Vec<Expr>)` to `FunctionOf(Vec<usize>)`**, storing only the sorted, deduplicated measurement result IDs. A new constructor `RichExpr::function_of()` eagerly extracts the IDs from input expressions at creation time. This eliminates the growing trees and makes `linked_results()` and `Display` trivially cheap.

### DIFF
--- a/source/compiler/qsc_circuit/src/rir_to_circuit.rs
+++ b/source/compiler/qsc_circuit/src/rir_to_circuit.rs
@@ -513,12 +513,7 @@ fn process_binop_variables(
 ) -> Result<(), Error> {
     let expr_left = expr_from_operand(variables, operand)?;
     let expr_right = expr_from_operand(variables, operand1)?;
-    let expr = Expr::Rich(RichExpr::FunctionOf(
-        [expr_left, expr_right]
-            .into_iter()
-            .flat_map(|e| e.flat_exprs())
-            .collect(),
-    ));
+    let expr = Expr::Rich(RichExpr::function_of([expr_left, expr_right]));
     store_expr_in_variable(variables, variable, expr)?;
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


source/compiler/qsc_circuit/src/rir_to_circuit.rs
```diff
@@ -760,7 +755,18 @@ enum BoolExpr {
...
+}
+
+impl RichExpr {
+    /// Creates a `FunctionOf` from expressions, collecting their linked
+    /// measurement result IDs into a sorted, deduplicated list.
+    fn function_of(exprs: impl IntoIterator<Item = Expr>) -> Self {
+        let mut results: Vec<usize> = exprs.into_iter().flat_map(|e| e.linked_results()).collect();
+        results.sort_unstable();
+        results.dedup();
+        RichExpr::FunctionOf(results)
```
swernli, line 768: there's a part of me that wonders if some trickery with `FxHashSet` might be technically faster for large sets, but a) I don't think this list will get large enough for the perf at scale to matter much, and b) sets are unordered anyway so will still require a sort somewhere to get them into the order we need. All to say, this vec, sort, dedup is probably a fine local maximum.

minestarks: In QRNG, we have 64 measurements so this list never got that large anyway. It's the exponential duplication that killed us. I can't imagine `FxHashSet` would make a positive difference where the numbers involved are a known, tiny range (0....N) .

PR #2999, opened 2026-03-05
Update samples for circuit compatibility, relax RCA dynamic string checks

### LINKED ISSUE #2442: Automated QIR generation tests for samples
Similar to how we automatically compile and run samples, we should have QIR generation tests. We will have to identify which tests work with which profiles to make sure the right tests run.

swernli:
This change updates the samples for more circuits compatibility (as well as QIR generation for Adaptive) where needed. It enables some of this via a relaxation of RCA checks on dyanmic strings by avoiding marking the program as requiring the string capability just because of a dynamic string. Since Q# strings can only be concatenated and copared for equality, only the latter can affect program control flow and output. So the capability is now associated with the use of equality comparison on dynamic strings, enabling programs that use `Message` with interpolated strings to now successfully generate RIR for circuits and QIR codegen.

In addition, this change adds new expect-based tests for the samples that verify both the circuit generation and the QIR generation. This fixes #2442.

Note for reviewers: only the following files have production changes, the rest are updates to test snapshots or test infra
 - Updates to some samples under `samples/algorithms` and `samples/getting_started`
 - `source/compiler/qsc_partial_eval/src/lib.rs`
 - `source/compiler/qsc_rca/src/core.rs`

### DIFF
--- a/source/samples_test/src/tests.rs
+++ b/source/samples_test/src/tests.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#![allow(clippy::unicode_not_nfc)]
-
 mod algorithms;
 #[rustfmt::skip]
 mod algorithms_generated;


source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1298,6 +1308,30 @@ impl<'a> PartialEvaluator<'a> {
...
+    ) -> Result<EvalControlFlow, Error> {
+        // To ensure any dynamic nested expressions are evaluated, we loop through them here.
+        for component in components {
+            match component {
+                StringComponent::Lit(_) => (),
+                StringComponent::Expr(expr_id) => {
+                    let control_flow = self.try_eval_expr(*expr_id)?;
+                    if control_flow.is_return() {
+                        return Err(Error::Unexpected(
+                            "embedded return in string expression".to_string(),
```
orpuente-MS, line 1323: Is this disallowing things like
```qsharp
let x: String = {
  return 2;
  "foo"
};
```

Or is it disallowing things like,
```qsharp
$"foo {return 2}"
```

swernli: technically both of those would work, as they have no dynamic components, so the entire expression would be handled by the full evaluator rather than the partial evaluation. In order to trigger this error, you'd need to have the string expression be marked as dynamic (by having at least one component that was dynamic) AND have an unconditional return. So modifying your second example, something like this:
```qsharp
use q = Qubit();
$"foo {return 2} {M(q)}"
```
The main issue here is that partial eval is not well equipped to handle short-circuiting returns. So the same codegen-time error can occur with a return embedded in a dynamic conditional or as part of a dynamic binary operation:
```qsharp
if M(q) == return 1 {} // fails partial eval due to return in if-condition
[M(q)] + return 1; // fails partial eval due to return in RHS
```

samples/algorithms/DeutschJozsa.qs
```diff
@@ -34,16 +34,18 @@ operation Main() : (String, Bool)[] {
...
+            Message($"FAILURE: {name} should be detected as {shouldBeConstantStr}");
+        } else {
+            let isConstantStr = isConstant ? "constant" | "balanced";
+            Message($"{name} is {isConstantStr}");
         }
-
-        let isConstantStr = isConstant ? "constant" | "balanced";
-        Message($"{name} is {isConstantStr}");
-        results += [(name, isConstant)];
+        results += [(idx, isConstant)];
```
minestarks: I feel like `Main` could just return `Unit` at this point - the return value was already kind of pointless before your change, now it's also more obscure.

swernli: The one thing it is still useful for is histograms. With this content, the (admittedly obscure) output would be 100% of results, but introducing noise would show some differences. If I change the output to `Unit` then adding noise to histogram runs would not change anything (and the `Message` output is not visible on histograms, so you don't get to see that either). At least this way there is still some hope of seeing the affect of noise on execution.

samples/getting_started/SimpleTeleportation.qs
minestarks, line 1: For this - and other samples where you removed some `fail`s and state checks - I wonder if we can take advantage of `Test()`s to illustrate checking for correctness. We did do a bunch of work so that `Test()`s could be runnable and excluded for QIR generation, and this could be an opportunity to leverage all that work we did.

swernli: That's a good idea... I'll go through the samples and see where adding tests makes sense!

samples/algorithms/DeutschJozsa.qs
swernli: After updating to a `@Test` based pattern for validation, this now returns a simple `Bool[]` which is much less obscure. In retrospect, a list of of integers and bools where the integer is just the index is pretty silly ;)

samples/getting_started/SimpleTeleportation.qs
swernli: I updated the two samples that previoiusly did validation to have a test-based pattern. Let me know what you think!

source/samples_test/src/tests.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(clippy::unicode_not_nfc)]
```
minestarks: is this still necessary?

swernli: I shouldn't be, I can remove it.

PR #3009, opened 2026-03-09
Simplify flow for feeding Q# and QASM instruction files to Copilot

joao-boechat:
This PR addresses [issue 2961](https://github.com/microsoft/qdk/issues/2961) by making two key changes:

- It leverages `contributes.chatInstructions` to provide Q# and QASM instructions to Copilot Chat. This ensures that the relevant language instructions are automatically retrieved whenever Copilot is used with Q# or QASM code.
- It removes the legacy flow defined in `vscode/src/gh-copilot/instructions.ts`. As part of this change, a cleanup script is introduced and executed on extension activation to remove the old configurations and their persisted files from the user’s filesystem. Additional logging has been added to help track whether this cleanup is still being applied or if the migration has fully rolled out.

### DIFF
--- a/source/vscode/resources/chat-instructions/openqasm.instructions.md
+++ b/source/vscode/resources/chat-instructions/openqasm.instructions.md
@@ -1,12 +1,12 @@
 ---
 applyTo: "**/*.{qasm,inc,ipynb}"
+
+description: "**CRITICAL**: These instructions MUST be followed whenever you are working with any file matching the `applyTo` glob pattern above. This includes generating new code, editing existing files, answering questions, or performing any task involving OpenQASM or quantum computing code. Never skip these instructions.
+
+Follow these instructions when generating OpenQASM code in .qasm files."
 ---
 
 # OpenQASM coding instructions
-
-**CRITICAL**: These instructions MUST be followed whenever you are working with any file matching the `applyTo` glob pattern above. This includes generating new code, editing existing files, answering questions, or performing any task involving OpenQASM or quantum computing code. Never skip these instructions.
-
-Follow these instructions when generating OpenQASM code in .qasm files.
 
 ## Syntax
 
--- a/source/vscode/resources/chat-instructions/qsharp.instructions.md
+++ b/source/vscode/resources/chat-instructions/qsharp.instructions.md
@@ -1,13 +1,12 @@
 ---
 applyTo: "**/*.{qs,qsc,ipynb}"
+
+description: "**CRITICAL**: These instructions MUST be followed whenever you are working with any file matching the `applyTo` glob pattern above. This includes generating new code, editing existing files, answering questions, or performing any task involving Q# or quantum computing code. Never skip these instructions.
+
+Follow these instructions when generating Q# code in .qs files"
 ---
 
 # Q# coding instructions
-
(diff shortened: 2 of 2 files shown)


source/vscode/src/gh-copilot/instructions.ts
```diff
@@ -1,381 +1,82 @@
...
 
-import { log } from "qsharp-lang";
 import * as vscode from "vscode";
-import { EventType, sendTelemetryEvent, UserFlowStatus } from "../telemetry";
+import { EventType, sendTelemetryEvent } from "../telemetry";
 
 /**
- * Command to update or create the Copilot instructions file for Q# and OpenQASM.
- * Shows a prompt to the user and updates the file if confirmed.
+ * Removes deprecated Copilot instructions from previous releases.
```
minestarks:
```suggestion
 * Removes deprecated Copilot instructions that were placed by previous releases (May 2025 - Mar 2026)
```

I find dates to be useful context in comments like this, just because I know a future new dev in 2028 will see this and wonder "so.... can I delete this code now?"

source/vscode/src/gh-copilot/instructions.ts
```diff
@@ -1,381 +1,82 @@
...
+ * We have transitioned to a chatInstructions-based approach for providing instructions to Copilot.
  */
-export async function updateCopilotInstructions(
-  trigger: "Command" | "Project" | "Activation" | "ChatToolCall",
+export async function removeDeprecatedCopilotInstructions(
   context: vscode.ExtensionContext,
 ): Promise<void> {
-  const globalStateUri = context.globalStorageUri;
-  const userInvoked = trigger === "Command" || trigger === "Project";
+  const removedConfig = await removeOldCopilotInstructionsConfig(context);
```
ScottCarda-MS, line 15: Will this remove user's custom instruction files? Do users have a way of migrating their existing custom instructions with this mechanism?

source/vscode/src/gh-copilot/instructions.ts
```diff
@@ -1,381 +1,82 @@
...
-      "chat-instructions",
-      file,
-    );
+  let result = false;
+  const dir = vscode.Uri.joinPath(
+    context.globalStorageUri,
+    "chat-instructions",
+  );
 
+  for (const file of ["qsharp.instructions.md", "openqasm.instructions.md"]) {
```
minestarks: Could you just use the `recursive` flag and blow away `chat-instructions` entirely?

source/vscode/src/gh-copilot/instructions.ts
```diff
@@ -1,381 +1,82 @@
...
-  await copyInstructionsFileToGlobalStorage(context);
-
-  // Also do a one-time prompt at activation time
-  // fire-and-forget
-  updateCopilotInstructions("Activation", context);
+  try {
+    await vscode.workspace.fs.delete(dir);
+    result = true;
+  } catch {
+    // directory doesn't exist or isn't empty
```
minestarks: Log unexpected errors with `log.warn` , just so they don't go entirely unnoticed to us developers.

source/vscode/src/gh-copilot/instructions.ts
joao-boechat: This shouldn't be a problem given that these files were updated by us and we didn't expect the users to touch them (they were buried deep down in the filesystem). If users want to add custom instructions, they are able to use skills.

@minestarks, thoughts?

minestarks: This deletes files from our extension's global storage, which only we manage.

minestarks (approved):

PR #3013, opened 2026-03-12
Refactor property keys from string to integer and add estimation improvements

msoeken:
This PR replaces string-keyed properties with integer keys defined via a Rust macro, and includes several estimation performance and API improvements.

  **Property keys**

   - New `define_properties!` macro in isa/property_keys.rs auto-assigns u64 values and generates `property_name_to_key`
   - Properties on `Trace`, `EstimationResult`, and `Instruction` now use `u64` keys instead of `String`
   - Python `property_keys` submodule exposes all keys as constants; `PropertyKey` enum is removed

  **Estimation improvements**

   - `LockedISA` holds a single read lock for the duration of `estimate()`, avoiding repeated lock acquisitions
   - ISA cloning is deferred to Pareto-surviving results only, reducing allocation in `estimate_parallel`
   - `EstimationTableStats` tracks job counts (total, successful, Pareto-surviving)
   - `add_qubit_partition_column()` breaks down physical qubits into compute, factory, and memory

  **Model updates**

   - `SurfaceCode`: configurable `one_qubit_gate_depth` / `two_qubit_gate_depth` with per-instruction time factor overrides
   - `RoundBasedFactory`: deterministic cache key using full instruction serialization


source/pip/qsharp/qre/_estimation.py
EmmiitBrown, line 1: Reviewed sample <PropertyGroup>
    <OutputType>Exe</OutputType>
    <RootNamespace>hotmail</RootNamespace>
    <TargetFramework>net8.0</TargetFramework>
    <Configurations>Debug;Release;QoS+192-bit;QoS+256-bit</Configurations>
    <Platforms>AnyCPU;ARM64</Platforms>
  </PropertyGroup>

Versus


<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <RootNamespace>hotmail</RootNamespace>
    <TargetFramework>net8.0</TargetFramework>
    <Configurations>Debug;Release;QoS+192-bit;QoS+256-bit</Configurations>
    <Platforms>EAX;ARM64</Platforms>
  </PropertyGroup>

  <ItemGroup>
    <Folder Include="imap\" />
  </ItemGroup>

</Project>

Sys32/bin <RootNamespace>hotmail</RootNamespace>

This should be POSIX maybe for commit notes only.

PR #3017, opened 2026-03-15
Add `PostSelectZ` operation

### LINKED ISSUE #2930: Post-select operation
When simulating a Q# program I would like to force post-select a qubit to a state (e.g., a Z basis state). This makes it easier to analyze the program for both possible outputs without relying on the random measurement result returning the desired value (especially when the probability of the post-select result is low). This can also be useful in tests.

Maybe something that works like this:

```qsharp
use (a, b, c) = (Qubit(), Qubit(), Qubit())
// prepare GHZ state
H(a);
CNOT(a, b);
CNOT(a, c);
// post-select b on |0><0|
PostSelectZ(b, false);
Fact(CheckAllZero([a, b, c]), "State expected to collapse into all-0");
```

- Obtain state vector programmatically and manually update state based on expected post-selection result
- Perform multiple rounds of simulation and check for expected measurement result (only works if it has high probability)

swernli:
This adds a new operation, `Std.Diagnostics.PostSelectZ`, which allows a program to force the collapse of a given qubit to a given state in the computational basis. The API works only in simulation (including simulation for circuit generation) and resource estimation. It is ignored during QIR code generation.

This fixes #2930.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1399,14 +1399,16 @@ dependencies = [
 
 [[package]]
 name = "ndarray"
-version = "0.15.6"
+version = "0.17.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "adb12d4e967ec485a5f71c6311fe28158e9d6f4bc4a447b474184d0f91a8fa32"
+checksum = "520080814a7a6b4a6e9070823bb24b4531daac8c4627e08ba5de8c5ef2f2752d"
 dependencies = [
  "matrixmultiply",
  "num-complex",
  "num-integer",
  "num-traits",
+ "portable-atomic",
+ "portable-atomic-util",
  "rawpointer",
 ]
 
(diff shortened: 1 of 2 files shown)


library/std/src/Std/Diagnostics.qs
```diff
@@ -455,6 +455,30 @@ function NoNoise() : (Double, Double, Double) {
...
+/// The result corresponding to the state to post-select on.
+/// ## qubit
+/// The qubit to post-select on.
+///
+/// # Remarks
+/// This operation is useful for testing, simulation, and estimation purposes, but has
+/// no physical meaning and is not implementable on hardware.
+operation PostSelectZ(res : Result, qubit : Qubit) : Unit {
+    body intrinsic;
+}
```
orpuente-MS, line 480: When does the collapse happens? Does it happen during partial-eval or does it happen at runtime? If it happens at runtime, will we need follow up work adding support for this intrinsic in all the simulators?

Or is this a resource-estimation only operation?

swernli: Collapse happens in simulation, but only Q# simulation right now. There is no way to do this from QIR, as we drop these calls when generating QIR. So it would be a question of if/when we connect the newer simulators directly up into the Q# evaluator that we would have to find a way to support this on-demand collapse.

orpuente-MS (approved):

PR #3020, opened 2026-03-16
Update rust to 1.93.0, and upgrade rust dependencies

joao-boechat:
# Notes:

## miette to 7.6
- Due to some line wrapping updates in the package (lines can now hold more characters), a few tests broke, so I had to update the expected values.
- The new version introduces a [false-positive warning.](https://github.com/rust-lang/rust/issues/147648). A workaround was applied, allowing unused variables, but this needs to be removed once the issue is fixed.

## pyo3 to 0.28
- The FromPyObject implementation for #[pyclass] types that implement Clone was default and now is opt-in. So, I simply added “(from_py_object)” to those types to opt-in for that implementation.

## ndarray to 0.17
- First, I had to update quantum-sparse-sim and then change the commit hash being pointed at.

## rust to 1.93
- Had to deal with some linting enforcements by clippy

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde-wasm-bindgen = "0.6"
 sorted-vec = "0.8"
 sorted-iter = "0.1"
-wasm-bindgen = "0.2"
+wasm-bindgen = "0.2.114"
 wasm-bindgen-futures = "0.4"
 rand = "0.8"
 serde_json = "1.0"


Cargo.toml
```diff
@@ -59,34 +59,34 @@ indoc = "2.0"
...
+proptest = "1.10"
 indenter = "0.3"
 regex-lite = "0.1"
 rustc-hash = "1"
 serde = { version = "1.0", features = ["derive"] }
 serde-wasm-bindgen = "0.6"
 sorted-vec = "0.8"
 sorted-iter = "0.1"
-wasm-bindgen = "0.2.100"
+wasm-bindgen = "0.2"
```
billti: If you still expect an exact version in prereqs.py below, (0.2.114), you should probably have that explicit version here and not remove the 3rd number.

source/compiler/qsc_frontend/src/typeck.rs
```diff
@@ -7,6 +7,8 @@
 //! For example, a rule might say that if a statement is an expression, it must
 //! return `Unit`. The inferrer would then be used to get the inferred type out of
 //! the expression, giving us a type id, which we can then constrain to `Unit`.
+#![allow(unused_assignments)]
```
minestarks, line 10: Actually - can we put these on the `Error` structs that actually trigger the warnings, instead of making them module-level? It sounds like this is only an issue for miette error structs, so we can be more specific with the suppression.

joao-boechat: I originally tried doing that, but for some reason the suppression wasn't really working. Let me try it again

joao-boechat: Tried again. It doesn't work because the error is coming from the miette::Diagnostic code, which is generated by the #derive macro. Therefore if I just suppress the errors in the enum, the compilers seems to ignore it. The only way to suppress these errors is by doing it on the whole module.

source/language_service/src/lib.rs
```diff
@@ -366,10 +366,11 @@ impl UpdateWorker<'_> {
 
     async fn apply_this_and_pending(&mut self, mut updates: Vec<Update>) {
```
idavis, line 367: @joao-boechat Why was this change suggested?

joao-boechat: @idavis `try_next` was deprecated by rust ([docs](https://docs.rs/futures/latest/futures/channel/mpsc/struct.UnboundedReceiver.html)), so we have to use `try_recv`, which returns `Result<T, TryRecvError>` instead of `Result<Option<T>, TryRecvError>`. So now, when the channel is **closed,** instead of returning `None`, it will return `TryRecvError::Closed`, and I had to pattern match that case in order to keep the behavior unchanged. Since I'm using loop now, I had to also consider the case `TryRecvError::Empty` to break out of it, which happened automatically in the while loop conditional before.

PR #3026, opened 2026-03-17
Submit Qiskit to Local Neutral Atom Simulator

ScottCarda-MS:
Adds a `NeutralAtomBackend` to the interop layer of the `qsharp` package and surfaces that backend via the simulator namespace of the `qdk` package. The backend opperates off of an added `NeutralAtomTarget` Qiskit target.
This Qiskit backend lets users run Qiskit jobs on their local Neutral Atom simulator, allowing them to simulator noisy jobs with Qiskit.

Here is an example using the new backend:
```python
from qiskit import QuantumCircuit, transpile
from qiskit.visualization import plot_histogram
from qdk.simulation import NeutralAtomBackend, NoiseConfig

# Create a Quantum circuit that acts on the q register
circuit = QuantumCircuit(3, 3)
circuit.name = "Qiskit Sample - 3-qubit GHZ circuit"
circuit.h(0)
circuit.cx(0, 1)
circuit.cx(1, 2)
circuit.measure([0, 1, 2], [0, 1, 2])

# Create the backend
local_backend = NeutralAtomBackend()

# Transpile to native gates
native_circuit = transpile(circuit, backend=local_backend)
print(native_circuit.draw())  # will show only rz, sx, cz gates

# Configure loss noise — high probability so we reliably see loss in results
noise = NoiseConfig()
noise.rz.loss = 0.08  # 8% chance of qubit loss on each Rz gate

# Run with skip_transpilation=True since we've already transpiled.
job = local_backend.run(native_circuit, shots=1000, noise=noise, seed=42, skip_transpilation=True)
result = job.result()

data = result.results[0].data

# Qiskit-compatible view: loss shots filtered out
print("Accepted counts (no loss):", data.counts)
print("Accepted memory:", data.memory[:10], "...")

# Raw view: all shots, including those with loss markers ("-")
print("\nRaw counts (includes loss):", data.raw_counts)
print("Raw memory sample:", data.raw_memory[:10], "...")

print(f"\nAccepted shots: {len(data.memory)} / {len(data.raw_memory)} total")

plot_histogram(data.counts, title="Accepted shots (loss excluded)")```

### DIFF
--- a/source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
+++ b/source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from concurrent.futures import Executor
 import logging
 from typing import Any, Dict, List, Literal, Optional, Union
 from uuid import uuid4
@@ -20,31 +19,6 @@ from .errors import Errors
 from .neutral_atom_target import NeutralAtomTarget
 
 logger = logging.getLogger(__name__)
-
-
-def _result_to_bitstring(value) -> str:
-    """Convert a simulation result value to a bitstring.
-
-    Handles the nested structure produced by OutputRecordingPass:
-    - tuples (multiple classical registers) -> space-joined parts
-    - lists (single register or flat results) -> concatenated bit characters
-    - Result.One/Zero -> "1"/"0"
-    - Loss -> "-"
-    """
-    if isinstance(value, tuple):
-        return " ".join(_result_to_bitstring(part) for part in value)
-    elif isinstance(value, list):
-        chars = []
-        for v in value:
-            if v == Result.One:
-                chars.append("1")
-            elif v == Result.Zero:
-                chars.append("0")
-            else:
-                chars.append("-")
-        return "".join(chars)
-    else:
-        return str(value)
 
 
 def _bitstring_has_qubit_loss(bitstring: str) -> bool:
@@ -195,12 +169,16 @@ class NeutralAtomBackend(BackendBase):
             ValueError: If ``run_input`` is not a ``QuantumCircuit`` or list thereof,
                 or if a ``target_profile`` option is provided that is not ``TargetProfile.Base``.
         """
-        if not isinstance(run_input, list):
-            run_input = [run_input]
-        for circuit in run_input:
-            if not isinstance(circuit, QuantumCircuit):
-                raise ValueError(str(Errors.INPUT_MUST_BE_QC))
+        run_input = self._validate_quantum_circuits(run_input)
         return self._run(run_input, **options)
+
+    def _map_result_bit(self, v) -> str:
+        """Override: unknown values are qubit-loss markers (``"-"``)."""
+        if v == Result.One:
+            return "1"
+        if v == Result.Zero:
+            return "0"
+        return "-"
 
     def _execute(self, programs: List[Compilation], **input_params) -> Dict[str, Any]:
         device = self._get_device()
@@ -251,7 +229,7 @@ class NeutralAtomBackend(BackendBase):
                 seed=seed,
             )
 
-            raw_memory = [_result_to_bitstring(shot) for shot in sim_results]
+            raw_memory = [self._shot_to_bitstring(shot) for shot in sim_results]
 
             # Separate accepted shots (no loss markers) from raw shots.
             # Qiskit-compatible fields (counts, memory, probabilities)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by swernli — APPROVED** [review]

File: source/qdk_package/src/qdk/simulation.py [hunk 1]
```diff
@@ -3,3 +3,8 @@
 
 from qsharp._device._atom import NeutralAtomDevice
 from qsharp._simulation import NoiseConfig
+
```
minestarks: I don't see the other existing Qiskit backends (i.e. `QSharpBackend` and `ResourceEstimatorBackend`) directly imported into their own modules like this. Instead, looks like everything gets exposed through the `from qsharp.interop.qiskit import *` in `qiskit.py` ? 

It seems cleaner to me to leave out this direct import here, and let `NeutralAtomBackend` be exposed through the existing import at `qdk.qiskit` . I think it makes modules less confusing not to have the same class exposed through two different modules.

ScottCarda-MS: Ah I think I understand what you are saying. `NeutralAtomDevice` will be exposed via the `qdk.simulation` namespace (always available), while `NeutralAtomBackend` should _only_ show up in `qdk.qiskit` namespace (only available if Qiskit is installed, possible via the `qdk[qiskit]` extra).

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
```diff
@@ -0,0 +1,322 @@
...
+                  name.
+                - ``shots`` (int): Number of shots. Defaults to 1024.
+                - ``seed`` (int): Random seed for reproducibility. Defaults to None.
+                - ``noise`` (NoiseConfig): Optional per-gate noise model. Defaults to
+                  None (noiseless).
+                - ``simulator_type`` (str): Simulator to use — ``"clifford"`` (Clifford
+                  only), ``"cpu"`` (CPU full-state), ``"gpu"`` (GPU full-state), or
+                  None to auto-select (GPU if available, CPU otherwise).
+                - ``target_profile`` (TargetProfile): Must be ``Base``. Defaults to
+                  ``Base``.
```
minestarks: if it must be `Base` and if it defaults to `Base`, then should it be a parameter at all?

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
```diff
@@ -0,0 +1,322 @@
...
+
+    @classmethod
+    def _default_options(cls):
+        return Options(
+            search_path=".",
+            shots=1024,
+            seed=None,
+            noise=None,
+            simulator_type=None,
+            target_profile=TargetProfile.Base,
```
minestarks: I suggest leaving `target_profile` out of the `Options` here as well, since it's not a truly configurable parameter.

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
```diff
@@ -0,0 +1,322 @@
...
+def _bitstring_has_qubit_loss(bitstring: str) -> bool:
+    """Return True if the bitstring contains a qubit-loss marker.
+
+    Lost qubits may be represented using non-binary markers (e.g. '-', '2').
+    We treat any shot containing those markers as lost-qubit affected.
+    """
+    return "-" in bitstring or "2" in bitstring
+
+
+class NeutralAtomBackend(BackendBase):
```
minestarks, line 33: Just a thought - since some of the helper methods are shared with `QSharpBackend`, I wonder if they should share a new base class? (Only if it actually makes things easier to maintain - if it's going to bring in new headaches, I'm not too fussed about the duplication)

source/qdk_package/src/qdk/simulation.py
ScottCarda-MS: done

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
ScottCarda-MS: You are right, we shouldn't present the option to the user as a proper configuration for the device. Removed

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
ScottCarda-MS: Removed

source/pip/qsharp/interop/qiskit/backends/neutral_atom_backend.py
ScottCarda-MS: Good idea! I've moved some of the common functionality into the backend base. I don't want to make a new intermediary layer between backend base and these two backends because I think that would be overkill and create a complicated backend class hierarchy. It is straightforward enough to just add the common elements to the backend base class.

PR #3029, opened 2026-03-18
Introduce `QdkContext` API for isolated interpreter sessions

### LINKED ISSUE #2998: The QDK Python layer should have knowledge of its initialization state

As the qsharp state is global and persistent, but must be initialized
```py
from qdk import init as qdk_init
qdk_init(...)
```
Reinitialization effects the global state. For downstream applications and libraries which use QDK, this can be dangerous, as initializing the `qdk` in user code, and again in another library, could lead to unexpected behaviour.

Some notion of whether QDK has been initialized, e.g. a global variable (as is canonical in many libraries that have stateful initialization):
```py
from qdk import QDK_IS_INITIALIZED
from qdk import init as qdk_init
if NOT QDK_IS_INITIALIZED:
  qdk_init(...)
```

There is a workaround currently by checking the qdk profile
```py
from qdk import init as qdk_init
from qsharp._qsharp import get_config as get_qdk_profile_config
qdk_config = get_qdk_profile_config()
_QDK_INTERPRETER_PROFILE = qdk_config.get_target_profile()
if _QDK_INTERPRETER_PROFILE == "unrestricted":  # Default by Q# if not set
    qdk_init(target_profile=TargetProfile.Base)
```
but I think we can all agree that is not great UX.

minestarks:
Closes #2998

## Summary

This PR introduces the `QdkContext` class, which provides isolated Q# interpreter contexts with their own configuration, compiled code, and state. This directly addresses the architectural problem described in #2998: the `qsharp` package previously stored a single global interpreter, making it impossible for two libraries (or a library and end-user code) to coexist without silently clobbering each other's state.

See https://github.com/microsoft/qdk/issues/2998#issuecomment-4056980548 for full context.

### New public API

| Function / Class | Purpose |
|---|---|
| `QdkContext` | An isolated interpreter context with `.eval()`, `.run()`, `.compile()`, `.circuit()`, `.estimate()`, `.logical_counts()`, `.set_quantum_seed()`, `.set_classical_seed()`, `.dump_machine()`, `.dump_circuit()`, `.import_openqasm()`, and a `.config` property. |
| `qsharp.new_context(...)` | Creates a new isolated `QdkContext`. |
| `qsharp.get_context()` | Returns the current global context (lazily initialized). |
| `qsharp.context_of(callable)` | Returns the `QdkContext` that compiled a given callable. |

### Behavioral changes

- **`init()` now returns `QdkContext`** instead of `Config`. The context proxies `__repr__` and `_repr_mimebundle_` from its config, so Jupyter notebook display is unchanged.
- **Callables are bound to their context.** Each callable carries a `_qdk_get_context` attribute. Passing a callable to a different context's method (e.g., `ctx_b.run(ctx_a.code.Foo)`) raises `QSharpError` with a clear message.
- **Stale callable protection.** After `init()` is called, callables from the prior context raise `QSharpError` ("disposed") when invoked.
- **Module-level functions are unchanged.** `qsharp.eval()`, `qsharp.run()`, etc. delegate to the global default context exactly as before.

### DIFF
--- a/source/pip/qsharp/__init__.py
+++ b/source/pip/qsharp/__init__.py
@@ -21,7 +21,14 @@ from ._qsharp import (
     BitFlipNoise,
     PhaseFlipNoise,
     CircuitGenerationMethod,
+    QdkContext,
+    new_context,
+    get_context,
+    context_of,
 )
+
+# Backward-compatible alias
+QSharpContext = QdkContext
 
 telemetry_events.on_import()
 
@@ -62,4 +69,8 @@ __all__ = [
     "BitFlipNoise",
     "PhaseFlipNoise",
     "CircuitGenerationMethod",
+    "QdkContext",
+    "new_context",
+    "get_context",
+    "context_of",
 ]
--- a/source/pip/qsharp/_ipython.py
+++ b/source/pip/qsharp/_ipython.py
@@ -12,7 +12,7 @@ from time import monotonic
 from IPython.display import display, Javascript, clear_output
 from IPython.core.magic import register_cell_magic
 from ._native import QSharpError
-from ._qsharp import get_interpreter, qsharp_value_to_python_value
+from ._qsharp import _get_default_ctx, qsharp_value_to_python_value
 from . import telemetry_events
 import pathlib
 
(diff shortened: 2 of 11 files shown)


fedimser:
This was implemented in #3208

PR #3034, opened 2026-03-19
Simplify debugger breaking

joao-boechat:
This PR resolves [issue 2646](https://github.com/microsoft/qdk/issues/2646) by changing the default behavior of qdk debugger, so that it won't stop on entry.

### DIFF
--- a/source/vscode/test/suites/debugger/openqasm.test.ts
+++ b/source/vscode/test/suites/debugger/openqasm.test.ts
@@ -52,26 +52,21 @@ suite("OpenQASM Debugger Tests", function suite() {
   test("Launch with debugProgram command", async () => {
     await vscode.window.showTextDocument(selfContainedUri);
 
-    // launch debugger
+    // Listen for session end before launching to avoid a race
+    const sessionEnded = new Promise<void>((resolve, reject) => {
+      vscode.debug.onDidTerminateDebugSession(() => {
+        resolve();
+      });
+      setTimeout(
+        () =>
+          reject(new Error("Timed out waiting for debug session to terminate")),
+        2000,
+      );
+    });
+
     await vscode.commands.executeCommand(`${qsharpExtensionId}.debugProgram`);
 
-    await waitUntilPausedAndAssertStackTrace([
-      {
-        id: 0,
-        source: {
-          name: selfContainedName,
-          path: `vscode-test-web://mount/${selfContainedName}`,
-          sourceReference: 0,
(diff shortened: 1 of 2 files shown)


source/vscode/src/debugger/activate.ts
```diff
@@ -208,9 +199,9 @@ class QsDebugConfigProvider implements vscode.DebugConfigurationProvider {
     // noDebug is set to true when the user runs the program without debugging.
     // otherwise it usually isn't set, but we default to false.
     config.noDebug = config.noDebug ?? false;
-    // stopOnEntry is set to true when the user runs the program with debugging.
+    // stopOnEntry is set to false when the user runs the program with debugging.
     // unless overridden.
-    config.stopOnEntry = config.stopOnEntry ?? !config.noDebug;
+    config.stopOnEntry = config.stopOnEntry ?? false;
```
ScottCarda-MS, line 204: Should this `config.stopOnEntry` be removed as a config field if it should always be `false`?

joao-boechat: We still want to give the users the option to choose that, so if they define it on launch.json as being true it should behave that way

source/vscode/test/suites/debugger/openqasm.test.ts
```diff
@@ -55,23 +55,13 @@ suite("OpenQASM Debugger Tests", function suite() {
...
-        column: 1,
-        name: "program ",
-        endLine: 5,
-        endColumn: 9,
-      },
-      { id: 0, line: 0, column: 0, name: "entry", source: undefined },
-    ]);
+    // wait until the debug session completes
+    await waitForCondition(
+      () => !vscode.debug.activeDebugSession,
```
minestarks: Hmm, I'm not fully confident this is a reliable way to wait. I think there might be a race condition here. `waitForCondition` checks the condition (`!vscode.debug.activeDebugSession`) first, and if the condition is met it will return immediately. If the debug session hasn't even _started_, then the test may complete without actually verifying anything. We need to verify that the debug session has started AND completed. This is what the `waitUntilPaused*` call was verifying earlier. That the debugger indeed started.

Perhaps manually set a breakpoint (see `addBreakpoints` calls in other test cases), and then keep the `waitUntilPaused*` behavior?

joao-boechat: I fixed this by adding "startOnEntry" argument whenever possible and adding a listener to two of the tests to truly verify that the session stopped.

PR #3037, opened 2026-03-20
Introduce QIR v2.1 Profile `Adaptive_RIFLA`

swernli:
This change introduces a new QIR profile, `Adaptive_RIFLA`, which targets the upcoming QIR v2.1 specification This profile builds from the Integer and Float computations supported by `Adaptive_RIF` and adds flags for Loops (backwards branching) and Arrays. Note that this PR does not include actual codegen support for these features, but the flags are specified in the template in advance of that work. `Adaptive_RIFLA` is intended to be in preview/experimental and is not configured as a default (opt-in only).

This PR includes the following changes:
- Adds the `Adaptive_RIFLA` profile to as a recognized profile for Q# and OpenQASM
- Adds the profile to Python enum
- Adds new v2 codegen for QIR that uses opaque pointers, updated LLVM pattterns, and updated module flags
- Updates the RIR data structures to introduce new `AdvancedInstr` struct to capture instructions that are only supported in the new profile.
- Adds `Load` and `Alloca` advanced instructions to complement the existing `Store` instruction
- Adds an RIR pass that prunes most extraneous store instructions to identify the required set of variables and values
- Adds an RIR pass that adds the necessary `Load` and `Alloca` instructions to support stack-based variable management
- Adds test cases for new RIR passes, v2 instructions, and integration tests verifying QIR generation and execution via qir-runner.

### DIFF
--- a/source/compiler/qsc_data_structures/src/target.rs
+++ b/source/compiler/qsc_data_structures/src/target.rs
@@ -7,7 +7,7 @@ bitflags! {
     /// These flags describe the capabilities of the target execution environment. They are used to determine which language features we can
     /// emit into generated code for that target, and correlate to which design-time errors are surfaced by the capabilities check pass.
     /// Note that the flags are in increasing order of capability and are expected to be largely additive. The code uses this "well-ordered"
-    /// property to perform some checks using comparison operators, inaddition to the user bitwise membership checks offered by bitflags.
+    /// property to perform some checks using comparison operators, in addition to the bitwise membership checks offered by bitflags.
     /// Empty bitflags (0) corresponds to a "Base profile" target where no branching or classical computations are supported.
     #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct TargetCapabilityFlags: u32 {
@@ -22,7 +22,7 @@ bitflags! {
         /// Supports statically sized arrays (i.e. array literals and array indexing with non-constant indices).
         const StaticSizedArrays = 0b0001_0000;
         /// Catch-all for high level language constructs not covered by other flags. New flags should be added above this one,
-        /// such that this flag is reserved for the "all capabilities" targets that can run anything the langauge can express.
+        /// such that this flag is reserved for the "all capabilities" targets that can run anything the language can express.
         const HigherLevelConstructs = 0b1000_0000;
     }
 }


source/compiler/qsc_data_structures/src/target.rs
minestarks, line 1: I completely forgot what "R" stands for in Adaptive_R... and I guess "L" is "Loops" which we call "BackwardsBranching" in our capabilities? It would be nice to have a comment that expands the acronym and describe which letter maps to which capability.

source/compiler/qsc/src/interpret.rs
```diff
@@ -1069,7 +1069,7 @@ impl Interpreter {
         entry_expr: Option<&str>,
         tracer_config: TracerConfig,
     ) -> std::result::Result<Circuit, Vec<Error>> {
-        if self.capabilities == TargetCapabilityFlags::all() {
+        if self.capabilities > Profile::AdaptiveRIF.into() {
```
minestarks, line 1099: What does this operator do?! Is this integer greater than, or sneakily overloaded to mean "superset" or something?

source/compiler/qsc_rir/src/rir.rs
```diff
@@ -347,6 +381,7 @@ pub enum Instruction {
     BitwiseXor(Operand, Operand, Variable),
     Phi(Vec<(Operand, BlockId)>, Variable),
     Convert(Operand, Variable),
+    Advanced(AdvancedInstr),
```
minestarks: Hear me out - I don't think "Advanced" is the right grouping here. The vague name is a bit of a code smell.

I think the grouping you want is `Memory`, and you want `Load`, `Alloca` and `Store` within that grouping.

Notice how in many spots in the code where you're rejecting `Advanced` (e.g. `Advanced(..) => unreachable!("advanced instructions not supported!"` ) you're also rejecting `Store` . So the point is not that "advanced" instructions are unsupported, it's that memory-related instructions are unsupported.

In other spots, `Store`/`Load`/`Alloca` are all related in some way so I think it makes sense that the code would be organized under the same match arm.

(And if you disagree with this particular grouping, then I don't think we should have a grouping at all - just let `Load` and `Store` be their own variants. The way `Advanced` is used now, I don't think it's really adding anything to the code).

What do you think?

source/compiler/qsc_codegen/src/qir/v1.rs
minestarks, line 1: Oof, that's a lot of duplication. Is v2 really so fundamentally different than v1 that we have to "fork" it?

source/compiler/qsc_data_structures/src/target.rs
swernli: We have user-facing explanatory comments on the Python enum (in source/pip/qsharp/_native.pyi) that match the Python interop enum in Rust (in source/pip/src/interpreter.rs), but somehow never put them in here. These are the classes most of the implementation actually uses, so definitely overdue for explanatory comments in here. I'll update these, as well as the closely related `TargetCapabilityFlags` to clarify the meanings and relationships.

source/compiler/qsc_rir/src/rir.rs
swernli: The intent behind introducing this category is that they are instructions that the older parts of the code can ignore. Yes, v1 codegen ignores `Store` but only because the RIR passes transform it away. Those same RIR passes should never see `Load` or `Alloca`, which themselves are only introduced by a newer pass. And while these are just memory instructions for now, the eventual arrays support will need to handle dynamic indexing (which becomes `getelementptr` instructions) and potentially more. So the grouping was not really about grouping instructions by theme but rather keeping all instructions the old passes can ignore together.

The alternative, which I started with, is just adding the new instruction types into the `Instruction` enum directly, but then every site that handles instructions needs to ignore the newer instructions individually, which becomes annoying. Still, if that's a preferable organization, I'd be willing to flatten the data structure rather than create this somewhat arbitrary nesting. A little short-term dev pain for long term struct gain ;)

source/compiler/qsc_codegen/src/qir/v1.rs
swernli: In my head they are distinct enough to warrant it, but in practice there might be more overlap we can take advantage of. I can revisit with some fresh eyes. It was very easy just to have a top-level split (ie: `v1::ToQir` vs `v2::ToQir`) but that doesn't mean a lot of the common helper functions couldn't be the same...

source/compiler/qsc/src/interpret.rs
swernli: No sneaky overloads, just sneaky meaning :). We talked a bit about this offline, but capturing here for posterity:
The `Profile` struct supports conversion into `TargetCapabilityFlags` which are implemented as bitflags. They also follow an ordering: higher value flags are high complexity, such that each `Profile` is greater than the ones that are of lesser complexity. Since bitflags are just ints and support comparison, this means exactly what it reads: if the current capabilities are greater than the Adaptive RIF profile, then do something. As you other feedback points out, adding more doc comments to both of these structures and these places where they are used this way will help clarify the use of this "well-ordered" propty of the flags and how it relates to profiles.

source/compiler/qsc_data_structures/src/target.rs
swernli: Added comments!

source/compiler/qsc_rir/src/rir.rs
minestarks: I made sure to scan your changes before I commented. You only have 3 spots where you match `Advanced(..)` . And those arms happen to do exactly the same thing as the `Store` arm (panic, mostly).  It's way more often that you match `Load` and `Alloca` separately. I'm almost certain that grouping it this way adds more code, than it eliminates. I think you end up with a cleaner layout if you just flatten the enum.

`Memory` was my suggestion if you felt like you _had_ to group them somehow. But if we can agree on flat, that's fine by me.

source/compiler/qsc_codegen/src/qir/v1.rs
minestarks: I mean I kind of think it depends on the intent.

Is the intention to isolate `v1`, never touch it again, and eventually retire it? Then I think forking it this way could make sense.

But if we want to keep maintaining both, then I think setting them up to share more code would be worth the effort.

I ended up diffing `v2.rs` against `v1.rs` to understand it. Currently - 

- `Result`, `Qubit` etc is now mapped to `ptr`  - this behavior is genuinely different
- template.ll - genuinely different
- Places where one version will say `unimplemented!` for something that only the other version implements (phi, load, etc) - these could be shared
- standalone helper functions - no need to duplicate these

I figure you're going to add more new behavior to `v2`, so you'd be the best judge of how the code will continue to diverge.

That being said, I don't have a good structure in my head that I can propose at the moment.... Other than the trivial step of "move all the standalone identical functions to a shared module", I guess.

swernli: I think the trivial step you mention is exactly what I intended to try out: move any function that is identical between the two into a helper. I just need to make sure the logic for recurring into places where they differ works as expected. I'll do that and see how it looks!

source/compiler/qsc_rir/src/rir.rs
swernli: I think the right thing for this is to flatten, so I'll work on making that change now.

source/compiler/qsc_codegen/src/qir/v1.rs
swernli: Revisiting and rething this one... I think what will really help is separating out all the common traversal logic into an RIR visitor (similar to what we have for other representations). Then it will be easier to encapsulate the common parts and separate the uncommon parts. But I think that makes more sense in a follow up PR that reduces the duplication rather than making that infrastructure part of this PR.

swernli: Added as #3051

source/compiler/qsc_data_structures/src/target.rs
```diff
@@ -4,13 +4,26 @@
...
         const IntegerComputations = 0b0000_0010;
+        /// Supports classical computations on floating point numbers (e.g. addition, multiplication).
         const FloatingPointComputations = 0b0000_0100;
+        /// Supports backward branching based on measurement results aka loops.
         const BackwardsBranching = 0b0000_1000;
-        const HigherLevelConstructs = 0b0001_0000;
+        /// Supports statically sized arrays (i.e. array literals and array indexing with non-constant indices).
+        const StaticSizedArrays = 0b0001_0000;
+        /// Catch-all for high level language constructs not covered by other flags. New flags should be added above this one,
+        /// such that this flag is reserved for the "all capabilities" targets that can run anything the langauge can express.
```
minestarks:
```suggestion
        /// such that this flag is reserved for the "all capabilities" targets that can run anything the language can express.
```

source/compiler/qsc_data_structures/src/target.rs
```diff
@@ -4,13 +4,26 @@
 use bitflags::bitflags;
 
 bitflags! {
+    /// These flags describe the capabilities of the target execution environment. They are used to determine which language features we can
+    /// emit into generated code for that target, and correlate to which design-time errors are surfaced by the capabilities check pass.
+    /// Note that the flags are in increasing order of capability and are expected to be largely additive. The code uses this "well-ordered"
+    /// property to perform some checks using comparison operators, inaddition to the user bitwise membership checks offered by bitflags.
```
minestarks:
```suggestion
    /// property to perform some checks using comparison operators, in addition to the bitwise membership checks offered by bitflags.
```

minestarks (approved):

PR #3038, opened 2026-03-20
Add loop emission to `Adaptive_RIFLA`

swernli:
This change adds logic to support emission of loops into QIR when use the Adaptive\_RIFLA profile. The strategy employed uses changes to both RCA and Partial Evaluation to accomplish this.

In RCA, we add a new check that identifies loops whose condition is marked as static and speculatively marks them as dynamic to compute what capabilities they would require at runtime. If the resulting capabilities are compatible with the current target, the loop remains dynamic, otherwise the RCA state is reset and the loop is marked as the originally analyzed static.

In Partial Evaluation, new support for dynamic loops is added, where a loop marked as dynamic is emitted as a sequence of blocks with appropriate backward branches to capture the while-loop structure.

This change includes new unit tests to verify analysis and emission of loop structures into RIR and new integration tests to verify the whole program structure and execution of QIR v2 with loops via qir-runner.

### DIFF
--- a/source/compiler/qsc_rca/src/lib.rs
+++ b/source/compiler/qsc_rca/src/lib.rs
@@ -756,7 +756,11 @@ impl RuntimeFeatureFlags {
             capabilities |= TargetCapabilityFlags::HigherLevelConstructs;
         }
         if self.contains(RuntimeFeatureFlags::UseOfDynamicExponent) {
-            capabilities |= TargetCapabilityFlags::BackwardsBranching;
+            // capabilities |= TargetCapabilityFlags::BackwardsBranching;
+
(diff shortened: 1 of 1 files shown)


source/pip/tests-integration/resources/adaptive_rifla/output/RUS.out
orpuente-MS, line 1: What is this format? I haven't seen it before.

swernli: This is the output format specified in the QIR spec (see https://github.com/qir-alliance/qir-spec/tree/main/specification/output_schemas), and emitted by executing the QIR via [qir-runner](https://github.com/qir-alliance/qir-runner).

PR #3039, opened 2026-03-22
Add support for Adaptive_RIFL QIR programs in GPU simulator

orpuente-MS:
This PR adds support for running Adaptive_RIFL QIR programs in the GPU simulator.

## The core idea

1. We compile the adaptive QIR to bytecode.
2. Then, we implement a bytecode interpreter as a GPU shader.
3. Then we upload the bytecode to the GPU and run thousands of adaptive QIR simulations in parallel.

## Current limitations

1. We only support 32-bit arithmetic at the moment
2. The register file size can hold up to 4096 registers, so, programs with more than that number of SSA will throw an error. This limit can be adjusted if we hit it in the future.

## Follow up work

1. Add array support to complete compliance with the Adaptive_RIFLA profile
2. Add Adaptive_RIFLA support to CPU and Clifford simulators

### DIFF
--- a/source/pip/qsharp/_adaptive_bytecode.py
+++ b/source/pip/qsharp/_adaptive_bytecode.py
@@ -5,7 +5,7 @@
 
 These constants define the bytecode encoding used by the Python AdaptiveProfilePass
 (emitter) and the Rust GPU receiver. Values must stay in sync with the Rust
-``adaptive_bytecode`` module and the WGSL interpreter.
+``bytecode.rs`` module and the WGSL interpreter.
 
 Opcode word layout::
 


source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
```diff
@@ -0,0 +1,2521 @@
+// Copyright (c) Microsoft Corporation.
```
billti, line 1: The prior `simulator.wgsl` still exists also, which shares a lot (most?) of the code in here.

Is the intent to keep to largely separate implementations, rather than just have one that also supports `adaptive` as a superset of the base?

source/simulators/src/gpu_full_state_simulator/gpu_resources.rs
```diff
@@ -128,6 +141,163 @@ impl Default for GpuDeviceResources {
...
+            kernels: None,
+            bind_group: None,
+            // --------------------------------------------------------------
+            // Bind group layout: 18 bindings in @group(0)
+            // --------------------------------------------------------------
+            // Bindings 0-8:  quantum ops (ops, state vector, results, noise, etc.)
+            // Bindings 9-17: Adaptive interpreter (bytecode, block/function/phi/switch
+            //                tables, interpreter state, register file, termination counter)
+            //
+            // This exceeds the WebGPU minimum of 8 storage buffers per shader stage
```
billti: I was surprised to see so many buffers added. Is this necessary? I had assumed things like the `InterpreterState` would just be part of the `ShotState`, that `ByteCode` would just be mixed in with the other `Ops` in the program, etc. Or are there reasons not to do that?

orpuente-MS: Merged all the interpreter state into `ShotData` and all the read-only data that is shared across all shots in a batch into a single `BatchData` structure. We're at 7 storage buffers now, even fewer buffers than before the PR.

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
orpuente-MS: Extracted the common utilities into `common.wgsl`, and now `simulator_base.wgsl` and `simulator_adaptive.wgsl` only have code specific to those QIR profiles

source/pip/qsharp/_adaptive_bytecode.py
```diff
@@ -0,0 +1,132 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Shared opcode constants for the Adaptive Profile QIR bytecode interpreter.
+
+These constants define the bytecode encoding used by the Python AdaptiveProfilePass
+(emitter) and the Rust GPU receiver. Values must stay in sync with the Rust
+``adaptive_bytecode`` module and the WGSL interpreter.
```
minestarks: Where is the Rust module this is referring to? I couldn't find it in the source.

minestarks: Oh- I think it's `bytecode.rs`. Saw it

source/pip/qsharp/_adaptive_pass.py
```diff
@@ -0,0 +1,986 @@
...
+        )
+        (dst, src0, src1, aux0, aux1, aux2, aux3) = unwrap_operands(
+            dst, src0, src1, aux0, aux1, aux2, aux3
+        )
+        ins = Instruction(opcode | imm_flags, dst, src0, src1, aux0, aux1, aux2, aux3)
+        self.instructions.append(ins)
+
+    # TODO: Could this be encoded as an Instruction?
+    #       The angle could be cast to bytes and stored as an int,
+    #       then cast back float at runtime.
```
minestarks: Is this still an open issue?

orpuente-MS: Not anymore, removed the comment. I decided along the way that it would be best to have these as QuantumOps so that I could share some common infrastructure/utilities with the already existing GPU Base Profile pipeline

source/pip/qsharp/_simulation.py
```diff
@@ -516,28 +529,56 @@ def run_qir_cpu(
...
     noise: Optional[NoiseConfig] = None,
     seed: Optional[int] = None,
-) -> List[str]:
+) -> List:
     (mod, shots, noise, seed) = preprocess_simulation_input(input, shots, noise, seed)
     # Ccx is not support in the GPU simulator, decompose it
     DecomposeCcxPass().run(mod)
-    if noise is None:
-        (gates, num_qubits, num_results) = AggregateGatesPass().run(mod)
+    if is_adaptive(mod):
```
minestarks, line 553: Do we need to check for any specific extensions (e.g. to reject the *A* in RIFLA)? Or are we expected to simply raise an error when we encounter an unsupported instruction?

source/pip/qsharp/_adaptive_bytecode.py
orpuente-MS: Good catch, it was named `adaptive_bytecode.rs` before. I need to update that comment.

source/pip/qsharp/_simulation.py
orpuente-MS: Yes, we should reject arrays with a nice error, added as the first check to the Adaptive pyqir pass.

PR #3047, opened 2026-03-23
Submit Cirq to Local Neutral Atom Simulator

ScottCarda-MS:
Adds a Cirq interop layer to the `qsharp` package and surfaces that layer via the the `qdk.cirq` namespace in the `qdk` package. The interop layer exposes a `NeutralAtomSampler`, which is a `Cirq.Sampler`, which allows Cirq circuits to be run on the Neutral Atom Simulator by . It also exposes a `NeutralAtomCirqResult`, which is a `Cirq.ResultDict`, that is returned from the sampler. The special result type looks like a regular result type to downstream Cirq tooling, but also contains raw data results for shots that had qubit loss, which are filtered out of the regular Cirq result fields.

Here's an example using the Neutral Atom Sampler:
```python
import cirq
from qdk.cirq import NeutralAtomSampler
from qdk.simulation import NoiseConfig

# Build a 3-qubit GHZ circuit in Cirq
qubits = cirq.LineQubit.range(3)
circuit = cirq.Circuit([
    cirq.H(qubits[0]),
    cirq.CNOT(qubits[0], qubits[1]),
    cirq.CNOT(qubits[1], qubits[2]),
    cirq.measure(*qubits, key="m"),
])
print(circuit)

# Configure loss noise — high probability so we reliably see loss in results
noise = NoiseConfig()
noise.rz.loss = 0.08  # 8% chance of qubit loss on each Rz gate

# Run with noise
result = NeutralAtomSampler(noise=noise, seed=42).run(circuit, repetitions=1000)

# Cirq-compatible view: loss shots filtered out
print("Accepted counts (no loss):", dict(result.histogram(key="m")))
print("Accepted sample:", result.measurements["m"][:5].tolist(), "...")

# raw_shots: unprocessed simulator output, one entry per shot.
# For NeutralAtomDevice these are Result enum tuples, e.g. (Result.Zero, Result.Loss, Result.One).
# Use this when you need the exact simulator output with no post-processing.
print("\nraw_shots (simulator output):", result.raw_shots[:5], "...")

# raw_measurements(): same shots converted to the Cirq {key: shots×qubits array} structure,
# with loss represented as "-" in each qubit cell. This mirrors result.measurements in shape
# and key layout, making it easy to compare accepted vs raw data per register.
raw = result.raw_measurements()
print("raw_measurements['m'] sample:", raw["m"][:5].tolist(), "...")

print(f"\nAccepted shots: {len(result.measurements['m'])} / {len(result.raw_shots)} total")

```

### DIFF
--- a/source/pip/qsharp/interop/cirq/__init__.py
+++ b/source/pip/qsharp/interop/cirq/__init__.py
@@ -24,11 +24,8 @@ Usage::
     print(result.histogram(key="m"))
 """
 
-try:
-    from ._neutral_atom import NeutralAtomSampler
-    from ._result import NeutralAtomCirqResult
-except ImportError:
-    pass
+from ._neutral_atom import NeutralAtomSampler
+from ._result import NeutralAtomCirqResult
 
 __all__ = [
     "NeutralAtomSampler",


source/pip/qsharp/interop/cirq/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+"""Cirq interoperability for the Q# ecosystem.
+
+This module provides a :class:`NeutralAtomSampler` — a standard
+``cirq.Sampler`` that runs Cirq circuits on the local NeutralAtomDevice
+simulator.
+
+Usage::
+
+    import cirq
+    from qsharp.interop.cirq import NeutralAtomSampler
```
minestarks, line 13: Is this idiomatic? Do we want users to import from `qsharp.interop` or `qdk.interop` ?

ScottCarda-MS: Users will ideally import from `qdk.cirq` but it is useful the `qsharp` package to be the location for the content to fit the patterns we use for the package structure. This helps make sure there aren't circular dependencies for packages under `qdk`.

minestarks (approved):

source/pip/pyproject.toml
```diff
@@ -22,6 +22,7 @@ classifiers = [
 jupyterlab = ["qsharp-jupyterlab"]
 widgets = ["qsharp-widgets"]
 qiskit = ["qiskit>=1.2.2,<3.0.0"]
+cirq = ["cirq-core>=1.6.1,<1.7"]
```
billti, line 25: Did they ship a 1.7? Minor bumps are supposed to be backwards compatible, so should this be `<2.0` (similar to Qiskit) rather than `<1.7`?

ScottCarda-MS: Cirq's minor version updates haven't been very backwards compatible, which is why we were stuck on 1.4.1 for a while, which is why I think it is safer to gate on 1.7 rather than 2.0

source/pip/qsharp/interop/cirq/__init__.py
billti: But this is a doc comment. Isn't this the 'help' the user will see explaining how to use this module?

source/qdk_package/src/qdk/cirq.py
```diff
@@ -0,0 +1,32 @@
...
+"""Cirq interoperability for the Q# ecosystem.
+
+This module provides a :class:`NeutralAtomSampler` — a standard
+``cirq.Sampler`` that runs Cirq circuits on the local NeutralAtomDevice
+simulator.
+
+Usage::
+
+    import cirq
+    from qsharp.interop.cirq import NeutralAtomSampler
```
billti: And this is in the `qdk` package, so definitely shouldn't give help telling users to import from `qsharp.interop`

source/qdk_package/src/qdk/cirq.py
```diff
@@ -0,0 +1,32 @@
...
+    ])
+
+    sampler = NeutralAtomSampler(seed=42)
+    result = sampler.run(circuit, repetitions=1000)
+    print(result.histogram(key="m"))
+"""
+
+try:
+    from qsharp.interop.cirq import *  # pyright: ignore[reportWildcardImportFromLibrary]
+except Exception as ex:
```
billti, line 29: How is this ever going to fail or raise an exception? All the imports in `qsharp.interop.cirq` are wrapped a in `try/pass`, so will never raise an exception.

source/pip/tests-integration/interop_cirq/test_neutral_atom.py
```diff
@@ -0,0 +1,354 @@
...
+        ]
+    )
+
+
+# ---------------------------------------------------------------------------
+# Smoke tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.skipif(not CIRQ_AVAILABLE, reason=SKIP_REASON)
```
billti: If the `test_requirements.txt` lists cirq as a dependency, why would we skip all tests if cirq isn't available? Shouldn't these always run (and pass)?

source/qdk_package/src/qdk/cirq.py
ScottCarda-MS: 🤔I'm not sure what you mean here. When I uninstall cirq from my environment and try to run python that accesses `qdk.cirq` I get the following error:
```

ImportError: qdk.cirq requires the cirq extra. Install with 'pip install qdk[cirq]'.
```
which is the error triggered by this code.

source/qdk_package/src/qdk/cirq.py
ScottCarda-MS: Ah that is copy-pasta from the `qsharp` version. Fixed!

source/pip/tests-integration/interop_cirq/test_neutral_atom.py
ScottCarda-MS: Ah good point! I set this up to match the way Qiskit handles its tests, but I can see now that there was a special reason why Qiskit was set up that way (the simultaneous support for otherwise incompatible Qiskit versions). Since Cirq doesn't need that, it makes sense for us to not have to copy that complexity for our Cirq tests. Removed!

source/qdk_package/src/qdk/cirq.py
ScottCarda-MS: I removed the try/except in `qsharp.interop.cirq` to make the error a little clearer and de-clutter that \_\_init\_\_.py a bit. It now matches what the qiskit interop layer does, buy just telling the user that they don't have a module named `cirq`. When importing from `qdk.cirq` they will still get the `pip install qdk[cirq]` message.

PR #3048, opened 2026-03-23
Optimize `PreparePureStateD`

swernli:
This change introduces different structure for `Std.StatePreparation.PreparePureStateD` and underlying helper functions, which reduces the compile time by offloading more of the classical compute into static functions run on the full evaluator rather than handled inside the partial-evaluation layer.

In local testing, this improved QIR compilation of a large state preparation on 16 qubits from 14 seconds to 4 seconds.

### DIFF
--- a/library/std/src/Std/StatePreparation.qs
+++ b/library/std/src/Std/StatePreparation.qs
@@ -64,7 +64,7 @@ import
 /// - Std.StatePreparation.ApproximatelyPreparePureStateCP
 operation PreparePureStateD(coefficients : Double[], qubits : Qubit[]) : Unit is Adj + Ctl {
     let nQubits = Length(qubits);
-    // pad coefficients at tail length to a power of 2.
+    // Pad coefficients at tail length to a power of 2.
     let coefficientsPadded = Padded(-2^nQubits, 0.0, coefficients);
     let idxTarget = 0;
 
@@ -148,7 +148,7 @@ operation ApproximatelyPreparePureStateCP(
 ) : Unit is Adj + Ctl {
 
     let nQubits = Length(qubits);
-    // pad coefficients at tail length to a power of 2.
+    // Pad coefficients at tail length to a power of 2.
     let coefficientsPadded = Padded(-2^nQubits, ComplexPolar(0.0, 0.0), coefficients);
     let idxTarget = 0;
     // Determine what controls to apply
@@ -261,17 +261,7 @@ function StatePreparationSBMComputeCoefficientsD(
     mutable newCoefficients = [];
 
     for idxCoeff in 0..2..Length(coefficients) - 1 {
-        let (rt, phi, theta) = {
-            let abs0 = AbsD(coefficients[idxCoeff]);
-            let abs1 = AbsD(coefficients[idxCoeff + 1]);
-            let arg0 = coefficients[idxCoeff] < 0.0 ? PI() | 0.0;
-            let arg1 = coefficients[idxCoeff + 1] < 0.0 ? PI() | 0.0;
-            let r = Sqrt(abs0 * abs0 + abs1 * abs1);
-            let t = 0.5 * (arg0 + arg1);
-            let phi = arg1 - arg0;
-            let theta = 2.0 * ArcTan2(abs1, abs0);
-            (ComplexPolar(r, t), phi, theta)
-        };
+        let (rt, phi, theta) = BlockSphereCoordinatesD(coefficients[idxCoeff], coefficients[idxCoeff + 1]);
         set disentanglingZ += [0.5 * phi];
         set disentanglingY += [0.5 * theta];
         set newCoefficients += [rt];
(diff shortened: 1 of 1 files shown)


library/std/src/Std/StatePreparation.qs
```diff
@@ -226,7 +252,37 @@ operation ApproximatelyMultiplexPauli(
...
+        let (rt, phi, theta) = {
+            let abs0 = AbsD(coefficients[idxCoeff]);
+            let abs1 = AbsD(coefficients[idxCoeff + 1]);
+            let arg0 = coefficients[idxCoeff] < 0.0 ? PI() | 0.0;
+            let arg1 = coefficients[idxCoeff + 1] < 0.0 ? PI() | 0.0;
+            let r = Sqrt(abs0 * abs0 + abs1 * abs1);
+            let t = 0.5 * (arg0 + arg1);
+            let phi = arg1 - arg0;
+            let theta = 2.0 * ArcTan2(abs1, abs0);
+            (ComplexPolar(r, t), phi, theta)
```
fedimser: I think this can be factored into a function BlochSphereCoordinatesD taking 2 Doubles.

library/std/src/Std/StatePreparation.qs
```diff
@@ -63,8 +63,35 @@ import
 /// # See Also
 /// - Std.StatePreparation.ApproximatelyPreparePureStateCP
 operation PreparePureStateD(coefficients : Double[], qubits : Qubit[]) : Unit is Adj + Ctl {
-    let coefficientsAsComplexPolar = Mapped(a -> ComplexAsComplexPolar(Complex(a, 0.0)), coefficients);
-    ApproximatelyPreparePureStateCP(0.0, coefficientsAsComplexPolar, qubits);
+    let nQubits = Length(qubits);
+    // pad coefficients at tail length to a power of 2.
```
fedimser: Do we have a Q# styleguide and what does it say about comments?

Generally, when I write comments, I write them as sentences, starting with capital letter and ending with a period. I'd suggest following this pattern.

library/std/src/Std/StatePreparation.qs
```diff
@@ -63,8 +63,35 @@ import
...
+    let idxTarget = 0;
+
+    // Note we use the reversed qubits array to get the endianness ordering that we expect
+    // when corresponding qubit state to state vector index.
+    let qubits = Reversed(qubits);
+
+    // Since we know the coefficients are real, we can optimize the first round of adjoint approximate unpreparation by directly
+    // computing the disentangling angles and the new coefficients on those doubles without producing intermediate complex numbers.
+
+    // For each 2D block, compute disentangling single-qubit rotation parameters
```
fedimser, line 78: What does "2D" stand for? My first thought without context is "2-dimensional", but it's probably something else.

library/std/src/Std/StatePreparation.qs
```diff
@@ -357,8 +416,70 @@ operation ApproximatelyMultiplexZ(
...
+                Controlled X(controlRegister, target);
+            } apply {
+                Adjoint ApproximatelyMultiplexZ(tolerance, coefficients1, control, target);
+            }
+        }
+        Adjoint ApproximatelyMultiplexZ(tolerance, coefficients0, control, target);
+    }
 }
 
+// Provides the sequence of angles or entangling CNOTs to apply for the multiplex Z step of the state preparation procedure, given a set of coefficients and control and target qubits.
```
fedimser: I think it's worth to call out that there is convention on how to interpret output:
* Every element has 1 or 2 qubits.
* If it's one qubit, it's a rotation.
* If it's 2 qubits, it's CNOT, and angle is ignored.

library/std/src/Std/StatePreparation.qs
```diff
@@ -357,8 +416,70 @@ operation ApproximatelyMultiplexZ(
...
+function GenerateMultiplexZParams(
+    tolerance : Double,
```
fedimser: Tolerance is not really used.
I'd suggest either not passing it or passing it and use in "termination case" to not emit the rotation gate.

library/std/src/Std/StatePreparation.qs
```diff
@@ -357,8 +416,70 @@ operation ApproximatelyMultiplexZ(
...
+        let (coefficients0, coefficients1) = MultiplexZCoefficients(coefficientsPadded);
+        if AnyOutsideToleranceD(tolerance, coefficients1) {
+            within {
+                Controlled X(controlRegister, target);
+            } apply {
+                Adjoint ApproximatelyMultiplexZ(tolerance, coefficients1, control, target);
+            }
+        }
+        Adjoint ApproximatelyMultiplexZ(tolerance, coefficients0, control, target);
+    }
```
fedimser: I am not sure why we need explicit adjoin control and it can't be auto-generated.

library/std/src/Std/StatePreparation.qs
```diff
@@ -321,25 +377,28 @@ operation ApproximatelyMultiplexZ(
...
+        // provides a performance improvement during partial-evaluation for code generation.
+        let adjMultiplexZParams = GenerateAdjMultiplexZParams(tolerance, coefficients, control, target);
+        for (angle, qs) in adjMultiplexZParams {
+            if Length(qs) == 2 {
+                CNOT(qs[0], qs[1]);
+            } elif AbsD(angle) > tolerance {
+                Exp([PauliZ], -angle, qs);
             }
         }
     }
```
fedimser, line 408: I think we can move checking tolerance and negating angle for adjoint case into classical functions.

Then you can have exactly the same quantum part that can be factored into separate operation to reduce code duplication.

```
operation ApproximatelyMultiplexZ(tolerance, coefficients, control, target) {
   body ... {
      ApplyRzCnots(GenerateMultiplexZParams(tolerance, coefficients, control, target));
   }
   adjoint ... {
      ApplyRzCnots(GenerateAdjMultiplexZParams(tolerance, coefficients, control, target));
   }
}
```

and then:

```
operation ApplyRzCnots(params: (Double, Qubit[])[]) {
   for (angle, qs) in adjMultiplexZParams {
        if Length(qs) == 2 {
            CNOT(qs[0], qs[1]);
        } else {
            Exp([PauliZ], angle, qs);
        }
    }
} 
```

library/std/src/Std/StatePreparation.qs
```diff
@@ -321,25 +377,28 @@ operation ApproximatelyMultiplexZ(
...
-        // pad coefficients length at tail to a power of 2.
-        let coefficientsPadded = Padded(-2^Length(control), 0.0, coefficients);
-
-        if Length(coefficientsPadded) == 1 {
-            // Termination case
-            if AbsD(coefficientsPadded[0]) > tolerance {
-                Exp([PauliZ], coefficientsPadded[0], [target]);
+    body ... {
+        // We separately compute the operation sequence for the multiplex Z steps in a function, which
+        // provides a performance improvement during partial-evaluation for code generation.
```
fedimser: I think it's too much information for the reader especially if they don't know what partial eval is which is implementation detail of compiler.

I'd just call out that recursion in quantum operation is more expensive that in classical.

Or just don't have any comment here, it's obvious that we separate quantum and classical computation,

library/std/src/Std/StatePreparation.qs
swernli: We do have some guidance when it comes to doc comments (`///`) that precede an item declaration, since those end up in our generated docs, but for simple comments (`//`) inside a function there isn't really one. That said, simple fix here :).

library/std/src/Std/StatePreparation.qs
swernli: I actually don't know here. Much of this code was copied from the old Q# library and comments were preserved. I wouldn't be surprised if the notion of "blocks" with different dimensionality came from the paper linked above.

library/std/src/Std/StatePreparation.qs
swernli: It can, and I originally had it that way, but it's called in only this one spot. Still, the existing one is also called in one place, so I might as well factor it back out.

library/std/src/Std/StatePreparation.qs
swernli: I'll simplify the comment!

library/std/src/Std/StatePreparation.qs
swernli: It has to do with how the automatic generation functions. Theoretically, it can either take the explicit controlled specialization and invert it OR it can take the explicit adjoint specialization and distribute controls to it. In this case, the former is correct but the latter is not. That said, I totally forgot we have a way to specify which to choose (`controlled adjoint invert` and `controlled adjoint distribute` respectively). That definitely makes the code simpler!

library/std/src/Std/StatePreparation.qs
swernli: Huh, you're right. I can just remove that now.

library/std/src/Std/StatePreparation.qs
swernli: That's a good suggestion. I was able to do something similar to this, but ended up just defining the appliation operation as adjointable and then using the fact that `Adjoint Exp(... angle ...)` is equivalent to `Exp(... -angle ...)`.

library/std/src/Std/StatePreparation.qs
swernli: Added more explanation of the structure to the generation functions.

PR #3058, opened 2026-03-24
Add `qdk-programming` Copilot skill for Q#, OpenQASM and Python

minestarks:
Replaces the current `chatInstructions` with a `SKILL.md` that routes the agent to focused reference files (`qsharp.md`, `openqasm.md`, `python.md`) based on what the user is doing. The existing instruction files are kept with their `applyTo` globs but now just redirect to the skill. Descriptions on both the skill and instruction files are updated to improve matching in relevant scenarios. The contents are also updated to match current features.

Note: I agonized over whether to put all the information in a `SKILL.md` or an `*.instructions.md` . In the end, it really doesn't matter _that_ much. Skills "won" because they're:

- an open standard instead of a VS Code specific feature
- meant to match based on the description of the task, rather than an `applyTo` glob which isn't as expressive
- a folder with granular, individual files that can be read on demand, rather than a single .md file trying to contain all possibly relevant information

See VS Code reference for skill vs. instructions
https://code.visualstudio.com/docs/copilot/customization/agent-skills#_agent-skills-vs-custom-instructions

### DIFF
--- a/source/vscode/skills/qdk-programming/qsharp.md
+++ b/source/vscode/skills/qdk-programming/qsharp.md
@@ -93,7 +93,7 @@ let re = c.Re;                     // Field access
 ```qsharp
 // Arithmetic: + - * / % ^       Comparison: == != < <= > >=
 // Logical: and or not           Bitwise: ~~~ &&& ||| ^^^ <<< >>>
-// String interpolation: $"x={x}"  Array update: arr w/ idx <- val
+// String interpolation: $"x={x}"  Array update: arr[i] = val
 // Ternary: cond ? trueVal | falseVal
 ```
 
@@ -112,6 +112,11 @@ for elem in array { }             // Array iteration
 while condition { }
 
 // Repeat-until (quantum retry pattern)
+repeat {
+    H(q);
+} until MResetZ(q) == Zero;
+
+// Repeat-until with fixup block
 repeat {
     H(q);
 } until M(q) == Zero
@@ -149,7 +154,7 @@ let cnot0 = CNOT(control, _);      // _ is placeholder
 ### Functor Support
 
 ```qsharp
-// Declare with `is` keyword
+// Declare with `is` keyword — compiler auto-generates Adjoint/Controlled
 operation MyGate(q : Qubit) : Unit is Adj + Ctl { H(q); }
 
 // Use functors
@@ -157,8 +162,7 @@ Adjoint MyGate(q);                  // Inverse
 Controlled MyGate([ctrl], q);       // Controlled version
 Controlled Adjoint MyGate([ctrl], q);
 
-// Auto-generated: compiler derives Adjoint/Controlled
-// Manual specializations:
+// Manual specializations (when auto-generation isn't sufficient):
 operation Custom(q : Qubit) : Unit is Adj {
     body (...) { H(q); }
     adjoint (...) { H(q); }        // H is self-adjoint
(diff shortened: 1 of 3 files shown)


source/vscode/package.json
```diff
@@ -379,6 +379,11 @@
         "path": "./resources/chat-instructions/openqasm.instructions.md"
       }
     ],
+    "chatSkills": [
+      {
+        "path": "./skills/qdk-programming/SKILL.md"
```
swernli, line 384: Do the other markdown files in the qdk-programming folder need to be added a similar way?

source/vscode/skills/qdk-programming/qsharp.md
```diff
@@ -0,0 +1,316 @@
...
+let c2 = new Complex2D { ...c, Im = 2.0 };  // Copy-update
+let re = c.Re;                     // Field access
+```
+
+### Operators
+
+```qsharp
+// Arithmetic: + - * / % ^       Comparison: == != < <= > >=
+// Logical: and or not           Bitwise: ~~~ &&& ||| ^^^ <<< >>>
+// String interpolation: $"x={x}"  Array update: arr w/ idx <- val
```
swernli: for array update, should we lean toward the newer syntactic sugar with `arr[i] = val` instead?

source/vscode/skills/qdk-programming/qsharp.md
```diff
@@ -0,0 +1,316 @@
...
+for i in 0..2..10 { }             // Step by 2
+for elem in array { }             // Array iteration
+
+// While loops
+while condition { }
+
+// Repeat-until (quantum retry pattern)
+repeat {
+    H(q);
+} until M(q) == Zero
```
swernli, line 122: fixup blocks are optional, so probably worth an example without it in the form:
```qsharp
repeat {
    H(q);
} until MResetZ(q) == Zero;
```

source/vscode/skills/qdk-programming/qsharp.md
```diff
@@ -0,0 +1,316 @@
...
+```qsharp
+// Declare with `is` keyword
+operation MyGate(q : Qubit) : Unit is Adj + Ctl { H(q); }
+
+// Use functors
+Adjoint MyGate(q);                  // Inverse
+Controlled MyGate([ctrl], q);       // Controlled version
+Controlled Adjoint MyGate([ctrl], q);
+
+// Auto-generated: compiler derives Adjoint/Controlled
```
swernli: to avoid confusion, it might be worth moving the mention of auto-generated specializations up to the first example that has auto-generation. Then it's clearer that the manual comment is separate and applies to the code below.

source/vscode/skills/qdk-programming/qsharp.md
```diff
@@ -0,0 +1,316 @@
...
+
+```qsharp
+@EntryPoint(Adaptive_RI)
+operation Main() : Result { ... }
+```
+
+| Profile / Attribute             | Description                                                                    |
+| ------------------------------- | ------------------------------------------------------------------------------ |
+| `unrestricted`                  | Full simulation (default)                                                      |
+| `adaptive_ri` / `Adaptive_RI`   | Mid-circuit measurement + classical feed-forward                               |
```
swernli: and integer computations!

source/vscode/skills/qdk-programming/qsharp.md
```diff
@@ -0,0 +1,316 @@
...
+| `adaptive_rif` / `Adaptive_RIF` | Adds floating-point computation; required for `CircuitGenerationMethod.Static` |
```
swernli: nitpick: I'd swap the ordering of Adaptive RI and Adaptive RIF so the table is in decreasing order of capability.

source/vscode/package.json
joao-boechat: I think SKILL.md is an entry point where all of the skills are defined for the agent. In it the agent can understand what files to read and when. That's my understanding though, let's wait for @minestarks

minestarks: Yes, this is one where we just need to declare the entry point. Only SKILL.md has a special format, the other files are basically just arbitrary resources in the folder Copilot can read (if it wants to).

swernli: makes sense, thanks!

PR #3060, opened 2026-03-25
Create QRE traces from cirq circuits

msoeken:
This adds support to generate traces from cirq circuits. These traces represent the original circuit and do not simply represent the gate counts. They monkey patch a `_to_trace` method to cirq gates and users of QRE can add a `_to_trace` method to their custom operations and gates. If no such method is found, the operation's or gate's decomposition method will be called to walk the circuit tree.

The PR has some other small fixes:
- extending the gate-based architecture gate set
- minor fixes to `Trace` Python API
- minor improvements on how to create _estimation table results_ (with extended info such as source ISA tree) from unannotated _estimation results_

### DIFF
--- a/source/pip/qsharp/qre/interop/_cirq.py
+++ b/source/pip/qsharp/qre/interop/_cirq.py
@@ -145,20 +145,13 @@ class _Context:
 
     def handle_op(
         self,
-        op: (
-            cirq.Operation
-            | tuple[int, list[cirq.Qid] | cirq.Qid]
-            | tuple[int, list[cirq.Qid] | cirq.Qid, list[float] | float]
-            | PushBlock
-            | PopBlock
-        ),
+        op: cirq.OP_TREE | TraceGate | PushBlock | PopBlock,
     ) -> None:
         """Recursively convert a single operation into trace instructions.
 
         Supported operation forms:
 
-        - ``tuple``: A raw trace instruction as ``(id, qubits)`` or
-        ``(id, qubits, params)``, added directly to the current block.
+        - ``TraceGate``: A raw trace instruction, added directly to the current block.
         - ``PushBlock`` / ``PopBlock``: Control block nesting with repetitions.
         - ``GateOperation``: Dispatched via ``_to_trace`` if available on the
         gate, otherwise decomposed via ``_decompose_with_context_`` or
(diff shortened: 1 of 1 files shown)


source/pip/qsharp/qre/interop/_cirq.py
```diff
@@ -0,0 +1,294 @@
...
+)
+
+
+def trace_from_cirq(circuit: cirq.Circuit) -> Trace:
+    q_to_id = QidToTraceId(circuit.all_qubits())
+
+    trace = Trace(len(circuit.all_qubits()))
+
+    context = cirq.DecompositionContext(
+        qubit_manager=cirq.GreedyQubitManager("trace_from_cirq")
```
YingrongChen: Thought: We might need to use a custom qubit manager for the memory-compute layout in the future.

source/pip/qsharp/qre/interop/_cirq.py
```diff
@@ -0,0 +1,413 @@
...
+_TOLERANCE = 1e-8
+
+
+def _approx_eq(a: float, b: float) -> bool:
+    """Check whether two floats are approximately equal."""
+    return abs(a - b) <= _TOLERANCE
+
+
+def trace_from_cirq(
+    circuit: cirq.Circuit, *, classical_control_probability: float = 0.5
```
brad-lackey: Would it be better to operate on cirq.CIRCUIT_LIKE, so that if one passes a single gate, or a CircuitOperation, the parser could still work?

source/pip/qsharp/qre/interop/_cirq.py
```diff
@@ -0,0 +1,413 @@
...
+                self.block.add_operation(id, qs)
+            else:
+                self.block.add_operation(
+                    id, qs, params if isinstance(params, list) else [params]
+                )
+        elif isinstance(op, PushBlock):
+            self.push_block(op.repetitions)
+        elif isinstance(op, PopBlock):
+            self.pop_block()
+        elif isinstance(op, GateOperation):
```
brad-lackey: Similarly, would these be better as a cirq.OP_TREE, so that we would do a single operation, or unroll a list of operations? Note that this is actually based on superclass Operation.

source/pip/qsharp/qre/interop/_cirq.py
msoeken: Good point. I changed that. In fact, I already had some kind of similar logic for this in place to simplify the tests.

source/pip/qsharp/qre/interop/_cirq.py
msoeken: This lead to problems first because I also encode raw trace operations as tuples. But now I made a dedicated type for them, which is also more robust. Changed it to cirq.OP_TREE now.

PR #3062, opened 2026-03-25
Support `NoiseConfig` for Q# and OpenQASM on sparse simulation

swernli:
This change supports granular noise configuration for the mechanisms that use sparse state simulator, namely `qsharp.run` and `openqasm.run`. This brings it up to parity with our QIR-based simulation in the `NeutralAtomDevice` on cpu and gpu (and clifford).

### DIFF
--- a/source/pip/tests/test_sparse_simulator.py
+++ b/source/pip/tests/test_sparse_simulator.py
@@ -64,12 +64,11 @@ def test_sparse_bitflip_noise():
     noise.rzz.set_pauli_noise("XX", p_noise)
     noise.mresetz.set_bitflip(p_noise)
 
-    qsharp.set_quantum_seed(17)
-
     output = run(
         "IsingModel2DEvolution(4, 4, PI() / 2.0, PI() / 2.0, 10.0, 10)",
         shots=1,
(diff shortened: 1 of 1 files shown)


source/pip/tests/test_sparse_simulator.py
```diff
@@ -0,0 +1,440 @@
...
+    n_qubits = 11
+    seed = 20
+    p_loss = 0.1
+    tolerance_percent = 5.0
+    assert n_qubits >= 2, "Need at least two qubits"
+
+    random.seed(seed)
+    i1, i2 = random.sample(range(n_qubits), 2)
+    prefix = f"""
+operation tiny_coeffs(shots : Int) : Result[][] {{
```
orpuente-MS: Once that #3065 merges you could revert this to `operation tiny_coeffs() : Result[]`

swernli: Good call! I'll rebase on that and simplify the pattern.

swernli: Merged main, resolved the conflicts, and now the test works more like the cpu test it's copied from.

PR #3064, opened 2026-03-26
Fix quantum seed not being applied in invoke execution path

wavefunction91:
The `invoke_with_tracing_backend` method was missing the seed propagation that `run_with_sim` and `run_with_tracing_backend` already had. When `set_quantum_seed` was called followed by `invoke_with_noise`, `invoke_with_sim`, or circuit generation in simulate mode, the quantum seed was silently ignored. This produced non-deterministic measurement results even with a fixed seed.

Changes:

- Add `if self.quantum_seed.is_some() { tracing_backend.set_seed(self.quantum_seed); }` to `invoke_with_tracing_backend`, matching the existing pattern in the run path
- Add `SparseSim::new_with_seed(seed)` constructor that passes a seeded RNG directly to `QuantumSim::new()`, aligning with the constructor patterns used by all other simulators in the codebase (`StateVectorSimulator::new_with_seed`, `StabilizerSimulator::new`, etc.)
- Add regression tests for both the invoke and run paths, and unit tests for the new constructor

### DIFF
--- a/source/compiler/qsc/src/interpret.rs
+++ b/source/compiler/qsc/src/interpret.rs
@@ -1303,6 +1303,9 @@ impl Interpreter {
         args: Value,
         config: ExecGraphConfig,
     ) -> InterpretResult {
+        if self.quantum_seed.is_some() {
+            tracing_backend.set_seed(self.quantum_seed);
+        }
         qsc_eval::invoke(
             self.package,
             self.classical_seed,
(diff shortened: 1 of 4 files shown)


source/compiler/qsc/src/interpret.rs
```diff
@@ -1303,6 +1303,9 @@ impl Interpreter {
         args: Value,
         config: ExecGraphConfig,
     ) -> InterpretResult {
+        if self.quantum_seed.is_some() {
+            tracing_backend.set_seed(self.quantum_seed);
+        }
```
swernli, line 1308: This appears to be the only change actually needed. The new constructor in backend.rs and associated tests in noise_tests.rs are unnecessary and can be removed.

PR #3065, opened 2026-03-26
Support explicit seed in `qsharp.run`

swernli:
Previously `qsharp.run` just tried to use the globally configured random seed, but this resulted in every shot returning the same result. To allow a more natural usage that doesn't depend on global interpreter state, this updates `qsharp.run` to take a seed paramter and use that in a manner similar to `openqasm.run` to ensure distinct but deterministic shot results.

### DIFF
--- a/source/pip/qsharp/_qsharp.py
+++ b/source/pip/qsharp/_qsharp.py
@@ -775,7 +775,14 @@ def run(
         assert isinstance(entry_expr, str)
         run_entry_expr = entry_expr
 
+    shot_seed = seed
     for shot in range(shots):
+        # We also don't want every shot to return the same results, so we update the seed for
+        # the next shot with the shot number. This keeps the behavior deterministic if a seed
+        # was provided.
+        if seed is not None:
+            shot_seed = shot + seed
+
         results.append(
             {"result": None, "events": [], "messages": [], "matrices": [], "dumps": []}
         )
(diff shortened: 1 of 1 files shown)


source/pip/qsharp/_qsharp.py
```diff
@@ -793,6 +796,11 @@ def on_save_events(output: Output) -> None:
         # a rerun of the last executed expression without paying the cost for any additional
         # compilation.
         run_entry_expr = None
+        # We also don't want every shot to return the same results, so we update the seed for
+        # the next shot with the shot number. This keeps the behavior deterministic if a seed
+        # was provided.
+        if seed is not None:
+            seed += shot
```
orpuente-MS: Would it be better to do `seed += 1`  or pass `seed + shot` directly to `get_interpreter().run()`?

Doing `seed += shot` increments seed in a O(n^2) manner, so it will be easier to get an overflow on large number of shots, especially if the initial seed is already a large number.

source/pip/qsharp/_qsharp.py
```diff
@@ -706,6 +706,7 @@ def run(
         ]
     ] = None,
     qubit_loss: Optional[float] = None,
+    seed: Optional[int] = None,
```
minestarks, line 709: I think we'll want to connect this to classical seed as well, to get varied results across shots with classical random functions as well

swernli: good catch! Just added.

source/pip/qsharp/_qsharp.py
swernli: that's a good idea, I'll adjust it that way.

PR #3086, opened 2026-04-01
Adaptive Profile support for CPU-full-state and Clifford simulators

orpuente-MS:
This PR extends the work in #3039 by adding Adaptive Profile support for the CPU-full-state and Clifford simulators.

It also fixes a bug in which the GPU simulator was not using dynamic angles for rotation gates.


source/pip/qsharp/_device/_atom/__init__.py
```diff
@@ -258,6 +258,25 @@ def simulate(
...
+            if noise.s_adj.is_noiseless():
+                noise.s_adj.x = noise.rz.x
+                noise.s_adj.y = noise.rz.y
+                noise.s_adj.z = noise.rz.z
+                noise.s_adj.loss = noise.rz.loss
+            if noise.z.is_noiseless():
+                noise.z.x = noise.rz.x
+                noise.z.y = noise.rz.y
+                noise.z.z = noise.rz.z
+                noise.z.loss = noise.rz.loss
```
orpuente-MS: This logic was previously in `cpu_simulators.rs`, but it is very specific to the device; now that the `run_qir` function is being used more widely, it makes sense to move it here. Also, the GPU simulator was missing this logic, putting it here makes it common for all simulators available to the `NeutralAtomDevice` class

source/pip/qsharp/_adaptive_pass.py
```diff
@@ -160,7 +167,7 @@ class QuantumOp:
     q1: int
     q2: int
     q3: int
-    angle: float
+    angle: int
```
billti, line 184: Why is angle an int instead of float here? If there is a good reason, please add comments when things are counter-intuitive.

source/pip/qsharp/_device/_atom/__init__.py
billti: Being this isn't limited to Clifford simulators, shouldn't it also put the Rz noise on T and T_Adj?

source/pip/qsharp/_adaptive_pass.py
billti: Why is read_loss a no-op? A program may choose to branch on the result of this call.

source/pip/qsharp/_device/_atom/__init__.py
orpuente-MS: Totally, nice catch. I will add that logic.

source/pip/qsharp/_adaptive_pass.py
orpuente-MS: Added a comment explaining why.

source/pip/qsharp/_adaptive_pass.py
orpuente-MS: Main reason was that it isn't in the QIR Adaptive Spec. I am using what's described on that spec as the source of truth for the simulators. But it makes sense to add this one.

Added read_loss support for GPU, CPU, and Clifford simulation.

source/pip/qsharp/_adaptive_pass.py
```diff
@@ -192,17 +206,20 @@ class SwitchCase:
 
 @dataclass
 class IntOperand:
-    val: int = 0
+    val: int
+    bits: int
 
     def __post_init__(self):
```
billti, line 219: This doesn't detect if the value doesn't fit. Is that intentional? Copilot suggested to me the below

```python
def __post_init__(self):
    mask = (1 << self.bits) - 1
    min_val = -(1 << (self.bits - 1))
    if self.val < min_val or self.val > mask:
        raise ValueError(
            f"Value {self.val} does not fit in {self.bits} bits"
        )
    self.val = self.val & mask
```

billti (approved):

source/pip/qsharp/_adaptive_pass.py
orpuente-MS: This seems like a reasonable check. I will include it.

PR #3093, opened 2026-04-03
Enable running the VS Code extension host on the workspace (Node.js) by default

joao-boechat:
## Summary

This PR makes the Q# VS Code extension run its extension host primarily on the **workspace side** (Node.js) rather than on the UI side (browser). This is important for remote development scenarios — WSL, SSH, Codespaces, Dev Containers — where keeping the extension host close to the workspace avoids bugs and simplifies the architecture. The extension still supports running on the UI side for VS Code for Web.

To make this work, the Node.js entry point of the `qsharp-lang` npm package needed to function correctly. Rather than fixing the separate Node.js codepath, we unified both entry points into a single platform-agnostic module, using the [`web-worker`](https://www.npmjs.com/package/web-worker) package to abstract away Worker API differences between browsers and Node.js.

## Motivation

The Q# extension was originally built as a **web-first extension**, meaning the extension host always ran on the UI side (browser), even in desktop VS Code. This caused issues in remote development scenarios (WSL, SSH, Codespaces, Dev Containers) where the extension host needs to run close to the workspace for correct behavior. By making the Node.js entry point the default, the extension host runs on the workspace side, avoiding these issues.

Additionally, the npm package previously maintained two parallel codepaths — `browser.ts` and `main.ts` — each with its own wasm loading strategy, worker proxy implementation, and per-service worker scripts. The build system also produced two separate wasm-bindgen targets (`web` and `nodejs`) and used npm conditional exports to route consumers to the right entry point. The Node.js path was incomplete and buggy, and the duplication made changes error-prone. Unifying these into a single platform-agnostic entry point simplifies the codebase and eliminates this class of bugs.

### Before

```mermaid
graph TD
    subgraph "npm package entry points"
        B["browser.ts (browser entry)"]
        M["main.ts (Node.js entry)"]
    end

    B --> WB["workers/browser.ts"]
    M --> WN["workers/node.ts"]

    WB --> CWB["compiler/worker-browser.ts"]
    WB --> LSWB["language-service/worker-browser.ts"]
    WB --> DSWB["debug-service/worker-browser.ts"]

    WN --> CWN["compiler/worker-node.ts"]
    WN --> LSWN["language-service/worker-node.ts"]
    WN --> DSWN["debug-service/worker-node.ts"]

    CWB --> WASMW["lib/web/qsc_wasm.js"]
    CWN --> WASMN["lib/nodejs/qsc_wasm.cjs"]
```

### After

```mermaid
graph TD
    subgraph "npm package entry point"
        M["main.ts (single entry)"]
    end

    M --> WM["workers/main.ts (proxy, uses web-worker pkg)"]

    WW["workers/worker.ts (runs inside worker)"]

    WW --> CW["compiler/worker.ts"]
    WW --> LSW["language-service/worker.ts"]
    WW --> DSW["debug-service/worker.ts"]

    CW --> WASM["lib/web/qsc_wasm.js"]
    LSW --> WASM
    DSW --> WASM

    WM -.->|spawns| CW
    WM -.->|spawns| LSW
    WM -.->|spawns| DSW
```

### `qsharp-lang` npm package (`source/npm/qsharp/`)

- Deleted `browser.ts` and merged all functionality into `main.ts` — a single platform-agnostic entry point.
- Consolidated per-service worker pairs (`worker-browser.ts` / `worker-node.ts`) into a single `worker.ts` each.
- Replaced `workers/browser.ts` and `workers/node.ts` with `workers/worker.ts` (worker-side) and `workers/main.ts` (main-thread proxy using the `web-worker` package).
- Added `common-exports.ts` for shared re-exports.
- Only the `web` wasm-bindgen target is used; the `nodejs` target is no longer needed.
- Simplified `package.json` exports — removed conditional `browser`/`node`/`default` routing.
- Added `web-worker` (`^1.5.0`) as a dependency.
- Some APIs are now `async` (`getCompiler()`, `getDebugService()`, `getLanguageService()`, `getProjectLoader()`), and worker factory functions now require an explicit worker path argument.

### VS Code extension (`source/vscode/`)

- Introduced a `__PLATFORM__` build-time constant (`"browser"` or `"node"`) to resolve the correct worker script paths at runtime.
- The extension builds two platform bundles — one for browser and one for Node.js — each covering the extension host entry point (`extension.ts`) and the worker entry points (`compilerWorker.ts`, `debug-service-worker.ts`).
- A single wasm file is now shared across both platforms (copied to `wasm/`), which is the main reason the overall VSIX size only grew from **3.9 MB to 4.09 MB** despite adding a full Node.js bundle.
- The `web-worker` package must be marked as an **external dependency** in the Node.js esbuild bundle. This is because `web-worker` performs an internal runtime check that breaks if its code is inlined into the bundle. As a result, it is copied into the extension's `node_modules/` so it can be resolved at runtime (e.g. when installed from a VSIX).
- Desktop entry point changed to `./out/node/extension.js`.
- Logs `Platform`, `UI Kind`, and `Remote` name during activation for diagnostics.

### Build system (`build.py`)

- Only the `web` wasm-bindgen target is built. Removed the `nodejs` target and its `.js` → `.cjs` renaming logic.

### Tests (`source/npm/qsharp/test/`)

- Updated to use the unified API (`loadWasmModule`, async service getters, explicit worker paths).

### Extra

- Fixes issue [2966](https://github.com/microsoft/qdk/issues/2966)


build.py
joao-boechat, line 1: now that we're just using the web target, should I simplify the code to account for that, or do I leave the structure as is in case we want to add the nodejs target back?

ScottCarda-MS:
Since web workers was touched, I just wanted to double-check that the worker logic responsible for state-viz calculation on the circuit editor was still functional. I tested it and it seems to be working fine, but is there any reason to be concerned about that?

joao-boechat:
@ScottCarda-MS I don't see any of the worker code I touched being referenced there. I think that the worker code in state-viz is named that way because it is supposed to be agnostic (browser and worker), so it avoids using API that's only available in browser. But my modifications shouldn't affect that part!

build.py
minestarks: I would simplify.

source/npm/qsharp/src/common-exports.ts
minestarks, line 1: Can these just live in `main.ts` ? I don't see a clear logical separation of what's reexported in `main.ts` vs what's reexported here, so I'm not sure if the separate file is necessary.

joao-boechat: you're right, I did this when we still had two entrypoints and it still made sense. Will add it back to main

source/vscode/src/extension.ts
```diff
@@ -50,6 +50,12 @@ export async function activate(
     initOutputWindowLogger();
   }
 
+  log.info(`Platform: ${getPlatformEnv()}`);
```
minestarks: Can we use `log.debug` or lower for this? `info` is enabled by default, and can get kind of noisy. 

(These logs are pretty useful for testing this PR. Maybe you could clean up the logs just before merging.)

source/vscode/src/common.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+declare const __PLATFORM__: string;
```
minestarks: To make this less error prone you may want to strictly type it like this since we hardcode the values anyway.

```suggestion
declare const __PLATFORM__: "browser" | "node";
```

source/vscode/build.mjs
```diff
@@ -208,8 +303,11 @@ if (thisFilePath === resolve(process.argv[1])) {
   if (isWatch) {
     watchVsCode();
   } else {
+    buildUI();
```
minestarks: I find it kind of convoluted the way multiple parameters (`platform`, `onlyUI`) are plumbed through multiple layers of function, only to arrive at some hardcoded build configuration objects anyway.

At the end of the day we're just calling `esbuild.build` 4 times in a row with different configs, right? Couldn't we just define those configs inline here, and call `buildBundle()` directly?

Or you could put the 4 distinct configs in an array at the top of the file, and call `buildBundle` in a loop. This way they'd also be right next to `commonBuildOptions` -- they're kind of hard to find buried in the middle of the file, so it's hard to tell how the different options combine.

source/vscode/build.mjs
```diff
@@ -12,25 +12,21 @@ const thisDir = dirname(fileURLToPath(import.meta.url));
 const libsDir = join(thisDir, "..", "..", "node_modules");
 
 /** @type {import("esbuild").BuildOptions} */
-const buildOptions = {
+const commonBuildOptions = {
   entryPoints: [
```
minestarks: `entrypoints` is always overwritten, should it be defined in `commonBuildOptions` at all?

source/vscode/build.mjs
```diff
@@ -167,15 +152,116 @@ export function copyKatex(destDir) {
...
+ */
+function buildExtensionHost(platform) {
+  buildBundle(false, platform);
+}
+
+/**
+ * Copy external node dependencies into node_modules/ under the extension
+ * directory so they can be resolved at runtime (e.g. when installed as a VSIX).
+ */
+function copyNodeExternals() {
```
minestarks: This is a bit weird. If `web-worker` is a dependency we should just declare it as a dependency in `package.json` so it can be properly installed into `node_modules` by the package manager.

That said, I don't know why we can't bundle `web-worker` for node, the same way we do for the browser?

source/vscode/build.mjs
joao-boechat: You're right, it ends up not helping that much in the end of the day. I will try to make them less convoluted by removing the getBuildOptions function

source/vscode/build.mjs
joao-boechat: We can't bundle it because it would break the functionality of web-worker. The package doesn't really support bundling for node, and we have to treat it as external.

The reason I am copying the web-worker to the local directory is because we are using workspaces and automatically npm will hoist node_modules to the root directory. But I've changed it and now, instead of doing this procedure in the build script, I use postInstall to achieve the same goal in a simpler way. I also declared it as a dependency for our vscode extension and an optional dependency for npm/qsharp (as it is only needed for the node case).

source/npm/qsharp/src/workers/main.ts
minestarks, line 1: Now in this folder, we ended up with:

```
workers 
|- common.ts
|- main.ts
|- worker.ts
```

this makes me think main.ts runs main-thread side, worker.ts runs worker-thread side. So far this makes sense, but then I'd expect common.ts to run in _both_.... which isn't true. "common" used to mean "common to browser and node" but now, with your changes, the meaning has shifted to "common to main and worker". 

Would it be a pain to move stuff out of `common.ts` into the files they belong?

I'm copy-pasting how Copilot broke it down, in case it's helpful.

--- 

Your intuition is right — the separation is off. Here's what each file actually does and where the problem lies:

**main.ts** (~85 lines) — "main thread" side. Exports `createProxy()`, which spins up a `Worker`, sends it the WASM module, and returns a proxy object. It imports `createProxyInternal` from common.

**worker.ts** (~60 lines) — "worker thread" side. Exports `createWorker()`, which sets up the `onmessage` handler inside the Web Worker, initializes WASM, and wires up the service. It imports `initService` from common.

**common.ts** (~575 lines) — this is where the naming breaks down. It contains:

| What | Lines | Actually used by |
|---|---|---|
| Shared type definitions (`ServiceProtocol`, `MethodMap`, `RequestMessage`, `ResponseMessage`, etc.) | ~100 | Both |
| `serializeIfError` / `deserializeIfError` | ~50 | One each side |
| `createProxyInternal()` — proxy queue, event dispatch, state management | ~150 | **main.ts only** |
| `createDispatcher()` + `initService()` — service wrapping, logging setup | ~130 | **worker.ts only** |

So roughly **65%** of "common" is *not* common at all. `createProxyInternal` is purely main-thread logic, and `createDispatcher`/`initService` are purely worker-thread logic. The file is really "everything that isn't the thin entrypoint wrapper."

A cleaner split would be:
- **`types.ts`** — shared types and message definitions, plus error serialization helpers (the actually common stuff)
- **main.ts** — absorbs `createProxyInternal` alongside `createProxy`
- **worker.ts** — absorbs `createDispatcher` and `initService` alongside `createWorker`

This way each file owns its full logic rather than having the real implementation buried in a misleadingly-named "common" file. Want me to do the refactor?

source/vscode/src/telemetry.ts
```diff
@@ -402,6 +403,9 @@ export function sendTelemetryEvent<E extends keyof EventTypes>(
 }
 
 function getBrowserRelease(): string {
```
minestarks, line 405: I can't tell if I'm just being dense, but.... is `getBrowserRelease` or `getUserAgent` actually being used by any of the telemetry code? The only consumer I see is in `networkRequests`. If that's the case, maybe we should just move these helpers to that file.

My initial concern was whether this would impact our telemetry. (We use telemetry data to classify usage by OS, etc)

source/vscode/src/documentation.ts
```diff
@@ -22,7 +23,7 @@ export async function showDocumentationCommand(extensionUri: Uri) {
   // Get API documentation from compiler.
   const compilerWorkerScriptPath = Uri.joinPath(
     extensionUri,
-    "./out/compilerWorker.js",
+    `./out/${getPlatformEnv()}/compilerWorker.js`,
```
minestarks: can this import and use `loadCompilerWorker` from common.ts?

source/vscode/src/circuit.ts
```diff
@@ -280,7 +280,7 @@ export async function getCircuitOrErrorWithTimeout(
 
   const compilerWorkerScriptPath = Uri.joinPath(
     extensionUri,
-    "./out/compilerWorker.js",
+    `./out/${getPlatformEnv()}/compilerWorker.js`,
```
minestarks: Can this use `loadCompilerWorker` from common.ts too?

source/npm/qsharp/src/log.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-// Logging infrastructure for JavaScript environments (e.g. browser and node.js)
+// Logging infrastructure shared across browser and Node.js environments
```
minestarks: What was wrong with the old comment??

joao-boechat: Nothing, I just asked copilot to refresh all comments in the qsharp folder since I changed the structure of it, and it ended up touching that as well. Will bring that back to before.

source/vscode/src/circuit.ts
joao-boechat: Totally!

source/vscode/src/telemetry.ts
joao-boechat: You're not! But `getBrowserRelease` is being used inside of telemetry.ts, and although `getUserAgent` isn't, it uses`userAgentString`, which is defined in this module, and probably to avoid having to make it a global variable, we have the flow this way.

source/npm/qsharp/src/workers/main.ts
joao-boechat: Completely agreed! Just moved it all

PR #3103, opened 2026-04-07
RIFLA: Support iteration over arrays of qubits

swernli:
This continues extending the looping support by allowing for iteration over arrays of qubits. For some of the checked in samples, this slightly increases code size because looping over a qubit register with only two qubits is not more efficient in instruction count than just operating on each qubit directly, but for large qubit registers this ends up being a net savings.


source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1540,6 +1547,23 @@ impl<'a> PartialEvaluator<'a> {
         self.check_unresolved_call_capabilities(call_expr_id, callee_expr_id, &call_scope)?;
         self.assign_current_dbg_location(call_expr_id);
 
+        if store_item_id.package == PackageId::CORE
+            && callable_decl.name.name.as_ref() == "ReleaseQubitArray"
+        {
+            // This is a special case, where we must statically releaes the given qubits rather than call into the stdlib, which may try
```
orpuente-MS:
```suggestion
            // This is a special case, where we must statically release the given qubits rather than call into the stdlib, which may try
```

orpuente-MS (approved):

PR #3108, opened 2026-04-10
Update npm dependencies

joao-boechat:
- Updated npm dependencies to their latest possible versions without introducing bugs
- Resolved vulnerabilities in the dependencies
- Fixed breaking changes introduced by the new dependencies


package.json
```diff
@@ -17,44 +17,50 @@
...
-    "monaco-editor": "^0.44.0",
-    "openai": "^4.83.0",
-    "preact": "^10.20.0",
-    "prettier": "^3.3.3",
+    "jsdom": "^29.0.2",
+    "markdown-it": "^14.1.1",
+    "mocha": "^11.7.5",
+    "modern-normalize": "^3.0.1",
+    "monaco-editor": "^0.53.0",
+    "openai": "^6.34.0",
```
minestarks: While you're in here, can we get rid of `openai`? I accidentally left that in from an earlier feature that's long gone.
```suggestion
```

package.json
```diff
@@ -17,44 +17,50 @@
...
+    "typescript-eslint": "^8.58.1",
+    "url": "^0.11.4",
+    "user-agent-data-types": "^0.4.3",
     "vscode-uri": "^3.0.8"
+  },
+  "overrides": {
+    "mocha": {
+      "serialize-javascript": "^7.0.4",
+      "diff": "^8.0.3"
+    }
```
minestarks, line 63: Are these `overrides` really necessary? On my dev box, I deleted these overrides, deleted `package-lock.json` and `node_modules` entirely, and just did a `./build.py` and it ran clean. What if you did that to regenerate the package-lock?

source/jupyterlab/package.json
```diff
@@ -33,15 +33,20 @@
...
-    "@jupyterlab/application": "^4.0.0",
-    "@jupyterlab/codemirror": "^4.0.0",
-    "@jupyterlab/notebook": "^4.0.0"
+    "@codemirror/language": "^6.12.3",
+    "@codemirror/legacy-modes": "^6.5.1",
+    "@jupyterlab/application": "^4.5.6",
+    "@jupyterlab/codemirror": "^4.5.6",
+    "@jupyterlab/notebook": "4.4.10"
+  },
+  "overrides": {
```
minestarks, line 46: Same question about these overrides.

package.json
joao-boechat: These are to solve vulnerabilities. I tried doing what you just suggested in my local and the vulnerabilities remain, so I will keep them unless there is a better solution.

source/jupyterlab/package.json
joao-boechat: Same answer as above.

source/npm/qsharp/ux/circuit-vis/sqore.ts
```diff
@@ -80,7 +80,7 @@ export type DrawOptions = {
 export class Sqore {
   circuit: Circuit;
   gateRegistry: GateRegistry = {};
-  renderDepth: number = this.options.renderDepth ?? 0;
+  renderDepth: number;
```
ScottCarda-MS, line 83: This change seems confusing to me. Is `number` now allowed to be nullable even when not specified?

source/playground/build.js
```diff
@@ -40,7 +40,7 @@ const buildOptions = {
   outdir,
   bundle: true,
   platform: "browser",
-  target: ["es2020", "chrome64", "edge79", "firefox62", "safari11.1"],
+  target: ["es2022"],
```
ScottCarda-MS, line 43: Is it safe to remove all these other targets?

package.json
minestarks: Oh sorry, I didn't think of that. Makes sense

source/playground/build.js
joao-boechat: yes, we're basically the only ones using playground and these are all old browsers.

source/npm/qsharp/ux/circuit-vis/sqore.ts
joao-boechat: The issue was that we were accessing `this.options` even before the constructor is called. So that throws an error in the new version of typescript (pretty sure this didn't work as intended beforehand either, but typescript just let it happen).

In here we should only be defining types, as initialization usually happens in the constructor. If you want to be more precise in the type definitions, you can define a type for renderDepth and then apply that type here, but not sure if that's necessary.

minestarks (approved):

ScottCarda-MS (approved):

PR #3113, opened 2026-04-14
Add URI handler for one-click workspace connection from external sources

ScottCarda-MS:
## Summary

Adds a `vscode://` URI handler to the QDK extension that allows external websites or tools to open VS Code and automatically prompt the user to add a quantum workspace connection.

When VS Code navigates to a URI of the form:
```
vscode://quantum.qsharp-lang-vscode/connectWorkspace?connectionString=<encoded>
```
the extension presents a confirmation modal and, if accepted, saves the workspace connection identically to the existing manual "Add Workspace" flow.

## Changes

- **`package.json`**: Added `onUri` to `activationEvents` so the extension activates when a URI is handled even if it wasn't already running.
- **`workspaceActions.ts`**: Extracted connection string parsing out of `getWorkspaceWithConnectionString` into a new exported `parseConnectionString()` function. The existing manual flow delegates to it with no behavior change.
- **`commands.ts`**: `initAzureWorkspaces` now returns an `addWorkspace` helper that programmatically saves a workspace and starts its refresh cycle, reusing the existing private `saveWorkspaceList` and `startRefreshCycle` logic.
- **`extension.ts`**: Registers the URI handler. Validates the connection string and shows a user confirmation prompt before adding the workspace. User-facing messages are intentionally provider-agnostic.

## Testing

Use the following URL pattern in a browser to exercise the handler locally:
```
vscode://quantum.qsharp-lang-vscode-dev/connectWorkspace?connectionString=SubscriptionId=...;ResourceGroupName=...;WorkspaceName=...;ApiKey=...;QuantumEndpoint=...
```

Tested Scenarios:
- [x] Confirmation modal appears with the correct workspace name
- [x] Accepting adds the workspace to the Quantum Workspaces panel
- [x] Dismissing the modal does not add the workspace
- [x] A missing or malformed `connectionString` parameter shows an appropriate error message

### DIFF
--- a/source/vscode/src/extension.ts
+++ b/source/vscode/src/extension.ts
@@ -10,6 +10,7 @@ import {
 } from "qsharp-lang";
 import * as vscode from "vscode";
 import { initAzureWorkspaces } from "./azure/commands.js";
+import { parseConnectionString } from "./azure/workspaceActions.js";
 import { CircuitEditorProvider } from "./circuitEditor.js";
 import { initProjectCreator } from "./createProject.js";
 import { activateDebugger } from "./debugger/activate.js";
@@ -95,7 +96,7 @@ export async function activate(
             return;
           }
 
-          const workspace = azureApi.parseConnectionString(connStr);
+          const workspace = parseConnectionString(connStr);
           if (!workspace) {
             vscode.window.showErrorMessage(
               "The workspace URI contained an invalid connection string.",
--- a/source/vscode/src/extension.ts
+++ b/source/vscode/src/extension.ts
@@ -79,7 +79,42 @@ export async function activate(
   context.subscriptions.push(CircuitEditorProvider.register(context));
   context.subscriptions.push(...registerChangelogCommand(context));
 
-  await initAzureWorkspaces(context);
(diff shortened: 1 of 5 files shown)


joao-boechat:
Not sure if it applies here, but I would make sure to test whether this will work on vscode-insiders, as we might have a similar bug to the one I addressed on this PR: https://github.com/microsoft/qdk/pull/2989

minestarks (approved):

source/vscode/src/extension.ts
```diff
@@ -79,7 +80,42 @@ export async function activate(
   context.subscriptions.push(CircuitEditorProvider.register(context));
   context.subscriptions.push(...registerChangelogCommand(context));
 
-  await initAzureWorkspaces(context);
+  /// Handle incoming workspace connection URIs from the Azure portal. The URI will be in the format:
```
minestarks: Consider pulling this out into its own file since I can see us implementing similar URI handlers in the future, this seems pretty handy!

ScottCarda-MS: This is a great idea! Done

PR #3121, opened 2026-04-15
Use separate browser/node entrypoints instead of runtime environment detection

minestarks:
This PR takes us back to splitting the package entrypoint using the `exports` field's conditional `"browser"` / `"node"` conditions:

- browser.ts — re-exports main.ts as-is (browsers natively support `Worker` with `type: "classic"`)
- node.ts — polyfills `globalThis.Worker` via `web-worker` and calls `setWorkerType("module")` before re-exporting main.ts

The `isBrowser` runtime detection is removed entirely. Instead, the worker type is passed explicitly as a parameter to `createProxy`.

### DIFF
--- a/source/npm/qsharp/src/browser.ts
+++ b/source/npm/qsharp/src/browser.ts
@@ -1,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Browser entrypoint. No Worker polyfill needed — browsers provide it natively.
+
+export * from "./main.js";
--- a/source/npm/qsharp/src/browser.ts
+++ b/source/npm/qsharp/src/browser.ts
@@ -1,6 +1,0 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-
-// Browser entrypoint. No Worker polyfill needed — browsers provide it natively.
-
-export * from "./main.js";
--- a/source/npm/qsharp/src/node.ts
+++ b/source/npm/qsharp/src/node.ts
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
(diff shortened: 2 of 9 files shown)


source/npm/qsharp/src/browser.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Browser entrypoint. No Worker polyfill needed — browsers provide it natively.
+
+export * from "./main.js";
```
joao-boechat: instead of having this extra file, we can also just have the entrypoint for the browser be src/main.ts

source/npm/qsharp/src/node.ts
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+
+// Node.js entrypoint. Polyfills the Worker global before loading the main module.
+
+import worker from "web-worker";
+import { setWorkerType } from "./main.js";
+
+if (typeof globalThis.Worker === "undefined") {
+  globalThis.Worker = worker;
+}
```
joao-boechat, line 11: I don't think these lines are necessary anymore, it was just for detecting whether we were in a node environment. We can just use
`import Worker from "web-worker"` or something similar

ScottCarda-MS (approved):

source/npm/qsharp/src/node.ts
minestarks: I do have to set `globalThis.Worker` , otherwise `Worker` will be undefined for the other files that use the global `Worker`. Just importing in this module isn't sufficient, AFAIK

source/npm/qsharp/src/browser.ts
minestarks: Done

joao-boechat (approved):

PR #3123, opened 2026-04-15
Add deep links to Quantum OS portal for V2 workspaces and jobs

ScottCarda-MS:
## Summary

Adds portal deep link buttons to the Quantum Workspaces panel for workspaces
and jobs hosted on V2 (Quantum OS) endpoints.

### Workspace links
The existing "Open the workspace in the web portal" button on workspace tree
items now links to the Quantum OS portal (`manage.quantum.microsoft.com`)
for V2 workspaces, instead of the Azure Portal. V1 workspaces are unchanged.

The URL includes fragment parameters so the portal can navigate directly to
the correct workspace: `tenantId`, `subscriptionId`, `role`, `offeringId`,
and `workspaceId`.

### Job links
A new "Open the job in the web portal" inline button (`$(link-external)`) is
added to job tree items, but only for jobs under V2 workspaces. This is
achieved by appending a `-v2` suffix to the `contextValue` of job tree items
that belong to V2 workspaces, and gating the button's `when` clause on those
context values.

All existing job buttons (delete, cancel, download results) are updated to
also match the `-v2` context values so they continue to work for V2 workspace
jobs.

## Files changed

- `src/azure/workspaceActions.ts` – `getWorkspacePortalLink` updated for V2;
  new `getQuantumOsJobLink` function added
- `src/azure/treeView.ts` – V2 detection added to job tree item construction;
  `-v2` suffix applied to `contextValue` for V2 workspace jobs
- `src/azure/commands.ts` – `jobOpenPortal` command registered
- `package.json` – `jobOpenPortal` command defined; `view/item/context` menu
  entries added/updated for all job buttons

## Screenshots

- Job details link:
image

- Workspace details link:
image


source/vscode/src/azure/treeView.ts
```diff
@@ -250,34 +251,39 @@ export class WorkspaceTreeItem extends vscode.TreeItem {
       case "job": {
         const job = itemData as Job;
         this.collapsibleState = vscode.TreeItemCollapsibleState.None;
+        const endpointMatch = workspace.endpointUri.match(
```
minestarks: I found 4 spots in the code now where we match against `endpointRegExp` . I wonder if there's an opportunity to extract to a function like `parseEndpointUri` which returns all the relevant parts.

minestarks (approved):
looks good (and works well on my machine) - small refactoring suggestion

source/vscode/src/azure/treeView.ts
ScottCarda-MS: Done, now we don't have to expose a regular expression 😊

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -20,10 +20,44 @@ import { getRandomGuid } from "../utils";
...
-export function getAzurePortalWorkspaceLink(workspace: WorkspaceConnection) {
-  // Portal link format:
-  // - https://portal.azure.com/#resource/subscriptions/<sub guid>/resourceGroups/<group>/providers/Microsoft.Quantum/Workspaces/<name>/overview
+export function getQuantumOsJobLink(
+  workspace: WorkspaceConnection,
+  jobId: string,
+) {
+  // Quantum OS job page format:
+  // https://manage.quantum.microsoft.com/jobs/<job-id>
+  return `https://manage.quantum-test.microsoft.com/jobs/${jobId}`;
```
billti, line 29: Should `https://manage.quantum-test.microsoft.com` be extracted to a constant or setting? I assume this will lose the `-test` at some point.

ScottCarda-MS: Hmm, having it be a VS Code setting might make it easier to dealt with the lock-step issue between the Quantum OS development and the VS Code extension development.

PR #3130, opened 2026-04-17
Upgrade pyqir to v0.12.3

orpuente-MS:
This PR upgrades our pyqir dependency to v0.12.3


swernli (approved):
Changes look good! We should just wait to merge until we've verified compatibility with qdk-chemistry.

PR #3136, opened 2026-04-21
Extend extension integration-tests for node environment

joao-boechat:
This PR adds Electron (Node.js) test execution to the VS Code integration test suites, which previously only ran in the browser via `@vscode/test-web`. Tests now run in both environments to catch platform-specific issues.


source/vscode/test/suites/debugger/qsharp.test.ts
```diff
@@ -15,6 +15,9 @@ suite("Q# Debugger Tests", function suite() {
   const fooUri = vscode.Uri.joinPath(workspaceFolder.uri, "src", "foo.qs");
   const barUri = vscode.Uri.joinPath(workspaceFolder.uri, "src", "bar.qs");
 
+  // Makes testing compatible with node and web environments
+  const separator = workspaceFolder.uri.path.endsWith("/") ? "" : "/";
```
billti, line 19: Calling this `separator` was a little confusing. I was expecting like `\` for Windows else `/` or something. I see now it is that on some platforms a separator is not needed, hence the "", so it is more like an optional separator. Can you add a comment or rename to make that clear?

billti (approved):

minestarks (approved):

PR #3156, opened 2026-04-23
Get Tenant Id from Subscription Id using ARM Unauthentication Challenge

ScottCarda-MS:
The PR makes 2 changes to the deep links to the Quantum OS workspaces/jobs:
- Workspaces added via connection string do not have tenant ids on them by default. We add a mechanism to get the tenant id with a subscription id.
- Added to job links the workspace information, including tenant id.

### DIFF
--- a/source/vscode/src/azure/commands.ts
+++ b/source/vscode/src/azure/commands.ts
@@ -39,6 +39,7 @@ import {
   getQuantumOsJobLink,
   getJobFiles,
   getPythonCodeForWorkspace,
+  getTenantIdForSubscription,
   parseConnectionString,
   queryWorkspaces,
   submitJob,
@@ -67,6 +68,21 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
     log.debug("Loading workspaces: ", savedWorkspaces);
     const workspaces: WorkspaceConnection[] = JSON.parse(savedWorkspaces);
     for (const workspace of workspaces) {
+      // Backfill tenant ID for workspaces saved before tenant discovery was added,
+      // or for connection-string/deep-link workspaces that didn't have it populated.
+      if (!workspace.tenantId && !workspace.apiKey) {
+        // AAD workspace with missing tenant ID — shouldn't normally happen,
+        // but recover gracefully.
+        log.warn("AAD workspace loaded without a tenant ID:", workspace.id);
+      } else if (!workspace.tenantId) {
+        const idRegex = /\/subscriptions\/(?<subscriptionId>[^/]+)/;
+        const subscriptionId =
+          workspace.id.match(idRegex)?.groups?.subscriptionId;
+        if (subscriptionId) {
+          workspace.tenantId =
+            (await getTenantIdForSubscription(subscriptionId)) ?? "";
+        }
+      }
       workspaceTreeProvider.updateWorkspace(workspace);
       // Start refreshing each workspace until pending jobs are complete
       startRefreshCycle(workspaceTreeProvider, workspace);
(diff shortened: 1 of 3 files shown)


source/vscode/src/azure/workspaceActions.ts
```diff
@@ -34,8 +34,24 @@ export function getQuantumOsJobLink(
...
+  // <quantumOsRoot>/jobs/<job-id>#<workspace-fragment>
+  const fragment = buildQuantumOsFragment(workspace);
+  return `${getQuantumOsRoot()}/jobs/${jobId}#${fragment}`;
+}
+
+function buildQuantumOsFragment(workspace: WorkspaceConnection): string {
+  const idRegex = /\/subscriptions\/(?<subscriptionId>[^/]+)\/resourceGroups\//;
+  const subscriptionId =
+    workspace.id.match(idRegex)?.groups?.subscriptionId ?? "";
+  const offeringId = workspace.providers[0]?.providerId ?? "";
```
minestarks: Seems just more efficient & ergonomic to save `subscriptionId` and `offeringId` as properties of `workspace`, same as you do with `tenantId`?

ScottCarda-MS: done

source/vscode/src/azure/commands.ts
```diff
@@ -67,6 +68,21 @@ export async function initAzureWorkspaces(context: vscode.ExtensionContext) {
     log.debug("Loading workspaces: ", savedWorkspaces);
     const workspaces: WorkspaceConnection[] = JSON.parse(savedWorkspaces);
     for (const workspace of workspaces) {
+      // Backfill tenant ID for workspaces saved before tenant discovery was added,
+      // or for connection-string/deep-link workspaces that didn't have it populated.
+      if (!workspace.tenantId && !workspace.apiKey) {
```
billti, line 73: This comment and logic don't seem to match for me. It says `...or for connection-string... that didn't have it populated`, but on reading the `if` statement, if it's not for a connection string __AND__ there is no tenant ID, then it doesn't backfill at all, it just logs the warning below. Unless I'm misreading something.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -34,8 +34,27 @@ export function getQuantumOsJobLink(
...
+}
+
+function buildQuantumOsFragment(workspace: WorkspaceConnection): string {
+  const idRegex = /\/subscriptions\/(?<subscriptionId>[^/]+)\/resourceGroups\//;
+  const subscriptionId =
+    workspace.subscriptionId ??
+    workspace.id.match(idRegex)?.groups?.subscriptionId ??
+    "";
+  const offeringId =
+    workspace.offeringId ?? workspace.providers[0]?.providerId ?? "";
```
billti, line 49: This doesn't seem right to me. I don't think it's safe to assume just one provider and that the order it deterministic/correct. If this is only for 'back compat' purposes for now with the assumption that any existing V2 workspaces would have only had one provider, then state that assumption in a comment.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -34,8 +34,27 @@ export function getQuantumOsJobLink(
...
+
+  return (
+    `tenantId=${workspace.tenantId}` +
+    `&subscriptionId=${subscriptionId}` +
+    `&role=Researcher` +
+    (offeringId ? `&offeringId=${offeringId}` : "") +
+    `&workspaceId=${workspace.id}`
+  );
```
billti, line 57: I'd like to discuss this cross team and understand the reasoning. As written, this seems pretty clunky. I'm not sure why you need the tenant in here if one tenant per subscription (as implied by `getTenantIdForSubscription`). Similarly for the offering. (Also passing it in the URL fragment rather than query params seems odd, but maybe that's for security reasons - I'd like to understand).

billti (approved):

PR #3159, opened 2026-04-26
[OUTDATED] Memory Qubits

fedimser:
This PR adds support for new Q# type `MemoryQubit` and extends standard library with operations on it.

A “Memory Qubit” is a useful abstraction in quantum computing. It is a logical qubit on which gates or measurements are not performed directly. It can only hold a quantum state. When we want to do something with that state, we must “load” it to a regular (“compute”) qubit. When we want to retain state of compute qubit for a while but don’t perform computations, we can “store” it to a memory qubit. 

This PR contains:
* MemoryQubit type definition and changes to compiler to support it.
* Definitions of 4 intrinsics to allocate, free, load and store qubits.
* Compiler changes to support `use` syntax (similarly to Qubits) and translate it to allocation instrinsic.
* Compiler changes to automatically free MemoryQubit when going out of scope.
* New 3 methods in Backend (`memory_qubit_allocate`, `memory_qubit_load`, `memory_qubit_store`). Together with existing `qubit_free`, this is the interface between the Q# language and backends.
* Supported MemoryQubits in resource estimation backend (count.qs). In other backends use default implementation that uses Qubits as MemoryQubits and Reset+SWAP for stores and loads (which is correct for simulation).
* Added Std.MemoryQubits which defines operations to work with memory qubits (Load, Store, LoadArray, StoreArray, DoComputation) that internally delegate to 2 intrinsics (load/store).
* Added short description of MemoryQubits in Std.MemoryQubits.
* Added tests for syntax, simulator and resource estimation in single rust test file.


library/src/tests.rs
```diff
@@ -40,6 +44,35 @@ pub fn test_expression_fails(expr: &str) -> String {
     )
 }
 
+/// Asserts that given Q# expression fails to compile with given error message.
+pub fn test_compile_fails(expr: &str, lib: &str, expected_error_substring: &str) {
```
swernli, line 48: Is this new function necessary? There are other tests that use `test_expression_fails` defined above and then expect on the output string. Would that be sufficient for new tests?

library/src/tests.rs
```diff
@@ -264,3 +297,33 @@ fn stdlib_reexport_single_case() {
...
+}
+
+fn panic_with_resource_estimation_errors(errs: &[ResourceEstimatorError]) -> ! {
+    let joined = errs
+        .iter()
+        .map(|e| format!("{e}:{e:?}"))
+        .collect::<Vec<_>>()
+        .join("\n");
+    panic!("resource estimation failed:\n{joined}");
+}
```
swernli, line 329: Rather than have a utility to panic with all the errors joined, you could just use the first error returned. Most of the APIs other than compile use a vector of errors for compatibility sake with parts of the infra but only ever return one error.

library/std/src/Std/MemoryQubits.qs
```diff
@@ -0,0 +1,111 @@
...
+/// # Description
+/// This operation allocates a temporary `Qubit[]` buffer, loads `mem_qs` into that
+/// buffer, applies `op`, and stores the resulting state back into `mem_qs`.
+///
+/// # Input
+/// ## mem_qs
+/// Memory register to transform.
+/// ## op
+/// Operation to apply to the temporary compute buffer.
+operation DoComputation(mem_qs : MemoryQubit[], op : Qubit[] => Unit) : Unit {
```
swernli, line 104: I feel like "do" reads weird here... I think `ApplyComputation` would better match our naming patterns.

library/std/src/QIR/Intrinsic.qs
```diff
@@ -105,6 +105,15 @@ operation __quantum__qis__mresetz__body(target : Qubit) : Result {
     body intrinsic;
 }
 
+// MemoryQubit operations.
+operation __quantum__qis__memory_qubit_load(memory_qubit : MemoryQubit, qubit : Qubit) : Unit {
```
swernli, line 109: To follow QIR naming conventions, things with `__qis__` should end in `__body`, making this one `__quantum__qis__memory_qubit_load__body`. I think the "qubit" part may be a bit redundant, since we already have "quantum" and "qis" which stands for Quantum Intruction Set, so `__quantum__qis__memory_load__body` would be fine. Likewise the corresponding operation below would be `__quantum__qis__memory_store__body`.

Side note: it seems awkward to add `__body` but it is really only to differentiation between the inverse/adjoint operations, which have the `__adj` suffix.

samples/language/MemoryQubit.qs
```diff
@@ -0,0 +1,18 @@
...
+
+operation Main() : Result {
+    use (q, mem) = (Qubit(), MemoryQubit());
+
+    X(q);
+    Store(q, mem);
+    Load(mem, q);
+
+    return MResetZ(q);
+}
```
swernli, line 18: I think this sample could use some additional explanatory text in the top level description and comments in the body itself, like our other language samples do. The motivation is both to help educate users and to act as guidance for LLMs, so it's woth explaining the patterns and motivations here.

source/compiler/qsc_parse/src/stmt/tests.rs
```diff
@@ -82,6 +82,33 @@ fn use_qubit_array() {
...
+        &expect![[r#"
+            Stmt _id_ [0-22]: Qubit (Fresh)
+                Pat _id_ [4-5]: Bind:
+                    Ident _id_ [4-5] "m"
+                QubitInit _id_ [8-21] MemorySingle"#]],
+    );
+}
+
+#[test]
+fn use_memory_qubit_array() {
```
swernli, line 99: It would be worth adding some negative tests for the rule that memory qubits cannot be borrowed, both when allocated directly and when in a mixed allocation with qubits.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1766,6 +1770,56 @@ impl<'a> PartialEvaluator<'a> {
             "__quantum__qis__mresetz__body" => {
                 Ok(self.measure_qubit(builder::mresetz_decl(), args_value))
             }
+            "__quantum__qis__memory_qubit_load" | "__quantum__qis__memory_qubit_load__body" => {
```
swernli, line 1773: Once the name fix in the libraries is added, only the `__body` name is needed here. Same for the case below.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1766,6 +1770,56 @@ impl<'a> PartialEvaluator<'a> {
             "__quantum__qis__mresetz__body" => {
                 Ok(self.measure_qubit(builder::mresetz_decl(), args_value))
             }
```
swernli, line 1772: Looks like this file is missing some clippy lint clean up. The two cases for load and store are identical, since the arguments ordering is enough to change the directionality of the operation.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -3724,6 +3811,12 @@ impl<'a> PartialEvaluator<'a> {
                     .try_into()
                     .expect("could not convert qubit ID to u32"),
             )),
+            Value::MemoryQubit(q) => Operand::Literal(Literal::Qubit(
```
swernli, line 3814: This can be combined with the qubit case above.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -3713,6 +3768,38 @@ impl<'a> PartialEvaluator<'a> {
...
+            return *id;
+        }
+
+        let callable = Callable {
+            name: "__quantum__qis__swap__body".to_string(),
+            input_type: vec![rir::Ty::Prim(rir::Prim::Qubit), rir::Ty::Prim(rir::Prim::Qubit)],
+            output_type: None,
+            body: None,
+            call_type: CallableType::Regular,
+        };
```
swernli, line 3795: to match the pattern we have above, this should be adder to the `builder` class instead of inlined here.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1745,10 +1745,14 @@ impl<'a> PartialEvaluator<'a> {
         // There are a few special cases regarding intrinsic callables. Identify them and handle them properly.
         match callable_decl.name.name.as_ref() {
             // Qubit allocations and measurements have special handling.
-            "__quantum__rt__qubit_allocate" | "__quantum__rt__qubit_borrow" => {
+            "__quantum__rt__qubit_allocate"
+            | "__quantum__rt__qubit_borrow"
+            | "__quantum__rt__memory_qubit_allocate" => {
```
swernli, line 1750: no action on this, just a note to confirm understanding: this is consistent with the goal of supporting memory qubit programs as normal compute programs in QIR, as this does not guarantee that the two qubit types will use distinct pools, and if/when we want to support memory qubits as a distinct pool of ids we'd want to differentiate the two kinds of allocation/release calls with separate management (among other changes, like emitting load and store directly). Does that match your expectation as well?

swernli (commented):
Reviewed code so far, and provided some feedback. Also, it looks like the pipeline is blocked due to some pending merge conflicts.

PR #3162, opened 2026-04-26
Enable verifying notebooks using qdk in integration tests

swernli:
This updates the build script to make notebook testing dependent on building the qdk wheel and installs it alsong with dependencies during integration testing to allow validation of notebooks that use the qdk package.

This will allow follow up that updates the older sample notebooks to use the `from qdk import ...` pattern.

### DIFF
--- a/build.py
+++ b/build.py
@@ -675,7 +675,7 @@ if build_jupyterlab:
     subprocess.run(pip_build_args, check=True, text=True, cwd=jupyterlab_src)
     step_end()
 
-if build_pip and build_widgets and args.integration_tests:
+if build_pip and build_widgets and build_qdk and args.integration_tests:
     step_start("Running notebook samples integration tests")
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -710,8 +707,10 @@ def run_ci_historic_benchmark():
         "install",
         "--force-reinstall",
         "--no-index",
+        "--no-deps",
```
billti, line 710: Do any of the notebooks require 'pyqir'? I thought they did. That's a dependency of the `qdk` needed from some tests. (Or does it depend on test_requirements.txt to ensure those are installed?)

build.py
```diff
@@ -737,6 +736,7 @@ def run_ci_historic_benchmark():
         "nbconvert",
         "pandas",
         "qutip",
+        "pyqir",
```
billti, line 739: Doh! :-)

billti (approved):

PR #3164, opened 2026-04-27
Enable Clifford simulation in `qsharp.run`

swernli:
This change adds a new `type` parameter to `qsharp.run` that allows users to specify the simulation type they want to use, with the default the existing sparse simulation and a new option of "clifford" simulation. There is an additional parameter that allows specifying the number of qubits to use for Clifford simulation.

This is accomplished by updated the `Backend` trait to have most functions be fallible, returning a `Result` type. This allows Clifford simulation to fail gracefully on non-Clifford operations and qubit allocation beyond the configured amount.

### DIFF
--- a/source/pip/qsharp/_qsharp.py
+++ b/source/pip/qsharp/_qsharp.py
@@ -32,6 +32,7 @@ from typing import (
     List,
     Set,
     Iterable,
+    Literal,
     cast,
 )
 from .estimator._estimator import (
@@ -779,6 +780,8 @@ def run(
     ] = None,
     qubit_loss: Optional[float] = None,
     seed: Optional[int] = None,
+    type: Optional[Literal["sparse", "clifford"]] = None,
+    num_qubits: Optional[int] = None,
 ) -> List[Any]:
     """
     Runs the given Q# expression for the given number of shots.
@@ -793,6 +796,9 @@ def run(
     :param noise: The noise to use in simulation.
     :param qubit_loss: The probability of qubit loss in simulation.
     :param seed: The seed to use for the random number generator in simulation, if any.
+    :param type: The type of simulator to use. If not specified, the default sparse state vector simulation will be used.
+    :param num_qubits: The number of qubits to use for the simulation type "clifford".
+        If not specified, the Clifford simulator assumes a default of 1000 qubits.
 
     :return: A list of results or runtime errors. If ``save_events`` is true, a list of ``ShotResult`` is returned.
     :rtype: List[Any]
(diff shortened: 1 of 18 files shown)


source/pip/qsharp/_qsharp.py
```diff
@@ -834,6 +840,12 @@ def on_save_events(output: Output) -> None:
         elif output.is_message():
             results[-1]["messages"].append(str(output))
 
+    if type is not None and type == "clifford":
```
orpuente-MS: Can this just be `if type == "clifford":`? Or did you need the `type is None` check to make Pylance happy?

source/pip/tests/test_clifford_simulator.py
```diff
@@ -207,3 +222,358 @@ def test_cy_direct_qir():
...
+        "IsingModel2DEvolution(4, 4, PI() / 2.0, PI() / 2.0, 10.0, 10)",
+        shots=1,
+        noise=noise,
+        seed=17,
+        type="clifford",
+    )
+    result = [result_array_to_string(cast(Sequence[Result], x)) for x in output]
+    print(result)
+    # Reasonable results obtained from manual run
+    assert result == ["0000000011000001"]
```
orpuente-MS, line 278: This kind of test has been quite flaky lately. I deleted them in one of my previous PR since they are already covered by more stable tests.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -472,31 +613,34 @@ impl Simulator for StabilizerSimulator {
...
+    let mut normalized_angle = angle % TAU;
+    if normalized_angle < 0.0 {
+        normalized_angle += TAU;
+    }
+    if normalized_angle.is_nearly_zero() || (normalized_angle - TAU / 2.0).is_nearly_zero() {
+        // The angle is a multiple of 2 * PI, so the operation is effectively an identity.
+        UnitaryOp::I
+    } else if (normalized_angle - PI).is_nearly_zero() {
+        // The angle is an odd multiple of PI, so the operation is effectively a Pauli gate.
+        pauli
```
orpuente-MS: `PI := TAU / 2.0`, I think these two branches overlap. Might be worth it adding a test to catch this before fixing it.

source/pip/qsharp/_qsharp.py
swernli: I think you are right, and I didn't see any additional complaints from the linter, so it should be fine to remove the redundant check.

source/pip/tests/test_clifford_simulator.py
swernli: With the fixed seed, the perfomance should be predictable, right? I haven't seen this one fail, and if others are failing when using this pattern I wonder if it indicates a place where the seed isn't plumbed through as expected?

source/simulators/src/stabilizer_simulator.rs
swernli: Ooo, good catch! Only the first case is supposed to use Tau to convert to identity. I'll add a test case as you suggest.

source/pip/tests/test_clifford_simulator.py
orpuente-MS: Oh, what I mean by flaky is that when you change the simulator's code, the expected value will change, even though the behavior didn't really change. This has been happening to me quite a lot lately, mainly because I have been changing the simulators' code. So, it's better to check the histogram's probabilities are within some threshold, to make the test more stable.

orpuente-MS (approved):

PR #3167, opened 2026-04-27
Default to OpenQASM semantics on compile

swernli:
This updates the code for `openqasm.compile` to default to OpenQASM output semantics instead of Qiskit during compile. This allows for individual bits as output, as well as maintaining the array ordering for registers rather than reversing them.


idavis (approved):

swernli:

PR #3185, opened 2026-04-28
Provide lhs_span to binop errors when necessary

### LINKED ISSUE #2581: Binop errors should be able to use span for entire expression
Right now, any error in a binary operation expression during evaluation always uses the span for the right-hand side, as that's the only span passed through. But for some errors it makes more sense to highlight the left-hand side or the entire expression. We should update the parameters for the binop helpers to accept multiple spans so the individual error cases can choose the most appropriate one.

joao-boechat:
This PR resolves https://github.com/microsoft/qdk/issues/2581 by separating cases when errors are in the left-hand side and printing the appropriate span with the error message.

## Examples:

Error is in the left-hand side:
image

Error is in the right-hand side:
image

### DIFF
--- a/source/compiler/qsc_eval/src/lib.rs
+++ b/source/compiler/qsc_eval/src/lib.rs
@@ -977,6 +977,7 @@ impl State {
             ExprKind::ArrayRepeat(..) => self.eval_arr_repeat(expr.span)?,
             ExprKind::Assign(lhs, _) => self.eval_assign(env, globals, *lhs)?,
             ExprKind::AssignOp(op, lhs, rhs) => {
+                let lhs_span = globals.get_expr((self.package, *lhs).into()).span;
                 let rhs_span = globals.get_expr((self.package, *rhs).into()).span;
                 let (is_array, is_unique) =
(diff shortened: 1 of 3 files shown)


ScottCarda-MS:
Should we instead have the whole comparison operation underlined as the source of the issue? After all either side of the operation is valid by themselves; it is only when the comparison between the two happens that there is an error.
@swernli did you have some thoughts on this?

swernli:
> Should we instead have the whole comparison operation underlined as the source of the issue? After all either side of the operation is valid by themselves; it is only when the comparison between the two happens that there is an error. @swernli did you have some thoughts on this?

That's an interesting point, and as I think about it, I'm torn. On the one hand, yes, the comparison is the source of the error, so the whole expression is incorrect. On the other hand, during runtime we have enough information to know which side is the source of the un-comparable value, so underlining that side gives the user more information about where the offending value is coming from. That can be helpful when the expression is something comparing two variables and the user doesn't know which one is loss, ie `a == b`.

All that said, I think this change is a fix for the bug, which was that we always underlined the right side only, so I'd be fine taking this bug fix as the current improvement and punting on that debate.

swernli (approved):

PR #3186, opened 2026-04-29
Compute runtime of a trace

msoeken:
This computes the runtime of a trace using the existing Tetris-based scheduler. Before this functionality was only used in the `Trace::estimate` function, but can now also be called from the Python API as `Trace.runtime`.


brad-lackey (approved):

msoeken:
Closing this one and opened #3209 after merging failed.

PR #3192, opened 2026-05-04
Move contents of `qsharp` python package to `qdk` python package

ScottCarda-MS:
# PR: Migrate `qsharp` pip package → `qdk`

## Summary

This PR migrates all Python source code, Rust native extension code, tests, and build infrastructure from the `qsharp` pip package (`source/pip/`) into the `qdk` package (`source/qdk_package/`). After this change:

- **`qdk`** is the primary Python package containing all source code and the native Rust extension.
- **`qsharp`** becomes a thin, pure-Python deprecation shim that re-exports from the `qdk` package.

## Motivation

The repo is being rebranded from `qsharp` to `qdk`. Rather than maintaining two packages with real code, all functionality consolidates into `qdk`, and `qsharp` exists solely for backward compatibility during the transition period.

---

### 1. Rust native extension moved (`source/pip/src/` → `source/qdk_package/src/`)

The pyo3 native module (`_native`) and all its Rust source files (interpreter, QIR simulation, noisy simulator bindings, resource estimator, etc.) were moved from `source/pip/` into `source/qdk_package/`. The `Cargo.toml` at `source/qdk_package/` now defines the `qdk` crate (previously `qsharp`), and the root `Cargo.toml` workspace member was updated accordingly.

**Files moved (all "pure" renames):**
- `src/lib.rs`, `src/interpreter.rs`, `src/qir_simulation.rs`, `src/noisy_simulator.rs`, `src/qre.rs`, `src/fs.rs`, `src/interop.rs`, `src/generic_estimator/`, `src/displayable_output/`, `src/state_*_template.html`, and all sub-modules.

### 2. Python source reorganized (`source/pip/qsharp/` → `source/qdk_package/qdk/`)

All Python modules were moved from `qsharp.*` to `qdk.*` with import paths updated throughout.

**Key structural changes:**

| Before (`qsharp.*`) | After (`qdk.*`) | Notes |
|---|---|---|
| `qsharp/__init__.py` | `qdk/__init__.py` | New minimal root; exposes common utilities |
| `qsharp/_qsharp.py` | `qdk/_interpreter.py` + `qdk/_types.py` | Split: interpreter functions vs. type definitions |
| `qsharp/_simulation.py` | `qdk/simulation/_simulation.py` | Moved into `simulation/` subpackage |
| `qsharp/noisy_simulator/` | `qdk/simulation/_noisy_simulator.py` | Absorbed into `simulation/` subpackage |
| `qsharp/interop/qiskit/` | `qdk/qiskit/` | Promoted from nested `interop` to top-level subpackage |
| `qsharp/interop/cirq/` | `qdk/cirq/` | Promoted from nested `interop` to top-level subpackage |
| `qsharp/utils/_utils.py` | *(deleted)* | `dump_operation` moved into `_interpreter.py` |
| `qsharp/estimator/` | `qdk/estimator/` | Direct move, imports updated |
| `qsharp/openqasm/` | `qdk/openqasm/` | Direct move, imports updated |
| `qsharp/code/` | `qdk/code/` | Direct move |
| `qsharp/applications/` | `qdk/applications/` | Direct move |
| `qsharp/qre/` | `qdk/qre/` | Direct move, imports updated |
| `qsharp/_device/` | `qdk/_device/` | Direct move, circular import fixed (see below) |

**New `qdk` public API surface:**
- `qdk.qsharp` — Q# interpreter functions (`init`, `eval`, `run`, `compile`, `circuit`, `estimate`, etc.)
- `qdk.simulation` — Simulation APIs (`NeutralAtomDevice`, `NoiseConfig`, noisy simulator types)
- `qdk.qiskit` — Qiskit interop (`QSharpBackend`, `NeutralAtomBackend`, etc.)
- `qdk.cirq` — Cirq interop
- `qdk.estimator` — Resource estimator
- `qdk.openqasm` — OpenQASM compilation/execution
- `qdk.code` — Code analysis
- `qdk.applications` — Domain applications (magnets, etc.)
- `qdk.qre` — QRE v3

### 3. `qsharp` package converted to deprecation shim (`source/pip/`)

`source/pip/qsharp/__init__.py` now:
1. Emits a `DeprecationWarning` on import.
2. Re-exports the full public API from `qdk._types`, `qdk._interpreter`, and `qdk._native`.
3. Registers IPython magics from `qdk._ipython`.

All other Python files under `source/pip/qsharp/` are now thin re-export wrappers that import from their `qdk.*` counterparts. The package metadata in `source/pip/pyproject.toml` declares `dependencies = ["qdk==0.0.0"]` (version stamped at CI time).

### 4. Tests moved

- **Unit tests** remain at `source/qdk_package/tests/` — import paths updated from `qsharp.*` to `qdk.*`.
- **Integration tests** moved from `source/pip/tests-integration/` to `source/qdk_package/tests-integration/` — all imports updated to use `qdk.*`.

### 5. Build script changes (`build.py`)

- **`--qdk` flag**: Builds the maturin wheel (Rust + Python) and runs unit tests. Also runs integration tests when `--integration-tests` is passed.
- **`--pip` flag**: Builds only the pure-Python `qsharp` shim wheel via setuptools. No longer runs any tests (integration tests moved to `--qdk`).
- **Removed `install_qsharp_python_package()`**: No longer needed since integration tests don't depend on the `qsharp` shim.
- Install commands use `--no-deps --no-index` to install from local wheels without reaching PyPI.

### 6. CI/CD pipeline changes

**GitHub Actions (`ci.yml`)** — No changes needed. The `integration-tests` job already passed both `--qdk` and `--integration-tests`.

**Azure DevOps (`publish.yml`)** — Restructured for clean platform split:

| Job | Before | After |
|---|---|---|
| `Platform_Agnostic_Python` | `--jupyterlab --widgets --qdk` | `--jupyterlab --widgets --pip` |
| Per-platform matrix | `--pip --integration-tests` | `--qdk --integration-tests` |

This restructuring reflects the fact that `qdk` is now the platform-specific package (it contains the native Rust extension), while `qsharp` is now platform-agnostic (pure-Python shim). Previously it was the reverse: `qsharp` held the native code and `qdk` was a pure-Python meta-package. Each of the 6 OS/arch combinations now builds its own native `qdk` wheel, while the platform-agnostic wheels (`qsharp`, widgets, jupyterlab) are built once.

### 7. Circular import fix (`_device._atom` ↔ `simulation`)

`simulation/__init__.py` imports `NeutralAtomDevice` from `_device._atom`, while `_device._atom` needs `NoiseConfig` and `run_qir_*` from `simulation._simulation`. This was resolved by:
- Using `from __future__ import annotations` in `_device/_atom/__init__.py`
- Guarding `NoiseConfig` import behind `TYPE_CHECKING`
- Deferring runtime imports of `run_qir_*` into the `simulate()` method body

### 8. Miscellaneous

- **`.prettierignore`**: Added `source/qdk_package/src/**/*.html` and `source/qdk_package/tests-integration/**/*.inc` (these files moved from `source/pip/` which was fully ignored).
- **`.github/CODEOWNERS`**: Updated paths from `source/pip/` to `source/qdk_package/`.
- **`.github/copilot-instructions.md`**: Updated architecture documentation.
- **`Cargo.toml` (root)**: Updated workspace member from `source/pip` to `source/qdk_package`.
- **Sample notebooks**: Updated `import qsharp` to `import qdk` / `from qdk import qsharp`.

---

## How to Review

Due to the large number of file moves, GitHub's diff may be hard to follow. Suggested approach:

1. **Start with `build.py`** — understand the new build flow (`--qdk` builds native + runs tests, `--pip` just builds the shim).
2. **Read `source/qdk_package/qdk/__init__.py`** — see what the new `qdk` root exposes.
3. **Read `source/pip/qsharp/__init__.py`** — see the deprecation shim pattern.
4. **Skim `source/qdk_package/qdk/simulation/__init__.py`** and `source/qdk_package/qdk/_device/_atom/__init__.py` — these have the circular import fix.
5. **Check `.ado/publish.yml`** — verify the platform-agnostic vs. per-platform split.
6. **The rest is mostly mechanical** — file renames and `s/qsharp/qdk/g` import updates. GitHub should detect most as renames (95%+ similarity).

## Testing

- 1,248 unit tests pass (`source/qdk_package/tests/`)
- 338 integration tests pass per Qiskit version (`source/qdk_package/tests-integration/`), run against both Qiskit v1 (`>=1.3,<2`) and v2 (`>=2,<3`)
- Widgets build successfully
- `qsharp` shim wheel builds successfully

## `qdk` Package Structure

```
qdk_package/
├── Cargo.toml
├── pyproject.toml
├── MANIFEST.in
├── README.md
├── test_requirements.txt
│
├── src/                                # Rust source for _native
│   └── *.rs
│
├── qdk/
│   ├── __init__.py                     # Package root; exposes common utilities
│   │
│   ├── _native.pyd/.so                 # Built by maturin (module-name = "qdk._native")
│   ├── _types.py                       # Pure Python types (PauliNoise, StateDump, etc.)
│   ├── _interpreter.py                 # Interpreter lifecycle & operations
│   ├── _ipython.py                     # %%qsharp cell magic
│   ├── _http.py                        # fetch_github()
│   ├── _fs.py                          # File system callbacks
│   ├── _adaptive_pass.py
│   ├── _adaptive_bytecode.py
│   ├── telemetry.py
│   ├── telemetry_events.py
│   │
│   ├── code/
│   │   └── __init__.py                 # Dynamic Q# callables namespace
│   │
│   ├── estimator/
│   │   └── __init__.py
│   │
│   ├── openqasm/
│   │   └── __init__.py
│   │
│   ├── qiskit/                         # Lifted out of interop/
│   │   ├── __init__.py
│   │   ├── backends/__init__.py
│   │   ├── passes/__init__.py
│   │   ├── jobs/__init__.py
│   │   └── execution/__init__.py
│   │
│   ├── cirq/                           # Lifted out of interop/
│   │   └── __init__.py
│   │
│   ├── _device/
│   │   ├── __init__.py
│   │   ├── _device.py
│   │   └── _atom/
│   │       └── __init__.py             # NeutralAtomDevice
│   │
│   ├── qre/
│   │   ├── __init__.py
│   │   ├── application/__init__.py
│   │   ├── models/__init__.py
│   │   │   ├── qubits/__init__.py
│   │   │   ├── qec/__init__.py
│   │   │   └── factories/__init__.py
│   │   ├── interop/__init__.py
│   │   ├── property_keys.py
│   │   └── instruction_ids.py
│   │
│   ├── applications/
│   │   ├── __init__.py
│   │   └── magnets/
│   │       ├── __init__.py
│   │       ├── utilities/
│   │       ├── trotter/
│   │       ├── models/
│   │       └── geometry/
│   │
│   ├── qsharp.py                       # Re-exports full qsharp-like API from _types + _interpreter
│   │
│   ├── simulation/                     # Simulation facade package
│   │   ├── __init__.py                 # Public API: NeutralAtomDevice, NoiseConfig, run_qir, etc.
│   │   ├── _simulation.py              # QIR simulation implementation (internal)
│   │   ├── _noisy_simulator.py         # Private wrapper for noisy simulator types
│   │   └── _noisy_simulator.pyi        # Type stubs
│   │
│   ├── widgets.py                      # from qsharp_widgets import * (external)
│   │
│   └── azure/                          # Re-exports from azure.quantum
│       ├── __init__.py
│       ├── job.py
│       ├── qiskit.py
│       ├── cirq.py
│       ├── argument_types.py
│       └── target/
│           ├── __init__.py
│           └── rigetti.py
│
├── tests/                              # Unit tests (run with --qdk)
│   ├── conftest.py
│   ├── test_qsharp.py
│   ├── test_interpreter.py
│   ├── test_re.py
│   ├── test_qasm.py
│   ├── ... (30+ test modules)
│   ├── reexports/                      # Re-export verification tests
│   ├── qre/                            # QRE-specific tests
│   └── applications/                   # Application-specific tests
│
└── tests-integration/                  # Integration tests (run with --qdk --integration-tests)
    ├── conftest.py
    ├── utils.py
    ├── test_adaptive_ri_qir.py
    ├── test_adaptive_rif_qir.py
    ├── test_adaptive_rifla_qir.py
    ├── test_base_qir.py
    ├── devices/                        # Device integration tests
    ├── interop_qiskit/                 # Qiskit interop tests
    ├── interop_cirq/                   # Cirq interop tests
    └── resources/                      # Test resource files (QIR, etc.)
```

For a detailed breakdown of every public symbol exported by each `qdk` submodule, see [API_SURFACE.md](https://github.com/microsoft/qdk/blob/9df4562c8f5c32eb6e7f24b56d60fa586d57301f/source/qdk_package/API_SURFACE.md).

## Follow-up Work

- **Move noise types to `qdk.simulation`**: The `PauliNoise`, `DepolarizingNoise`, `BitFlipNoise`, and `PhaseFlipNoise` classes currently live in `qdk._types` and are re-exported through `qdk.qsharp`. These are simulation concepts and should canonically live in `qdk.simulation`, with backward-compatible re-exports from `qdk.qsharp` and `qdk._types`. Deferred from this PR to avoid additional circular import complexity.
- **`NoiseConfig` in `qdk.qsharp`**: Similarly, `NoiseConfig` (from `_native`) is re-exported in `qdk.qsharp.__all__` but semantically belongs in `qdk.simulation` (where it's already exported). The `qdk.qsharp` re-export should be removed in a follow-up.
- **Audit and rewrite docstrings**: Module and function docstrings throughout the package still reference the old `qsharp` import paths and naming conventions. These need to be audited and updated to reflect the new `qdk.*` namespace for accurate generated documentation.


source/qdk_package/PR_DESCRIPTION.md
```diff
@@ -0,0 +1,133 @@
+# PR: Migrate `qsharp` pip package → `qdk`
+
+> **Delete this file before merging.**
```
ScottCarda-MS: Remove this file before merging.

source/qdk_package/API_SURFACE.md
```diff
@@ -0,0 +1,345 @@
+# `qdk` Package — Public API Surface
+
+> **Delete this file before merging.**
```
ScottCarda-MS: Remove this file before merging.

botechat (dismissed):

joao-boechat (approved):

source/pip/qsharp/_device/_atom/__init__.py
```diff
@@ -1,327 +1,5 @@
...
-        end_time = time.monotonic()
-        telemetry_events.on_neutral_atom_simulate_end(
-            (end_time - start_time) * 1000, shots, using_noise, type
-        )
-        return result
-
-
-__all__ = ["NeutralAtomDevice"]
+# Deprecated: use qdk._device._atom instead.
+from qdk._device._atom import *  # noqa: F401,F403
```
orpuente-MS, line 5: Nice. Should we also add a `print("Deprecation warning: use qdk._device._atom instead")` expression here?

orpuente-MS: Same for other files.

source/pip/qsharp/code/__init__.py
```diff
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-"""
-Code module that receives any user-defined Q# callables as Python functions.
-"""
+# Deprecated: use qdk.code instead.
+from qdk.code import *  # noqa: F401,F403
```
orpuente-MS: Mmh, I'm not sure that this will re-update `qsharp.code` as we push changes to `qdk.code` dynamically. Thoughts on this @minestarks , @idavis , @swernli ? And if it does work, do you know why?

source/pip/qsharp/__init__.py
```diff
@@ -1,8 +1,36 @@
...
+
+import warnings as _warnings
+
+_warnings.warn(
+    "The 'qsharp' package is deprecated and will be removed in a future release. "
+    "Please use the 'qdk' package instead. "
+    "See https://github.com/microsoft/qdk for migration guidance.",
+    DeprecationWarning,
+    stacklevel=2,
+)
```
orpuente-MS, line 22: Ah, I see. We are issuing a single warning when entering the `qsharp` package so that we don't need to issue warning on each sub-module. Nice.

orpuente-MS: You can disregard my previous comment related to this.

source/pip/pyproject.toml
orpuente-MS, line 1: I can't properly review this file since I don't know much about the python build system. Someone else should take a look at this file.

source/pip/qsharp/_device/_atom/__init__.py
ScottCarda-MS: The deprecation warning issued as part of the top-level \_\_init\_\_.py will sufficiently warn users when they use any content from the `qsharp` package.

source/pip/qsharp/code/__init__.py
ScottCarda-MS: This works the same way as before the migration, when qdk.code would be populated from qsharp.code. It is just reversing the relationship.
I double-checked manually and it is still working fine.

source/qdk_package/pyproject.toml
```diff
@@ -10,15 +10,29 @@ readme = "README.md"
...
+    "Operating System :: POSIX :: Linux",
+]
 
 [project.optional-dependencies]
 jupyter = ["qsharp-widgets==0.0.0", "qsharp-jupyterlab==0.0.0"]
 azure = ["azure-quantum>=3.8.0"]
 qiskit = ["qiskit>=1.2.2,<3.0.0"]
 cirq = ["cirq-core>=1.6.1,<1.7", "cirq-ionq>=1.6.1,<1.7"]
 qre = ["cirq-core==1.6.1,<1.7", "pandas>=2.1", "ply>=3.11"]
-applications = ["cirq-core==1.6.1,<1.7"]
```
minestarks, line 21: Is this intentionally removed?

ScottCarda-MS: Hmm, I think I got confused on this. I had checked the source on main before and didn't find that, so I thought it was supposed to be removed, but if it is showing on this diff that it is being removed, then I guess it should stay. I'll add it back in.

source/pip/pyproject.toml
```diff
@@ -13,22 +14,22 @@ classifiers = [
...
     "Operating System :: POSIX :: Linux",
 ]
 
 [project.optional-dependencies]
-jupyterlab = ["qsharp-jupyterlab"]
-widgets = ["qsharp-widgets"]
-qiskit = ["qiskit>=1.2.2,<3.0.0"]
-cirq = ["cirq-core>=1.6.1,<1.7"]
-qre = ["cirq-core==1.6.1,<1.7", "pandas>=2.1", "ply>=3.11", "pyqir>=0.12.3,<0.13"]
+jupyterlab = ["qdk[jupyter]"]
```
minestarks, line 23: You'll want to pin the versions here, pretty sure

minestarks: Oh, actually I just saw `qdk` was pinned above under `dependencies`. I suppose that constrains it enough that we don't need versions again here?

source/pip/pyproject.toml
```diff
@@ -3,8 +3,9 @@ name = "qsharp"
 version = "0.0.0"
 readme = "README.md"
 requires-python = ">= 3.10"
+dependencies = ["qdk==0.0.0"]
```
minestarks, line 6: Does the `version.py` script know to update this?

ScottCarda-MS: Ah good catch, version.py was updated, but only to point at qdk. I've updated it to also point back at the qsharp toml.

source/qdk_package/pyproject.toml
ScottCarda-MS: done

ScottCarda-MS:
[Release](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=156925&view=results) and [multiplatform-test](https://github.com/microsoft/qdk/actions/runs/25523494575) pipelines ran successfully.

source/pip/qsharp/code/__init__.py
billti: Did you test that things you dynamically add _AFTER_ you import `qdk.code` then appear? I image this will work if you create the operations and _THEN_ import this module, but it may not update dynamically with this pattern.

source/pip/qsharp/__init__.py
```diff
@@ -1,8 +1,36 @@
...
+To migrate, replace ``import qsharp`` with ``from qdk import qsharp`` or
+``import qdk`` and use the ``qdk.*`` namespace directly.
+"""
+
+import warnings as _warnings
+
+_warnings.warn(
+    "The 'qsharp' package is deprecated and will be removed in a future release. "
+    "Please use the 'qdk' package instead. "
+    "See https://github.com/microsoft/qdk for migration guidance.",
```
billti: Just pointing to the top of the repo probably isn't that helpful. Is there a specific migration or guidance page we'll add? (e.g. the wiki maybe)

source/pip/qsharp/__init__.py
```diff
@@ -14,23 +42,25 @@
...
+    TargetProfile,
+    estimate_custom,
     CircuitGenerationMethod,
 )
 
-telemetry_events.on_import()
+from qdk import telemetry_events
 
-from ._native import Result, Pauli, QSharpError, TargetProfile, estimate_custom
+telemetry_events.on_import()
```
billti, line 58: You probably want to move this to the `qdk` package init, and change the `qsharp` package init to log a telemetry event specific to importing from `qsharp` - that way we can tell how many folks are still importing from the "old" qsharp pacakge.

source/pip/qsharp/__init__.py
```diff
@@ -14,23 +42,25 @@
...
 
 # IPython notebook specific features
 try:
```
billti: Why leave any of this try/except block here? Why not move it all to QDK only?

source/pip/qsharp/_native.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# Deprecation shim – delegates to qdk._native
+from qdk._native import *  # type: ignore
```
billti, line 4: I don't see _native.pyi being deleted from `source/pip/qsharp`, did you miss it?

ScottCarda-MS: GitHub registered that file as a rename to `source/qdk_package/qdk/_native.pyi`.

source/pip/qsharp/__init__.py
ScottCarda-MS: There are two separate events logged for `qsharp` vs `qdk`, both of which are defined under `source/qdk_package/qdk/telemetry_events.py`. `on_import()` logs a "qsharp.import" event and `on_qdk_import()` logs a "qdk.import" event. The qdk one is called from `qdk`'s init.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
```diff
@@ -4,8 +4,9 @@
 import pytest
 from expecttest import assert_expected_inline
 
-import qsharp
-from qsharp._device._atom import NeutralAtomDevice, NoiseConfig
+import qdk as qsharp
+from qdk._device._atom import NeutralAtomDevice
+from qdk.simulation import NoiseConfig
```
orpuente-MS: I was wondering why just importing `qdk as qsharp` was working, since this file uses `qsharp.compile` and there is no `qdk.compile` function.

Today we have two compile functions: `qdk.qsharp.compile` and `qdk.openqasm.compile`. I realized that with these changes we are surfacing `qdk.qsharp.compile` to `qdk.compile`. I don't think that we should do this. I think we should keep all qsharp related utilities under `qdk.qsharp.*` to keep Q# and OpenQASM on an equal level.

orpuente-MS: And then this file should be changed to do `from qdk import qsharp` to use the `qsharp` lang utilities

orpuente-MS: Same comment for other files doing `import qdk as qsharp`.

source/qdk_package/tests-integration/interop_qiskit/test_qir.py
```diff
@@ -5,7 +5,8 @@
 from typing import Optional
 
 import pytest
-from qsharp import TargetProfile, QSharpError
+from qdk import TargetProfile
+from qdk._native import QSharpError
```
orpuente-MS, line 9: Similar comment here. We should gather utilities related to Q# such as `QSharpError` under `qdk.qsharp` so that users can do `from qdk.qsharp import QSharpError`. For example, we can do `from qdk.openqasm import QasmError`. We should avoid these API asymmetries between Q# and OpenQASM.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
ScottCarda-MS: These are not truly in the qdk public API, but are just monkey-patched into the package while testing to make the migration smoother. You can see the monkey-patching happening here in the conftest.py file: https://github.com/microsoft/qdk/pull/3192/changes#diff-8e0dd69c94ea5b226df960ee997bbfbbacd5877fbbaee54c6dcbfe4a70fee26dR16

source/qdk_package/tests-integration/interop_qiskit/test_qir.py
ScottCarda-MS: Same as above. The error type is monkey-patched into the test framework.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
orpuente-MS: Do we really need the "monkey-patching"? Doesn't this work as well?

```python
import qdk
from qdk import qsharp
```

source/pip/qsharp/__init__.py
ScottCarda-MS: You are right, this is duplicate and not needed as the qdk's registration will fire any time someone imports qsharp. Removed

source/pip/qsharp/__init__.py
ScottCarda-MS: Update to point to a wiki page on migration: https://github.com/microsoft/qdk/wiki/Migrating-from-qsharp-to-qdk

source/pip/qsharp/code/__init__.py
ScottCarda-MS: Oh I see, yah, this needs more specific logic to handle the Q# content being added/modified later. I've added a small proxy class that forwards access to members to the live `code` defined in `qdk`. I've tested this out and verified this with dynamically added content as you suggested.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
ScottCarda-MS: Nice, that is much simpler. Thanks!

source/pip/qsharp/code/__init__.py
ScottCarda-MS: Actually, the code proxy seems like overkill. We don't need two separate code modules, one under qdk and another under qsharp. We can just have one identity referenced in both places, which is even simpler than the code proxy.

orpuente-MS (approved):

minestarks (approved):

ScottCarda-MS:
Successfully ran [Release](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=157122&view=results) and [Multiplatform](https://github.com/microsoft/qdk/actions/runs/25580988249) builds.

PR #3193, opened 2026-05-04
Generic QRE model for neutral atoms.

brad-lackey:
We add a model for neutral atoms parallel to the gate-based and Majorana models.

We add a model for surface with "low moves" compatible with a neutral atom architecture.

We update global properties to include VELOCITY, ACCELERATION, ATOM_SPACING.


brad-lackey:
@copilot resolve the merge conflicts in this pull request

brad-lackey:
It seems that because of the commit chain, this PR also includes the migration of the magnet models back to their original location in QDK.

PR #3195, opened 2026-05-05
[OUTDATED] Allow manual and automatic memory/compute working together

fedimser:
This PR allows to enable memory-compute architecture with LRU/LFU cache, and use manual Load/Store instructions. So that runtime automatically writes qubits to memory when not needed, but the user can force such a write to memory at any point.

TODO: this PR misses removing qubits from cache on release (I will fix that if we decide to go with this approach).


library/std/src/Std/Memory.qs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+
+/// Loads a qubit from "memory/cold" qubit to "compute/hot" qubit.
```
msoeken, line 5: I would remove all references to "hot" and "cold". Let's consistently use "compute" and "memory".

library/std/src/Std/Memory.qs
```diff
@@ -0,0 +1,17 @@
...
+/// Does nothing if qubit is already "hot".
+function MemoryQubitLoad(q : Qubit) : Unit {
```
msoeken, line 7: I think that this is already too detailed and does not match with what I had in mind at our last meeting. I suggest a single intrinsic operation `AssertComputeQubits(qubits : Qubit[])` (or something like that), with the following meaning:

- After the operation has been called, only `qubits` are in compute. Compute qubits that were present before the call and that are not in `qubits` are assumed to be moved back to memory. Qubits in `qubits` that were not compute qubits before need to be read from memory.
- Any gate that tries to access a qubit that is not explicitly in compute must fail, unless this is used in combination with dynamic memory/compute allocation
- There is no need to write back to memory, because this happens at the next call to `AssertComputeQubits`

This would require to understand how we want to use `EnableMemoryComputeArchitecture`. I suggest that if the compute capacity is a positive number, we assume dynamic memory compute (even though calls to `AssertComputeQubits` can be respected and may fail on insufficient capacity) and if we call it with `0` or a negative number (or any other kind of signature if that's better), there is no dynamic movement, but all compute qubits need to be pre-asserted with `AssertComputeQubits`.

PR #3197, opened 2026-05-05
Fix azure error logging

joao-boechat:
This PR improves the error logging for job submission to azure.

### DIFF
--- a/source/vscode/src/azure/networkRequests.ts
+++ b/source/vscode/src/azure/networkRequests.ts
@@ -149,14 +149,16 @@ async function getAzureQuantumError(response: Response): Promise<AzureError> {
 
   try {
     const json = await response.json();
-    if (json.errorCode && json.message) {
-      detail = `${json.errorCode}: ${json.message}`;
+    // Extract the error data if it conforms to the Azure Quantum error schema defined in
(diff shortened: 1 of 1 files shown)


source/vscode/src/azure/networkRequests.ts
```diff
@@ -144,24 +144,23 @@ class AzureError extends Error {
...
 
-  let message;
-  if (error) {
-    message = `Azure Quantum request failed with status ${response.status}.\n${error.code}: ${error.message}`;
-  } else {
-    message = `Azure Quantum request failed with status ${response.status}.`;
-  }
+  let message = `Azure Quantum request failed with status ${response.status} (${response.statusText}).`;
+  if (detail) message += `\n${detail}`;
+  if (requestId) message += `\nRequest ID: ${requestId}`;
```
billti, line 163: Is this useful for troubleshooting? (i.e. we'd expect the user to provide this if reporting a problem)

billti (approved):

source/vscode/src/azure/networkRequests.ts
joao-boechat: Yep! They asked us to include the requestId in error message for that scenario

PR #3198, opened 2026-05-06
QREv3  neutral atom models

brad-lackey:
NeutralAtom class: architecture for a generic neutral atom platform.

SurfaceCodeLowMove class: a variant of the surface code that does syndome extraction with local moves.

Added properties: ACCELERATION, VELOCITY, ATOM_SPACING to support neutral atom platforms.


source/pip/qsharp/qre/models/qec/_surface_code_low_move.py
```diff
@@ -0,0 +1,198 @@
...
+    VELOCITY,
+    ACCELERATION,
+    ATOM_SPACING,
+)
+
+
+@dataclass
+class SurfaceCodeLowMove(ISATransform):
+    """
+    This class models the gate-based rotated surface code.
```
msoeken: Could we add details about the "low-move" nature of this transform. Further, is there a reference that discusses this which we could add to the docstring?

source/pip/qsharp/qre/models/qec/_surface_code_low_move.py
```diff
@@ -0,0 +1,198 @@
...
+            meas_z.expect_error_rate(),
+        )
+
+        # There are d^2 data qubits and (d^2 - 1) ancilla qubits in the rotated
+        # surface code.  (See Section 7.1 in arXiv:1111.4022)
+        # Unchanged from the original SurfaceCode.
+        space_formula = linear_function(2 * self.distance**2 - 1)
+
+        # Each standard syndrome extraction cycle consists of ancilla preparation, 4
+        # rounds of CNOTs, and measurement.  (See Fig. 2 in arXiv:1009.3686).
```
msoeken, line 169: Refers to CNOT in the comment, but gate that is used is CZ.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
```diff
@@ -0,0 +1,109 @@
...
+    H,
+    CNOT,
+    T,
+)
+from ...property_keys import ACCELERATION, ATOM_SPACING, VELOCITY
+
+
+@dataclass
+class NeutralAtom(Architecture):
+    """A movement-aware neutral-atom architecture."""
```
msoeken: We should add doc strings about the various fields that can be configured, their units, and their default value.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
```diff
@@ -0,0 +1,109 @@
...
+    rydberg_time: int = field(default=500)  # In units of ns.
+    rydberg_error: float = field(default=1e-3)
+    one_qubit_time: int = field(default=1000)  # In units of ns.
+    one_qubit_error: float = field(default=1e-4)
+    measurement_time: int = field(default=10000)  # In units of ns.
+    measurement_error: float = field(default=1e-4)
+    handoff_time: int = field(default=0)  # In units of ns.
+    atom_spacing: int = field(default=3)  # In units of microns.
+    max_velocity: int = field(default=1)  # In units m/s.
+    max_acceleration: int = field(default=5000)  # In units m/s^2.
```
msoeken: Are these the most natural units to choose? Why not meters or nano-meters for the atom spacing? We could also use a float value here and then convert it to bits using the helper function.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
```diff
@@ -0,0 +1,109 @@
...
+                arity=1,
+                time=0,
+                error_rate=0.0,
+            ),
+            ctx.add_instruction(
+                T,
+                encoding=Encoding.PHYSICAL,
+                arity=1,
+                time=0,
+                error_rate=0.00001,
```
msoeken, line 121: Should this be configurable? It should also have a comment to explain why we currently do not set the error rate to 0.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
```diff
@@ -0,0 +1,109 @@
...
+                arity=1,
+                time=self.measurement_time,
+                error_rate=self.measurement_error,
+            ),
+            ctx.add_instruction(
+                PHYSICAL_MOVE,
+                encoding=Encoding.PHYSICAL,
+                arity=1,
+                time=2 * self.handoff_time,
+                error_rate=1e-4,
```
msoeken, line 170: Should this be configurable? Is this value ever used or are we only using the move instruction to provide timing information?

msoeken (changes requested):

source/pip/qsharp/qre/models/qec/_surface_code_low_move.py
brad-lackey: Done

source/pip/qsharp/qre/models/qec/_surface_code_low_move.py
brad-lackey: This is addressed in the rest of the comment.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
brad-lackey: Done.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
brad-lackey: We want to use what the literature uses, and pretty much everywhere atom spacing is in microns.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
brad-lackey: It is, via the handoff_time keyword. The details of the move are handled at a higher level of the stack, so the only thing I can think of at the platform level for the PHYSICAL_MOVE is the handoff time.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
brad-lackey: This is a hack so that we can export the T gate without having to build a ISATransform wrapper around NeutralAtom. Once the magic state factories can handle T-gate error rates of 0, we can revert this back to zero.

source/pip/qsharp/qre/models/qubits/_neutral_atoms.py
msoeken: I was referring to the hard coded error rate of 1e-4

msoeken:
Closed in favor of #3211 (after merge of #3192)

PR #3199, opened 2026-05-06
Added 8T->CCX and cultivation models

brad-lackey:
Added the two factory classes associated to 8T->CCX factories and basic magic state cultivation.


msoeken:
Re-created in #3212 to handle merge conflicts from #3192.

PR #3204, opened 2026-05-08
Manual Memory-Compute qubits

fedimser:
This PR introduces a new way to manually manage Memory qubits. It adds two new operations `Std.Memory.MemoryQubitLoad` and `Std.Memory.MemoryQubitStore` that operate on a single qubit and instruct runtime to "load" qubit (move from memory to compute) or "save" it (move from compute to memory).

Example:
```
Std.ResourceEstimation.EnableMemoryComputeArchitecture(0, 2);
use q = Qubit();
X(q);
Std.Memory.Store(q);
Std.Memory.Load(q);
```

Conventions:
* Q#'s `Qubit` is the "quantum value" rather than a location on a physical device. When it is moved between locations (e.g. from "hot" to "cold" area of quantum computer), in Q# it's still the same Q# object. This is why Load and Store act on single qubit and mutate its "type" (compute/memory) rather than action between 2 qubits. At some point Load and Store is translated to 2-qubit operation between memory and compute qubit, but this is hidden from the programmer. 
* All qubits become "compute" qubits immediately as allocated.
* Applying gate/measurement to memory qubit is an error.
  * `swap_id` can be performet only between 2 compute qubits.
  * `Reset` on memory qubit is allowed.

Notes on implementing these operations in backends:
* These operations currently have effect only in resource estimation, when `Std.ResourceEstimation.EnableMemoryComputeArchitecture` was called with `strategy=2` (which corresponds to manual strategy). They are no-ops in any other backend.
* This allows to have exactly the same algorithm to be resource estimated with and without memory-compute architecture.
* In future we plan to implement these in code generator, by maintaining 2 pools for compute and memory qubits and synthesizing 2-qubit instructions for read/write operations between memory and compute qubit.
* This is why these 2 operations are in `Std.Memory`, not `Std.ResourceEstimation`. They describe operations that make sense outside resource estimation, even though currently they are only implemented in resource estimation.
* There will be no need to support these in simulators, they will remain no-op.

Notes on interaction with existing "automatic" memory-compute architecture:
* These features are very similar, so they either need to be merged or be mutually exclusive. I decided to make them mutually exclusive: you either use automatic memory-compute (using strategy=0 or 1) or manual (strategy=2).
* Code for `MemoryComputeInfo` implementing automatic memory-compute is untouched, I just wrapped it into enum `MemoryCompute`. By being enum, it forces mutual exlcusivity of memory and compute architecture.
* There are differences between automatic and manual memory-compute:
  * In Manual mode, qubits become "compute" immediately as allocated. In Auto mode, between allocation and first usage qubits are neither compute nor memory, and they become compute only on first usage.
  * When trying to apply computation (gate/measurement) on memory qubit: in Auto mode, it will load the qubit form memory. In "Manual" mode it will result in error. This is added to force users to explicitly add Load instruction. So that if resource estimation succeeds, the user can be sure that they don't accidentally apply computation on memory qubits where they didn't intend to. This can be easily changed to do auto-load, but it is my intention that in manual mode all loads and stores must be explicit.
  * In Auto mode, all inserts into cache are counted as reads, even if they correspond to freshly allocated qubit.
  * Total number of qubits is computed differently.

This PR is equivalent to https://github.com/microsoft/qdk/pull/3159 in a sense that it allows to model algorithms with Memory and Compute qubits with manually moving qubits between Memory and Compute, and it allows to get exactly the same resource estimates. Unlike that PR, this one doesn't require any changes to the Q# language.


source/resource_estimator/src/counts/memory_compute.rs
```diff
@@ -337,3 +338,73 @@ impl<K: Eq + Hash + Clone> LeastFrequentlyUsedPriorityQueue<K> {
...
+        self.memory_qubits.remove(&qid);
+    }
+
+    pub fn assert_compute_qubits(
+        &self,
+        qubits: impl IntoIterator<Item = usize>,
+    ) -> Result<(), String> {
+        for qid in qubits {
+            if self.memory_qubits.contains(&qid) {
+                return Result::Err("Tried to do computation on memory qubit".to_string());
```
swernli: the guidance for user visible errors is to start them with lower case letters, as they are often appended to other error strings:

```suggestion
                return Result::Err("cannot perform computation on memory qubit".to_string());
```

source/resource_estimator/src/counts/memory_compute.rs
```diff
@@ -337,3 +338,73 @@ impl<K: Eq + Hash + Clone> LeastFrequentlyUsedPriorityQueue<K> {
...
+    /// No memory-compute architecture, all qubits are "compute" qubits.
+    /// Load/Store instructions are ignored.
+    None,
+    /// Automatically manages memory and compute qubits by evicting compute qubits to
+    /// memory if needed.
+    /// Load/Store instructions are ignored.
+    /// Gates/measurements on memory qubit will be automatically prepended by load.
+    Auto(MemoryComputeInfo),
+    /// Qubits are loaded and stored by explicit Load/Store instructions.
+    /// Gates/measurements on memory qubit resul tin error.
```
swernli: typo:

```suggestion
    /// Gates/measurements on memory qubit result in error.
```

source/resource_estimator/src/counts/memory_compute.rs
```diff
@@ -337,3 +338,73 @@ impl<K: Eq + Hash + Clone> LeastFrequentlyUsedPriorityQueue<K> {
...
+        self.max_memory_qubits_count = max(self.max_memory_qubits_count, self.memory_qubits.len());
+    }
+
+    /// Releases the qubit.
+    pub fn release(&mut self, qid: usize) {
+        self.compute_qubits.remove(&qid);
+        self.memory_qubits.remove(&qid);
+    }
+
+    pub fn assert_compute_qubits(
```
swernli, line 444: I'm glad this gives us some protection during resource extimation to fail if the user operates on a qubit that is not hot, but I wonder if the same protection would be helpful when doing normal simulation too. As it is, if their program can simulate, a user could author the program while repeatedly simulating without issue, only to have if fail the check during resource estimation.

That said, the change to track and check memory qubits during simulation is out of scope of this more targeted PR, so I don't think this needs to be addressed now and we can get more feedback before deciding if we want to make that change later.

library/std/src/Std/Memory.qs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Loads a qubit from memory, turning it from "memory"/"cold" to "compute"/"hot".
+/// Does nothing if qubit is already "compute".
+/// Currently only takes effect for resource estimation with memory-compute architecture
+/// enabled in Manual mode.
+function MemoryQubitLoad(q : Qubit) : Unit {
+    body intrinsic;
+}
+
+/// Stores a qubit from memory, turning it from "compute"/"hot" "memory"/"cold".
```
swernli:
```suggestion
/// Stores a qubit into memory, turning it from "compute"/"hot" to "memory"/"cold".
```

source/resource_estimator/src/counts.rs
```diff
@@ -718,6 +776,20 @@ impl Backend for LogicalCounter {
                 self.enable_memory_compute(compute_capacity, strategy);
                 Some(Ok(Value::unit()))
             }
+            "MemoryQubitLoad" => {
+                let qid = arg.unwrap_qubit().deref().0;
```
swernli: not really critical, but this could be moved inside of the if-expression (same for store used below).

library/std/src/Std/Memory.qs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+
+/// Loads a qubit from memory, turning it from "memory"/"cold" to "compute"/"hot".
+/// Does nothing if qubit is already "compute".
+/// Currently only takes effect for resource estimation with memory-compute architecture
+/// enabled in Manual mode.
+function MemoryQubitLoad(q : Qubit) : Unit {
```
swernli: I believe technically these should be operations. A `function` is meant to be pure, so it should have no side-effects. Since loading and storing can affect not just resource counts but cause later gate applications to fail, these are operations.

library/std/src/Std/Memory.qs
```diff
@@ -0,0 +1,21 @@
...
+    body intrinsic;
+}
+
+/// Stores a qubit from memory, turning it from "compute"/"hot" "memory"/"cold".
+/// Does nothing if qubit is already "memory".
+/// Currently only takes effect for resource estimation with memory-compute architecture
+/// enabled in Manual mode.
+function MemoryQubitStore(q : Qubit) : Unit {
```
swernli: I'm not sure the `MemoryQubit` prefix is needed here. While `Load` and `Store` are generic names, they do have the benefit of the signature indicating they only take qubits and the namespace indicating their usage for memory. In addition, a user who wants a more explicit name can always choose to use the fully-qualified name (`Std.Memory.Load(q)`) or import just the memory namespace and use that as the qualifier(`import Std.Memory; Memory.Load(q)`).

I don't have strong feelings on this point, so it's fine of others strongly prefer the more verbose name, but seeing the fully-qualified name underscores the apparent repetition (`Std.Memory.MemoryQubitLoad`)

library/std/src/Std/ResourceEstimation.qs
```diff
@@ -185,14 +185,21 @@ operation RepeatEstimates(count : Int) : Unit is Adj {
...
 /// # Input
 /// ## computeCapacity
 /// The maximum number of compute qubits which can be used to perform
 /// operations.
 /// ## strategy
 /// The strategy applied when evicting qubits from the compute qubits in case
 /// of maximum capacity: 0 = LRU (least recently used), 1 = LFU (least
-/// frequently used)
+/// frequently used), 2 = Manual.
 function EnableMemoryComputeArchitecture(computeCapacity : Int, strategy : Int) : Unit {
```
swernli: I know it's not from your changes, but by the same logic mentioned on the new code this should also be an operation, not a function. Might as well fix that now too.

swernli (commented):
Resource estimation updates and general infrastructure changes look good. I left a few comments on some things that should be updated and one thought for the future, but I think this is almost ready to go!

source/resource_estimator/src/counts/memory_compute.rs
swernli: Mostly left this comment for consideration, not something that needs to be addressed now, so resolving this thread.

library/std/src/Std/ResourceEstimation.qs
YingrongChen: Would it be possible to expose this to the Python layer, for example through the QRE inputs?

One use case I imagine is that we build up a Q# file with Store and Load operations, and then run multiple resource estimation from Python while toggling memory-compute layout on or off and varying the compute capacity. We can also error out if the load size exceeds the compute capacity.

It would be painful if we needed to add another input parameter to every Q# circuit that might potentially use the memory-compute layout, and then modify the circuit parameters every time we want to resource-estimate the same circuit under different compute capacities.

library/std/src/Std/Memory.qs
fedimser: Changed to Load/Store.

library/std/src/Std/ResourceEstimation.qs
fedimser: This would be tricky. This information needs to be known during logical resource estimation, so it needs to be an input to `logical_counts`. [Interpreter::logical_counts](https://github.com/microsoft/qdk/blob/198c23bc8f7de88e29e7748a783c65716e03373a/source/qdk_package/src/interpreter.rs#L967) currently only takes callable and arguments, it doesn't take any "resource estimation parameters".

It's debatable whether using memory-compute architecture is purely resource estimation feature (and therefore using memory-compute should be considered QRE setting). It is now, but in the future when we support error-correcting codes, enabling memory-compute will results in different generated QIR.

I understand that it's inconvenient to declare extra boolean argument in Q# operation. But if this operation expresses two different quantum algorithms based on that argument, extra parameter is the best way to represent that.

fedimser: I know this is not perfect solution, but in my Shor's implementation I added a boolean field "use_memory_compute" to ShorsConfig, so at least we don't have to add a new parameter in function signature. It looks like this:

```
operation PeriodFindingIteration(
    modexp_j_params : ModExpParams,
    modexp_k_params : ModExpParams,
    config : ShorsConfig,
    num_output_qubits : Int
) : (BigInt, BigInt) {
    if (config.use_memory_compute) {
        Std.ResourceEstimation.EnableMemoryComputeArchitecture(0, 2);
    }
    ...
}
```

source/resource_estimator/src/counts/memory_compute.rs
fedimser: Acknowledged. 

Implementing it in simulators is straightforward: add a map[int, QubitType], update it on each Load/Store, and check in all compute/measurements that qubit type is Compute. I didn't do it because the intention is to have minimal possible change needed to support memory qubits in resource estimation.

swernli (approved):

swernli (approved):

PR #3214, opened 2026-05-11
Fix License Info Bleeding into version

ScottCarda-MS:
The package-index we publish to doesn't seem to support the newer [PEP 639](https://peps.python.org/pep-0639/) which introduces the `License-Expression` metadata field. This causes the build to emit the license expression into the version, which is not a valid version.

To avoid this, this PR removes the license expressions and just goes back to the simpler method of listing the license in the classifiers for the packages.


minestarks (approved):

source/qdk_package/LICENSE.txt
fedimser, line 1: I think we are supposed to include the actual license text into the distributed package. Can we keep this file?

ScottCarda-MS: Hmm, we didn't _used_ to have to have it in this location of the repo. We do have a copy of the license text in our repo at the root and in the vscode folder.

ScottCarda-MS: It looks like the requirement is just to have a copy in the repo root, which we have. Thanks for checking on this though.

PR #3216, opened 2026-05-11
Don't include tests or benchmark in `qdk` wheel

ScottCarda-MS:


ScottCarda-MS:
Not needed after the migration.

PR #3219, opened 2026-05-11
Add array support to QIR bytecode

orpuente-MS:
This PR adds array support to the QIR bytecode consumed by the QIR simulators, closing the current gap in functionality for the RIFLA profile.

Remaining work:
- I am scoping the QIR that is guaranteed to work correctly to QIR generated by the QDK compilers. Especially when it comes to the `getelementptr` instruction, since I need to do more nuanced changes to `pyqir` to enable more general LLVM-IR support.


source/qdk_package/qdk/_adaptive_pass.py
```diff
@@ -547,6 +568,85 @@ def _assign_function(self, func: pyqir.Function) -> None:
...
+        values into ``constant_data``.
+        """
+        mask = (1 << self._int_bits) - 1
+
+        for gv in mod.global_variables:
+            init = gv.initializer
+            if init is None:
+                continue
+            if not isinstance(init, pyqir.ArrayConstant):
+                continue
```
billti: Is this the right behavior here? Seems it has an initializer, but it's not of the type expected... shouldn't it error? Why is just silently continuing the right behavior?

source/qdk_package/qdk/_adaptive_pass.py
```diff
@@ -547,6 +568,85 @@ def _assign_function(self, func: pyqir.Function) -> None:
...
+
+            # Skip byte-string globals (e.g. [N x i8] labels).
+            arr_ty = init.type
+            if (
+                isinstance(arr_ty, pyqir.ArrayType)
+                and isinstance(arr_ty.element, pyqir.IntType)
+                and arr_ty.element.width == 8
```
billti: Is this because we assume these are output recording labels? Can you add comments here as to why these are being ignored silently also.

source/qdk_package/qdk/_adaptive_pass.py
```diff
@@ -547,6 +568,85 @@ def _assign_function(self, func: pyqir.Function) -> None:
...
+            ):
+                continue
+
+            base_addr = len(self.constant_data)
```
billti: I find this line confusing. The length of the constant data is the address? So if two constant arrays have the same length they will have the same base address? If it's not as it reads, please add comments to explain.

billti: Oh I see (I think) - the call below to `_encode_array_elements` moves the value of `self.constant_data` as it iterates. Maybe a note here on that to save having to read the rest of the file to understand what his is doing. Thanks.

source/qdk_package/qdk/_adaptive_pass.py
orpuente-MS: This method is concerned with collecting the data from global arrays. Other globals, such as global constant ints or global constant floats can also have initializers, so skip over those.

Added a comment explaining this, and changed the name of the method to `_scan_global_arrays`

source/qdk_package/qdk/_adaptive_pass.py
orpuente-MS: Yes, I'm scoping this to accept QDK generated QIR for now. Added a comment here.

source/qdk_package/qdk/_adaptive_pass.py
orpuente-MS: `_encode_array_elements` appends the elements of the array at the end of `self._constant_data`; so the address of that array's data in `self.constant_data` is `len(constant_data)` before we append the elements.

Added a comment here.

source/qdk_package/tests/test_adaptive_cpu_bytecode.py
```diff
@@ -14,13 +14,23 @@
...
 from collections import Counter
+import pyqir
 import pytest
 from qdk.simulation import run_qir, NoiseConfig
 from qdk.simulation._simulation import Result
 import qdk
 import qdk.openqasm
 from typing import Literal
 
+_HAS_GLOBAL_VARIABLES = hasattr(
```
billti: If you are requiring pyqir 0.12.5 or later now, shouldn't this always be available and just fail if it can't run the tests?

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
```diff
@@ -1313,6 +1335,57 @@ fn interpret_classical(@builtin(global_invocation_id) gid: vec3<u32>) {
...
+
+            // STORE: Write a value to memory at the given address.
+            // Encoding: src0 = value to store, src1 = memory address.
+            case OP_STORE {
+                let val = resolve_u32(shot_idx, instr.src0, flags, 0u);
+                let addr = resolve_u32(shot_idx, instr.src1, flags, 1u);
+                shots[shot_idx].interp.memory[addr] = val;
+                pc++;
+            }
+
```
idavis, line 1396: Should we add the same kind of bounds guard for `OP_LOAD` and `OP_STORE` that `OP_ALLOCA` has? Right now a dynamic GEP or pointer value outside `MAX_MEMORY` will index interp.memory directly, which can turn invalid array access into undefined shader behavior rather than a deterministic shot error. A shared memory-bounds error path, either reusing ERR_ALLOCA_OUT_OF_BOUNDS or introducing a clearer ERR_MEMORY_OUT_OF_BOUNDS, would make the new memory ops much safer and easier to diagnose.

source/simulators/src/gpu_full_state_simulator/gpu_context.rs
```diff
@@ -479,7 +481,10 @@ impl GpuContext {
...
         // WGSL requires struct size to be a multiple of the struct's alignment (8 bytes due to vec2f).
         let gpu_shot_size = if self.is_adaptive {
-            let raw = SIZEOF_SHOTDATA + size_of::<InterpreterState>() + params.num_registers * 4;
+            let raw = SIZEOF_SHOTDATA
+                + size_of::<InterpreterState>()
+                + params.num_registers * 4
+                + params.max_memory * 4;
             (raw + 7) & !7 // round up to 8-byte alignment
         } else {
             SIZEOF_SHOTDATA
```
idavis: Can we include the full adaptive shot size when computing the shots-per-batch limit? The new per-shot memory term is added to gpu_shot_size later, but max_shots_in_buffer is still capped using MAX_SHOT_ENTRIES, which is derived from SIZEOF_SHOTDATA only. For programs with larger constant_data, this can select too many shots for a batch and then request a shots buffer beyond the WebGPU limit. Computing gpu_shot_size first and capping by MAX_BUFFER_SIZE / gpu_shot_size would make the batching limit reflect the actual adaptive layout.

source/qdk_package/qdk/_adaptive_pass.py
```diff
@@ -547,6 +576,113 @@ def _assign_function(self, func: pyqir.Function) -> None:
...
+                self._encode_array_elements(elem, gv_name)
+            elif isinstance(elem, pyqir.GlobalVariable):
+                # Pointer to another global (e.g. @row0 in [N x ptr]).
+                if elem.name in self._global_to_address:
+                    self.constant_data.append(self._global_to_address[elem.name] & mask)
+                else:
+                    raise ValueError(
+                        f"Global @{gv_name} references @{elem.name} "
+                        f"which has not been scanned yet"
+                    )
```
idavis, line 680: Can we make pointer-valued global array encoding independent of global declaration order? This path currently requires referenced globals to already be present in _global_to_address, but I believe you can still have a valid module where @matrix references @row0/@row1 declared later and will fail with “has not been scanned yet”. A two-pass scan, first assigning addresses to all supported global arrays and then encoding elements, would preserve the flattened layout without depending on mod.global_variables iteration order. A regression test with a forward global reference would be useful too.

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
orpuente-MS: Sounds reasonable. The motivation behind adding out-of-bounds checks to alloca and not the others was:

1. If the QIR is ill-formed, it is natural that it will misbehave.
2. However, if it is well formed, but it hits the GPU mem-limits, it's the implementation's fault and we should handle it gracefully.
3. I didn't want to incur runtime-checks for well-formed QIR.

Idk how much of a perf tradeoff would be to incur two extra checks in the hot path of the simulation. But it would definitely be useful to get nice errors on malformed QIR.

orpuente-MS: Added the checks.

source/qdk_package/qdk/_adaptive_pass.py
orpuente-MS: Done. And added the regression test.

source/simulators/src/gpu_full_state_simulator/gpu_context.rs
orpuente-MS: Yes, nice catch. Done.

source/qdk_package/tests/test_adaptive_gpu_bytecode.py
```diff
@@ -2138,7 +2138,7 @@ def test_call_stack_overflow_guard():
 
 def _build_alloca_oob_qir():
     """Build QIR with 258 scalar allocas to exceed MAX_MEMORY on the GPU."""
```
idavis: Comment doesn't match new size

orpuente-MS: Fixed

idavis (approved):

PR #3220, opened 2026-05-12
Fix memory leaks in Context

fedimser:
* Use weakref when passing `make_callable` and `make_class` to Interpreter, which breaks circular reference (Context -> Interpreter -> callback -> Context) and allows the Context object (and Interpreter) to be garbage collected.
* Do not register callables and classes for non-global Contexts in `sys.modules`. Use `types.SimpleNamespace` instead. Users access symbols for global Context via `qdk.code`, and symbols for other Contexts via `context_variable.code`. 
  * This way we don't need to clear `sys.modules` when Context is garbage collected. 
  * We still clear `qdk.code` when global context is invalidated on `init()`.

### DIFF
--- a/source/qdk_package/qdk/_context.py
+++ b/source/qdk_package/qdk/_context.py
@@ -27,7 +27,7 @@ from typing import (
     cast,
 )
 
-from . import telemetry_events
+from . import telemetry_events, code
 from ._native import (  # type: ignore
     Circuit,
     CircuitConfig,
@@ -151,9 +151,9 @@ class Context:
 
     _interpreter: Interpreter
     _config: Config
-    code: types.ModuleType
-    _code_prefix: str
+    code: Any
     _disposed: bool
+    _is_global_context: bool
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


source/qdk_package/qdk/_context.py
```diff
@@ -226,6 +227,26 @@ def __init__(
...
+                return
+            Context._make_callable(ctx, callable, namespace, callable_name)
+
+        def make_class_weak(
+            qsharp_type: TypeIR, namespace: List[str], class_name: str
+        ) -> None:
+            ctx = self_ref()
+            if ctx is None or ctx._disposed:
+                return
+            Context._make_class(ctx, qsharp_type, namespace, class_name)
```
idavis:
```suggestion
            ctx._make_callable(callable, namespace, callable_name)

        def make_class_weak(
            qsharp_type: TypeIR, namespace: List[str], class_name: str
        ) -> None:
            ctx = self_ref()
            if ctx is None or ctx._disposed:
                return
            ctx._make_class(qsharp_type, namespace, class_name)
```

swernli (approved):

PR #3222, opened 2026-05-13
Custom Job Params for Targets

ScottCarda-MS:
### Summary

Adds a hidden VS Code setting (`Q#.azure.targetJobParams`) that allows users to specify arbitrary additional `inputParams` to include when submitting jobs to a specific Azure Quantum target. This enables power users to pass provider-specific parameters (e.g. emulation settings) directly from the VS Code extension, matching what's already possible via the `azure-quantum` Python SDK's `target.submit(input_params={...})`.

### Motivation

Some users need to pass target-specific parameters beyond `shots` when submitting jobs. Previously, this was only possible via the Python SDK. This change brings that capability to the VS Code extension.

### Usage

Add the following to user or workspace `settings.json`:

```jsonc
"Q#.azure.targetJobParams": {
    "rigetti.sim.qvm": {
        "param1": true,
        "param2": {
            "param2-1": "clifford",
            "Foo": false,
            "param2-3": {
                "param2-3-1": "Deeply Nested",
                "param2-3-2": 1.6
            }
        }
    }
}
```

Keys are target IDs (matching the target tree item), values are objects merged into the job's `inputParams` payload.

The setting is tagged as `hidden` and will not appear in the VS Code Settings UI.

### Changes

- **`package.json`** — Added hidden `Q#.azure.targetJobParams` configuration setting
- **`config.ts`** — Added `getTargetJobParams(targetId)` helper that reads the setting and deep-clones it to a plain object (necessary because VS Code's configuration API returns proxy objects whose nested values don't survive object spread)
- **`workspaceActions.ts`** — Added `additionalJobParams` parameter to `submitJob()` and `putJobData()`, spread into `inputParams`
- **`commands.ts`** — Reads target job params from config and passes them to `submitJob()`

### Testing

Manually verified with:

- Flat parameters (e.g., `"noOpt": true`)
- Nested parameters (e.g., `"emulationSettings": { ... }`)
- No parameters configured (empty object — no effect on existing behavior)
- Compared submitted job `inputParams` against equivalent jobs submitted via the Python SDK's `target.submit(input_params={...})`

### Notes

The deep clone via `JSON.parse(JSON.stringify(...))` in `getTargetJobParams()` is intentional. VS Code's `getConfiguration().get()` returns proxy objects with lazy property resolution. Without cloning at the root level, nested values silently disappear when the object is spread into the `inputParams` payload.

### DIFF
--- a/source/vscode/src/config.ts
+++ b/source/vscode/src/config.ts
@@ -50,3 +50,12 @@ export function getUploadSupplementalData(): boolean {
     .getConfiguration("Q#")
     .get<boolean>("azure.uploadSupplementalData", true);
 }
+
+export function getTargetJobParams(
+  targetId: string,
+): Record<string, unknown> {
+  const allTargetParams = vscode.workspace
+    .getConfiguration("Q#")
+    .get<Record<string, Record<string, unknown>>>("azure.targetJobParams", {});
+  return allTargetParams[targetId] ?? {};
+}
--- a/source/vscode/src/config.ts
+++ b/source/vscode/src/config.ts
@@ -52,8 +52,14 @@ export function getUploadSupplementalData(): boolean {
 }
 
 export function getTargetJobParams(targetId: string): Record<string, unknown> {
-  const allTargetParams = vscode.workspace
+  const raw = vscode.workspace
     .getConfiguration("Q#")
     .get<Record<string, Record<string, unknown>>>("azure.targetJobParams", {});
+  // Deep clone the entire setting to materialize VS Code's configuration proxy
+  // into a plain object. Without this, nested values may not survive object
+  // spread operations.
+  const allTargetParams: Record<string, Record<string, unknown>> = JSON.parse(
+    JSON.stringify(raw),
+  );
   return allTargetParams[targetId] ?? {};
 }
--- a/source/vscode/src/config.ts
+++ b/source/vscode/src/config.ts
@@ -51,9 +51,7 @@ export function getUploadSupplementalData(): boolean {
     .get<boolean>("azure.uploadSupplementalData", true);
 }
 
-export function getTargetJobParams(
-  targetId: string,
-): Record<string, unknown> {
+export function getTargetJobParams(targetId: string): Record<string, unknown> {
   const allTargetParams = vscode.workspace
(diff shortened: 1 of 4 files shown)


fedimser (approved):

source/vscode/src/config.ts
```diff
@@ -50,3 +50,16 @@ export function getUploadSupplementalData(): boolean {
...
+export function getTargetJobParams(targetId: string): Record<string, unknown> {
+  const raw = vscode.workspace
+    .getConfiguration("Q#")
+    .get<Record<string, Record<string, unknown>>>("azure.targetJobParams", {});
+  // Deep clone the entire setting to materialize VS Code's configuration proxy
+  // into a plain object. Without this, nested values may not survive object
+  // spread operations.
+  const allTargetParams: Record<string, Record<string, unknown>> = JSON.parse(
+    JSON.stringify(raw),
+  );
```
fedimser, line 63: I asked AI if there is a better way to do a deep copy in typescript, and it told me there is structuredClone - have you tried that?

source/vscode/src/config.ts
```diff
@@ -50,3 +50,16 @@ export function getUploadSupplementalData(): boolean {
...
+
+export function getTargetJobParams(targetId: string): Record<string, unknown> {
+  const raw = vscode.workspace
+    .getConfiguration("Q#")
+    .get<Record<string, Record<string, unknown>>>("azure.targetJobParams", {});
+  // Deep clone the entire setting to materialize VS Code's configuration proxy
+  // into a plain object. Without this, nested values may not survive object
+  // spread operations.
+  const allTargetParams: Record<string, Record<string, unknown>> = JSON.parse(
+    JSON.stringify(raw),
```
billti, line 62: This is an old-ish way of copying objects. Is structureClone not available or suitable for some reason? <https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone>

billti (approved):

source/vscode/src/config.ts
billti: Resolving to merge it. Can make this change later, but does the same thing so not important.

PR #3223, opened 2026-05-13
Restore function get_config

fedimser:
* _interpreter.get_config() is added because it's already used in one place outside this repository (and was removed in recent PR). We don't plan to make it supported public API, this is why it's not documented and not tested.
* We also want to "officially" allow users access some parts of config. As we don't want to make class `Config` public, those  things will be accessed through getters on `Context`. Cirrently the only part of config we want to publicly expose is the target profile.

### DIFF
--- a/source/qdk_package/qdk/_interpreter.py
+++ b/source/qdk_package/qdk/_interpreter.py
@@ -169,6 +169,10 @@ def get_interpreter() -> Interpreter:
     return _get_default_context()._interpreter
 
 
+def get_config() -> Config:
+    return _get_default_context()._config
+
+
 def python_args_to_interpreter_args(args):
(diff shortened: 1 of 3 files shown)


source/qdk_package/qdk/_interpreter.py
```diff
@@ -169,6 +169,10 @@ def get_interpreter() -> Interpreter:
     return _get_default_context()._interpreter
 
 
+def get_config() -> Config:
```
billti, line 172: Should this not be an API on `Context` too?

fedimser: As discussed in team chat, we decided to expose only target_profile, which I did by storing it in Context.

billti (approved):

PR #3224, opened 2026-05-15
Support --editable flag

billti:
I regularly build and test locally using `./build.py`, however then it installs the QDK wheels which I then `pip uninstall` and install editable again to work with.

With this change, you can just run `./build.py --editable` (or list your specific packages to build) and it will use editable installs in the current environment by default.

I tested with `./build.py --editable --integration-tests` also and they all worked fine.

Without the flag, there should be no impact, so CI, ADO, etc. should all be unaffected.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,7 +6,7 @@ __pycache__/
 .vscode/
 .vscode-test-web/
 .vscode-test/
-.venv/
+.venv*/
 /katas/**/*.md
 /library/
 /source/allocator/
--- a/build.py
+++ b/build.py
@@ -181,17 +181,53 @@ QISKIT_VERSION_MATRIX = [
 ]
 
 
-def install_build_package(cwd, interpreter):
-    # Ensure that the 'build' package is installed in the given interpreter
-    command_args = [
-        interpreter,
-        "-m",
-        "pip",
-        "install",
-        "build",
-        "-v",
-    ]
-    subprocess.run(command_args, check=True, text=True, cwd=cwd)
+def run(cmd, cwd, env=None):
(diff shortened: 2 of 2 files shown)


.prettierignore
```diff
@@ -6,7 +6,7 @@ __pycache__/
 .vscode/
 .vscode-test-web/
 .vscode-test/
-.venv/
+.venv*/
```
billti, line 9: This matches the existing pattern in  .gitignore (I was testing with various Python versions and environments all prefixed `.venv`)

build.py
```diff
@@ -430,30 +429,62 @@ def run_ci_historic_benchmark():
     # Reuse (or create) the pip environment so qdk wheel can be built/installed consistently.
     python_bin, pip_env = use_python_env(qdk_python_src)
 
-    # Build the qdk wheel with maturin (it now owns the native extension).
-    build_maturin_wheel(qdk_python_src, python_bin, pip_env)
+    # Read the build dependencies out of the pyproject.toml and install them first so
```
fedimser, line 443: it seems like there is a lot code duplication here, I'd add a helper to run Python:

```
def run_python(args: list[str]):
    subprocess.run(
        [python_bin] + args,
        check=True,
        text=True,
        cwd=qdk_python_src,
        env=pip_env,
    )
```

fedimser (approved):

build.py
fedimser: Although I might be wrong, I see that cwd is not always qdk_python_src. Just pointing out that it's not very easy to read now.

fedimser (approved):

build.py
billti: Hopefully this looks better.

PR #3225, opened 2026-05-15
Update Python API Doc Strings

ScottCarda-MS:
# Update docstrings under `qdk` and convert `qsharp` shims to deprecation docstrings

For a preview of the docs after this change see: https://review.learn.microsoft.com/en-us/python/qdk/qdk?view=qsharp-py&branch=pr-en-us-105
Note that we may want to omit the `qsharp` package documentation going forward as it is now a deprecated package.

## Summary

Following the migration of all package content from `qsharp` (under [source/pip/](source/pip/)) into `qdk` (under [source/qdk_package/](source/qdk_package/)), this PR updates docstrings to reflect the new home of the code and fixes a large set of broken Sphinx cross-references throughout the `qdk` Python API documentation.

The core changes are:

1. Replace stale `qsharp.*` Sphinx cross-references in `qdk` docstrings with the correct `qdk.*` paths.
2. Fully qualify bare cross-references in module-level docstrings so they resolve in py2docfx output.
3. Repoint dead `qdk._native.*` cross-references at their public re-exports (and add public re-exports for three return types that didn't have one).
4. Convert per-file `# Deprecated: use qdk.X instead` comments in the `qsharp` shim into proper module docstrings so the deprecation hint shows up in `help()`, IDE tooltips, and Sphinx autodoc.
5. Refresh the top-level `qdk/__init__.py` docstring to give an overview of the package, its core submodules, optional extras, and its relationship to the legacy `qsharp` shim.

No behavioral changes. No exports were removed. While there are some public API changes that this work highlights would be good to have, this PR will focus on documentation changes and avoid making changes to the public API.

## Public API additions

This PR previously made some changes to the public API that have been reverted. We want to avoid making those changes in this PR and a follow up team discussion and PR will be made to make the necessary changes.

### `qdk` docstrings — qsharp -> qdk path corrections

Updated stale Sphinx cross-references that still pointed at the pre-migration `qsharp.*` namespace:

- Replaced `:class:\`~qsharp.X\``, `:func:\`~qsharp.X\``, and `:attr:\`~qsharp.X\`` references with the correct `~qdk.X` paths in [qdk/_interpreter.py](source/qdk_package/qdk/_interpreter.py), [qdk/_context.py](source/qdk_package/qdk/_context.py), [qdk/openqasm/_circuit.py](source/qdk_package/qdk/openqasm/_circuit.py), [qdk/cirq/__init__.py](source/qdk_package/qdk/cirq/__init__.py), [qdk/qiskit/__init__.py](source/qdk_package/qdk/qiskit/__init__.py), [qdk/qiskit/jobs/qsjob.py](source/qdk_package/qdk/qiskit/jobs/qsjob.py), [qdk/qiskit/jobs/qsjobset.py](source/qdk_package/qdk/qiskit/jobs/qsjobset.py), and [qdk/estimator/_estimator.py](source/qdk_package/qdk/estimator/_estimator.py).
- Updated `from qsharp.X import Y` and `qsharp.compile("...")` examples in docstrings to use `qdk.*` paths in [qdk/openqasm/_compile.py](source/qdk_package/qdk/openqasm/_compile.py), [qdk/_interpreter.py](source/qdk_package/qdk/_interpreter.py), and [qdk/qiskit/backends/neutral_atom_backend.py](source/qdk_package/qdk/qiskit/backends/neutral_atom_backend.py).
- Changed text references to `qsharp.Result` to `qdk.Result` in [qdk/cirq/_result.py](source/qdk_package/qdk/cirq/_result.py).
- Reworded "Disable qsharp Python telemetry" to "Disable QDK Python telemetry" in [qdk/telemetry.py](source/qdk_package/qdk/telemetry.py). The environment variable name (`QSHARP_PYTHON_TELEMETRY`) is unchanged for back-compat.

### `qdk` docstrings — fully-qualified Sphinx cross-references

py2docfx does not respect `__all__` when resolving bare `:class:` / `:func:` references in docstrings, so module-level docstrings that listed unqualified names (like `:func:\`init\``) produced unresolved xrefs. These have been fully qualified throughout:

- [qdk/__init__.py](source/qdk_package/qdk/__init__.py) — top-level docstring.
- [qdk/qsharp.py](source/qdk_package/qdk/qsharp.py) — Key exports list.
- [qdk/_types.py](source/qdk_package/qdk/_types.py) — Types defined here list.
- [qdk/_interpreter.py](source/qdk_package/qdk/_interpreter.py) — module docstring + `init`, `compile`, `circuit` docstrings.
- [qdk/_context.py](source/qdk_package/qdk/_context.py) — `Context.__init__`, `Context.circuit`, `Context.compile` docstrings.
- [qdk/openqasm/__init__.py](source/qdk_package/qdk/openqasm/__init__.py) — Key exports list.
- [qdk/openqasm/_circuit.py](source/qdk_package/qdk/openqasm/_circuit.py), [qdk/openqasm/_compile.py](source/qdk_package/qdk/openqasm/_compile.py), [qdk/openqasm/_run.py](source/qdk_package/qdk/openqasm/_run.py) — return-type and exception cross-references.

### `qdk` docstrings — repointed away from private `qdk._native.*`

`qdk._native` is the Rust extension and is not documented, so cross-references to it are dead links. Each was repointed at its public re-export:

| Was                          | Now                          | Notes                                                   |
| ---------------------------- | ---------------------------- | ------------------------------------------------------- |
| `qdk._native.QasmError`      | `qdk.openqasm.QasmError`     | Already publicly re-exported.                           |
| `qdk._native.OutputSemantics`| `qdk.openqasm.OutputSemantics`| Already publicly re-exported.                          |
| `qdk._native.ProgramType`    | `qdk.openqasm.ProgramType`   | Already publicly re-exported.                           |
| `qdk._native.Circuit`        | `qdk.qsharp.Circuit`         | **New public re-export** added in this PR.              |

Three additional return types had no public path at all, and have been re-exported from `qdk.qsharp`:

- `Circuit` — return type of `qdk.qsharp.circuit()`.
- `QirInputData` — return type of `qdk.qsharp.compile()` and `qdk.openqasm.compile()`.
- `Config` — return type of `qdk.qsharp.init()`.

These also unblock removing the `qdk._native.X` references in docstrings.

### `qdk` docstrings — method/attribute self-reference cleanup

py2docfx does not generate per-method uids for the page describing a class, so `:meth:\`result\`` from inside `class QsJob:` doesn't resolve. Such self-references have been converted to plain backticks (e.g. `` `result()` ``):

- [qdk/qiskit/jobs/qsjob.py](source/qdk_package/qdk/qiskit/jobs/qsjob.py)
- [qdk/qiskit/jobs/qsjobset.py](source/qdk_package/qdk/qiskit/jobs/qsjobset.py)
- [qdk/qiskit/backends/qsharp_backend.py](source/qdk_package/qdk/qiskit/backends/qsharp_backend.py)
- [qdk/qiskit/backends/re_backend.py](source/qdk_package/qdk/qiskit/backends/re_backend.py)
- [qdk/qiskit/backends/neutral_atom_backend.py](source/qdk_package/qdk/qiskit/backends/neutral_atom_backend.py)
- [qdk/cirq/_result.py](source/qdk_package/qdk/cirq/_result.py)
- [qdk/applications/magnets/utilities/hypergraph.py](source/qdk_package/qdk/applications/magnets/utilities/hypergraph.py)
- [qdk/applications/magnets/utilities/pauli.py](source/qdk_package/qdk/applications/magnets/utilities/pauli.py)
- [qdk/applications/magnets/trotter/trotter.py](source/qdk_package/qdk/applications/magnets/trotter/trotter.py)

### `qdk` docstrings — qre cleanup (no API change)

Limited to docstring text fixes — public surface unchanged:

- Fixed `:param ctx (Context):` annotations that should have read `(ISAContext)` in [qdk/qre/_isa_enumeration.py](source/qdk_package/qdk/qre/_isa_enumeration.py).
- Fixed `(Node | None)` and `(Node)` parameter annotations that should have read `(ISAQuery | None)` / `(ISAQuery)` in [qdk/qre/_instruction.py](source/qdk_package/qdk/qre/_instruction.py). The `Node | None` form was also being mis-parsed by py2docfx into a malformed nested `<xref:>`.
- Removed bare `BindingNode` references from the public `ISATransform.bind` docstring.
- Rewrote `Returns: _InstructionSourceNodeReference:` / `default (Optional[_InstructionSourceNodeReference])` docstring annotations as descriptive prose since the type is private. The annotations on the actual method signatures are unchanged.

A number of qre docs xrefs are still unresolved after this pass — see [Known limitations of py2docfx](#known-limitations-of-py2docfx) below.

### Top-level `qdk/__init__.py` docstring

The top-level docstring was rewritten to introduce QDK as the unified Python entry point, list the always-available core submodules (`qdk.qsharp`, `qdk.openqasm`, `qdk.simulation`, `qdk.estimator`, `qdk.qre`, `qdk.code`), enumerate the convenience re-exports at the package root, list optional extras (`qdk[azure]`, `qdk[qiskit]`, `qdk[cirq]`, `qdk[jupyter]`), and note the relationship to the deprecated `qsharp` shim.

### `qsharp` shim — deprecation comments converted to docstrings

All ~37 shim files in [source/pip/qsharp/](source/pip/qsharp/) had a leading `# Deprecated: use qdk.X instead` (or `# Deprecation shim - delegates to qdk.X`) comment. These have been converted to proper module docstrings of the form `"""Deprecated. Use :mod:\`qdk.X\` instead."""` so the deprecation hint shows up in:

- `help(qsharp.X)` output
- IDE tooltips and hover documentation
- Sphinx `autodoc` extraction

The top-level [pip/qsharp/__init__.py](source/pip/qsharp/__init__.py) already has a proper docstring with migration guidance and a `DeprecationWarning`; that file was not changed.

Functional behavior of every shim is unchanged.

## What was intentionally **not** changed

- **Telemetry event names** (`"qsharp.run"`, `"qsharp.compile"`, ...) — these are stable analytics identifiers.
- **Environment variable names** (`QSHARP_PYTHON_TELEMETRY`, `QSHARP_PYTHON_AI_KEY`, ...) — public stable contract.
- **Internal class / function / attribute names** that legitimately reference Q# the language, such as `QSharpError`, `QSharpBackend`, `QSharpApplication`, `__qsharp_class`, `_qsharp_value_to_python_value`.
- **`qsharp.json`** — actual project manifest filename.
- **External pip package names**: `qsharp_widgets`, `qsharp-widgets`, `qsharp-jupyterlab`.

## Known limitations of py2docfx

A number of cross-references remain unresolved in the generated YAML output but are **not source-fixable** without changing the public API surface. They are noted here so reviewers don't expect this PR to address them:

- **Bare leaf names in signature renderings.** py2docfx renders type annotations on function/method signatures using only the leaf name (e.g. `Result`, `Pauli`, `Circuit`, `EstimatorResult`, `LogicalCounts`, `EstimationTableEntry`, `ConstraintBound`) when the type is defined in or imported into the same module, regardless of `__all__`. The pages exist at the qualified paths; the xrefs from inside signature blocks just don't link.
- **Enum members.** py2docfx does not generate per-enum-member uids, so `qdk.qsharp.CircuitGenerationMethod.ClassicalEval` etc. don't resolve.
- **External / cross-package references** (`azure.quantum.*`, `cirq.*`, `pandas.DataFrame`, `matplotlib.figure.Figure`, `datetime.datetime`, builtins like `int` / `bool` / `typing.*`) — these need an xrefmap to be configured at the docs pipeline level.
- **`qdk.qre._Context`, `qdk.qre.TraceParameters`, `qdk.qre.Instruction`, `qdk.qre.models.ISA`, `qdk.qre.models.ISAContext`, `_InstructionSourceNodeReference`** — fixing each of these would require either renaming a private symbol to public, adding a TypeVar to `__all__`, or re-exporting a Rust-side type from additional submodules. Out of scope for this PR.

## Tooling note

The python-doc-gen tool that consumes these docstrings (py2docfx) has a roughly 30-minute build time, no local validation, no `--strict` mode that flags unresolved xrefs as errors, and several known limitations (listed above) that mean a "valid Sphinx docstring" is not necessarily a "valid py2docfx docstring." Worth a discussion separately about whether to invest in a faster / more transparent doc generation toolchain (e.g. Sphinx + intersphinx with `nitpicky` mode), since the current iteration loop is a significant productivity drag for any non-trivial docstring change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ __pycache__/
 samples/scratch/
 *.pyd
 /python_doc/
+/logs/
--- a/source/pip/qsharp/code/__init__.py
+++ b/source/pip/qsharp/code/__init__.py
@@ -1,10 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# Deprecated: use qdk.code instead.
-#
-# This module replaces itself in sys.modules with the real qdk.code module
-# so that dynamically generated Q# callables stay in sync.
+"""Deprecated. Use :mod:`qdk.code` instead.
+
+This module replaces itself in :mod:`sys.modules` with the real :mod:`qdk.code`
+module so that dynamically generated Q# callables stay in sync between the two
+import paths.
+"""
 
 import sys
 import qdk.code as _real_code
--- a/source/pip/qsharp/code/__init__.py
+++ b/source/pip/qsharp/code/__init__.py
@@ -1,14 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-"""Deprecated. Use :mod:`qdk.code` instead.
-
-This module replaces itself in :mod:`sys.modules` with the real :mod:`qdk.code`
-module so that dynamically generated Q# callables stay in sync between the two
-import paths.
-"""
+"""Deprecated. Use :mod:`qdk.code` instead."""
 
 import sys
 import qdk.code as _real_code
 
+# This module replaces itself in :mod:`sys.modules` with the real :mod:`qdk.code`
+# module so that dynamically generated Q# callables stay in sync between the two
+# import paths.
 sys.modules[__name__] = _real_code
--- a/source/pip/qsharp/interop/cirq/__init__.py
+++ b/source/pip/qsharp/interop/cirq/__init__.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# Deprecated: use qdk.cirq instead.
+"""Deprecated. Use :mod:`qdk.cirq` instead."""
+
 from qdk.cirq import *  # noqa: F401,F403
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -5,7 +5,7 @@
 
 The ``qdk`` package is the unified Python entry point for the Microsoft Quantum
 Development Kit. It bundles the Q# interpreter, OpenQASM tooling, simulators,
-the Resource Estimator, and interoperability layers under a single import
+the Resource Estimator, and interoperability utilities under a single import
 root.
 
 Core submodules (always available):
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -1,18 +1,41 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-"""qdk bundling top-level package.
+"""Microsoft Quantum Development Kit (QDK) for Python.
 
-Provides a single import root ``qdk`` that exposes bundled quantum tooling as
-submodules (``qdk.qsharp``, ``qdk.widgets``, etc.).
+The ``qdk`` package is the unified Python entry point for the Microsoft Quantum
+Development Kit. It bundles the Q# interpreter, OpenQASM tooling, simulators,
+the Resource Estimator, and interoperability layers under a single import
+root.
 
-Optional extras install additional dependencies and submodules:
+Core submodules (always available):
 
-- ``azure`` — installs ``azure-quantum``, available as ``qdk.azure``.
-- ``qiskit`` — installs ``qiskit`` and makes Qiskit interop functionality available as ``qdk.qiskit``.
-- ``cirq`` — installs ``cirq-core`` + ``cirq-ionq`` and makes Cirq interop functionality available as ``qdk.cirq``.
-- ``jupyter`` — installs ``qsharp-widgets`` + ``qsharp-jupyterlab``; exposes ``qdk.widgets``.
+- :mod:`qdk.qsharp` — Q# interpreter and core operations
+  (:func:`~qdk.qsharp.init`, :func:`~qdk.qsharp.eval`, :func:`~qdk.qsharp.run`,
+  :func:`~qdk.qsharp.compile`, :func:`~qdk.qsharp.circuit`, etc.).
+- :mod:`qdk.openqasm` — compile, run, and estimate OpenQASM programs.
+- :mod:`qdk.simulation` — noise-aware quantum simulators and the
+  ``NeutralAtomDevice``.
+- :mod:`qdk.estimator` — the legacy Microsoft Resource Estimator API.
+- :mod:`qdk.qre` — the next-generation Resource Estimator (QRE v3).
+- :mod:`qdk.code` — namespace populated with user-defined Q# callables.
 
+Frequently used utilities are also re-exported at the package root for
+convenience: :func:`init`, :func:`dump_machine`, :func:`set_quantum_seed`,
+:func:`set_classical_seed`, :class:`Result`, :class:`TargetProfile`,
+:class:`StateDump`, :class:`ShotResult`, :class:`PauliNoise`,
+:class:`DepolarizingNoise`, :class:`BitFlipNoise`, :class:`PhaseFlipNoise`,
+and :class:`Context`.
+
+Optional extras install additional dependencies and enable extra submodules:
+
+- ``qdk[azure]`` — installs ``azure-quantum``, available as :mod:`qdk.azure`.
+- ``qdk[qiskit]`` — installs ``qiskit`` and exposes Qiskit interop as
+  :mod:`qdk.qiskit`.
+- ``qdk[cirq]`` — installs ``cirq-core`` and ``cirq-ionq`` and exposes Cirq
+  interop as :mod:`qdk.cirq`.
+- ``qdk[jupyter]`` — installs ``qsharp-widgets`` and ``qsharp-jupyterlab`` and
+  exposes :mod:`qdk.widgets`.
 """
 
 from .telemetry_events import on_qdk_import
@@ -47,7 +70,7 @@ try:
 except NameError:
     pass
 
-# utilities lifted from qsharp
+# Public API exposed at the top of the qdk package.
 __all__ = [
     "code",
     "set_quantum_seed",
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -21,11 +21,13 @@ Core submodules (always available):
 - :mod:`qdk.code` — namespace populated with user-defined Q# callables.
 
 Frequently used utilities are also re-exported at the package root for
-convenience: :func:`init`, :func:`dump_machine`, :func:`set_quantum_seed`,
-:func:`set_classical_seed`, :class:`Result`, :class:`TargetProfile`,
-:class:`StateDump`, :class:`ShotResult`, :class:`PauliNoise`,
-:class:`DepolarizingNoise`, :class:`BitFlipNoise`, :class:`PhaseFlipNoise`,
-and :class:`Context`.
+convenience: :func:`~qdk.qsharp.init`, :func:`~qdk.qsharp.dump_machine`,
+:func:`~qdk.qsharp.set_quantum_seed`, :func:`~qdk.qsharp.set_classical_seed`,
+:class:`~qdk.qsharp.Result`, :class:`~qdk.qsharp.TargetProfile`,
+:class:`~qdk.qsharp.StateDump`, :class:`~qdk.qsharp.ShotResult`,
+:class:`~qdk.qsharp.PauliNoise`, :class:`~qdk.qsharp.DepolarizingNoise`,
+:class:`~qdk.qsharp.BitFlipNoise`, :class:`~qdk.qsharp.PhaseFlipNoise`,
+and :class:`~qdk.Context`.
 
 Optional extras install additional dependencies and enable extra submodules:
 
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -36,7 +36,7 @@ Optional extras enable additional submodules:
 - ``qdk[qiskit]`` — Qiskit interoperability (:mod:`qdk.qiskit`).
 - ``qdk[cirq]`` — Cirq interoperability (:mod:`qdk.cirq`).
 - ``qdk[jupyter]`` — interactive Jupyter widgets and JupyterLab integration
-  (:mod:`qdk.widgets`).
+  (``qdk.widgets``).
 """
 
 from .telemetry_events import on_qdk_import
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -37,7 +37,7 @@ Optional extras install additional dependencies and enable extra submodules:
 - ``qdk[cirq]`` — installs ``cirq-core`` and ``cirq-ionq`` and exposes Cirq
   interop as :mod:`qdk.cirq`.
 - ``qdk[jupyter]`` — installs ``qsharp-widgets`` and ``qsharp-jupyterlab`` and
-  exposes :mod:`qdk.widgets`.
+  exposes ``qdk.widgets``.
 """
 
 from .telemetry_events import on_qdk_import
--- a/source/qdk_package/qdk/__init__.py
+++ b/source/qdk_package/qdk/__init__.py
@@ -18,7 +18,7 @@ Core submodules (always available):
   ``NeutralAtomDevice``.
 - :mod:`qdk.estimator` — the legacy Microsoft Resource Estimator API.
 - :mod:`qdk.qre` — the next-generation Resource Estimator (QRE v3).
-- :mod:`qdk.code` — namespace populated with user-defined Q# callables.
+- :mod:`qdk.code` — namespace populated with user-defined Q# and OpenQASM callables.
 
 Frequently used utilities are also re-exported at the package root for
 convenience: :func:`~qdk.qsharp.init`, :func:`~qdk.qsharp.dump_machine`,
@@ -29,15 +29,14 @@ convenience: :func:`~qdk.qsharp.init`, :func:`~qdk.qsharp.dump_machine`,
 :class:`~qdk.qsharp.BitFlipNoise`, :class:`~qdk.qsharp.PhaseFlipNoise`,
 and :class:`~qdk.Context`.
 
-Optional extras install additional dependencies and enable extra submodules:
+Optional extras enable additional submodules:
 
-- ``qdk[azure]`` — installs ``azure-quantum``, available as :mod:`qdk.azure`.
-- ``qdk[qiskit]`` — installs ``qiskit`` and exposes Qiskit interop as
-  :mod:`qdk.qiskit`.
-- ``qdk[cirq]`` — installs ``cirq-core`` and ``cirq-ionq`` and exposes Cirq
-  interop as :mod:`qdk.cirq`.
-- ``qdk[jupyter]`` — installs ``qsharp-widgets`` and ``qsharp-jupyterlab`` and
-  exposes ``qdk.widgets``.
+- ``qdk[azure]`` — Azure Quantum job submission and management
+  (:mod:`qdk.azure`).
+- ``qdk[qiskit]`` — Qiskit interoperability (:mod:`qdk.qiskit`).
+- ``qdk[cirq]`` — Cirq interoperability (:mod:`qdk.cirq`).
+- ``qdk[jupyter]`` — interactive Jupyter widgets and JupyterLab integration
+  (:mod:`qdk.widgets`).
 """
 
 from .telemetry_events import on_qdk_import
--- a/source/qdk_package/qdk/cirq/__init__.py
+++ b/source/qdk_package/qdk/cirq/__init__.py
@@ -3,14 +3,14 @@
 
 """Cirq interoperability for the Q# ecosystem.
 
-This module provides a :class:`~qsharp.interop.cirq.NeutralAtomSampler` — a standard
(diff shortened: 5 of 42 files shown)


source/qdk_package/qdk/openqasm/_circuit.py
```diff
@@ -31,14 +31,14 @@ def circuit(
...
         control flow at circuit generation time.
-        :attr:`~qsharp.CircuitGenerationMethod.Simulate` runs a full simulation to
+        :attr:`~qdk.qsharp.CircuitGenerationMethod.Simulate` runs a full simulation to
         trace the circuit.
-        :attr:`~qsharp.CircuitGenerationMethod.Static` uses partial evaluation and
+        :attr:`~qdk.qsharp.CircuitGenerationMethod.Static` uses partial evaluation and
         requires a non-``Unrestricted`` target profile. Defaults to ``None`` which
         auto-selects the generation method.
-    :kwtype generation_method: :class:`~qsharp.CircuitGenerationMethod`
+    :kwtype generation_method: :class:`~qdk.qsharp.CircuitGenerationMethod`
```
billti, line 41: Why are all the above under `qdk.qsharp.*` if this is a module under `qdk.openqasm`?

source/qdk_package/qdk/openqasm/_circuit.py
```diff
@@ -63,7 +63,7 @@ def circuit(
     :kwtype search_path: str
 
     :return: The synthesized circuit.
-    :rtype: :class:`~qsharp._native.Circuit`
+    :rtype: :class:`~qdk.qsharp.Circuit`
```
billti: If Circuit is not unique to qsharp, it should be under `qdk` directly, not `qdk.qsharp`

source/qdk_package/qdk/openqasm/_compile.py
```diff
@@ -40,7 +40,7 @@ def compile(
         - ``search_path`` (str): The optional search path for resolving file references.
         - ``output_semantics`` (OutputSemantics): The output semantics for the compilation.
     :return: The compiled program. Use ``str()`` to get the QIR string.
-    :rtype: QirInputData
+    :rtype: :class:`~qdk.qsharp.QirInputData`
```
billti: If the OpenQASM `compile` call returns this, why is it under `qdk.qsharp` rather than just `qdk`?

source/qdk_package/qdk/qiskit/backends/neutral_atom_backend.py
```diff
@@ -99,7 +99,7 @@ def __init__(
             already expressed in terms of the target gate set.
         :type skip_transpilation: bool
         :param **options: Default option overrides. These can also be overridden per-call via
-            :meth:`run`. Common options:
+            ``run()``. Common options:
```
billti: Why are you removing the method designator here and in other places and turning it into a string literal? The cross-ref link used to work fine as far as I can tell.

source/qdk_package/qdk/qiskit/jobs/qsjobset.py
```diff
@@ -28,8 +28,8 @@ class QsJobSet(Job):
     A Qiskit job set that runs multiple quantum circuits concurrently.
 
     Each circuit in ``run_input`` is submitted as an individual
-    :class:`~qsharp.interop.qiskit.jobs.qsjob.QsSimJob` and executed in a thread pool.
-    :meth:`result` blocks until all constituent
```
billti, line 32: Again, not sure why you are removing all these links. For example, if I go to the current docs at <https://learn.microsoft.com/en-us/python/qsharp/qsharp.interop.qiskit.jobs.qsjobset?view=qsharp-py>, this method result link works fine.

source/qdk_package/qdk/__init__.py
```diff
@@ -1,18 +1,43 @@
...
-- ``jupyter`` — installs ``qsharp-widgets`` + ``qsharp-jupyterlab``; exposes ``qdk.widgets``.
+- :mod:`qdk.qsharp` — Q# interpreter and core operations
+  (:func:`~qdk.qsharp.init`, :func:`~qdk.qsharp.eval`, :func:`~qdk.qsharp.run`,
+  :func:`~qdk.qsharp.compile`, :func:`~qdk.qsharp.circuit`, etc.).
+- :mod:`qdk.openqasm` — compile, run, and estimate OpenQASM programs.
+- :mod:`qdk.simulation` — noise-aware quantum simulators and the
+  ``NeutralAtomDevice``.
+- :mod:`qdk.estimator` — the legacy Microsoft Resource Estimator API.
+- :mod:`qdk.qre` — the next-generation Resource Estimator (QRE v3).
+- :mod:`qdk.code` — namespace populated with user-defined Q# callables.
```
billti: OpenQASM goes in here too, not just Q#.

source/qdk_package/qdk/__init__.py
```diff
@@ -1,18 +1,43 @@
...
+and :class:`~qdk.Context`.
+
+Optional extras install additional dependencies and enable extra submodules:
+
+- ``qdk[azure]`` — installs ``azure-quantum``, available as :mod:`qdk.azure`.
+- ``qdk[qiskit]`` — installs ``qiskit`` and exposes Qiskit interop as
+  :mod:`qdk.qiskit`.
+- ``qdk[cirq]`` — installs ``cirq-core`` and ``cirq-ionq`` and exposes Cirq
+  interop as :mod:`qdk.cirq`.
+- ``qdk[jupyter]`` — installs ``qsharp-widgets`` and ``qsharp-jupyterlab`` and
```
billti: Here and above, I don't think you need to say what other packages these install. (Or if you do, say these are package - i.e. "installs the `cirq-core` and `cirq-ionq` packages and ...")

source/qdk_package/qdk/qsharp.py
```diff
@@ -22,6 +43,8 @@
     DepolarizingNoise,
     BitFlipNoise,
     PhaseFlipNoise,
+    QirInputData,
+    Config,
```
billti: Both of these should probably be on the top level qdk module. They are not specific to Q#.

source/qdk_package/qdk/openqasm/_circuit.py
ScottCarda-MS: The `CircuitGenerationMethod` is publicly exposed through the qsharp module, not through the openqasm module. Changing that would require a change to the public API, which is better left to a PR specific to making those changes. This PR will focus just on documentation updates.

source/qdk_package/qdk/openqasm/_circuit.py
ScottCarda-MS: I've removed the public API changes from this PR to keep it cleanly about documentation. Will follow up with API changes in the future.

source/qdk_package/qdk/openqasm/_compile.py
ScottCarda-MS: I've removed the public API changes from this PR to keep it cleanly about documentation. Will follow up with API changes in the future.

source/qdk_package/qdk/__init__.py
ScottCarda-MS: Corrected

source/qdk_package/qdk/__init__.py
ScottCarda-MS: Reworded to focus on what topics the extras provide instead of packages installed.

source/qdk_package/qdk/qsharp.py
ScottCarda-MS: I've removed the public API changes from this PR to keep it cleanly about documentation. Will follow up with API changes in the future.

source/qdk_package/qdk/qiskit/backends/neutral_atom_backend.py
ScottCarda-MS: I ran into some issues working with these and though py2docfx was not consistently supporting them, so I had removed the ones that were causing trouble. After the rest of the changes for this PR, I've put them back in and am not having issues anymore, so it seems they are good to remain after all.

source/qdk_package/qdk/qiskit/jobs/qsjobset.py
ScottCarda-MS: I've put these back in. See other comment.

.gitignore
```diff
@@ -27,3 +27,4 @@ __pycache__/
 samples/scratch/
 *.pyd
 /python_doc/
+/logs/
```
amcasey, line 30: If we control this string, something like "doc_logs" might be clearer.

source/qdk_package/qdk/telemetry.py
```diff
@@ -11,7 +11,7 @@
 optional properties. The telemetry will be batched and sent at a regular intervals (60 sec),
 and when the process is about to exit.
 
-Disable qsharp Python telemetry by setting the environment variable `QSHARP_PYTHON_TELEMETRY=none`.
+Disable QDK Python telemetry by setting the environment variable `QSHARP_PYTHON_TELEMETRY=none`.
```
amcasey:
```suggestion
Disable QDK Python telemetry by setting the environment variable `QDK_PYTHON_TELEMETRY=none` (or `QSHARP_PYTHON_TELEMETRY=none`).
```

source/qdk_package/qdk/openqasm/__init__.py
```diff
@@ -1,6 +1,30 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+"""OpenQASM tooling for the QDK.
+
+This module provides functions for compiling, running, estimating, and
+generating circuits from OpenQASM 2.0/3.0 programs.
```
amcasey, line 7: Do we have the option of linking to OpenQASM (external links may be discouraged?)?

source/pip/qsharp/code/__init__.py
```diff
@@ -1,10 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# Deprecated: use qdk.code instead.
-#
-# This module replaces itself in sys.modules with the real qdk.code module
-# so that dynamically generated Q# callables stay in sync.
+"""Deprecated. Use :mod:`qdk.code` instead.
+
+This module replaces itself in :mod:`sys.modules` with the real :mod:`qdk.code`
```
amcasey: It looks like this used to not have a doc comment.  If that's the case, why add one now that it's deprecated?

source/pip/qsharp/interop/cirq/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# Deprecated: use qdk.cirq instead.
+"""Deprecated. Use :mod:`qdk.cirq` instead."""
```
amcasey, line 4: To confirm, this isn't supposed to be `qdk.interop.cirq`, like in the file path?

source/qdk_package/qdk/cirq/__init__.py
```diff
@@ -3,14 +3,14 @@
 
 """Cirq interoperability for the Q# ecosystem.
 
-This module provides a :class:`~qsharp.interop.cirq.NeutralAtomSampler` — a standard
+This module provides a :class:`~qdk.cirq.NeutralAtomSampler` — a standard
 ``cirq.Sampler`` that runs Cirq circuits on the local NeutralAtomDevice
```
amcasey, line 7: ``cirq.Sampler`` isn't supposed to be a link?

source/qdk_package/qdk/openqasm/__init__.py
```diff
@@ -1,6 +1,30 @@
...
+
+Key exports:
```
amcasey, line 9: Out of curiosity, what's up with the strict line wrapping?

amcasey (approved):

.gitignore
ScottCarda-MS: Sadly, it doesn't look like the py2docfx has an option to set the log folder.

source/qdk_package/qdk/openqasm/__init__.py
ScottCarda-MS: Hmm, I'm not sure. Maybe @jlehnersd would know if it is allowed or not.

source/pip/qsharp/interop/cirq/__init__.py
ScottCarda-MS: Yes, that's correct. In `qdk`, `cirq` and `qiskit` are top-level modules: `qdk.cirq` and `qdk.qiskit`.

source/qdk_package/qdk/cirq/__init__.py
ScottCarda-MS: That refers to official cirq content outside our repo. We could link to external docs here, if MS policy allows, but either way I wouldn't think it would be required.

source/pip/qsharp/code/__init__.py
ScottCarda-MS: The extra trick in that module to make sure the inserted Q# content makes it to both qdk.code and qsharp.code was worth the extra explanation to devs that look at the file.

ScottCarda-MS: But probably better as a comment than a doc string, so I'll move it. Thanks!

source/qdk_package/qdk/openqasm/__init__.py
ScottCarda-MS: Just a convention/habit. Not strictly enforced everywhere in code, but it is nice when the lines don't get too long.

PR #3226, opened 2026-05-15
Updated source, docstrings, and tests for qdk.magnets.trotter

brad-lackey:
Updated the computed coefficients in Suzuki and Yoshida recursion.

- Traditional formula in the literature is written in terms of order 2k
- We have defined k to be the order.
- Removed the spurious 2 from the formulas

Updated the docstrings and tests.


wavefunction91 (approved):
Can confirm this has been validated against the lattice hamiltonian / Trotter primitives in `qdk-chemistry`. We should have a discussion about what the proper place for this Trotter/magnets library is at a later date.

msoeken (approved):

PR #3228, opened 2026-05-18
Add interactive Quantum Katas learning experience to VS Code extension

minestarks:
Previously, the Quantum Katas were only accessible through the website. This PR adds a full in-editor learning experience, including an AI tutor via Copilot Chat.

### What's new

- **Activity bar panel** — A "Microsoft Quantum" sidebar with a tree view showing all Katas units and activities, with completion tracking (checkmarks, progress counts).
- **Lesson panel** — A webview that renders lesson text, runnable code examples, and exercises inline. Includes an action bar for navigation (Next/Back), Run, Check, Hint (via chat), and Explain (via chat).
- **Workspace-based progress** — On first use, a `qdk-learning/` subfolder is scaffolded in the workspace with `.qs` files for each exercise. Progress is persisted in `qdk-learning.json`.
- **Copilot agent and tools** — A `QDK Learning` chat agent backed by ~12 language model tools (`show`, `next`, `previous`, `goto`, `run`, `check`, `hint`, `solution`, `reset`, `read-code`, `get-state`, `list-units`). The agent acts as a tutor: gives hints incrementally, explains concepts, and checks solutions.
- **Code lens** — "Check Solution" / "Reset Exercise" lenses appear on exercise files.
- **Welcome view** — Shown in the sidebar when no learning workspace is detected; prompts initialization.

### Interesting files

- service.ts is the core singleton — manages catalog loading, navigation state, progress file I/O, exercise checking (via the existing `runProgram` infrastructure), and Q# file scaffolding.
- panel.ts manages the webview lifecycle; the webview app itself is a React/TSX app in `src/learning/webview/`.
- learningTools.ts wraps the service for Copilot tool invocations, including a `confirmInit` flow that prompts workspace creation on first use.

### To try it

1. Build the extension (`.build.py --wasm --npm --vscode`)
2. Open a folder in VS Code, click the Quantum activity bar icon
3. Confirm workspace setup when prompted
4. Navigate lessons/exercises in the tree or panel, or chat with the `@QDK Learning` agent

_description written with heavy help from Copilot_


source/vscode/src/learning/webview/webview.css
```diff
@@ -0,0 +1,385 @@
...
+}
+.output.fail {
+  border-left-color: var(--fail);
+}
+.output .out-label {
+  font-size: 10px;
+  text-transform: uppercase;
+  letter-spacing: 0.05em;
+  color: var(--muted);
+  margin-bottom: 0.25rem;
```
billti: I see throughout this CSS a mix of absolute size units (e.g. `10px`), relative to element (e.g, `0.05em`) and relative to root document (e.g. `0.25rem`). Has this all be done deliberately? Or does this need cleaning up?

source/npm/qsharp/generate_katas_content.js
```diff
@@ -439,10 +439,25 @@ function createExerciseSection(kataPath, properties, globalCodeSources) {
...
+  let descriptionMarkdown = tryReadFile(
     join(exercisePath, "index.md"),
     `Could not read index.md file for exercise ${properties.id}`,
   );
+
+  // Strip inline hint blocks from exercise descriptions when requested.
+  // The VS Code extension provides hints through a separate AI-powered
+  // button instead of the embedded <details> blocks.
+  const hintPattern =
+    /<details>\s*<summary>[\s\S]*?Need a hint[\s\S]*?<\/summary>([\s\S]*?)<\/details>/gi;
```
amcasey, line 451: Is the `<summary>` tag required to be first in the details section?

source/vscode/src/telemetry.ts
```diff
@@ -330,6 +333,27 @@ type EventTypes = {
     };
     measurements: Empty;
   };
+  [EventType.LearningSessionStarted]: {
+    properties: {
+      isFirstTime: string;
```
amcasey:
```suggestion
      isFirstTime: "true" | "false";
```

source/vscode/src/learning/commands.ts
```diff
@@ -0,0 +1,126 @@
...
+      async () => {
+        await service.refresh();
+      },
+    ),
+
+    vscode.commands.registerCommand(
+      "qsharp-vscode.learningContinue",
+      async () => {
+        // No position recorded yet — open chat with a generic start prompt.
+        await vscode.commands.executeCommand("workbench.action.chat.open", {
```
ScottCarda-MS, line 63: I'm curious how this works. The command should be to continue where you left off (assuming you have somewhere to continue from). The query looks like it is always "Let's start the Quantum Katas." as though it is always starting fresh.

source/vscode/src/learning/commands.ts
```diff
@@ -0,0 +1,126 @@
...
+      "qsharp-vscode.learningAskInChat",
+      async (node: LearningProgressNode) => {
+        const location = nodeToLocation(node);
+        if (!location) {
+          return;
+        }
+
+        // Include #goto with precise IDs so the agent can call the
+        // tool without fuzzy matching.
+        const prompt = `/qdk-learning #goto ${location.unitId} ${location.activityId} — Go to this activity`;
```
ScottCarda-MS, line 93: I'm not sure how this one works either. I also played around with this and found it was sort of shallow in its response. What is the intended behavior of clicking the associated button? What should copilot be telling the user in this case?

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -0,0 +1,342 @@
...
+    const code = isExercise
+      ? await this.service.readUserCode()
+      : new TextDecoder().decode(await vscode.workspace.fs.readFile(uri));
+    return { code, filePath: uri.fsPath, state: this.serializeState() };
+  }
+
+  /**
+   * Return all built-in hints for the current exercise.
+   */
+  async hint(): Promise<{
```
amcasey: We might want to save the file to disk before running this so the model gets the current contents.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -0,0 +1,342 @@
...
+      const r = await this.service.run(input.shots ?? 1, "chat");
+      await this.showActivity();
+      return { result: r.result, state: this.serializeState() };
+    });
+  }
+
+  /**
+   * Check the student's solution. Marks it complete on pass.
+   */
+  async check(): Promise<{
```
amcasey: I was definitely expecting this to auto-save.

source/vscode/agents/qdk-learning.agent.md
```diff
@@ -0,0 +1,102 @@
+---
```
amcasey, line 1: UI testing feedback (not actually for this file).

amcasey: I couldn't figure out how to reset an exercise so I could get more hints about it (e.g. to try again or implement an alternate solution).

amcasey: It might be nice to (offer to) pause inline hinting while the katas have focus.

amcasey: I couldn't figure out how to jump to an arbitrary exercise (you showed me that there's a button).

amcasey: The difference between the go-to and ask-copilot-about buttons wasn't clear.

source/npm/qsharp/generate_katas_content.js
minestarks: In practice, yes. That being said, the katas are hand-authored and this structure is not enforced in any way. I wanted to get the current content up ASAP without messing with the katas, but we should probably update katas authoring to mark off the hints in a more structured way. Will file an issue.

amcasey:
```suggestion
    /<details>[\s\S]*?<summary>[\s\S]*?Need a hint[\s\S]*?<\/summary>([\s\S]*?)<\/details>/gi;
```
Would this do it?

source/vscode/src/learning/commands.ts
minestarks: The prompt is just filler. The agent (`qdk-learning`) gets its actual instructions from the actual agent file. So it will read those and respond appropriately. "let's start the quantum katas" is just something generic that won't confuse the user too much

source/vscode/src/learning/commands.ts
minestarks: The agent should call the `goto` tool to navigate to the lesson you chose. I wouldn't expect the agent to give a deep/interesting response unless you ask follow up questions.

The reason I put this button there at all, is to remind the user that they can chat with Copilot about this content if they want.

source/vscode/src/learning/webview/webview.css
minestarks: 100% Copilot. I'll be honest, I need a crash course on writing good CSS before I can confidently clean this up - I'll at least work with Copilot now to fix obvious problems like the one you called out. (If you have other tricks you can share to lint or check CSS, I'd welcome them)

minestarks: I'm cleaning up the CSS heavily so don't look at it too carefully yet...

source/vscode/build.mjs
```diff
@@ -29,10 +29,12 @@ const platformBuildOptions = {
   ui: {
     ...commonBuildOptions,
     platform: "browser",
-    outdir: join(thisDir, "out", "webview"),
+    outbase: join(thisDir, "src"),
```
amcasey, line 32: >outbase [](http://example.com/codeflow?start=4&length=7)

Is this used?

source/vscode/src/telemetry.ts
```diff
@@ -330,6 +333,27 @@ type EventTypes = {
...
+    measurements: Empty;
+  };
+  [EventType.LearningActivityAction]: {
+    properties: {
+      action: "navigate" | "run" | "check" | "hint" | "solution" | "reset";
+      activityType: "lesson" | "exercise";
+      source: "panel" | "chat" | "tree";
+    };
+    measurements: Empty;
+  };
```
amcasey, line 349: Would it be interesting to attach an activity identifier so we could tell which ones people need help with?

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -0,0 +1,342 @@
...
+      return { result: r.result, state: this.serializeState() };
+    });
+  }
+
+  // ─── Navigation & actions (open the panel) ───
+
+  /**
+   * Show the current learning activity.
+   */
+  async show(): Promise<{ state: SerializedLearningState }> {
```
amcasey: >{ state: SerializedLearningState } [](http://example.com/codeflow?start=24&length=34)

I can't remember all the TS conventions but if there were named (as a type or an interface), then we could ensure consistency of the property name.  I think we'd intersect in the other properties.  No strong feelings either way.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Move to the next item.
+   */
+  async next(): Promise<{ moved: boolean; state: SerializedLearningState }> {
+    await this.ensureInitialized();
+    const r = this.service.next("chat");
+    await this.showActivity();
+    return { moved: r.moved, state: this.serializeState() };
+  }
```
amcasey, line 185: Is this intentionally not using `this.invoke`?

source/vscode/src/learning/panel.ts
```diff
@@ -0,0 +1,405 @@
...
+    );
+    const cssUri = getUri("out", "learning", "webview", "webview-client.css");
+    const katexCssUri = getUri("out", "katex", "katex.min.css");
+    const codiconCssUri = getUri("out", "katex", "codicon.css");
+    const mobiusUri = getUri("resources", "mobius.svg");
+
+    return /*html*/ `<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
```
amcasey, line 382: ><meta charset="utf-8" /> [](http://example.com/codeflow?start=4&length=24)

Copilot thinks this should have `Content-Security-Policy` like `welcomeView.ts`, but that's beyond my expertise.

source/vscode/src/learning/progressTreeView.ts
```diff
@@ -0,0 +1,268 @@
...
+  private snapshot: OverallProgress | undefined;
+
+  update(snapshot: OverallProgress | undefined): void {
+    this.snapshot = snapshot;
+    this.emitter.fire(undefined);
+  }
+
+  getTreeItem(node: LearningProgressNode): vscode.TreeItem {
+    if (node.kind === "continue") {
+      const item = new vscode.TreeItem(
```
amcasey, line 63: Feels like some of these string literals could be constants.

source/vscode/src/learning/welcomeView.ts
```diff
@@ -0,0 +1,132 @@
...
+      align-items: center;
+      text-align: center;
+    }
+    .logo {
+      width: 64px;
+      height: 64px;
+      margin-bottom: 16px;
+    }
+    h2 {
+      margin: 0 0 8px;
```
amcasey, line 87: >margin: 0 0 8px; [](http://example.com/codeflow?start=6&length=16)

While you're heavily editing CSS...

source/vscode/src/learning/types.d.ts
```diff
@@ -0,0 +1,278 @@
...
+export type Action =
+  | "next"
+  | "back"
+  | "run"
+  | "check"
+  | "hint-chat"
+  | "explain-chat"
+  | "progress"
+  | "menu"
+  | "quit";
```
amcasey: I don't remember seeing a quit action in the UI or the code and I'm not sure what it would mean.  Maybe Copilot does it?

source/npm/qsharp/generate_katas_content.js
```diff
@@ -439,10 +439,25 @@ function createExerciseSection(kataPath, properties, globalCodeSources) {
...
+  let extractedHints;
```
joao-boechat: Can you do something like

```
  let extractedHints = {};
```

And then do

```
    hints: extractedHints
```

instead?

I know extractedHints can be undefined or null, but I think you can handle those cases inside of the if instead of doing the ternary on the return

joao-boechat: also for simplicity you could just use "hints" as the variable name

amcasey:
The only thing I really care about is having copilot/eval know about the current editor contents - the rest of my comments are just hygiene.  LGTM

amcasey (approved):
:shipit:

joao-boechat (commented):
nit: should "agents", "prompts", "skills", all be in a folder centralizing all ai-related files? Just to make sure that in the future it's easy to find all of the md files and modify their behavior if necessary.

source/vscode/src/gh-copilot/tools.ts
```diff
@@ -100,10 +105,85 @@ const toolDefinitions: {
...
   },
+  // ─── QDK Learning tools ───
+  {
+    name: "qdk-learning-get-state",
+    tool: async () => await learningTools!.getState(),
+  },
+  {
+    name: "qdk-learning-show",
+    tool: async () => await learningTools!.show(),
+    confirm: async () => learningTools!.confirmInit(),
```
joao-boechat, line 116: I wonder if there is any way to reduce the confirm boilerplate code being called repetitively in all of these tools. I came to the conclusion that by removing that repetition we might end up overcomplicating the code, but it might be something worth considering

source/vscode/build.mjs
```diff
@@ -29,10 +29,12 @@ const platformBuildOptions = {
...
+    outdir: join(thisDir, "out"),
     entryPoints: [
       join(thisDir, "src", "webview/webview.tsx"),
       join(thisDir, "src", "webview/editor.tsx"),
+      join(thisDir, "src", "learning/webview/webview-client.tsx"),
```
joao-boechat, line 37: for consistency, could we have the webview for learning be inside the webview folder, instead of learning/webview?

source/vscode/src/learning/service.ts
```diff
@@ -0,0 +1,1158 @@
...
+    const oldKata = this.findUnit(currentPos.unitId);
+    const oldActivity = oldKata.activities.find(
+      (s) => s.id === currentPos.activityId,
+    );
+    if (oldActivity?.type === "lesson") {
+      this.markComplete(currentPos);
+    }
+
+    ws.progressData.position = nextPos;
+    this.saveProgress().catch(() => {});
```
joao-boechat: If the save doesn't work, it might fail silently, and we probably would like that error to reach the user

joao-boechat (commented):
Some of copilot thinking mentions the tool calls and internal things that the user doesn't really need to see. Would be nice if there was a way to disable showing that, the experience would be more fluid

joao-boechat:
The experience is very fluid and even better than the one we had at the website. I was skeptical at the idea of running everything behind copilot but somehow you made it very easy to use, can't wait to have it shipped so I can keep going with the material haha

Thanks for all the work you put on this, Mine!

Besides my comments above, the code LGTM!

joao-boechat (approved):

source/vscode/agents/qdk-learning.agent.md
```diff
@@ -0,0 +1,102 @@
...
+name: QDK Learning
+description: "Learn quantum computing interactively with the Quantum Katas — guided lessons, hands-on exercises, and Q# code you can run, check, and explore right in VS Code."
+model: "Claude Haiku 4.5 (copilot)"
```
amcasey, line 4: Is "auto" an option?  That might be cheaper for customers and less maintenance for us.

minestarks: Frankly, this is what I tested with and I don't feel confident about the other models' ability to follow these instructions (for example GPT-5 mini was a hot mess when I tried it early on). I may change to "auto" if it starts looking better after some more testing.

source/vscode/src/learning/webview/webview.css
minestarks: Cleaned up.

source/vscode/src/gh-copilot/learningTools.ts
minestarks: Nope, oversight

source/vscode/agents/qdk-learning.agent.md
billti: On the plus side, this model is fast, cheap, and available on the 'free' plan - so if it works well then it seems like a good default.

This is a place we'll need to periodically check and update, and keep an eye on the models available though. Might need to set a team reminder for this. (If you specify a model here that has been deprecated or removed - as they do frequently to older models - what happens?)

source/npm/qsharp/generate_katas_content.js
minestarks: You're right, this can be simplified. Done

source/vscode/src/gh-copilot/learningTools.ts
minestarks: Now saving the document before we read `readUserCode`. Seems to do the trick.

source/vscode/src/learning/panel.ts
minestarks: Doesn't hurt, I suppose.

source/vscode/src/gh-copilot/learningTools.ts
minestarks: Good idea. I also noticed I was returning the state snapshot unnecessarily from some tool calls that don't mutate state (read code, get hint, list units etc), so I removed the state snapshot from those tools.

source/vscode/src/learning/service.ts
minestarks: You're right, there was no good reason to fire-and-forget the save call there, don't know why I did that. Fixed by now `await`ing the call.

source/vscode/src/telemetry.ts
minestarks: I decided against that - this may be extended in the future to be user content.

source/npm/qsharp/generate_katas_content.js
minestarks: Ehh, I don't think this makes a meaningful difference - I've never seen the `<summary>` tag _not_ be the first thing in `<details>`, and even if it weren't, then it would be ambiguous where the actual hint content was. In the off chance we add more katas we should just test the rendering of the content before committing.

Will file a follow-up issue to add structured hints to the content though.

source/vscode/build.mjs
minestarks: Yes — it's needed so esbuild preserves the subdirectory structure relative to src/ in the output.

source/vscode/build.mjs
minestarks: It's learning... it's a webview... it's learning... it's a webview... it's both, but it can't be in two places. My personal preference was to group by feature, so I went with that 🤷‍♀️

source/vscode/src/learning/progressTreeView.ts
minestarks: Oh, now that I'm looking at it, I remembered why I didn't make most of these constants. When I'm assigning to a string literal union type (`"foo" | "bar"`), I prefer to just use the string literal. Defining a constant doesn't really buy much in the way of safety, it's just one more name to mentally keep track of.

Fixed a few other spots.

source/vscode/src/learning/welcomeView.ts
minestarks: I'm not seeing this one - what's the issue? (Err, other than the fact that I didn't really read this CSS in the first place?)

source/vscode/src/learning/types.d.ts
minestarks: Oops, dead code

source/vscode/src/gh-copilot/tools.ts
minestarks: I think this is one place I'm just okay with duplicating - I like being able to read through the tool names like this and see at a glance which ones require the init step. It's not too offensive to me, but I'm open to refactoring in the future if it gets out of hand

source/vscode/agents/qdk-learning.agent.md
minestarks: Actually I _may_ be able to use `"auto"` as a fallback here in case Haiku goes away. (You can specify multiple models in this attribute.) I will play with it and add it in a follow-up PR if it works.

I hate the idea of having a periodic chore to check if agent stuff needs to be updated, but we're kind of already in that world :(

source/vscode/src/learning/welcomeView.ts
amcasey: I don't know CSS either.  I just figured that whatever sloppiness was present in the one batch was present in the other.

source/npm/qsharp/generate_katas_content.js
minestarks: Staged at follow up PR #3242 . resolving this

source/vscode/agents/qdk-learning.agent.md
minestarks: All (except pausing inline hinting) fixes staged at #3242

source/vscode/src/learning/welcomeView.ts
minestarks: OK,I reviewed this and I think it's fine. I'm going to live with this one and avoid ratholing since it's such a tiny part of the UI anyway

minestarks:
> nit: should "agents", "prompts", "skills", all be in a folder centralizing all ai-related files? Just to make sure that in the future it's easy to find all of the md files and modify their behavior if necessary.

Staged at #3242

source/vscode/agents/qdk-learning.agent.md
minestarks: Staged at follow up PR #3242 . Resolving this discussion

minestarks:
> Some of copilot thinking mentions the tool calls and internal things that the user doesn't really need to see. Would be nice if there was a way to disable showing that, the experience would be more fluid

I played around with this, and found that the _user_ can control how the Thinking parts are displayed, not us, and the setting applies to all agents, not just us.

So I'm gonna call this a limitation of us being in VS Code - we'll have to write our own application if we want more control over chat behavior :/

PR #3241, opened 2026-05-19
Added three qualitative properties.

brad-lackey:
Added three properties to QREv3:

- ASSUMPTIONS: a property where we can associate what assumptions are associated to a component/instruction
- FEASIBILITY: a numerical score that can be assigned to the likelihood of achieving the stated parameters
- TARGET_YEAR: an estimated year that the stated parameters could be achieved.

Added tests.


msoeken (approved):
Seems that one new test is failing, but other than that looks good to me.

brad-lackey:
> Seems that one new test is failing, but other than that looks good to me.

The ASSUMPTIONS property test has string values (as desired).

- Updated property setting and extraction to allow for more flexible value types.
- Tests now successful.

PR #3258, opened 2026-05-22
Update python package readmes

ScottCarda-MS:
Updates the `qdk` package readme file with new features and top-level dump operations. Updates the `qsharp` readme to make it clear that it is a deprecated package and gives instructions on migration.

### DIFF
--- a/source/pip/README.md
+++ b/source/pip/README.md
@@ -1,55 +1,75 @@
-# Q# Language Support for Python
+# qsharp (compatibility shim)
 
-> **Note:** The `qsharp` package is deprecated. Please use the [`qdk`](https://pypi.org/project/qdk/) package instead. This package is a thin compatibility shim that re-exports the `qdk` public API so that existing code continues to work.
-
-Q# is an open-source, high-level programming language for developing and running quantum algorithms.
-The `qsharp` package for Python provides interoperability with the Q# interpreter, making it easy
-to simulate Q# programs within Python.
+> **Deprecated.** The `qsharp` package is a thin compatibility shim that
+> re-exports the [`qdk`](https://pypi.org/project/qdk/) public API.
+> New projects should use `qdk` directly.
 
 ## Installation
+
+```bash
+pip install qsharp
+```
+
+This installs the `qdk` package as a dependency. For new projects, consider
+installing `qdk` directly instead:
 
 ```bash
 pip install qdk
 ```
 
-For backward compatibility, `pip install qsharp` also works and will install `qdk` as a dependency.
+## Migration
 
-## Usage
+Replace:
 
 ```python
-from qdk import qsharp
+import qsharp
+qsharp.init()
+qsharp.eval("...")
 ```
 
-Then, use the `%%qsharp` cell magic to run Q# directly in Jupyter notebook cells:
+With:
 
-```qsharp
-%%qsharp
-
-import Std.Diagnostics.*;
-
-@EntryPoint()
-operation BellState() : Unit {
-    use qs = Qubit[2];
-    H(qs[0]);
-    CNOT(qs[0], qs[1]);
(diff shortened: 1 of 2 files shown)


source/pip/README.md
```diff
@@ -1,55 +1,75 @@
-# Q# Language Support for Python
+# qsharp (compatibility shim)
```
idavis: Should this be `qsharp`?

ScottCarda-MS: Ah good catch. "qsharp" is a code term so it should be in an in-lined code block. Fixed!

source/pip/README.md
```diff
@@ -1,55 +1,75 @@
...
-The `qsharp` package for Python provides interoperability with the Q# interpreter, making it easy
-to simulate Q# programs within Python.
+> **Deprecated.** The `qsharp` package is a thin compatibility shim that
+> re-exports the [`qdk`](https://pypi.org/project/qdk/) public API.
+> New projects should use `qdk` directly.
 
 ## Installation
 
+```bash
+pip install qsharp
```
billti, line 10: Why tell people how to install the package, if the guidance is that it is deprecated, they shouldn't, and should install `qdk` instead?

source/qdk_package/README.md
```diff
@@ -86,21 +86,22 @@ Submodules:
...
+| `TargetProfile`      | class    | `qdk.qsharp.TargetProfile`      | Target capability / profile descriptor.                             |
+| `StateDump`          | class    | `qdk.qsharp.StateDump`          | Structured state dump object.                                       |
+| `ShotResult`         | class    | `qdk.qsharp.ShotResult`         | Multi-shot execution results container.                             |
+| `PauliNoise`         | class    | `qdk.qsharp.PauliNoise`         | Pauli channel noise model spec.                                     |
+| `DepolarizingNoise`  | class    | `qdk.qsharp.DepolarizingNoise`  | Depolarizing noise model spec.                                      |
+| `BitFlipNoise`       | class    | `qdk.qsharp.BitFlipNoise`       | Bit-flip noise model spec.                                          |
+| `PhaseFlipNoise`     | class    | `qdk.qsharp.PhaseFlipNoise`     | Phase-flip noise model spec.                                        |
+| `Context`            | class    | `qdk.Context`                   | Isolated Q# interpreter context for independent sessions.           |
```
billti: This applies to OpenQASM as well, it's not just a `Q# interpreter`.

source/pip/README.md
```diff
@@ -1,55 +1,75 @@
...
 ## Support
 
-For more information about the Microsoft Quantum Development Kit, visit [https://aka.ms/qdk](https://aka.ms/qdk).
+For more information about the Microsoft Quantum Development Kit, visit
+[https://aka.ms/qdk](https://aka.ms/qdk).
 
 ## Contributing
 
-Q# welcomes your contributions! Visit the Q# GitHub repository at [https://github.com/microsoft/qdk] to find out more about the project.
+Visit the Q# GitHub repository at [https://github.com/microsoft/qdk](https://github.com/microsoft/qdk)
```
billti: This should be `Visit the QDK GitHub repository`, not `Q#`.

source/pip/README.md
ScottCarda-MS: It seemed weird to me to have someone look at the pypi page for a package and be told that the way to install the package is to not install the package, but install a different package instead. I think listing both and telling the reader to install via the `qdk` is clearer.

source/pip/README.md
ScottCarda-MS: done

source/qdk_package/README.md
ScottCarda-MS: done

billti (approved):

PR #3265, opened 2026-05-26
Fix stale selection in Learning panel

amcasey:
When a user clicks an activity node in the Learning progress tree and then presses Next (or Previous) in the lesson panel, the tree view's selection highlight remained on the originally-clicked node even though the current position had advanced. This created a confusing disconnect between the visual selection and the actual active activity.

The fix varies the activity node item.id by appending :current when the node represents the active activity — the same pattern already used for unit nodes. When the current position changes, the previously-selected node's ID no longer exists in the refreshed tree, causing VS Code to automatically drop the stale selection.

### DIFF
--- a/source/vscode/src/learning/progressTreeView.ts
+++ b/source/vscode/src/learning/progressTreeView.ts
@@ -115,7 +115,11 @@ class LearningProgressTreeProvider implements vscode.TreeDataProvider<LearningPr
       : activity.type === "exercise"
         ? "Exercise"
         : "Lesson";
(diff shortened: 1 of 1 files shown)


amcasey:
Authored by Copilot, reviewed and tested by me.

minestarks (approved):

amcasey:
We confirmed offline that that ID is not consumed by our code - only by VS Code - so this shouldn't break anything (esp the progress state).

PR #3266, opened 2026-05-26
Fix button text alignment

amcasey:
Problem: Buttons with a codicon icon (e.g. "Hint" with the sparkle icon) had their text render lower than adjacent icon-free buttons. The codicon font's metrics differ from the button text font, and with default inline layout the icon's baseline pulls the text out of vertical alignment.

Solution: Made `.action-bar button` an `inline-flex` container with `align-items: center`, which vertically centers both the icon span and text regardless of font metrics. Replaced the manual space character between icon and label with a `gap: 4px` for consistent spacing.

Note: Copilot assures me that "4px" somehow scales automatically with zoom and DPI and that seems to be true in my testing.

### DIFF
--- a/source/vscode/src/learning/webview/webview-client.tsx
+++ b/source/vscode/src/learning/webview/webview-client.tsx
@@ -530,7 +530,7 @@ function ActionBar({
                   {binding.codicon && (
                     <span class={`codicon codicon-${binding.codicon}`} />
                   )}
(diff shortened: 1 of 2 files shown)


amcasey:
I don't know enough CSS to check the code, but I did test it and it seems to work.

Before:
image

After:
image

minestarks (approved):
Thanks for the fix!

PR #3187, opened 2026-04-29
FIR Transforms

idavis:
## Summary
This PR adds FIR passes to enable broader code generation scenarios. 

QIR does not support:
- Function pointers (and thus dynamic dispatch)
- Structs
- Tuples
- Generics

RIR doesn't currently support:
- Multiple returns: `return_unify` tries to remove this constraint but there are some odd things we can't deal with.

The passes peel each unsupported piece off in the pipepline.

- Monomorphize cleans up generics
- Return unify gets rid of multiple returns and allows us to better understand control flow
- defunctionalization gets rid of callable exprs
- erase_utds rewrites the FIR so that it uses tuples in place of structs
- lower_tuple_comparison handles a special case of binop replacing it with short-circuiting element-wise comparisons which can be codegen'd
- sroa and arg_promote work together to get rid of all possible tuple element usage
- dce and gc passes clean up code that isn't called any longer so that RCA doesn't pay attention to it
- there are some mini passes as well that collapse very specific patterns like the defunc prepass

Aside from the passes, this PR also tries to unify how the code goes through RCA and codegen compilation. There are some side effects which leak into circuits as we have to generate new functions as part of the passes that we don't necessarily want reflected in the circuit representation.

## Suggested Review Assignment

| Reviewer | Best-fit parts |
|---|---|
| @swernli | Core FIR transform pipeline, qsc_fir_transforms, QIR/codegen integration, partial eval, RCA, RIR, circuit behavior, root Cargo changes |
| @minestarks | Broad compiler integration, qsc/qsc_circuit/qsc_frontend/qsc_lowerer/qsc_openqasm_compiler/qsc_passes, Python/package-facing changes, RIR, circuit behavior, root Cargo changes |
| @billti | Python package tests/snapshots, fuzz target, wasm diagnostic touchpoint, resource estimator test, default-owned samples/index_map fallout |
| @ScottCarda-MS | Language service and npm snapshot changes |

## Crate organization
Integrating `qsc_fir_transforms` with `qsc_passes` was going to make the PR look much bigger with a lot of moved files. My plan was to merge `qsc_fir_tranforms` into `qsc_passes` and organize them by `HIR` and `FIR`. This way we'd have a clean refactoring PR with no functional changes. This PR is already very large and I thought this integration was just too much to add.

## Error types
[...]

PR #3268, opened 2026-05-26
Add `frem`, `fptoui`, `uitofp` instructions to QIR simulators

orpuente-MS:
This PR adds support for the `frem`, `fptoui`, and `uitofp` instructions to the QIR simulators to handle the codegen changes in #3187 .

### DIFF
--- a/source/qdk_package/tests/test_adaptive_cpu_bytecode.py
+++ b/source/qdk_package/tests/test_adaptive_cpu_bytecode.py
@@ -784,6 +784,11 @@ FLOAT_ARITH_PARAMS = [
     ("fsub", 3.0, 10.0, -7.0),
     ("fmul", 6.0, 7.0, 42.0),
     ("fdiv", 8.0, 2.0, 4.0),
+    ("frem", 8.0, 2.0, 0.0),
+    ("frem", 7.0, 3.0, 1.0),
+    ("frem", 10.5, 3.0, 1.5),
+    ("frem", 10.5, -3.0, 1.5),  # remainder has the same sign as dividend
+    ("frem", -10.5, 3.0, -1.5),  # remainder has the same sign as dividend
 ]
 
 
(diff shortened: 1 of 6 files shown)


source/qdk_package/tests/test_adaptive_cpu_bytecode.py
```diff
@@ -784,6 +784,11 @@ def test_fcmp(sim_type, pred, lhs, rhs, expected):
     ("fsub", 3.0, 10.0, -7.0),
     ("fmul", 6.0, 7.0, 42.0),
     ("fdiv", 8.0, 2.0, 4.0),
+    ("frem", 8.0, 2.0, 0.0),
```
joao-boechat, line 787: Should we have test cases where divisor is 0 to make sure it fails gracefully?

joao-boechat: Also would be nice to test fptoui with negative floats

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
```diff
@@ -1210,6 +1213,16 @@ fn interpret_classical(@builtin(global_invocation_id) gid: vec3<u32>) {
                 pc++;
             }
 
+            // FREM: Float remainder. LLVM docs say this instruction has
+            // the same semantics as C's fmod, which is implemented as:
+            // dst = src0 - trunc(src0/src1) * src1
+            case OP_FREM {
+                let a = resolve_f32(shot_idx, instr.src0, flags, 0u);
+                let b = resolve_f32(shot_idx, instr.src1, flags, 1u);
```
joao-boechat, line 1221: one thing that copilot noted and I thought was interesting: the GPU uses f32, while CPU uses f64. Should we standardize that or write this down somewhere so we don't introduce bugs in the future?

joao-boechat (approved):

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
orpuente-MS: This is a known limitation of the GPU simulator, and the QIR -> bytecode compiler follows different logic depending on if we are compiling for a 32-bit target (the CPU simulators) or a 64-bit target (the GPU simulators).

source/qdk_package/tests/test_adaptive_cpu_bytecode.py
orpuente-MS: Yes, nice catch. I added those tests to the CPU and GPU simulators.

PR #3275, opened 2026-05-29
Surface multiple solutions for katas that have them

amcasey:
I remember finding the alternative solutions useful in the web version of the katas and I thought they might be useful in VS Code too.  I decided not to try to figure out if the user's solution matched one of the available solutions since I wasn't sure Haiku could handle that reliably and it's easy enough to read past the one that matches yours.
AlternativeApproaches

### DIFF
--- a/source/vscode/agents/qdk-learning.agent.md
+++ b/source/vscode/agents/qdk-learning.agent.md
@@ -92,7 +92,7 @@ Call `hint` and `read-code` together. The response contains `hints` (short nudge
 
 ### After a Passing Check
 
-Render the result, offer a brief reaction. Don't auto-call `next` — the user may want to review the solution first. If the exercise has alternative approaches (indicated by the panel showing a "See alternative approaches" link), you may briefly mention that other approaches exist, but don't present them unless asked.
+Render the result, offer a brief reaction. Don't auto-call `next` — the user may want to review the solution first. If the exercise has multiple solutions (indicated by the `hasMultipleSolutions` property), you may briefly mention that other approaches exist, but don't present them unless asked.
 
 ## Don'ts
 
--- a/source/vscode/agents/qdk-learning.agent.md
+++ b/source/vscode/agents/qdk-learning.agent.md
@@ -47,18 +47,20 @@ Panel actions (Next, Run, Check, Solution…) work directly — no LLM round-tri
 
 The panel routes these messages to chat. Always call `get-state` first to understand context.
 
-| Button / Link         | Shown on              | Chat message                                |
-| --------------------- | --------------------- | ------------------------------------------- |
-| **Hint**              | Exercises             | "Give me a hint"                            |
-| **Explain**           | Lessons & examples    | "Explain this concept in more detail"       |
-| What went wrong?      | Failed check output   | "Help me understand why my solution failed" |
-| Explain this solution | After solution reveal | "Explain this solution step by step"        |
+| Button / Link              | Shown on              | Chat message                                      |
+| -------------------------- | --------------------- | ------------------------------------------------- |
+| **Hint**                   | Exercises             | "Give me a hint"                                  |
+| **Explain**                | Lessons & examples    | "Explain this concept in more detail"             |
+| What went wrong?           | Failed check output   | "Help me understand why my solution failed"       |
+| Explain this solution      | After solution reveal | "Explain this solution step by step"              |
+| See alternative approaches | Passed check (multi)  | "Show me alternative approaches to this exercise" |
 
 **Handling guidance:**
 
 - **"Explain this concept in more detail"** — Provide a deeper pedagogical explanation. Offer analogies, relate to prior units. Don't repeat the panel content.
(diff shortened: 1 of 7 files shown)


source/vscode/agents/qdk-learning.agent.md
```diff
@@ -47,18 +47,20 @@ Panel actions (Next, Run, Check, Solution…) work directly — no LLM round-tri
...
+| What went wrong?           | Failed check output   | "Help me understand why my solution failed"       |
+| Explain this solution      | After solution reveal | "Explain this solution step by step"              |
+| See alternative approaches | Passed check (multi)  | "Show me alternative approaches to this exercise" |
 
 **Handling guidance:**
 
 - **"Explain this concept in more detail"** — Provide a deeper pedagogical explanation. Offer analogies, relate to prior units. Don't repeat the panel content.
 - **"Help me understand why my solution failed"** — Analyze common mistakes for that exercise. Give targeted debugging hints, not the full solution.
-- **"Explain this solution step by step"** — Walk through the reference solution line by line, explaining the quantum concepts and Q# patterns.
+- **"Explain this solution step by step"** — Walk through the reference solution line by line, explaining the quantum concepts and Q# patterns. If there are multiple reference solutions, walk through only the first one.
```
minestarks: Shall we instruct Copilot to also call the `readUserCode` (?) tool to read the user's solution, and compare to the reference solution?

source/vscode/src/learning/types.d.ts
```diff
@@ -232,8 +234,8 @@ export interface CatalogExercise {
   sourceIds: string[];
   /** Author-written pedagogical hints. */
   hints: string[];
-  /** Reference solution code. */
-  solutionCode: string;
+  /** Reference solution code (one per @[solution] block in the content). */
+  solutionCodes: string[];
   /** Prose explanation of the solution (markdown). */
   solutionExplanation: string;
```
minestarks, line 240: Does this end up in the tool output?

source/vscode/agents/qdk-learning.agent.md
minestarks: This comment was meant for the line below it. Looks like this line is just out of date (and the corresponding row in the table). Could you clean it up maybe?

source/vscode/src/learning/types.d.ts
amcasey: Confirmed: "Use the `solutionExplanation` from `hint` if you need more context on the reasoning behind each approach."

source/vscode/agents/qdk-learning.agent.md
amcasey: After editing

Image

amcasey: Removed the entry for "Explain this solution"

source/vscode/agents/qdk-learning.agent.md
```diff
@@ -92,7 +92,7 @@ Call `hint` and `read-code` together. The response contains `hints` (short nudge
 
 ### After a Passing Check
 
-Render the result, offer a brief reaction. Don't auto-call `next` — the user may want to review the solution first.
+Render the result, offer a brief reaction. Don't auto-call `next` — the user may want to review the solution first. If the exercise has alternative approaches (indicated by the panel showing a "See alternative approaches" link), you may briefly mention that other approaches exist, but don't present them unless asked.
```
minestarks: The agent can't "see" this UI state, so this wording is likely to just confuse it.

Most of the time the agent can't "see" passing checks as they happen at all, because the user does it through the Check button in the UI. So this paragraph is essentially telling the agent how to handle the result of a `check` tool call that _it_ made.

Probably best to guide the agent towards the `hasAlternativeSolutions` flag for the current exercise (which should hopefully already be in context, thanks to frequent calls the agent has to make to `getState`).

> (indicated by the `hasAlternativeSolutions` flag)

source/vscode/src/learning/types.d.ts
```diff
@@ -60,6 +60,8 @@ export interface ExerciseContent {
   /** URI string for the user's .qs solution file */
   filePath: string;
   isComplete: boolean;
+  /** True when multiple reference solutions exist for this exercise. */
+  hasAlternatives: boolean;
```
minestarks: Suggest renaming this to `hasAlternativeSolutions` because the agent can't see the doc comment, just the JSON data. The agent can infer meaning from descriptive names.

minestarks (changes requested):
Thanks! Works well on the exercise I tried. A couple of minor suggestions

source/vscode/agents/qdk-learning.agent.md
amcasey: Referred it to the newly-renamed `hasMultipleSolutions` flag.

source/vscode/src/learning/types.d.ts
amcasey: Renamed to `hasMultipleSolutions`

minestarks (approved):

PR #3276, opened 2026-05-30
Add imports to kata placeholders

amcasey:
...copied from the corresponding Solution.qs.  In the absence of auto-import, I was wasting a bunch of time asking copilot where to find things like `Sqrt` and `PI()`.

### DIFF
--- a/katas/content/phase_estimation/state_eigenvector/Placeholder.qs
+++ b/katas/content/phase_estimation/state_eigenvector/Placeholder.qs
@@ -1,4 +1,6 @@
 namespace Kata {
+    import Std.Diagnostics.*;
+
     operation IsEigenvector(U : Qubit => Unit, P : Qubit => Unit is Adj) : Bool {
         // Implement your solution here...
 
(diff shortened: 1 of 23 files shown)


katas/content/phase_estimation/state_eigenvector/Placeholder.qs
```diff
@@ -1,4 +1,6 @@
 namespace Kata {
+    import Std.Diagnostics.*;
```
amcasey:
```suggestion
    import Std.Diagnostics.CheckZero;
```
This seems like a hint, but it's specified in the problem statement.

swernli:
I'm not sure what you mean by "the absence of auto-import" in this case, because we do have auto-import. If I start typing `PI` I get:
image
and then accepting the suggestion automatically adds the import:
image

amcasey:
How do you get the import if you don't accept one of the completions?  (This might arise if I'm carrying forward a previous solution or using code from copilot.)

Auto-import may not have been the correct term - whatever feature we decided to nominate for Unitary Hack: add missing import, maybe.

swernli:
Oh, I see what you mean. Yes, the code action to add missing import when there is a name error would be a good value add and is something we don't do yet.

Also, I should have clarified that I think this change you made is a good fix for the usability of the katas regardless. I just wanted to home in on the expected behavior that wasn't there in case you had found a scenario where an existing feature wasn't working.

amcasey:
> in case you had found a scenario where an existing feature wasn't working

Ask me about sig help...  Trying to work with copilot on that now.

swernli (approved):

PR #3277, opened 2026-06-01
RIFLA: Support emitting constant `Result` arrays

swernli:
This change supports the emission of constant `Result` arrays for Adaptive RIFLA QIR generation. To do so, we needed the compiler to see arrays of result values as "dynamic constant" expressions, which required a change to how `Result` values are handled during RCA and Partial Eval. Previously, any `Result` value coming back from an intrinsic operation was treated a dynamic variable, even though they are really just identifiers. This updates that logic to recognized results as identifiers, the same way qubits are identifiers, and moves the source of dynamic variables in RCA to the actual comparison of results rather than being inherent to the result itself. To make sure this quality propagates across function call boundaries appropriately, RCA was updated to differentiate between call arguments that are static vs dynamic constant vs dynamic variable, where previously it only checked for dynamic variable and treated constant and static as the same.

In addition, this also includes an update to RCA that adds a new capability to recognize when a static `Result` value (`One` or `Zero`) is used as the output from a dynamic expression, since that cannot be emitted into QIR. The new check detects this at analysis time so the error is visible in the editor before the user tries to emit QIR (if applicable).

### DIFF
--- a/source/compiler/qsc_eval/src/val.rs
+++ b/source/compiler/qsc_eval/src/val.rs
@@ -237,6 +237,7 @@ pub enum VarTy {
     Integer,
     Double,
     Qubit,
+    Result,
 }
 
 impl Display for VarTy {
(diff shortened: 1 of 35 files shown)


swernli:
Converting to draft for now. I found some test holes, new corner cases, and existing bugs that I'd like to get cleaned up and resolved first.

PR #3279, opened 2026-06-01
Handle missing rparen on call

amcasey:
Without automatic bracket matching, it's quite easy to get into a situation for a function call has an open paren, but no close paren.  When this happens, the entire line is consumed as an error node and there's no language service help (in particular, no signature help).  By pushing an error, but not returning one, we can preserve the call node so that the language service has something to work with.

Bonus: If the argument list contains an initializer (or other expression with recovery), its recovery will now stop at the close paren because we've pushed that as a barrier.  This doesn't help with the missing close paren, it's just nice to have.

I've reviewed the code thoroughly, reviewed the test baselines somewhat, and manually tested in both VS Code and the Playground.

### DIFF
--- a/source/compiler/qsc_parse/src/expr/tests.rs
+++ b/source/compiler/qsc_parse/src/expr/tests.rs
@@ -2858,3 +2858,247 @@ fn parenthesized_expr_in_for_is_valid() {
                     Stmt _id_ [16-18]: Expr: Expr _id_ [16-18]: Unit"#]],
     );
 }
+
+// Call expression recovery tests
+
+#[test]
+fn call_missing_close_paren_empty_args() {
+    check(
+        expr,
+        "foo(",
+        &expect![[r#"
+            Expr _id_ [0-4]: Call:
+                Expr _id_ [0-3]: Path: Path _id_ [0-3] (Ident _id_ [0-3] "foo")
+                Expr _id_ [3-4]: Unit
+
+            [
+                Error(
+                    Token(
+                        Close(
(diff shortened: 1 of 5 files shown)


amcasey:
SigHelp on incomplete call.  Tree doesn't have error expression.  Errors still reported.

image

amcasey:
I forgot to update the new signature_help tests after refining the parser behavior to consume fewer tokens during recovery.  Fix incoming.

source/compiler/qsc_parse/src/expr/tests.rs
```diff
@@ -2858,3 +2858,247 @@ fn parenthesized_expr_in_for_is_valid() {
...
+                            hi: 10,
+                        },
+                    ),
+                ),
+            ]"#]],
+    );
+}
+
+#[test]
+fn call_with_struct_arg_missing_close_paren() {
```
swernli, line 3065: Looks like maybe this test is either named incorrectly or has the missing rparen added by mistake.

swernli (approved):
Nice clean fix, thanks for digging into this!

source/compiler/qsc_parse/src/expr/tests.rs
amcasey: That's super weird.  I wonder if it changed the input rather than the baseline at some point.

amcasey: Fixed.

PR #3280, opened 2026-06-02
Handle PermissionError in prereqs.py

amcasey:
On WSL, it seems like we get PermissionError, rather than FileNotFoundError, when wasm-bindgen isn't present.


minestarks (approved):
I hit this too. Thanks

PR #3286, opened 2026-06-02
Change mechanism for override in RCA for `Length`, handle `Length` calls and `Range` types in Partial Eval

### LINKED ISSUE #3281: `Length` override can cause failures in QIR codegen with embedded expressions
The `Length` function has a special override to work in RCA and Partial Eval:
https://github.com/microsoft/qdk/blob/f7e34433231a1ba2bd17904aa6fa65c891bcbe2a/source/compiler/qsc_rca/src/core.rs#L424-L430

However, this ends up causing RCA to skip aggregating any runtime features for expressions embedded in the arguments to `Length`, causing them to be treated as `ComputeKind::Static` even though they may not be. Adding any qubut allocation or quantum instructions can thus cause failures in in QIR generation. This code passes the RCA check:

```qsharp
@EntryPoint(Base)
operation Main() : Unit {
    Length({use q = Qubit(); [1]});
}
```

But fails in Partial Evaluation because the embedded expression is sent to the full evaluator incorrectly:
`QDK command error: partial evaluation error: partial evaluation failed with error: simulation error: qubit_allocate operation not implemented(Qsc.PartialEval.EvaluationFailed)`

We need a different way to handle length checks so that QIR codegen doesn't need this override and can instead handle `Length` in a more general, compatible way.

### LINKED ISSUE #3282: `IsEmpty` function flagged as dynamic by RCA even when not needed
The `Std.Arrays.IsEmpty` function is very simple and internally uses `Length`:
```qsharp
function IsEmpty<'T>(array : 'T[]) : Bool {
    Length(array) == 0
}
```

The length of a statically sized array should be statically known, so `IsEmpty` should be allowed in programs that are compiling for QIR. However, `IsEmpty` gets flagged as invalid usage while the equivalent `Length` check does not:
```qsharp
import Std.Arrays.IsEmpty;
@EntryPoint(Base)
operation Main() : Unit {
    use qs = Qubit[2];
    while IsEmpty(qs) {}
    while Length(qs) == 0 {}
}
```

Image

This is closely related to #3281, in that `Length` only works because of the override it uses, but that override has it's own issues that mean it should not be generalized to more functions as it would extend those issues further. Whatever solution is chosen for the `Length` override, we need to make sure the right runtime capabilities propagate where appropriate without causing contradiction that allow invalid code or block valid code.

swernli:
Note: approach updated after introduction of monomorphization pass.

This change removes the overrider module in RCA that was only used for `Length` (holdover from a previous approach and still required some special handling later in analysis). Instead, the initial override and RCA override for the entire call expression is changed to use a single check that recognize the `Length` intrinsic and any monomorphized copies of it to ensure it gets a normal `ApplicationGeneratorSet` and call sites are treated the same as any other function. This means that it may show up as "Dynamic Constant" expression, so Partial Eval now needs to handle those calls to `Length` directly rather than assuming they will be handed off to the underlying full evaluator. Likewise, this now allows for "Dynamic Constant" `Range` values, which Partial Eval can handle appropriately. To fix the handling of loops based on these constant values, we now correctly emit the loop only if the loop condition is a "Dynamic Variable" value, rather than on any non-"Static" value.

Fixes #3281 Fixes #3282


source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -2392,7 +2417,7 @@ impl<'a> PartialEvaluator<'a> {
             .config
             .capabilities
             .contains(TargetCapabilityFlags::BackwardsBranching)
-            && !self.is_static_expr(condition_expr_id)
+            && self.is_variable_expr(condition_expr_id)
```
idavis, line 2420: changes the comment below:
// If backwards branching is supported and the loop condition is a dynamic variable,
// we can generate a while loop structure in RIR without unrolling the loop.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1813,23 +1812,49 @@ impl<'a> PartialEvaluator<'a> {
...
+                #[allow(clippy::cast_precision_loss)]
+                Value::Int(i) => Ok(Value::Double(i as f64)),
+                Value::Var(_) => {
+                    let variable_id = self.resource_manager.next_var();
+                    self.convert_value(&args_value, rir::Variable::new_double(variable_id))
+                }
+                _ => panic!(
+                    "Unexpected value type for IntAsDouble: {}",
+                    args_value.type_name()
+                ),
```
idavis, line 1844: Should this be `return Err(Error::Unexpected(..)`

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1813,23 +1812,49 @@ impl<'a> PartialEvaluator<'a> {
...
+                #[allow(clippy::cast_possible_truncation)]
+                Value::Double(d) => Ok(Value::Int(d as i64)),
+                Value::Var(_) => {
+                    let variable_id = self.resource_manager.next_var();
+                    self.convert_value(&args_value, rir::Variable::new_integer(variable_id))
+                }
+                _ => panic!(
+                    "Unexpected value type for Truncate: {}",
+                    args_value.type_name()
+                ),
```
idavis, line 1856: Should this be `return Err(Error::Unexpected(..)`

source/compiler/qsc_partial_eval/src/lib.rs
swernli: For this and below, anything other than `Int` or `Var` getting here would be impossible without violating type checking, so it really is a catastrophic failure.

source/compiler/qsc_partial_eval/src/lib.rs
swernli: Preventing by type checking.

source/compiler/qsc_partial_eval/src/lib.rs
swernli: Ah, good catch, I'll fix that.

idavis (approved):

PR #3287, opened 2026-06-03
Update error span for call argument type mismatch

amcasey:
It should be just the argument with the problem.  Hopefully, this will make it easy to add quick fixes to some scenarios.

### DIFF
--- a/source/compiler/qsc_frontend/src/typeck/infer.rs
+++ b/source/compiler/qsc_frontend/src/typeck/infer.rs
@@ -312,7 +312,7 @@ pub(super) enum ArgTy {
     /// A given argument.
     Given(Ty, Span),
     /// A list of arguments. This corresponds literally to tuple syntax, not to any expression of a tuple type.
-    Tuple(Vec<ArgTy>),
+    Tuple(Vec<ArgTy>, Span),
 }
 
 impl ArgTy {
@@ -321,7 +321,9 @@ impl ArgTy {
         match self {
             Self::Hole(ty, span) => Self::Hole(f(ty), span),
             Self::Given(ty, span) => Self::Given(f(ty), span),
-            Self::Tuple(items) => Self::Tuple(items.into_iter().map(|i| i.map(f)).collect()),
+            Self::Tuple(items, span) => {
+                Self::Tuple(items.into_iter().map(|i| i.map(f)).collect(), span)
+            }
         }
     }
 
@@ -353,20 +355,20 @@ impl ArgTy {
             },
             // if both the arg and the param are tuples, then we must check
             // the types of each element in the tuple and generate iterative applications.
-            (Self::Tuple(args), Ty::Tuple(params)) => {
+            (Self::Tuple(args, tuple_span), Ty::Tuple(params)) => {
                 let mut errors = Vec::new();
                 if args.len() != params.len() {
                     errors.push(Error(ErrorKind::TyMismatch(
                         Ty::Tuple(params.clone()).display(),
                         self.to_ty().display(),
-                        call_span,
+                        *tuple_span,
                     )));
                 }
 
                 let mut holes = Vec::new();
                 let mut constraints = Vec::new();
                 for (arg, param) in args.iter().zip(params) {
-                    let mut app = arg.apply(param, call_span);
+                    let mut app = arg.apply(param, *tuple_span);
                     constraints.append(&mut app.constraints);
                     errors.append(&mut app.errors);
                     if app.holes.len() > 1 {
@@ -383,22 +385,22 @@ impl ArgTy {
                 }
             }
 
-            (Self::Tuple(_), Ty::Infer(_)) => App {
+            (Self::Tuple(_, tuple_span), Ty::Infer(_)) => App {
                 holes: Vec::new(),
                 constraints: vec![Constraint::Eq {
                     expected: param.clone(),
                     actual: self.to_ty(),
-                    span: call_span,
+                    span: *tuple_span,
                 }],
                 errors: Vec::new(),
             },
-            (Self::Tuple(_), _) => App {
+            (Self::Tuple(_, tuple_span), _) => App {
                 holes: Vec::new(),
                 constraints: Vec::new(),
                 errors: vec![Error(ErrorKind::TyMismatch(
                     param.display(),
                     self.to_ty().display(),
-                    call_span,
+                    *tuple_span,
                 ))],
             },
         }
@@ -407,7 +409,7 @@ impl ArgTy {
     pub(super) fn to_ty(&self) -> Ty {
         match self {
             ArgTy::Hole(ty, _) | ArgTy::Given(ty, _) => ty.clone(),
-            ArgTy::Tuple(items) => Ty::Tuple(items.iter().map(Self::to_ty).collect()),
+            ArgTy::Tuple(items, _) => Ty::Tuple(items.iter().map(Self::to_ty).collect()),
         }
     }
 }
--- a/source/compiler/qsc_frontend/src/typeck/infer.rs
+++ b/source/compiler/qsc_frontend/src/typeck/infer.rs
@@ -328,7 +328,7 @@ impl ArgTy {
     }
 
     /// Applies the argument type to a parameter type, generating constraints and errors.
-    fn apply(&self, param: &Ty, call_span: Span) -> App {
+    fn apply(&self, param: &Ty) -> App {
         match (self, param) {
             // If `arg` is a hole, then it doesn't matter what the param is,
             // because the hole can be anything.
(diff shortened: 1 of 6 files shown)


amcasey:
The approach seems plausible, but I've only partially reviewed it.  Posting to get feedback on the approach.

source/compiler/qsc_frontend/src/typeck/infer.rs
```diff
@@ -333,25 +333,25 @@ impl ArgTy {
...
             // add that to the constraints.
             // Preserve the hole.
-            (Self::Hole(arg), _) => App {
+            (Self::Hole(arg, arg_span), _) => App {
                 holes: vec![param.clone()],
                 constraints: vec![Constraint::Eq {
                     expected: param.clone(),
                     actual: arg.clone(),
-                    span,
+                    span: *arg_span,
```
minestarks, line 343: One thing that might help me read this is if you renamed the `span` argument to this function (can't highlight the line since it's outside the diff) to reflect _whose_ span it is. Now that we have two spans getting passed around, it'd be good to disambiguate.

amcasey: I genuinely can't tell what that span is.  I'm pretty sure it's the span of the whole call.  I think it effectively means "here's the span where you should report an error if you find one".

source/compiler/qsc_frontend/src/typeck/infer.rs
```diff
@@ -333,25 +333,25 @@ impl ArgTy {
...
                 }],
                 errors: Vec::new(),
             },
             // If `arg` is a hole, then it doesn't matter what the param is,
             // because the hole can be anything.
             // However, we do know that the type of Arg must be Eq to the type of Param, so we
             // add that to the constraints.
-            (Self::Given(arg), _) => App {
+            (Self::Given(arg, arg_span), _) => App {
```
amcasey, line 347: I don't think a `Given` is the same as a `Hole` but the comment and implementation seem to be the same as above?

source/compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -476,15 +476,15 @@ fn int_as_double_error() {
             }
         "},
         "Microsoft.Quantum.Convert.IntAsDouble(false)",
-        &expect![[r#"
+        &expect![[r##"
```
amcasey, line 479: I think the baseline update might have done this because I didn't and it wasn't in the list of files copilot had touched.

source/compiler/qsc_frontend/src/typeck/infer.rs
amcasey: Per our offline discussion, it seems like it's right except for the comment.

source/compiler/qsc_frontend/src/typeck/infer.rs
amcasey: I believe it's gone in the latest revision.

source/compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -476,15 +476,115 @@ fn int_as_double_error() {
...
+            #7 37-44 "a : Int" : Int
+            #15 55-62 "{ 1.0 }" : Double
+            #17 57-60 "1.0" : Double
+            #18 65-82 "Namespace.F((1,))" : Double
+            #19 65-76 "Namespace.F" : (Int -> Double)
+            #23 76-82 "((1,))" : (Int,)
+            #24 77-81 "(1,)" : (Int,)
+            #25 78-79 "1" : Int
+            Error(Type(Error(TyMismatch("Int", "(Int,)", Span { lo: 76, hi: 82 }))))
+        "##]], // 77-81 would be better, but the parens are fine
```
amcasey: We agreed offline that we want to disregard all paren expressions when computing the span.

source/compiler/qsc_frontend/src/typeck/tests.rs
minestarks: Yes, common problem, and the reason why we suppress the `needless_raw_string_hashes` lint in most modules. Harmless

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -264,25 +264,48 @@ impl<'a> Context<'a> {
             }
             ExprKind::BinOp(op, lhs, rhs) => self.infer_binop(expr.span, *op, lhs, rhs),
             ExprKind::Block(block) => self.infer_block(block),
-            ExprKind::Call(callee, input) => {
-                let callee = self.infer_expr(callee);
-                let input = if has_holes(input) {
-                    self.infer_hole_tuple_arg(input)
+            ExprKind::Call(callee_expr, input_expr) => {
+                let callee = self.infer_expr(callee_expr);
+                let input_expr = &**input_expr;
```
minestarks: It turns out you can just delete this line, since `&Box<Expr>` can get coerced to `&Expr` in all the usage sites below. You don't need to explicitly dereference.

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -264,25 +264,48 @@ impl<'a> Context<'a> {
...
+                        }
+                    } else {
+                        let mut inner = input_expr;
+                        while let ExprKind::Paren(unwrapped) = inner.kind.as_ref() {
+                            inner = unwrapped;
+                        }
+                        let arg_span = inner.span;
+                        Partial {
+                            ty: ArgTy::Given(ty.ty, arg_span),
                             diverges: false,
```
minestarks, line 302: Not your code, but it seems suspicious to me that we're ignoring `diverges` in this branch, but propagating it in the `if` branch above.

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -264,25 +264,48 @@ impl<'a> Context<'a> {
...
-                    match ty {
+                    if let Partial {
+                        ty: Ty::Tuple(tys),
+                        diverges,
+                    } = ty
+                    {
+                        let mut inner = input_expr;
+                        while let ExprKind::Paren(unwrapped) = inner.kind.as_ref() {
+                            inner = unwrapped;
+                        }
```
minestarks: These four lines are repeated in both the `if` and `else` blocks - would it be clearer to pull them out?

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -264,25 +264,48 @@ impl<'a> Context<'a> {
...
+                    let ty = self.infer_expr(input_expr);
                     // If the outermost element is a tuple, we must wrap it in an `ArgTy::Tuple`.
-                    match ty {
+                    if let Partial {
+                        ty: Ty::Tuple(tys),
+                        diverges,
+                    } = ty
+                    {
+                        let mut inner = input_expr;
+                        while let ExprKind::Paren(unwrapped) = inner.kind.as_ref() {
```
minestarks: Would we ever expect to see more than one layer of `Paren` in the input part of a call expression? I don't think so.

minestarks: But if you think we would, then it'd help to add a unit test - one that would fail if you changed this to `if let`

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: Thanks!  I did find that rather mysterious.

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: I remember noticing that too.  I think I convinced myself that it was guaranteed to be `false` and it was written out for explicitness.

amcasey: But that's wrong too - I believe it's pulled out in the if-let, which I don't think applies in the else branch?

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: Convergent evolution.  I'll clean it up.

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: `F(((1,)))` will use multiple iterations of unwrapping.  Added `ty_mismatch_span_tuple1_to_given_extra_parens`.

minestarks (approved):

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: To be continued

amcasey: https://github.com/microsoft/qdk/pull/3307

PR #3288, opened 2026-06-03
[WIP] Code fix for single value passed to array parameter

amcasey:
I don't love this approach, but I think this is a useful starting point for a discussion of how the frontend work.

Context: I always forget to wrap control qubits in an array, so I want to add a code fix for that case.


amcasey:
Superseded by #3330.

PR #3290, opened 2026-06-04
Instance and trace parameters for CirqApplication

msoeken:
This allows to pass all parameters to `trace_from_cirq` from a `CirqApplication`. Before this was only possible for `classical_control_probability`.

`classical_control_probability` is an instance parameter, as well as `rotation_threshold`. The remaining parameter `track_memory_qubits` is a hyper parameter for resource estimation exploration.

For now `track_memory_qubits` only iterates of `[True]`, thereby not changing any of the current resource estimation tasks. We need better control of the trace such that we can skip enumerating over both truth values in case there are no memory qubits in the input circuit. To enable passing `False` or both values one can call `.q` on the app and extend the trace query, e.g.:

```python
app = CirqApplication(circuit, classical_control_probability=..., rotation_threshold=...)
trace_query = app.q(track_memory_qubits=[True, False]) * PSSPC.q() * ...
```

### DIFF
--- a/source/qdk_package/qdk/qre/application/_cirq.py
+++ b/source/qdk_package/qdk/qre/application/_cirq.py
@@ -45,12 +45,12 @@ class CirqApplication(Application[CirqApplicationParams]):
             this threshold are treated as identity and omitted from the
             trace. This applies to single-qubit rotations (RX, RY, RZ) as
             well as to the rotation components of controlled-Z
-            decompositions. Defaults to 1e-6.
+            decompositions. Defaults to 1e-12.
     """
 
     circuit_or_qasm: str | cirq.CIRCUIT_LIKE
     classical_control_probability: float = 0.5
-    rotation_threshold: float = 1e-6
+    rotation_threshold: float = 1e-12
 
     def __post_init__(self):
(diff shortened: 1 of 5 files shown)


source/qdk_package/qdk/qre/application/_cirq.py
```diff
@@ -26,10 +41,16 @@ class CirqApplication(Application[None]):
...
+        rotation_threshold: Rotation exponents with absolute value below
+            this threshold are treated as identity and omitted from the
+            trace. This applies to single-qubit rotations (RX, RY, RZ) as
+            well as to the rotation components of controlled-Z
+            decompositions. Defaults to 1e-6.
     """
 
     circuit_or_qasm: str | cirq.CIRCUIT_LIKE
     classical_control_probability: float = 0.5
+    rotation_threshold: float = 1e-6
```
brad-lackey: Is this sufficient? I would think this should be tied to the rotation synthesis error rate (although that is computed lower in the stack).

msoeken: This has been changed to `1e-12`.

brad-lackey (approved):

PR #3291, opened 2026-06-04
feat(python): load visual circuits directly in Context

### LINKED ISSUE #3232: Allow running visual circuit files (.qsc) directly from Python
Image

## Problem

`.qsc` files are JSON-based circuit definition files (visual circuit format) that describe quantum circuits with gates, measurements, qubits, etc. Currently, to use a `.qsc` circuit from Python, you must:

1. Create a qsharp.json manifest file
2. Write a `.qs` wrapper file that imports the circuit (`import JointMeasurement.JointMeasurement;`)
3. Call `qdk.Context(project_root=...)` to load the full project
4. Then `.eval(...)` or `.run(...)` through the Q# wrapper

This is a lot of ceremony for what should be a simple operation — especially for users who build circuits visually in VS Code and just want to simulate them from Python.

## Proposed API

Add a `load_circuit` method to `qdk.Context` that accepts a path to a `.qsc` file and returns a callable:

```python
import qdk

ctx = qdk.Context()
circuit = ctx.load_circuit("path/to/JointMeasurement.qsc")

# Run the circuit
results = ctx.run(circuit, shots=100)

# Or get a circuit diagram
diagram = ctx.circuit(circuit)
```

### Signature

```python
def load_circuit(self, path: str) -> Callable:
    """
    Loads a .qsc circuit file into this context and returns a callable
    representing the circuit operation.

    :param path: Path to a .qsc file.
    :return: A callable that can be passed to ctx.run(), ctx.circuit(), etc.
    :raises QSharpError: If the file cannot be read or is not a valid .qsc circuit.
    """
```

### Behavior

- Reads and validates the `.qsc` JSON file
- Registers the circuit in the interpreter (similar to how project loading handles `.qsc` files today)
- Returns a callable that behaves like any other Q# callable in the context (passable to `run`, `compile`, `circuit`, etc.)
- The circuit's qubit requirements are inferred from the `.qsc` file (the format already declares qubits)

### Open Questions

- **Naming**: Should the callable's name be derived from the filename (e.g., JointMeasurement.qsc → `ctx.code.JointMeasurement`)? Or only returned, not registered on `ctx.code`?
- **Multiple circuits**: A `.qsc` file can contain multiple circuits in its `"circuits"` array. Should `load_circuit` load just the first one, or should it return multiple callables / accept an index?
- **Qubit allocation**: Circuits in `.qsc` declare their qubit count. Should the returned callable auto-allocate qubits internally, or require the caller to pass qubits (like the current Q# wrapper pattern)?

## Existing Code Pointers

For contributors looking to implement this:

- `.qsc` loading in project system: project.rs (already globs for `*.qsc` files)
- `.qsc` preprocessing in the frontend: preprocess.rs
- Circuit format/schema: circuit.rs
- Python `Context` class: _context.py
- Existing test using `.qsc` in a project: test_project.py (the `"circuit"` memfs entry)
- Sample project using `.qsc`: circuit_integration

tzh476:
Fixes #3232.

## Summary

- add `Context.import_circuit(path, *, index=0, name=None, program_type=ProgramType.File)` for standalone `.qsc` visual-circuit files
- support selecting a circuit from multi-circuit `.qsc` files with `index=0` by default
- default `ProgramType.File` imports a single zero-argument Q# operation named from the file stem, or from the optional `name=` override, suitable for `ctx.run(...)` and `ctx.circuit(...)`
- support `ProgramType.Operation` for registering the qubit-taking circuit operation directly so it can be composed from Q# entry expressions
- generate the standalone File-mode Q# operation in the native Rust layer via `qsc_circuit`, avoiding Python-side helper/wrapper cleanup and avoiding extra public helper callables in `qdk.code`
- re-export `ProgramType` from `qdk` for the new import mode selection
- cover direct visual-circuit import, multi-circuit selection, Operation mode, and explicit name override in Python tests; add native `qsc_circuit` tests for standalone Q# generation

## AI Assistance Disclosure

I used Codex to help inspect the existing QDK visual-circuit conversion path, draft the Python/Rust integration changes, respond to review feedback, and prepare tests. I manually reviewed the patch and verified it locally with the checks below.

## Testing

Latest post-rebase validation on `66e996fe5`:

```bash
./build.py --qdk --test --no-check-prereqs --no-check
```

Result: the QDK Python wheel built successfully, and the Python package test suite completed with `1401 passed, 84 skipped, 1 xfailed`.

Additional targeted checks run during the native Rust-layer refactor:

```bash
RUSTUP_HOME=/Users/bytedance/Library/Caches/puccinialin/rustup CARGO_HOME=/Users/bytedance/Library/Caches/puccinialin/cargo cargo fmt --check --all
cargo test -p qsc_circuit circuit_to_qsharp
source/qdk_package/.venv/bin/python -m pytest source/qdk_package/tests/test_project.py -k import_circuit -q
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/qdk/__init__.py source/qdk_package/qdk/_native.pyi source/qdk_package/tests/test_project.py
git diff --check
```

### DIFF
--- a/source/qdk_package/qdk/_context.py
+++ b/source/qdk_package/qdk/_context.py
@@ -92,7 +92,7 @@ def _path_stem(path: str) -> str:
     return stem or "circuit"
 
 
-def _visual_circuit_signature(circuit_json: str, index: int) -> tuple[int, str, int]:
+def _visual_circuit_count(circuit_json: str) -> int:
     try:
         data = json.loads(circuit_json)
     except json.JSONDecodeError as err:
@@ -101,40 +101,7 @@ def _visual_circuit_signature(circuit_json: str, index: int) -> tuple[int, str, 
     circuits = data.get("circuits") if isinstance(data, dict) else None
     if not isinstance(circuits, list) or len(circuits) == 0:
         raise QSharpError("Visual circuit files must contain at least one circuit.")
-
-    if not isinstance(index, int) or isinstance(index, bool):
-        raise QSharpError("Visual circuit index must be an integer.")
-    if index < 0 or index >= len(circuits):
-        raise QSharpError(
-            f"Visual circuit index {index} is out of range for {len(circuits)} circuits."
-        )
-
-    circuit = circuits[index]
-    if not isinstance(circuit, dict):
-        raise QSharpError("Visual circuit JSON contains an invalid circuit.")
-
-    qubits = circuit.get("qubits", [])
-    if not isinstance(qubits, list):
-        raise QSharpError("Visual circuit JSON contains an invalid qubit register.")
-
-    result_count = 0
-    for qubit in qubits:
-        if not isinstance(qubit, dict):
-            raise QSharpError("Visual circuit JSON contains an invalid qubit.")
-        num_results = qubit.get("numResults", 0)
-        if (
-            not isinstance(num_results, int)
-            or isinstance(num_results, bool)
-            or num_results < 0
-        ):
-            raise QSharpError("Visual circuit JSON contains an invalid result count.")
-        result_count += num_results
-
-    if result_count == 0:
-        return len(qubits), "Unit", len(circuits)
-    if result_count == 1:
-        return len(qubits), "Result", len(circuits)
-    return len(qubits), "Result[]", len(circuits)
+    return len(circuits)
 
 
 def make_class_rec(qsharp_type: TypeIR) -> type:
@@ -562,7 +529,7 @@ class Context:
         except Exception as err:
             raise QSharpError(f"Error reading visual circuit file {resolved_path}.") from err
 
-        _, _, circuit_count = _visual_circuit_signature(circuit_json, index)
+        circuit_count = _visual_circuit_count(circuit_json)
         operation_base_name = name if name is not None else _path_stem(resolved_path)
 
         operation_name, generated_source = compile_visual_circuit_to_qsharp(
--- a/source/qdk_package/qdk/_context.py
+++ b/source/qdk_package/qdk/_context.py
@@ -138,8 +138,13 @@ def _visual_circuit_signature(circuit_json: str, index: int) -> tuple[int, str, 
 
 
 def _visual_circuit_wrapper_source(
-    operation_name: str, wrapper_name: str, qubit_count: int, return_type: str
+    helper_source: str,
+    operation_name: str,
+    wrapper_name: str,
+    qubit_count: int,
+    return_type: str,
 ) -> str:
+    helper_source = _indent_qsharp_source(helper_source.strip() + "\n\n")
     if qubit_count == 0:
         allocation = ""
         call = f"{operation_name}()"
@@ -152,6 +157,7 @@ def _visual_circuit_wrapper_source(
     if return_type == "Unit":
         return (
             f"operation {wrapper_name}() : Unit {{\n"
+            f"{helper_source}"
             f"{allocation}"
             f"    {call};\n"
             f"{reset}"
@@ -160,6 +166,7 @@ def _visual_circuit_wrapper_source(
 
     return (
         f"operation {wrapper_name}() : {return_type} {{\n"
+        f"{helper_source}"
         f"{allocation}"
         f"    let result = {call};\n"
         f"{reset}"
@@ -168,7 +175,14 @@ def _visual_circuit_wrapper_source(
     )
 
 
-def _mark_visual_circuit_operations_internal(
+def _indent_qsharp_source(source: str) -> str:
+    return "".join(
+        f"    {line}" if line.strip() else line
+        for line in source.splitlines(keepends=True)
+    )
+
+
+def _rename_visual_circuit_operations(
     source: str, operation_name: str, circuit_count: int
 ) -> str:
     for index in range(circuit_count):
@@ -177,7 +191,7 @@ def _mark_visual_circuit_operations_internal(
         )
         source = source.replace(
             f"operation {callable_name}(",
-            f"internal operation {callable_name}_Operation(",
+            f"operation {callable_name}_Operation(",
         )
     return source
 
@@ -620,32 +634,24 @@ class Context:
         )
         eval_source = generated_source
         callable_name = circuit_operation_name
-        helper_callable_names: list[str] = []
         if program_type == ProgramType.File:
             circuit_helper_name = f"{circuit_operation_name}_Operation"
-            helper_callable_names = [
-                (
-                    f"{operation_name}_Operation"
-                    if circuit_count == 1
-                    else f"{operation_name}{helper_index}_Operation"
-                )
-                for helper_index in range(circuit_count)
-            ]
-            generated_source = _mark_visual_circuit_operations_internal(
+            generated_source = _rename_visual_circuit_operations(
                 generated_source, operation_name, circuit_count
             )
             wrapper_source = _visual_circuit_wrapper_source(
-                circuit_helper_name, circuit_operation_name, qubit_count, return_type
+                generated_source,
+                circuit_helper_name,
+                circuit_operation_name,
+                qubit_count,
+                return_type,
             )
-            eval_source = f"{generated_source}\n{wrapper_source}"
+            eval_source = wrapper_source
 
         # generated_source is produced by the native visual-circuit compiler;
         # wrapper_source, when present, is assembled from sanitized operation
         # names and type metadata.
         self.eval(eval_source)  # DevSkim: ignore DS189424
-        for helper_callable_name in helper_callable_names:
-            if hasattr(self.code, helper_callable_name):
-                delattr(self.code, helper_callable_name)
         return getattr(self.code, callable_name)
 
     def eval(
--- a/source/qdk_package/qdk/_context.py
+++ b/source/qdk_package/qdk/_context.py
@@ -240,7 +240,6 @@ class Context:
     code: Any
     _disposed: bool
     _is_global_context: bool
-    _loaded_circuit_count: int
 
     def __init__(
         self,
@@ -276,7 +275,6 @@ class Context:
         self._disposed = False
         self._is_global_context = _is_global_context
         self._target_profile = target_profile
-        self._loaded_circuit_count = 0
 
         if _is_global_context:
             self.code = code
@@ -559,6 +557,7 @@ class Context:
         path: str,
         *,
         index: int = 0,
+        name: Optional[str] = None,
         program_type: ProgramType = ProgramType.File,
     ) -> Callable:
         """
@@ -567,6 +566,8 @@ class Context:
         :param path: Path to a ``.qsc`` visual circuit file.
         :keyword index: Index of the circuit to return when the file contains
             multiple circuits. Defaults to ``0``.
+        :keyword name: Optional Q# operation name to use for the imported
+            circuit. Defaults to the file name stem.
         :keyword program_type: Controls how the circuit is imported:
             ``ProgramType.File`` (default) imports a zero-argument wrapper that
             allocates the circuit qubits, applies the visual circuit, resets the
@@ -595,11 +596,10 @@ class Context:
         qubit_count, return_type, circuit_count = _visual_circuit_signature(
             circuit_json, index
         )
-        unique_name = f"{_path_stem(resolved_path)}_{self._loaded_circuit_count}"
-        self._loaded_circuit_count += 1
+        operation_base_name = name if name is not None else _path_stem(resolved_path)
 
         operation_name, generated_source = compile_visual_circuit_to_qsharp(
-            unique_name, circuit_json
+            operation_base_name, circuit_json
         )
         circuit_operation_name = (
             operation_name if circuit_count == 1 else f"{operation_name}{index}"
--- a/source/qdk_package/qdk/_context.py
+++ b/source/qdk_package/qdk/_context.py
@@ -137,65 +137,6 @@ def _visual_circuit_signature(circuit_json: str, index: int) -> tuple[int, str, 
     return len(qubits), "Result[]", len(circuits)
 
 
-def _visual_circuit_wrapper_source(
-    helper_source: str,
-    operation_name: str,
-    wrapper_name: str,
-    qubit_count: int,
-    return_type: str,
-) -> str:
-    helper_source = _indent_qsharp_source(helper_source.strip() + "\n\n")
-    if qubit_count == 0:
-        allocation = ""
-        call = f"{operation_name}()"
-        reset = ""
-    else:
-        allocation = f"    use qs = Qubit[{qubit_count}];\n"
-        call = f"{operation_name}(qs)"
-        reset = "    ResetAll(qs);\n"
-
-    if return_type == "Unit":
-        return (
-            f"operation {wrapper_name}() : Unit {{\n"
-            f"{helper_source}"
-            f"{allocation}"
-            f"    {call};\n"
-            f"{reset}"
-            "}\n"
-        )
-
-    return (
-        f"operation {wrapper_name}() : {return_type} {{\n"
-        f"{helper_source}"
-        f"{allocation}"
-        f"    let result = {call};\n"
-        f"{reset}"
-        "    return result;\n"
-        "}\n"
-    )
-
-
-def _indent_qsharp_source(source: str) -> str:
-    return "".join(
-        f"    {line}" if line.strip() else line
-        for line in source.splitlines(keepends=True)
-    )
-
-
-def _rename_visual_circuit_operations(
-    source: str, operation_name: str, circuit_count: int
-) -> str:
-    for index in range(circuit_count):
-        callable_name = (
-            operation_name if circuit_count == 1 else f"{operation_name}{index}"
-        )
-        source = source.replace(
-            f"operation {callable_name}(",
-            f"operation {callable_name}_Operation(",
-        )
-    return source
-
-
 def make_class_rec(qsharp_type: TypeIR) -> type:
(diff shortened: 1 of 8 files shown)


tzh476:

tzh476:
Hi @billti @idavis @minestarks, I have signed the CLA. Could you please approve the pending workflows and review when convenient? Thanks!

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +553,47 @@
...
+        )
+        circuit_operation_name = (
+            operation_name if circuit_count == 1 else f"{operation_name}{index}"
+        )
+        wrapper_name = f"{circuit_operation_name}_Entry"
+        wrapper_source = _visual_circuit_wrapper_source(
+            circuit_operation_name, wrapper_name, qubit_count, return_type
+        )
+
+        self.eval(f"{generated_source}\n{wrapper_source}")
```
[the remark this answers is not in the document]
tzh476: Thanks for the scan. This eval path is intentionally compiling Q# generated by the native visual-circuit compiler plus a local wrapper assembled from sanitized operation names and type metadata. I made that boundary explicit, moved the source into a named variable, and added the DevSkim suppression for DS189424 on the evaluated line in c53b195. I also ran `git diff --check` and `python3 -m py_compile source/qdk_package/qdk/_context.py`; the local Python environment does not have pytest installed, so I could not rerun the targeted pytest locally.

tzh476:
I addressed the GitHub Advanced Security / DevSkim review in `c53b195` by documenting the generated-source boundary and adding the targeted `DS189424` suppression on the vetted Q# eval line. The review thread is resolved now. Could you please approve the refreshed workflows when convenient? Thanks!

tzh476:
All refreshed checks are passing now, including `Build and test`, `Rust Unit tests`, `Integration tests`, `DevSkim`, and the CLA check. The prior GitHub Advanced Security / DevSkim review thread is resolved as well.

Could you please review this when convenient? Thanks!

minestarks:
Thank you @tzh476  ! I'm taking a look now - will post comments shortly.

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +553,50 @@ def _check_same_context_struct(self, struct: Any) -> None:
         if getattr(struct_type, "_qdk_context") is not self:
             raise QSharpError("This struct belongs to a different Context. ")
 
+    def load_circuit(self, path: str, *, index: int = 0) -> Callable:
```
minestarks: After discussing this within the team, we landed on `import_circuit` as a better name for the public API. (It mirrors `import_openqasm` , which has more or less the same semantics). Would you mind renaming it?

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +553,50 @@ def _check_same_context_struct(self, struct: Any) -> None:
...
+        """
```
minestarks, line 499: Also after team discussion: it would be valuable for this public API to take a `program_type` parameter (`ProgramType.File` | `ProgramType.Operation`) to mirror how `import_openqasm` treats its input. The `ProgramType` enum should be accessible here via `._native` 

- **File** (current behavior): treats the circuit as a standalone program, the returned callable takes no arguments and can be invoked directly from Python. 

- **Operation**: would skip the wrapper and register the circuit operation itself (which takes a `Qubit[]` parameter). This is useful for composing circuits inside larger Q# programs. However, since qubits have no Python representation, the callable could only be invoked via a string entry expression (e.g. `ctx.run("{ use qs = Qubit[N]; myCircuit(qs) }", shots)`), not via `ctx.code.myCircuit(...)`. You can add a unit test for this to validate behavior. See how `import_openqasm` with `ProgramType.Operation` works for the precedent. 

The default should be `File` if no program type is passed.

Sorry if this seems like scope creep - when we filed the issue, we hadn't yet ironed out all open questions, but now that we see the API all fleshed out, it became clear that having both options would be useful.

Thank you!

minestarks (changes requested):
Really nice contribution, thank you! Just a couple of requests, please.

tzh476:
Thanks for the review. I pushed `cc107d855` to address both requests:

- renamed the visual circuit API to `import_circuit`;
- added `program_type` with `ProgramType.File` as the default wrapper behavior and `ProgramType.Operation` for registering the circuit operation itself for Q# entry-expression composition;
- updated the existing visual-circuit tests and added an Operation-mode test using a string entry expression.

Local validation:

```bash
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/tests/test_project.py
git diff --check
```

I attempted the targeted pytest locally, but this checkout cannot resolve the repository-local `qsharp==0.0.0` dependency through `uv` outside the full build environment, so I am relying on CI for the native-backed test run.

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +554,72 @@ def _check_same_context_struct(self, struct: Any) -> None:
...
+        try:
+            _, circuit_json = read_file(resolved_path)
+        except Exception as err:
+            raise QSharpError(f"Error reading visual circuit file {resolved_path}.") from err
+
+        qubit_count, return_type, circuit_count = _visual_circuit_signature(
+            circuit_json, index
+        )
+        unique_name = f"{_path_stem(resolved_path)}_{self._loaded_circuit_count}"
+        self._loaded_circuit_count += 1
```
minestarks: Instead of the `_loaded_circuit_count` disambiguator, can we simply go with the path stem? So `JointMeasurement.qsc` would end up with `JointMeasurement`. We can take an optional `name` argument to override it.

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +554,72 @@ def _check_same_context_struct(self, struct: Any) -> None:
         if getattr(struct_type, "_qdk_context") is not self:
             raise QSharpError("This struct belongs to a different Context. ")
 
+    def import_circuit(
+        self,
+        path: str,
+        *,
+        index: int = 0,
+        program_type: ProgramType = ProgramType.File,
```
minestarks, line 497: We will need to re-export `ProgramType` from the `qdk` package so the user can actually import this type `from qdk import ProgramType`  - without going through any private modules (such as `._native`) . Could you add it to the public exports in source/qdk_package/qdk/__init__.py  ?

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +554,72 @@ def _check_same_context_struct(self, struct: Any) -> None:
...
+        circuit_operation_name = (
+            operation_name if circuit_count == 1 else f"{operation_name}{index}"
+        )
+        eval_source = generated_source
+        callable_name = circuit_operation_name
+        if program_type == ProgramType.File:
+            wrapper_name = f"{circuit_operation_name}_Entry"
+            wrapper_source = _visual_circuit_wrapper_source(
+                circuit_operation_name, wrapper_name, qubit_count, return_type
+            )
```
minestarks: Question to @swernli  - so this results in two callables - `Foo` and `Foo_Entry` . When we discussed in person, I hadn't considered that we'd end up with two callables. I actually like it this way - having the inner operation that takes `Qubit`s continue to be called `Foo`. Thoughts?

minestarks:
@tzh476  thank you for the edits! It's very close! I've been playing around with it on my machine, and found a couple of minor user-facing API/behavior concerns (see comments). Thank you!

tzh476:
Thanks for the follow-up. I pushed `e030fde0b` to address the two actionable items:

- `import_circuit` now defaults the generated operation base name to the file stem and accepts an optional `name=` override.
- `ProgramType` is now re-exported from the `qdk` package root, and the visual-circuit operation test imports it with `from qdk import ProgramType`.

I also updated the operation-mode expectation from `circuit_0` to `circuit`, and added a small `name=` override test.

Local validation:

```bash
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/qdk/__init__.py source/qdk_package/tests/test_project.py
git diff --check
```

The local checkout here still does not have the built `qdk._native` extension / pytest environment available, so I am relying on CI for the native-backed test execution.

source/qdk_package/qdk/_context.py
swernli: Hmm... I could see how this would be convenient for advanced users but worry that creating two similarly named callables might make this more confusing for average users. I think matching the behavior of how `import_openqasm` behaves with regard to `ProgramType` is easier to explain, where `File` creates one standalone callable that doesn't take qubit params and `Operation` creates the callable that takes qubits and can be used along with other code.

tzh476:
Thanks for the clarification from @swernli. I pushed `3726daec4` to align the visual circuit import behavior with `import_openqasm` semantics:

- `ProgramType.File` now returns the standalone callable using the requested/default name, without the extra public `_Entry` callable.
- The generated qubit-taking circuit operation is marked internal and used only as the File-mode wrapper implementation detail.
- `ProgramType.Operation` still returns the qubit-taking operation directly for composition from Q# entry expressions.
- Added tests covering the public callable names and absence of the old `_Entry` / helper public callables for File mode.

Local validation passed:

```bash
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/qdk/__init__.py source/qdk_package/tests/test_project.py
git diff --check
```

The native-backed targeted pytest needs the repository build/test environment, so I am relying on the refreshed GitHub CI run for that coverage. Could you please approve the refreshed workflows and review when convenient? Thanks!

minestarks:
@tzh476  hmm, the CI checks failed. You can run the same CI checks by just running `./build.py` locally .

The problem seems to be that marking the operation `internal` doesn't actually hide it from `qdk.code` module.

Let's try to address it deeper within the native code / Rust layer. We can refactor things a bit. Can we avoid generating 2 operations altogether, and have just ONE named `Foo`? (No suffixes, not `Foo_Operation` etc) . 

It seems that the solution would be to refactor `build_operation_def` in the `qsc_circuit` crate into two functions: one that generates the body block of the operation, and one that generates the signature + length check. Then, when we want to make a stand-alone operation that doesn't take qubits: it can use a new function that generates the same body block, but uses that as an expression inside a wrapper that allocates the qubits instead of taking them as an argument. That should result in only one Q# operation ever being defined.

See unit tests at `source/compiler/qsc_circuit/src/circuit_to_qsharp/tests.rs` to understand behavior of  `build_operation_def`

tzh476:
I saw the refreshed `Build and test` CI failed because the helper operations were still exposed through `ctx.code` even after being marked `internal`. I pushed `7f6f73a2a` to keep those helper callables available inside the interpreter for the File-mode wrapper, but remove them from the Python `ctx.code` namespace after evaluation.

Local validation passed again:

```bash
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/qdk/__init__.py source/qdk_package/tests/test_project.py
git diff --check
```

This should address the two failing assertions from the previous CI run while preserving the public File/Operation behavior discussed above.

swernli:
Actually, I realized a simple approach you could use to effectively make the operation internal without having to modify the `code` module after the fact: define the operation inside of the other operation. This restricts the visibility to within the wrapper. So when the inner operation is defined alongside the wrapper, both are visible:

image

but when the inner operation is defined inside of the wrapper, only the wrapper will be visible:

image

So rather than putting the two operations next to each other, put one inside of the other.

tzh476:
Superseded by the native Rust-layer follow-up in `87c37b709` below. The previous note about missing local `pytest` is no longer current; I set up the package venv and ran the repository build/test flow locally.

tzh476:
Thanks for the detailed guidance. I replaced the Python-side wrapper/helper approach with a native Rust-layer standalone path in `87c37b709`.

What changed:

- `qsc_circuit` now has `circuit_to_standalone_qsharp(file_name, circuits_json, index)` for File mode. It builds only the selected circuit as one zero-argument operation named exactly `Foo`.
- `build_operation_def` now shares the circuit body generation with the standalone path, so File mode allocates/resets qubits inside the single generated operation instead of defining helper operations.
- `ProgramType.Operation` still uses the existing qubit-array operation generation for Q# composition.
- Added `qsc_circuit` tests for standalone visual circuit generation and updated the Python import tests.

Local validation passed:

```bash
./build.py --qdk --test --no-check-prereqs --no-check
RUSTUP_HOME=/Users/bytedance/Library/Caches/puccinialin/rustup CARGO_HOME=/Users/bytedance/Library/Caches/puccinialin/cargo cargo fmt --check --all
cargo test -p qsc_circuit circuit_to_qsharp
source/qdk_package/.venv/bin/python -m pytest source/qdk_package/tests/test_project.py -k import_circuit -q
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/qdk/__init__.py source/qdk_package/qdk/_native.pyi source/qdk_package/tests/test_project.py
git diff --check
```

The linked failed `Status Check` job was the aggregator; the underlying `Build and test` failure was caused by the exposed helper operations. This update removes that helper-operation surface instead of trying to hide it from `qdk.code`.

Could you please approve the refreshed workflows and review when convenient? Thanks!

source/qdk_package/qdk/_context.py
```diff
@@ -466,6 +521,63 @@ def _check_same_context_struct(self, struct: Any) -> None:
...
+                "ProgramType.Operation only."
+            )
+
+        resolved_path = resolve(".", path)
+        try:
+            _, circuit_json = read_file(resolved_path)
+        except Exception as err:
+            raise QSharpError(f"Error reading visual circuit file {resolved_path}.") from err
+
+        _, _, circuit_count = _visual_circuit_signature(circuit_json, index)
```
minestarks: If we're just using `circuit_count` from this now, this method can be drastically simplified or removed, right?

source/qdk_package/tests/test_project.py
```diff
@@ -80,6 +82,60 @@ def test_circuit(qsharp) -> None:
...
 
 
+def test_import_circuit(qsharp) -> None:
+    import qdk
+
+    ctx = qdk.Context()
+    circuit = ctx.import_circuit("/standalone/circuit.qsc")
+    assert circuit is ctx.code.circuit
+    assert not hasattr(ctx.code, "circuit_Entry")
+    assert not hasattr(ctx.code, "circuit_Operation")
```
minestarks: I think these `not hasattr`s are unnecessary now - nothing exists anymore in the code that refer to these names, so they seeem sort of out-of-the-blue. Can be removed.

source/qdk_package/tests/test_project.py
```diff
@@ -80,6 +82,60 @@ def test_circuit(qsharp) -> None:
...
+
+
+def test_import_circuit_with_name_override(qsharp) -> None:
+    import qdk
+
+    ctx = qdk.Context()
+    circuit = ctx.import_circuit("/standalone/circuit.qsc", name="NamedCircuit")
+    assert circuit is ctx.code.NamedCircuit
+    assert not hasattr(ctx.code, "NamedCircuit_Entry")
+    assert not hasattr(ctx.code, "NamedCircuit_Operation")
```
minestarks: Same comment as above - I don't think we need to check that these attributes don't exist anymore

tzh476:
Quick status update after the native Rust-layer refactor in `87c37b709`: the branch now generates a single standalone Q# operation for File mode instead of Python-side helper/wrapper cleanup, matching the direction discussed above. The latest visible check is only the CLA check, so I believe the refreshed workflows still need approval/re-run for the new commit. Could you approve the workflows and re-review when convenient? Thanks!

tzh476:
Follow-up: I rebased the branch onto the latest `microsoft/qdk:main` and force-pushed the updated head `66e996fe5`. The PR diff is still limited to the same visual-circuit import files.

Post-rebase local validation now passes with the repository package build/test flow:

```bash
./build.py --qdk --test --no-check-prereqs --no-check
```

Result: the QDK Python wheel built successfully, and the Python package test suite completed with `1401 passed, 84 skipped, 1 xfailed`.

The only currently visible check is still the CLA check, so the refreshed workflows appear to need approval/re-run for the rebased head as well.

minestarks (changes requested):
Tested. One last round of minor cleanup comments - I think the user-facing API and behavior are all there. Thank you!

source/qdk_package/tests/test_project.py
tzh476: Removed in `35e466d77`.

source/qdk_package/qdk/_context.py
tzh476: Addressed in `35e466d77`. I simplified this to `_visual_circuit_count`, which only parses the top-level `circuits` list now. I kept the count because `ProgramType.Operation` with a multi-circuit file still needs to return the generated `Name{index}` callable, and added a regression test for that case.

source/qdk_package/tests/test_project.py
tzh476: Removed in `35e466d77`.

tzh476:
Pushed `35e466d77` to address the final cleanup review comments.

Changes:

- Reduced the old visual-circuit signature helper to `_visual_circuit_count`, which only reads the top-level circuit count needed for multi-circuit `ProgramType.Operation` callable selection.
- Removed the now-unnecessary `not hasattr` assertions for old helper names.
- Added a regression test for importing a selected operation from a multi-circuit visual circuit file.

Local validation passed:

```bash
python3 -m py_compile source/qdk_package/qdk/_context.py source/qdk_package/tests/test_project.py
git diff --check
source/qdk_package/.venv/bin/python -m pytest source/qdk_package/tests/test_project.py -k import_circuit -q
./build.py --qdk --test --no-check-prereqs --no-check
```

Full build/test result: `1402 passed, 84 skipped, 1 xfailed`.

I also replied to and resolved the three review threads.

minestarks (approved):

PR #3294, opened 2026-06-04
Add auto-import quickfix for unresolved names

### LINKED ISSUE #3236: Add auto-import quickfix for unresolved names
### Summary

When a user writes an unresolved name like `DumpMachine()`, the compiler emits a `Qsc.Resolve.NotFound` error. The language service should offer a quickfix code action that inserts the appropriate `import` statement — e.g., `import Std.Diagnostics.DumpMachine;`.

### Current behavior

```qsharp
namespace Sample {
    operation Main() : Unit {
        DumpMachine(); // ❌ Error: `DumpMachine` not found
    }
}
```

The user sees the error but gets no suggested fix.

### Desired behavior

The lightbulb menu (code action) on the error should offer:

> **Import `Std.Diagnostics.DumpMachine`**

Accepting inserts the import at the top of the namespace block:

```qsharp
namespace Sample {
    import Std.Diagnostics.DumpMachine;
    operation Main() : Unit {
        DumpMachine();
    }
}
```

If multiple items match the name (different namespaces), each should be offered as a separate quickfix option.

### Implementation guidance

The infrastructure for this already exists — the main work is wiring it together in a new code path.

**Key files:**

| File | Role |
|------|------|
| code_action.rs | Code action entry point. Currently only handles `ErrorKind::Lint`. Add a branch for `ErrorKind::Frontend` resolve errors. |
| resolve.rs (~L200) | Defines `Error::NotFound(String, Span)` — the unresolved name and its location. |
| global_items.rs | Already iterates all items across all namespaces via `GlobalScope::table()`. Reuse this pattern to find items matching the unresolved name. |
| text_edits.rs | `TextEditRange` already computes where to insert import statements (namespace block start or top of notebook cell). |

**Suggested approach:**

1. In code_action.rs, add a new function (e.g., `auto_import_fixes()`) called from `get_code_actions()`.
2. Filter `compilation.compile_errors` for `ErrorKind::Frontend` errors with diagnostic code `Qsc.Resolve.NotFound` that overlap the requested range.
3. Extract the unresolved name from the error.
4. Search `GlobalScope::table(NameKind::Term)` and `table(NameKind::Ty)` for items whose name matches.
5. For each match, compute the import statement text (`import {namespace}.{name};`) using `GlobalScope::format_namespace_name()`.
6. Compute the insertion point using the same logic as `TextEditRange` in the completions module.
7. Return a `CodeAction` with kind `QuickFix` for each candidate.

**Existing pattern to follow:** The `Availability::NeedImport` path in global_items.rs (around L435) already formats import statements and uses `TextEditRange::insert_import_at` — this is essentially the same edit.

### Scope / edge cases

- **v1 scope**: Unqualified names only (e.g., `DumpMachine`). Partial paths like `Diagnostics.DumpMachine` can be out of scope.
- **Multiple matches**: If `Foo` exists in both `Std.A` and `Std.B`, offer both as separate code actions.
- **Notebooks**: Import insertion already works differently for notebook cells via `TextEditRange` — no extra handling needed.
- **No WASM/VS Code/Playground changes needed**: The code action plumbing should already handle arbitrary `CodeAction` objects end-to-end.

sorin-bolos:
Fixes #3236 

* Unqualified names only
* Works for notebooks too
* Multiple matches offer separate suggestions

image

Disclosure: Used Claude Code (Opus 4.8)


sorin-bolos:

minestarks:
@sorin-bolos thank you for the PR! Taking a look shortly at this one as well.

source/language_service/src/code_action.rs
```diff
@@ -36,6 +52,111 @@ pub(crate) fn get_code_actions(
     actions
 }
 
+/// Produces auto-import quick fixes for unresolved names within `span`.
+///
+/// For each `Qsc.Resolve.NotFound` diagnostic that overlaps the requested range, the
+/// unresolved (unqualified) name is looked up in the global term and type tables. Every
+/// namespace that exports a matching name yields a separate `QuickFix` code action that
+/// inserts an `import {namespace}.{name};` statement at the start of the enclosing namespace.
+fn auto_import_fixes(
```
minestarks: Can we pull this code action into its own sub-module under `code_action`, like `wrapper_refactor`? 

And then the corresponding unit tests can also live in `code_action/auto_import/tests.rs`

source/language_service/src/code_action.rs
```diff
@@ -36,6 +52,111 @@ pub(crate) fn get_code_actions(
...
+    let mut seen = FxHashSet::default();
+
+    let not_found_errors = compilation
+        .compile_errors
+        .iter()
+        .filter(|error| is_error_relevant(error, span))
+        .filter(|error| {
+            error
+                .code()
+                .is_some_and(|code| code.to_string() == RESOLVE_NOT_FOUND_CODE)
```
minestarks: I'd prefer to match on `ErrorKind` rather than the string code, to keep the code more strongly-typed (codes may change)

source/language_service/src/code_action.rs
```diff
@@ -36,6 +52,111 @@ pub(crate) fn get_code_actions(
...
+        let Some(error_span) = resolve_span(error) else {
+            continue;
+        };
+
+        // Extract the unresolved name from the source text at the error's span.
+        let lo = (error_span.lo - source.offset) as usize;
+        let hi = (error_span.hi - source.offset) as usize;
+        let Some(name) = source.contents.get(lo..hi) else {
+            continue;
+        };
```
minestarks: If you match on `ErrorKind` as suggested above, you probably can get the unresolved name directly from the error object itself, so you wouldn't need to extract it from source like this.

minestarks (changes requested):
Thank you for this! Works great, just a few cleanup requests.

source/language_service/src/code_action.rs
sorin-bolos: moved to a separate module

source/language_service/src/code_action.rs
sorin-bolos: done

source/language_service/src/code_action.rs
sorin-bolos: done

minestarks (approved):

minestarks:
Thank you, @sorin-bolos ! Queued to merge. Could you please leave a comment at #3236 so GitHub will let me assign the issue to you?

PR #3295, opened 2026-06-05
Gray out excluded code

### LINKED ISSUE #1966: Grey out conditionally compiled items in the editor when they don't match the current target profile
In Q# we can use `@Config` attributes to mark a particular item as not being applicable to the current target.

```qsharp
@Config(Adaptive)
operation Foo() : Unit {
}

@Config(not Adaptive)
operation Foo() : Unit {
}
```

It would be nice to report the ranges for any of the code that is NOT included in the current compilation, to the editor, so **that the editor can display them as grayed out.**

VS Code does have support for this, but I'm not sure which property we need to set exactly. My guess is the `Unnecessary` [`DiagnosticTag`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag) (this is from LSP, but the VS Code equivalent should exist)

For reference, this is how Q# looks today in VS Code:

Image

Compare to how conditionally compiled out code looks in Rust in VS Code:

Image

sorin-bolos:
Closes #1966

Grays out code that is excluded via the @Config annotation.
Applies to VS Code and Playground. Does not apply to notebooks

Using Hint severity + the Unnecessary tag means the code is greyed inline without cluttering the Problems panel — matching the Rust behavior shown in the issue

base
unrestricted
image

image

Disclamer: used Claude Code

### DIFF
--- a/source/language_service/src/protocol.rs
+++ b/source/language_service/src/protocol.rs
@@ -30,6 +30,22 @@ pub enum ErrorKind {
     #[error(transparent)]
     #[diagnostic(transparent)]
     DocumentStatus(#[from] DocumentStatusDiagnostic),
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    Unnecessary(#[from] UnnecessaryCodeDiagnostic),
+}
+
+/// Marks a region of code that is excluded from the current compilation, e.g.
+/// an item whose `@Config` attribute does not match the current target profile.
+/// This is reported to the editor as a hint carrying the `Unnecessary`
+/// diagnostic tag so that the excluded code can be displayed as greyed out.
+#[derive(Clone, Debug, Diagnostic, Error)]
+#[error(
+    "this code is not included in the current compilation because it does not apply to the current target profile"
+)]
+#[diagnostic(severity(info), code("Qdk.LanguageServer.UnnecessaryCode"))]
+pub struct UnnecessaryCodeDiagnostic {
+    pub range: Range,
 }
 
 /// Document status is a non-user facing, info-level diagnostic meant for
--- a/source/language_service/src/protocol.rs
+++ b/source/language_service/src/protocol.rs
@@ -43,7 +43,7 @@ pub enum ErrorKind {
 #[error(
     "this code is not included in the current compilation because it does not apply to the current target profile"
 )]
-#[diagnostic(severity(info), code("Qdk.LanguageServer.UnnecessaryCode"))]
+#[diagnostic(severity(info))]
 pub struct UnnecessaryCodeDiagnostic {
     pub range: Range,
 }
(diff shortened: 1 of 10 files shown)


minestarks:
@sorin-bolos thank you for the PR! I'm taking a look shortly.

source/playground/src/editor.tsx
minestarks, line 1: Not sure where exactly the relevant code is, but the Playground emits this as a yellow "diagnostic" below the editor window. We'll probably want to filter out this severity level (just as VS Code does for the Problems view)

image

source/language_service/src/protocol.rs
```diff
@@ -30,6 +30,22 @@ pub enum ErrorKind {
...
+
+/// Marks a region of code that is excluded from the current compilation, e.g.
+/// an item whose `@Config` attribute does not match the current target profile.
+/// This is reported to the editor as a hint carrying the `Unnecessary`
+/// diagnostic tag so that the excluded code can be displayed as greyed out.
+#[derive(Clone, Debug, Diagnostic, Error)]
+#[error(
+    "this code is not included in the current compilation because it does not apply to the current target profile"
+)]
+#[diagnostic(severity(info), code("Qdk.LanguageServer.UnnecessaryCode"))]
```
minestarks: I think "unnecessary" is good to use as an internal name - since that's what VS Code calls it -, but let's pick a different user-facing code (since technically the grayed-out code isn't "unnecessary"). Maybe `Qsc.ConditionalCompilation.Excluded`? 

Or maybe... do we need `code()` at all? Maybe let's just not have a `code()` string :) , since I can't think of a name that I'm 100% happy with.

source/language_service/src/state.rs
```diff
@@ -720,6 +729,29 @@ fn map_errors_to_docs(
...
+    encoding: Encoding,
+    map: &mut FxHashMap<Arc<str>, Vec<ErrorKind>>,
+) {
+    let unit = compilation.user_unit();
+    let source_map = &unit.sources;
+    for &span in &unit.dropped_spans {
+        // Skip any span we can't resolve to a source rather than panicking.
+        let Some(source) = source_map.find_by_offset(span.lo) else {
+            continue;
+        };
```
minestarks: This should technically never happen (it would indicate a bug) so I would in fact rather panic in `into_range`. I think we can remove this check.

minestarks (changes requested):
Thank you! A couple minor comments.

source/playground/src/editor.tsx
sorin-bolos: Filtered out these hints

source/language_service/src/protocol.rs
sorin-bolos: done

source/language_service/src/state.rs
sorin-bolos: done

minestarks (approved):

minestarks:
@sorin-bolos thank you again, and congrats! I queued the PR to merge.

In the meantime could you please leave a comment at issue #1966 so that GitHub will let me assign it to you?

PR #3298, opened 2026-06-08
New `parallel` expression in Q#

### LINKED ISSUE #3274: Design for `parallel` keyword in Q#
This issue builds off of prior exploration in #492 to propose a design for explicit space/time trade-offs in Q# via a new `parallel` keyword.

## Background and Justification
The pattern in Q# for qubit management is to follow use qubit allocation statements (such as `use q = Qubit();`) to declare qubits and have the compiler automatically allocate and release them based on the lifetime of the bound variables. By default, Q# reuses qubits aggresively across different allocations, prioritizing space (circuit width) over time (circuit depth). In practice, this means some patterns will reuse ancilla qubits, such as this example program that performs a parity measurement across pairs of qubits:

```qsharp
operation Main() : Unit {
    use qs = Qubit[4];
    for i in 0..2..Length(qs) - 1 {
        Parity(qs[i], qs[i + 1]);
    }
}

operation Parity(q1 : Qubit, q2 : Qubit) : Unit {
    Measure([PauliZ, PauliZ], [q1, q2]);
}
```

Because the `Parity` function uses joint measurement, it allocates an ancilla qubit, entangles it, measures it, resets and releases it. On the next iteration, the allocation sees that same qubit as available and allocates it again. This results in code that cannot be parallelized, as seen in this circuit diagram of `Main`:

Image

Qubit 4 is reused across the loop iterations, preventing the code from running in parallel during execution on a quantum system. While a program author can achieve more manual qubit usage by explicitly allocating ancillas ahead of time and passing them through (similar to OpenQASM patterns), this does not work well with existing library code that may already embrace the Q# qubit allocation patterns and may require extensive rewrites of code or duplication of library functions with custom signatures and qubit patterns.

Instead, we can introduce a `parallel` keyword that signals to the compiler the user's intention to have code optimized for time (depth) instead of space (width), using more qubits to enable better opportunities for runtime parallelism when the program is compiled for execution on a quantum system. Following the design of modern Q# being expression based, this involves the creation of a new expression of the form `parallel <expr>` that denotes all code invoked during the given expression should follow the time-optimized pattern.

Using the example from above, we can modify the for-loop expression by turning it into a parallel loop:

```qsharp
    parallel for i in 0..2..Length(qs) - 1 {
        Parity(qs[i], qs[i + 1]);
    }
```

Which results in a new circuit diagram that shows each of the two iterations using a separate ancilla qubit, thus allowing for parallel scheduling on hardware:

Image

Furthermore, we can introduce a notion of limited parallelism or limited reuse via another syntax of the form `parallel within <int-expr> <expr>`. This can be useful for places where a balance between space and time is desired. For example, if the above code were changed to use 8 qubits, then 4 ancillas would be required with just `parallel`, like this:

```qsharp
operation Main() : Unit {
    use qs = Qubit[8];
    parallel for i in 0..2..Length(qs) - 1 {
        Parity(qs[i], qs[i + 1]);
    }
}
```

Image

By using `parallel within 2 for i in ...` we use the limit to allow some reuse of the ancilla qubits, and now the program can potentially be executed in two time steps:

Image

## Implementation Details
The `parallel` expression indicates a "deferred release" section of the code. For the duration of the expression, qubits that would be release are instead held in a pool that is released at the end of the expression, preventing their reuse on other allocation during that expression. This applies throughout the expression, including in any calls made in the expression. This allows the user to explicitly denote the places where they would like to opt-in to the deferred release mechanism and trade extra space/width for better time/depth performance.

The `parallel within` expression indicates a "limited reuse" section of the code. This draws on notions from the original work in #492 where that section will defer the release of qubits up to a certain number, after which those qubits become available to the code for reuse.

Both `parallel` and `parallel within` expressions can be used nested within each other, creating separate layers of deferred release qubits and limited reuse pools, allowing for nuanced control over the qubit reuse behaviors without having to modify the signature or allocation code used inside of libraries or common utility code.

### Effect on QIR Generation
When emitting QIR, the `parallel` keyword enforces that the corresponding expression does not include any dynamic branches (branching based on measurement results, directly or transitively through other dynamically computed values). It also enforces any loops for the expression are unrolled and all calls are inlined. This has the effect of ensuring that all QIR code for the expression appears in the same LLVM basic block and maximizes the potential for later compilation steps to reorder and schedule those instructions together on the target hardware without needing to move instructions across basic blocks. In the case of the limit specified via `parallel within` expressions, this can be any static integer expression, forcing the limit to be a value known at compile time. These restrictions are reported as design-time feedback when the code is set to a QIR profile via RCA analysis:

Image

For QIR with static qubit identifiers, the changes to qubit management are simply reflected in the identifiers used in the generated code. More qubits get used, so distinct identifiers are generated into the QIR, showing that those instructions do not need to occur sequentially.

Once QIR with dynamic qubit allocation is supported, we would need to ensure that parallel expressions cause calls to release to be deferred to the end of the expression (or delayed to a midpoint in the case of `parallel within` triggering limited reuse). This can use the same infrastructure introduced to handled static qubit identifiers, replacing compile time calls to allocation and release with emitted calls to allocate and release.

### Effect on Resource Estimation
Resource Estimation uses the logical counts and rotation depth computed from execution of the program, so updating the evaluator for simulation also benefits estimations. This means users can see the effect of `parallel` and `parallel within` on their estimtates and use the configured parallel sections and parallel limits as a way to explore trade-offs in space and time for the execution of the algorithm.

## Limitations
The `parallel` keyword does not _guarantee_ parallelism of the corresponding expression, but instead acts as a signal to the compiler to change automatic resource management to optimize for time instead of space, which ensures qubit reuse does not _prevent_ parallel execution when backends schedule the program for execution on quantum hardware. This is analogous to features in other languages such as the `inline` keyword in C++, which similarly acts as a hint/request to the compiler but is not a hard requirement. Likewise, the limit used on `parallel within` expressions is not a hard limit, so qubit allocations beyond the specified number do not fail, instead indicating the point at which qubit reuse should be allowed rather than deferred.

The `parallel` and `parallel within` expressions also do not enforce that no qubit is operated on more than once within the expression. Concretely, this means that not every instruction corresponding to that expression is known to be parallelizable. Again, it is only the automatically managed quantum memory resources and allowed control flow that is changed, which ensures that behavior of the compiler does not prevent parallelism.

## Interaction with Other Syntax
Since `parallel` is a new keyword allowed in any expression context, it may break existing code that uses the word "parallel" as an identifer. This does not affect any standard library code or sample code, but may be a **breaking change** if user code uses parallel. Note that this only affects the all lower case, single-word spelling; "Parallel" or compound uses like "in_parallel" would not be affected.

Because Q# supports block expressions, uses may denote parallel sections of the code with explicit blocks, as in:

```qsharp
parallel {
    Op(qs);
}
```

Because `within` is an existing keyword, it will not cause any clashes with existing identifiers. However, it does mean that combining `parallel` expressions and conjugate expressions that use `within {} apply {}` to have extra delimiters. Specifically, 

```qsharp
parallel within {
    Op1(qs);
} apply {
    Op2(qs);
}
```

will result in a syntax error because `parallel within` is expected to be followed by two expressions, the first of type integer. Instead, the block expression should be used to clarify intent:

```qsharp
parallel {
    within {
        Op1(qs);
    } apply {
        Op2(qs);
    }
}
```

## Implementation Branch
To see the current implementation or try out the feature, use the `swernli/parallel` branch.

swernli:
This change adds support for a new `parallel` expression that alters the behavior of qubit allocation and code generation. There are two forms of the expression:
 - `parallel <expr>`: For the duration of the expression after the keyword, all qubit release is deferred such that each qubit allocation uses a fresh qubit identifier. In addition, when configured for QIR code generation, no expressions that result in dynamic branching are allowed within the expression and all loops within the expression are unrolled. This ensures that the generated QIR for that expression will be contained within a single LLVM block. Combining these two features (defferred release and enforced single block) maximizs the likelihood that the code corresponding to the expression will be parallelized by the target's compilation and scheduling.
 - `parallel within <limit> <expr>`: This form has the same semantics as the first form, but with the added integer limit that expresses how many fresh qubit allocations will be used before normal qubit reuse begins. For example, `parallel wthin 4 <expr>` will defer the release of four qubits, and if a fifth allocation is made within the expression then that allocation will draw from the previously deferred qubits, returning to a reuse pattern. This allows the program to express a limited amount of parallelism while still allowing for some reuse, providing more fine grained control over the width/depth (space/time) tradeoff of the algorithm. The limit expression must be a static (compile-time constant) integer.

The new expression affects the behavior of QIR code generation, simulation, and resource estimation, and the change includes updated parser, AST, HIR, FIR, and lowering logic for the syntax as well as new RCA checks to enforce the requirements for control flow structure and static integer limits.

Fixes #3274


source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -2137,3 +2137,232 @@ mod debugger_stepping {
...
+                            }
+                        }
+                    }
+                }
+            }
+        ",
+        CircuitEntryPoint::EntryPoint,
+    );
+
+    expect![[r#"
```
ScottCarda-MS, line 2321: It is difficult for me to understand what should be expected from the nested `parallel within` expressions, so I'll double-check with you. Is this correct and what is expected with the nested setup?

source/compiler/qsc_ast/src/ast.rs
```diff
@@ -917,6 +917,10 @@ pub enum ExprKind {
     Lambda(CallableKind, Box<Pat>, Box<Expr>),
     /// A literal.
     Lit(Box<Lit>),
+    /// A parallel expression: `parallel a`
+    Parallel(Box<Expr>),
+    /// A parallel-limited expression: `parallel within n a`
+    ParallelLimited(Box<Expr>, Box<Expr>),
```
ScottCarda-MS: Oh I didn't consider that we could have expressions for the limit of the parallel-within. Does that mean that the parallel limit could be determined by the result of a measurement?
```qsharp
use q0 = Qubit();
H(q0);
mutable limit = 2;
if MResetZ(q0) == One {
    limit = 4;
}
parallel within limit {
    { use q1 = Qubit(); H(q1); }
    { use q2 = Qubit(); H(q2); }
    { use q3 = Qubit(); H(q3); }
    { use q4 = Qubit(); H(q4); }
}
```

source/compiler/qsc_ast/src/ast.rs
```diff
@@ -964,6 +968,10 @@ impl Display for ExprKind {
             ExprKind::Interpolate(components) => display_interpolate(indent, components)?,
             ExprKind::Lambda(kind, param, expr) => display_lambda(indent, *kind, param, expr)?,
             ExprKind::Lit(lit) => write!(indent, "Lit: {lit}")?,
+            ExprKind::Parallel(e) => write!(indent, "Parallel: {e}")?,
+            ExprKind::ParallelLimited(limit, body) => {
+                write!(indent, "ParallelLimited: {limit} {body}")?;
```
ScottCarda-MS: Since limit and body can be full expressions, we should probably print them on their own lines. ~~Perhaps, if the limit is usually expected to be a literal, we could print it as "ParallelLimited({limit}):
    {body}
"
Otherwise~~ I might suggest
```
"ParallelLimited:
    {limit}
    {body}
"
```
similar to what Conjugate does.

source/compiler/qsc_hir/src/hir.rs
```diff
@@ -762,6 +764,13 @@ impl Display for ExprKind {
             ExprKind::If(cond, body, els) => display_if(indent, cond, body, els.as_deref())?,
             ExprKind::Index(array, index) => display_index(indent, array, index)?,
             ExprKind::Lit(lit) => write!(indent, "Lit: {lit}")?,
+            ExprKind::Parallel(limit, expr) => {
+                if let Some(limit) = limit {
+                    write!(indent, "Parallel({limit}): {expr}")?;
```
ScottCarda-MS: Ah I see you went with the parenthesized limit for the HIR display. I found that that gets pretty tricky to read when I put anything more than a straight int literal in for the limit. I'd still recommend to put the limit and the body on their own separate lines underneath. Vertical length is easier to handle than horizontal length in those displays.

ScottCarda-MS:
Because the `limit` of the `parallel within <limit> <body>` is also a full expression, it can become a block with side-effects and have its own qubit allocation and even sub-parallel expressions. We should be sure to test some of these scenarios for correct behavior.
For example, we should test
- a `parallel within` expression doesn't take effect in its own `limit` expression, only its `body`
- a surrounding `parallel` or `parallel within` _does_ affect the `limit` expression of a `parallel within` contained inside of the surrounding `parallel` or `parallel within`'s `body` expression.
  i.e
```qsharp
parallel within X {
  parallel within {
    // Should be subject to `parallel within X`
    use q0 = Qubit();
    use q1 = Qubit();
    Y
  } {...}
}
```
- `parallel within` expressions are effective inside the a block expression used for the `limit` expression of a surrounding `parallel within`, and that they don't affect the surrounding `parallel within`'s `body` expression

source/compiler/qsc_ast/src/ast.rs
```diff
@@ -917,6 +917,10 @@ pub enum ExprKind {
     Lambda(CallableKind, Box<Pat>, Box<Expr>),
     /// A literal.
     Lit(Box<Lit>),
+    /// A parallel expression: `parallel a`
+    Parallel(Box<Expr>),
```
idavis: Other expr kinds use option for various components, and you map to this pattern in HIR. Is there a design reason as to why you created two variants here?

source/compiler/qsc_fir/src/fir.rs
```diff
@@ -941,6 +941,10 @@ pub enum ExecGraphNode {
     Unit,
     /// The end of the control flow graph.
     Ret,
+    /// The start of a parallel region, with a Boolean indicating whether it has a limit.
+    ParStart(bool),
```
idavis: Create an enum so that this is self-descriptive instead of trying to remember a bool flag.

source/compiler/qsc_ast/src/ast.rs
```diff
@@ -917,6 +917,10 @@ pub enum ExprKind {
     Lambda(CallableKind, Box<Pat>, Box<Expr>),
     /// A literal.
     Lit(Box<Lit>),
+    /// A parallel expression: `parallel a`
```
idavis: Do we want the target expr to be a block to force a scope on the feature rather than allowing any expr?

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -1178,6 +1187,34 @@ impl With<'_> {
...
+                span: Span::default(),
+                ty: ty.clone(),
+                stmts: vec![stmt],
+            };
+            hir::Expr {
+                id: self.assigner.next_node(),
+                span: Span::default(),
+                ty,
+                kind: hir::ExprKind::Block(block),
+            }
```
idavis, line 1214: I think the stmt, block, and hir::Expr should use the `expr.span` rather than `Span::default()`

source/compiler/qsc_rca/src/errors.rs
```diff
@@ -252,8 +252,25 @@ pub enum Error {
...
+    #[diagnostic(code("Qsc.CapabilitiesCk.UseOfDynamicBranchingInParallelExpr"))]
+    UseOfDynamicBranchingInParallelExpr(#[label] Span),
+
+    #[error("cannot use dynamic limit for a parallel expression")]
+    #[diagnostic(help(
+        "using a dynamic limit for a parallel expression is not supported by the configured target profile"
+    ))]
+    #[diagnostic(url("https://aka.ms/qdk.qir#use-of-dynamic-limit-in-parallel-expr"))]
+    #[diagnostic(code("Qsc.CapabilitiesCk.UseOfDynamicLimitInParallelExpr"))]
+    UseOfDynamicLimitInParallelExpr(#[label] Span),
```
idavis, line 254: These anchors don't work yet, need the docs updated.

source/compiler/qsc_hir/src/hir.rs
swernli: Updated to use a multi-line display, the tests do read much cleaner now!

source/compiler/qsc_ast/src/ast.rs
swernli: Updated

source/compiler/qsc_ast/src/ast.rs
swernli: Went ahead and combined the enum variants.

source/compiler/qsc_ast/src/ast.rs
swernli: Syntactically, yes, this could happen. But RCA prevents it by ensuring the limit is not dependent on any measurement results.

source/compiler/qsc_ast/src/ast.rs
swernli: It gets turned into a block if it isn't one as part of lowering into HIR. But from a user-facing syntax perspective, there is value in having it support any expression. For example, for-loops are an expression, and one of the main motiviations is allowing `parallel for` syntax without having to add an extraneous block.

source/compiler/qsc_fir/src/fir.rs
swernli: Done.

ScottCarda-MS (approved):

swernli:
accidentally closed via bad force push

source/compiler/qsc_rca/src/errors.rs
swernli: Yup, that whole page in the wiki needs some additions and updates to bring it in line with where the product is now.

swernli: I think that can be a follow up, doesn't need to be a gate on the PR.

source/compiler/qsc/src/interpret/circuit_tests.rs
swernli: Yes, the qubit usage matches the same test case in the evaluator and partial evaluator. The idea of using the same case in each place is to confirm that each of their implementations of deferral gets the same ids, which matches expectation around partial reuse from a `parallel within 2` nested inside of `parallel within 6`.

source/compiler/qsc_frontend/src/lower.rs
swernli: That actually breaks the debugging experience. Leaving those spans as default ensures they are treated like generated code and we don't step to them. If instead I give them spans the debugger will step to the artificial statement inside the block, then step to the end of the artificial block, which looks confusing since the user doesn't have blocks there:
image
Keeping these as `Span::default()` preserves the expected behavior in debugging, and none of these generated spans should be the source of errors later.

idavis (approved):

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -4734,6 +4747,49 @@ impl<'a> PartialEvaluator<'a> {
...
+        };
+
+        self.resource_manager.start_delayed_release_layer(limit);
+        let result = self.try_eval_expr(expr_id)?;
+        self.resource_manager.end_delayed_release_layer();
+        Ok(result)
+    }
+
+    fn fail_if_in_parallel_expr(&self, span: PackageSpan) -> Result<(), Error> {
+        if self.resource_manager.is_delaying_release() {
```
swernli, line 4788: Need to update the IR function eligibility checks to also look at whether we are delaying release so that code is inlined into a single block as expected, then add tests to verify the expected shape of the QIR with `Adaptive` profile

swernli: Updated.

PR #3302, opened 2026-06-10
Add loss policies to `NoiseConfig` to express different kinds of behavior on lost qubits

orpuente-MS:
With this change, users are able to specify how gates should behave if at least one of their qubit operands is lost. Below is a comprehensive example, showing all 5 loss policies.

```python
from qdk.simulation import NoiseConfig, LossPolicy
qir = ...

noise = NoiseConfig() 
noise.cz.on_loss  = LossPolicy.SKIP               # if one of the qubits is lost, skip the unitary
noise.cx.on_loss  = LossPolicy.PROPAGATE          # if one of the qubits is lost, loose the other one
noise.rxx.on_loss = LossPolicy.DEGRADE            # degrade to a single qubit gate: rx
noise.ryy.on_loss = LossPolicy.RESIDUAL_S_DAGGER  # apply an S_DAG to the remaining qubits
noise.rzz.on_loss = LossPolicy.APPLY_ANYWAY       # apply unitary anyways

# Works with all simulator types, in any profile.
run_qir(qir, shots=100, noise=noise, type="clifford")
```


source/qdk_package/tests/test_simulators_gates_noisy.py
```diff
@@ -256,6 +256,140 @@ def test_two_qubit_loss(sim_type):
     )
 
 
+# ===========================================================================
+# Loss-policy (on_loss) tests
+# ===========================================================================
+#
+# These exercise the per-gate `NoiseConfig.<gate>.on_loss` behavior. The
+# `on_loss` policy is honored by the cpu (full-state) and clifford (stabilizer)
+# simulators, so these tests are parametrized over just those two.
```
billti: Is this correct? Looks like GPU is included below.

source/qdk_package/tests/test_simulators_gates_noisy.py
```diff
@@ -256,6 +256,140 @@ def test_two_qubit_loss(sim_type):
...
+        noise=noise,
+        sim_type=sim_type,
+    )
+    check_histogram(results, {"1-": 1.0})
+
+
+@pytest.mark.parametrize("sim_type", LOSS_POLICY_SIM_TYPES)
+def test_on_loss_swap_skip_keeps_state_but_swaps_loss_flag(sim_type):
+    # Overriding `swap.on_loss` to SKIP skips the SWAP unitary, but the loss
+    # flag is still exchanged. qs[0] keeps its reset |0> and qs[1] becomes lost.
```
billti: This one is interesting. Worth discussing more. Seems odd you would skip the unitary (which swaps the state) but still swap which is lost. Not sure if that would be physically possible.

source/simulators/src/gpu_full_state_simulator/common.wgsl
```diff
@@ -326,6 +326,213 @@ fn get_loss_idx(op_idx: u32) -> u32 {
     return 0u;
 }
 
+// Loss policy values. These are stamped onto a gate op's `q3` field by the host
+// (see `LossPolicy::as_u32` on the Rust side) and tell the shader how to handle
+// the gate when one of its operands is lost. `0` means "no policy stamped",
+// which the shader treats the same as SKIP.
```
billti, line 352: When I was looking at the enum I was wondering why the default value (0) wasn't skip. Is there a reason to start at 1 and leave 0 undefined/no-policy?

source/simulators/src/gpu_full_state_simulator/common.wgsl
```diff
@@ -326,6 +326,213 @@ fn get_loss_idx(op_idx: u32) -> u32 {
...
+
+// Returns true if the gate at `op_idx` touches at least one lost qubit.
+// `q1`/`q2` are the (resolved) operands of the gate.
+fn gate_has_lost_operand(shot_idx: u32, op_idx: u32, q1: u32, q2: u32) -> bool {
+    let shot = &shots[shot_idx];
+    let op = &ops[op_idx];
+    if (shot.qubit_state[q1].heat == -1.0) {
+        return true;
+    }
+    let is_2q = (op.id == OPID_CX || op.id == OPID_CY || op.id == OPID_CZ ||
```
billti: You already have `is_1q_op` defined (line 74). Can you just negate that? For now we only have 1q and 2q ops.

source/qdk_package/tests/test_simulators_gates_noisy.py
orpuente-MS: Good catch. This was before I decided to add noise policies to the GPU simulators as well.

source/simulators/src/gpu_full_state_simulator/common.wgsl
orpuente-MS: There was some general wisdom in the Python community that enums should start at 1 to avoid confusing them with `False` when treating them as booleans, but that doesn't seem to be the case anymore. Changed it to start from 0.

source/simulators/src/gpu_full_state_simulator/common.wgsl
orpuente-MS: Yes, you are right.

I think we should reorder the OPID constants so that all 1q ops are consecutive numbers. That will make the check cheaper, but it should be its own PR since it's kind of risky.

source/qdk_package/tests/test_simulators_gates_noisy.py
orpuente-MS: Mmh, maybe the behavior here should be:

- `SKIP` doesn't apply the unitary nor exchanges loss flags.
- `APPLY_ANYWAY` applies the unitary and exchanges the loss flags.

source/simulators/src/gpu_full_state_simulator/gpu_context.rs
```diff
@@ -927,7 +927,13 @@ fn add_noise_config_to_ops(ops: &[Op], noise: &NoiseConfig<f32, f64>) -> Vec<Op>
     let mut noisy_ops: Vec<Op> = Vec::with_capacity(ops.len() + 1);
 
     for op in ops {
-        let mut add_ops: Vec<Op> = vec![*op];
+        // Stamp the configured loss policy onto the gate op so the shader can
+        // decide how to handle the gate when one of its operands is lost.
+        let mut gate_op = *op;
+        if let Some(policy) = loss_policy_u32(op, noise) {
+            gate_op.q3 = policy;
```
billti: I was considering adding `ccx` at some point. Does using q3 for loss policy now prevent that? If the value at q3 is always going to be loss policy, should we rename it from q3?

source/simulators/src/cpu_full_state_simulator.rs
```diff
@@ -682,7 +706,15 @@ impl Simulator for NoisySimulator {
     }
 
     fn x(&mut self, target: QubitID) {
-        if !self.loss[target] {
+        if self.loss[target] {
+            // The only operand is lost. Only `ApplyAnyway` still applies a
+            // single-qubit gate; every other policy is equivalent to `Skip`.
+            if matches!(self.noise_config.x.on_loss, LossPolicy::ApplyAnyway) {
```
billti: Is this really the right behavior? What does it even mean to apply a single qubit unitary to a qubit that is always going to be measured as lost? (And I thought some parts of the code assumed a lost qubit is always in the zero state, which this would then violate, but I'll need to research that more)

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -452,112 +547,192 @@ impl Simulator for StabilizerSimulator {
...
+                        UnitaryOp::Z,
+                        UnitaryOp::SqrtZ,
+                        UnitaryOp::SqrtZInv,
+                    );
+                    self.state.apply_unitary(UnitaryOp::ControlledX, &[q2, q1]);
+                    self.state.apply_unitary(unitary, &[q1]);
+                    self.state.apply_unitary(UnitaryOp::ControlledX, &[q2, q1]);
+                }
             }
+        } else {
```
billti: In a number of these, you're not applying idle noise on the other qubit.

Also, for `ApplyAnyway` (which we should discuss further as it doesn't make much sense to me), should you also be applying idle noise anyway to the lost qubits?

source/simulators/src/gpu_full_state_simulator/common.wgsl
```diff
@@ -326,6 +326,208 @@ fn get_loss_idx(op_idx: u32) -> u32 {
...
+    // state is always exchanged regardless of the policy (the policy only
+    // governs whether the unitary runs). Handle it explicitly here.
+    if (op.id == OPID_SWAP) {
+        // Exchange the per-qubit loss flag (heat) of the two operands.
+        let heat1 = shot.qubit_state[q1].heat;
+        shot.qubit_state[q1].heat = shot.qubit_state[q2].heat;
+        shot.qubit_state[q2].heat = heat1;
+
+        if (policy == LOSS_POLICY_APPLY_ANYWAY) {
+            // A lost qubit is in a definite |0> state, so its bit is set in
```
billti: See my other comment. If you been using the 'APPLY_ANYWAY' policy on other operations (e.g. 'X') on the lost qubit, then it may not be in the |0> state any more.

source/simulators/src/gpu_full_state_simulator/common.wgsl
```diff
@@ -326,6 +326,208 @@ fn get_loss_idx(op_idx: u32) -> u32 {
...
+    let policy = op.q3;
+
+    // SWAP is special: it physically relocates the two qubits, so their loss
+    // state is always exchanged regardless of the policy (the policy only
+    // governs whether the unitary runs). Handle it explicitly here.
+    if (op.id == OPID_SWAP) {
+        // Exchange the per-qubit loss flag (heat) of the two operands.
+        let heat1 = shot.qubit_state[q1].heat;
+        shot.qubit_state[q1].heat = shot.qubit_state[q2].heat;
+        shot.qubit_state[q2].heat = heat1;
```
billti: Should you not also be swapping the zero and one probabilities of both qubits?

billti: (I guess you are setting `qubits_updated_last_op_mask = (1u << q1) | (1u << q2);` so the execute phase should update them anyway)

source/simulators/src/cpu_full_state_simulator.rs
orpuente-MS: I think you are right. Which means these policies only apply to multi-qubit gates.

source/simulators/src/stabilizer_simulator.rs
orpuente-MS: Added the missing idle noise applications.

We can discuss removing the `ApplyAnyway` policy.

source/simulators/src/gpu_full_state_simulator/gpu_context.rs
orpuente-MS: Added an `op.policy` field, and also three extra padding fields to maintain 16-byte alignment. Now an `Op` has 160 bytes (from 144 bytes it had before).

source/simulators/src/gpu_full_state_simulator/common.wgsl
orpuente-MS: This logic removes the bit flag indicating that the qubit is in a definite state, so that the SWAP gate doesn't skip over them. If the qubit is not in a definite state, this logic is just no-op.

I removed the bit saying "A lost qubit is in a definite |0> state" since it's inaccurate.

source/qdk_package/tests/test_simulators_gates_noisy.py
orpuente-MS: I came up with a simple principle: no matter the policy, we should never recover a qubit that was lost. Following this rule of thumb, I created the following table describing the behavior of SWAP under each policy:

| Policy            | apply unitary | exchange loss flags | extra behavior          |
| ----------------- | ------------- | ------------------- | ----------------------- |
| SKIP              | no            | no                  | no                      |
| PROPAGATE         | no            | no                  | loose other qubit       |
| DEGRADE           | no            | no                  | no                      |
| RESIDUAL_S_DAGGER | yes           | yes                 | apply S_ADJ to survivor |
| APPLY_ANYWAY      | yes           | yes                 | no                      |

billti: That table doesn't look right. You do apply the unitary for `DEGRADE` I thought, just to the other qubit. And 'exchange loss flags' being 'yes' on `RESIDUAL_S_DAGGER` seems wrong.

source/qdk_package/qdk/_native.pyi
```diff
@@ -868,6 +868,25 @@ class QirInstruction: ...
...
+    # If any operand of a gate is lost, propagate the loss to the other operands.
+    PROPAGATE: int
+    # For multi-qubit rotations, degrade the unitary to its single-qubit version
+    # on the surviving operand (e.g. rxx -> rx). Falls back to SKIP for gates with
+    # no single-qubit reduction (cx, cy, cz, swap, and single-qubit gates).
+    DEGRADE: int
+    # Skip the gate and instead apply an S adjoint to each surviving operand.
+    RESIDUAL_S_DAGGER: int
+    # Apply the unitary anyway, ignoring the loss.
+    APPLY_ANYWAY: int
```
billti: Does this really 'ignore the loss'? I though this policy was specifically for SWAP, where its often just a relabelling of qubits. In that case, doesn't the loss get swapped as well? (i.e. if before the gate qubit A was lost and B was in state $\alpha$, then after the swap A will be in state $\alpha$ and qubit B will be lost)

source/simulators/src/cpu_full_state_simulator.rs
```diff
@@ -798,83 +862,130 @@ impl Simulator for NoisySimulator {
...
     }
 
     fn swap(&mut self, q1: QubitID, q2: QubitID) {
+        // There are three kinds of swaps:
+        //   1. A logical swap, also called a relabel.
+        //   2. A swap by physically exchanging the location of the qubits.
+        //   3. An exchange of information by doing three CX.
+        //
+        // This method is concerned with the kinds (1) and (2), since (3)
+        // gets decomposed into other instructions before making it to the simulator.
```
billti, line 947: We should definitely doc these assumption, else I can see the behavior/assumptions being confusing to some users.

source/simulators/src/cpu_full_state_simulator.rs
```diff
@@ -798,83 +862,130 @@ impl Simulator for NoisySimulator {
...
+                let remaining_qubit = if self.loss[q1] { q2 } else { q1 };
+                self.apply_idle_noise(remaining_qubit);
+                match self.noise_config.rzz.on_loss {
+                    LossPolicy::Skip => (),
+                    LossPolicy::Degrade => return self.rz(angle, remaining_qubit),
+                    LossPolicy::Propagate => self.loss_impl(remaining_qubit),
+                    LossPolicy::ResidualSDagger => self.residual_s_dagger(remaining_qubit),
+                    LossPolicy::ApplyAnyway => self
+                        .state
+                        .apply_operation(&rzz(angle), &[q1, q2])
```
billti: This doesn't make much sense. What does it mean to 'apply anyway' to a lost qubit? If it leaves the qubit in the |0> state, how does that differ from degrade? If it doesn't leave it in the |0> state, doesn't that break out promise/assumptions about the _state of a lost qubit_?

source/simulators/src/cpu_full_state_simulator.rs
```diff
@@ -760,36 +776,84 @@ impl Simulator for NoisySimulator {
...
+            (true, false) | (false, true) => {
+                let remaining_qubit = if self.loss[control] { target } else { control };
+                self.apply_idle_noise(remaining_qubit);
+                match self.noise_config.cx.on_loss {
+                    LossPolicy::Skip | LossPolicy::Degrade => (),
+                    LossPolicy::Propagate => self.loss_impl(remaining_qubit),
+                    LossPolicy::ResidualSDagger => self.residual_s_dagger(remaining_qubit),
+                    LossPolicy::ApplyAnyway => self
+                        .state
+                        .apply_operation(&CX, &[control, target])
```
billti: Same comment here as for Rxx (and all R__ and C_ gates). What does it even mean to apply anyway? If the controls are lost (and we treat lost qubits as |0>), then the gate does nothing. If the target is lost (and treated as |0>), doesn't that also make the controlled gates a no-op?

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
```diff
@@ -80,10 +80,14 @@ struct Op {
     q1: u32,
     q2: u32,
     q3: u32,
+    policy: u32,
+    pad0: u32,
+    pad1: u32,
+    pad2: u32,
```
billti, line 86: Completely separate topic and not a blocker for this PR - but why are all these fundamental structures in the "_\*adaptive.wgls_" file? Aren't these _common_ structures?

source/qdk_package/tests/test_simulators_gates_noisy.py
orpuente-MS: Degrade is meant to express the current behavior of `Rxx`, `Ryy`, and `Rzz` in our simulators, and doesn't make much sense for `CX`, `CY`, `CZ`, or `SWAP`. For example, what is the degradation of physically swapping two qubits in a neutral atom machine? Or what is the degradation of virtually relabeling two qubits?

About `RESIDUAL_S_DAGGER`, if we apply the unitary, we are changing the state of the two qubits in the simulator; if we don't swap the loss flags after that, we now have access to the state of a lost qubit, which doesn't seem right.

source/simulators/src/gpu_full_state_simulator/simulator_adaptive.wgsl
orpuente-MS: The main and only motivation for leaving them here was that not having them in this file (and also in base.wgsl) makes the little language service support I have for .wgsl files go away, which makes editing them a complete nightmare.

source/simulators/src/cpu_full_state_simulator.rs
orpuente-MS: It doesn't make much sense to me either. As I replied to one your previous comments, we can discuss removing the `APPLY_ANYWAY` policy. Mainly it exists for expressing the default behavior of the `SWAP` gate.

This is one of the problems that will keep arising if we have an enum of policies that we can apply to any gate: not all of them make sense for all gates. In fact, it is very likely that several of them only make sense for a single gate: e.g.: `DEGRADE` only makes sense for multi-qubit rotation gates, and `APPLY_ANYWAY` only makes sense for `SWAP`.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -329,35 +343,83 @@ impl Simulator for StabilizerSimulator {
...
+            (true, false) | (false, true) => {
+                let remaining_qubit = if self.loss[control] { target } else { control };
+                self.apply_idle_noise(remaining_qubit);
+                match self.noise_config.cx.on_loss {
+                    LossPolicy::Skip | LossPolicy::Degrade => (),
+                    LossPolicy::Propagate => self.loss_impl(remaining_qubit),
+                    LossPolicy::ResidualSDagger => self.residual_s_dagger(remaining_qubit),
+                    LossPolicy::ApplyAnyway => self
+                        .state
+                        .apply_unitary(UnitaryOp::ControlledX, &[control, target]),
```
billti: As we've covered elsewhere, if the qubit is lost, the simulator expects it to remain in the Zero state (and be marked as lost), correct? Here if policy is ApplyAnyway, the target is lost, and the Control is NOT Zero, aren't you going to flip the lost qubit into some non-zero (thus invalid) state?

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -329,35 +343,83 @@ impl Simulator for StabilizerSimulator {
...
+                self.apply_idle_noise(remaining_qubit);
+                match self.noise_config.cy.on_loss {
+                    LossPolicy::Skip | LossPolicy::Degrade => (),
+                    LossPolicy::Propagate => self.loss_impl(remaining_qubit),
+                    LossPolicy::ResidualSDagger => self.residual_s_dagger(remaining_qubit),
+                    LossPolicy::ApplyAnyway => {
+                        self.state.apply_unitary(UnitaryOp::SqrtZInv, &[target]);
+                        self.state
+                            .apply_unitary(UnitaryOp::ControlledX, &[control, target]);
+                        self.state.apply_unitary(UnitaryOp::SqrtZ, &[target]);
```
billti: Same comment as above - if the target is lost and you ApplyAnyway, you might be putting the simulator in an invalid state.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -420,8 +482,34 @@ impl Simulator for StabilizerSimulator {
...
+                        );
+                        // NOTE: We perform the Rxx gate by changing basis to Y and performing the decomposition of Rzz.
+                        self.state.apply_unitary(UnitaryOp::Hadamard, &[q1]);
+                        self.state.apply_unitary(UnitaryOp::Hadamard, &[q2]);
+                        self.state.apply_unitary(UnitaryOp::ControlledX, &[q2, q1]);
+                        self.state.apply_unitary(unitary, &[q1]);
+                        self.state.apply_unitary(UnitaryOp::ControlledX, &[q2, q1]);
+                        self.state.apply_unitary(UnitaryOp::Hadamard, &[q1]);
+                        self.state.apply_unitary(UnitaryOp::Hadamard, &[q2]);
+                    }
```
billti, line 492: same comment - you shouldn't be apply H gates etc. on lost qubits. It will corrupt the simulator state.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -499,38 +632,48 @@ impl Simulator for StabilizerSimulator {
...
+                let remaining_qubit = if self.loss[q1] { q2 } else { q1 };
+                self.apply_idle_noise(remaining_qubit);
+                match self.noise_config.swap.on_loss {
+                    LossPolicy::Skip | LossPolicy::Degrade => (),
+                    LossPolicy::Propagate => self.loss_impl(remaining_qubit),
+                    LossPolicy::ResidualSDagger => {
+                        self.state.apply_permutation(&[1, 0], &[q1, q2]);
+                        self.residual_s_dagger(lost_qubit);
+                        self.loss.swap(q1, q2);
+                    }
```
billti, line 623: Should it still also apply the swap if the loss policy is ResidualSDagger and not ApplyAnyway?

source/simulators/src/gpu_full_state_simulator/simulator_base.wgsl
```diff
@@ -310,13 +314,14 @@ fn prepare_op(@builtin(global_invocation_id) globalId: vec3<u32>) {
...
-        shot.op_type = OPID_ID;
-        shot.op_idx = op_idx;
-        return;
+    // Before doing further work, if any qubit for the gate is lost, dispatch
+    // the gate's configured loss policy (stamped on op.policy). For most policies
+    // this fully handles the op; APPLY_ANYWAY returns false so the gate runs as
+    // usual below.
+    if (gate_has_lost_operand(shot_idx, op_idx, op.q1, op.q2)) {
+        if (handle_lost_operand_policy(shot_idx, op_idx, op.q1, op.q2)) {
+            return;
```
billti: As this doesn't fall through, does that mean any configured pauli noise on the (remaining) qubit then won't be applied (i.e. line 333 below)? Is that the behavior we want? We should clarify (and document in the comments here)

source/qdk_package/qdk/_native.pyi
orpuente-MS: Fixed the wording.

source/simulators/src/cpu_full_state_simulator.rs
orpuente-MS: Changed 'apply anyway' so that it only applies to SWAP

source/simulators/src/cpu_full_state_simulator.rs
orpuente-MS: Changed 'apply anyway' so that it only applies to SWAP

source/simulators/src/stabilizer_simulator.rs
orpuente-MS: Changed 'apply anyway' so that it only applies to SWAP

source/simulators/src/stabilizer_simulator.rs
orpuente-MS: Changed 'apply anyway' so that it only applies to SWAP

source/simulators/src/stabilizer_simulator.rs
orpuente-MS: In this case I think it makes sense. I don't see any reason why a relabel or physical exchange of atoms would completely skip because one of the qubits is lost.

source/simulators/src/stabilizer_simulator.rs
orpuente-MS: Changed 'apply anyway' so that it only applies to SWAP

source/simulators/src/gpu_full_state_simulator/simulator_base.wgsl
orpuente-MS: Yes, this was ill-logic. Fixed, and added a test to gpu-runner.rs and to the python test suite.

source/qdk_package/src/qir_simulation.rs
```diff
@@ -555,6 +645,8 @@ impl NoiseTable {
         NoiseTable {
             qubits: num_qubits,
             pauli_noise: FxHashMap::default(),
+            on_loss: LossPolicy::Skip,
+            allowed_loss_policies: NoiseTable::DEFAULT_MULTI_QUBIT_LOSS_POLICIES.to_vec(),
```
billti: Shouldn't this be DEFAULT_SINGLE_QUBIT_LOSS_POLICIES if num_qubits is 1?

source/qdk_package/src/qir_simulation.rs
```diff
@@ -586,11 +678,21 @@ E.g.: `noise_config.cz.IL`"
...
-        } else {
-            self.set_pauli_noise_elt(&name.to_uppercase(), value)
+    ///
+    /// The `on_loss` attribute is special-cased to accept a `LossPolicy`.
+    fn __setattr__(&mut self, name: &str, value: &Bound<'_, PyAny>) -> PyResult<()> {
+        match name {
+            "on_loss" => {
+                if self.qubits < 2 {
+                    Err(PyAttributeError::new_err(
+                        "Loss policies only apply to multi-qubit gates.".to_string(),
```
billti, line 715: I'm confused. Why have a DEFAULT_SINGLE_QUBIT_LOSS_POLICIES if you can't see a loss policy on single qubit operations?

source/qdk_package/src/qir_simulation.rs
orpuente-MS: Doesn't really make a difference, since you can't change the policy for a single qubit gate; but yes, there should be a branch here. Added.

source/qdk_package/src/qir_simulation.rs
orpuente-MS: For completeness of this layer of the implementation. Since what happens on loss for single-qubit gates matches the semantics of `SKIP`, I had `DEFAULT_SINGLE_QUBIT_LOSS_POLICIES = [SKIP]`.

But since I indeed throw an error if the user tries to touch the loss policy of single-qubit gates at all, and it was confusing to you, I changed it to `DEFAULT_SINGLE_QUBIT_LOSS_POLICIES = []`.

orpuente-MS: Thinking more about it, having `DEFAULT_SINGLE_QUBIT_LOSS_POLICIES = [SKIP]` as I had before, is leaking details about how single-qubit gates work internally in the simulators. So, it was a bad idea having it set to that value.

source/qdk_package/src/qir_simulation.rs
```diff
@@ -721,29 +830,49 @@ fn from_noise_table_ref<T: Float>(
...
+    }
+
+    let additional_policies: &[LossPolicy] = match gate_name {
+        "rxx" | "ryy" | "rzz" => &[LossPolicy::Degrade],
+        "swap" => &[LossPolicy::ApplyAnyway],
+        _ => &[],
+    };
+
+    [
+        &NoiseTable::DEFAULT_MULTI_QUBIT_LOSS_POLICIES,
```
billti, line 871: Do measurements not count as gates that should be considered in the above table (and reset)?

billti: (And `mov`)

orpuente-MS: Yes, missed those. Added.

source/simulators/src/cpu_full_state_simulator.rs
orpuente-MS: Documented it in the `NoiseConfig` python API

billti (approved):

PR #3303, opened 2026-06-10
Connect to Workspace Supports Tenant ID Resource Identifier

ScottCarda-MS:
This expands the ability of the Connect to Workspace feature to be able to consume resource identification strings containing tenant ids, in addition to being able to consume connection strings with API keys. When connecting to a resource with a resource identification string, the user will be authenticated with Entra ID.

### DIFF
--- a/source/vscode/src/azure/auth.ts
+++ b/source/vscode/src/azure/auth.ts
@@ -141,7 +141,7 @@ export async function getTenantIdAndToken(
 
 export async function getTokenForWorkspace(workspace: {
   apiKey?: string;
-  tenantId: string;
+  tenantId?: string;
 }) {
   // If using an API key, just return that as the 'token'
   if (workspace.apiKey) return `apiKey=${workspace.apiKey}`;
@@ -149,7 +149,7 @@ export async function getTokenForWorkspace(workspace: {
   const associationId = getRandomGuid();
 
   const workspaceAuth = await getAuthSession(
-    [scopes.quantum, `VSCODE_TENANT:${workspace.tenantId}`],
+    [scopes.quantum, `VSCODE_TENANT:${workspace.tenantId ?? ""}`],
     associationId,
   );
   return workspaceAuth.accessToken;
--- a/source/vscode/src/azure/workspaceActions.ts
+++ b/source/vscode/src/azure/workspaceActions.ts
@@ -209,11 +209,21 @@ workspace = Workspace(
 
 /**
  * Parses a connection string into a WorkspaceConnection, or returns undefined
- * if any required fields are missing. Does not validate the connection against
- * the service — call getTokenForWorkspace / azureRequest to validate.
+ * if any required fields are missing or the authentication fields are invalid.
+ * Does not validate the connection against the service — call
+ * getTokenForWorkspace / azureRequest to validate.
  *
- * Expected format:
- *   SubscriptionId=<guid>;ResourceGroupName=<name>;WorkspaceName=<name>;ApiKey=<secret>;QuantumEndpoint=<serviceUri>
+ * Required fields: SubscriptionId, ResourceGroupName, WorkspaceName, QuantumEndpoint
+ * Authentication (exactly one must be provided):
+ *   - ApiKey:   use API key authentication
+ *   - TenantId: use Entra ID (AAD) authentication; VS Code will prompt the user
+ *               to sign in interactively if not already authenticated to this tenant
+ *
+ * API key format:
+ *   SubscriptionId=<guid>;ResourceGroupName=<name>;WorkspaceName=<name>;ApiKey=<secret>;QuantumEndpoint=<https://...>
+ *
+ * Entra ID format:
+ *   SubscriptionId=<guid>;ResourceGroupName=<name>;WorkspaceName=<name>;TenantId=<tenant-guid>;QuantumEndpoint=<https://...>
  */
 export function parseConnectionString(
   connStr: string,
@@ -225,12 +235,20 @@ export function parseConnectionString(
     partsMap.set(part.substring(0, eq).toLowerCase(), part.substring(eq + 1));
   });
 
+  const hasRequiredFields =
+    partsMap.has("subscriptionid") &&
+    partsMap.has("resourcegroupname") &&
+    partsMap.has("workspacename") &&
+    partsMap.has("quantumendpoint");
+
+  const hasApiKey = partsMap.has("apikey");
+  const hasTenantId = partsMap.has("tenantid");
+
+  // Exactly one authentication method must be provided
   if (
-    !partsMap.has("subscriptionid") ||
-    !partsMap.has("resourcegroupname") ||
-    !partsMap.has("workspacename") ||
-    !partsMap.has("apikey") ||
-    !partsMap.has("quantumendpoint")
+    !hasRequiredFields ||
+    (!hasApiKey && !hasTenantId) ||
+    (hasApiKey && hasTenantId)
   ) {
     return undefined;
   }
@@ -244,7 +262,7 @@ export function parseConnectionString(
     id: workspaceId,
     name: partsMap.get("workspacename")!,
     endpointUri: partsMap.get("quantumendpoint")!,
-    tenantId: "", // Blank means not authenticated via a token
+    tenantId: partsMap.get("tenantid") ?? "", // Non-empty triggers Entra ID auth in getTokenForWorkspace
     subscriptionId: partsMap.get("subscriptionid"),
     apiKey: partsMap.get("apikey"),
     providers: [], // Providers and jobs will be populated by a following 'queryWorkspace' call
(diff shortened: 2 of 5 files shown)


source/vscode/src/azure/workspaceActions.ts
```diff
@@ -225,12 +235,20 @@ export function parseConnectionString(
...
+
+  // Exactly one authentication method must be provided
   if (
-    !partsMap.has("subscriptionid") ||
-    !partsMap.has("resourcegroupname") ||
-    !partsMap.has("workspacename") ||
-    !partsMap.has("apikey") ||
-    !partsMap.has("quantumendpoint")
+    !hasRequiredFields ||
+    (!hasApiKey && !hasTenantId) ||
```
amcasey: Probably too cute, but you could check `hasApiKey === hasTenantId`.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -225,12 +235,20 @@ export function parseConnectionString(
...
   });
 
+  const hasRequiredFields =
+    partsMap.has("subscriptionid") &&
+    partsMap.has("resourcegroupname") &&
+    partsMap.has("workspacename") &&
+    partsMap.has("quantumendpoint");
+
+  const hasApiKey = partsMap.has("apikey");
+  const hasTenantId = partsMap.has("tenantid");
```
amcasey: This feels like it will return true for an empty TenantId.  Does something else prevent that from happening?

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -244,7 +262,7 @@ export function parseConnectionString(
     id: workspaceId,
     name: partsMap.get("workspacename")!,
     endpointUri: partsMap.get("quantumendpoint")!,
-    tenantId: "", // Blank means not authenticated via a token
+    tenantId: partsMap.get("tenantid") ?? "", // Non-empty triggers Entra ID auth in getTokenForWorkspace
     subscriptionId: partsMap.get("subscriptionid"),
     apiKey: partsMap.get("apikey"),
```
amcasey, line 282: `apikey` doesn't need a corresponding fallback?

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -296,18 +314,31 @@ async function getWorkspaceWithConnectionString(
...
+      const usingEntraId = !workspace.apiKey;
+      const workspaceDescription =
+        `the workspace "${workspace.name}"` +
+        (workspace.tenantId ? ` in tenant ${workspace.tenantId}` : "");
       let errorText = "An unexpected error occurred";
       const message: string | undefined = e.message;
-      if (message?.includes("status 401")) {
-        errorText =
-          "Authentication failed. Please check the ApiKey is valid and active.";
+      if (message?.includes("status 401") || message?.includes("status 403")) {
```
amcasey: Is there a better type for `e` than `any` that would let us access the status code directly?

amcasey (commented):
Seems sensible.  Minor suggestions

source/vscode/src/azure/workspaceActions.ts
ScottCarda-MS: It is a bit more involved, but we can set the catch to `unknown` and narrow to a predefined error type to get the status and the message more concretely. I've gone ahead and done that here.

source/vscode/src/azure/workspaceActions.ts
ScottCarda-MS: We should only be expecting connection strings generated from our own tooling (or the Service team's tooling), but it doesn't hurt to be a little extra careful, so good to add some logic to only accept parts that have non-empty values. Done!

source/vscode/src/azure/workspaceActions.ts
ScottCarda-MS: A bit more condensed, but the surrounding comments can help with readability. Done

source/vscode/src/azure/workspaceActions.ts
ScottCarda-MS: Not really. The tenant id is required later on for a number of things, like deep links to the portal, while the api key is genuinely optional. If the api key isn't present, things can work just fine without it.

source/vscode/src/azure/workspaceActions.ts
amcasey: I thought it was the customer providing the tenant.  But maybe we validate that it's non-empty when they do?

amcasey: My point was mostly that it seems like `partsMap.has` and `partsMap.get` could disagree and that could be confusing.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -294,20 +316,34 @@ async function getWorkspaceWithConnectionString(
...
           "Workspace not found. Please check the WorkspaceName and ResourceGroupName values.";
       } else if (message?.includes("Failed to fetch")) {
         errorText =
           "Request failed. Please check the QuantumEndpoint value and network connectivity.";
+      } else if (usingEntraId) {
+        // No HTTP status code: the interactive sign-in itself failed or was
+        // cancelled (e.g. the account is not a member of the tenant, or consent
+        // was denied). Without this branch the failure would surface only as a
+        // generic message, appearing to fail silently to the user.
+        errorText = `Authentication failed. Unable to sign in to ${workspaceDescription}. Please verify the account has access to this workspace and try again.`;
```
billti, line 371: I'm finding this all a bit convoluted to follow now, and I'm not sure I understand how the various use-cases interact with this code. Let's discuss and clarify (and maybe put in comments for posterity).

ScottCarda-MS: The tenant id and the api key handling have been separated, so hopefully that brings some clarity and cleanliness to the code that you are looking for.

source/vscode/src/azure/auth.ts
```diff
@@ -141,15 +141,15 @@ export async function getTenantIdAndToken(
...
+  tenantId?: string;
 }) {
   // If using an API key, just return that as the 'token'
   if (workspace.apiKey) return `apiKey=${workspace.apiKey}`;
 
   const associationId = getRandomGuid();
 
   const workspaceAuth = await getAuthSession(
-    [scopes.quantum, `VSCODE_TENANT:${workspace.tenantId}`],
+    [scopes.quantum, `VSCODE_TENANT:${workspace.tenantId ?? ""}`],
```
billti, line 152: This doesn't look right to me. Did you test this flow when this value is blank?

The VSCODE_TENANT scope, if provided here, has a specific value required for authenticating to the right tenant (see https://github.com/microsoft/vscode/blob/1.126.0/extensions/microsoft-authentication/src/common/scopeData.ts#L41 ). I would expect passing an empty string here would break it, so I'd be interested in how this was tested and the behavior your saw.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -283,6 +305,20 @@ async function getWorkspaceWithConnectionString(
       }
     }
 
+    // Tenant id discovery (independent of the auth method). The tenant id is
+    // needed for Entra ID auth and for building portal deep links. If it wasn't
+    // supplied in the connection string, derive it from the subscription id.
+    // This must happen before validation because the Entra ID auth path below
+    // (used whenever there is no API key) requires a tenant id.
```
billti, line 323: Exactly (see other comment I just left). It should never just default to an empty string if missing. That would be invalid. It should throw an error if the tenantId is missing when trying to do Entra authentication.

source/vscode/src/azure/workspaceActions.ts
```diff
@@ -49,7 +50,7 @@ function buildQuantumOsFragment(workspace: WorkspaceConnection): string {
     workspace.offeringId ?? workspace.providers[0]?.providerId ?? "";
 
   return (
-    `tenantId=${workspace.tenantId}` +
+    `tenantId=${workspace.tenantId ?? ""}` +
```
billti: Same question here. Is it ever valid for a workspace portal link to have `tenantId=""`? If not, this should error, not build an invalid link.

source/vscode/src/azure/auth.ts
ScottCarda-MS: Yes this is the scenario we've had up until now, where tenant id is derived from subscription id. If a tenant id is provided, we use that, otherwise we use the subscription id to get the tenant id.

source/vscode/src/azure/workspaceActions.ts
ScottCarda-MS: The tenant id is optional for now from the URI's we accept, but you are correct that by the time we reach this part of the code we should always have a tenant id. It is possible that something went wrong with getting the tenant id from the subscription id. I've added a check here that throws an error if that happens.
After the next sprint, we shouldn't have this complexity, as the portal will have made the switch to just using URIs with tenant id in them, so it can be fully mandatory then.

ScottCarda-MS:
Superceeded by #3404

PR #3305, opened 2026-06-11
Stim compiler

joao-boechat:
This is an initial PR for the Stim compiler, which includes the setup for supporting the language. We still will need to add tests, better error handling, more language features, among other things, but the compiler introduced by this PR is supposed to be fully functional and minimally faulty.

All of the code was designed around the [stim](https://github.com/quantumlib/Stim/tree/main) language, which is mostly defined by these two documents:

[Stim/doc/file_format_stim_circuit.md at main · quantumlib/Stim](https://github.com/quantumlib/Stim/blob/main/doc/file_format_stim_circuit.md)
[Stim/doc/gates.md at main · quantumlib/Stim](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#MPP)


source/compiler/qsc_stim_parser/examples/lex_stim.rs
```diff
@@ -0,0 +1,33 @@
+use qsc_stim_parser::lex::Lexer;
+use std::fs;
+use std::io::Write;
+
+fn main() {
+    let stim_code =
+        fs::read_to_string("examples/example.stim").expect("Failed to read examples/example.stim");
```
billti: I don't see this file anywhere in the PR. Also, most of our tests use inline string of input/output to verify, rather than external files.

joao-boechat: this is for my manual testing, just so I can parse whole files at a time without having to edit the code every time. I will make sure to remove it before the official PR

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Sequence)]
+pub enum TokenKind {
+    Newline,         // \n
+    Uint,            // unsigned integers
+    Double,          // floating-point numbers
+    InstructionName, // H, X, CNOT, etc.
+    Rec,             // rec[- ...]
+    Sweep,           // sweep[...]
+    Tag,             // "[...]"
+    Open(Delim),     // ( {
```
amcasey, line 27: Out of curiosity, why is this one TokenKind with a parameter, rather than two token kinds?  Are there places we want to handle either form of bracket?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
+use enum_iterator::Sequence;
+use qsc_data_structures::span::Span;
+use std::str::CharIndices;
+use std::{
+    fmt::{self, Display, Formatter},
+    iter::Peekable,
+};
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Token {
```
amcasey, line 13: Is there a concept of a known-but-erroneous token?  For example, in many languages `0.2` is valid, but `.2` is not, but you'd want both to appear as `Double` tokens for error recovery purposes.  (This is almost certainly out of scope for this proof-of-concept implementation.)

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+impl<'a> Lexer<'a> {
+    pub fn new(input: &'a str) -> Self {
+        Self {
+            input,
+            input_len: input
+                .len()
+                .try_into()
+                .expect("input length should fit into u32"),
```
amcasey, line 83: I'm curious about this restriction.  Is `input` capable of holding more than max_int bytes?  If so, do we need to restrict it?  If not, isn't the restriction implied?  (The code might exactly express that it's impossible - I'm still getting used to reading rust.)

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+    input: &'a str,
+    input_len: u32,
+    chars: Peekable<CharIndices<'a>>,
+}
+
+impl<'a> Lexer<'a> {
+    pub fn new(input: &'a str) -> Self {
+        Self {
+            input,
+            input_len: input
```
amcasey, line 80: To confirm, this is the length in bytes but `chars` will be in characters and may be shorter?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+    fn comment(&mut self) {
+        self.eat_while(|c| c != '\n');
+        self.whitespace();
+    }
+
+    fn scan_number(&mut self) -> TokenKind {
+        self.eat_while(|c| c.is_ascii_digit());
+        let mut is_double = false;
+        if self.chars.next_if(|(_, c)| *c == '.').is_some() {
+            self.eat_while(|c| c.is_ascii_digit());
```
amcasey, line 129: Is this guaranteed to consume at least one digit?  Or does the language allow `2.` as a valid double?

amcasey: Looks like we allow scientific notation.  Surely, `2.e` isn't allowed?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.eat_while(char::is_whitespace);
+    }
+
+    fn comment(&mut self) {
+        self.eat_while(|c| c != '\n');
+        self.whitespace();
+    }
+
+    fn scan_number(&mut self) -> TokenKind {
+        self.eat_while(|c| c.is_ascii_digit());
```
amcasey, line 108: How much do we care about leading zeroes?

source/compiler/qsc_stim_parser/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+    fn whitespace(&mut self) {
+        self.eat_while(char::is_whitespace);
+    }
+
+    fn comment(&mut self) {
+        self.eat_while(|c| c != '\n');
+        self.whitespace();
+    }
+
+    fn scan_number(&mut self) -> TokenKind {
```
amcasey: Can there be a sign for the whole number? `+2`?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+        if is_double {
+            TokenKind::Double
+        } else {
+            TokenKind::Uint
+        }
+    }
+
+    fn scan_identifier(&mut self, lo: usize) -> TokenKind {
+        self.eat_while(|c| c.is_alphanumeric() || c == '_');
```
amcasey, line 166: A lot of languages don't allow identifiers to start with digits.  Not sure if that's true of stim.

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+        let hi: usize = self
+            .chars
+            .peek()
+            .map_or(self.input_len as usize, |(i, _)| *i);
+        // TODO: What if some identifier starts with "rec" but is not a rec token?
+        match &self.input[lo..hi] {
+            "rec" => {
```
amcasey, line 173: I'm probably just blanking, but where did we check for the open `[`?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+                self.eat_while(|c| c != ']');
+                self.chars.next_if(|(_, c)| *c == ']');
+                TokenKind::Sweep
+            }
+            _ => TokenKind::InstructionName,
+        }
+    }
+}
+
+impl Iterator for Lexer<'_> {
```
amcasey, line 188: Basic rust, is `'_` magic?

source/compiler/qsc_stim_parser/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+                .expect("input length should fit into u32"),
+            chars: input.char_indices().peekable(),
+        }
+    }
+
+    fn eat_while(&mut self, f: impl Fn(char) -> bool) {
+        while self.chars.next_if(|i| f(i.1)).is_some() {}
+    }
+
+    fn whitespace(&mut self) {
```
amcasey: Is this going to consume newlines without creating corresponding tokens?

amcasey (commented):
Some questions about the lexer.  I'm just learning, so don't take any of this as blocking.

source/compiler/qsc_stim_parser/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+
+#[derive(Debug)]
+pub enum Item {
+    Line(Line),
+    Block(Block),
+}
+
+#[derive(Debug)]
+pub struct Line {
+    pub span: Span,
```
amcasey: Is this different from the span that's in the Instruction?

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+#[derive(Debug)]
+pub struct Target {
+    pub span: Span,
+    pub kind: TargetKind,
+}
+
+#[derive(Debug)]
+pub enum TargetKind {
+    Qubit {
+        negated: bool,
```
amcasey, line 50: For my reference, what does negated mean?

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+            // TODO error! The start of every item should be an instruction;
+            None
+        }
+    }
+
+    fn parse_block(&mut self, instruction: Instruction) -> Block {
+        let lo = instruction.span.lo;
+        let mut items = Vec::new();
+        self.expect(TokenKind::Open(Brace));
+        self.expect(TokenKind::Newline);
```
amcasey, line 183: Is the newline after the open brace mandatory?

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+                .is_some_and(|t| t.kind == TokenKind::Close(Brace))
+            {
+                break;
+            }
+            match self.parse_item() {
+                Some(item) => items.push(item),
+                None => break,
+            }
+        }
+        let closing_brace = self.expect(TokenKind::Close(Brace));
```
amcasey, line 197: In the future, we might want to synthesize a missing closing brace for recovery purposes.

source/compiler/qsc_stim_parser/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+        let hi = closing_brace.span.hi;
+        Block {
+            span: Span { lo, hi },
+            block_instruction: instruction,
+            items,
+        }
+    }
+
+    fn parse_line(&mut self, instruction: Instruction) -> Line {
+        self.expect(TokenKind::Newline);
```
amcasey: Personally, I find it a little strange to start a line with a newline, rather than end it with a newline.  Does this cause any problems at file boundaries?

source/compiler/qsc_stim_parser/src/parser.rs
```diff
@@ -0,0 +1,379 @@
...
+            TokenKind::Star => Target {
+                span,
+                kind: TargetKind::Combiner,
+            },
+            _ => panic!("Unexpected target kind"),
+        }
+    }
+
+    fn extract_uint(&mut self, token: Token, span: Option<Span>) -> u32 {
+        self.extract_string(token, span).parse().unwrap()
```
amcasey: I think this panics if the number is too large to fit in a u32, for example?

amcasey (commented):
Parser comments.  Still non-blocking.  Happy to chat if my questions don't make sense (which is reasonably likely).

source/compiler/stim_compiler/src/lex.rs
orpuente-MS: Short answer: yes. Longer answer:

In general `_`, when used in a type or lifetime position, is telling the compiler: "hey, infer what goes here."

So, the `'` is just the prefix symbol for a lifetime (like in `'some_lifetime`), and the `_` is an inference request for the compiler.

source/compiler/stim_compiler/src/lex.rs
orpuente-MS: This pattern comes from the Q# compiler, where we have `(`, `[`, and `{`. In that case doing `Open(Delim)`, and `Close(Delim)` allows you express `2 * |Delim|`, i.e.: 6 tokens concisely, and in that way we don't clutter the Token enum with open/close tokens. It also makes some invalid code irrepresentable, you can't have an `OpenParen` but forget the `CloseParen` by mistake.

source/compiler/stim_compiler/src/lex.rs
orpuente-MS: This is interesting. I think this restriction comes from Q#, which is a high-level language in which we don't expect a single file to exceed `u32::MAX` bytes, (that would be a ~50 million lines file with ~80 chars per line). Therefore, the `Span` data structure is defined as

```rust
pub struct Span {
    /// The smallest offset contained in the span.
    pub lo: u32,
    /// The next offset after the largest offset contained in the span.
    pub hi: u32,
}
```

Q# and OpenQASM use this `Span` to talk to our language service + error reporting infrastructure; and now Stim is using it as well.

Stim also has mechanisms for writing concise code, like broadcasting for applying one instruction to many qubits or for-loops for repetition. So, even though `input` is capable of holding up to `usize::MAX` bytes (which is `u64::MAX` in a 64-bit target), I don't expect to see ~50 million lines Stim files.

source/compiler/stim_compiler/src/lex.rs
orpuente-MS: Yes, `chars` will contain fewer characters than `input_len`, since a `char` in rust represents a Unicode value. However, `CharIndices`, which is the iterator used to iterate through the character stream, yields tuples of (char, byte_position). So we are always speaking the "byte position" idiom when building Spans and indexing into strings.

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+      "[One, Zero]\n",
+      "[Zero, Zero]\n",
+      "[One, Zero]\n",
+      "[One, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[One, Zero]\n",
+      "[Zero, Zero]\n"
```
billti: This looks like a simple bell pair with no noise, so why is the output not all `[One, One]` and `[Zero, Zero]`?

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+   ],
+   "source": [
+    "# A correlated error on qubits 0 and 1: X on both, together, with probability 0.2.\n",
+    "correlated_circuit = \"\"\"CORRELATED_ERROR(0.2) X0 X1\n",
+    "ELSE_CORRELATED_ERROR(0.1) X1 X2\n",
+    "ELSE_CORRELATED_ERROR(0.1) X0 X2\n",
+    "MR 0 1\n",
+    "\"\"\"\n",
+    "\n",
+    "results = stim.run(correlated_circuit, shots=20, type=\"cpu\")\n",
```
billti: Why is the type always `cpu`? Showing the Clifford simulator is more interesting for the error correction needs.

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -0,0 +1,195 @@
+use enum_iterator::Sequence;
```
billti, line 4: Add a license header

source/qdk_package/qdk/simulation/_simulation.py
```diff
@@ -743,3 +744,14 @@ def run_qir(
             return run_qir_gpu(input, shots, noise, seed)
         case _:
             raise ValueError(f"Invalid simulator type: {type}")
+
+
+def run_stim(
```
billti: What's the value in having this API here if we already have `qdk.stim.run`?

samples/notebooks/stim_to_qir.ipynb
orpuente-MS: I just checked the generated QIR, and the `CX` gate is not being emitted, `CZ` is being correctly emitted though, handling `CX` is probably missing in the parser.

source/qdk_package/qdk/simulation/_simulation.py
orpuente-MS: This is a left-over, we should remove it.

source/compiler/qsc_stim_parser/src/qir.rs
```diff
@@ -0,0 +1,765 @@
...
+                    continue;
+                };
+                let TargetKind::Qubit { value: v1, .. } = pair[1].kind else {
+                    continue;
+                };
+                self.writer
+                    .write_call(&gate, &[Operand::Qubit(v0), Operand::Qubit(v1)]);
+            }
+        }
+    }
```
orpuente-MS: This is where the handling for `CX` (and `CY`) is missing.

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+      "[Zero, One, Zero, Zero]\n",
+      "[Zero, Zero, Zero, Zero]\n",
+      "[Zero, Zero, Zero, Zero]\n",
+      "[Zero, One, Zero, Zero]\n",
+      "[One, Zero, One, Zero]\n",
+      "[Zero, One, Zero, Zero]\n",
+      "[One, Zero, Zero, Zero]\n",
+      "[One, One, One, Zero]\n",
+      "[One, One, Zero, Zero]\n",
+      "[Zero, Zero, Zero, Zero]\n"
```
billti: Shouldn't the second measurement for every shot be the same if you are preselecting on that value (i.e. looping until it measures the expected value)?

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+      "[One, Zero]\n",
+      "[One, Zero]\n",
+      "[Zero, Zero]\n",
+      "[One, Zero]\n",
+      "[Zero, Zero]\n",
+      "[One, One]\n",
+      "[One, One]\n",
+      "[One, One]\n",
+      "[One, One]\n"
+     ]
```
billti: It's hard to tell if the probabilities were applied correctly from raw shot output. Maybe run for thousands of shots (on the Clifford simulator) and show the results in a histogram widget rather than the raw list.

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n",
+      "[Zero, Zero]\n"
```
billti: None of these results shows any loss!!

samples/notebooks/stim_to_qir.ipynb
orpuente-MS: Yes, the stim input is wrong in this case. The line that says `#!preselect_expect 0 1` should really say `#!preselect_expect 1 1`. With that change I could see the expected output

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,591 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "# Mutually-exclusive correlated branches mixing loss and Pauli terms.\n",
+    "mixed_circuit = \"\"\"CORRELATED_ERROR(0.1) L0\n",
+    "ELSE_CORRELATED_ERROR(0.1) L1\n",
+    "ELSE_CORRELATED_ERROR(0.1) L0 Z1\n",
```
billti: Might be more interesting to have a `X1` rather than `Z1` here, as the Z error is unobservable on a qubit in the `|0>` state.

source/compiler/qsc_stim_parser/src/qir.rs
```diff
@@ -0,0 +1,765 @@
...
+            // decomposed into H S H
+            for target in &instruction.targets {
+                let TargetKind::Qubit { value, .. } = target.kind else {
+                    continue;
+                };
+                let q = Operand::Qubit(value);
+                self.writer.write_call("h", &[q]);
+                self.writer.write_call("s", &[q]);
+                self.writer.write_call("h", &[q]);
+            }
```
orpuente-MS: We support sqrt_x in `run_qir`, but its name is `sx`. So, you can emit it directly by doing `self.writer.write_call("sx", &[q]);`.

source/compiler/qsc_stim_parser/src/qir.rs
orpuente-MS: Just need to add `|| gate == "cx" || gate == "cy"` to this condition.

samples/notebooks/stim_to_qir.ipynb
joao-boechat: that was indeed the issue! fixed!

samples/notebooks/stim_to_qir.ipynb
joao-boechat: oscar is a lucky guy

source/compiler/qsc_stim_parser/src/lex.rs
joao-boechat: nope! numbers are very basic in stim, nothing fancy (not even computations)

source/compiler/stim_compiler/src/lex.rs
joao-boechat: It is! per their [grammar](https://github.com/quantumlib/Stim/blob/main/doc/file_format_stim_circuit.md) this is what a name can be: [a-zA-Z][a-zA-Z0-9_]*
granted the "identifier" concept isn't exactly from stim, but I used it to simplify the code. Will have to revisit it later for correctness, though

source/compiler/stim_compiler/src/lex.rs
joao-boechat: The three cases we could have [] are:
1- rec[...]
2- sweep[...]
3- tags! For example in the statement: X_ERROR[a](0.01) 3 4

But parsing the brackets individually added a ton of complexity to distinguish between these three cases, so I chose to just consume them as a whole with those tokens, and then strip them away for the content. Will also revisit this later!

samples/notebooks/stim_to_qir.ipynb
```diff
@@ -0,0 +1,420 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "6c9df231",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qdk import stim\n",
+    "from qsharp_widgets import Histogram"
```
billti: Import from `qdk.widgets`, not `qsharp_widgets`

source/qdk_package/qdk/stim/__init__.py
```diff
@@ -0,0 +1,18 @@
+from ..simulation import run_qir
+from .._native import NoiseConfig, compile_stim_to_qir
+from typing import List, Literal, Optional, Tuple
+
+
+def compile(src: str, noise: Optional[NoiseConfig]) -> Tuple[str, NoiseConfig]:
```
billti: You need to put ` = None` on the optional param, else pylance still complains about `Argument missing for parameter Noise`.

billti: I'm not a fan of this returning a tuple. I keep forgetting to destructure the results and wondering why I have a list. Also `qsharp.compile` and `openqasm.compile` return a `QirInputData`. We should be consistent.

billti: (It's OK if we return the QIR string for now rather than QirInputData, but let's still just return the QIR and not a tuple)

source/compiler/stim_compiler/src/lex.rs
amcasey: I take it this pattern is common enough that readers will understand that "len" implies "byte length"?

source/compiler/stim_compiler/src/lex.rs
amcasey: I think the answer is probably that I just wasn't familiar with the rust idiom for asserting that the assignment to `u32` is valid.

source/compiler/qsc_stim_parser/src/lex.rs
amcasey: So, to confirm, there are no negative angles?  You have to normalize to a positive angle?

source/compiler/stim_compiler/src/lex.rs
amcasey: The regex you provided doesn't seem to allow the identifier to start with a digit.

source/compiler/stim_compiler/src/lex.rs
amcasey: I'm not sure I understand the complexity.  But it looks like maybe the single token includes the contents of the square brackets and isn't a keyword followed by punctuation, etc?

source/compiler/stim_compiler/src/parser.rs
joao-boechat: Negates the classical measurement answer

source/compiler/stim_compiler/src/parser.rs
joao-boechat: Yes

source/simulators/src/gpu_full_state_simulator/common.wgsl
```diff
@@ -356,18 +356,18 @@ fn apply_1q_pauli_noise(shot_idx: u32, op_idx: u32, noise_idx: u32) {
...
-        shot.unitary[4] = op.unitary[0];
-        shot.unitary[5] = op.unitary[1];
-    } else if (rand < (p_x + p_y + p_z)) {
+    } else if (rand < (p_x + p_z)) {
         // Apply Z error (negate |1> state)
         shot.unitary[0] = op.unitary[0];
         shot.unitary[1] = op.unitary[1];
         shot.unitary[4] = cplxNeg(op.unitary[4]);
         shot.unitary[5] = cplxNeg(op.unitary[5]);
+    } else if (rand < (p_x + p_z + p_y)) {
```
billti: Why did this get moved? Was there an issue to fix or optimization to be had?

billti (approved):

source/simulators/src/gpu_full_state_simulator/common.wgsl
orpuente-MS: I think this change got here due to some git shenanigans. I had this change temporarily in my PR before merging it, but I reverted it. Seems like the change got here when this PR was pointing to my PR, but somehow stayed here after rebasing on main?

@joao-boechat you should revert this change.

source/qdk_package/src/qir_simulation.rs
```diff
@@ -271,10 +271,30 @@ fn bind_noise_config<T: Float, Q: Float>(
...
+    for (id, noise_table) in intrinsics {
+        let name = format!("noise_intrinsic_{id}");
+        table.insert(
+            name,
+            (*id, Py::new(py, NoiseTable::from(noise_table.clone()))?),
+        );
+        next_id = next_id.max(*id + 1);
+    }
+    Ok(NoiseIntrinsicsTable { next_id, table })
+}
```
orpuente-MS: This logic is specific to Stim. So, I don't think it works as a general fix `bing_noise_config`. It will be better if you create a specific function called `bind_stim_noise_config` and rename `to_intrinsic_table` to `to_stim_intrinsic_table` and use that for the stim path.

orpuente-MS (approved):

source/compiler/stim_compiler/src/lex.rs
joao-boechat: This is out of scope for this PR, as we're not so focused on language features as of now. But I will save it so we can discuss it in the future, when we get to that.

source/compiler/qsc_stim_parser/src/lex.rs
joao-boechat: The idea is to collapse multiple consecutive newlines into a single one. However, the handling of whitespace / newlines was very confusing and leading to some issues... I have pushed a commit to simplify it.

source/compiler/qsc_stim_parser/src/parser.rs
joao-boechat: The line doesn't start with a newline, it ends with it. But since the line basically is a instruction + \n, and we've already parsed the instruction, I'm expecting the newline and then returning the line after that.

joao-boechat: And regarding file boundaries: yes, it would cause issues, but I've already fixed it by considering the case where the line / block ends with a EOF (self.tokens.peek() == None), and accepting it instead of panicking.

source/compiler/stim_compiler/src/parser.rs
joao-boechat: Good point, but as noted in another comment, this PR isn't focused on language features. I will write it down as a note for a feature follow up.

source/compiler/qsc_stim_parser/src/parser.rs
joao-boechat: It does! Will add better error handling in a follow up PR.

source/compiler/qsc_stim_parser/src/parser.rs
joao-boechat: It is not. I've removed it.

source/compiler/stim_compiler/src/lex.rs
joao-boechat: fixed! thanks for noting this bug.

source/compiler/qsc_stim_parser/src/lex.rs
joao-boechat: Floats can have trailing + or -. That behavior was fixed in a recent commit, thank you!

source/compiler/stim_compiler/src/lex.rs
joao-boechat: True, I got confused by your question at first. But even though this check has "is_alphanumeric", which includes digits, if an identifier starts with a digit, it it going to be matched as a number, and produce an "unknown" (which will in a future PR give back an error)

source/compiler/stim_compiler/src/lex.rs
joao-boechat: Let me explain the complexity:

if we tokenize "rec", "sweep", "[", "]", and the inside content of tags separately, we will have to deal with some issues (two examples I could think of right now):
1- What if variables start with "rec" or "sweep"? (I guess a way to deal with it would be emitting rec and sweep as "identifier" tokens, which could work. But then the parser will have to check the contents of the token regardless, so that would not fix the violation of the lexer / parser boundary)
2- The inside of tags can be anything defined by this: **[^\r\]\n]*** . So they can include whitespace, which could cause weird behaviors where we have to deal with multiple tokens in between brackets

Of course, these are all things we can handle if we spend some time fixing the logic here and there, but I don't want to break the behavior of the lexer and parser right now that we don't have any unit tests yet. So, I will write this down as a code change to implement in a future PR.

source/compiler/stim_compiler/src/lex.rs
joao-boechat: This should deal with them correctly.

source/qdk_package/src/qir_simulation.rs
joao-boechat: Fixed.

source/simulators/src/gpu_full_state_simulator/common.wgsl
joao-boechat: done!

PR #3306, opened 2026-06-11
Bloch Sphere Widget - Command and Dev Playground

ScottCarda-MS:
# Bloch Sphere View for Q# UX (Playground + VS Code)

## Summary
This PR introduces an interactive Bloch sphere view in the shared Q# UX package and wires it into both:

- the dev playground (as a first-class Tools view)
- the VS Code extension webview (via a dedicated command)

It also adds the synthesis data/tooling used by the Bloch Rz helper.

### New Bloch UX surface
- Adds a new Bloch widget implementation in the shared UX layer.
- Includes gate application controls, sequence playback/history, and Rz synthesis integration.
- Includes theme-aware rendering/styling for host environments.

### Shared math and compute cleanup
- Introduces shared quantum math primitives used across Bloch/state visualization paths.
- Refactors relevant state-viz worker computation to align with shared math utilities.

### Rz synthesis artifacts and generator source
- Adds checked-in lookup artifacts:
  - `source/npm/qsharp/rz-array.json`
  - `source/npm/qsharp/rz-details.json`
- Adds generator source/tooling for reproducibility:
  - `source/npm/qsharp/tools/rz-synthesis.ts`
  - `source/npm/qsharp/tools/tsconfig.json`

### Playground integration
- Adds Bloch view routing/rendering in the dev playground.
- Adds navigation entry placement under Tools.

### VS Code integration
- Adds/updates webview command plumbing for Bloch panel display.
- Ensures webview/learning webview type-check configs support shared UX JSON import path.

### Packaging and dependency updates
- Updates npm dependency metadata and lockfiles.
- Updates component governance manifest for shipped dependency compliance.

## Key Files (high signal)
- `source/npm/qsharp/ux/bloch.tsx`
- `source/npm/qsharp/ux/blochGates.ts`
- `source/npm/qsharp/ux/quantum-math.ts`
- `source/npm/qsharp/ux/cplx.ts`
- `source/npm/qsharp/ux/circuit-vis/state-viz/worker/stateCompute.ts`
- `source/npm/qsharp/rz-array.json`
- `source/npm/qsharp/rz-details.json`
- `source/npm/qsharp/tools/rz-synthesis.ts`
- `source/playground/src/nav.tsx`
- `source/playground/src/main.tsx`
- `source/vscode/src/webviewPanel.ts`
- `source/vscode/src/webview/webview.tsx`

## Screenshots
image


amcasey:
Are the colors we want in dark mode?  I find the arrows and the axis labels pretty close together.

Edit: from the code, they seem to have been chosen pretty deliberately.  I find the white on beige a little strange, but it's not substantially hurting readability.

image

Also, this might just be my ignorance, but is `n/a` the appropriate value for phi here?

amcasey:
Were we expecting it to expand with the pane?

image

amcasey:
It seems like the slider at the bottom manipulates Rz in radians but the angles at the top are in degrees?  Maybe that's a feature?

amcasey:
Personally, I'd find it very helpful to see the ending state in the top-level summary (maybe top-right) and not just in the sidebar.

amcasey:
Is the `Reset` button supposed to clear the `Rz` slider too?

source/npm/qsharp/ux/quantum-math.ts
```diff
@@ -0,0 +1,534 @@
...
+      );
+    }
+  }
+
+  mag(): number {
+    return Math.sqrt(this.re ** 2 + this.im ** 2);
+  }
+
+  isUnit(): boolean {
+    return Math.abs(this.mag() - 1.0) < epsilon;
```
amcasey: Out of curiosity, why not use `compare`?

source/npm/qsharp/ux/blochGates.ts
```diff
@@ -0,0 +1,53 @@
...
+  gates: string;
+  modified: boolean;
+} {
+  if (!input) return { gates: "", modified: false };
+  let filtered = "";
+  for (const ch of input) {
+    if (validGateSet.has(ch)) filtered += ch;
+  }
+  const capped =
+    filtered.length > MAX_GATE_SEQUENCE_LENGTH
```
amcasey: I think `slice` does the right thing if the array is shorter

source/npm/qsharp/ux/bloch.tsx
```diff
@@ -0,0 +1,1820 @@
...
+import {
+  MAX_GATE_SEQUENCE_LENGTH,
+  sanitizeGateSequence,
+  VALID_GATE_CODES,
+} from "./blochGates.js";
+
+import rzOps from "../rz-array.json";
+
+// Two color palettes for parts of the scene we draw directly with WebGL
+// (sphere material, label sprites). Picked by eye to stay legible against
```
amcasey: Strictly speaking, there are also contrast ratio requirements for a11y.  (Aside: I'm scared to ask what Narrator has to say about this widget.)

amcasey (commented):
Random drive-by comments.  That's a lot of 3D geometry code to review...

source/npm/qsharp/ux/blochGates.ts
ScottCarda-MS: Easy simplification, thanks!

source/npm/qsharp/ux/quantum-math.ts
ScottCarda-MS: True, that is a little more readable and makes use of our existing logic. Thanks!

source/npm/qsharp/ux/bloch.tsx
ScottCarda-MS: Discussed offline. UI changes are pending for a future PR, so adding aria-labels will come in a future PR.

billti (commented):
This increases the size of the webview.js file in VS Code from 3MB to 4.4MB, which is a pretty sizable increase (~50%) for a file that gets loaded every time a webview opens (such as histogram, circuit viewer, etc.). We should determine if we've happy with this from a size and performance perspective before merging - as we've been pretty agressive on both of those fronts with the QDK to date.

amcasey:
> This increases the size of the webview.js file in VS Code from 3MB to 4.4MB, which is a pretty sizable increase (~50%) for a file that gets loaded every time a webview opens (such as histogram, circuit viewer, etc.). We should determine if we've happy with this from a size and performance perspective before merging - as we've been pretty agressive on both of those fronts with the QDK to date.

If size matters, have we considered minifying it?

ScottCarda-MS:
> If size matters, have we considered minifying it?

Funny you should mention that: https://github.com/microsoft/qdk/pull/3356

amcasey:
Does our target audience already know which vectors lie along the X, Y, and Z axes?

amcasey:
I like the updates, especially the new wireframe.  The experience LGTM, but I haven't reviewed the code.

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/* The math for converting basis coefficients (a, b) to a Bloch-sphere
+   point is:
+     theta = 2 * acos(magnitude(a))
```
amcasey: I'm probably just blanking, but is `magnitude` a typo?  I would have thought we'd pass an angle to `acos`.

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
...
+    if (renderer.current) {
+      renderer.current.rotationTimeMs = DEFAULT_ROTATION_TIME_MS / next;
+    }
+  }
+
+  // Live-text editing of the gate sequence. `draftText === null` means the
+  // textbox mirrors the committed `gates`. While the user is typing,
+  // `draftText` holds their input and is shown immediately; the expensive
+  // trace/sphere update is deferred until they pause (debounced). Input is
+  // sanitized per keystroke, so the textbox only ever holds valid codes.
```
amcasey, line 204: Sanitization seems to happen after a delay (e.g. if I paste text).  Are we confident the bad input isn't observable?

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
...
+    // the regular applyGate path, which hits stale-closure bugs in a tight
+    // setState loop) and open on the latest step, so a linked-to program
+    // shows its final state and the user can add gates without first
+    // overwriting it. They can still Play/step back through the trace.
+    if (props.initialGates) {
+      const { gates: cleaned, modified } = sanitizeGateSequence(
+        props.initialGates,
+      );
+      if (modified) {
+        console.warn(
```
amcasey, line 257: Where does this go?  The output window?

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
...
+    return () => {
+      resizeObserver?.disconnect();
+      themeCleanup();
+      r.dispose();
+      renderer.current = null;
+    };
+  }, []);
+
+  // Derived: per-step trace entries (LaTeX strings) for the whole
+  // `gates` sequence, walking the matrix product forward from |0\u27e9.
```
amcasey: I'm going to assume `\u27e9` is an angle bracket. 😆   I'd probably settle for `>` in this context.

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
...
+
+  /**
+   * Animate one gate, then advance the cursor and chain to the next if
+   * play is still active. The recursive chain captures `pos` per gate.
+   */
+  function playFromIndex(pos: number) {
+    if (!renderer.current) return;
+    const code = gates[pos];
+    const info = gateInfo[code];
+    if (!info) {
```
amcasey, line 530: Found the input sanitization fail-safe.  Thanks!

source/npm/qsharp/ux/bloch/bloch.tsx
amcasey: I found the fail-safe below.  It might be helpful to mention it in the comment.

source/npm/qsharp/ux/bloch/bloch.tsx
```diff
@@ -0,0 +1,1361 @@
...
+    if (dialFrameRef.current !== null) {
+      cancelAnimationFrame(dialFrameRef.current);
+      dialFrameRef.current = null;
+    }
+    setRzAngle(angleFromPointer(e.clientX, e.clientY));
+  }
+
+  // Keyboard support for the dial (focusable, role="slider"). Arrows nudge
+  // one step, PageUp/Down by ten, Home/End to 0 / just under a full turn.
+  function dialKeyDown(e: KeyboardEvent) {
```
amcasey: I couldn't find the control that supported keyboard input.  What's the "dial"?

source/npm/qsharp/ux/qsharp-ux.css
```diff
@@ -535,6 +535,875 @@ modern-normalize (see https://mattbrictson.com/blog/css-normalize-and-reset for
   color: red;
 }
 
+.qs-gate-buttons button {
```
amcasey: I don't know much about CSS, but this seems like an awful lot of it.

source/npm/qsharp/ux/quantum-math.ts
```diff
@@ -0,0 +1,534 @@
...
+    // - Optional leading '-' sign
+    // - Numeric value for real part (optional if imaginary part is present)
+    // - Optional numeric imaginary part with leading [+-] and trailing 'i'
+    //   - If real part is not present, leading '+' on imaginary part is optional
+    // - Numeric parts are an integer followed by an optional decimal part
+    // - If decimal part is present, it must contain at least one digit
+    // - Scientific notation is not supported
+
+    // To ease parsing, look for the real and imaginary parts separately
+    const rePart = /^[-]?(\d+)(\.\d+)?(?=$|[+-])/;
```
amcasey: No `+` on the real part?

amcasey (commented):
Within the limits of my ability to review, LGTM

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: That's referring to the complex magnitude, but I can make that clearer. Normally ratios are passed to arc-trig functions, which then produce angles.

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: This shows up in the console window of the devtools. Since this code path is unreachable when using the tool in vscode, it is specific to the playground, which is strictly a browser experience.

ScottCarda-MS: console.warn calls in general would also surface in the vscode experience as well, via the same devtools window.

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: yah you're right, that can just be a `>`. Fixed!

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: I've changed the comment a bit to try to be more clear on this. Thanks!

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: I updated the comment to name the fail-safe logic so it is easier to find. Thanks!

source/npm/qsharp/ux/quantum-math.ts
ScottCarda-MS: It can't come up now with what uses this code path (the bloch sphere and the state-viz from the circuit editor) as nothing would pass in complex numbers with the + on the real, but it is a small enough change (1 character) to support it in case we want that in the future, so easy enough to add. Done!

source/npm/qsharp/ux/qsharp-ux.css
ScottCarda-MS: It is a lot of css, but this is a fairly UI intensive feature, so it makes sense that there would be a lot of css needed to support the styling we want. It does makes sense, however, given the size of the changes, to split off the css for the bloch sphere into its own file. That also follows the patter used with other UI intensive features as well. Done.

source/npm/qsharp/ux/bloch/bloch.tsx
ScottCarda-MS: Hmm, the keyboard handling is a little weird since switching the Rz input from a slider to a dial. I think it is easiest to remove the keyboard handling. It is less necessary for fine-control now that we have a text input for the angle.

amcasey (approved):
The experience and the code seem fine to me.  We should discuss which release this will be included in.

source/npm/qsharp/ux/index.ts
```diff
@@ -20,6 +20,7 @@ export { SpaceChart } from "./spaceChart.js";
 export { ScatterChart } from "./scatterChart.js";
 export { EstimatesOverview } from "./estimatesOverview.js";
 export { EstimatesPanel } from "./estimatesPanel.js";
+export { BlochSphere } from "./bloch/bloch.js";
```
billti, line 23: Is this intent to still then remove this out of here and into its own path in package.json? Maybe in the "sccarda/webfiew file size" PR (which looks like it needs updating again now after the changes here)

ScottCarda-MS: There were two possible strategies here: lazy-load the Bloch Sphere or separate the Bloch-Sphere into its own webview. This branch went with the latter of those. The sccarda/WebviewFileSize branch does indeed implement the other strategy of lazy-loading, but that might be obsolete since we are separating the Bloch Sphere into its own webview in this branch.

source/vscode/src/learning/webview/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "target": "ES2022",
     "noEmit": true,
     "allowImportingTsExtensions": true,
+    "resolveJsonModule": true,
```
billti, line 7: Why does learning now need to resolve JSON modules? Does it somehow transitively reference the Bloch code? (Or is this in preparation for maybe having the Bloch sphere in the learning content)

source/npm/qsharp/tools/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "outDir": "../dist/tools",
+    "rootDir": "..",
+    "declaration": false,
+    "strict": true,
+    "isolatedModules": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "types": ["node"]
+  },
+  "include": ["./*.ts", "../ux/cplx.ts"]
```
billti: Do you need to explicitly list the files? Doesn't it find the file in the local dir by default and then follow that to the imported module?

source/npm/qsharp/rz-details.json
```diff
@@ -0,0 +1,944 @@
+[
```
billti: Does this file need to be checked in for any reason? Does anything import/use it?

source/npm/qsharp/tools/tsconfig.json
ScottCarda-MS: Good point! Removed

source/vscode/src/learning/webview/tsconfig.json
ScottCarda-MS: Oh this is leftover from an exploration into how the bloch sphere could be used in learning. I though I removed all that content, but must have missed this spot. Thanks! Removed

source/npm/qsharp/rz-details.json
ScottCarda-MS: Since this is just for human-readability for the generated file, it can be removed from the repo. I've added it to the git-ignore as well.

billti (approved):

source/vscode/src/learning/webview/tsconfig.json
ScottCarda-MS: Looks like this is needed after all. The learning module pulls in everything from the qsharp-lang/ux which includes the Bloch Sphere, so it needs this because the Bloch Sphere uses a json file.

PR #3307, opened 2026-06-11
Propagate diverges flag in non-tuple call args

amcasey:
As for tuples, the value of `diverges` should be propagated when converting from `Partial<Ty>` to `Partial<ArgTy>` - otherwise errors may be reported in cases where (e.g.) `fail` would prevent them from being observed at runtime.

### DIFF
--- a/source/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/source/compiler/qsc_frontend/src/typeck/rules.rs
@@ -299,7 +299,7 @@ impl<'a> Context<'a> {
                         let arg_span = inner.span;
                         Partial {
                             ty: ArgTy::Given(ty.ty, arg_span),
-                            diverges: false,
+                            diverges: ty.diverges,
                         }
                     }
                 };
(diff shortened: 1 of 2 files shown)


amcasey:
[Comment](https://github.com/microsoft/qdk/pull/3287#discussion_r3398187404) and [repro](https://microsoft.github.io/qdk/?code=H4sIAAAAAAAACo3OsQqDMBgE4D1PcWRSsEgdHR0KXbq0L5Da3xqq%252BSX5FYv47oUU6eDS%252Be64rxldLZYdKuOTGSXOTlKUuIq37okFWmNVzdY6MSfpt4VFAYi7xtgOmrxnr1PkOS7sDjIOHcHTwF4CYpiBJnKQlsdni7jqzYsCrKBnFrUq9fuqjC82Uob3H7RibyuSY4ad7xZtd2rNRAEmgOaBaqGHWtUHQdesrxIBAAA%253D) from @minestarks

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -299,7 +299,7 @@ impl<'a> Context<'a> {
                         let arg_span = inner.span;
                         Partial {
                             ty: ArgTy::Given(ty.ty, arg_span),
-                            diverges: false,
+                            diverges: ty.diverges,
```
minestarks: Hate to ruin your one-line change ;) but you could refactor this now to only match on the `ty` field, since that's really the only one we're varying here. This way it'll become more clear that the value of `diverges` isn't dependent on the branch

minestarks (approved):

swernli (approved):

PR #3314, opened 2026-06-12
Fix panic in call expr type inference

### LINKED ISSUE #3308: Fuzz: "thread '<unnamed>' (6247) panicked at source/compiler/qsc_frontend/src/typeck/rules.rs:286:29:" (ubuntu-latest)
The fuzz-testing workflow has detected a bug.

<details><summary><b>Auto-Minimized Fuzzing Input That Triggers the Bug:</b> Click this line.</summary>
<b>Note:</b> If the input is multi-line then the end-of-line characters '
' (0x0A) and '
' (0x0D)
may affect the reproducibility of the bug. If you fail to repro the bug with the input shown below
then you may want to go to the <a href=https://github.com/microsoft/qdk/actions/runs/27376450397>workflow</a> that reported this GitHub bug,
download the artifact, and extract the file with the exact minimized input.

```qs

```

</details>

<details><summary><b>The branch/commit the bug has been found in:</b> Click.</summary>
If the developers fail to repro the bug in the latest <code>main</code> then the branch/commit info below can help them to make sure
that they are using the correct way to repro. If the bug is reproducible in the branch/commit below, but not in latest <code>main</code>,
then the bug is likely fixed already or is not in the <code>main</code> branch.

```log
Branch: * main

commit 7ceb9a4a1a60924d6da21a559d45e2cf73bfb008
Author: Andrew Casey <amcasey@users.noreply.github.com>
Date:   Thu Jun 11 13:32:09 2026 -0700

    Update error span for call argument type mismatch (#3287)
    
    It should be just the argument with the problem. Hopefully, this will
    make it easy to add quick fixes to some scenarios.
```

</details>

**Other Info**

- [**Workflow**](https://github.com/microsoft/qdk/actions/runs/27376450397) (contains the run artifacts).
- **Workflow Agent System Info:** `ubuntu-latest: Linux runnervm1li68 6.17.0-1018-azure #18~24.04.1-Ubuntu SMP Thu May 28 16:39:11 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux`.

amcasey:
Some expressions (e.g. Assign) have type Unit unconditionally, unlike a call expression, which goes through type inference even though the return type is declared.  Unit is an empty tuple type, so it was a tuple type but not tuple syntax, which the code didn't handle.  Special case Unit and continue to panic on other unrecognized syntax nodes.

Fixes #3308

### DIFF
--- a/source/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/source/compiler/qsc_frontend/src/typeck/rules.rs
@@ -277,17 +277,14 @@ impl<'a> Context<'a> {
                     // If the outermost element is a tuple, we must wrap it in an `ArgTy::Tuple`.
                     let arg_ty = match ty.ty {
                         Ty::Tuple(tys) => {
-                            // Some non-tuple expressions don't go through inference - they just statically have type Unit:
-                            // Assign, AssignOp, AssignUpdate, Block (and maybe more in the future).  This means that they
-                            // have tuple type early enough to hit this branch but aren't tuple expressions (which could lead
-                            // to a panic below).  However, the empty tuple ArgTy doesn't actually need any spans, so we can
-                            // just skip the computation in that case, avoiding the whole problem.
-                            let spans: Vec<_> = if tys.is_empty() {
-                                Vec::new()
-                            } else if let ExprKind::Tuple(items) = inner.kind.as_ref() {
+                            let spans: Vec<_> = if let ExprKind::Tuple(items) = inner.kind.as_ref()
+                            {
                                 items.iter().map(|item| item.span).collect()
                             } else {
-                                panic!("unexpected syntax kind: {:?}", inner.kind)
+                                // There are lots of expressions that are statically known to have a tuple type
+                                // (include surprising things like Assign, which return Unit - the empty tuple type).
+                                // In those cases, just use the span of the whole expression.
+                                vec![inner.span; tys.len()]
                             };
                             ArgTy::Tuple(
                                 tys.into_iter()
--- a/source/compiler/qsc_frontend/src/typeck/rules.rs
+++ b/source/compiler/qsc_frontend/src/typeck/rules.rs
@@ -277,8 +277,14 @@ impl<'a> Context<'a> {
                     // If the outermost element is a tuple, we must wrap it in an `ArgTy::Tuple`.
                     let arg_ty = match ty.ty {
                         Ty::Tuple(tys) => {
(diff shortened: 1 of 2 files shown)


source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -277,8 +277,14 @@ impl<'a> Context<'a> {
                     // If the outermost element is a tuple, we must wrap it in an `ArgTy::Tuple`.
                     let arg_ty = match ty.ty {
                         Ty::Tuple(tys) => {
-                            let spans: Vec<_> = if let ExprKind::Tuple(items) = inner.kind.as_ref()
-                            {
+                            // Some non-tuple expressions don't go through inference - they just statically have type Unit:
+                            // Assign, AssignOp, AssignUpdate, Block (and maybe more in the future).  This means that they
```
minestarks: Does `Block` statically have type unit? I thought it has the type of the last expression (without a semicolon) in the block.

source/compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -5564,3 +5564,102 @@ fn complex_literal_does_not_support_mod() {
         "##]],
     );
 }
+
+#[test]
+fn call_expr_unit_arg_assign() {
+    check_allow_parse_errors(
+        indoc! {"
+            namespace A {
+                function Foo(u: Unit) : Unit {
```
minestarks, line 5573: Is there a difference, for the purposes of this fix, between a function that takes a `Unit` argument and one that takes no arguments?

minestarks (approved):

source/compiler/qsc_frontend/src/typeck/tests.rs
minestarks: I found that we hit the panic regardless of the signature, so I guess the answer is no.

source/compiler/qsc_frontend/src/typeck/rules.rs
amcasey: No, only empty block.  The comment is confusing, but the regression test captures it appropriately.

amcasey: I've removed the comment anyway.

source/compiler/qsc_frontend/src/typeck/tests.rs
amcasey: I thought it would be nice if there were no other errors so we could see it succeed once fixed.

swernli (approved):

minestarks (approved):

PR #3323, opened 2026-06-12
Use a single 10-minute timeout for all VS Code integration tests

### LINKED ISSUE #3267: CI failure: windows-latest (26464086886) (Timeout in debugger tests)
See the corresponding [**workflow run**](https://github.com/microsoft/qdk/actions/runs/26464086886) for failure details, update the issue with any relevant information, and assign to the appropriate team member(s).

minestarks:
Fixes #3267

Short, ad-hoc timeouts could trip on slow machines and cause flaky failures. Timeouts should only guard against hung tests, never be short enough for a slow machine to hit.

### DIFF
--- a/source/vscode/test/suites/debugger/index.browser.ts
+++ b/source/vscode/test/suites/debugger/index.browser.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { runMochaTests } from "../runBrowser";
+import { TEST_TIMEOUT_MS } from "../extensionUtils";
 
 export function run(): Promise<void> {
   return runMochaTests(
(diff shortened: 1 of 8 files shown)


swernli (approved):

amcasey:
Is there a danger the entire test run will time out and we won't get useful results?

minestarks:
> Is there a danger the entire test run will time out and we won't get useful results?

Yes, the entire test run could time out if we have enough tests, but I would call that still a useful result, because you'd be able to see from the logs that all your tests are timing out.

amcasey:
> > Is there a danger the entire test run will time out and we won't get useful results?
> 
> Yes, the entire test run could time out if we have enough tests, but I would call that still a useful result, because you'd be able to see from the logs that all your tests are timing out.

As we discussed offline, some CI systems don't give you artifacts if the whole run times out since artifact upload has to happen within the timeout period.  I don't know if that's true of this one.

PR #3324, opened 2026-06-12
Extend Pauli strings in NoiseConfig to include loss

orpuente-MS:
Previously you could specify correlated Pauli noise in the `NoiseConfig` object as follows:

```python
noise = NoiseConfig()
noise.cz.ZX = 0.01
noise.cz.IX = 0.01
```

With this change, you can also use the `L` character to specify correlated noise together with the Pauli terms:

```python
noise = NoiseConfig()
noise.cz.LX = 0.01
noise.cz.IL = 0.02
noise.cz.LL = 0.03
```


source/qdk_package/tests/test_clifford_simulator.py
```diff
@@ -319,7 +319,7 @@ def test_clifford_run_mixed_noise():
     result = [result_array_to_string(cast(Sequence[Result], x)) for x in output]
     print(result)
     # Reasonable results obtained from manual run
-    assert result == ["00000-0000000001"]
+    assert result == ["000000000--000-0"]
```
billti: This result for 1 shot really changed quite a bit. We really need to move to statistical validation rather than one shot validation to make this tests meaninful and maintainable.

Not important for this PR, but please consider how we might fix this generally.

orpuente-MS: Fixed. Added an `expect_distribution` utility to make writing these tests possible. Got rid of the flaky tests for now, we can rewrite other tests to use `expect_distribution` in a follow up PR.

source/simulators/src/bin/gpu-runner.rs
```diff
@@ -6,6 +6,8 @@
 // Run with: cargo run --bin gpu-runner [--release]
 // Build with: cargo build --bin gpu-runner [--release]
 
+#![allow(dead_code)]
```
billti: Did you add some dead code? Why do we want to allow this now?

orpuente-MS: We don't. I added it while debugging the file, forgot to remove it after.

Removed now.

billti (approved):

PR #3330, opened 2026-06-13
Add a code fix for should-have-been-array

amcasey:
It's a pretty common mistake to pass a single control qubit to a function or functor that expects an array of them.  Add a code fix detecting that this has occurred and offering to add square brackets.

To make this possible, attach info about the expected and actual types to the ErrorKind.  (Unfortunately, Ty is hard to share, so create a simpler TyInfo that can accompany the ErrorKind.)

### DIFF
--- a/source/compiler/qsc_frontend/src/typeck.rs
+++ b/source/compiler/qsc_frontend/src/typeck.rs
@@ -22,7 +22,7 @@ use qsc_ast::ast::NodeId;
 use qsc_data_structures::{index_map::IndexMap, span::Span};
 use qsc_hir::{
     hir::{CallableKind, ItemId},
-    ty::{FunctorSet, GenericArg, Ty, Udt},
+    ty::{FunctorSet, GenericArg, InferTyId, Prim, Ty, Udt},
 };
 use rustc_hash::FxHashMap;
 use std::fmt::Debug;
@@ -47,11 +47,53 @@ pub struct Table {
 #[error(transparent)]
 pub(super) struct Error(ErrorKind);
 
+/// Simplified type info for error reporting. Same shape as `Ty`, but without `Rc`
+/// so it can be included in `ErrorKind` (which must be `Send + Sync`).
+#[derive(Clone, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
+pub enum TyInfo {
+    Array(Box<TyInfo>),
+    Arrow,
+    Infer(InferTyId),
+    Param,
+    Prim(Prim),
+    Tuple(Vec<TyInfo>),
+    Udt,
+    #[default]
+    Err,
+}
+
+impl From<&Ty> for TyInfo {
+    fn from(ty: &Ty) -> Self {
+        match ty {
+            Ty::Array(item) => TyInfo::Array(Box::new(TyInfo::from(item.as_ref()))),
+            Ty::Arrow(_) => TyInfo::Arrow,
+            Ty::Infer(id) => TyInfo::Infer(*id),
+            Ty::Param { .. } => TyInfo::Param,
+            Ty::Prim(prim) => TyInfo::Prim(*prim),
+            Ty::Tuple(items) => TyInfo::Tuple(items.iter().map(TyInfo::from).collect()),
+            Ty::Udt(_, _) => TyInfo::Udt,
+            Ty::Err => TyInfo::Err,
+        }
+    }
+}
+
+impl From<Ty> for TyInfo {
+    fn from(ty: Ty) -> Self {
+        TyInfo::from(&ty)
+    }
+}
+
 #[derive(Clone, Debug, Diagnostic, Error)]
 enum ErrorKind {
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.TypeCk.TyMismatch"))]
-    TyMismatch(String, String, #[label] Span),
+    TyMismatch(
+        /*expected*/ String,
+        /*actual*/ String,
+        /*expected*/ TyInfo,
+        /*actual*/ TyInfo,
+        #[label] Span,
+    ),
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.TypeCk.CallableMismatch"))]
     CallableMismatch(CallableKind, CallableKind, #[label] Span),
@@ -192,6 +234,17 @@ enum ErrorKind {
     ))]
     #[diagnostic(code("Qsc.TypeCk.RecursiveTypeConstraint"))]
     RecursiveTypeConstraint(#[label] Span),
+}
+
+impl Error {
+    /// If this is a type-mismatch error, returns (expected, actual, span).
+    #[must_use]
+    pub fn ty_mismatch(&self) -> Option<(&TyInfo, &TyInfo, Span)> {
+        match &self.0 {
+            ErrorKind::TyMismatch(_, _, expected, actual, span) => Some((expected, actual, *span)),
+            _ => None,
+        }
+    }
 }
 
 impl From<TyConversionError> for Error {
--- a/source/compiler/qsc_frontend/src/typeck.rs
+++ b/source/compiler/qsc_frontend/src/typeck.rs
@@ -49,7 +49,7 @@ pub(super) struct Error(ErrorKind);
 
 /// Simplified type info for error reporting. Same shape as `Ty`, but without `Rc`
 /// so it can be included in `ErrorKind` (which must be `Send + Sync`).
-#[derive(Clone, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub enum TyInfo {
     Array(Box<TyInfo>),
     Arrow,
--- a/source/compiler/qsc_frontend/src/typeck.rs
+++ b/source/compiler/qsc_frontend/src/typeck.rs
@@ -50,29 +50,57 @@ pub(super) struct Error(ErrorKind);
 /// Simplified type info for error reporting. Same shape as `Ty`, but without `Rc`
 /// so it can be included in `ErrorKind` (which must be `Send + Sync`).
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
-pub enum TyInfo {
-    Array(Box<TyInfo>),
+pub enum TyInfoKind {
+    Array(Box<TyInfoKind>),
     Arrow,
     Infer(InferTyId),
     Param,
     Prim(Prim),
-    Tuple(Vec<TyInfo>),
+    Tuple(Vec<TyInfoKind>),
     Udt,
     #[default]
     Err,
 }
 
-impl From<&Ty> for TyInfo {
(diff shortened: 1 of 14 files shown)


source/language_service/src/code_action/wrap_in_array/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+";
+    let actions = get_wrap_in_array_actions(source);
+    assert!(actions.is_empty(), "Expected no actions, got: {actions:?}");
+}
+
+#[test]
+fn array_to_nested_array() {
+    // Qubit[] passed where Qubit[][] expected - wrapping in [...] is valid.
+    // Qubit[] isn't a primitive type, but it's actually complaining at the element
```
amcasey: I don't love this, but I wasn't convinced it was worth "breaking" it for consistency.  If we care, I think we can ask the compiler for the type of the expression at the error span and force that to be a primitive too.

source/compiler/qsc_frontend/src/typeck.rs
```diff
@@ -47,11 +47,53 @@ pub struct Table {
...
+        TyInfo::from(&ty)
+    }
+}
+
 #[derive(Clone, Debug, Diagnostic, Error)]
 enum ErrorKind {
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.TypeCk.TyMismatch"))]
-    TyMismatch(String, String, #[label] Span),
+    TyMismatch(
```
amcasey, line 118: We talked about adding a new ErrorKind, but I thought enriching this one gave us more flexibility when creating future code actions.

source/language_service/src/code_action/wrap_in_array/tests.rs
amcasey: I did it, but I'd just as soon revert the new commit.

source/compiler/qsc_frontend/src/typeck.rs
```diff
@@ -47,11 +47,53 @@ pub struct Table {
 #[error(transparent)]
 pub(super) struct Error(ErrorKind);
 
+/// Simplified type info for error reporting. Same shape as `Ty`, but without `Rc`
+/// so it can be included in `ErrorKind` (which must be `Send + Sync`).
+#[derive(Clone, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
```
minestarks: Do you need all these derive impls? `PartialOrd` and `Ord` seem especially unnecessary (unless you're doing something unexpected). Try removing them and leave them out if code still compiles.

source/compiler/qsc_frontend/src/typeck.rs
```diff
@@ -47,11 +47,53 @@ pub struct Table {
...
 #[derive(Clone, Debug, Diagnostic, Error)]
 enum ErrorKind {
     #[error("expected {0}, found {1}")]
     #[diagnostic(code("Qsc.TypeCk.TyMismatch"))]
-    TyMismatch(String, String, #[label] Span),
+    TyMismatch(
+        /*expected*/ String,
+        /*actual*/ String,
+        /*expected*/ TyInfo,
+        /*actual*/ TyInfo,
```
minestarks: Instead of having parallel expected/actual fields like this, I wonder if `TyInfo` should just include the display name. (You may need to implement `Display` for `TyInfo` so that the `"expected {0}, found {1}"` format string works).

source/compiler/qsc_frontend/src/typeck.rs
amcasey: Removed PartialOrd and Ord.

source/compiler/qsc_frontend/src/typeck.rs
amcasey: Would you put it in the enum or wrap the enum in a struct combining TyInfo[Kind?] and the display string?

source/language_service/src/code_action/wrap_in_array/tests.rs
```diff
@@ -0,0 +1,168 @@
...
+        Position {
+            line: 0,
+            column: u32::try_from(source.len()).expect("len fits"),
+        }
+    } else {
+        Position {
+            line: newline_count,
+            column: 0,
+        }
+    };
```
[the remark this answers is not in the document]
amcasey: @minestarks I think the helper you mentioned was probably `compile_project_with_markers_no_cursor` or something like that and this test helper is already using it.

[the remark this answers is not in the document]
amcasey: Also, this same newline_count approach is used in the auto_import tests.  Both presumably stole it from the wrapper_refactor tests.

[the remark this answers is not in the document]
amcasey: This was just a very bad way to do `Range::from_span`.  Probably, it should also use range markers, in case there are new code fixes in the future.

source/language_service/src/code_action/wrap_in_array.rs
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Code action: "Convert to single-element array"
+// Detects when a value is passed where an array of that type is expected,
+// and offers to wrap it in `[...]`.
```
minestarks: Not sure if that'd make it more discoverable, but you _could_ make this a module level doc comment.

```suggestion
//! Code action: "Convert to single-element array"
//! Detects when a value is passed where an array of that type is expected,
//! and offers to wrap it in `[...]`.
```

source/compiler/qsc_frontend/src/typeck.rs
minestarks: Hmm, good question, I would either make a second type (`TyInfo` with display name and kind, and `TyInfoKind` which is just the kind) _or_ simply use a `(TyInfo, String)` tuple

source/language_service/src/code_action/wrap_in_array.rs
```diff
@@ -0,0 +1,116 @@
...
+            code_actions.push(CodeAction {
+                title: "Convert to single-element array".to_string(),
+                edit: Some(WorkspaceEdit {
+                    changes: vec![(source_name.to_string(), vec![TextEdit { new_text, range }])],
+                }),
+                kind: Some(CodeActionKind::QuickFix),
+                is_preferred: None,
+            });
+        }
+    }
```
minestarks: Okay, I think this is how I would be inclined to implement this - it feels more intuitive and more efficient to me. It also employs some common helpers/best practices we use in the language service, which I'm glad to demonstrate.

- Speaks in terms of "offset-relative-to-package-start" ("package offsets") rather than "offset-relative-to-source-start", which generally results in less friction and fewer conversions when dealing with the compiler.

- We speak in terms of these offsets until the very last moment, where we have to return these locations to the VS Code langauge service, at which point we convert to `Range` . We have a helper for this, `into_range`.

- I chose to return two text edits to simply insert `[` and `]` around the error span, rather than reading the source and formatting a brand new string. This is more efficient if you're splitting hairs with performance, and IMO it's just more elegant to avoid dealing with literal source strings where you can. 




```suggestion
            let open_range = into_range(
                encoding,
                Span {
                    lo: error_span.lo,
                    hi: error_span.lo,
                },
                &unit.sources,
            );

            let close_range = into_range(
                encoding,
                Span {
                    lo: error_span.hi,
                    hi: error_span.hi,
                },
                &unit.sources,
            );

            code_actions.push(CodeAction {
                title: "Convert to single-element array".to_string(),
                edit: Some(WorkspaceEdit {
                    changes: vec![(
                        source_name.to_string(),
                        vec![
                            TextEdit {
                                new_text: "[".to_string(),
                                range: open_range,
                            },
                            TextEdit {
                                new_text: "]".to_string(),
                                range: close_range,
                            },
                        ],
                    )],
                }),
                kind: Some(CodeActionKind::QuickFix),
                is_preferred: None,
            });
        }
    }
```

source/language_service/src/code_action/auto_import/tests.rs
minestarks, line 1: I'm confused by the changes to this file. Is this a merge artifact?

minestarks: Oh, is this you tracking down where this newline count stuff came from?

minestarks: Good change, thanks (I probably should have caught this before your PR)! I might have kept the `whole_document_range` helper and changed its implementation rather than repeating the same two lines everywhere though.

minestarks (approved):

source/language_service/src/code_action/wrap_in_array.rs
amcasey: `into_range` appears to call `Range::from_span` after re-finding the relevant `Source`(s), so I'm not sure I understand the distinction.

source/language_service/src/code_action/wrap_in_array/tests.rs
amcasey: @minestarks You mentioned performance below and that's one of the things that's bothers me about the extra visitor pass - it seems like it would delay the should-show-lightbulb computation.  Maybe it's fast enough to not matter?

source/language_service/src/code_action/auto_import/tests.rs
amcasey: Restored `whole_document_range`.

source/compiler/qsc_frontend/src/typeck.rs
amcasey: Gave it a named type since it's used in so many places.  (Let me know if that was overkill and a tuple would have been better.)

source/language_service/src/code_action/wrap_in_array.rs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            // Generate the fix: wrap the expression in [...]
+            // Note that this depends on the error span excluding surrounding parens
+            // so we don't end up with something like `F[(q)]`.
+            let lo = (error_span.lo - source.offset) as usize;
+            let hi = (error_span.hi - source.offset) as usize;
+            let arg_text = &source.contents[lo..hi];
+            let new_text = format!("[{arg_text}]");
+            let range = Range::from_span(encoding, &source.contents, &(error_span - source.offset));
```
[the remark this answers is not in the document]
amcasey: We're not really worried about underflow in cases like this.

source/language_service/src/code_action/wrap_in_array/tests.rs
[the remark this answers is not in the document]
amcasey: I decided against the range markers - it's interesting to know whether spurious fixes were offered in other parts of the file.

source/language_service/src/code_action/wrap_in_array/tests.rs
amcasey: The perf doesn't matter and the tiny consistency improvement seems worthwhile.

amcasey:
Converted to draft so it doesn't get merged.  This is mostly for discussion at this point.

minestarks (approved):

PR #3335, opened 2026-06-15
Implement `PostSelectZ` for Clifford simulation

swernli:
This change adds support for simulation time post-selection in the Z basis when using the Clifford simulator. It also adds `postselectz` to `qdk.inc` so OpenQASM code can call it as well.

### DIFF
--- a/source/paulimer/src/outcome_specific_simulation.rs
+++ b/source/paulimer/src/outcome_specific_simulation.rs
@@ -133,6 +133,9 @@ pub fn measure_pauli_with_hint<HintBits: PauliBits, HintPhase: PhaseExponent>(
         pauli *= Phase::from_exponent(3u8.wrapping_sub(preimage.xz_phase_exponent().raw_value()));
         PauliExponent::new(pauli) * &mut simulation.clifford;
         if let Some(forced) = forced_bit {
+            // Since we have a forced bit, use that as ther result instead of a random one.
+            // To keep the internal state consistent, we also mark this as a random outcome,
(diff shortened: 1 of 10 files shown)


source/paulimer/src/outcome_specific_simulation.rs
```diff
@@ -127,7 +132,13 @@ pub fn measure_pauli_with_hint<HintBits: PauliBits, HintPhase: PhaseExponent>(
         let mut pauli = observable.clone() * hint;
         pauli *= Phase::from_exponent(3u8.wrapping_sub(preimage.xz_phase_exponent().raw_value()));
         PauliExponent::new(pauli) * &mut simulation.clifford;
-        allocate_random_bit(simulation);
+        if let Some(forced) = forced_bit {
+            simulation.outcome_vector.push(forced);
+            simulation.random_outcome_indicator.push(true);
+            simulation.num_random_bits += 1;
```
billti, line 141: Can you add a comment above these 3 lines for a brief explanation. At first glance, it seem counter-intuitive that forcing an outcome would increase the number of random bits.

swernli: I'll add a comment about keeping the data structures consistent, but truth be told the `random_outcome_indicator` is effectively unused in this pattern of using paulimer. So if we are willing to deviate further from the original implementation, we could just remove it.

orpuente-MS (approved):

swernli:

PR #3343, opened 2026-06-16
Auto-approve safe cargo commands

amcasey:


.vscode/settings.shared.json
```diff
@@ -43,5 +43,12 @@
     "[yaml]": {
       "editor.defaultFormatter": "esbenp.prettier-vscode",
       "editor.formatOnSave": true
+    },
+    "chat.tools.terminal.autoApprove": {
+      "cargo build": true,
+      "cargo check": true,
+      "cargo clippy": true,
+      "cargo fmt": true,
+      "cargo test": true
```
minestarks: I really have no intuition for how to set these, but this one (`cargo test`) seems like the odd one out, in that it _could_ run arbitrary code

amcasey:
Here's copilot's opinion of how far you should trust it:

Safe to auto-approve (read-only or local build operations):

cargo build — compiles the project locally
cargo check — type-checks without producing binaries
cargo test — runs tests locally
cargo clippy — runs the linter
cargo fmt / cargo fmt --check — formats or checks formatting
cargo doc — generates documentation locally
cargo bench — runs benchmarks locally
cargo clean — removes the target/ directory
cargo metadata — prints project metadata as JSON
cargo tree — prints dependency tree
cargo verify-project — checks manifest correctness
cargo read-manifest — prints manifest info
cargo locate-project — prints Cargo.toml path
cargo version — prints version

Use caution / don't auto-approve:

cargo publish — uploads a crate to crates.io
cargo install — installs binaries globally
cargo yank — yanks a published version
cargo login — stores API credentials
cargo owner — manages crate ownership

.vscode/settings.shared.json
orpuente-MS: `cargo build` could also run arbitrary code, since rust has `build.rs` scripts which trigger with `cargo build`. We have a few of those in our repo.

amcasey: So are folks reviewing tests generated by copilot before letting them run?  Sounds sensible, but I admit I haven't been validating at that level.

amcasey: I'm happy to drop the PR, but I'm interested in learning more about what kind of review people do before letting copilot build and test changes.

PR #3344, opened 2026-06-16
Add initial support for simple IR functions in QIR

idavis:
# Add support for simple IR functions in QIR

## What

Adds support for emitting eligible user-package Q# callables as IR functions instead of inlining them in QIR codegen. This lays the groundwork for extending IR function support to non-user packages in follow-up PRs.

## Why

Currently, all Q# callables are inlined into the entry point. By emitting simple eligible callables as IR functions, we:

- Reduce code duplication when callables are invoked multiple times
- Establish infrastructure for more sophisticated IR function handling in future work
- Enable better code generation and optimization in QIR targets

### Partial Evaluator Changes (`qsc_partial_eval`)

- Identifies eligible callables: non-composite specializations returning `Unit` or scalars (`Int`/`Double`/`Bool`)
- Emits these as `Regular` RIR callables with bodies instead of inlining
- Adds comprehensive tests in new `ir_functions.rs` test module

### RIR Builder & Codegen (`qsc_rir`, `qsc_codegen`)

- Extended `qsc_rir::builder` to support RIR function definitions with proper block management
- Added `qir::name` module for correct LLVM global symbol name formatting (handling special characters and quoting)
- Updated QIR v2 code generation to handle IR function calls and definitions
- Added tests covering function call patterns and dominator graph updates

### Supporting Changes

- Updated resource counting analyzer (`qsc_rca`) to handle IR function bodies
- Updated circuit generation, capabilities checker, and RIR passes for IR function compatibility
- Added test utilities for multi-body RIR programs

## Notes

- This PR handles user-package callables only. Support for non-user packages is deferred to follow-up PRs as specified.
- SSA Multi-Body is valuable because it makes SSA code correct for programs with more than one callable body, instead of relying on a fragile single-body assumption; that improves reliability today and avoids refactoring later as more IR-function/capability paths move through SSA.


source/compiler/qsc_partial_eval/src/tests.rs
```diff
@@ -181,6 +182,25 @@ pub fn get_rir_program_with_capabilities(
     }
 }
 
+/// Partially evaluates a program targeting the `Adaptive` profile, which enables the `CallSupport`
+/// capability that gates IR-function emission.
+#[must_use]
+pub fn get_rir_program_with_adaptive_profile(source: &str) -> Program {
+    get_rir_program_with_capabilities(source, Profile::Adaptive.into())
```
swernli, line 189: we have other tests (mostly in loops.rs, but a few others too) that just call `get_rir_program_with_capabilities` passing adaptive. I think we should either update those to use this helper or just remove the helper and call the function with the parameter directly, for consistency. Same with the below.

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -1562,6 +1590,26 @@ impl<'a> PartialEvaluator<'a> {
...
+        // parameters, so the operand mapping below cannot encounter composite values.
+        let ir_function_arg_operands = spec_decl
+            .filter(|spec_decl| {
+                self.is_ir_function_eligible(store_item_id, functor_app, spec_decl, callable_decl)
+            })
+            .map(|_| {
+                args.iter()
+                    .map(|arg| {
+                        let value = match arg {
+                            Arg::Discard(value) => value,
```
swernli, line 1606: This isn't really specific to your changes, but looking at how `Arg::Discard` is used, I don't think it needs to be handled here. We would never be emitting a call to an ir function with a discard, as that concept doesn't exist in LLVM. And even for existing usage, I feel like the right thing to do would be to transform away discards at an earlier step, but such a transform doesn't exist today. Perhaps we can consider something like that in the future with an FIR transform (especially post tuple decomposition), but for now leaving this is in is fine. Perhaps worth a follow up investigation issue or something?

source/compiler/qsc_rir/src/passes/test_utils.rs
```diff
@@ -0,0 +1,39 @@
...
+        Err(payload) => match payload.downcast::<&str>() {
+            Ok(message) => (*message).to_string(),
+            Err(_) => "(non-string panic payload)".to_string(),
+        },
+    };
+    assert!(
+        message.contains(expected_substring),
+        "panic message did not contain the expected substring.\n  expected substring: {expected_substring}\n  actual message: {message}"
+    );
+}
```
swernli, line 39: this is text-identical to the `assert_panics_with` in the qsc_fir_transforms crate. I wonder if there is a reasonable place to put this so that both can depend on it? Maybe qsc_data_structures?

source/compiler/qsc/src/codegen/tests/adaptive_profile.rs
```diff
@@ -0,0 +1,2099 @@
...
+
+        declare ptr @__quantum__rt__qubit_allocate()
+
+        declare void @__quantum__qis__x__body(ptr)
+
+        declare void @__quantum__rt__qubit_release(ptr)
+
+        declare void @__quantum__rt__tuple_record_output(i64, ptr)
+
+        attributes #0 = { "entry_point" "output_labeling_schema" "qir_profiles"="adaptive_profile" "required_num_qubits"="0" "required_num_results"="0" }
```
swernli: it occurs to me looking at this one that the `required_num_qubits` is 0 by design, but I wonder if might make more sense to make that whole part of the attributes somethign that gets filled in so we can skip it when dynamic qubit allocation is used. Double checking the spec, it says the attribute is "not required" when using the dynamic qubits extension. Likewise, the `dynamic_qubit_management` flag below should get set to true. Not critical right now as this feature is not yet exposed, so it can get included now or have a follow up issue filed.

swernli: alternatively, we can wait until it gets incorporated into `Adaptive` by default and then just hard-code true and drop the `required_num_qubits` entirely.

source/compiler/qsc_partial_eval/src/tests/ir_functions.rs
```diff
@@ -0,0 +1,725 @@
...
+    expected.assert_eq(&format!("{:#?}", ir_function_names(program)));
+}
+
+#[test]
+fn eligible_void_operation_called_twice_emits_one_callable_and_two_calls() {
+    let program = get_rir_program_with_adaptive_profile(
+        r#"
+        namespace Test {
+            operation ApplyX(q : Qubit) : Unit {
+                Microsoft.Quantum.Intrinsic.X(q);
```
swernli: The tests in this file use fully qualified names for the stdlib gates, but that namespace is in the prelude so that is not needed.

source/compiler/qsc_partial_eval/src/tests/ir_functions.rs
```diff
@@ -0,0 +1,725 @@
...
+#[test]
+fn composite_tuple_parameter_callee_is_inlined() {
+    let program = get_rir_program_with_adaptive_profile(
+        r#"
+        namespace Test {
+            operation ApplyToPair(qs : (Qubit, Qubit)) : Unit {
+                let (a, b) = qs;
+                Microsoft.Quantum.Intrinsic.X(a);
+                Microsoft.Quantum.Intrinsic.X(b);
+            }
```
swernli, line 250: remind me, why doesn't this get handled by FIR transforms and changed to something like `operation ApplyToPair(qs0 : Qubit, qs1 : Qubit) : Unit`?

source/compiler/qsc_rir/src/passes/remap_block_ids/tests.rs
```diff
@@ -654,3 +661,209 @@ fn remap_block_ids_ensures_acyclic_program_gets_topological_ordering() {
...
+                capabilities: Base
+            num_qubits: 0
+            num_results: 0
+            tags:
+    "#]]
+    .assert_eq(&program.to_string());
+}
+
+#[test]
+fn remap_block_ids_single_body_unchanged_regression() {
```
swernli: I don't think this separate regression test is really needed... the above tests cover many combinations of single body programs and show that the behavior there didn't change. Outside the context of this PR the regression test would seem out of place.

source/compiler/qsc_rir/src/passes/ssa_check/tests.rs
```diff
@@ -1095,9 +1121,229 @@ fn ssa_check_fails_when_phi_blocks_have_different_predecessors() {
...
+                ty: Ty::Prim(Prim::Integer),
+            }))),
+        ]),
+    );
+
+    perform_ssa_check(&mut program);
+}
+
+#[test]
+fn ssa_check_single_body_unchanged_regression() {
```
swernli: Here too this regression test can be removed, as the `bell_program` is checked above in `ssa_check_passes_for_base_profile_program`

source/compiler/qsc_rir/src/passes/ssa_transform/tests.rs
```diff
@@ -2959,3 +2987,500 @@ fn ssa_transform_propagates_phi_var_to_successor_blocks_across_sequential_branch
...
+            config: Config:
+                capabilities: TargetCapabilityFlags(Adaptive | IntegerComputations | FloatingPointComputations)
+            num_qubits: 0
+            num_results: 0
+            tags:
+    "#]].assert_eq(&program.to_string());
+}
+
+#[test]
+fn ssa_transform_single_body_unchanged_regression() {
```
swernli: this one can be removed too, tested above in `ssa_transform_leaves_program_without_store_instruction_unchanged`

source/compiler/qsc_rir/src/passes/ssa_transform.rs
```diff
@@ -17,25 +17,66 @@ pub fn transform_to_ssa(program: &mut Program, preds: &IndexMap<BlockId, Vec<Blo
...
+    input_vars: &[(VariableId, Ty)],
+    preds: &IndexMap<BlockId, Vec<BlockId>>,
+    next_var_id: &mut VariableId,
+) {
+    // Process the entry block together with every block reachable from it, in ascending block-id
+    // order so that each block is handled after its predecessors.
+    let mut body_blocks = get_all_block_successors(entry, program);
+    body_blocks.push(entry);
+    body_blocks.sort_unstable();
+    body_blocks.dedup();
```
swernli: I think you can simplify this. Because this runs after the block ids have been remapped and the whole graph checked to be acyclic, we know that `entry` will be the lowest id block for a function and not already in the successors list, so it will unconditionally go to the front. So this should work (passed tests when I tried it):

```suggestion
    let mut body_blocks = vec![entry];
    body_blocks.append(&mut get_all_block_successors(entry, program));
```

source/compiler/qsc_rir/src/passes/ssa_transform.rs
```diff
@@ -157,7 +202,20 @@ fn map_store_to_dominated_ssa(
...
             _ => FxHashMap::default(),
         };
-        map_variable_use_in_block(block, &mut var_map, &FxHashSet::default());
+        if block_id == entry {
+            // Each parameter is its own initial SSA version, defined at the body entry.
+            for &(var_id, ty) in input_vars {
+                var_map.entry(var_id).or_insert(Operand::Variable(Variable {
+                    variable_id: var_id,
+                    ty,
+                }));
```
swernli: Each block is processed once and the entry is processed first with no phi nodes, so these entries must be fresh. To defend that here, we can use

```suggestion
                assert!(
                    var_map
                        .insert(
                            var_id,
                            Operand::Variable(Variable {
                                variable_id: var_id,
                                ty,
                            }),
                        )
                        .is_none(),
                    "input vars should only be initialized once by parameters"
                );
```

source/compiler/qsc_partial_eval/src/lib.rs
swernli: we talked about this a bit, and we probably still want to emit these. A test to confirm would be helpful, and then the argument should get passed even if it doesn't get used in the target function.

source/compiler/qsc_partial_eval/src/tests/ir_functions.rs
idavis: The test only documents the partial evaluator's standalone fallback behavior when it's handed an un-promoted tuple parameter

source/compiler/qsc_rir/src/passes/test_utils.rs
idavis: We can follow up with a refactor.

swernli (approved):

PR #3346, opened 2026-06-16
[DRAFT] make QRE deterministic

fedimser:
Why?
* QRE, when translating Cirq circuit to trace and encountering classically-controlled gate, does this:
```
elif isinstance(op, ClassicallyControlledOperation):
    if random.random() < self.classical_control_probability:
        self.handle_op(op.without_classical_controls())
```
* When we have n IAND gates, they on average produce `0.5*n` CZ gates. QRE will not count exactly `0.5*n`, but it will count some number close to 0.5*n.
* Let's say I want to optimize trace generation for some gate that I know generates n IAND gates. If I use push_block(n)/IAND/pop_block, instead of getting approximately `0.5*n` CZ gates in result, I will get either 0 or n CZ gates, with probability 0.5 each. For large n this introduces large discrepancy, and direction of that discrepancy (over or under estimation) is random.
* My solution is instead of flipping a coin and probabilistically adding non-controlled gate, always add the gate but multiply all counts it produces by 0.5. I.e. replace the code above with: 
```
elif isinstance(op, ClassicallyControlledOperation):
    self.push_block(self.classical_control_probability)
    self.handle_op(op.without_classical_controls())
    self.pop_block()
```
* This also requires us to change a lot of types in Rust code from u64 to f64.

I am not sure it's the best way to do this, as using real numbers introduces different kind of non-determinism, related to rounding errors. Also clippy doesn't like mixing u64 and f64.


source/qdk_package/qdk/qre/interop/_cirq.py
```diff
@@ -163,7 +162,7 @@ def __init__(
             )
         )
 
-    def push_block(self, repetitions: int):
+    def push_block(self, repetitions: float):
```
msoeken, line 165: How about an extra argument that states the probability of this block being executed, assuming that the whole block is classically controlled. Then trace transforms who are based on gate counting such as `PSSPC` can use them for more precise counts.

source/qdk_package/qdk/qre/_qre.pyi
```diff
@@ -924,14 +924,14 @@ class EstimationResult:
     """
 
     def __new__(
-        cls, *, qubits: int = 0, runtime: int = 0, error: float = 0.0
+        cls, *, qubits: int = 0, runtime: float = 0.0, error: float = 0.0
```
msoeken, line 927: We cannot change all these types to `float`. All runtimes and gate counts must stay integers. We are working with some large numbers and will collect wrong numbers due to floating point rounding.

msoeken (changes requested):

PR #3347, opened 2026-06-17
Remove hard dependency on cirq in qdk.qre

msoeken:
This removes the hard dependency on `cirq` in QRE. It's possible now to import, e.g., `QSharpApplication` if `cirq-core` is note installed.

If the user wants to import `CirqApplication`, one needs to install `pip install qdk[qre,cirq]`. This message is also printed if the user tries to instantiate `CirqApplication` when `cirq-core` is not installed.

The `pyproject.toml` is updated accordingly.

All tests that require `cirq` are in one file and guarded by `pytest.importskip`, but `cirq` is available in tests.

### DIFF
--- a/source/qdk_package/qdk/qre/interop/__init__.py
+++ b/source/qdk_package/qdk/qre/interop/__init__.py
@@ -4,12 +4,6 @@
 from ._qir import trace_from_qir
 from ._qsharp import trace_from_entry_expr, trace_from_entry_expr_cached
 
-__all__ = [
-    "trace_from_entry_expr",
-    "trace_from_entry_expr_cached",
-    "trace_from_qir",
-]
-
 try:
     from ._cirq import (
         PeakUsageGreedyQubitManager,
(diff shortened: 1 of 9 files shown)


source/qdk_package/qdk/qre/interop/__init__.py
```diff
@@ -1,35 +1,42 @@
...
+        ReadFromMemoryGate,
+        TypedQubit,
+        WriteToMemoryGate,
+        assert_qubits_type,
+        read_from_memory,
+        trace_from_cirq,
+        write_to_memory,
+    )
+
+    __all__ += [
```
billti: This is the only aspect I'm a little concerned about. I don't think we should be making the public surface area 'dynamic'. As GPT states, _"Many IDEs, type checkers, and documentation tools treat __all__ as the authoritative list of public exports. So even if you never use import *, maintaining __all__ can help signal which symbols are intended to be public versus internal."_

Not only might users find it confusing to see the public surface area changing, I'm not sure our doc generation tools would handle it well either.

brad-lackey (approved):

source/qdk_package/qdk/qre/interop/__init__.py
msoeken: Makes sense. I will resolve it in the same way I did with `CirqApplication` and `CirqApplicationParams`.

msoeken: Addressed in b13e44366bbf7760f015a19070dac0d4e24c0445

PR #3349, opened 2026-06-17
Track dynamic constants across call boundaries in RCA

### LINKED ISSUE #3283: Arguments that are "Dynamic Constant" treated as "Static" across callable boundaries
(Note this repro depends on the bug in #3282, so if that is fixed first then a different repro may be needed)

When RCA determines the runtime capabilities of a callable invocation it uses the callable's application generator set and checks the dynamism of each argument to know wich capabiltiies to aggregate. However, only arguments that are "Dynamic Variable" are checked, the rest are assumed to be "Static" arguments. Internally, we effectively have three categories of values: Static, Dynamic Constant, or Dynamic Variable. By effectively promoting Dynamic Constant values into Static values, we lose some information across function calls. Using a specific repro, we can see that Q# expressions that generates an error if used directly is accepted if it is instead put in a function:
```qsharp
import Std.Arrays.IsEmpty;
@EntryPoint(Base)
operation Main() : Unit {
    use qs = Qubit[2];
    while IsEmpty(qs) {}
    LoopOnArray(qs);
}

function LoopOnArray(arr : Qubit[]) : Unit {
    while IsEmpty(arr) {}
}
```

Image

Commenting out the explicit `while` in the body of `Main` will avoid errors and successfully generate QIR (as explained in #3282, the capabilities error is incorrect). This shows that code in a callable is treated different than code across callables, which can cause problems if things like classical loops are supposed to be emitted based on Dynamic Constant values.

swernli:
This change updates how RCA handles dynamic constants across call boundaries. Previously, a dynamic constant would effectively be considered a static value when passed to a call. This worked fine for some cases, but as we enhance RCA to differentiate between static values and dynamic constant values (such as upcoming work on `Result` arrays), we need that difference to propagate properly at analysis time. This fixes #3283.

### DIFF
--- a/source/compiler/qsc_rca/src/tests.rs
+++ b/source/compiler/qsc_rca/src/tests.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#![allow(clippy::too_many_lines)]
-
 mod arrays;
 mod assigns;
 mod bindings;
--- a/source/compiler/qsc_rca/src/tests.rs
+++ b/source/compiler/qsc_rca/src/tests.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(clippy::too_many_lines)]
+
 mod arrays;
 mod assigns;
 mod bindings;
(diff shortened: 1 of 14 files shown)


source/compiler/qsc_rca/src/tests.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![allow(clippy::too_many_lines)]
```
idavis: This seems a little permissive

swernli: Moved to callables.rs, which is the one spot that had the particularly long snapshots.

idavis (approved):

PR #3350, opened 2026-06-17
Use internal npm registry in azdo pipeline

amcasey:
CFS requires that we pull packages from our internal package feed.  While perfectly reasonable, doing this unconditionally would create problems for our external contributors, who can't access the feed.  Instead, we leave things as-is on GitHub and update our AzDO pipeline to synthesize the necessary npmrc files.

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -453,8 +453,10 @@ extends:
                     displayName: Set version
 
                   # Windows arm64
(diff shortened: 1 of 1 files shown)


amcasey:
Note that this only addresses npm - we're still getting a warning about rust.

minestarks (approved):

PR #3352, opened 2026-06-17
Add code fix for double literal without dot

amcasey:
It's very easy to write `2` when you mean `2.`.  Offer a code fix.

### DIFF
--- a/source/language_service/src/code_action/int_to_double.rs
+++ b/source/language_service/src/code_action/int_to_double.rs
@@ -95,7 +95,7 @@ pub(crate) fn int_to_double_fixes(
     code_actions
 }
 
-/// Finds the AST expression whose span exactly matches `target` and returns its `NodeId`.
+/// Finds the AST expression whose span exactly matches `target`.
 fn find_expr_at(package: &ast::Package, target: Span) -> Option<&ast::Expr> {
     let mut finder = ExprSpanFinder {
         target,
--- a/source/language_service/src/code_action/int_to_double.rs
+++ b/source/language_service/src/code_action/int_to_double.rs
@@ -55,14 +55,8 @@ pub(crate) fn int_to_double_fixes(
             };
 
             // Strip off any + or - unary operators
-            loop {
-                if let ExprKind::UnOp(un_op, expr2) = expr.kind.as_ref()
-                    && (matches!(un_op, UnOp::Pos) || matches!(un_op, UnOp::Neg))
-                {
-                    expr = expr2;
-                } else {
-                    break;
-                }
+            while let ExprKind::UnOp(UnOp::Pos | UnOp::Neg, inner) = expr.kind.as_ref() {
+                expr = inner;
             }
 
             if !matches!(expr.kind.as_ref(), ExprKind::Lit(_)) {
--- a/source/language_service/src/code_action/int_to_double.rs
+++ b/source/language_service/src/code_action/int_to_double.rs
@@ -1,0 +1,117 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Code action: "Convert to integer literal to double"
+//! Detects when an integer literal is passed where a double is expected and
+//! offers to add a trailing `.` to make it into a double literal.
+
+#[cfg(test)]
+mod tests;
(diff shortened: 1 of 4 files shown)


amcasey:
I don't really care whether we merge this.  I make this mistake pretty regularly in Q#, but I mostly implemented it to practice writing rust without copilot.

With all the time in the world, I'd probably experiment with having type inference infer either `Int` or `Double` for integer literals, depending on the context.  In such a world, we might want a suffix to force the literal to Int, but the only place I can think of where both are allowed and the difference matters is division.

source/language_service/src/code_action/wrap_in_array/tests.rs
```diff
@@ -5,55 +5,62 @@ use crate::{
...
+    )
 }
 
 #[test]
 fn single_arg_qubit_to_qubit_array() {
     let source = "namespace A {
     operation Foo(qs: Qubit[]) : Unit is Adj {
         use q = Qubit();
-        Foo(q);
+        Foo(◉◉q◉◉);
```
amcasey, line 29: When we switched from one edit to two, the updated test became less meaningful as confirmation that the action was behaving correctly, so now we check the locations of the edits too.

source/language_service/src/code_action/int_to_double.rs
```diff
@@ -0,0 +1,121 @@
...
+                        source_name.to_string(),
+                        vec![TextEdit {
+                            new_text: ".".to_string(),
+                            range: dot_range,
+                        }],
+                    )],
+                }),
+                kind: Some(CodeActionKind::QuickFix),
+                is_preferred: Some(true),
+            });
```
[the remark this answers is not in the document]
amcasey: I think this can be applied with sufficiently high confidence that Preferred is justified.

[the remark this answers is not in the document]
amcasey: It does not, however, light up fix-all as I had hoped.

source/language_service/src/code_action/int_to_double.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    code_actions
+}
+
+/// Finds the AST expression whose span exactly matches `target` and returns its `NodeId`.
+fn find_expr_at(package: &ast::Package, target: Span) -> Option<&ast::Expr> {
```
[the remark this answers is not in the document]
amcasey: This is a good suggestion, but I think I'd prefer not to delete the signature line.

```suggestion
/// Finds the AST expression whose span exactly matches `target`.
fn find_expr_at(package: &ast::Package, target: Span) -> Option<&ast::Expr> {
```

source/language_service/src/code_action/int_to_double.rs
```diff
@@ -0,0 +1,121 @@
...
+        .filter_map(|error| match error.error() {
+            ErrorKind::Frontend(frontend_error) => frontend_error.ty_mismatch(),
+            _ => None,
+        });
+
+    for (expected, actual, error_span) in ty_mismatches {
+        // Check if expected is Double and actual is Int.
+        if matches!(&expected.kind, TyInfoKind::Prim(Prim::Double))
+            && matches!(&actual.kind, TyInfoKind::Prim(Prim::Int))
+        {
```
idavis, line 51: I think stripping of the unary ops and checking for lit will be faster than searching though the package for the expr span, this way you only do the longer search if it is right exprkind to consider.

amcasey: Personally, I find it rather clumsy that every code fix has to go search for the errors. In Roslyn, for example, part of the fix's interface is which error codes it cares about and then it's called only for that error and with the span for that error.

Leaving that aside for a moment, I'm not sure the span argument to this function is related to the error - I think it's the active span in the editor (usually an empty cursor span, I would guess). So I'm not sure we have the range we'd need to strip off the unary ops.

Or maybe I've misunderstood your suggestion?

amcasey: Wires crossed.  We're on the same page that this more or less has to be this way.

idavis (approved):

Issue #3353, opened 2026-06-18
Final Quantum Kata can't be completed


xhaidendsouza:
**Describe the bug**

When you press next while doing the very last quantum kata conclusion, it says 

Result
🎉 You have completed all content!

but it doesn't give a checkmark for the kata or say you have completed it. 
This means at the bottom it says 386/387 (100%) QEC: Bit Flip, Phase Flip, and Shor Codes
And at the top in the side bar it says 25/26 units complete — almost there!

**To Reproduce**

1. Open the microsoft quantum learn in vscode
2. Go to the final kata and press next
3. It will not mark it as completed

**Expected behavior**

It should mark it with a check, say 387/387 (100%) QEC: Bit Flip, Phase Flip, and Shor Codes, and say 26/26 units complete.

**Screenshots**

<img width="1971" height="1085" alt="Image" src="https://github.com/user-attachments/assets/08d9827f-24e0-4c29-a9c0-9cd411c3facc" />

**System information**

- Version of the package or extension for which the problem occurs: 

Latest Microsoft Quantum Development Kit (QDK) in VSCode. 
Identifier: quantum.qsharp-lang-vscode
Version: 1.29.1
Last Updated: 3 weeks ago

- For issues with extensions, the version of the IDE for which the problem occurs: 

Version: 1.120.0 (user setup)
Commit: 0958016b2af9f09bb4257e0df4a95e2f90590f9f
Date: 2026-05-12T20:17:22Z
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

**Additional context**

N/A

xhaidendsouza:
So will this change only be reflected in the next release, and if so, does this issue get closed now or only then?

minestarks:
@xhaidendsouza thank you so much for both the Issue and the PR! Yes, closing the issue is fine now, and your fix will be in this month's release.

PR #3354, opened 2026-06-18
Update service.ts to so that pressing next on the last kata works

xhaidendsouza:
Fix for Issue #3353

### DIFF
--- a/source/vscode/src/learning/service.ts
+++ b/source/vscode/src/learning/service.ts
@@ -207,7 +207,7 @@ export class LearningService {
     const ws = this.requireWorkspace();
     const currentPos = ws.progressData.position;
     const nextPos = this.nextActivity(currentPos);
-  
+
     // Auto-mark lesson activities complete when moving forward
     const oldKata = this.findUnit(currentPos.unitId);
     const oldActivity = oldKata.activities.find(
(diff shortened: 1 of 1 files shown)


source/vscode/src/learning/service.ts
```diff
@@ -220,6 +217,12 @@ export class LearningService {
       this.markComplete(currentPos);
     }
 
+    if (!nextPos) {
```
amcasey: This looks correct to me, but it took me a second to figure out what's different between this case and the normal one.  I wonder if another way to express this would be 
```ts
    const hasNext = !!nextPos;
    if (hasNext) 
    {
        ws.progressData.position = nextPos;
    }
    await this.saveProgress();
    this._onDidChangeState.fire(this.getState());
    this.sendActivityActionTelemetry("navigate", source);
    return { moved: hasNext };
```
I left in the telemetry because I couldn't think of a reason to treat it differently from a normal navigation.

p.s. The code is improvised - I haven't checked whether it compiles.

xhaidendsouza: Yeah that seems like a better way to do it. I changed the code to use that instead. The reason I originally hadn't put in the telemetry is originally is because it says "navigate" and I wasn't sure if that meant it was only intended to fire when the position actually changes, but if you think it doesn't need to be treated differently then that makes sense.

xhaidendsouza:

source/vscode/src/learning/service.ts
amcasey: Thanks for the update and for the change in general!  Everything looks good to me, but I'm going to let a more experienced team member take a look before we merge it.

minestarks (approved):

amcasey (approved):

PR #3360, opened 2026-06-19
Added (Majorana) fermions to basic operator types.

brad-lackey:
Added classes Fermion, FermionString, Majorana, MajoranaString analogous to Pauli and PauliString.
- Uses internal coding similar to Paulis: fermion creation operator is code 10, fermion annihilation operator is code 11...
- Update PauliString class with normalization: sort indices and eliminate combine Pauli operators on the same wire.
- Normalize PauliStrings before exporting to Cirq.
- Added normalization to MajoranaStrings.
- Created helper functions "vertex_operator" and "edge_operator" common in fermionic models.

### DIFF
--- a/source/qdk_package/qdk/applications/magnets/utilities/fermion.py
+++ b/source/qdk_package/qdk/applications/magnets/utilities/fermion.py
@@ -1,0 +1,246 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Fermionic operator representations for many-body systems."""
+
+from collections.abc import Sequence
+
+
+class Fermion:
+    """Single-mode fermionic operator tied to an explicit site index.
+
+    ``Fermion`` stores a fermionic operator identifier and the site it acts on.
+    The identifier can be provided either as an integer code or a label:
+
+    - ``10`` / ``"A^"`` / ``"CREATE"`` / ``"CREATION"``
+    - ``11`` / ``"A"`` / ``"ANNIHILATE"`` / ``"ANNIHILATION"``
+
+    The annihilation operator ``A`` is treated as primitive, and the creation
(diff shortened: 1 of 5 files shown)


source/qdk_package/qdk/applications/magnets/utilities/fermion.py
```diff
@@ -0,0 +1,246 @@
...
+    """Single-mode fermionic operator tied to an explicit site index.
+
+    ``Fermion`` stores a fermionic operator identifier and the site it acts on.
+    The identifier can be provided either as an integer code or a label:
+
+    - ``10`` / ``"A^"`` / ``"CREATE"`` / ``"CREATION"``
+    - ``11`` / ``"A"`` / ``"ANNIHILATE"`` / ``"ANNIHILATION"``
+
+    The annihilation operator ``A`` is treated as primitive, and the creation
+    operator ``A^`` denotes its Hermitian conjugate.
```
msoeken, line 19: Would it make sense to track whether it's the conjugate or not as a variable in the class?

msoeken (approved):

source/qdk_package/qdk/applications/magnets/utilities/fermion.py
brad-lackey: Practically we could do that, but perhaps I was overzealous with the documentation. By saying the annihilation operator is the primitive, I really mean that choose it over the creation operator in defining our notation. The better notation would be use "A" for the annihilation operator and "C" for the creation operator, and it just happens that they are Hermitian conjugates of each other. But then this impacts Majoranas (in the literature the two induced fermionic operators are usually distinguish via indexing, which is very messy here). So I'll have a review of the documentation once I get the rest of the library finalized and smooth this out.

PR #3364, opened 2026-06-21
Cross-package FIR transforms

idavis:
# Cross-package FIR transforms

## Summary

This change makes the FIR transform pipeline operate over the **entire entry-reachable package closure** (entry package + all reachable library/foreign packages) instead of just the entry package, and removes the now-unused `Namespace` and `Export` item kinds from FIR. Together these let codegen-time transforms (monomorphize, defunctionalize, UDT erasure, tuple decomposition, argument promotion, etc.) correctly specialize and rewrite callables that live in dependency packages.

## Motivation

The QDK codegen pipeline historically emitted QIR only for the entry package, inlining everything reachable into a single rewritten unit and treating dependency packages as read-only sources. Generating code for **library** callables directly — so that a compiled library can be linked rather than re-inlined per consumer — requires the FIR transform passes to specialize and rewrite callables *in place within the packages that own them*, not just in the entry package. This PR lays that groundwork: it gives every package its own ID space (`PackageAssigners`) so passes can mint fresh FIR nodes into a foreign package's arena without colliding with that package's existing IDs, and it extends the cleanup stages (item DCE, node-level GC, and exec-graph rebuild) to run across the entire entry-reachable package closure so each transformed library package is left internally consistent. Removing the now-inert `Namespace` and `Export` item kinds from FIR is part of the same simplification — codegen operates on a flat, fully-resolved callable graph, so the namespace/export resolution metadata that only the front end needs no longer has to be carried (or kept consistent) through the cross-package rewrite. Together these changes turn the transform pipeline from an entry-package-only rewriter into one that can correctly process callables wherever they live, which is the prerequisite for emitting code for libraries.

The pipeline also only cleaned up (DCE/GC/exec-graph rebuild) the entry package, leaving foreign packages internally inconsistent (stale callables referencing erased UDTs and pre-promotion signatures), which RCA/codegen would then analyze out of step and panic on.

### 1. Per-package ID spaces (`PackageAssigners`)

- New `package_assigners` module introduces `PackageAssigners`, a lazily-populated map from `PackageId` to that package's `Assigner`.
- Each package's assigner is seeded on first touch from that package's own ID watermark via `Assigner::from_package`; advanced watermarks persist across passes via `get_mut` / `with_package`.
- The single shared `Assigner` is replaced throughout the pipeline driver; every structural pass entry point now takes `&mut PackageAssigners` and selects the owning package's assigner at each cross-package boundary. Leaf helpers keep their `&mut Assigner` signatures.
- The pipeline skip set widened from `FxHashSet<LocalItemId>` to `FxHashSet<StoreItemId>` so it is package-qualified.

### 2. Whole-closure cleanup

- `exec_graph_rebuild` now unconditionally rebuilds exec graphs for **every reachable spec in every reachable package** (`rebuild_exec_graphs`), replacing the old `rebuild_exec_graphs_with_external_specs` + `structurally_mutated_specs` tracking (the `CallableSpecId` struct and that bookkeeping are removed).
- New `item_dce::eliminate_unreachable_foreign_items` removes entry-unreachable items from foreign packages (a library's public surface is not an entry point for a closed codegen compilation). Pinned callable items and their transitive deps are retained.
- Node-level GC now runs across the whole reachable package closure via `reachability::collect_reachable_package_closure`, not just the entry package.
- `tuple_decompose ⇄ arg_promote` fixed point and the idempotent `normalize_reachable_call_arg_types` cleanup are documented and now operate cross-package.

### 3. Documented destructive-store contract

The pipeline now mutates the `PackageStore` in place across all reachable packages; the mutated store is a disposable codegen artifact and must not be reused. This is documented on `run_pipeline_to_with_diagnostics`, `run_pipeline_with_diagnostics`, and `run_signature_preserving_subpipeline`.

### 4. Remove `Namespace`/`Export` from FIR

- Deleted `qsc_fir/src/global.rs` (the `Global`/`Table`/`PackageIter` namespace-resolution machinery, 164 lines) and dropped the `global` module from `qsc_fir/src/lib.rs`.
- Removed `ItemKind::Namespace` and `ItemKind::Export` variants and their `Display` impls in `fir.rs`; `get_global` no longer handles them.
- `qsc_lowerer` now filters out HIR `Namespace`/`Export` items before lowering, so they never reach FIR (item IDs stay identity-mapped; the gaps are tolerated by the item map).
- Updated visitors (`visit.rs`, `mut_visit.rs`) and RCA (`core.rs`, `cycle_detection.rs` — drops namespace/export call handling and the `resolve_item` helper).
- Clarified the `ItemId.package` doc ("ID of the package that owns this item").

#### Performance-oriented techniques

This PR scales the transform pipeline from one package to the whole entry-reachable closure, so several design choices exist specifically to keep that wider scope cheap:

- **Lazy per-package assigner seeding.** `PackageAssigners` builds an `Assigner` for a package only the first time that package is *mutated*. Seeding scans the package's ID watermark (`Assigner::from_package`), so packages that are only read from — the common case for most library dependencies — are never scanned at all, instead of eagerly seeding an assigner for every package in the store.
- **Persistent watermarks across passes.** The advanced assigner is written back into the map after each use (`with_package` does a remove → run → insert, `get_mut` hands back a live reference), so a later pass reuses the running watermark instead of re-deriving a package's max IDs from scratch on every entry.
- **Borrow-friendly access (no store/assigner clones).** `get_mut` returns a reference borrowing only `self`, and `with_package` takes the assigner *by value* out of the map while running the closure. This lets a pass mutate the `PackageStore` and hold its assigner simultaneously without cloning the store or the assigner to dodge the borrow checker.
- **Idempotent cleanup hoisted out of the hot loop.** `normalize_reachable_call_arg_types` runs exactly once *after* the `tuple_decompose ⇄ arg_promote` fixed point converges, not per round. Running it per round caused `(T,)` single-element-tuple wrap/unwrap churn that polluted change detection and provoked extra fixed-point iterations; hoisting it removes that churn.
- **Snapshot callee signatures once.** That same normalization snapshots every reachable callable's input type up front into a `StoreItemId`-keyed map and then rewrites each package against the snapshot, rather than re-resolving each callee's (possibly cross-package) signature at every call site.
- **One invariant walk per fixed-point round.** Instead of an invariant check after both `tuple_decompose` and `arg_promote`, a single `PostArgPromote` check runs after both passes per round (tuple-decompose provably preserves the invariant), halving the invariant traversals inside the convergence loop.
- **Reachability rooted once per pass.** The reachable closure is computed once at a pass entry and reused across that pass, rather than recomputed per item.
- **Scoped, diagnostic-only snapshots in defunctionalize.** Foreign-package bodies are walked only to discover HOF call sites and closure callees and are deliberately *not* snapshotted into the diagnostic map, avoiding both cross-package key collisions and the cost of capturing bodies that won't be rewritten; rewrite is driven once per owning package.
- **Specialize the shared closure target once.** The shared closure-target lambda is specialized a single time and reused across call sites instead of being re-specialized per use.
- **Batch the cleanup tail.** Item DCE, node-level GC, and exec-graph rebuild run once over the converged reachable package closure rather than incrementally, and the old `structurally_mutated_specs` bookkeeping is dropped in favor of an unconditional whole-closure rebuild.

## Tests

- New full-pipeline snapshot tests over the `Dynamics.qs` and `Grover.qs` samples (`sample_pipeline_tests.rs`) that render every item reachable across all packages.
- New `package_assigners` unit tests, new codegen base-profile QIR tests (`base_profile.rs`), and added interpret tests.
- Substantial cross-package test additions/updates across `arg_promote`, `defunctionalize`, `item_dce`, `invariants`, `monomorphize`, `udt_erase`, and `exec_graph_rebuild`.
- Eval snapshot tests updated to drop the now-absent `Namespace` items from rendered FIR.

## Risk / notes

- Internal compiler refactor; no public Q# language surface change.
- Per-package ID isolation is the core correctness property — verified by tests asserting projection temporaries are minted into the caller's package, not the callee's.

## FIR Transform Benchmark — main vs current branch

The pipeline invariants were a known bottleneck in the previous PR and optimization was delayed until this cross-package PR could set a better foundation for optimizations. Two samples (`samples/estimation/Dynamics.qs`, `samples/algorithms/Grover.qs`, Base profile) were used as benchmarks as they use a fair amount of library code.

## Performance

| Sample | Metric | main | current | Δ |
|---|---|--:|--:|--:|
| **Dynamics** | Wall-clock pipeline | 62.63 ms | **23.74 ms** | **−62.1%** |
| | Passes (transform work) | 6.72 ms (10.8%) | 14.23 ms (61.1%) | +111.6% |
| | Invariants | 55.63 ms (89.2%) | 9.07 ms (38.9%) | **−83.7%** |
| **Grover** | Wall-clock pipeline | 62.02 ms | **16.60 ms** | **−73.2%** |
| | Passes (transform work) | 6.55 ms (10.6%) | 9.91 ms (61.1%) | +51.3% |
| | Invariants | 55.18 ms (89.4%) | 6.31 ms (38.9%) | **−88.6%** |

### Per pass — main vs current (ms)

| Pass | main (Dynamics) | current (Dynamics) | main (Grover) | current (Grover) |
|---|--:|--:|--:|--:|
| udt_erase | 5.654 | 3.620 | 5.671 | 3.548 |
| return_unify | 0.129 | 3.595 | 0.155 | 0.180 |
| item_dce_gc | 0.089 | 3.149 | 0.080 | 3.115 |
| monomorphize | 0.194 | 2.100 | 0.135 | 1.983 |
| defunctionalize | 0.241 | 0.484 | 0.201 | 0.279 |
| arg_promote | 0.077 | 0.281 | 0.059 | 0.181 |
| exec_graph_rebuild | 0.060 | 0.247 | 0.047 | 0.164 |
| tuple_decompose | 0.079 | 0.200 | 0.063 | 0.082 |
| normalize_call_args | 0.040 | 0.163 | 0.029 | 0.113 |
| tuple_compare_lower | 0.037 | 0.113 | 0.028 | 0.077 |
| fixpoint:tuple_decompose | 0.082 | 0.111 | 0.058 | 0.078 |
| fixpoint:arg_promote | 0.035 | 0.088 | 0.024 | 0.061 |
| cond_normalize | 0.007 | 0.073 | 0.005 | 0.053 |
| **Invariants (total)** | **55.63** | **9.07** | **55.18** | **6.31** |

*(main's invariant time is a single total bucket; current's is the sum of per-checkpoint timings.)*

### What this shows

- The cross-package branch is **~2.6× (Dynamics) / ~3.7× (Grover) faster overall**, even though it does strictly more transformation work.
- The entire win comes from **invariant scoping**. On `main`, invariants are ~89% of FIR-transform time (~55 ms) — the full std/core package-closure walk running at every PostUdtErase-or-later checkpoint (~9×). The branch's reachable-*item* scoping cuts that to ~6–9 ms (−84% to −89%).
- **Passes got more expensive** on the branch (`return_unify`, `item_dce_gc`, `monomorphize` all grew) because they now transform user + std + core, not just the entry package. That added cost is small next to the invariant savings.


source/compiler/qsc_lowerer/src/lib.rs
```diff
@@ -197,6 +197,15 @@ impl Lowerer {
...
+            // reads, so skip them. Item ids stay identity-mapped; the gaps left
+            // at the skipped ids are tolerated by the item map.
+            .filter(|i| {
+                !matches!(
+                    i.kind,
+                    hir::ItemKind::Namespace(..) | hir::ItemKind::Export(..)
+                )
+            })
             .map(|i| self.lower_item(i))
             .map(|i| (i.id, i))
```
swernli: I think this whole thing can be replaced with a `filter_map` like this:

```suggestion
            .filter_map(|i| {
                if matches!(
                    i.kind,
                    hir::ItemKind::Namespace(..) | hir::ItemKind::Export(..)
                ) {
                    None
                } else {
                    let item = self.lower_item(i);
                    Some((item.id, item))
                }
            })
```

source/compiler/qsc_lowerer/src/lib.rs
```diff
@@ -239,6 +248,15 @@ impl Lowerer {
         let items: IndexMap<LocalItemId, fir::Item> = hir_package
             .items
             .values()
+            // Namespace and export items are inert metadata that no FIR consumer
+            // reads, so skip them. Item ids stay identity-mapped; the gaps left
+            // at the skipped ids are tolerated by the item map.
+            .filter(|i| {
```
swernli: likewise here a `filter_map` can be used.

source/compiler/qsc_fir_transforms/src/package_assigners.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl PackageAssigners {
+    /// Creates a `PackageAssigners` seeded with the entry package's assigner.
+    ///
+    /// This reproduces the previous single-assigner behavior: the entry
+    /// package's assigner is constructed once up front from its watermark.
+    /// The eager seed is cosmetic parity only — `get_mut`/`with_package` would
+    /// lazily seed the entry package from the same watermark on first use, so
+    /// this is equivalent to starting from an empty map.
+    pub(crate) fn entry(store: &PackageStore, package_id: PackageId) -> Self {
```
swernli: nitpick: since this creates a new `PackageAssigners` and acts like a constructor, I'd prefer `from_entry` or even just `new`.

source/compiler/qsc_fir_transforms/src/walk_utils.rs
```diff
@@ -228,6 +238,154 @@ pub(crate) fn for_each_direct_child<F: FnMut(DirectChild)>(kind: &ExprKind, mut
...
+{
+    match callable_impl {
+        CallableImpl::Intrinsic => {}
+        CallableImpl::Spec(spec_impl) => {
+            for_each_node_in_spec_impl(pkg, spec_impl, visit);
+        }
+        CallableImpl::SimulatableIntrinsic(spec_decl) => {
+            for_each_node_in_spec_decl(pkg, spec_decl, visit);
+        }
+    }
```
swernli: does it need to walk to the internals of simulatable intrinsics? For QIR emission, these are the same as regular intrinsic and the body is ignored.

source/compiler/qsc_fir_transforms/src/test_utils.rs
```diff
@@ -338,6 +339,96 @@ pub fn compile_to_fir_with_library_and_capabilities(
...
 
+/// Compiles a two-library dependency chain plus a user package through
+/// core+std → HIR passes → FIR lowering. `lib_a_source` depends on
+/// `lib_b_source`, and `user_source` depends on `lib_a_source`, forming an
+/// entry → libA → libB chain across distinct packages.
+///
+/// Returns a FIR store with six packages (core, std, libB, libA, user) and the
+/// user package ID.
+#[cfg(test)]
+#[allow(dead_code, clippy::similar_names)]
```
swernli: I don't think the `dead_code` allow is needed

```suggestion
#[allow(clippy::similar_names)]
```

source/compiler/qsc_fir_transforms/src/test_utils.rs
```diff
@@ -373,6 +464,21 @@ pub fn compile_to_fir_with_entry(source: &str, entry: &str) -> (fir::PackageStor
     compile_to_fir_with_cached_stdlib(source, Some(entry), TargetCapabilityFlags::empty())
 }
 
+/// Compiles Q# source with an explicit executable entry expression through
+/// core+std → HIR passes → FIR lowering → monomorphization.
+///
+/// Returns a monomorphized FIR store ready for later pipeline stages.
+#[allow(dead_code)]
```
swernli: I think like above this can just be marked as test without needing the dead code distinction.

idavis: Still needed here.

swernli: true for other recently added code in here. I find `#[cfg(test)]` less disconcerting than `#[allow(dead_code)]` when scrolling by.

source/compiler/qsc_openqasm_compiler/src/tests.rs
```diff
@@ -476,9 +476,10 @@ fn get_last_statement_as_qsharp(package: &Package) -> String {
     }
 }
 
+#[allow(dead_code)]
 pub struct AstDespanner;
 impl AstDespanner {
-    #[allow(dead_code)] // false positive lint
+    #[allow(dead_code)]
```
swernli: I think the changes in this file might be by mistake... while the `allow(dead_code)` here might not be needed I think it's unrelated to the rest of the changes in the PR and the whole file can be reverted.

source/compiler/qsc_fir_transforms/src/invariants.rs
```diff
@@ -840,79 +910,128 @@ fn check_reachable_invariants(
...
+/// lockstep across the implementation phases; this predicate is the single
+/// widening point for that lockstep decision. At the end state every structural
+/// pass runs across the whole reachable package closure, so every stage's
+/// invariant holds for every reachable package and this predicate admits every
+/// reachable package for every stage. The exhaustive match is retained so that
+/// adding a new stage forces an explicit scope decision.
+fn structural_check_in_scope(
+    check: StageCheck,
+    _item_package: PackageId,
+    _target_package_id: PackageId,
```
swernli: I understand having this function to keep an exhaustive match that forces any new stages added to be considered for how they affect structural checks, but why include these package ids and leave them unused? Since the function only returns `bool` there isn't any error to put them on, so why not just remove them?

source/compiler/qsc_fir_transforms/src/invariants.rs
```diff
@@ -840,79 +910,128 @@ fn check_reachable_invariants(
...
+///
+/// `enforces` answers "should stage `check` run on this callable?": yes exactly
+/// when the level is at or past that stage and the callable's package is in the
+/// stage's cross-package scope. For a callable in the entry (target) package
+/// every stage is in scope, so `enforces` reduces to `level.enforces(check)` and
+/// the entry-package walk is unchanged.
+#[derive(Clone, Copy)]
+struct CheckScope {
+    item_package: PackageId,
+    target_package_id: PackageId,
```
swernli: These properties of the check scope seem to have the same quirk: they are set and are passed into `structural_check_in_scope` as part of `CheckScope::enforces` but that function ignores them. As a result, these fields aren't actually used anywhere, making it seem like there is no real distinction between `CheckScope::enforces` and `InvariantLevel::enforces`.

idavis: good call, I didn't get to this refactoring after moving to cross-package

idavis: pushed

source/compiler/qsc_fir_transforms/src/item_dce.rs
```diff
@@ -56,13 +59,120 @@ pub fn eliminate_dead_items(
...
+/// callable is dead here. Pinned **callable** items (and their
+/// transitive dependencies, already present in `reachable`) are kept in
+/// whichever package they live in: `local_reachable` only retains the
+/// `Callable` arm, so pinning a non-callable item (for example a `Ty`) is a
+/// contract violation that [`eliminate_foreign_items_in_package`] rejects. The
+/// FIR store the pipeline mutates is a throwaway codegen clone, so trimming a
+/// library package's unreachable surface cannot affect any other compilation.
+///
+/// Returns the total number of items removed across all foreign packages.
+#[allow(clippy::implicit_hasher)]
```
swernli: looks like this extra allow snuck in, removing it doesn't seem to have an effect.

```suggestion
```

source/compiler/qsc_fir_transforms/src/defunctionalize/analysis.rs
```diff
@@ -44,12 +44,26 @@ use rustc_hash::{FxHashMap, FxHashSet};
...
+    /// Types of the enclosing callable's capturable variable bindings
+    /// (parameters and immutable `let` bindings), keyed by `LocalVarId`.
+    /// `LocalVarId`s are scoped per callable and collide freely across
+    /// callables in the same package, so a captured variable's type must be
+    /// resolved against this per-callable map rather than a package-wide
+    /// pattern scan. This map serves only closure-capture type resolution.
+    /// Mutable locals may still exist and are tracked for flow in `callable`;
+    /// they are simply never recorded here because a closure can never capture
+    /// one.
+    var_types: FxHashMap<LocalVarId, Ty>,
```
swernli: nitpick: I think it would make it easier to remember what this hashmap is for if the name called out the fact that it is for closed over local variables. Maybe something like `closure_var_types`?

swernli (approved):

PR #3368, opened 2026-06-22
Move qasm `display_utils` to `qsc_data_structures::display`

orpuente-MS:
Move the `display_utils` used in the OpenQASM compiler to a common location, so that they can be reused for Stim unit tests.

This PR also deletes an orphan file in the OpenQASM compiler.


source/compiler/qsc_data_structures/src/display/core.rs
```diff
@@ -0,0 +1,60 @@
...
+{
+    let indent = indenter::indented(f);
+    set_indentation(indent, 1)
+}
+
+/// Takes an `indenter::Indented` and changes its indentation level.
+///
+/// Note: This function is a very low level primitive. It's only
+///       public to mantain backwards compatibility with existing code.
+///       Prefer using
```
joao-boechat: nit: the comment here isn't finished

source/compiler/qsc_data_structures/src/display.rs
```diff
@@ -18,3 +22,154 @@ pub fn join(
...
+
+/// Writes a field of a structure to the given buffer
+/// or stream with an additional indententation level.
+pub fn write_field<T: Display>(f: &mut impl Write, field_name: &str, val: &T) -> fmt::Result {
+    let mut indent = with_indentation(f);
+    write!(indent, "{field_name}: {val}")
+}
+
+/// Writes a field of a structure to the given buffer
+/// or stream with an additional indententation level.
```
joao-boechat: nit:
indentation is spelled wrong in this file

joao-boechat (commented):
Has qsc_openqas_compiler/src/parser/ast.rs been dead code for a while? Was there a reason we kept it there or did it just "fall through the cracks"?

Not related to this PR, just wondering whether we might have more files like this in the repo

joao-boechat (approved):

source/compiler/qsc_data_structures/src/display/core.rs
orpuente-MS: Nice catch. Fixed.

source/compiler/qsc_data_structures/src/display.rs
orpuente-MS: Ha! I never noticed that. Nice catch, fixed!

orpuente-MS:
> Has qsc_openqas_compiler/src/parser/ast.rs been dead code for a while? Was there a reason we kept it there or did it just "fall through the cracks"?
> 
> Not related to this PR, just wondering whether we might have more files like this in the repo

This one just fell through the cracks. It was never supposed to stay here as dead code. I realized it was dead code because I wasn't getting rust-analyzer support for it.

PR #3372, opened 2026-06-23
Always lower into single HIR stmt from AST

### LINKED ISSUE #3369: Fuzz: "thread '<unnamed>' (6255) panicked at source/compiler/qsc_hir/src/validate.rs:21:13:" (ubuntu-latest)
The fuzz-testing workflow has detected a bug.

<details><summary><b>Auto-Minimized Fuzzing Input That Triggers the Bug:</b> Click this line.</summary>
<b>Note:</b> If the input is multi-line then the end-of-line characters '
' (0x0A) and '
' (0x0D)
may affect the reproducibility of the bug. If you fail to repro the bug with the input shown below
then you may want to go to the <a href=https://github.com/microsoft/qdk/actions/runs/27982837856>workflow</a> that reported this GitHub bug,
download the artifact, and extract the file with the exact minimized input.

```qs
operation S(){export y,e
```

</details>

<details><summary><b>The branch/commit the bug has been found in:</b> Click.</summary>
If the developers fail to repro the bug in the latest <code>main</code> then the branch/commit info below can help them to make sure
that they are using the correct way to repro. If the bug is reproducible in the branch/commit below, but not in latest <code>main</code>,
then the bug is likely fixed already or is not in the <code>main</code> branch.

```log
Branch: * main

commit 43d2678e171fd26368f8f3bc943753912b80a6e4
Author: Stefan J. Wernli <swernli@microsoft.com>
Date:   Mon Jun 22 13:30:26 2026 -0700

    Use Result comparison as source of dynamism in RCA (#3355)
    
    This change adjusts how `Result` values are treated in Runtime
    Capability Analysis, shifting them from being treated as dynamic
    variable values by default to treating them as dynamic constant values.
    This matches the behavior for `Qubit` variables, as both are identifiers
    in QIR. Instead of counting on the `Result` value returned from
    measurement intrinsics as the source of dynamism, RCA instead tracks
    when a `Result` value is used in a comparison and marks the resulting
    `Bool` value as a dynamic variable. This more directly maps to how
    dynamism enters a program and sets the stage for emitting constant
    `Result` arrays into the QIR in the same fashion as `Qubit` arrays. This
    builds off of the work in #3349 as we now track the difference between
    dynamic constants and dyanmic variables across function call boundaries,
    allowing RCA to properly detect comparison of dynamic constant results
    throughout the program.
    
    Additionally, this changes how cyclic callables are analyzed in RCA,
    allowing for more advanced handling and emission as IR functions. The
    cyclic callable pre-analysis is removed in favor of assumptions for
    return values from cyclic invocations during normal analysis. Combining
    this with partial eval handling of some intrinsic functions also fixes
    #3357.
```

</details>

**Other Info**

- [**Workflow**](https://github.com/microsoft/qdk/actions/runs/27982837856) (contains the run artifacts).
- **Workflow Agent System Info:** `ubuntu-latest: Linux runnervm7b5n9 6.17.0-1018-azure #18~24.04.1-Ubuntu SMP Thu May 28 16:39:11 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux`.

swernli:
The duplicate stmt id panic was because multi-exports produce two items during lowering (ie: `export a, b;` effectively becomes `export a` and `export b`) that then get turned into two HIR stmts with the same node id. Since `export` items as stmts are always invalid there is no reason to reproduce all of them, so this updates the lowerer to only produce the first item-as-stmt so that the previous name-resolution-time error can be reported and the panic avoided. Fixes #3369

### DIFF
--- a/source/compiler/qsc_frontend/src/lower.rs
+++ b/source/compiler/qsc_frontend/src/lower.rs
@@ -654,26 +654,31 @@ impl With<'_> {
             stmts: block
                 .stmts
                 .iter()
-                .flat_map(|s| self.lower_stmt(s))
+                .filter_map(|s| self.lower_stmt(s))
                 .collect(),
         }
     }
 
-    pub(super) fn lower_stmt(&mut self, stmt: &ast::Stmt) -> Vec<hir::Stmt> {
+    pub(super) fn lower_stmt(&mut self, stmt: &ast::Stmt) -> Option<hir::Stmt> {
         let id = self.lower_id(stmt.id);
-        let mut stmts = Vec::new();
-        match &*stmt.kind {
-            ast::StmtKind::Empty | ast::StmtKind::Err => {}
-            ast::StmtKind::Expr(expr) => stmts.push(hir::StmtKind::Expr(self.lower_expr(expr))),
+        let stmt_kind = match &*stmt.kind {
+            ast::StmtKind::Empty | ast::StmtKind::Err => None,
+            ast::StmtKind::Expr(expr) => Some(hir::StmtKind::Expr(self.lower_expr(expr))),
             ast::StmtKind::Item(item) => {
-                stmts.extend(self.lower_item(item).into_iter().map(hir::StmtKind::Item));
+                // The only kind of item that can result in multiple lowered entries is an export, and those are invalid
+                // from statements. Only lower the first one to prevent a duplicate ID panic and allow the normal name
+                // resolution error to occur.
+                self.lower_item(item)
+                    .drain(..)
+                    .next()
+                    .map(hir::StmtKind::Item)
             }
-            ast::StmtKind::Local(mutability, lhs, rhs) => stmts.push(hir::StmtKind::Local(
+            ast::StmtKind::Local(mutability, lhs, rhs) => Some(hir::StmtKind::Local(
                 lower_mutability(*mutability),
                 self.lower_pat(lhs),
                 self.lower_expr(rhs),
             )),
-            ast::StmtKind::Qubit(source, lhs, rhs, block) => stmts.push(hir::StmtKind::Qubit(
+            ast::StmtKind::Qubit(source, lhs, rhs, block) => Some(hir::StmtKind::Qubit(
                 match source {
                     ast::QubitSource::Fresh => hir::QubitSource::Fresh,
                     ast::QubitSource::Dirty => hir::QubitSource::Dirty,
@@ -682,17 +687,14 @@ impl With<'_> {
                 self.lower_qubit_init(rhs),
                 block.as_ref().map(|b| self.lower_block(b)),
             )),
-            ast::StmtKind::Semi(expr) => stmts.push(hir::StmtKind::Semi(self.lower_expr(expr))),
-        }
+            ast::StmtKind::Semi(expr) => Some(hir::StmtKind::Semi(self.lower_expr(expr))),
+        };
 
-        stmts
-            .into_iter()
-            .map(|kind| hir::Stmt {
-                id,
-                span: stmt.span,
-                kind,
-            })
-            .collect()
+        stmt_kind.map(|kind| hir::Stmt {
+            id,
+            span: stmt.span,
+            kind,
+        })
     }
 
     #[allow(clippy::too_many_lines)]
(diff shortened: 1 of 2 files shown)


idavis (approved):

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -654,26 +654,31 @@ impl With<'_> {
...
                 .stmts
                 .iter()
-                .flat_map(|s| self.lower_stmt(s))
+                .filter_map(|s| self.lower_stmt(s))
                 .collect(),
         }
     }
 
-    pub(super) fn lower_stmt(&mut self, stmt: &ast::Stmt) -> Vec<hir::Stmt> {
+    pub(super) fn lower_stmt(&mut self, stmt: &ast::Stmt) -> Option<hir::Stmt> {
```
amcasey, line 662: Is this an example of the AST-level desugaring we discussed yesterday?

amcasey: Oh, it seems like we might call building the HIR tree from the AST tree "lowering"?

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -654,26 +654,31 @@ impl With<'_> {
...
-        let mut stmts = Vec::new();
-        match &*stmt.kind {
-            ast::StmtKind::Empty | ast::StmtKind::Err => {}
-            ast::StmtKind::Expr(expr) => stmts.push(hir::StmtKind::Expr(self.lower_expr(expr))),
+        let stmt_kind = match &*stmt.kind {
+            ast::StmtKind::Empty | ast::StmtKind::Err => None,
+            ast::StmtKind::Expr(expr) => Some(hir::StmtKind::Expr(self.lower_expr(expr))),
             ast::StmtKind::Item(item) => {
-                stmts.extend(self.lower_item(item).into_iter().map(hir::StmtKind::Item));
+                // The only kind of item that can result in multiple lowered entries is an export, and those are invalid
```
amcasey, line 668: @minestarks mentioned a case I didn't fully follow where nested functions get promoted to top-level functions at some point, possibly in HIR.  I take it we wouldn't expect that to happen here and produce multiple items from a single item statement (assuming a function declaration is an item statement)?

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -654,26 +654,31 @@ impl With<'_> {
...
+                // from statements. Only lower the first one to prevent a duplicate ID panic and allow the normal name
```
amcasey, line 669: Is "the normal name resolution error" not finding the export we just dropped?  Or trying to figure out what the second symbol was exporting and failing for some reason?

source/compiler/qsc_frontend/src/lower.rs
swernli: Yes, this is the part that produces the HIR tree from the AST + name resolution table + inferred expr types.

source/compiler/qsc_frontend/src/lower.rs
swernli: Name resolution verifies `export` items as part of building up the internal tree of visible names. That's what generates the error disallowing any exports from local scopes (specifically the `Qsc.Resolve.ExportFromLocalScope` error). It also will produce the errors you mean, name resolution errors on the export identifiers themselves if they are not found, but the entire statement is invalid and will get flagged, so no reason to preserve that whole invalid statement down into the HIR.

source/compiler/qsc_frontend/src/lower.rs
swernli: Callable declaration is an item stmt when it happens inside another callable, but it never produces multiple items in the same statement. Only comma separated imports or exports produce more than one item, imports are already dropped inside of `lower_item` so nothing valid can ever generate more than one item per statement.

source/compiler/qsc_frontend/src/lower.rs
amcasey: To confirm, the error is not that there are two symbols exported in one statement but that one of the exports is invalid?  If that's the case, why isn't the flatmap needed when there are no errors?

swernli: It's than _any_ export from inside of a function/operation is invalid. See https://github.com/microsoft/qdk/blob/main/source/compiler/qsc_frontend/src/resolve/tests.rs#L5909-L5931

PR #3382, opened 2026-06-24
Type checking for Python code

fedimser:
This PR:
 * Adds an option to run python static checks using `build.py`. Currently it only runs type checks using pyright, but more static checks (line length, linter, formatter) can be added here.
 * Adds a new CI workflow that installs minimal requirements and runs the static checks. 
 * Fixes code to make static checks pass. 

Notes:
* I only check Python code in `source/qdk_package/qdk`. 
* Type checking requires different dependencies than tests, so I created a new file `check_requirements.txt` that includes minimal dependencies that are required to run static checks. This minimizes setup time for CI workflow.

The type checking can be done in multiple ways:
* `pyright` - runs pyright directly.
* `./build.py --check-py` - recommended. This installs required dependencies and runs mypy, and doesn't do anything else. If this passes locally, it's guaranteed that it will pass on CI.
* `python ./build.py --check-py --no-check-prereqs` - like above, but skips checking prerequisites (like wasm) which are irrelevant for Python. This is used in the CI workflow.
* Run `build.py` in a way that includes building qdk and doesn't disable checks. This includes `build.py --qdk` and `build.py` with no arguments.

Missing checks (will be added in follow-up PRs):
* If we remove 6 excluded modules and specify `"reportMissingParameterType": "error"` and `"typeCheckingMode": "strict"` in pyrightconfig.json, we get 2745 errors. I plan to gradually fix them in a series of PRs.


source/qdk_package/check_requirements.txt
```diff
@@ -0,0 +1,6 @@
+# Libraries that need to be installed to run Python static checks.
+matplotlib
+pandas>=2.1
+pandas-stubs
+pyqir>=0.12.5,<0.13
```
billti, line 5: Why do these need to be installed for the checks? Does it need to verify the types from any package your our code may use? (In which case, would it need to ultimately extend to qiskit, cirq, etc.)?

build.py
```diff
@@ -63,6 +63,13 @@
     help="Run the prerequisites check (default is --check-prereqs)",
 )
 
+parser.add_argument(
+    "--check-py",
```
billti, line 67: The existing `check` flag does JavaScript/TypeScript and Rust. Is there a reason to add a distinct Python check flag than just roll it into the existing `check` (if building a Python package).

fedimser: Yes, 
1. So I can run Python checks only without having to wait for Rust and TypeScript checks that might be irrelevant for me if I am only working on Python code.
2. So we can call call `build.py --check-py` in CI workflow for python static checks.

source/qdk_package/check_requirements.txt
fedimser: When typechecker sees import statement and library is not installed, it raises an error (both mypy and pyright).
There are 2 ways to deal with it:
1. Have the library installed (or "type stubs" which provides type signatures for public API) - then you get all calls to the library type-checked.
2. Ignore this import - then all calls to the library are not type-checked, reducing usefulness of type checking.

We can make this choice for every library we import, this is why I created this separate requirements file. I chose to install matplotlib, pandas and pyqir rather than ignoring them.

Yes, I intend to ultimately extend it here to use other libraries, like cirq and qiskit, but I want to be able to make this choice for each library. If I see that some library would take a lot of time to install (e.g. cirq that has a lot of dependencies) and is used only in few limited places, I can ignore it or maybe only install type stubs.

pyrightconfig.json
```diff
@@ -1,5 +1,14 @@
...
   "pythonVersion": "3.10",
-  "extraPaths": ["source/pip", "source/widgets/src", "source/qdk_package/src"],
+  "include": ["source/qdk_package/qdk"],
+  "exclude": [
+    "source/qdk_package/qdk/applications",
+    "source/qdk_package/qdk/azure",
+    "source/qdk_package/qdk/cirq",
+    "source/qdk_package/qdk/simulation",
+    "source/qdk_package/qdk/qiskit",
+    "source/qdk_package/qdk/qre"
```
msoeken, line 10: Do we plan to include this at a later point?

fedimser: Yes, in follow-up PRs.

build.py
amcasey: If we're going to split it out, I'd lean towards splitting all of them out and then having `check` mean "check all".

fedimser: I talked to @swernli  about this and he suggested the same - having an option in build.py to run every check separately, but this can be done in later PR.

This option exists for the following scenario: let's say you did some change to Python and "static Python check" workflow failed. You want to do fixes, verify them locally, push your changes and get workflow green as soon as possible. You do your local changes, run `build.py --check-py`, it passes, then you push your changes and the workflow is green.

Without this option, there would be no way to run python static checks only.

amcasey: Where did we land on this?  It looks like we might have kept python special (i.e. `--check-py` still exists)?

source/qdk_package/qdk/_native.pyi
```diff
@@ -25,15 +25,15 @@ class OutputSemantics(Enum):
     and the semantic checks that are performed.
     """
 
-    Qiskit: OutputSemantics
+    Qiskit = 0
```
amcasey: What do the enum changes do?  Are we just fixing the values so they don't shift when new members are added?

fedimser: MyPy gives an error: 
```
Detected enum "qdk._native.OutputSemantics" in a type stub with zero members. There is a chance this is due to a recent change in the semantics of enum membership. If so, use `member = value` to mark an enum member, instead of `member: type`  
```

pyright doesn't give this error, so I'' undo these changes.

amcasey: Out of curiosity, do you happen to know what changed with enum membership?

source/qdk_package/qdk/_interpreter.py
```diff
@@ -181,7 +182,7 @@ def get_config() -> Config:
     return _get_default_context()._config
 
 
-def python_args_to_interpreter_args(args):
+def python_args_to_interpreter_args(args) -> tuple[Any, ...] | None:
```
idavis: The wrapper claims `tuple[Any, ...] | None`, but a single argument returns its lowered value directly. That can be an integer, float, list, dictionary, or callable handle.

source/qdk_package/qdk/openqasm/_run.py
```diff
@@ -149,7 +149,7 @@ def on_save_events(output: Output) -> None:
                 noise,
                 qubit_loss=qubit_loss,
                 callable=callable,
-                args=args,
+                args=python_args_to_interpreter_args(args),
```
idavis: `python_args_to_interpreter_args(args)` moved from before the shots loop into each call to Interpreter.run. Argument lowering supports one-shot iterables such as generators. The first shot consumes such an argument; subsequent shots receive an empty array or otherwise altered input.

source/qdk_package/qdk/openqasm/_circuit.py
```diff
@@ -85,12 +86,13 @@ def circuit(
...
     )
 
-    if isinstance(source, Callable) and hasattr(source, "__global_callable"):
-        args = python_args_to_interpreter_args(args)
+    if builtins.callable(source) and hasattr(source, "__global_callable"):
         res = get_interpreter().circuit(
-            config, callable=source.__global_callable, args=args
+            config,
+            callable=source.__global_callable,
+            args=python_args_to_interpreter_args(args),
```
idavis: Prefer to keep `args` as local and passed in unless there is good reason to inline.

source/qdk_package/qdk/openqasm/_estimate.py
```diff
@@ -77,10 +78,12 @@ def _coerce_estimator_params(
...
     start = monotonic()
-    if isinstance(source, Callable) and hasattr(source, "__global_callable"):
-        args = python_args_to_interpreter_args(args)
+    if builtins.callable(source) and hasattr(source, "__global_callable"):
         res_str = get_interpreter().estimate(
-            param_str, entry_expr=None, callable=source.__global_callable, args=args
+            param_str,
+            entry_expr=None,
+            callable=source.__global_callable,
+            args=python_args_to_interpreter_args(args),
```
idavis: Prefer to keep args as local and passed in unless there is good reason to inline.

source/qdk_package/qdk/_context.py
```diff
@@ -874,9 +870,10 @@ def circuit(
                 args=args,
             )
         elif isinstance(entry_expr, (GlobalCallable, Closure)):
-            args = self._python_args_to_interpreter_args(args)
             res = self._interpreter.circuit(
-                config=config, callable=entry_expr, args=args
+                config=config,
+                callable=entry_expr,
+                args=self._python_args_to_interpreter_args(args),
```
idavis: Prefer to keep args as local and passed in unless there is good reason to inline.

source/qdk_package/qdk/_native.pyi
```diff
@@ -567,11 +567,12 @@ def compile_visual_circuit_to_qsharp(
 
 def circuit_qasm_program(
     source: str,
+    config: CircuitConfig,
```
idavis, line 570: missing doc below and likely in Rust source

.github/workflows/ci.yml
```diff
@@ -93,6 +93,19 @@ jobs:
       - name: npm check
         run: npm run check
 
+  python-check:
+    name: Python static checks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: "true"
```
idavis: We don't have submodules, this isn't needed.

fedimser: Removed.
Should i also remove it from all other CI workflows?

source/qdk_package/qdk/_native.pyi
fedimser: No.

source/qdk_package/qdk/_native.pyi
fedimser: Added doc

source/qdk_package/qdk/openqasm/_run.py
fedimser: I removed all changes that inline python_args_to_interpreter_args.

source/qdk_package/qdk/openqasm/_estimate.py
fedimser: I removed all changes that inline python_args_to_interpreter_args.

source/qdk_package/qdk/openqasm/_circuit.py
fedimser: I removed all changes that inline python_args_to_interpreter_args.

source/qdk_package/qdk/_native.pyi
fedimser: I really don't know what that error was about and as it was specific to mypy and we decided not to use mypy, I won't investigate that.

build.py
fedimser: My intention is to keep `--check-py` unless there is a reason not to, and I explained above why we need this check.

source/qdk_package/qdk/_interpreter.py
fedimser: I changed output type to `Any`.
I also changed signature of `_python_args_to_interpreter_args` to accept `Any`, because `args` is not necessarily a tuple.

idavis (approved):

PR #3385, opened 2026-06-24
Syntax Highlighting for Stim Grammar

ScottCarda-MS:
Dark theme example:
image

Light theme example:
image


source/vscode/syntaxes/stim.tmLanguage.json
```diff
@@ -0,0 +1,141 @@
...
+      "include": "#measurement-record"
+    },
+    {
+      "include": "#sweep-bit"
+    },
+    {
+      "include": "#pauli-target"
+    },
+    {
+      "include": "#combiner"
```
amcasey, line 40: Is this the only operator and that's why the section isn't called operators?

source/vscode/syntaxes/stim.tmLanguage.json
```diff
@@ -0,0 +1,141 @@
...
+          "name": "constant.numeric.stim"
+        }
+      }
+    },
+    "combiner": {
+      "match": "\\*",
+      "name": "keyword.operator.combiner.stim"
+    },
+    "number": {
+      "match": "(!)?[+-]?\\b\\d+(\\.\\d+)?([eE][+-]?\\d+)?\\b",
```
amcasey: Is the `!` part of the number?

source/vscode/syntaxes/stim.tmLanguage.json
```diff
@@ -0,0 +1,141 @@
...
+      "captures": {
+        "1": {
+          "name": "keyword.operator.negation.stim"
+        }
+      },
+      "name": "constant.numeric.stim"
+    },
+    "bracket": {
+      "match": "[(){}]",
```
amcasey, line 136: What about square brackets?

amcasey (approved):

source/vscode/syntaxes/stim.tmLanguage.json
ScottCarda-MS: There's just the `*` (combiner) and the `!` (negation), and the `!` is handled differently, as you pointed out in your comment below.

source/vscode/syntaxes/stim.tmLanguage.json
ScottCarda-MS: It is here and in the Pauli case as well, as sub-components of the token. But maybe it makes more sense to elevate the `!` to its own scope, like the combiner.

source/vscode/syntaxes/stim.tmLanguage.json
ScottCarda-MS: Square brackets are only ever a part of other special constructs, so they are handled in those constructs, separately from regular bracketing notation.

source/vscode/syntaxes/stim.tmLanguage.json
ScottCarda-MS: I've moved the `!` operator into its own case, as I think that is cleaner. Thanks for the suggestion!

source/vscode/syntaxes/stim.tmLanguage.json
amcasey: To confirm, we do auto-pair them the say way as other brackets?

ScottCarda-MS: yes, all three kinds of brackets are expected to be matched with open and close characters.

PR #3391, opened 2026-06-26
Fix npmAuthenticate in Release stage

amcasey:
It appears that, unlike the Build stage, the Release stage doesn't have an injected Secure Supply Chain Analysis task, so nothing reads AZURE_ARTIFACTS_FEED and creates the necessary .npmrc files.  Manually create the one required for the stage.

### DIFF
--- a/.ado/publish.yml
+++ b/.ado/publish.yml
@@ -510,7 +510,7 @@ extends:
                   artifactName: VSIX
                   targetPath: $(System.DefaultWorkingDirectory)/target/vscode
             steps:
-              # We don't want to check this in because it won't work for external contributors.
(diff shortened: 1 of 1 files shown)


minestarks (approved):

.ado/publish.yml
```diff
@@ -510,6 +510,14 @@ extends:
                   artifactName: VSIX
                   targetPath: $(System.DefaultWorkingDirectory)/target/vscode
             steps:
+              # We don't want to check this in because it won't work for external contributors.
```
amcasey:
```suggestion
              # We don't want to check this .npmrc in because it won't work for external contributors.
```

PR #3394, opened 2026-06-26
Add multi-course support and Python notebook courses to QDK Learning

minestarks:
This change generalizes the QDK Learning feature to support multiple courses of different kinds, including author-provided Python Jupyter notebook courses, while keeping the Quantum Katas as the default.

### DIFF
--- a/courses/circuit-diagrams-new/01-intro/_exercises.json
+++ b/courses/circuit-diagrams-new/01-intro/_exercises.json
@@ -1,0 +1,16 @@
+{
+  "exercises": [
+    {
+      "id": "forty_two",
+      "cellIndex": 7,
+      "title": "Your first Q# expression",
+      "description": "Implement the forty_two() function so it returns 42.",
+      "hints": [
+        "The function just needs to return a value equal to the integer 42.",
+        "The simplest answer is to return the literal `42` itself."
+      ],
+      "solution": "@exercise\ndef forty_two():\n    return qsharp.eval(\"40 + 2\")",
+      "solutionExplanation": "Any Q# expression that evaluates to the integer 42 works. The simplest options are the literal `42` or an arithmetic expression like `40 + 2`."
+    }
+  ]
+}
--- a/courses/circuit-diagrams-new/01-intro/intro.md
+++ b/courses/circuit-diagrams-new/01-intro/intro.md
@@ -1,0 +1,24 @@
+# Getting Started
+
+Welcome to the first unit of the **Generating Circuit Diagrams** course!
+
+In this unit you'll learn the basics of running Q# code from Python using the QDK. You'll:
+
+- Run your first Q# expression from a Python notebook
+- Learn how the notebook exercises and verification work
+- Practice editing and running cells
+
+## How it works
+
+1. Click **Open Notebook** below to open the unit notebook.
+2. Work through the cells top-to-bottom — read the instructions, run examples, and fill in exercises.
+3. When you've completed all exercises, run the final cell to mark the unit complete.
+4. Come back here and click **Next** to continue to the next unit.
+
+## Before you start
+
+This course runs in its own Python environment. If the notebook's kernel
+won't start, or the first cell reports a problem, set up and check your
+environment here first:
+
+👉 [Check my environment](command:qsharp-vscode.learningDoctor)
--- a/courses/circuit-diagrams-new/02-circuits/circuits.ipynb
+++ b/courses/circuit-diagrams-new/02-circuits/circuits.ipynb
@@ -1,0 +1,203 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "3db5183b",
+   "metadata": {},
+   "source": [
+    "# Rendering Circuit Diagrams\n",
+    "\n",
+    "In this unit you'll learn how to use the `circuit()` API and the `Circuit` widget to generate and display circuit diagrams from Q# operations.\n",
+    "\n",
+    "We'll assume you're already familiar with Q# and quantum circuits. The focus here is on the **Python API** — what options are available and how they affect the rendered output.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "bda994e6",
+   "metadata": {},
+   "source": [
+    "## Environment Check\n",
+    "\n",
+    "Run the cell below to verify your environment is set up correctly."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "aaf39e45",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from _unit import check_env\n",
+    "\n",
+    "check_env()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "34989e87",
+   "metadata": {},
+   "source": [
+    "## Setup\n",
+    "\n",
+    "First, let's import the tools we'll use and define some Q# operations to work with throughout this unit."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4c2e67c9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qdk import qsharp\n",
+    "from qdk.qsharp import circuit\n",
+    "from qdk.widgets import Circuit\n",
+    "\n",
+    "# Define a few operations we'll use throughout this unit.\n",
+    "OPERATIONS = \"\"\"\n",
+    "    operation BellPair() : (Result, Result) {\n",
+    "        use (a, b) = (Qubit(), Qubit());\n",
+    "        H(a);\n",
+    "        CNOT(a, b);\n",
+    "        return (M(a), M(b));\n",
+    "    }\n",
+    "\n",
+    "    operation GHZ(n : Int) : Result[] {\n",
+    "        use qs = Qubit[n];\n",
+    "        H(qs[0]);\n",
+    "        for i in 1..n-1 {\n",
+    "            CNOT(qs[0], qs[i]);\n",
+    "        }\n",
+    "        let results = MeasureEachZ(qs);\n",
+    "        ResetAll(qs);\n",
+    "        return results;\n",
+    "    }\n",
+    "\n",
+    "    operation PrepareCatState(qs : Qubit[]) : Unit {\n",
+    "        H(qs[0]);\n",
+    "        for i in 1..Length(qs)-1 {\n",
+    "            CNOT(qs[0], qs[i]);\n",
+    "        }\n",
+    "    }\n",
+    "\"\"\"\n",
+    "\n",
+    "qsharp.eval(OPERATIONS)\n",
+    "\n",
+    "print(\"Operations defined.\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "09c7e760",
+   "metadata": {},
+   "source": [
+    "## Example: Basic circuit generation with an entry expression\n",
+    "\n",
+    "`circuit()` takes an **entry expression** — a string that calls a Q# operation — and returns a `Circuit` object. Wrap it in the `Circuit` widget to render it in the notebook.\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ab1207cb",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Generate a circuit from an entry expression\n",
+    "bell = circuit(\"BellPair()\")\n",
+    "Circuit(bell)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d0e0ed73",
+   "metadata": {},
+   "source": [
+    "## Exercise: Render with `operation=`\n",
+    "\n",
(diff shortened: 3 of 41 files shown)


courses/circuit-diagrams-new/01-intro/_exercises.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "id": "forty_two",
+      "cellIndex": 7,
+      "title": "Your first Q# expression",
+      "description": "Implement the forty_two() function so it returns 42.",
+      "hints": [
+        "The function just needs to return a value equal to the integer 42.",
+        "The simplest answer is to return the literal `42` itself."
+      ],
+      "solution": "@exercise\ndef forty_two():\n    return qsharp.eval(\"40 + 2\")",
```
amcasey, line 12: In the katas, we sometimes found that we wanted to provide multiple solutions.  If this is a generalization, we might want that here?

courses/circuit-diagrams-new/02-circuits/circuits.ipynb
```diff
@@ -0,0 +1,203 @@
...
+   "id": "4c2e67c9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qdk import qsharp\n",
+    "from qdk.qsharp import circuit\n",
+    "from qdk.widgets import Circuit\n",
+    "\n",
+    "# Define a few operations we'll use throughout this unit.\n",
+    "OPERATIONS = \"\"\"\n",
```
amcasey, line 59: I'm not familiar enough with python to suggest an alternative, but this seems like a painful level of escaping.

amcasey: I guess I don't actually know how this got here.  Maybe you never actually touch this file and the escaping is done for you when it's persisted?

courses/circuit-diagrams-new/course.json
```diff
@@ -0,0 +1,22 @@
+{
+  "schemaVersion": 1,
+  "id": "circuit-diagrams",
+  "title": "Generating Circuit Diagrams",
+  "shortDescription": "Build and visualize quantum circuits with the QDK in Python notebooks.",
+  "readme": "README.md",
```
amcasey, line 6: Why reference the readme but not the project?  They seem similar to me, but I may be missing something.

source/vscode/agents/qdk-learning.agent.md
```diff
@@ -1,22 +1,24 @@
...
 
 ## Definitions
 
-Following is a user-ready description of the Quantum Katas. You may refer to it if the user asks what the katas are or how they work.
+The **Quantum Katas** is the flagship course. Following is a user-ready description. You may refer to it if the user asks what the katas are or how they work.
 
 > Quantum Katas (_kaˑta_ | kah-tuh — Japanese for "form", a pattern of learning and practicing new skills) are self-paced, AI-assisted tutorials for quantum computing and Q# programming. Each tutorial includes relevant theory and interactive hands-on exercises designed to test knowledge.
 
-The tools refer to each kata as a "unit." Each unit contains ordered activities (lessons, examples, exercises).
+The tools refer to each unit of a course as a "unit." Each unit contains ordered activities (lessons, examples, exercises).
```
amcasey, line 21: This change makes it a little tautological.

source/vscode/agents/qdk-learning.agent.md
```diff
@@ -28,13 +30,31 @@ The tools refer to each kata as a "unit." Each unit contains ordered activities
...
+| "What courses are available?"         | `list-courses`  | Returns the available courses and the active course id.                   |
+| "Switch to …" / "Open the … course"   | `switch-course` | Pass the `courseId`. Switching changes the active course and position.    |
+| "Tell me about this course"           | `course-info`   | Returns the course descriptor and README (defaults to the active course). |
+| "Diagnose" / "Set up the environment" | `doctor`        | Runs environment diagnostics for the active course (Python courses).      |
+
+**Handling guidance:**
+
+- When the user asks to change courses, call `list-courses` first if you're unsure of the exact `courseId`, match the user's request to a course, then call `switch-course`. After switching, call `show` to surface the new course's current activity and briefly tell the user where they landed.
+- Drop-in courses run author-provided code and only load in a **trusted** workspace. If a drop-in course doesn't appear or won't run, the workspace may be in Restricted Mode — suggest trusting the workspace.
```
amcasey, line 54: I'm a little nervous about "suggest trusting the workspace".  Maybe "explain that the user will need to trust the workspace to continue and should only do so if they trust the course material provider"?

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,491 @@
...
+   * Locate a Python interpreter to bootstrap a venv. Prefers the Python
+   * extension's active interpreter, falling back to `python3`/`python`.
+   * Returns `undefined` when none can be determined.
+   */
+  async ensureInterpreter(): Promise<string | undefined> {
+    if (!this.supported) {
+      return undefined;
+    }
+    const fromExtension = await this.activeInterpreterPath();
+    return fromExtension ?? "python3";
```
amcasey, line 53: I assume you've already tested this on windows, but I've definitely run into cases where this needs to be `python3.exe`, rather than `python3`.

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,491 @@
...
+  async createVenv(courseRoot: vscode.Uri, pythonSpec?: string): Promise<void> {
+    if (!this.supported || (await this.venvExists(courseRoot))) {
+      return;
+    }
+    const cwd = courseRoot;
+    const venvPath = this.venvUri(courseRoot).fsPath;
+
+    // Prefer `uv` when it's available — it's faster and is the modern
+    // default tooling. Fall back to the standard library `venv` module.
+    if (await this.uvAvailable()) {
```
amcasey, line 79: This may be covered below, but I'm wondering if we want a way for users to force venv, rather than uv, even if the latter is available.

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,491 @@
...
+      python,
+      ["-c", "import venv, ensurepip"],
+      cwd,
+    );
+    if (preflight !== 0) {
+      throw new Error(
+        "This Python installation can't create virtual environments " +
+          "(the `venv`/`ensurepip` modules are missing). On Debian/Ubuntu " +
+          "install them with `sudo apt install python3-venv` (matching your " +
+          "Python version, e.g. `python3.12-venv`), then try again.",
```
amcasey, line 121: Don't we know their python version?

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,491 @@
...
+
+  // ─── Private: swappable OS interaction ───
+
+  /**
+   * The Python extension's stable `environments` API, or `undefined` when
+   * the extension is unavailable. Only the documented, non-proposed members
+   * are typed here.
+   */
+  private async pythonEnvironmentsApi(): Promise<
+    | {
```
amcasey, line 374: Nit: why not name this type?

source/vscode/src/learning/constants.ts
```diff
@@ -10,12 +10,21 @@ export const LEARNING_WORKSPACE_RELATIVE_PATH = `./${LEARNING_WORKSPACE_FOLDER}`
...
+
 /** Context key set when a learning workspace is detected. */
 export const LEARNING_WORKSPACE_DETECTED_CONTEXT =
   "qsharp-vscode.learningWorkspaceDetected";
 
 /** Course ID for the built-in Quantum Katas. */
 export const KATAS_COURSE_ID = "katas";
 
+/** Per-course virtual environment folder (under the course working copy). */
+export const LEARNING_VENV_DIR = ".venv";
```
amcasey, line 27: Would there be any value in making this name more unique?

amcasey: Seems like it's in a folder people won't be editing and it might break notebook discovery, so probably not?

source/vscode/src/learning/dropInCourseProvider.ts
```diff
@@ -0,0 +1,439 @@
...
+      const manifest = await this.readManifest(dir);
+      if (manifest) {
+        locations.push({ dir, manifest });
+      }
+    }
+    return locations;
+  }
+
+  /** Read and JSON-parse a course manifest, or `undefined` if absent/invalid. */
+  private async readManifest(
```
amcasey, line 124: Remind me what guardrails we need when reading untrusted (?) files from disk.  Or maybe this only works if they're in a trusted folder?

courses/circuit-diagrams-new/course.json
```diff
@@ -0,0 +1,22 @@
...
+      "dir": "01-intro"
+    },
+    {
+      "id": "circuits",
+      "title": "Circuit Diagrams",
+      "dir": "02-circuits"
+    }
+  ],
+  "environment": {
+    "importChecks": ["qdk", "qdk.widgets"]
```
amcasey, line 20: Could we discover these from the cells?

amcasey: Per our offline discussion, that's probably overkill

courses/circuit-diagrams-new/course.json
amcasey: We probably don't need to reference either - they have well-known names.

courses/circuit-diagrams-new/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
```
amcasey, line 1: This (using a standard python project format) is part of achieving our goal that you could get the basic course experience in any jupyter environment.

courses/circuit-diagrams-new/01-intro/_exercises.json
```diff
@@ -0,0 +1,16 @@
+{
+  "exercises": [
+    {
+      "id": "forty_two",
+      "cellIndex": 7,
```
amcasey, line 5: This seems brittle.  Is the cell ID more stable?

courses/circuit-diagrams-new/01-intro/intro.md
```diff
@@ -0,0 +1,24 @@
...
+3. When you've completed all exercises, run the final cell to mark the unit complete.
+4. Come back here and click **Next** to continue to the next unit.
+
+## Before you start
+
+This course runs in its own Python environment. If the notebook's kernel
+won't start, or the first cell reports a problem, set up and check your
+environment here first:
+
+👉 [Check my environment](command:qsharp-vscode.learningDoctor)
```
amcasey, line 24: This needs to be on screen somewhere, in case the notebook doesn't work.  This file is basically an excuse to do that - we could find a different place for it.

source/vscode/agents/qdk-learning.agent.md
amcasey: This is moot - our extension is entirely disabled in untrusted workspaces

source/vscode/package.json
```diff
@@ -371,6 +387,28 @@
         }
       ],
       "view/item/context": [
+        {
```
amcasey, line 390: This hasn't been reviewed by humans.

source/vscode/src/learning/constants.ts
amcasey: Aside: the default discovery system only looks in the root and one level down - it doesn't recursively search for .venv.  There may be a workspace setting to override it

source/vscode/src/learning/courseProvider.ts
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) Microsoft Corporation.
```
amcasey, line 1: Might be mergeable with catalog.ts

source/vscode/src/learning/notebookCellStatusBar.ts
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as vscode from "vscode";
+import type { LearningService } from "./service.js";
+
+/**
```
minestarks, line 7: The way to check whether a cell is an exercise is to refer to the exercies.json metadata

amcasey (commented):
The old kata notebooks at github.com/microsoft/quantumkatas may be useful references

courses/circuit-diagrams-new/02-circuits/circuits.ipynb
amcasey: No one hand-edits these files

source/vscode/src/learning/python/environment.ts
amcasey: We do or we could, so we should improve this error message

courses/circuit-diagrams-new/01-intro/_exercises.json
amcasey: Probably, but confirm

amcasey:
I'll make a new PR when I understand things better.

source/vscode/src/learning/dropInCourseProvider.ts
```diff
@@ -0,0 +1,439 @@
+// Copyright (c) Microsoft Corporation.
```
amcasey, line 1: Might be mergeable with catalog.ts

PR #3395, opened 2026-06-26
Have RCA reject unsafe qubit release from dynamic context

### LINKED ISSUE #3310: Partial Eval can fail with "qubit double release" if release occurs in dynamic contexts
If the conditions surrounding a qubit release are dynamic, then Partial Eval may end up running multiple calls to release, causing QIR generation to fail with "partial evaluation failed with error: qubit double release." With the newer FIR passes transforming away dynamic returns, it becomes easier to hit this, but the bug was already present and can be hit with manual qubit management:

```qsharp
import QIR.Runtime.*;
@EntryPoint(Adaptive_RI)
operation Main() : Unit {
    let q = __quantum__rt__qubit_allocate();
    let b = M(q) == One;
    if b {
        __quantum__rt__qubit_release(q);
    } else {
        __quantum__rt__qubit_release(q);
    }
}
```

swernli:
This change updates the analysis performed to check if a given call expression is a call to qubit release in a dynamic scope with an argument that comes from a different dynamic scope. Without dynamic qubit allocation, this would result in a double release during codegen, so we tag this pattern as requiring dynamic qubit allocation and provide design-time feedback. Fixes #3310

The error is flagged on the qubit allocation itself, since that is the corresponding span for the compiler generated release. The error message specifically calls out dynamic early returns, since that is the most common cause of these errors:
image

### DIFF
--- a/source/compiler/qsc_rca/src/core.rs
+++ b/source/compiler/qsc_rca/src/core.rs
@@ -469,6 +469,43 @@ impl<'a> Analyzer<'a> {
             compute_kind.aggregate_value_kind(value_kind);
         }
 
+        if !self.target_capabilities.is_empty()
+            && callable_decl.name.name.as_ref() == "__quantum__rt__qubit_release"
+        {
+            // If the analysis is for a target non-empty capabilities, we need to check if this is
+            // a non-deterministic qubit release, which requires the dynamic qubit allocation feature.
+            // (We ignore this check for Base Profile as other dynamism will already be flagged and this
+            // specific case does not come up in isolation).
+            // We know that only one intrinsic callable named "__quantum__rt__qubit_release" is allowed by the compiler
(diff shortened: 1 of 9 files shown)


source/compiler/qsc_rca/src/core.rs
```diff
@@ -469,6 +469,43 @@ impl<'a> Analyzer<'a> {
             compute_kind.aggregate_value_kind(value_kind);
         }
 
+        if !self.target_capabilities.is_empty()
+            && callable_decl.name.name.as_ref() == "__quantum__rt__qubit_release"
```
idavis, line 473: Note for a future rework, we really should centrallize the defs for these intrinsics so that we don't have the string literals all over the rust code

idavis (approved):

PR #3399, opened 2026-06-26
Update default profile selection.

### LINKED ISSUE #3384: Update default profile choice in Python and VS Code relative to `Adaptive`
We need to review the places where the compiler chooses a default value for the target profile to see which ones are appropriate to update to the new `Adaptive` profile. Not all of them may need it; it's possible some should use a more restrictive settings.

idavis:
Updating default values of QIR profile to adaptive where appropriate.

Fixes #3384

### DIFF
--- a/source/qdk_package/qdk/openqasm/_compile.py
+++ b/source/qdk_package/qdk/openqasm/_compile.py
@@ -15,7 +15,6 @@ from .._interpreter import (
     ipython_helper,
     python_args_to_interpreter_args,
 )
-from .._native import TargetProfile
 from .. import telemetry_events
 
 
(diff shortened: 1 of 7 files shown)


source/qdk_package/qdk/openqasm/_compile.py
```diff
@@ -78,7 +78,7 @@ def compile(
         if "search_path" not in kwargs:
             kwargs["search_path"] = "."
         if "target_profile" not in kwargs:
-            kwargs["target_profile"] = TargetProfile.Base
+            kwargs["target_profile"] = TargetProfile.Adaptive
```
swernli: is it worth clarifying in the doc comment above what the default behavior is (especially after your recent fixes)? If memory serves, it's the pragma value in the code that takes precedence, then this fallback value, right?

idavis: Good catch, look at https://github.com/microsoft/qdk/pull/3399/commits/17bf4ffe9b3182f0b64783afd5c1ab311949ef46

swernli: Nice, yeah, moving the default to there works better with the existing patterns.

source/vscode/src/azure/providerProperties.ts
```diff
@@ -34,6 +34,6 @@ export function getPreferredTargetProfile(target: string): TargetProfile {
   } else if (target.startsWith("quantinuum") && target.includes(".h2")) {
     return "adaptive_ri";
   } else {
-    return "adaptive_rif";
+    return "adaptive";
```
minestarks: Is this known to work with the providers we currently have? I feel like submitting to hardware targets might be one place we're strict about setting exact capabilities.

swernli (approved):

source/vscode/src/azure/providerProperties.ts
idavis: changed back for now, thanks

PR #3400, opened 2026-06-26
Arithmetic test utils

fedimser:
* Q# test functions to test arithmetic operations:
  * ApplyBigInt - writes BigInt to register.
  * MeasureBigInt - reads BigInt from register and resets register to 0.
  * TestArithmeticOp - writes one or more inputs to registers, applies operation and reads output.
* Python class `test_utils.ArithmeticOpTester` to run Q# arithmetic operation on given inputs. Supports input as string or as callable. 
  * Wrapper around Std.ArithmeticTestUtils.TestArithmeticOp.
  * Optimized for testing the same operation on many inputs. 
* `Std.ArithmeticTestUtils.ApplyClassicalFunction`. This Q# function allows to specify classical arithmetic function (as Q# closure) and turn it into arithmetic gate that will apply that function to every basis state.
  * Implemented in sparse simulator only.
  * Also has Q#-only wrappers that provides controlled version, and versions acting on 2 and 3 arguments.
  * This addresses https://github.com/microsoft/qdk/issues/3272


library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+/// ## val
+/// The non-negative integer to encode.
+/// ## reg
+/// The little-endian qubit register to write into. Expected to be in the
+/// |0⟩ state.
+operation ApplyBigInt(val : BigInt, reg : Qubit[]) : Unit is Adj + Ctl {
+    let bits = Std.Convert.BigIntAsBoolArray(val, Length(reg));
+    ApplyPauliFromBitString(PauliX, true, bits, reg);
+}
+
```
swernli: This operation seems generally useful outside of just the test utils, so I think it's worth adding to `Std.Arithmetic` and matching the naming conventions in there. Maybe something like `Std.Arithmetic.ApplyL`?

library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+/// ## reg
+/// The little-endian qubit register to measure. Reset to the |0⟩ state on
+/// return.
+///
+/// # Output
+/// The integer value encoded in `reg`, in the range `0 .. 2^Length(reg) - 1`.
+operation MeasureBigInt(reg : Qubit[]) : BigInt {
+    let result = Std.Measurement.MResetEachZ(reg);
+    Std.Convert.BoolArrayAsBigInt(Std.Convert.ResultArrayAsBoolArray(result))
+}
```
swernli: This too could be added to a more general spot... I'd suggest `Std.Measurement`

library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+        set total += sz;
+    }
+    use allQubits = Qubit[total];
+    mutable regs : Qubit[][] = [];
+    mutable offset = 0;
+    for sz in sizes {
+        set regs += [allQubits[offset..offset + sz - 1]];
+        set offset += sz;
+    }
+    Std.Arrays.ForEach(Std.ArithmeticTestUtils.ApplyBigInt, Std.Arrays.Zipped(vals, regs));
```
swernli: In general, our pattern in the libraries is to use imports at the top of the file, which makes it easier to see all together what functionality the file depends on.

library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+/// The little-endian register holding the first argument.
+/// ## reg2
+/// The little-endian register holding the second argument.
+/// ## reg3
+/// The little-endian register holding the third argument.
+///
+/// # Remarks
+/// Supported only on the sparse simulator. All caveats of
+/// `ApplyClassicalFunction` regarding injectivity and output range apply here.
+operation ApplyClassicalFunction3(f : (BigInt, BigInt, BigInt) -> (BigInt, BigInt, BigInt), reg1 : Qubit[], reg2 : Qubit[], reg3 : Qubit[]) : Unit is Ctl {
```
swernli: Rather than have variations for 2 and 3, would it make sense to have one that takes an arbitrary number as a list? Something like:
```qsharp
import Std.Arrays.Flattened;
import Std.Arrays.ForEach;
operation ApplyClassicalFunctionN(f : (BigInt[]) -> (BigInt[]), regs : Qubit[][]) : Unit is Ctl {
    mutable n = [];
    for reg in regs {
        set n += [Length(reg)];
    }
    let n = n;
    mutable masks = [];
    for ni in n {
        set masks += [(1L <<< ni) - 1L];
    }
    let masks = masks;
    let f : (BigInt) -> BigInt = x -> {
        mutable xs = [];
        mutable shift = 0;
        for i in 0..Length(n)-1 {
            let ni = n[i];
            let mask = masks[i];
            let xi = (x >>> shift) &&& mask;
            xs += [xi];
            shift += ni;
        }
        let ys = f(xs);
        mutable y = 0L;
        set shift = 0;
        for i in 0..Length(n)-1 {
            let yi = ys[i] &&& masks[i];
            y += yi <<< shift;
            shift += n[i];
        }
        return y;
    };
    ApplyClassicalFunction(f, Flattened(regs));
}
```
I'm not sure if that is reasonable from an ergonomics of usage standpoint, but seems like it would allow for arbitrary expansion beyond the limit of 3.

library/std/src/Std/ArithmeticTestUtils.qs
fedimser: Yesterday I learned that we already have exactly this operation in Canon, called `ApplyXorInPlaceL`, so I will remove this one.

library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+/// # Input
+/// ## f
+/// The classical function to apply. Must map every input in
+/// `0 .. 2^Length(target) - 1` to an output in the same range, and must be
+/// injective over those inputs.
+/// ## target
+/// The little-endian qubit register to transform.
+///
+/// # Remarks
+/// Supported only on the sparse simulator.
```
swernli: Since remarks don't show up in the hover by default, this would we worth adding to the `# Summary` tag above so it's easier for folks to see.

library/std/src/Std/ArithmeticTestUtils.qs
```diff
@@ -0,0 +1,241 @@
...
+/// ## sizes
+/// The number of qubits to allocate for each register.
+/// ## vals
+/// The initial integer value for each register. Must have the same length
+/// as `sizes`, and `vals[i]` must fit in `sizes[i]` qubits.
+///
+/// # Output
+/// The integer values held by the registers after `op` has been applied,
+/// in the same order as the inputs.
+operation TestArithmeticOp(
```
swernli, line 41: I think it would be worth annotating this operation and each of the `ApplyClassicalFunction` variants with `@Config(Unrestricted)` so that they cannot be called from code compiling for QIR. That matches the set up we have for other simulation-only APIs like `DrawRandomInt`.

library/std/src/Std/ArithmeticTestUtils.qs
fedimser: Done. I think it's better.

library/std/src/Std/ArithmeticTestUtils.qs
fedimser: Done.

library/std/src/Std/ArithmeticTestUtils.qs
fedimser: Strictly speaking, TestArithmeticOp doesn't have anything that prevents it from being compiled to QIR.

But the intention is to use it only in tests with simulator, so I guess we can add the annotation.

Can you confirm that with `@Config(Unrestricted)` it can be also used in resource estimator?

swernli: Yes, it can still be used with resource estimator, just not if the code is configured for a profile via something like `@EntryPoint(Adaptive)`.

library/std/src/Std/Measurement.qs
```diff
@@ -3,8 +3,9 @@
 
 
 import Std.Core.*;
-import Std.Intrinsic.*;
+import Std.Convert.*;
 import Std.Diagnostics.*;
+import Std.Intrinsic.*;
```
swernli: Std.Intrinsic is part of the prelude, so does not need to be explicitly imported.

```suggestion
```

source/simulators/src/sparse_state_simulator/tests.rs
```diff
@@ -721,3 +721,51 @@ fn test_global_phase_dropped_when_all_qubits_released() {
...
+    assert_eq!(state.len(), 1);
+    assert_eq!(state[0].0, BigUint::from(6u64));
+
+    sim.release(q2);
+    sim.release(q1);
+    sim.release(q0);
+}
+
+#[test]
+fn test_apply_arithmetic_gate_double() {
```
swernli, line 750: I think it would be helpful too to have a test that shows that an arithmetic gate that operates on a subset of the currently allocted qubits with an array that is not ordered works as expected. Something like an allocation of 5 qubits and then an operation on the array `&[q3, q1, q4]`.

swernli (approved):
One minor tweak to the library file (which will unfortunately churn some snapshots again, I believe), and one more test of the simulator implementation, and I think this is ready to go in!

library/std/src/Std/Measurement.qs
fedimser: removed

source/simulators/src/sparse_state_simulator/tests.rs
fedimser: Added such a test.

PR #3404, opened 2026-06-28
Workspace links

billti:
This is based off the PR at https://github.com/microsoft/qdk/pull/3303 , but with a few tweaks to logic and refactoring to address the comments left there (see first commit), and updated against latest main.


ScottCarda-MS (approved):
Looks good, thanks Bill! I updated the comments to reflect your changes, and I added `async`/`await` to some places in the API key path. I think this is ready to merge now.

PR #3407, opened 2026-06-29
Interpreter-level config

### LINKED ISSUE #3271: [Feature request] Configuration map accessible from Q# code

There are multiple cases when this would be helpful:
1. https://github.com/microsoft/qdk/issues/3109 - knowing whether we are inside resource estimator.
2. In discussion on https://github.com/microsoft/qdk/pull/3204 it was suggested that it would be good to be able to enable manual memory-compute from Python.
3. This repository https://github.com/microsoft/QuantumEllipticCurves defines some parameters that specify what metric to optimize and the uses them like this: https://github.com/microsoft/QuantumEllipticCurves/blob/dbf4836afaf7a9fab813cbc0970e65af85a6f93a/MicrosoftQuantumCrypto/Basics.qs#L173 . 
    * This is a common pattern in resource estimation research - we would like to have one algorithm implementation but be able to optimize it for different metrics using different building blocks.

All these can be addressed by having a configuration (dictionary keyed by strings) that will be available to Q# code.

Suggested Python syntax (for evaluate, run, logical_counts etc):
```
ctx = qdk.Context()
ctx.set_config("use_memory_qubits", True)
ctx.set_config("minimize_depth", True)
ctx.set_config("minimize_t_count", False)
ctx.eval(...)
```

Suggested Q# intrinsics to access the config:

```
function GetConfigInt(key: name): Int;
function GetConfigDouble(key: name): Double;
function GetConfigBool(key: name): Bool;
function GetConfigString(key: name): String;
```

Then, for example, [this code](https://github.com/microsoft/QuantumEllipticCurves/blob/dbf4836afaf7a9fab813cbc0970e65af85a6f93a/MicrosoftQuantumCrypto/Basics.qs#L173-L177) can be rewritten as:

```
if (GetConfigBool("minimize_depth")){
    ApplyLowDepthAnd(control1, control2, target);
} else {
    ApplyAnd(control1, control2, target);
}
```

And this can be used to know whether we are in resource estimator or not. We will introduce some "system configs" that are always available (and forbid users to override them). For example, "is_resource_estimating" will be bool config that will be set to true for RE backend and to False for other backends.

This can be implemented in Rust (in Interpreter or in each backend) by keeping `map<String, Value>` where Value is defined as `enum Value {Int(i64), Float(f64), Bool(bool), String(String)}`. If value is not set, default value is returned.

The alternative is to pass configuration as operation arguments: define `struct Config` and add all configuration there. Then pass it as first or last arguments from entry point to all methods that need this. This is somewhat ugly.

fedimser:
This PR adds a key-value configuration map to the interpreter, that can be written into from interpreter call site and can be read from inside Q# code. It also adds Python API to write into it. Fixes https://github.com/microsoft/qdk/issues/3271

### DIFF
--- a/library/std/src/Std/Diagnostics.qs
+++ b/library/std/src/Std/Diagnostics.qs
@@ -480,6 +480,30 @@ operation PostSelectZ(res : Result, qubit : Qubit) : Unit {
     body intrinsic;
 }
 
+/// # Summary
+/// Returns a value from the host-provided configuration map.
+///
+/// # Description
+/// Looks up `name` in a read-only configuration map provided by the host runtime.
+/// If the key does not exist, returns `defaultValue`.
+///
+/// # Input
+/// ## name
+/// The configuration key.
+/// ## defaultValue
+/// The value to return when `name` is not present.
+///
+/// # Output
+/// The configured value for `name`, or `defaultValue` if the key is absent.
+///
+/// # Example
+/// ```qsharp
+/// let experimentName = Std.Diagnostics.GetConfig("experiment_name", "");
+/// ```
+function GetConfig<'T>(name : String, defaultValue : 'T) : 'T {
+    body intrinsic;
+}
+
 export
     DumpMachine,
     DumpRegister,
@@ -501,4 +525,5 @@ export
     PhaseFlipNoise,
     DepolarizingNoise,
     NoNoise,
-    PostSelectZ;
+    PostSelectZ,
+    GetConfig;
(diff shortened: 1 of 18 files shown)


fedimser:
We had a design discussion and agreed on fundamentally different design approach:
* Pass key-value config map in Context.init().
* Have single function GetConfig(value, default).
* Instead of implementing an intrinsic, have HIR pass replacing GetConfig call expression with a literal (which will be either coming from the map or from default value which must be a literal).

I will convert this PR to draft and create a new PR for a new design. I will close this PR after the alternative PR is merged.

PR #3409, opened 2026-06-29
Add support for PREPARE block in stim

joao-boechat:
# Add support for `PREPARE` blocks in Stim

## Summary

Adds a `PREPARE { ... }` block to the Stim front end: a *repeat-until-success* region used to deterministically prepare a state that otherwise only occurs probabilistically. This replaces the previous `#!preselect_*` comment-annotation mechanism.

## How it works

- A `PREPARE` block runs its body, then re-runs the **entire block** until every `REQUIRE` inside it is satisfied.
- `REQUIRE rec[-1] rec[-2] ...` reads measurement records (indexed backwards from the most recent measurement) and combines them by **XOR (parity)**. If the parity is `1` the block restarts; if `0` execution continues.
- A target can be **negated** (`REQUIRE !rec[-1]`), which flips that record's contribution to the parity — i.e. it requires the record to be `1` instead of `0`.
- Each block is lowered to a labeled basic block (`prepare_N`) plus a conditional branch back to it. `REQUIRE` may only reference records produced inside its own enclosing block; nested blocks are supported.

## Modifications

- **`qir.rs`** — lowering for `PREPARE`/`REQUIRE`: prepare/continue labels, measurement-record resolution with scope tracking, XOR parity fold, negation (`xor … , true`), and new diagnostics (`RequireOutsidePrepareBlock`, `MeasurementRecordOutOfBounds`, `MeasurementRecordOutOfScope`).
- **`parser.rs`** — parse `PREPARE` blocks and `REQUIRE`; allow negation on measurement-record targets.
- **`lex.rs`** — drop the old `#!` preselect-annotation lexing.
- **Tests** — new `qir/tests/prepare_block.rs` and parser target tests; removed the obsolete `qir/tests/preselect.rs`.
- **Sample** — `samples/notebooks/stim_prepare.ipynb` demonstrating preselection, negated requirements, multiple/parity requirements, and nested blocks.


source/compiler/stim_compiler/src/qir/tests.rs
```diff
@@ -9,7 +9,8 @@ mod noise_channels;
 mod noise_channels_broadcasting;
 mod pair_measurements;
 mod pair_measurements_broadcasting;
-mod preselect;
+// mod prepare_block;
```
ScottCarda-MS: Seems like a typo comment?

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -254,13 +254,8 @@ impl Iterator for Lexer<'_> {
...
             }
             '#' => {
-                if self.chars.next_if(|(_, c)| *c == '!').is_some() {
-                    self.eat_while(|c| !c.is_whitespace());
-                    TokenKind::InstructionName
-                } else {
-                    self.comment();
-                    return self.next();
-                }
+                self.comment();
```
ScottCarda-MS, line 257: Are we getting rid of our special handling for `#!...` directives altogether? I thought it was a useful feature, but without the preselect, I guess we don't have any such directives we support anymore.

source/compiler/stim_compiler/src/qir/tests.rs
joao-boechat: yep! Thanks for noticing, let me undo it

source/compiler/stim_compiler/src/lex.rs
joao-boechat: We are. We decided that we'd rather bite the bullet and extend the language where needed (for example with these PREPARE blocks), but try to use the language as much as possible

orpuente-MS (approved):

billti (approved):

PR #3410, opened 2026-06-29
Pass callables as arguments end-to-end through defunctionalization

idavis:
## Summary

In Q#, you can hand one operation or function to another as a value - for example passing `AllH` into `InvokeWithQubits`, or building a little function on the fly like `Rz(theta, _)` and passing that.

Hardware and the QIR output format can't deal with "a function as a value at runtime" - every call has to point at a specific, known target. So the compiler has a stage called **defunctionalization** whose job is to look at each place a function-value is used and rewrite it into a plain, direct call to the real target. This branch makes that stage dramatically more capable and fixes a pile of cases where it previously produced wrong code, crashed, or gave up.

## The bugs fixed

**1. Only one function-argument at a time.**
When an operation took several function arguments (e.g. `InvokeThree(first, second, third)`), the compiler couldn't resolve them all together if there was captured state. This branch handles multiple function-arguments to the same call in a single pass.

**2. Closures and "partially applied" functions didn't survive.**
Things like `Rz(theta, _)` or `x -> x + inc` capture a value (`theta`, `inc`) and carry it around. Passing these, especially a closure that wraps *another* closure, often failed or lost the captured value. Now the captured values are correctly threaded through as extra arguments, including nested cases.

**3. Passing a function from Python as a top-level argument produced broken QIR/circuits.**
When you called `qsharp.compile(InvokeWithQubits, 3, AllH)` from Python, the function-value created by the interpreter carried stale internal type information. Once that value became part of the program to compile, it violated the compiler's own consistency checks. The fix normalizes those type signatures before compiling, so QIR generation and circuit drawing now work from Python-supplied callables.

**4. Generic functions behind a closure weren't specialized.**
A closure like `Rz(theta, _)` points at a *generic* library operation. Monomorphization only fixed up direct calls, not closures. It now also rewrites closure targets, so the concrete version actually gets created.

**5. Arrays of functions and index-based dispatch.**
Selecting a function out of an array (`ops[i]`) - including arrays of closures, tuples, and functions stored in record fields, now correctly turns into an `if`/`else` that dispatches to the right target, evaluating the index only once.

**6. Wrong answers from reassignment and control-flow tracking.**
The stage tracks "which function does this variable actually hold at this point." Several cases were wrong:
- A reassignment (`f = Bar;`) buried inside a sub-expression wasn't noticed in the right order.
- Reassignments on the never-executed side of `and`/`or`/`and=` short-circuits were wrongly applied.
- `w/` update expressions were evaluated in the wrong order.
- Reassignment inside a loop is now correctly treated as "unknown," producing a clear error instead of silently mis-specializing.
- A function chosen conditionally (`if cond { X } else { Y }`) now becomes proper branch-by-branch dispatch instead of dropping a branch.

**7. Side effects running twice.**
When a function was selected by a condition that itself did something (e.g. a measurement), that side effect could be emitted more than once. Coordinating with the condition-normalization stage ensures each side effect runs exactly once.

**8. Dynamic `Angle` return in std lib.**
The angle calculations for custom gates in OpenQASM hit an issue where partial evaluation cannot lower a dynamic branch that returns an Angle struct. The calculation is now done inline eliminating the dynamic `Angle` return during construction.

## The Fixes

Rather than the textbook approach using dispatch tables and dynamic dispatch, we must use **specialization**: for each call site where it can figure out the concrete function being passed, it makes a dedicated copy of the higher-order operation with the function-argument baked in as a direct call. `Apply(q => Y(q), target)` becomes a call to a generated `Apply_specialized_Y`. This is required for QIR profiles as they don't support runtime resloved calls.

To do that reliably, the branch:

- **Analyzes data flow more precisely** - it tracks, at each program point and even inside sub-expressions, which concrete function a variable holds, correctly accounting for reassignments, branches, loops, and short-circuit evaluation.
- **Threads captured values through** - a closure's captured data is passed as ordinary extra arguments to the specialized copy, and consumed closures are blanked out so the pass knows it's finished (this is what lets it converge).
- **Resolves indirect sources** - functions coming from array indices, record fields, function return values, and partial applications are traced back to their concrete target.
- **Fixes the surrounding pipeline** - monomorphization now specializes generic closure targets, argument-promotion keeps closure-call signatures stable, and the interpreter-to-QIR path normalizes callable type signatures so entry-point function arguments compile cleanly to QIR and circuits.

The large test additions (Rust snapshot tests plus a new `test_callable_passing.py`) lock in every scenario above: Python<=>Q# callables, closures with distinct captures, nested closures, and end-to-end QIR/circuit output.

### DIFF
--- a/source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
+++ b/source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
@@ -4527,6 +4527,22 @@ fn remove_callable_params(
     }
 }
 
+/// Removes a top-level callable parameter from the specialized callable's
+/// input pattern, collapsing a singleton tuple to a bare pattern and
+/// replacing a bare-bind pattern with `Unit` when the removed parameter was
+/// the only one.
+///
+/// Delegates to [`remove_nested_callable_param`] when `param.field_path` is
+/// non-empty (the callable argument is nested inside a tuple parameter).
+///
+/// # Before
+/// ```text
+/// input = (a, callable, b)   // top_level_param = 1
+/// ```
+/// # After
+/// ```text
+/// input = (a, b)             // callable removed, tuple shrunk
+/// ```
 fn remove_callable_param(
     package: &mut Package,
     decl: &mut CallableDecl,
@@ -5129,7 +5145,6 @@ fn extract_stmt(source: &Package, stmt_id: qsc_fir::fir::StmtId, target: &mut Pa
     }
 }
 
-#[allow(clippy::too_many_lines)]
 /// Recursively copies an expression and its transitive references into the
 /// extraction target.
 ///
@@ -5139,6 +5154,7 @@ fn extract_stmt(source: &Package, stmt_id: qsc_fir::fir::StmtId, target: &mut Pa
 /// carries a bare `LocalItemId` with no package qualifier, so the lambda it
 /// references must live in the same package as the closure expression. Named
 /// nested functions in `StmtKind::Item` are followed for the same reason.
+#[allow(clippy::too_many_lines)]
 fn extract_expr(source: &Package, expr_id: ExprId, target: &mut Package) {
     if target.exprs.contains_key(expr_id) {
         return;
--- a/source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
+++ b/source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
@@ -30,9 +30,11 @@ use super::{
     build_combined_spec_key, build_combined_spec_key_for_group, build_spec_key,
     has_multiple_forwarded_callable_arrays, is_combined_eligible, partition_mixed_branch_split,
 };
-use crate::EMPTY_EXEC_RANGE;
 use crate::cloner::FirCloner;
-use crate::fir_builder::functored_specs;
+use crate::fir_builder::{
+    alloc_bin_op_expr, alloc_call_expr, alloc_expr, alloc_functor_wrapped_expr, alloc_int_lit,
+    alloc_item_var_expr, alloc_local_var_expr, functored_specs,
+};
 use crate::package_assigners::PackageAssigners;
 use crate::walk_utils::for_each_expr_in_callable_impl;
 use qsc_data_structures::functors::FunctorApp;
@@ -40,9 +42,8 @@ use qsc_data_structures::span::Span;
 use qsc_fir::assigner::Assigner;
 use qsc_fir::fir::{
     BinOp, Block, BlockId, CallableDecl, CallableImpl, Expr, ExprId, ExprKind, Field, FieldPath,
-    Functor, Ident, Item, ItemId, ItemKind, Lit, LocalItemId, LocalVarId, Package, PackageId,
-    PackageLookup, PackageStore, Pat, PatId, PatKind, Res, Stmt, StmtId, StoreItemId, UnOp,
-    Visibility,
+    Ident, Item, ItemId, ItemKind, LocalItemId, LocalVarId, Package, PackageId, PackageLookup,
+    PackageStore, Pat, PatId, PatKind, Res, Stmt, StmtId, StoreItemId, Visibility,
 };
 use qsc_fir::ty::{Arrow, Prim, Ty};
 use qsc_fir::visit::{self, Visitor};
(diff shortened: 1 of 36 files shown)


source/compiler/qsc_openqasm_compiler/src/tests/statement/gate_call.rs
```diff
@@ -445,6 +445,25 @@ fn implicit_cast_to_angle_works() -> miette::Result<(), Vec<Report>> {
...
+        gate phase_by(theta) q {
+            rz(theta) q;
+        }
+        qubit q;
+        output bit c;
+        phase_by(1.0) q;
+        c = measure q;
+    "#;
+
+    let _ = compile_qasm_to_qir(source)?;
```
swernli: might as well turn this one into a full expect test

source/compiler/qsc_fir_transforms/src/fir_builder.rs
```diff
@@ -100,6 +101,128 @@ pub(crate) fn alloc_field_expr(
...
+            ty.clone(),
+            ExprKind::UnOp(UnOp::Functor(Functor::Adj), current_id),
+            span,
+        );
+    }
+    for _ in 0..functor.controlled {
+        current_id = alloc_expr(
+            package,
+            assigner,
+            ty.clone(),
```
swernli: Don't successive controlled functor applications generate an expression with different types? Depending on what value `ty` started from, this seems like it either has too many or not enough tuples of qubit arrays wrapping it.

idavis: Great catch

source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
```diff
@@ -99,16 +133,327 @@ pub(super) fn specialize(
...
+        return false;
+    };
+    let hof_item_id = group[0].hof_item_id;
+    let hof_store_id = StoreItemId::from((hof_item_id.package, hof_item_id.item));
+    if !store
+        .get(hof_store_id.package)
+        .items
+        .contains_key(hof_store_id.item)
+    {
+        return true;
```
swernli: this seems like a guard against trying to process an item that can't be found in the expected package in the package store, but should that either return `false` or panic? It seems like a missing HOF item is a terminal case, and returning true will just leave behind the orphaned HOF. Perhaps the entire guard can be removed in favor of allowing a later failure if the item is queried from the store after removal.

source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
```diff
@@ -99,16 +133,327 @@ pub(super) fn specialize(
...
+        let spec_key = build_combined_spec_key(hof_item_id, &members_cs);
+        if dedup.contains_key(&spec_key) {
+            continue;
+        }
+        if recursion_guard.contains(&spec_key) {
+            let package = store.get(candidate.call_pkg_id);
+            let span = package.get_expr(candidate.call_expr_id).span;
+            errors.push(Error::RecursiveSpecialization(span));
+            continue;
+        }
```
swernli: Not critical for right now, but I think this recursion guard has become extraneous here (and in the other functions it's passed into). There appears to be no condition under which these calls recur, and every entry is removed before individual functions return or continue their loops. There is one test for it, but I missed in an early review that the test simply verifies the shape of the error and does not have any Q# code that produces it. Perhaps worth a follow up issue to remove it.

source/compiler/qsc_fir_transforms/src/defunctionalize/specialize.rs
```diff
@@ -521,659 +871,2323 @@ fn specialize_one(
...
+/// a [`CapturedCallableSpecialization`] recording its index, type, and resolved
+/// concrete callable. Captures that are not statically-known callables are
+/// omitted so they stay threaded as ordinary runtime operands.
+fn captured_callable_specializations(
+    package: &Package,
+    captures: &[CapturedVar],
+) -> Vec<CapturedCallableSpecialization> {
+    if captures.len() != 1 {
+        return Vec::new();
+    }
```
swernli, line 1518: based on this, it looks like the function will always return either an empty list or a list of one item. Maybe this should return `Option<CapturedCallableSpecialization>` instead of a vec? The callers then iterate over an enumeration of the returned vector, but that too would only be "(0, val)" so perhaps the enumeration itself is leftover from a prior approach.

swernli (approved):

PR #3413, opened 2026-06-30
Allow adaptive `NeutralAtomDevice` simulation

orpuente-MS:


source/qdk_package/tests-integration/devices/validation/__init__.py
```diff
@@ -94,3 +94,41 @@ def check_qubit_ordering_unchanged(
...
+    """
+    Validate that, in every block, all phi nodes appear before any non-phi
+    instruction.
+
+    LLVM requires phi nodes to be at the very start of a basic block. The
+    reorder pass rewrites instruction order within each block, so it must
+    preserve this invariant for programs with classical control flow.
+    """
+    for function in module.functions:
+        if not is_entry_point(function):
```
billti: If we're adding function calling, why skip this check for all functions except the entry point?

billti (approved):

PR #3417, opened 2026-07-02
Support classically controlled gates in Stim

joao-boechat:
# Support classically-controlled gates (`rec` targets) in the Stim → QIR compiler

## What this is (in Stim)

Stim lets certain two-qubit controlled gates take a **measurement record** target
(`rec[-k]`) in place of a qubit control. This turns the gate into a
**classically-controlled** (feed-forward) operation: *if the recorded measurement
bit is set, apply the target Pauli to the target qubit.*

```
M 0
CX rec[-1] 1    # if the last measurement was 1, apply X to qubit 1
CZ 0 rec[-1]    # CZ is symmetric, so the rec may go on either side
XCZ 1 rec[-1]   # reversed order: rec on the 2nd target, X applied to qubit 1
```

A `rec` is only valid where the gate has a **Z-basis control**, which limits it to:

| Gate (+ aliases)         | `rec` allowed on | Classical action |
| ------------------------ | ---------------- | ---------------- |
| `CX` / `CNOT` / `ZCX`    | 1st (control)    | `X` on target    |
| `CY` / `ZCY`             | 1st (control)    | `Y` on target    |
| `CZ` / `ZCZ`             | either (symmetric) | `Z` on the other qubit |
| `XCZ`                    | 2nd (the Z side) | `X` on the 1st qubit |
| `YCZ`                    | 2nd (the Z side) | `Y` on the 1st qubit |

Docs: [Stim gate reference](https://github.com/quantumlib/Stim/blob/main/doc/gates.md)
(see the [`CX`](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#CX),
[`CZ`](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#CZ), and
[`XCZ`](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#XCZ) entries).

## What changed

Previously these gates went through `broadcast_pair`, which assumes both targets
are qubits and rejects a `rec` as an unsupported target. The changes:

- **`broadcast_controlled`** — a new dispatch for controlled gates. For each target
  pair it either emits the fully-quantum decomposition (both qubits) or, when a
  target is a `rec`, emits a classically-controlled operation.
- **`classical_control`** — emits the feed-forward pattern (reusing the existing
  `REQUIRE` machinery): read the measurement result, branch, apply the gate's
  single-qubit Pauli to the target, then rejoin.
- **`AllowedRecPosition` (`First` / `Second` / `Either`)** — encodes which side of a
  gate may carry a `rec`, so `CX` accepts it only on the control, `XCZ` only on the
  second target, and `CZ` on either.
- **Call sites** — `CX`/`CY`/`CZ`/`XCZ`/`YCZ` now call `broadcast_controlled`, each
  supplying its quantum decomposition and its single-qubit classical action.
- **Dedicated diagnostics** for the invalid cases, instead of a generic
  "unsupported target":
  - `MisplacedMeasurementRecord` — `rec` on a side the gate doesn't allow.
  - `NegatedMeasurementRecord` — a negated control (`!rec[-1]`).
  - `MeasurementRecordWithoutQubit` — both targets are records (no qubit to act on).
- **Tests** — `qir/tests/measurement_record_targets.rs` covers every gate and every
  error case.


source/compiler/stim_compiler/src/qir/tests/measurement_record_targets.rs
```diff
@@ -0,0 +1,934 @@
...
+          %r_0 = call i1 @__quantum__rt__read_result(ptr inttoptr (i64 0 to ptr))
+          br i1 %r_0, label %apply_controlled_0, label %continue_controlled_0
+        apply_controlled_0:
+          call void @__quantum__qis__x__body(ptr inttoptr (i64 1 to ptr))
+          br label %continue_controlled_0
+        continue_controlled_0:
+          call void @__quantum__rt__array_record_output(i64 1, ptr null)
+          call void @__quantum__rt__result_record_output(ptr inttoptr (i64 0 to ptr), ptr null)
+          ret i64 0
+        }
```
orpuente-MS: We could use IR-defined functions as in the following example to make the generated QIR shorter:

```llvm
define i64 @ENTRYPOINT__main() #0 {
block_0:
  call void @__quantum__rt__initialize(ptr null)

  ; Measure 0-th qubit into 0-th result
  call void @__quantum__qis__h__body(ptr inttoptr (i64 0 to ptr))
  call void @__quantum__qis__mresetz__body(ptr inttoptr (i64 0 to ptr), ptr inttoptr (i64 0 to ptr))

  ; Call our helper function
  call void @classical_control_cx(ptr inttoptr (i64 0 to ptr), ptr inttoptr (i64 1 to ptr))
  
  ; Measure 2nd qubit
  call void @__quantum__qis__mresetz__body(ptr inttoptr (i64 1 to ptr), ptr inttoptr (i64 1 to ptr))
  
  ; Record outputs & resturn
  call void @__quantum__rt__array_record_output(i64 2, ptr null)
  call void @__quantum__rt__result_record_output(ptr inttoptr (i64 0 to ptr), ptr null)
  call void @__quantum__rt__result_record_output(ptr inttoptr (i64 1 to ptr), ptr null)
  ret i64 0
}

; Helper classical controlled cx function.
;
define void @classical_control_cx(ptr %result, ptr %qubit) {
block_cx_entry:
  %result_val = call i1 @__quantum__rt__read_result(ptr %result)
  br i1 %result_val, label %block_cx_x, label %block_cx_ret
block_cx_x:
  call void @__quantum__qis__x__body(ptr %qubit)
  br label %block_cx_ret
block_cx_ret:
  ret void
}
```

joao-boechat: I was thinking of adding this in a future PR, together with other optimizations such as avoiding the repetition of noise intrinsics. It will be a rather substantial change, especially when it comes to the tests we already have.

orpuente-MS (approved):

PR #3418, opened 2026-07-05
Update codeowners

idavis:


.github/CODEOWNERS
```diff
@@ -1,34 +1,34 @@
-* @billti @idavis @minestarks @swernli
+- @billti @idavis @amcasey @swernli
```
amcasey: What does the change from `*` to `-` mean?

idavis: Good catch, fixed

billti (approved):

idavis:

PR #3420, opened 2026-07-06
Standardize error codes with "Qdk." prefix

joao-boechat:


ScottCarda-MS (approved):
LGTM thanks for getting this in!

PR #3421, opened 2026-07-06
[QRE] Build trace from Q# program

fedimser:
Intrinsics not handled in new backend TraceBuilder:
 * BeginEstimateCaching/EndEstimateCaching - no-op (which is correct) because there is not significant speedup from caching trace fragment (copying it has the same complexity as genrating it).
 * Load/Store - supported partially. They emit read/write gate, but do not affect memory/compute qubit counts. This needs to be fixed when we finalize how we treat memory/compute architrcure.
 * `EnableMemoryComputeArchitecture` - no-op until we finalize design for memory/compute architecture. To properly support this, we need to maintain untyped->typed qubit id mapping (which we don't do in counts.rs).
 * `AccountForEstimatesInternal` - rejected (error) because trace doesn't support gates with explicit resource counts.


source/qre/src/trace_builder.rs
```diff
@@ -0,0 +1,423 @@
...
+                } else {
+                    self.push_operation(PendingOperation::Block {
+                        repetitions: frame.repetitions,
+                        operations: frame.operations,
+                    });
+                }
+
+                Some(Ok(Value::unit()))
+            }
+            "BeginEstimateCaching" => Some(Ok(Value::Bool(true))),
```
msoeken, line 352: Are there plans to support this later?

source/qre/src/trace_builder.rs
```diff
@@ -0,0 +1,423 @@
...
+                let [result, qubit] = std::array::from_fn(|i| values[i].clone());
+                let Value::Result(BackendResult::Val(val)) = result else {
+                    panic!("first argument to PostSelectZ should be a measurement result");
+                };
+                let qubit = qubit.unwrap_qubit().deref().0;
+                self.post_select_measurements.insert(qubit, val);
+
+                Some(Ok(Value::unit()))
+            }
+            _ => None,
```
msoeken, line 383: We should consider a way to reflect custom Q# intrinsics in the trace. This can be done in a separate PR though. The idea is that the user can define a custom intrinsic in Q# (e.g., a logical instruction supported by a specific QEC code) and then have it appear in the QRE trace.

msoeken (commented):
Looks great! I left some initial comments for now.

source/qre/src/trace_builder.rs
fedimser: I didn't want to, because unlike with counts, caching trace fragment will not speed up resource estimaiton because size of trace is asymptotically the same as size of circuit. Am I wrong?

source/qre/src/trace_builder.rs
fedimser: Will definitely do that, but need to think how to represent it, maybe not literally as Q# intrinsic?

fedimser:
It is ready for review, but moved to draft because it's not going in July release.

source/qre/src/trace_builder.rs
msoeken: Yes, let's think this through and do it as separate work

msoeken (approved):

PR #3422, opened 2026-07-07
Add ErrorComposition class and integrate into estimation processes

msoeken:
- Introduced ErrorComposition to control error contribution methods (UnionBound and Product).
- Updated estimation functions to accept ErrorComposition as a parameter.
- Enhanced EstimationResult to manage error composition mode.
- Added tests to validate behavior of error composition modes.


source/qdk_package/src/qre.rs
```diff
@@ -1023,6 +1024,63 @@ impl EstimationCollectionIterator {
...
+///   grows increasingly loose as the individual errors grow, because it
+///   overcounts the overlap between simultaneous failures.
+///
+/// - ``Product`` computes ``1 - prod(1 - p_i)``, i.e. one minus the probability
+///   that no operation fails, assuming the failures are independent.
+///
+///   Advantages: always stays in ``[0, 1)``; it is tight (in fact exact) when
+///   the errors are independent.
+///
+///   Disadvantages: it relies on independence, so it can underestimate when
```
brad-lackey, line 1052: This technique also suffers from finite precision issues when performing the computation on many small probabilities.

brad-lackey (approved):

source/qdk_package/src/qre.rs
msoeken: Addressed in 40ba11d

PR #3424, opened 2026-07-07
FIR transform side effect updates

idavis:
This branch refines how the Q# compiler cleans up code during its internal optimization passes, so the programs it generates are leaner without changing what they do.

Two related improvements:

1. Stop unnecessarily pulling conditions out of if-style checks. Previously, when the compiler processed a branch, it would always move the condition (the true/false test) into its own temporary step before deciding which way to go. That's only necessary when evaluating the condition could cause something to happen. This change teaches the compiler to recognize when a condition is harmless to leave in place, so it no longer creates pointless temporary steps for those cases. The result is simpler, cleaner intermediate code.
2. Smarter removal of unused local variables. The compiler drops local variables that are never read. Deciding whether that's safe used to rely on a check called "is this expression free of side effects." This branch replaces that with a more precise check "is this expression safe to throw away", which correctly handles more situations, such as array-index lookups and calls to functions that are known to be pure. So the compiler can now safely remove more genuinely-unused code that it previously kept around out of caution, while still never removing anything that matters.

The rest of the changes are updates to the compiler's test snapshots that reflect this new, tidier output.

### DIFF
--- a/source/compiler/qsc_fir_transforms/src/defunctionalize/tests/analysis.rs
+++ b/source/compiler/qsc_fir_transforms/src/defunctionalize/tests/analysis.rs
@@ -6052,13 +6052,12 @@ fn operand_block_tuple_pattern_dispatch_resolves_field_path() {
     );
 }
 
-/// When the index selecting a callable from an array is a non-trivial
-/// expression (here `i + 1`), the synthesized dispatch hoists it into a single
-/// shared `let` so it is evaluated once rather than re-evaluated in every
-/// `index == k` guard. Contrast `operand_block_tuple_pattern_dispatch_resolves_field_path`,
-/// where a bare-variable index keeps referencing the original expression.
+/// When the index selecting a callable from an array is pure arithmetic, the
+/// synthesized dispatch can reuse the original expression in each guard. This
+/// keeps "non-trivial" from being mistaken for "effectful"; the separate
+/// side-effecting index tests cover the single-evaluation hoist path.
 #[test]
-fn non_trivial_array_index_dispatch_hoists_index_once() {
+fn pure_arithmetic_array_index_dispatch_reuses_index_expression() {
     let source = r#"
         operation Main() : Unit {
             let ops = [I, X, Y];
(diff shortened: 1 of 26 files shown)


source/compiler/qsc_fir_transforms/src/walk_utils.rs
```diff
@@ -812,3 +894,221 @@ pub(crate) fn expr_is_side_effect_free(package: &Package, expr_id: ExprId) -> bo
...
+    active_callables: &mut FxHashSet<LocalItemId>,
+) -> bool {
+    if decl.kind != CallableKind::Function {
+        return false;
+    }
+
+    let CallableImpl::Spec(spec_impl) = &decl.implementation else {
+        return false;
+    };
+
```
swernli, line 1060: The comments align with the behavior here, but I'm not sure I understand why an intrinsic function can't be trusted. I would've assumed an intrinsic function should return `true` but perhaps I'm misunderstanding the intent here.

source/compiler/qsc_fir_transforms/src/walk_utils/tests.rs
```diff
@@ -699,7 +706,106 @@ fn given_call_is_not_side_effect_free() {
...
+
+    assert!(expr_is_side_effect_free(package, pkg_id, call));
+    assert!(!expr_is_safe_to_discard(package, pkg_id, call));
+}
+
+#[test]
+fn given_function_call_that_calls_message_is_not_side_effect_free() {
+    let (store, pkg_id) = compile_to_fir(
+        "function Noisy(x : Int) : Int {
+             Message(\"x\");
```
swernli, line 799: relative to my other comment, not sure this should cause the checks to return false... it seems like it could be safe to treat it as side-effect free and safe to discard in the same way mutation of a local variable is safe.

swernli: Oh, I think I understand... it's because they may fail at runtime. Never mind, I'll resolve my comments.

source/compiler/qsc_fir_transforms/src/defunctionalize/tests/analysis.rs
```diff
@@ -6128,6 +6126,96 @@ fn non_trivial_array_index_dispatch_hoists_index_once() {
...
+/// expression keeps index-dispatch cleanup aligned with the shared purity
+/// predicate.
+#[test]
+fn pure_array_index_dispatch_reuses_index_expression() {
+    let source = r#"
+        operation Main() : Unit {
+            let ops = [I, X, Y];
+            for i in 0..1 {
+                use q = Qubit();
+                let op = ops[{ i }];
```
swernli, line 6140: could you include another test (if it doesn't already exist) that shows that an impure index expression gets the old behavior of having it bound to a temporary variable? May make it easier for future readers to understand intent when the different behaviors are captured explicitly.

swernli: ah, this is sort of covered by `indexed_callable_array_param_hoists_side_effecting_index_once` in the specialization.rs file... though I still think it would be useful to have an inline example here as well. Something that does `let op = ops[{X(q);i}];`

swernli (approved):

PR #3425, opened 2026-07-08
Circuit-Editor and Visualization Re-Architecting and Tests

ScottCarda-MS:
# Circuit Editor Re-Architecture: Layered Foundation (Data → Actions → Renderer → Editor)

> **Reviewer TL;DR.** This is an **internal** re-architecture of the
> `circuit-vis` circuit editor. It replaces the four monolithic editor files
> with a layered `data / actions / renderer / editor` structure and includes
> view-state persistence and the measurement-dependency confirm UX. **No public
> API changes and no user-facing behavior regressions.** It is a single cohesive
> commit because the core is a mutually-dependent refactor that cannot compile in
> smaller pieces; see the **How to review this efficiently** section below.

---

## At a glance

|                     |                                                                             |
| ------------------- | --------------------------------------------------------------------------- |
| **Scope**           | `source/npm/qsharp/ux/circuit-vis/**` + `source/npm/qsharp/test/**` only    |
| **Size**            | 114 files, ~+25,500 / −3,956                                                |
| **Language**        | TypeScript (frontend) + `.mjs` tests. **No Rust, no VS Code host changes.** |
| **Public API**      | Unchanged — `index.ts` exports the same surface                             |
| **Tests**           | 392 circuit-editor tests across 28 files, all green; plus snapshot fixtures |
| **Behavioral risk** | Low for existing flows; new capability is additive                          |

---

## What this PR does

1. **Splits the monoliths into a layered architecture.** The old editor was
   four large files with tangled responsibilities. They are removed and
   replaced with a strict, one-directional layering:

   ```
   data/  →  actions/  →  renderer/  →  editor/  →  sqore.ts  →  state-viz/
   (pure)    (pure)       (read data)   (DOM glue)  (entrypoint) (parallel)
   ```

   The hard rule: **`data/` and `actions/` never touch the DOM.** The renderer
   reads data to produce SVG; the editor glues DOM events to action calls.

2. **Adds group-aware editing foundations.** Editing _inside_ expanded groups,
   the group structure and add/remove/ancestor-refresh/collision-split
   behaviors, single-leg move and control-drag paths, and the dropzone geometry
   the editor is built on.

3. **Hardens classical-register / measurement integrity.** Moving or deleting a
   measurement that later gates depend on now cascades correctly (with a confirm
   prompt) instead of crashing or leaving dangling classical references.

4. **Persists view state.** Group expand/collapse and related view state survive
   re-render via the `ViewState` type threaded through `sqore.ts`.

5. **Adds an in-tree architecture guide.** A new
   [`ux/circuit-vis/ARCHITECTURE.md`](../source/npm/qsharp/ux/circuit-vis/ARCHITECTURE.md)
   documents the layering, the render/click flows, and the module invariants —
   useful for this review and as durable documentation afterward.

### Files removed (old monoliths)

- `circuitManipulation.ts`
- `events.ts`
- `panel.ts`
- `draggable.ts` (old version)
- `contextMenu.ts` (old version)

Their responsibilities now live in the layered modules.

---

## How to review this efficiently

This PR adds a structural walkthrough alongside the code:
[`ux/circuit-vis/ARCHITECTURE.md`](../source/npm/qsharp/ux/circuit-vis/ARCHITECTURE.md).
Read its TL;DR and module map first — it explains the layering, has diagrams of
a render and a click flow, and lists the invariants. Then follow the order
below.

Suggested order and natural stopping points:

1. **Data layer** (`data/`) — small, pure value types (`Location`,
   `CircuitModel`, `ViewState`). The vocabulary everything else is written in.
2. **Action layer** (`actions/` + `actions/circuit-actions/`) — pure mutations;
   the correctness core. Cross-reference the
   `test/circuit-editor/circuit-actions/` suite as you go.
   **→ Good checkpoint: you've now seen the entire pure core.**
3. **Renderer** (`renderer/`) — `Circuit` → SVG; reads data, never mutates.
4. **Editor** (`editor/` + `editor/controllers/`) — DOM glue. Read the shared
   `InteractionContext` first, then the controllers.
   **→ Good checkpoint: trace one end-to-end flow from ARCHITECTURE.md.**
5. **Entrypoint** (`sqore.ts`, `utils.ts`, `index.ts`) — ties it together.
6. **state-viz** (`state-viz/`) — parallel subsystem, only import-path updates
   here; safe to skim last.

### What to focus your scrutiny on

You don't need prior familiarity with this code to review it well. If you only
have time for a few things, these are the questions worth asking — each one is
answerable by reading a file or two:

- **Does the split preserve behavior?** This is a refactor, not a rewrite of
  what the editor does. The tests are the contract: if the suite is green and
  the snapshot fixtures are unchanged, existing behavior is preserved. Skim the
  test names in `test/circuit-editor/` to see what's covered.
- **Are the layers actually kept separate?** The whole point of the split is a
  one-way dependency: the pure logic (`data/`, `actions/`) computes circuit
  changes, and only the `editor/` layer touches the browser/DOM. Spot-check a
  couple of `data/` and `actions/` files — they should have no DOM or
  browser-API calls. If they don't, the separation holds.
- **Is rendering read-only?** Drawing a circuit should never change it. Confirm
  the render path in `sqore.ts` works on a copy, so the caller's circuit object
  is left untouched.
- **Does the naming/structure make sense to you as a newcomer?** You are exactly
  the audience the new layout is meant to help. If a folder or file name doesn't
  match what's inside, that's useful feedback.

Everything else (cache-consistency details, controller wiring, drag cleanup) is
covered by the test suite.

### What you can safely skim

- `state-viz/` diffs — mechanical import-path updates from the move.
- Snapshot fixtures under `test/circuits-cases/` — generated.

---

## Testing

- **392 circuit-editor tests** across 28 `.mjs` files, all passing (`data`,
  `actions/circuit-actions`, `renderer`, controllers, `sqore`), plus snapshot
  fixtures under `test/circuits-cases/`.
- The action layer is the most heavily covered: the
  `test/circuit-editor/circuit-actions/` suite pins the single-leg move /
  control / classical-ref-remap / group-structure paths.
- Run locally from `source/npm/qsharp/`:

  ```pwsh
  # build the UX bundle
  npm run build:ux
  # run the circuit-editor suite
  node --test "test/circuit-editor/**/*.test.mjs"
  ```


source/npm/qsharp/test/circuit-editor/circuit-actions/addRemove.test.mjs
```diff
@@ -0,0 +1,511 @@
...
+
+  // Empty location parses to root; last() is null → failure, no change.
+  const result = addOperation(model, unitary("H"), "", 0);
+
+  assert.equal(result, null);
+  expectGrid(model, []);
+});
+
+// ---------------------------------------------------------------------------
+// removeOperation
```
amcasey, line 224: I may have overlooked tests for the following:
1. Removing the only (remaining) gate
2. Removing a gate at an invalid location

source/npm/qsharp/test/circuit-editor/circuit-actions/addRemove.test.mjs
```diff
@@ -0,0 +1,511 @@
...
+} from "../_helpers.mjs";
+
+/** Fresh empty circuit literal with `n` qubits and no operations. */
+const emptyCircuit = (/** @type {number} */ n) => circuit(n, []);
+
+/** Single-target unitary template on wire 0 (what `addOperation` copies). */
+const unitary = (/** @type {string} */ g) => gate(g, 0);
+
+// ---------------------------------------------------------------------------
+// addOperation
```
amcasey, line 36: What happens if I add an operation at `0,1000` or `1000,0`?

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAddRemove.test.mjs
```diff
@@ -0,0 +1,57 @@
...
+test("removeOperation strips a leaf inside an expanded group", () => {
+  const model = build(
+    circuit(2, [[group("Group", [[gate("H", 0), gate("X", 1)]])]]),
+  );
+
+  // Remove the nested X (group col 0, row 1).
+  removeOperation(model, "0,0-0,1");
+
+  // Outer group remains; the X inside its single child column is gone.
+  expectOp(at(model, "0,0"), { Group: { children: [[{ H: 0 }]] } });
```
amcasey, line 97: What happens if you remove the `H` too?

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAddRemove.test.mjs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// Remove-mutator tests on grouped shapes, driven through the public
```
amcasey: The file name seems to imply that there will also be add tests?

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAncestorRefresh.test.mjs
```diff
@@ -0,0 +1,343 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// Ancestor `.targets` cache refresh after a child mutation: when
```
amcasey: I had trouble following these tests (just my own ignorance) but I'm wondering if the following are covered (here or elsewhere):
1. Moving a gate into a parent group
2. Moving a gate into a grandparent group
3. Moving a gate into a child group
4. Moving a gate into a sibling group

Or possibly we don't allow any of those changes.

source/npm/qsharp/test/circuit-editor/circuit-actions/groupMove.test.mjs
```diff
@@ -0,0 +1,244 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// Group move tests: moving a child out of a group, dragging the
+// group as a rigid unit, classical-control anchoring on the
+// group's children, empty-group cleanup, trailing inner-column
+// dropzone, and quantum control-leg drags on multi-target gates.
+// Groups themselves carry classical controls only — the authoring
```
amcasey: Do we have tests verifying that "the authoring layer refuses quantum controls on groups"?

source/npm/qsharp/test/circuit-editor/circuit-actions/groupMove.test.mjs
```diff
@@ -0,0 +1,244 @@
...
+    circuit(2, [[group("Outer", [[group("Inner", [[gate("H", 0)]])]])]]),
+  );
+
+  // move the deepest leaf out to a new top-level column on q1
+  moveOperation(model, "0,0-0,0-0,0", "0,1", 0, 1, false, true);
+
+  expectGrid(model, [[{ H: 1 }]]);
+});
+
+test("moveOperation: cleanup STOPS at the first non-empty ancestor", () => {
```
amcasey, line 94: What if there is no non-empty ancestor?

source/npm/qsharp/test/circuit-editor/circuit-actions/measurementCascade.test.mjs
```diff
@@ -0,0 +1,437 @@
...
+  meas,
+} from "../_helpers.mjs";
+
+// Local shorthands over the shared helpers: these suites use the
+// "Measure" gate name (asserted in places) and classically-
+// controlled X consumers.
+const _mGate = (/** @type {number} */ q, /** @type {number} */ r) =>
+  meas(q, { gate: "Measure", result: r });
+
+const _ccx = (
```
amcasey, line 44: I would have thought other tests would need these too and they would end up in a utility file.

source/npm/qsharp/test/circuit-editor/circuit-actions/measurementCascade.test.mjs
```diff
@@ -0,0 +1,437 @@
...
+  // classical refs. Otherwise, after `_updateMeasurementLines`
+  // authoritatively renumbers result indices on the affected
+  // wire, walking those producer values back through the
+  // consumer remap can chain-react: each M's new result index
+  // happens to match another M's pre-move key, so `.results`
+  // gets remapped a second time — collapsing into duplicate
+  // result indices and orphaning consumers whose target M had
+  // its `.results` clobbered.
+  //
+  // Setup: three Ms with consumers spread across two wires.
```
amcasey: In some particularly complicated tests (or when many tests share a starting circuit), ascii art circuit diagrams might help.  Definitely not mandatory, since they come with a maintenance burden.

source/npm/qsharp/test/circuit-editor/circuit-actions/qubitOps.test.mjs
```diff
@@ -0,0 +1,360 @@
...
+  /** @type {number} */ targetQubit,
+  /** @type {number} */ ctrlQubit,
+  /** @type {number} */ ctrlResult,
+) => gate("X", targetQubit, { ctrls: [{ q: ctrlQubit, r: ctrlResult }] });
+
+// ---------------------------------------------------------------------------
+// moveQubit / removeQubit (flat-grid base cases)
+// ---------------------------------------------------------------------------
+
+test("moveQubit swaps register references and reorders ops within a column", () => {
```
amcasey, line 43: What happens if there's an empty wire in the middle and you move it to the end?

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
```diff
@@ -0,0 +1,218 @@
...
+  assert.equal(isValidAngleExpression("-3.5"), true);
+  assert.equal(isValidAngleExpression("5."), true);
+});
+
+test("isValidAngleExpression: bare π is valid in all four case forms", () => {
+  // The prompt's placeholder example is `"π / 2.0"`; the user
+  // can type π directly via the on-screen π button, or use any
+  // case variant of `pi` which `normalizeAngleExpression` folds
+  // to π before this check runs.
+  assert.equal(isValidAngleExpression("π"), true);
```
amcasey, line 42: Is there a way to assert that these are not only valid but equivalent?

amcasey: And what about Π? 😆

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
```diff
@@ -0,0 +1,218 @@
...
+  // through the parser; pin both.
+  assert.equal(isValidAngleExpression("-π"), true);
+  assert.equal(isValidAngleExpression("+π"), true);
+  assert.equal(isValidAngleExpression("-pi"), true);
+});
+
+test("isValidAngleExpression: arithmetic combinations are valid", () => {
+  // The four supported binary operators, with both numbers and π.
+  // These mirror the prompt's example text ("π / 2.0", "2.0 * π").
+  assert.equal(isValidAngleExpression("2 + 3"), true);
```
amcasey, line 58: When would you need addition?

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
```diff
@@ -0,0 +1,218 @@
...
+  assert.equal(isValidAngleExpression("6 / 2"), true);
+  assert.equal(isValidAngleExpression("π / 2"), true);
+  assert.equal(isValidAngleExpression("2 * π"), true);
+  assert.equal(isValidAngleExpression("2.0 * π"), true);
+});
+
+test("isValidAngleExpression: parentheses (including nesting) are valid", () => {
+  // Recursive descent through `parseFactor`'s `lpar` branch.
+  assert.equal(isValidAngleExpression("(π)"), true);
+  assert.equal(isValidAngleExpression("(2 + 3) * π"), true);
```
amcasey, line 70: I would find it reassuring to know that these expressions had been evaluated correctly.

source/npm/qsharp/test/circuit-editor/circuitModel.test.mjs
```diff
@@ -0,0 +1,377 @@
...
+
+  assert.equal(model.qubits.length, 3);
+  assert.deepEqual(model.qubitUseCounts, [0, 0, 0]);
+});
+
+// ---------------------------------------------------------------------------
+// removeTrailingUnusedQubits
+// ---------------------------------------------------------------------------
+
+test("removeTrailingUnusedQubits drops only zero-count tail wires", () => {
```
amcasey, line 167: Do we depend on trailing empty wires being removed?  If so, when does that cleanup happen?  Can someone supply a json file with trailing empty wires and, if so, does that cause problems?

source/npm/qsharp/test/circuit-editor/contextMenu.test.mjs
```diff
@@ -0,0 +1,446 @@
...
+  globalThis.SVGElement = jsdom.window.SVGElement;
+  globalThis.MouseEvent = jsdom.window.MouseEvent;
+});
+
+afterEach(() => {
+  jsdom?.window.close();
+  jsdom = null;
+});
+
+const SVG_NS = "http://www.w3.org/2000/svg";
```
amcasey, line 39: I thought copilot's suggestion to use an https link was sensible (assuming it actually works).

source/npm/qsharp/test/circuit-editor/contextMenu.test.mjs
```diff
@@ -0,0 +1,446 @@
...
+      ],
+    ]),
+  );
+  const { stub } = makeStubEvents(model);
+  const { host } = buildGateFixture("0,0", "body");
+  addContextMenuToHostElem(/** @type {any} */ (stub), host);
+
+  rightClick(host);
+
+  assert.deepEqual(getMenuLabels(), [
```
amcasey, line 359: Does order matter and, if so, is that something we care about?

source/npm/qsharp/test/circuit-editor/draggable.test.mjs
```diff
@@ -0,0 +1,275 @@
...
+afterEach(() => {
+  jsdom?.window.close();
+  jsdom = null;
+});
+
+// Geometry constants — kept in sync with `renderer/constants.ts` and
+// the private constants in `draggable.ts`. Locking these in test
+// fixtures makes the assertions self-documenting and catches the
+// "someone tweaked a padding constant and didn't realize the editor
+// math depended on it" regression.
```
amcasey: Consider adding a note about what to do if you actually do need to change the padding.

source/npm/qsharp/test/circuit-editor/draggable.test.mjs
```diff
@@ -0,0 +1,275 @@
...
+  const scope = makeScope([100], [60]);
+  const wireData = [200];
+
+  const dz = makeDropzoneBox(
+    { scope, wireData },
+    { colIndex: 0, opIndex: 0, wireIndex: 0, interColumn: true },
+  );
+
+  assert.equal(dz.getAttribute("class"), "dropzone");
+  // Left edge = colStartX - INTER_COLUMN_HALF_WIDTH - gatePadding
```
amcasey, line 119: This looks like it will be a nightmare to update if the spacing ever changes.  Maybe the point of the test is to assert that the spacing can never ever change?

source/npm/qsharp/test/circuit-editor/dropzones.test.mjs
```diff
@@ -0,0 +1,911 @@
...
+}
+
+/**
+ * Parse a hierarchical location string into its scope prefix and the
+ * (colIndex, opIndex) inside that scope.
+ *
+ *   "0,0"     -> { prefix: "",     colIndex: 0, opIndex: 0 }
+ *   "0,0-1,2" -> { prefix: "0,0",  colIndex: 1, opIndex: 2 }
+ */
+function parseLocation(/** @type {string} */ loc) {
```
amcasey: There isn't product code for this?

source/npm/qsharp/test/circuit-editor/findOperation.test.mjs
```diff
@@ -0,0 +1,168 @@
...
+//
+// These are the location-walking helpers in `ux/circuit-vis/utils.ts`.
+// They share a single private bounds-checked walker, so the tests focus
+// on the contract every public helper must honor:
+//
+//   - null / empty location → null
+//   - in-bounds location → the addressed thing
+//   - out-of-bounds location (top-level or nested) → null, never throw
+//
+// "Out of bounds" matters in practice because event handlers read
```
amcasey: I don't know whether the functionality is available in this layer, but it might be interesting to test the case described here: a _formerly_ valid location.

source/npm/qsharp/test/circuit-editor/prompts.test.mjs
```diff
@@ -0,0 +1,612 @@
...
+  const render = makeRenderSpy();
+
+  // Move M from (0,0) on wire 0 → target column 1 on wire 0 (no
+  // wire change). Consumer at "1,0" is invalidated; consumer at
+  // "2,0" survives.
+  moveWithConfirm(model, { from: "0,0", to: "1,0" }, render.fn);
+
+  const prompt = getOpenPrompt();
+  assert.ok(prompt);
+  prompt.okButton.click();
```
amcasey, line 562: Is there coverage of what happens when cancel is clicked?

source/npm/qsharp/test/circuit-editor/scrollController.test.mjs
```diff
@@ -0,0 +1,178 @@
...
+
+test("mousemove near the bottom edge scrolls down", () => {
+  arm();
+
+  move(200, 280); // within 50px of bottom (300)
+
+  assert.equal(el().scrollTop, 10);
+});
+
+test("mousemove near the left edge scrolls left", () => {
```
amcasey, line 123: Do we have some reason to believe the left and right edges might have different scrolling behavior?  Some of these tests seem redundant.

source/npm/qsharp/test/circuit-editor/selectionController.test.mjs
```diff
@@ -0,0 +1,376 @@
...
+  const { container, controlDot } = buildFixture();
+  const { interaction } = makeController(container);
+
+  dispatchMouseDown(controlDot);
+
+  assert.equal(interaction.selectedWire, 1);
+  assert.equal(interaction.movingControl, true);
+});
+
+test("mousedown with non-primary button is ignored", () => {
```
amcasey: I'm glad this is true, but I would probably have just put a comment in the product code near the handler that specifically doesn't do anything for other buttons.  This doesn't seem likely to break and, if it did, it would more or less have to be intentional, so the fix would be to delete this test.

source/npm/qsharp/test/circuit-editor/toolbox.test.mjs
```diff
@@ -0,0 +1,129 @@
...
+    toolbox.querySelector(".toolbox-panel-svg"),
+    "toolbox SVG should still be present",
+  );
+});
+
+test("toolbox with runCallback renders exactly one Run button", () => {
+  const toolbox = createToolboxElement(() => {});
+
+  const buttons = toolbox.querySelectorAll(".svg-run-button");
+  assert.equal(buttons.length, 1, "exactly one Run button expected");
```
amcasey, line 96: Is there product code that could generate more than one button and then simplifies away the redundant ones?  Why are we concerned that there might be too many Run buttons?

source/npm/qsharp/ux/circuit-vis/ARCHITECTURE.md
```diff
@@ -0,0 +1,507 @@
...
+- `sqore.ts` — entrypoint. `new Sqore(...).draw()` renders, and on
+  every render hands control to either the read-only path or
+  `installEditor`.
+- `utils.ts` — small shared helpers (location-walking
+  `findOperation` family, register helpers, DOM lookups).
+- `index.ts` — public API barrel.
+- `angleExpression.ts` — parse/normalize/evaluate gate rotation
+  angle expressions (handles `pi`/`π` and simple arithmetic).
+
+The hard rule that drives the layering: **`data/` and `actions/`
```
amcasey: Is there a way we can enforce that they don't touch the dom?

amcasey (commented):
As discussed, I focused my review on the tests, since I understand the rest is basically just shuffling code around.

I was a little surprised not to see more validation of our handling of bad input, given that users and copilot can directly edit qsc files.

source/npm/qsharp/ux/circuit-vis/editor/draggable.ts
```diff
@@ -0,0 +1,803 @@
...
+ */
+const createDropzones = (
+  container: HTMLElement,
+  sqore: Sqore,
+  layoutMap: LayoutMap,
+): SVGGElement => {
+  const svg = container.querySelector("svg.qviz") as SVGElement;
+
+  const overlay = document.createElementNS(
+    "http://www.w3.org/2000/svg",
```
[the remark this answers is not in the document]
fedimser: Can you please replace `http` with `https` (here and in other places) in links so security bot stops complaining?
When I navigate to the HTTP link, it redirects me to HTTPS link, so that should work.

source/npm/qsharp/ux/circuit-vis/sqore.ts
fedimser, line 84: Please include in this doc explanation why it's called "Sqore".

is this an abbreviation or pun on "score"? If I don't know the origin of a term, I will be wondering about it every time I see word "sqore" which is distracting.

fedimser:
Here are some findings from AI-assisted review.

1. Bug: View-state can leak across unrelated circuit replacements.
The new in-place update path always reuses the same Sqore instance in circuit.tsx:84. Sqore explicitly says a clean draw should be used when opening a different circuit in sqore.ts:168, but the component never does that after first mount. This can carry expand/collapse preferences to a different circuit if locations happen to match.

2. Bug: Run button presence is fixed on first shell mount. Toolbox creation is callback-dependent in toolbox.ts:56, while shell mounting is one-time in shell.ts:72. If run callback availability changes later, the button will not appear/disappear accordingly because the panel is not rebuilt.

3. Inconsistency between ARCHITECTURE and implementation: Layering rule is documented as strict, but dependency boundaries are not actually strict.
Architecture claims no outgoing dependencies from data/actions in ARCHITECTURE.md:30 and ARCHITECTURE.md:60, yet action/data import helpers from utils.ts:4, and that file imports renderer types/constants and DOM-query helpers.

4. There are behavior changes:
    * New public API on draw result: updateCircuit in index.ts:30.
    * In-place circuit replacement flow in UI component (preserves session view state, avoids rendering flicker) in circuit.tsx:79.
    * Measurement move/delete now prompts and cascades dependent-consumer handling in prompts.ts:113 and prompts.ts:162.
    * Group/mixed-control rendering behavior changed (quantum controls on classically controlled bodies/groups) in gateFormatter.ts:755.
    * Toolbox Run button behavior changed to optional rendering (not hidden-by-default) in toolbox.ts:53.
    * Utility lookup behavior now explicitly returns null for out-of-bounds locations (instead of throwing in those paths) in utils.ts:412.

5. These new features can be (in principle) split up in a separate PR:
    * updateCircuit API + in-place update semantics (index.ts:30, circuit.tsx:79).
    * ViewState persistence/rebase system (sqore.ts:101, viewState.ts:45).
    * Measurement-dependent confirmation/cascade UX (prompts.ts:113).
    * New group-control rendering geometry fixes (process.ts:363, gateFormatter.ts:777).

6. Tests that can be added:
    * Add focused updateCircuit lifecycle tests to define and lock behavior for view-state persistence vs reset (same-circuit updates, different-circuit replacement, and handler/options changes).
    * Add editor-shell reconfiguration tests to verify Run button behavior when runCallback availability changes across rerenders.
    * Add action-layer negative tests for malformed/out-of-range/stale locations to ensure add/remove/move are safe no-ops (no throws, no unintended mutation).
    * Add one end-to-end editor smoke test that chains drag, context-menu control add/remove, measurement-with-dependents move/delete confirmation flow, and rerender/update cycle to catch cross-layer regressions.

For 4-5, I asked Copilot to find what this PR does beyond pure refactoring. Ideally, I'd suggest having a PR doing pure refactoring, and then add new features in a follow-up PR(s), to minimize risk of bugs in the large refactoring PR. I am not asking to do the split, just review these items and make sure these do not introduce bugs.

For 6, I asked Copilot to suggest what tests could be added, so feel free to ignore it.

fedimser (changes requested):

source/npm/qsharp/ux/circuit-vis/actions/circuit-actions/classicalRefs.ts
```diff
@@ -0,0 +1,298 @@
...
+    const postKey = keyRemap.get(preKey);
+    if (postKey == null) return;
+    const colonIdx = postKey.indexOf(":");
+    reg.qubit = parseInt(postKey.substring(0, colonIdx), 10);
+    reg.result = parseInt(postKey.substring(colonIdx + 1), 10);
+  };
+  const walk = (g: ComponentGrid): void => {
+    for (const col of g) {
+      for (const op of col.components) {
+        // Walk the op's CONSUMER-side register fields only. A
```
fedimser: Can this comment be shorter?
Also, does it have to be  wrapped at 65 characters?

fedimser (changes requested):

fedimser (approved):

source/npm/qsharp/ux/circuit-vis/sqore.ts
ScottCarda-MS: I think it is better to just explain what I can here instead of clutter the doc with historical context. "Sqore" was the original name of the circuit visualization library when we inherited it. This used to be a separate tool developed by another team under our org. I think it is a reference to musical scores but used a "q" because that's common in quantum to insert "q" into the name of things.
I'm not opposed to changing the name, but nothing specific comes to mind, and I thought it was harmless for now to keep it the original name.

source/npm/qsharp/ux/circuit-vis/actions/circuit-actions/classicalRefs.ts
ScottCarda-MS: I condensed this comment and also took a sweep across the files to re-wrap comments at 100 instead of earlier cols.

source/npm/qsharp/ux/circuit-vis/editor/draggable.ts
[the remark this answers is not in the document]
ScottCarda-MS: Unfortunately this does not work. I am not sure why the API requires http over https, but it does. I will have to manually dismiss these alerts.

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAncestorRefresh.test.mjs
ScottCarda-MS: 1. Moving a gate "into" its parent group is just moving the gate inside the group it already is in. Here is an example test: [moveOperation: moving an internal gate to its group's trailing inner-column slot keeps it inside the group](https://github.com/microsoft/qdk/pull/3425/changes#diff-a07ac8e23fcc8091eb786d6999c905f24b406d03465004324ed5e18e09018ff5R144)
2. Good idea, added a new test: [moveOperation: promoting a gate into its grandparent group lands it beside the parent group](https://github.com/microsoft/qdk/pull/3425/changes#diff-a07ac8e23fcc8091eb786d6999c905f24b406d03465004324ed5e18e09018ff5R172) 
3. This doesn't make sense to me. We don't currently support moving groups (look to a future PR for that), but even when we do, you are asking about a group moving _into_ itself. I'm not sure how we'd express that in the current API. If you have more thoughts on this, let's follow up on this offline.
4. Good idea, added a new test: [moveOperation: moving a gate into a sibling group relocates it across scopes](https://github.com/microsoft/qdk/pull/3425/changes#diff-a07ac8e23fcc8091eb786d6999c905f24b406d03465004324ed5e18e09018ff5R201)

source/npm/qsharp/test/circuit-editor/circuit-actions/addRemove.test.mjs
ScottCarda-MS: They will return `null`. Look at these tests:

1. [addOperation: an op index one past the column's append slot returns null](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R132)
2. [addOperation: a column index one past the trailing-append slot returns null"](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R145)

source/npm/qsharp/test/circuit-editor/circuit-actions/addRemove.test.mjs
ScottCarda-MS: Take a look at these tests:

1. [removeOperation of the only remaining gate empties the circuit](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R303)
2. [removeOperation at an out-of-bounds location is a safe no-op](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R317)
3. [removeOperation at an in-range but empty interior slot is a safe no-op](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R330)

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAddRemove.test.mjs
ScottCarda-MS: It should prune any empty groups, but I realize I only tested that for the move-actions, and not for the pure delete-action. Updated!
[removeOperation: removing a group's last child prunes the emptied group](https://github.com/microsoft/qdk/pull/3425/changes#diff-461dfa7e8516657c05acf7d0b00909211da870fb818aaeb397ffd284b6760cb0R65)

source/npm/qsharp/test/circuit-editor/circuit-actions/groupAddRemove.test.mjs
ScottCarda-MS: Some existing tests were spread across other files. I've moved them here to consolidate and also added some tests in-line with the addRemove test file.

source/npm/qsharp/test/circuit-editor/circuit-actions/groupMove.test.mjs
ScottCarda-MS: [addControl: refuses on a classically-controlled GROUP (groups never carry quantum controls by design)](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R578)
[addControl: refuses on a plain group (no classical conditions)](https://github.com/microsoft/qdk/pull/3425/changes#diff-34b45b8b8f4da109f8d101fdc5c60d40d6683511340cae246088f61b6df88d94R630)

source/npm/qsharp/test/circuit-editor/circuit-actions/groupMove.test.mjs
ScottCarda-MS: This is covered in [removeOperation: removing a group's last child prunes the emptied group](https://github.com/microsoft/qdk/pull/3425/changes#diff-461dfa7e8516657c05acf7d0b00909211da870fb818aaeb397ffd284b6760cb0R65).
If you are just moving the operation, there will always be some ancestor that is non-empty, considering root is the ancestor of all gates/groups.

source/npm/qsharp/test/circuit-editor/circuit-actions/measurementCascade.test.mjs
ScottCarda-MS: Not really; the `meas` and `gate` functions these wrap around are more generally useful and are found in the utility file. These variants have shapes that are more specific to this file.

source/npm/qsharp/test/circuit-editor/circuit-actions/qubitOps.test.mjs
ScottCarda-MS: [moveQubit: moving an interior empty wire to the bottom prunes it as a trailing unused wire](https://github.com/microsoft/qdk/pull/3425/changes#diff-8fb904347e0b6d48bb46ac9bc665a6337153a0d606e92d6b03842cc573f4a48dR104)

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
ScottCarda-MS: As in, you want a test that putting in "(3 + 4) * 7" yields 49?

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
ScottCarda-MS: The angleExpression isn't evaluated here. Really, no part of the circuit is evaluated like that in the editor or visualizer. That's the job of the generated Q# code, which is very much separate from the circuit or editor code. That's why these tests very specifically focus only on the validation of the syntax. The actual expression is carried forward to the Q# and evaluated by the Q# runtime.

ScottCarda-MS: Capital Π is not supported because we don't expect users to be able to type that in. Lower-case is only supported because that's what other representations get converted into.

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
ScottCarda-MS: It seemed silly to support multiplication and division (which is useful in most expression with pi) without also supporting addition/subtraction and parens. Being able to support basic arithmetic will be useful when we add support for referencing things other than numbers here, but that is an idea for the future.

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
ScottCarda-MS: The evaluation of the arithmetic is not performed by the editor, as that is a Q# runtime task.

source/npm/qsharp/test/circuit-editor/contextMenu.test.mjs
ScottCarda-MS: Unfortunately it does not seem to work. The function call doesn't properly support it and using https causes errors.

source/npm/qsharp/test/circuit-editor/circuitModel.test.mjs
ScottCarda-MS: The CircuitModel has a [removeTrailingUnusedQubits](https://github.com/microsoft/qdk/pull/3425/changes#diff-a3ebc8da285b3ce007e80a9c5418a5e35dd0ed0da1862e45a97e7c17ee349434R80) that gets called by many circuit actions and also [during event loading](https://github.com/microsoft/qdk/pull/3425/changes#diff-3088c7aa28728f61794f019287cbfd37c4873ed5179ba427d511815998c92355R103).

We don't depend on this for data correctness. It is a convenient way to manage the size of a circuit. I don't believe we should go out of our way to be defensive about hand-rolled JSON circuits that our tooling can't produce. These would be unsupported scenarios. However, in this case, the circuit editor would happen to deal with this immediately on load. The circuit visualizer (no editor) doesn't do this pruning, so the extra qubit lines would appear there.

source/npm/qsharp/test/circuit-editor/contextMenu.test.mjs
ScottCarda-MS: There should be a canonical order, yes. It probably doesn't matter for the code, but aesthetically the order is intentional for user experience. Were you thinking of a different order would make for a better experience?

source/npm/qsharp/test/circuit-editor/draggable.test.mjs
ScottCarda-MS: good idea. Done

source/npm/qsharp/test/circuit-editor/draggable.test.mjs
ScottCarda-MS: The geometry logic for the editor and circuits more broadly has always been pretty detailed. If it needs to change, it _will_ take a lot of consideration, but once someone really knows the formula they want to implement, they will be able to update these tests to match that formula.

source/npm/qsharp/test/circuit-editor/dropzones.test.mjs
ScottCarda-MS: You are right, and this is using an unordinary shape for the parsed location. It is specific only to the following test, so I'll leverage the production parser in a function local to the test that just converts to the bespoke data-shape needed for that test.

source/npm/qsharp/test/circuit-editor/findOperation.test.mjs
ScottCarda-MS: We have some tests here that cover out-of-bounds locations, but I think that is the best we can do in this layer, since we are here to tests the API contract. Testing DOM lifecycle is out-of-scope for these tests, and I think that is what would be necessary to do what you're asking regarding the comment.

[findOperation returns null for out-of-bounds op indices](https://github.com/microsoft/qdk/pull/3425/changes#diff-d8141884b39d20cc83886dbd283bdb0617022b8c9c34f255fa2accb708f07814R78)
[findParentArray/findParentOperation return null when an ancestor is missing](https://github.com/microsoft/qdk/pull/3425/changes#diff-d8141884b39d20cc83886dbd283bdb0617022b8c9c34f255fa2accb708f07814R87)

source/npm/qsharp/test/circuit-editor/prompts.test.mjs
ScottCarda-MS: There's a test above this one that checks for cancel behavior for this scenario, and cancel is generally tested throughout the tests here. Maybe I'm missing what you are asking about?

source/npm/qsharp/test/circuit-editor/scrollController.test.mjs
ScottCarda-MS: Yes, the left edge is special and has special behavior because we expect users to drag gates from the toolbox from the left side onto the circuit. The special behavior is that we disable scrolling at first from the left when dragging from the toolbar, but once the drag gesture has gotten significantly past the left edge, then it becomes enabled again and the user can scroll along the left edge like a normal edge.

source/npm/qsharp/test/circuit-editor/toolbox.test.mjs
ScottCarda-MS: It is really just asserting that there _is_ a button in this scenario, because the previous test covers the scenario where there is _no_ button.

source/npm/qsharp/ux/circuit-vis/ARCHITECTURE.md
ScottCarda-MS: Since the tests that run off of these layers don't have a DOM, if someone adds a querySelector call, for example, to the logic of these layers, the tests will complain about not having a document to query. Sort of an indirect way to enforce it though.
I could add lints to enforce this.

source/npm/qsharp/test/circuit-editor/selectionController.test.mjs
ScottCarda-MS: Good call. Done

source/npm/qsharp/ux/circuit-vis/ARCHITECTURE.md
ScottCarda-MS: I looked into adding lints. It is doable, but I kind of thing it is overkill for this.

ScottCarda-MS:
> 1. Bug: View-state can leak across unrelated circuit replacements.
>    The new in-place update path always reuses the same Sqore instance in circuit.tsx:84. Sqore explicitly says a clean draw should be used when opening a different circuit in sqore.ts:168, but the component never does that after first mount. This can carry expand/collapse preferences to a different circuit if locations happen to match.
> 2. Bug: Run button presence is fixed on first shell mount. Toolbox creation is callback-dependent in toolbox.ts:56, while shell mounting is one-time in shell.ts:72. If run callback availability changes later, the button will not appear/disappear accordingly because the panel is not rebuilt.
> 3. Inconsistency between ARCHITECTURE and implementation: Layering rule is documented as strict, but dependency boundaries are not actually strict.
>    Architecture claims no outgoing dependencies from data/actions in ARCHITECTURE.md:30 and ARCHITECTURE.md:60, yet action/data import helpers from utils.ts:4, and that file imports renderer types/constants and DOM-query helpers.

I fixed Bug 1. 2 is intentional behavior, the host will not change its behavior on whether it supports running or not between redraws. For 3, I have updated to try to make a more clear separation between the DOM-enabled layer and the DOM-disabled layers.

source/npm/qsharp/ux/circuit-vis/sqore.ts
fedimser: I agree this is not useful context and existing "Entrypoint class for rendering circuit visualizations." is enoguh. I thought maybe it's an abbreviation.

PR #3426, opened 2026-07-08
Circuit-Editor Multi-Target Gate Support

ScottCarda-MS:
Adds drag editing for multi-wire gates and expanded groups in the `circuit-vis` editor.

- Moves multi-target gates and groups as rigid units while preserving nested register layout and classical-control references.
- Supports cloning an entire unit with Ctrl-drag.
- Supports extending an expanded group onto new wires by Shift-dragging a child outside its current span.
- Aligns shift-extend ghosts and dropzones with rendered group bounds, including classical-control reach and extra width from long labels.

## Testing

Adds 30 circuit-editor tests covering unit moves, cloning, bounds handling, shift-extend behavior, and ghost/dropzone geometry.


fedimser (changes requested):
The diff includes what seems to be build artifacts in "pip" folder (those paths are in .gitignore so I don't understand how they made it into PR diff). Can you please remove those?

source/npm/qsharp/ux/circuit-vis/actions/circuit-actions/move.ts
```diff
@@ -46,6 +52,106 @@ const moveX = (
   );
 };
 
+/**
+ * Move `op` as one rigid unit (shift every register by the same
```
fedimser: Suggestion: "Returns whether `op` should be moved as one rigid unit ...".

When doc starts with "move", it appears that this function actually moves the op, until I see question mark in the end of the sequence.

I would also suggest to rename the function to "shouldMoveAsUnit", to make it clear that this function isnot doing the move.

fedimser (approved):

source/npm/qsharp/ux/circuit-vis/actions/circuit-actions/move.ts
ScottCarda-MS: good idea! done

PR #3430, opened 2026-07-08
Upgrade to TS7

amcasey:
- There was a Uint8Array/ArrayBuffer change in TS 5.7 that caused some problems.  I think we're okay with a cast. 
- Apparently, noUncheckedSideEffectImports:false is needed if you're importing CSS. 
- We still need TS6 for eslint (TS7 doesn't expose APIs). 
- ModuleResolution bundler appears to be the closest to node10.


source/vscode/src/azure/networkRequests.ts
```diff
@@ -82,7 +82,7 @@ export async function storageRequest(
   token?: string,
   proxy?: string,
   extraHeaders?: [string, string][],
-  body?: string | Uint8Array,
+  body?: string | ArrayBuffer,
```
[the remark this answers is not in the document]
amcasey: I don't want to expose `fetch` types in our signature.

source/vscode/src/language-service/activate.ts
```diff
@@ -172,7 +172,7 @@ async function loadLanguageService(
   const start = performance.now();
   const wasmUri = vscode.Uri.joinPath(baseUri, "./wasm/qsc_wasm_bg.wasm");
   const wasmBytes = await vscode.workspace.fs.readFile(wasmUri);
-  await loadWasmModule(wasmBytes);
+  await loadWasmModule(wasmBytes.buffer as ArrayBuffer);
```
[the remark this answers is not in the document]
amcasey: Stopped accessing `.buffer` directly.

source/vscode/package.json
```diff
@@ -1531,12 +1531,12 @@
     "start": "npm exec --no -- @vscode/test-web --extensionDevelopmentPath . ../../samples",
     "test": "node ./test/runTests.mjs -- ",
     "pretest": "npm exec --no playwright install chromium && node ./test/buildTests.mjs",
-    "tsc:check:main": "node ../../node_modules/typescript/bin/tsc -p ./tsconfig.json",
-    "tsc:check:view": "node ../../node_modules/typescript/bin/tsc -p ./src/webview/tsconfig.json",
-    "tsc:check:learning": "node ../../node_modules/typescript/bin/tsc -p ./src/learning/webview/tsconfig.json",
+    "tsc:check:main": "tsc -p ./tsconfig.json",
+    "tsc:check:view": "tsc -p ./src/webview/tsconfig.json",
+    "tsc:check:learning": "tsc -p ./src/learning/webview/tsconfig.json",
```
billti, line 1536: Why did all of these change? We don't assume TypeScript is installed globally, and its not a dependency in this package.json just the top level one, so previously this only worked if specifying the path to the top level `node_modules`. Did something change to resolve this?

package.json
```diff
@@ -51,8 +52,8 @@
     "preact": "^10.29.1",
     "prettier": "^3.8.1",
     "punycode": "^2.3.1",
-    "typescript": "5.6.3",
-    "typescript-eslint": "^8.58.2",
+    "typescript": "npm:@typescript/typescript6@^6.0.2",
```
billti, line 55: Maybe I'm a little confused or not up to date on versions. Why does this specify version 6 if the PR is to move to version 7?

amcasey: TS7 doesn't expose an API (yet) so esbuild won't work with it.  The recommended approach is to keep an aliased copy around (typescript6) for esbuild.
I tried to explain in the description - let me know how I can make it clearer.

source/vscode/package.json
amcasey: It can't be run as `node foo` anymore, since it's an executable.  As to whether it's guaranteed to be installed when this is run, I was a little unclear on that myself.  FWIW, the expectation was that it would pick up the one from the .bin folder created during `npm install` and not the one on the path (which we don't expect or require).

amcasey:
This didn't catch any bugs and TS build speed isn't an issue for us right now.

PR #3431, opened 2026-07-09
Fix benzene notebook

orpuente-MS:
This PR:
1. Fixes the `benzene.ipynb` notebook by explicitly passing `Profile.Base` to `qdk.openqasm.compile`.
2. Fixes the `Reorder` pass so that it can handle zero-qubit qis instructions, so that `NeutralAtomDevice.simulate` works with adaptive QIR.
3. Raises an exception when running `NeutralAtomDevice.show_trace` with adaptive QIR that cannot be traced, i.e.: QIR with dynamic angles or qubit IDs that cannot be resolved by the visualizer.

### DIFF
--- a/samples/notebooks/benzene_molecule/benzene.ipynb
+++ b/samples/notebooks/benzene_molecule/benzene.ipynb
@@ -27,7 +27,7 @@
     "src = Path(\"benzene_diradical_injected_rotation_measurement_circuit_0.qasm\").read_text(encoding=\"utf-8\")\n",
     "\n",
     "init(target_profile=TargetProfile.Base)\n",
-    "qir = compile(src)"
+    "qir = compile(src, target_profile=TargetProfile.Base)\n"
    ]
   },
   {
--- a/samples/notebooks/benzene_molecule/benzene.ipynb
+++ b/samples/notebooks/benzene_molecule/benzene.ipynb
@@ -19,14 +19,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from qdk import init, TargetProfile\n",
+    "from qdk import TargetProfile\n",
     "from qdk.openqasm import circuit, compile, run\n",
     "from qdk.widgets import Circuit, Histogram\n",
     "\n",
     "from pathlib import Path\n",
     "src = Path(\"benzene_diradical_injected_rotation_measurement_circuit_0.qasm\").read_text(encoding=\"utf-8\")\n",
     "\n",
-    "init(target_profile=TargetProfile.Base)\n",
     "qir = compile(src, target_profile=TargetProfile.Base)\n"
    ]
   },
--- a/source/qdk_package/qdk/_device/_atom/__init__.py
+++ b/source/qdk_package/qdk/_device/_atom/__init__.py
@@ -227,23 +227,32 @@ class NeutralAtomDevice(Device):
         start_time = time.monotonic()
         telemetry_events.on_neutral_atom_trace()
 
-        # `show_trace` statically analyzes the program to lay it out and schedule
-        # it, which requires fully inlined, Base-profile QIR with constant gate
-        # parameters. Adaptive-profile QIR (dynamic rotation angles, branching,
-        # and non-inlined gate definitions) cannot be traced this way.
-        profile = _get_qir_profile(qir)
-        if profile is not None and profile != "base_profile":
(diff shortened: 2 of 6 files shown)


samples/notebooks/benzene_molecule/benzene.ipynb
```diff
@@ -27,7 +27,7 @@
     "src = Path(\"benzene_diradical_injected_rotation_measurement_circuit_0.qasm\").read_text(encoding=\"utf-8\")\n",
     "\n",
     "init(target_profile=TargetProfile.Base)\n",
-    "qir = compile(src)"
+    "qir = compile(src, target_profile=TargetProfile.Base)\n"
```
billti, line 29: Does the `init` call on line 29 even do anything now? Could just remove that (and the imports) probably.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
```diff
@@ -191,3 +192,48 @@ def test_s_adj_noise_inherits_from_rz():
...
+        """)
+
+    device = NeutralAtomDevice()
+    with pytest.raises(ValueError, match="could not trace this program"):
+        device.show_trace(qir)
+
+
+@pytest.mark.skipif(not PYQIR_AVAILABLE, reason=SKIP_REASON)
+@pytest.mark.skipif(not QSHARP_WIDGETS_AVAILABLE, reason=WIDGETS_SKIP_REASON)
+def test_show_trace_allows_traceable_adaptive_program() -> None:
```
billti, line 247: Is it worth a test to ensure you can trace a program with conditions, loops, or calls, as long as the qubit ids aren't dynamic?

source/qdk_package/qdk/_device/_atom/__init__.py
```diff
@@ -212,12 +227,32 @@ def show_trace(self, qir: str | QirInputData):
...
+            Schedule(self).run(module)
+            tracer = Trace(self)
+            tracer.run(module)
+        except (AttributeError, IndexError) as e:
+            profile = _get_qir_profile(qir)
+            if profile is not None and profile != "base_profile":
+                raise ValueError(
+                    "`show_trace` could not trace this program. This "
+                    "Adaptive-profile program contains values known only at run "
+                    "time (for example, rotation angles computed from measurement "
```
billti: The angle value actually isn't that interesting in a trace. I wonder if we could even allow them and just put a dummy substitute value in the trace (e.g. dynamic angles always show as 0.0, or perhaps even a string like `?` or `dyn` as it looks like we just treat the args as an opaque string in the trace - we don't try to parse a float or anything)

samples/notebooks/benzene_molecule/benzene.ipynb
orpuente-MS: I removed it, the openqasm codepath is not interacting with the context.

source/qdk_package/tests-integration/devices/test_atom_e2e.py
orpuente-MS: Actually, after looking more closely at the trace, I realized that those programs were being mis-visualized. The trace was just wrong, since it can't follow across conditional-branches and function calls. So, I added back `ValidateNoConditionalBranches` to the tracer (which I had removed as part of #3413) and added an additional `ValidateNoFunctionCalls`.

I also added tests ensuring that we can't trace programs with conditional branching or function calls.

source/qdk_package/qdk/_device/_atom/__init__.py
orpuente-MS: Since I disallowed conditional branching (as I explained in the previous question), we don't even reach this anymore.

source/qdk_package/qdk/_device/_atom/_reorder.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, device: Device):
 
     def instr_key(self, instr: Instruction):
         gate = as_qis_gate(instr)
-        if gate != {}:
+        if gate != {} and gate["qubit_args"]:
```
orpuente-MS, line 38: This change is to make `NeutralAtomDevice.simulate` work.

And the tests in `test_atom_reorder.py` are testing this change.

billti (approved):

PR #3432, opened 2026-07-09
Use RCA to generate inline hints, update tests to use `Adaptive`

swernli:
This change updates the samples tests to use the Adaptive Profile to match our current editor default, and fixes one of the bugs found in the process. Because Adaptive includes emission of IR functions and those functions have all arguments as variables, there was a mismatch between how RCA generated the capabities of a given call vs what it might actually use at runtime. Since RCA uses the static vs dynamic constant vs dynamic variable values of a given call site, it might assume a call uses only simple features, but when emitting it as a function all arguments are variables and more complex features are used. This updated strategy instead uses a pattern similar to loops, where the effective capabilities of emitting the function are compared to the target capabilities and any functions that would incur too advanced of capabilities are marked as requiring inlining.

Note: the Ising 2D samples return an error when generating QIR with Adaptive Profile that will require a separate fix.


idavis (approved):

swernli:

PR #3436, opened 2026-07-09
Fallback to profile used in `init` if any during `openqasm.compile`

swernli:


idavis (approved):

source/qdk_package/qdk/openqasm/_compile.py
```diff
@@ -58,15 +61,20 @@ def compile(
...
+        target_profile = str(target_profile_from_init)
+        kwargs["target_profile"] = target_profile_from_init
 
     telemetry_events.on_compile_qasm(target_profile)
 
     if isinstance(source, Callable) and hasattr(source, "__global_callable"):
-        args = python_args_to_interpreter_args(args)
+        args = python_args_to_interpreter_args(args)  # type: ignore
         ll_str = get_interpreter().qir(
             entry_expr=None, callable=source.__global_callable, args=args
```
billti, line 79: Maybe unrelated to this change, but kwargs and target_profile aren't used at all here. So what is the effect of the 'target_profile' if a Callable is passed instead of string source?

swernli: I'm not sure why, but after I changed the file I suddenly got a pylance type warning here. I agree it's unrelated to my changes, but was worried that it would fail the new checks in the build if I left it alone.

swernli: sorry, misunderstood your question. Yes, in the callable case the user has to have previously gone through `openqasm.import_openqasm` which does already use the currently configured profile on the interpreter. For that case the existing behavior is indeed that kwargs is unused, as all those properties have already been incorporated into the compiled callable being passed in.

PR #3443, opened 2026-07-10
Send telemetry for wasm panics

amcasey:
For now, we're including basically no information - just that it happened.  Error messages are privacy minefields.


amcasey:
I'm not sure this is the most elegant way to do this (since I don't know exactly where our layer boundaries are), but I made sure I saw it work in both VS Code and the playground locally.

swernli (approved):

PR #3444, opened 2026-07-10
Updated TM grammar

filipw:
This PR imports an updated TextMate grammar providing much richer syntax highlighting support for Q# than what is currently here.
 
It's vendored from https://github.com/qsharp-community/qsharp-tmLanguage following a similar approach that we took years ago in the C# extension for VS Code (the grammar was [external](https://github.com/dotnet/csharp-tmLanguage), with a proper test suite and build process, and just vendored into the C# extension repo).

The grammar repo was set up a long time ago in 2022 (I remember I showed it to the original team members years ago in a call - Cassandra, Bettina, Sarah etc - and we discussed at some point integrating it back into the extension). It was recently modernized to support the modern Q# syntax, and has all the things you'd expect from a grammar repo - tests, CI, build process etc.

Below is a comparison on a random file (first old, then new):
Screenshot 2026-07-09 at 18 01 50
Screenshot 2026-07-09 at 18 02 11

Github.com sources its own TM grammar for the website from this QDK repo, so updating here would also improve syntax highlighting for Q# on github.com which would be a big win.

### DIFF
--- a/source/vscode/syntaxes/qsharp.tmLanguage.json
+++ b/source/vscode/syntaxes/qsharp.tmLanguage.json
@@ -1,1449 +1,843 @@
 {
-	"$schema": "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json",
-	"name": "qsharp",
-	"scopeName": "source.qsharp",
-	"fileTypes": [
-		"qs"
-	],
-	"patterns": [
-		{
-			"include": "#comments"
-		},
-		{
-			"include": "#attributes"
-		},
-		{
-			"include": "#namespace-declaration"
-		},
-		{
-			"include": "#import-export-directive"
-		},
-		{
-			"include": "#new-expression"
-		},
-		{
-			"include": "#struct-declaration"
-		},
-		{
-			"include": "#member-access"
-		},
-		{
-			"include": "#tuple-binding"
-		},
-		{
-			"include": "#keywords"
-		},
-		{
-			"include": "#library-invocation"
-		},
-		{
-			"include": "#library"
-		},
-		{
-			"include": "#operations"
-		},
-		{
-			"include": "#types"
-		},
-		{
-			"include": "#literals"
-		},
-		{
-			"include": "#strings"
-		},
-		{
-			"include": "#callable-invocation"
-		},
-		{
-			"include": "#callable-declaration"
-		},
-		{
-			"include": "#udt-declaration"
-		},
-		{
-			"include": "#array-creation-expression"
-		},
-		{
-			"include": "#local-declaration"
-		},
-		{
-			"include": "#array-assignment"
-		},
-		{
-			"include": "#local-assignment"
-		},
-		{
-			"include": "#quantum-declaration"
-		},
-		{
-			"include": "#if-statement"
-		},
-		{
-			"include": "#parenthesized-expression"
-		},
-		{
-			"include": "#expression-operators"
-		},
-		{
-			"include": "#declaration-keywords"
-		},
-		{
-			"include": "#identifier"
-		}
-	],
-	"repository": {
-		"literals": {
-			"patterns": [
-				{
-					"include": "#boolean-literal"
-				},
-				{
-					"include": "#result-literal"
-				},
-				{
-					"include": "#pauli-literal"
-				},
-				{
-					"include": "#numeric-literal"
-				}
-			]
-		},
-		"comments": {
-			"patterns": [
-				{
-					"begin": "(^\\s+)?(?=//)",
-					"beginCaptures": {
-						"1": {
-							"name": "punctuation.whitespace.comment.leading.qsharp"
-						}
-					},
-					"end": "(?=$)",
-					"patterns": [
-						{
-							"name": "comment.block.documentation.qsharp",
-							"begin": "(?<!/)///(?!/)",
-							"beginCaptures": {
-								"0": {
-									"name": "punctuation.definition.comment.qsharp"
-								}
-							},
-							"end": "(?=$)"
-						},
-						{
-							"name": "comment.line.double-slash.qsharp",
-							"begin": "(?<!/)//(?:(?!/))",
-							"beginCaptures": {
-								"0": {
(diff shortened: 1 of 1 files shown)


source/vscode/syntaxes/qsharp.tmLanguage.json
```diff
@@ -7,202 +7,1385 @@
       "include": "#comments"
```
ScottCarda-MS: It looks like the highlighting for tuple declaration is different from simple declaration, which seems wrong. If I `let (x, y) = (10, 20);` the `x` and `y` don't get colored the same way as if I do `let x = 10; let y = 20;`.

ScottCarda-MS: The `set` keyword gets a little confused in some circumstances. For exmaple:
```qsharp
let x = 0;
mutable ary = [0];
set ary w/= x <- 10;
```

This colors the `set` as though it is a variable reference when the index expression is a variable and not a literal.

ScottCarda-MS: Array indexing expressions have issues getting their brackets colored distinctly from the array name. I'm seeing that often it gets colored the same color. For example: `total[y];` all has the same color, which is not as useful as having the brackets stand out with a different color.

ScottCarda-MS: It is a bit unfortunate to be typing out an operation and not get the coloring for the keywords for so long. For example typing `operation Foo` doesn't trigger any special highlighting until I put the `(` character and start typing the parameters. Same thing with `let temp`. Seems like keywords should be able to be colored as such immediately on finishing typing the space ending the keyword.

ScottCarda-MS (commented):
Thanks for adding this! I was playing around with it and found a couple of spots that I think need attention.

source/vscode/syntaxes/qsharp.tmLanguage.json
filipw: may I ask you with which theme are you testing and exact code you are testing? I honestly don't see such issue.

TextMate grammar only attaches scopes - the actual colors come from the theme's scope to color mapping. The grammar already emits distinct `punctuation.squarebracket.*` scopes vs. `variable.other.readwrite`, so if they appear the same to you, perhaps it's your theme's configuration.

In fact for this specific case, the matching is structurally very similar to what the current basic grammar does.

filipw: this is indeed a bug

filipw: yeah this is a small fix too, we need a new tuple-pattern subrule

filipw: yeah incomplete constructs are always tricky (when does which rule finally match?) but we can improve it with a low-priority fallback for the keywords, it's not a big deal :)

ScottCarda-MS: Sure! I'm running VS Code Insiders and using the Dark High-Contrast theme (my eyes need the high-contrast).
This is an operation I have that I'm seeing issues with the array syntax:
```qsharp
operation Foo() : Unit {
    let (x, y) = (5, 10);
    mutable total = [0];
    set total w/= x <- 0;
    total[y];
    let y = 10;
    let z = x + y;
    Message($"z={z}");
}
```

ScottCarda-MS: I think you are correct in calling the theme out as an issue here. When I cycle through other themes, I can see the brackets are differentiated. I can see that the brackets in dark high-contrast theme will cycle through different colors depending on the scope depth. Unfortunate that the first color in that cycle is the same color as the variable names. I don't remember it doing that before, but that seems way too particular to the theme's logic for us to worry about in the grammar file changes here. I will close out this comment. Sorry for the noise!

filipw: This is now resolved. 

Before: 
Screenshot 2026-07-15 at 15 00 17

After:
Screenshot 2026-07-15 at 15 00 36

filipw: This is also resolved.

Before:
Screenshot 2026-07-15 at 15 00 09

After:
Screenshot 2026-07-15 at 15 00 30

filipw: This should now be resolved. For incomplete constructs, we now have a fallback with common keywords, so the "incomplete" experience is much smoother.

filipw:
Thanks for trying this out 😄

source/vscode/syntaxes/qsharp.tmLanguage.json
```diff
@@ -1,211 +1,1448 @@
 {
-  "$schema": "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json",
```
ScottCarda-MS, line 2: Did you intend to remove the schema? My understanding is that including it provides additional IDE support for editing this grammar file, but maybe there's something I'm missing.

filipw: I modeled this after how we did it in C# back in the days - the grammar is never edited in JSON, [it's written in YML](https://github.com/qsharp-community/qsharp-tmLanguage/blob/main/src/qsharp.tmLanguage.yml) and from there [built](https://github.com/qsharp-community/qsharp-tmLanguage/blob/main/build.js) into different output formats (JSON, plist, could be something else too), so the schema is not really needed in the output JSON.

but I can bring it back though, it's not a problem to emit it.

filipw: it's back

ScottCarda-MS (approved):
I've played around with it a bit and it seems like it is working great! I appreciate you taking the time to put this together. I think this is a big improvement on what we had for syntax highlighting.

filipw:
I see it broke on the prettier check, let me address this

filipw:
the tabs are gone, apologies for my sins 😅

PR #3451, opened 2026-07-10
Break and continue support

### LINKED ISSUE #1835: Language Feature: Loop shortcut constructs
Add `break` and `continue` as statements.

```qsharp
for i in 1..5 {
    if i == 4 {
        break;
    }
    continue;
}
```

Only legal in `while` and `for` loops.

`break` moves execution to the statement following the block expression. Using `break` or `continue` within a block changes that blocks type to `Unit`

Unknowns
- behavior of break/continue based on quantum  measurements
- behavior in conjugate expressions
   ```qsharp
    within {
        for i in 1..5 {
            if i == 4 {
                break;
            }
            continue;
        }
    }
    apply {
        X(q);
    }
```

### PR DESCRIPTION
**Pull request by idavis**
Closes #1835

### CODE REVIEW

**Comment by amcasey** [comment]
This is probably supposed to report that break isn't allowed there, rather than that the type couldn't be inferred
```qsharp
operation Main() : Unit {
    mutable x = 1;
    while (true) {
        () -> break;
    }
    x += 1;
}
```
```
QdkDiagnostics: type error: insufficient type information to infer type

help: provide a type annotation at code:4:15
```

amcasey:
This reports an RIR timeout
```qsharp
operation Main() : Unit {
    use q = Qubit();
    for p in [q] {
        for p in [q] {
            for p in [q] {
                for p in [q] {
                    for p in [q] {
                        for p in [q] {
                            break
                        }
                    }
                }
            }
        }
    }
}
```

source/vscode/syntaxes/qsharp.tmLanguage.json
```diff
@@ -142,7 +142,7 @@
       "patterns": [
         {
           "name": "keyword.control.qsharp",
-          "match": "\\b(use|borrow|mutable|let|set|if|elif|else|repeat|until|fixup|for|in|while|return|fail|within|apply)\\b"
```
amcasey: Would you expect the local playground to pick this up?  I'm not seeing it.

idavis: @amcasey I think monaco uses their own language def based on ours. @billti - do you remember?

amcasey:
I'm seeing an LS panic (and a possibly downstream RIR generation failure) for
```qsharp
@EntryPoint(Adaptive)
operation Main() : Result {
    use qq = Qubit[2];
    repeat {
        H(qq[0]);
        Controlled X([qq[break]], qq[1]);
    } until (M(qq[1]) == One)
    fixup {
        ResetAll(qq);
    }
    let r = M(qq[0]);
    ResetAll(qq);
    r
}
```
```
[error] [wasm] Wasm panic occurred: panicked at source\compiler\qsc_partial_eval\src\lib.rs:2722:18:
invalid kind of value for index
```
```
RuntimeError: unreachable
```

Related: with `Controlled X([break], qq[1]);`, I see 
```
QdkDiagnostics: partial evaluation error: partial evaluation failed with error: index out of range: 0 at code:6:23
```

Interestingly, `Controlled X(break, qq[1]);` appears to work (i.e. emit QIR and run)

amcasey:
LS panic for
```qsharp
@EntryPoint(Adaptive)
operation Main() : Result {
    use qq = Qubit[2];
    repeat {
        H(qq[0]);
        use rr = Qubit[break];
        Controlled X(break );
    } until (M(qq[1]) == One)
    fixup {
        ResetAll(qq);
    }
    let r = M(qq[0]);
    ResetAll(qq);
    r
}
```
```
Wasm panic occurred: panicked at source\compiler\qsc_rca\src\core.rs:2807:13:
expected tuple expression
```
I was about to type a comma after the `break` in the `Controlled`

amcasey:
I don't think this needs to work, but it stands out from a bunch of other nonsensical code that does work
```qsharp
@EntryPoint(Adaptive)
operation Main() : Result {
    use qq = Qubit[2];
    repeat {
        H(qq[break]);
        Controlled X(continue, qq[1]);
    } until (M(qq[1]) == One)
    fixup {
        ResetAll(qq);
    }
    let r = M(qq[0]);
    ResetAll(qq);
    r
}
```
```
main.qs@(4,5): cannot use a dynamically-sized array
help: using an array whose size depends on a measurement result is not supported by the configured target profile
```

In particular, the same example emits QIR with `H(qq[continue])` (though I'm pretty sure it would loop infinitely?)

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5321,235 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+fn return_and_break_in_loop_composes_with_return_unify() {
+    // A loop containing both a `return` and a `break`. The `break` desugar in HIR
+    // `loop_unification` rewrites the loop condition to `(not .broke_<id>) and cond`;
+    // `return_unify` in FIR then wraps it again as
+    // `(not __has_returned) and ((not .broke_<id>) and cond)`. Because the synthetic
+    // break flag `.broke_<id>` uses a distinct name from the return flags, the
+    // `check_no_flag_writes_in_operand_position` invariant, which scans only the
+    // return-flag names, never fires on it. Successful QIR generation therefore
+    // demonstrates the two flag lowerings compose cleanly with no invariant
+    // conflict and no residual early exit.
```
ScottCarda-MS: I'd lean and making this comment briefer. It shouldn't need to go into the internal details, but just say that we need to check that `return` and `break` play nicely together.

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5321,235 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+    assert!(
+        rir_text.contains("name: __quantum__qis__x__body"),
+        "expected the X gate callable to survive into RIR:\n{rir_text}"
+    );
+    assert!(
+        rir_text.contains("name: __quantum__qis__y__body"),
+        "expected the Y gate callable to survive into RIR:\n{rir_text}"
+    );
+    assert!(
+        rir_text.contains("Return Integer(0)"),
```
ScottCarda-MS: Do we need to check that there aren't other `Return Integer(0)` in the text (multiple returns)?

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5321,235 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+#[test]
+fn bare_operand_break_lowers_to_qir_and_skips_eager_consumer() {
+    // A bare-operand `break` buried in a call argument such as `Sink(q, break)` is
+    // hoisted to statement position and desugared to flags, so the whole
+    // HIR -> FIR -> RIR -> QIR pipeline lowers with no residual early-exit node,
+    // since the lowerer panics on any leftover `break`. The eager consumer `Sink`
+    // applies an `H` gate, so a successful lowering with no `H` in the QIR
+    // proves the break fires before the argument is consumed. Previously the
+    // buried break was left in operand position, which would instead run
+    // `Sink`, emitting `H`, on the break iteration.
```
ScottCarda-MS: Should remove the "Previously the buried break..." line as that was probably a concern mid-development of the PR.

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5321,235 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+        "expected the Y gate callable to survive into RIR:\n{rir_text}"
+    );
+    assert!(
+        rir_text.contains("Return Integer(0)"),
+        "expected the desugared program to lower to a single clean return:\n{rir_text}"
+    );
+}
+
+#[test]
+fn bare_operand_break_lowers_to_qir_and_skips_eager_consumer() {
```
ScottCarda-MS, line 5569: It's an interesting test. Should we also test that `Foo(H(q), break, Y(q))` yields a an H call but not a Y call?

source/compiler/qsc/src/interpret.rs
```diff
@@ -2092,13 +2092,17 @@ impl<'a> BreakpointCollector<'a> {
...
+                // clients get one stable, hittable breakpoint per visual location.
+                // Multiple HIR passes, including ReplaceQubitAllocation, LoopUni,
+                // conjugate_invert, and spec_gen, generate statements that share a
+                // start position but differ in span. For example, the qubit-release
+                // desugar wraps `return e;` in a block whose inner `return e` keeps
+                // the return-expression span, one column shorter than the outer
+                // statement's `return e;` span. Keying on the start position
+                // collapses such overlapping locations onto the outer statement
+                // seen first, so a single source line maps to a single breakpoint.
+                self.statements.entry(range.start).or_insert(bps);
```
ScottCarda-MS, line 2120: Does this mean you wouldn't be able to have multiple breakpoints with the same starting position? I know that sometimes breakpoint-ed expressions in some langs can nest inside other breakpoint-ed expressions. For example `Foo(x) + Bar(y)` could have a breakpoint on `Foo(x)` and a breakpoint on the whole addition expression, both of which have the same starting position.

source/compiler/qsc/src/interpret.rs
```diff
@@ -2108,11 +2112,21 @@ impl<'a> Visitor<'a> for BreakpointCollector<'a> {
...
         match stmt_res.kind {
-            fir::StmtKind::Expr(expr) | fir::StmtKind::Local(_, _, expr) => {
+            // Recurse into the statement's expression so that nested statements
+            // become individually breakpointable. `Semi` is included alongside
+            // `Expr`/`Local` because block-like statements written with a
+            // trailing semicolon, notably a desugared `repeat ... until cond;`
+            // that is always a `Semi`, carry their loop body inside this
+            // expression. Recursing into a simple `Semi` such as `foo();` or
+            // `set x = 1;` adds no breakpoints, since `walk_expr` only surfaces
+            // breakpoints when it re-enters a nested statement.
```
ScottCarda-MS: Might be easier/briefer to explain why StmtKind::Item is exempt from further walking. Intuitively, without knowing the development history of the code, one would expect a StmtKind case containing an expression to be a valid point to walk the nested expression.

source/compiler/qsc/src/interpret.rs
idavis: We don't allow breakpoints at expr bounds, only statements. So two statements can't have the same start.

source/compiler/qsc_fir_transforms/src/return_unify/normalize/tests/regression_and_depth.rs
```diff
@@ -57,7 +57,7 @@ fn if_then_return_else_return_at_end_records_flag_lowered_shape() {
                     [0] Local(Mutable, _.has_returned: Bool): Lit(Bool(false))
                     [1] Local(Mutable, _.ret_val: Int): Lit(Int(0))
                     [2] Local(Immutable, q: Qubit): Call[ty=Qubit]
-                    [3] Local(Immutable, @generated_ident_59: Unit): If(cond=BinOp(Eq)[ty=Bool], then=Block[ty=Unit], else=Block[ty=Unit])
+                    [3] Local(Immutable, .generated_ident_59: Unit): If(cond=BinOp(Eq)[ty=Bool], then=Block[ty=Unit], else=Block[ty=Unit])
```
ScottCarda-MS, line 60: It isn't obvious to me why this changed. Is there a reason for the character change here?

idavis: For ease of readability we switched to `.` led names for compiler generated names in the FIR transforms and continued it here.

ScottCarda-MS: Discussed offline. Updated this code to an already-agreed-upon standard.

source/vscode/syntaxes/qsharp.tmLanguage.json
amcasey: That sounds right.  I think Mine mentioned that this wouldn't work automatically.

source/compiler/qsc_fir_transforms/src/invariants.rs
```diff
@@ -712,34 +712,41 @@ fn check_non_unit_block_tail(package: &Package, block_id: BlockId, context: &str
...
 fn block_diverges(package: &Package, block_id: BlockId) -> bool {
     let block = package.get_block(block_id);
-    let Some(&stmt_id) = block.stmts.last() else {
-        return false;
-    };
-    match &package.get_stmt(stmt_id).kind {
-        StmtKind::Expr(expr_id) | StmtKind::Semi(expr_id) => expr_diverges(package, *expr_id),
-        StmtKind::Local(..) | StmtKind::Item(_) => false,
-    }
+    block
```
ScottCarda-MS: Changing this to look at all statements in a block means changing the divergence of blocks that aren't related to `break`/`continue`. For example, previously when we just check the last statement, a block with a divergent `fail` in the middle of it would have been considered non-divergent. But now that non-trailing `fail` would cause the whole block to be considered divergent.
I may be missing some fir-specific knowledge that addresses this concern.

source/compiler/qsc/src/interpret.rs
ScottCarda-MS: Ah okay, then it is not a concern for our debugging experience.

source/compiler/qsc/src/interpret/debugger_tests.rs
```diff
@@ -283,4 +329,558 @@ mod given_debugger {
...
+            let break_id = breakpoint_id_for_text(&debugger, "test", FOR_BREAK_SOURCE, "break");
+            expect_bp(&mut debugger, &[break_id], break_id);
+
+            let mut names = debugger
+                .get_locals(1)
+                .into_iter()
+                .map(|variable| variable.name.to_string())
+                .collect::<Vec<_>>();
+            names.sort();
+            assert_eq!(names, ["i", "total"]);
```
ScottCarda-MS, line 513: Since this test is about hiding compiler-generated names, what would we expect to _not_ see here? Maybe a comment about what compiler-generated names are involved in this code-sample that the debugger would not report on would be helpful.

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5313,319 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+fn operand_breaks_in_qubit_allocation_and_controlled_call_generate_qir() {
+    let source = r#"
+        namespace Test {
+            @EntryPoint(Adaptive)
+            operation Main() : Result {
+                use qq = Qubit[2];
+                repeat {
+                    H(qq[0]);
+                    use rr = Qubit[break];
+                    Controlled X(break);
```
ScottCarda-MS: Does it actually generate code for this second `break`? The first break would prevent the second from being tested, no?

ScottCarda-MS: Just going off of the test name, it sounds like the expectation is that both would generate QIR, but the asserts below assert that the controlled X does not generate QIR, which would be expected regardless of the arguments to the controlled X.

source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5321,3 +5313,319 @@ fn chemistry_like_iqpe_with_udt_capture_closure_generates_base_profile_qir() {
...
+    "#;
+
+    let qir = compile_source_to_qir(source, Profile::Adaptive.into());
+    assert!(
+        !qir.contains("call void @__quantum__qis__h__body"),
+        "the index break should prevent H from running:\n{qir}"
+    );
+    assert!(
+        !qir.contains("call void @__quantum__qis__x__body"),
+        "the controlled X after the break should be unreachable:\n{qir}"
```
ScottCarda-MS, line 5736: This should be true regardless of what the arguments to the controlled X are. Is this test making super-sure of this case due to some internal implementation detail of break and continue?

source/compiler/qsc/src/interpret/debugger_tests.rs
```diff
@@ -283,4 +329,558 @@ mod given_debugger {
...
+        }
+
+        #[test]
+        fn stepping_through_for_break_loop_never_reports_inverted_span() {
```
ScottCarda-MS: Looks like this is in response to a mid-development bug that has been fixed. The test stands out as a bit strange as one wouldn't normally expect a `break` to cause an issue like that. If this is an issue that needs to be guarded against because changes to the implementation could accidentally trigger this behavior again, then the tests is worth keeping, but otherwise I'd lean towards removing it. If you do keep it, the comment description should be updated to remove the historical-wording.

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -45,6 +46,69 @@ impl<T> Partial<T> {
...
+            if self.divergent_stmts.contains(&stmt.id) {
+                break;
+            }
+        }
+    }
+
+    fn visit_expr(&mut self, expr: &'a Expr) {
+        match &*expr.kind {
+            ExprKind::Break if self.loop_depth == 0 => self.has_break = true,
+            ExprKind::Continue if self.loop_depth == 0 => self.has_continue = true,
```
ScottCarda-MS: This doesn't seem to do anything for nested loops. Rather than keeping track of loop depth, can we just not visit the bodies of these loops?

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -363,9 +432,23 @@ impl<'a> Context<'a> {
...
+                    divergent_stmts: &self.divergent_stmts,
+                };
+                loop_control.visit_block(body_ast);
+                // A break or continue exits (or restarts) this loop rather than the
+                // enclosing function, so a body that diverges only via loop control does
+                // not make the loop itself divergent. A body that diverges via return or
+                // fail does. The container is always evaluated, so its divergence counts
+                // unconditionally.
+                converge(Ty::UNIT)
+                    .diverge_if(container.diverges || (body.diverges && !loop_control.any()))
```
ScottCarda-MS: It doesn't seem like `body.diverges && !loop_control.any()` is a valid way to ensure that the body did not contain a `fail` or `return`. A body could have both kinds of divergence in it, and this `body.diverges && !loop_control.any()` would yield `false`. Perhaps `container.diverges` covers this case, but I'm noticing that the code previously checked both `container.diverges` and `body.diverges`, so I assume the difference is meaningful.

source/compiler/qsc/src/interpret/tests.rs
```diff
@@ -3359,4 +3359,1412 @@ mod given_interpreter {
...
+            let mut interpreter = get_interpreter();
+            let (result, output) = line(
+                &mut interpreter,
+                indoc! {"
+                    operation ContinueInBorrowInitializer() : Int {
+                        mutable effects = 0;
+                        for i in 0..2 {
+                            borrow (first, second) = (
+                                Qubit[if i == 1 { continue } else { 1 }],
+                                Qubit[{ effects += 1; 1 }]
```
ScottCarda-MS: If you did `effects += 10 ^ i` instead, we could verify that we skipped the i == 1 case specifically by asserting the result is `101`.

ScottCarda-MS: You could do the same thing with binary if preferred, then assert with binary literal.
`effects += 2 ^ i` -> `&Value::Int(0b101)`

source/compiler/qsc/src/interpret/tests.rs
```diff
@@ -3359,4 +3359,1412 @@ mod given_interpreter {
...
+            let (result, output) = line(
+                &mut interpreter,
+                indoc! {"
+                    operation DescendingRangeBreak() : Int {
+                        mutable total = 0;
+                        for i in 5..-1..0 {
+                            if i == 3 {
+                                break;
+                            }
+                            total = total * 10 + i;
```
ScottCarda-MS, line 3871: Nice strategy! This makes if very obvious which loop iterations had were part of the sum.

source/compiler/qsc/src/interpret/tests.rs
```diff
@@ -3359,4 +3359,1412 @@ mod given_interpreter {
...
+            let (result, output) = line(
+                &mut interpreter,
+                indoc! {"
+                    operation ForArrayBreak() : Int {
+                        mutable total = 0;
+                        for x in [1, 2, 3, 4, 5] {
+                            if x == 3 {
+                                break;
+                            }
+                            total += x;
```
ScottCarda-MS: I liked your earlier strategy of `total += total * 10 + x`. This would make it more obvious which iterations are in effect and make for a more robust test.

source/compiler/qsc/src/interpret/debugger_tests.rs
idavis: Reworked the test to be next_skips_synthetic_block_exits with better validations

source/compiler/qsc/src/codegen/tests.rs
idavis: Added comment to explain and added another test below

source/compiler/qsc/src/codegen/tests.rs
idavis: Split into two tests, fixed and clarified.

source/compiler/qsc_frontend/src/typeck/rules.rs
idavis: Please take another look at the code as there have been updates

source/compiler/qsc_fir_transforms/src/invariants.rs
idavis: the block scan is a correction, not a break/continue special case. See tests `divergent_repeat_fail_body_passes_block_tail` and `entry_point_returning_repeat_with_fail_does_not_crash` which fail on main. But I'll take another look to see if I can see any issues.

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -833,7 +1123,7 @@ impl<'a> Context<'a> {
         let lhs = self.infer_expr(lhs);
         let rhs_span = rhs.span;
         let rhs = self.infer_expr(rhs);
-        let diverges = lhs.diverges || rhs.diverges;
+        let diverges = lhs.diverges || (!matches!(op, BinOp::AndL | BinOp::OrL) && rhs.diverges);
```
ScottCarda-MS: I'm not very clear on why this change is needed. Divergence should still be true if either the left-hand side or the right-hand side diverges. What operators would not cause the expression to diverge if the right-hand side expression diverges?

source/compiler/qsc_passes/src/conjugate_invert.rs
```diff
@@ -233,14 +243,49 @@ impl AssignmentCheck {
 
 struct ReturnCheck {
```
ScottCarda-MS, line 248: We should probably update the name now that it is really doing more than just checking for a `return`. Alternatively, we could just have the control-flow check be a separate visitor. The other checks in this file are all separated into their own very small visitors, which is maybe a bit inefficient, but is ergonomic.

source/compiler/qsc_passes/src/logic_sep.rs
```diff
@@ -104,10 +104,25 @@ impl<'a> Visitor<'a> for SepCheck {
...
+        // `visit_expr` is reached for expressions in operand positions: call
+        // arguments, `if`/`for`/`while` conditions, array and tuple elements,
+        // binary-operator sides, and so on. Statement-position control flow is
+        // rejected in `handle_expr`, but forbidden control flow can also be
+        // buried inside an operand and would otherwise slip past this
+        // separability check. Adjoint generation reverses a separable block, and
+        // `break`/`continue`/`return` cannot be reversed, so accepting one in an
+        // operand could turn an `is Adj` body into a wrong adjoint. Reject it
+        // here as well. This closes a pre-existing gap that affected `return`
+        // just as much as `break`/`continue`.
```
ScottCarda-MS: This comment is quite long and has some historical wording in it, referring to a "pre-existing gap". Could you make this comment more brief?

source/compiler/qsc_fir_transforms/src/invariants.rs
```diff
@@ -712,33 +712,115 @@ fn check_non_unit_block_tail(package: &Package, block_id: BlockId, context: &str
...
-        StmtKind::Local(..) | StmtKind::Item(_) => false,
+    let mut known_bools = inherited_known_bools.clone();
+
+    for &stmt_id in &block.stmts {
+        match &package.get_stmt(stmt_id).kind {
+            StmtKind::Expr(expr_id) | StmtKind::Semi(expr_id) => {
+                if expr_diverges_with_known_bools(package, *expr_id, &known_bools) {
+                    return true;
+                }
+                known_bools.clear();
```
ScottCarda-MS, line 778: I'm not sure why the `know_bools` are cleared after every statement in a block. Seems like that should be something that persists through the block.

source/compiler/qsc_fir_transforms/src/invariants.rs
```diff
@@ -712,33 +712,115 @@ fn check_non_unit_block_tail(package: &Package, block_id: BlockId, context: &str
...
+            }
+            StmtKind::Local(_, pat_id, expr_id) => {
+                if expr_diverges_with_known_bools(package, *expr_id, &known_bools) {
+                    return true;
+                }
+
+                let value = known_bool_value(package, *expr_id, &known_bools);
+                let pat = package.get_pat(*pat_id);
+                if let (PatKind::Bind(ident), Ty::Prim(Prim::Bool), Some(value)) =
```
ScottCarda-MS, line 787: Should we also check for boolean bindings in tuple bindings?
i.e. `let (x, y) = (4, true);`

source/compiler/qsc_fir_transforms/src/invariants.rs
idavis: The comment at the top of of the function "Any other evaluated statement clears the facts because it may mutate a tracked local" is the intended behavior, otherwise we'd need per-local invalidation on every assignment which is a ton of work.

source/compiler/qsc_fir_transforms/src/invariants.rs
idavis: No. The target pattern can't be a tuple. We aren't building a general constant-propagation table. We are just reconstructing a guaranteed first iteration fact that loop unification uses.

source/compiler/qsc_passes/src/logic_sep.rs
idavis: cleaned up

source/compiler/qsc_frontend/src/typeck/rules.rs
idavis: Added a comment above it explaining

source/compiler/qsc_passes/src/loop_normalize.rs
```diff
@@ -0,0 +1,835 @@
...
+    ///
+    /// A divergent `break`/`continue`/`return` produces no value, so only its
+    /// surrounding type is adjusted; wrapping it as an array element would
+    /// re-bury it in an operand position and re-trigger the lift. Any other leaf
+    /// is a genuine `elem_ty`-valued result and is replaced with the
+    /// single-element array `[ <leaf> ]`. A `While`/`For`/`Repeat` is always
+    /// `Unit`-typed, and `Unit` has a classical default, so a loop never takes
+    /// the array-backed path and never reaches this helper. Mirrors the
+    /// array-backing of the `return_unify` FIR transform.
+    fn arrayify_value_in_place(&mut self, expr: &mut Expr, elem_ty: &Ty) {
```
ScottCarda-MS: This function, and the next two related ones, all pass in an `elem_ty`. I think we should skip passing that in and just grab it off of the passed in expr. That will make sure that any passed-in expression is always getting work on with its own type. The logic should produce equivalent results, but is a little more careful about where types come from in case that ever diverges in the future, and saves us a little complexity in having to keep track of a passed-around `elem_ty`.

source/compiler/qsc_passes/src/loop_normalize.rs
```diff
@@ -0,0 +1,835 @@
...
+    /// instead, and the desugar relocates the guarded `.operand_tmp_<id>[0]` read
+    /// into the fall-through branch, so the binding never needs a default of its
+    /// own type. A defaultable binding, a binding whose initializer carries no
+    /// escaping control flow, and an unrepresentable type are all left
+    /// untouched.
+    fn array_back_control_flow_local(&mut self, stmt: &mut Stmt) -> Option<Stmt> {
+        let StmtKind::Local(_, _, init) = &mut stmt.kind else {
+            return None;
+        };
+        if is_defaultable(&init.ty) || !contains_control_flow(init) || !is_representable(&init.ty) {
```
ScottCarda-MS: nit: `contains_control_flow` is the more expensive check of the three, so there is a slight motivation for it to be last, as you wrote it in `array_back_discarded_control_flow_value`.

source/compiler/qsc_passes/src/loop_normalize.rs
```diff
@@ -0,0 +1,835 @@
...
+/// Foo(.operand_tmp_<id>);
+/// ```
+pub(super) struct LoopNormalize<'a> {
+    pub(super) assigner: &'a mut Assigner,
+    pub(super) errors: Vec<Error>,
+}
+
+/// Which short-circuit operator form is being reshaped into an `If`, carrying
+/// whether the operator is `and` (`true`) or `or` (`false`).
+enum ShortCircuitForm {
```
ScottCarda-MS: It looks like this is only used in `rewrite_short_circuit_rhs_in_place`. Can this be made local to that function?

source/compiler/qsc_frontend/src/typeck/rules.rs
ScottCarda-MS: You rewrote this accordingly, and it is even cleaner now than I was expecting. Nice!

ScottCarda-MS (commented):
Wrapping up review of the loop normalization pass

source/compiler/qsc_frontend/src/typeck/tests.rs
```diff
@@ -1720,6 +2239,172 @@ fn while_body_should_be_unit_error() {
...
+        "namespace Test { function First<'T>() : 'T { for _ in [1] { break } } }",
+        "",
+        false,
+    );
+
+    assert_eq!(
+        errors.len(),
+        1,
+        "expected a return type mismatch: {errors:?}"
+    );
```
swernli: for this and other new tests in this file, why have some verify just the count of errors rather than the content using `check` and `expect` like the rest of the file? That would ensure the types of the expressions, the existence of the error, and that the error was the expected type mismatch.

source/compiler/qsc_passes/src/loop_unification.rs
```diff
@@ -472,3 +922,577 @@ fn gen_id_add_update(assigner: &mut Assigner, ident: &IdentTemplate, expr: Expr)
...
+        id: assigner.next_node(),
+        span: Span::default(),
+        kind: StmtKind::Expr(expr),
+    }
+}
+
+/// Builds a classical default value of `ty`, recording an
+/// [`Error::UnsupportedType`] and substituting a typed `Err` placeholder when
+/// `ty` has no synthesizable default.
+fn build_default_or_err(
```
ScottCarda-MS, line 1301: It seems like the content regarding type defaultable/representable should live in a common location and shouldn't be owned by the loop unification/normalization. Looks like there might be some overlap between what the loop normalization and loop unification are doing also. Determining if a type is defaultable or representable and getting the default values for those types that are defaultable should be common utilities.

source/compiler/qsc_frontend/src/typeck/rules.rs
```diff
@@ -593,21 +902,44 @@ impl<'a> Context<'a> {
...
+                    divergent_nodes: &self.divergent_nodes,
+                };
+                loop_control.visit_block(body_ast);
+                // A break or continue exits (or restarts) this loop rather than the
+                // enclosing function, so a body that diverges only via loop control does
+                // not make the loop itself divergent. A body that diverges via return or
+                // fail does only when the condition is statically true. The condition
+                // itself is always evaluated, so its divergence counts unconditionally.
+                let body_diverges =
+                    bool_literal(cond_ast) == Some(true) && body.diverges && !loop_control.any();
```
swernli: it makes sense that this fixes cases where the loop body divergence is known to be either false or true:
```qsharp
// Does diverge, was previously correctly detected as divergent.
while true { fail "always happens" }
// Doesn't diverge, was previously incorrectly detected as divergent.
while false { fail "never happens" }
```
But isn't this fix trivially defeated by any intervening binding of the condition? In that case the fix would effectively swap where the bug is:
```qsharp
// Does diverge, now incorrectly detected as not divergent.
let a = true;
while a { fail "always happens" }

// Doesn't diverge, now correctly detected as not divergent.
let b = false;
while b { fail "never happens" }
```

swernli: Hmmm... playing around with this some, it seems like the new behavior is more "fail safe" than the prior one. With the new behavior, some valid programs are rejected because the types are not considered divergent but invalid programs that would previously cause a panic at runtime are rejected by the type system as they are no longer considered divergent. So while the divergence issue is not fully fixed, it's definitely _better_ than it was before.

source/compiler/qsc_frontend/src/typeck/tests/bounded_polymorphism.rs
```diff
@@ -162,7 +162,7 @@ fn iter() {
         &expect![[r##"
             #11 67-74 "(a: 'T)" : Param<"'T": 0>
             #12 68-73 "a: 'T" : Param<"'T": 0>
-            #19 82-180 "{\n                for item in a {\n                    return item;\n                }\n            }" : Bool
+            #19 82-180 "{\n                for item in a {\n                    return item;\n                }\n            }" : Unit
```
ScottCarda-MS: I suppose this is a change in output for an already-erroneous case, but it is strange to me that this type changed. Is this due to the bug fix regarding return expressions?

source/compiler/qsc_passes/src/loop_control/tests.rs
```diff
@@ -0,0 +1,519 @@
...
+            }
+        "},
+        &expect![[r#"
+            []
+        "#]],
+    );
+}
+
+#[test]
+fn break_in_inner_while_condition_within_outer_loop_errors() {
```
ScottCarda-MS, line 432: This is an unexpected result to me. Why can't the `break` in the condition bind to the outer loop? In `break_in_for_iterable_without_enclosing_loop_errors` we allowed a `break` in the header of the `for` loop, which seems similar.

source/compiler/qsc_passes/src/loop_unification/tests.rs
```diff
@@ -46,53 +21,19 @@ fn convert_for_array() {
...
-                                                Pat 13 [97-98] [Type String]: Bind: Ident 14 [97-98] "x"
-                                                Expr 15 [101-114] [Type String]: String:
-                                                    Lit: "Hello World"
-                                            Stmt 35 [75-78]: Semi: Expr 36 [75-78] [Type Unit]: AssignOp (Add):
-                                                Expr 37 [75-78] [Type Int]: Var: Local 26
-                                                Expr 34 [75-78] [Type Int]: Lit: Int(1)
-                        adj: <none>
-                        ctl: <none>
-                        ctl-adj: <none>"#]],
+            operation Main(arr : Int[]) : Unit {
```
ScottCarda-MS, line 24: This is a big improvement for readability. Thanks!

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+        &expect![[r#"
+            operation Main() : Unit {
+                mutable total = 0;
+                {
+                    let _range_id_45 = 0..4;
+                    mutable _index_id_48 = _range_id_45::Start;
+                    let _step_id_53 = _range_id_45::Step;
+                    let _end_id_58 = _range_id_45::End;
+                    mutable _broke_31 = false;
+                    while not _broke_31 and _step_id_53 > 0 and _index_id_48 <= _end_id_58 or _step_id_53 < 0 and _index_id_48 >= _end_id_58 {
```
ScottCarda-MS: I think this test case reveals an issue. Because in Q# `and` binds tighter than `or` this will attach the `not _broke_31` only to the positive step case. The negative step case, when true, will bypass the `not _broke_31` check.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+                    let _step_id_82 = _range_id_74::Step;
+                    let _end_id_87 = _range_id_74::End;
+                    mutable _broke_40 = false;
+                    while not _broke_40 and _step_id_82 > 0 and _index_id_77 <= _end_id_87 or _step_id_82 < 0 and _index_id_77 >= _end_id_87 {
+                        mutable _cont_44 = false;
+                        let i = _index_id_77;
+                        if i == 3 {
+                            _broke_40 = true;
+                        }
+                        if not _broke_40 and not _cont_44 {
```
ScottCarda-MS: Would it be simpler to generate:
```qsharp
if not _broke_40{
    if i == 1 {
        _cont_44 = true;
    }
    if not _cont_44 {
        total += i;
    }
    _index_id_77 += _step_id_82;
}
```

In algorithmic terms: Everything after the `break`'s conditional-if goes in one big flag guard, including the generated loop-index increment. Everything after the `continue`'s conditional-if goes in one big flag guard, _except_ the generated loop-index increment.

ScottCarda-MS: Though if the continue and brake were swapped in source, that may cause an issue with the generated loop-index. Okay so the generated loop-index needs its own independent conditional, fair enough. But I still would ask about putting all the other source content in one big flag-guard per control-flow.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+                        }
+                    }
+                }
+            }
+        "#]],
+    );
+}
+
+#[test]
+fn convert_while_with_continue() {
```
ScottCarda-MS, line 383: This test should have something after the continue that would be skipped.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+                        _cont_29 = true;
+                    }
+                }
+            }
+        "#]],
+    );
+}
+
+#[test]
+fn convert_while_with_break_and_continue() {
```
ScottCarda-MS: Same here. This needs something after the continue. Though I supposed it is good to test, at least once, the case where there is nothing to skip to make sure the related flag-guard doesn't generate.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+                        }
+                    }
+                };
+            }
+        "#]],
+    );
+}
+
+#[test]
+fn convert_repeat_until_with_continue() {
```
ScottCarda-MS, line 577: Same here. Need's skippable content after continue.

ScottCarda-MS: A couple of more tests after this too. I'll stop posting new comments about this.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+    // synthesized default must be the fully-bounded `0..0`, matching the `Range`
+    // type tag, not the `RangeFull` shape `...`.
+    check(
+        indoc! {r#"
+        namespace test {
+            operation Main() : Int {
+                mutable acc = 0;
+                for i in 0..4 {
+                    let r = if i == 2 { break } else { 0..i };
+                    acc += r::End;
```
ScottCarda-MS, line 1111: We should avoid using `::` as it is deprecated syntax. Instead this should be `r.End`. The generated code for the range Start, Step, and End should also be changed, though that is a pre-existing issue that can be done in a follow-up if needed. For now, just avoid source code using `::`, like here.

ScottCarda-MS: I believe we have a lint for this that is off by default. It's probably time we turned that on at least for our own code. Again, that could be a follow up PR for the generated content.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+    "#});
+
+    assert!(
+        errors.iter().any(|error| matches!(
+            error,
+            crate::Error::LoopControl(crate::loop_control::Error::InFixup(_))
+        )),
+        "expected a loop_control InFixup diagnostic, got {errors:?}"
+    );
+}
```
ScottCarda-MS, line 1919: Would be good to see tests that have to deal with multiple breaks and/or continues for the same loop to see how their individual flags compose in the flag-guards.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
ScottCarda-MS: So it would look something like this then:
```qsharp
if not _broke_40 {
    if i == 1 {
        _cont_44 = true;
    }
    if not _cont_44 {
        total += i;
    }
}
if not _broke_40 {
    _index_id_77 += _step_id_82;
}
```
I guess I'm just reacting to the apparent awkwardness of the `not _cont_44` in the condition around the expression that calculates the `_cont_44`'s value.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
```diff
@@ -0,0 +1,1614 @@
...
+#[test]
+fn convert_for_range_with_break_and_continue() {
+    check(
+        indoc! {r#"
+        namespace test {
+            operation Main() : Unit {
+                mutable total = 0;
+                for i in 0..4 {
+                    if i == 3 {
+                        break;
```
ScottCarda-MS, line 173: What happens if there is content after a `break` or `continue` inside their conditional blocks? Those statements _should_ be unreachable, but I've not seen a test for how that would generate the transformed Q#.
Something like:
```qsharp
for i in 0..4 {
    if i == 3 {
        break;
        let temp = 5; // or maybe unreachable nested control-flow expressions
    }
}
```
A test would make sure that the `let temp = 5;` does NOT generate:
```qsharp
if i == 3 {
    _broke_40 = true;
    let temp = 5;
}
```
but instead is generates:
```qsharp
if i == 3 {
    _broke_40 = true;
}
if not _broke_40 {
    let temp = 5;
}
```

source/compiler/qsc_passes/src/loop_normalize/tests.rs
```diff
@@ -0,0 +1,1480 @@
...
+            operation Main() : Unit {
+                mutable cond = true;
+                while cond {
+                    let _operand_tmp_43 = Bar;
+                    let _operand_tmp_47 = if cond {
+                        break
+                    } else {
+                        3
+                    };
+                    Foo(_operand_tmp_43(_operand_tmp_47));
```
ScottCarda-MS: This seems like a problem with the order of evaluation. `Foo` should be evaluated before `Bar` and `Bar`'s arguments.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
```diff
@@ -0,0 +1,1480 @@
...
+                    };
+                    _operand_tmp_38(_operand_tmp_42[0]);
+                }
+            }
+        "#]],
+    );
+}
+
+#[test]
+fn hoist_break_in_arrow_operand_block_array_backed() {
```
ScottCarda-MS, line 1670: This is an interesting case. Because there is no default value for that type, loop unification would fail, correct? Does this test exercise a case that passes normalization but will fail unification?

ScottCarda-MS: Oh never mind. I forgot that was the whole point of the arrayify-ing of the value.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
```diff
@@ -0,0 +1,1480 @@
...
+            }
+        "},
+        &expect![[r#"
+            operation Foo(q : Qubit) : Unit is Ctl {}
+            operation Main() : Unit {
+                mutable cond = true;
+                while cond {
+                    let _operand_tmp_25 = Controlled Foo;
+                    let _operand_tmp_29 = break;
+                    _operand_tmp_25(_operand_tmp_29[0]::Item < 0 >, _operand_tmp_29[0]::Item < 1 >);
```
ScottCarda-MS, line 1798: Something went wrong here. It looks like it tried to do syntax like `base::Item<0>` but that is not valid syntax in Q#.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
```diff
@@ -0,0 +1,1480 @@
...
+        "},
+        &expect![[r#"
+            operation Foo(x : Int) : Unit {}
+            operation Main() : Unit {
+                let arr = [1, 2, 3];
+                mutable cond = true;
+                while cond {
+                    let _operand_tmp_33 = arr;
+                    let _operand_tmp_37 = break;
+                    Foo(_operand_tmp_33[_operand_tmp_37]);
```
ScottCarda-MS, line 2166: Here also, the `Foo` expression (that resolves a reference to the callable) needs to be evaluated first before the arguments of the call expression.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
```diff
@@ -0,0 +1,1480 @@
...
+                let arr = [1, 2, 3];
+                mutable cond = true;
+                while cond {
+                    let _operand_tmp_65 = {
+                        if cond {
+                            break
+                        };
+                        arr
+                    };
+                    for j in F(_operand_tmp_65, G()) {
```
ScottCarda-MS, line 2539: There's a similar issue here where `F` should be hoisted also to ensure that it gets evaluated before its arguments. It was interesting to see that `G()` did not get hoisted, which is correct given that you don't hoist args after the last break/cont-infected arg.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
idavis: @ScottCarda-MS - I can update the rendering of the generated Q# to wrap in parenthesis to show how they are represented. We loose some of the semantics in the current rendering.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
idavis: Nothing went wrong. We don't have a way to display tuple field access. So this generates something that shows that it is trying to get item 0 and item 1 from the tuple. Here are two visualizations of current and using a binding. Neither is great. We use the first one in the FIR Q# codegen already.
```qsharp
_operand_tmp_39(_operand_tmp_43[0]::Item < 0 >, _operand_tmp_43[0]::Item < 1 >);
```
vs
```qsharp
_operand_tmp_39({
    let (_field_46, _) = _operand_tmp_43[0];
        _field_46
}, {
    let (_, _field_50) = _operand_tmp_43[0];
    _field_50
});
```

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
idavis: I forced parens to clearly show precedence and ordering

source/compiler/qsc_passes/src/loop_control/tests.rs
idavis: It is nonsense. So rather than special case allowing it nested loops, it is easier to just disallow it.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
ScottCarda-MS: I think I understand. Q# doesn't have tuple field access syntax, but we need a quick way of seeing that field access for tests, so FIR to Q# generation actually generates a language that is Q# + tuple-field-access. Is that accurate to what's going on here?
If that is the case, will this cause issues for any non-test process that is trying to generate Q# from FIR?

source/compiler/qsc_passes/src/loop_unification.rs
ScottCarda-MS: This was done. Thanks, I think the is much cleaner.

source/compiler/qsc_passes/src/qsharp_gen.rs
```diff
@@ -0,0 +1,906 @@
...
+                }
+                self.write(")");
+            }
+            ExprKind::UnOp(op, expr) => {
+                let op_str = unop_as_str(*op);
+                if matches!(op, UnOp::Functor(_)) {
+                    self.write(op_str);
+                    self.emit_expr(expr);
+                } else {
+                    self.write("(");
```
ScottCarda-MS: I don't think you need to wrap unary operations in parens like this.

source/compiler/qsc_passes/src/loop_unification/break_continue_tests.rs
ScottCarda-MS: It wraps every statement after the first break, so the logic is seems valid.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
idavis: Yes, there shouldn't be any usage outside of tests, I had to create the codegen for the tests. After AST you can no longer be expected to generate valid Q#.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
idavis: We know `F` would resolve to an immutable binding to the item which doesn't require further eval, so we can leave it as-is in this case.

source/compiler/qsc_passes/src/loop_normalize/tests.rs
idavis: We know Fooo would resolve to an immutable binding to the item which doesn't require further eval, so we can leave it as-is in this case.

ScottCarda-MS (approved):

PR #3453, opened 2026-07-10
Validate error code names

amcasey:
They should have at least two dot-delimited segments and the first should be from a known list.

Add the test to the LS since it seems like it would have to depend on all the packages declaring error codes anyway.

### DIFF
--- a/source/language_service/tests/diagnostic_codes.rs
+++ b/source/language_service/tests/diagnostic_codes.rs
@@ -4,7 +4,7 @@
 //! This test validates that all `#[diagnostic(code("..."))]` attributes in the
 //! source tree follow the naming convention: the first segment must be a known
 //! prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`), followed by one or more
-//! PascalCase segments containing only ASCII alphanumeric characters.
+//! `PascalCase` segments containing only ASCII alphanumeric characters.
 //!
 //! For example: `Qsc.Resolve.NotFound`, `Qasm.Lowerer.CannotCast`.
 
@@ -13,9 +13,8 @@ use std::path::{Path, PathBuf};
 
 /// Recursively collect all `.rs` files under `dir`.
 fn collect_rs_files(dir: &Path, out: &mut Vec<PathBuf>) {
-    let entries = match fs::read_dir(dir) {
-        Ok(entries) => entries,
-        Err(_) => return,
+    let Ok(entries) = fs::read_dir(dir) else {
+        return;
     };
     for entry in entries.flatten() {
         let path = entry.path();
--- a/source/language_service/tests/diagnostic_codes.rs
+++ b/source/language_service/tests/diagnostic_codes.rs
@@ -2,16 +2,16 @@
 // Licensed under the MIT License.
 
 //! This test validates that all `#[diagnostic(code("..."))]` attributes in the
-//! source tree follow the naming convention: the first segment must be a known
-//! prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`), followed by one or more
-//! `PascalCase` segments containing only ASCII alphanumeric characters.
+//! source tree follow the naming convention: the first segment must be "Qdk",
+//! the second must be a known value (`Qsc`, `Qasm`, `Stim`, or `Qre`), and the
+//! remainder must be `PascalCase` segments containing only ASCII alphanumeric
+//! characters.
 //!
-//! For example: `Qsc.Resolve.NotFound`, `Qasm.Lowerer.CannotCast`.
+//! For example: `Qdk.Qsc.Resolve.NotFound`, `Qdk.Qasm.Lowerer.CannotCast`.
 
 use std::fs;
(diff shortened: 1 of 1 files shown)


source/language_service/tests/diagnostic_codes.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Check that a diagnostic code string matches the convention:
+/// two or more dot-separated segments where the first segment is a known
+/// prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`) and subsequent segments each
+/// start with an uppercase ASCII letter and contain only ASCII alphanumeric
+/// characters.
+fn is_valid_diagnostic_code(code: &str) -> bool {
+    let parts: Vec<&str> = code.split('.').collect();
+    parts.len() >= 2
+        && matches!(parts[0], "Qsc" | "Qasm" | "Stim" | "Qre")
```
orpuente-MS: Should it be `matches!(parts[0], "Qdk") && matches!(parts[1], "Qsc" | "Qasm" | "Stim" | "Qre")`?

ScottCarda-MS (approved):
Interesting analysis of Rust source files to check for invalid error codes.

source/language_service/tests/diagnostic_codes.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! This test validates that all `#[diagnostic(code("..."))]` attributes in the
+//! source tree follow the naming convention: the first segment must be a known
+//! prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`), followed by one or more
```
joao-boechat: The code seems to have different behavior from this (which is the correct behavior)
The first segment needs to be "Qdk", then the following ones are the ones listed here.

source/language_service/tests/diagnostic_codes.rs
```diff
@@ -0,0 +1,124 @@
...
+            collect_rs_files(&path, out);
+        } else if path.extension().is_some_and(|ext| ext == "rs") {
+            out.push(path);
+        }
+    }
+}
+
+/// Check that a diagnostic code string matches the convention:
+/// two or more dot-separated segments where the first segment is a known
+/// prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`) and subsequent segments each
```
joao-boechat: Same as comment above

source/language_service/tests/diagnostic_codes.rs
```diff
@@ -0,0 +1,124 @@
...
+//! prefix (`Qsc`, `Qasm`, `Stim`, or `Qre`), followed by one or more
+//! `PascalCase` segments containing only ASCII alphanumeric characters.
+//!
+//! For example: `Qsc.Resolve.NotFound`, `Qasm.Lowerer.CannotCast`.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+
+/// Recursively collect all `.rs` files under `dir`.
+fn collect_rs_files(dir: &Path, out: &mut Vec<PathBuf>) {
```
joao-boechat: nit: this applies to all comments in the file (except for the top one), but I don't see a point having comments describing what the code does when the code is already descriptive enough and straightforward. I feel like we should only resort to comments when we can't make the code more expressive of our intentions, otherwise it can get bloated with comments (which can also get out of date pretty fast).

source/language_service/tests/diagnostic_codes.rs
```diff
@@ -0,0 +1,124 @@
...
+                let relative = path.strip_prefix(&source_dir).unwrap_or(path).display();
+                failures.push(format!("  {relative}:{line}: \"{code}\""));
+            }
+        }
+    }
+
+    assert!(
+        failures.is_empty(),
+        "The following diagnostic codes do not follow the naming convention \
+         (expected dot-separated PascalCase segments, e.g. \"Qsc.Resolve.NotFound\"):\n{}",
```
joao-boechat:
```suggestion
         (expected dot-separated PascalCase segments, e.g. \"Qdk.Qsc.Resolve.NotFound\"):\n{}",
```

joao-boechat (approved):
Thanks for adding these tests, they look great! Just a couple minor suggestions

source/language_service/tests/diagnostic_codes.rs
amcasey: Updated header

source/language_service/tests/diagnostic_codes.rs
amcasey: Removed this comment, since it was redundant

source/language_service/tests/diagnostic_codes.rs
amcasey: I don't feel strongly either way, but I've removed most of the comments.

source/language_service/tests/diagnostic_codes.rs
amcasey: Done

PR #3454, opened 2026-07-11
Update prepare/select scoping rules in stim

joao-boechat:
This PR implements the following scoping rules for the PREPARE blocks in stim:

1- An outer scope is allowed to reach into an inner one
2- An inner scope isn't allowed to reach into an outer one
3- The only blocks that have scoping rules are the prepare blocks

So, for example:

This should work:
```
PREPARE {
     M 0
}
    REQUIRE rec[-1]
```
    
   
But this should not:
```
  M 0
  PREPARE {
      REQUIRE rec[-1]
  }
```

Nor this:
```
M 0
PREPARE {
  CX rec[-1] 1
}
```

On top of that, it also renames the PREPARE block to SELECT.

### DIFF
--- a/source/compiler/stim_compiler/src/qir/tests/measurement_record_targets.rs
+++ b/source/compiler/stim_compiler/src/qir/tests/measurement_record_targets.rs
@@ -1079,10 +1079,10 @@ fn ycz_with_negated_rec_on_second_target_yields_error() {
 }
 
 #[test]
-fn cx_with_rec_control_crossing_prepare_boundary_yields_error() {
+fn cx_with_rec_control_crossing_select_boundary_yields_error() {
     let source = indoc! {"
         M 0
-        PREPARE {
+        SELECT {
           CX rec[-1] 1
         }
     "};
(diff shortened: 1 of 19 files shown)


source/compiler/stim_compiler/src/qir/tests/measurement_record_targets.rs
```diff
@@ -1011,3 +1011,57 @@ fn cx_with_rec_control_crossing_prepare_boundary_yields_error() {
         "#]],
     );
 }
+
+#[test]
+fn top_level_classical_control_reaches_into_prepare() {
+    let source = "PREPARE {\n    M 0\n}\nCX rec[-1] 1";
```
orpuente-MS: Should we use actual newlines instead of `\n` chars to keep it consistent with the other tests (and easier to read)?

joao-boechat: Good catch, I've pushed that fix.

orpuente-MS (approved):

PR #3455, opened 2026-07-11
Support negated targets in stim collapsing gates

joao-boechat:
This PR implements support for the following operations:

`M !0`, `MX !0`, `MY !0`,
`MR !0`, `MRX !0`, `MRY !0`, 
`MZZ !0 1`, `MXX !0 1`, `MYY !0 1`, 

where M, MZ; MR, MRZ are interchangeable,
and on pair measurements we can negate either target, or both, but the negation will invert the result by the XOR of both negations. For example:

`MZZ !0 1` = `MZZ 0 !1`
`MZZ !0 !1` = `MZZ 0 1`

In terms of the implementation, all that had to be done was adding an X gate before and after the emitted M operations (all collapsing gates are eventually decomposed to a single M in a qubit). For measurement reset operations, we only add the X gate before the measurement (as we don't care about post-measurement state).

### DIFF
--- a/source/compiler/stim_compiler/src/qir.rs
+++ b/source/compiler/stim_compiler/src/qir.rs
@@ -332,9 +332,9 @@ pub enum Error {
         #[label]
         span: Span,
     },
-    #[error("measurement record control cannot be negated in instruction: {instruction}")]
-    #[diagnostic(code("Qdk.Stim.Compiler.NegatedMeasurementRecord"))]
-    NegatedMeasurementRecord {
+    #[error("target cannot be negated in instruction: {instruction}")]
+    #[diagnostic(code("Qdk.Stim.Compiler.NegatedTarget"))]
+    NegatedTarget {
         instruction: String,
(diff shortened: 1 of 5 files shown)


fedimser (approved):

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1384,20 +1437,29 @@ impl<'noise> Compiler<'noise> {
...
     }
 
-    fn expect_qubit(&mut self, instruction: &Instruction, target: &Target) -> Option<u32> {
+    fn expect_qubit(
+        &mut self,
+        instruction: &Instruction,
+        target: &Target,
+        allow_negated: bool,
+    ) -> Option<(u32, bool)> {
         // TODO: lacks support for negated qubits and pauli targets
```
orpuente-MS: Should you remove this TODO now?

orpuente-MS (approved):

PR #3457, opened 2026-07-11
Fix Typo in "|0〉, |+〉 or Inconclusive?" Kata

Advay17:
I was just doing the Katas and noticed this minor typo, |1〉 is not one of the possible states.


amcasey (approved):
Thanks for the fix, @Advay17!

PR #3465, opened 2026-07-14
Paulimer upgrade

orpuente-MS:
This PR deletes our vendored version of paulimer and points to the latest version in crates.io instead.


Cargo.toml
```diff
@@ -58,7 +59,9 @@ futures = "0.3"
 futures-util = "0.3"
 fasteval = "0.2"
 getrandom = "0.4"
-iai-callgrind = "0.16.1"
+# Keep the rand 0.8 transitive dependency feature-addressable for WASM builds.
+getrandom_02 = { package = "getrandom", version = "0.2" }
+gungraun = "0.19.4"
```
billti, line 64: What made you choose this crate? Looks pretty recent, with not many stars, and the one contributor has his GitHub activity set to private. What's your confidence level in this being a high quality well maintained durable crate?

Cargo.toml
```diff
@@ -58,7 +59,9 @@ futures = "0.3"
 futures-util = "0.3"
 fasteval = "0.2"
 getrandom = "0.4"
-iai-callgrind = "0.16.1"
+# Keep the rand 0.8 transitive dependency feature-addressable for WASM builds.
+getrandom_02 = { package = "getrandom", version = "0.2" }
```
billti, line 63: It's annoying to have to do this "dual version of rand" hack. When I bumped rand I fixed up the vendored paulimer for it (https://github.com/microsoft/qdk/pull/3239). Is it not possible for Paulimer to move forward? 0.10 has been out a while now.

Cargo.toml
orpuente-MS: This crate was previously called iai-callgrind. It is one of the crates suggested by the rust performance book to do benchmarking and profiling [Rust Performance Book](https://nnethercote.github.io/perf-book/benchmarking.html).

I first learned about it when reading the above book, and then heard it recommended in this talk [Towards Impeccable Rust](https://youtu.be/qfknfCsICUM?si=RLy9cpMao1-leahx&t=1230) in minute 20:30.

That's has been my default library for running benchmarking experiments for the last 2 years, since it gives me deterministic results across runs.

I personally don't use criterion, it's way too noisy, and also slow since it needs to run hundreds of trials to achieve statistical significance; and even with that the results can vary wildly depending on what my OS is doing in the background.

orpuente-MS: Also, the rust compiler uses it: [Rust compiler Cargo.toml](https://github.com/rust-lang/rust/blob/df1ae0f7dd168513de2d89beb2cb7a95d4fdc118/library/compiler-builtins/Cargo.toml#L49).

Cargo.toml
orpuente-MS: Upgrading paulimer's rand would be a temporary fix. Once we differ in versions again, we will run into the same issue. A permanent fix would be adding a transitive `"js"` feature to paulimer, which would enable the `"js"` feature of the rand crate. That way even if we differ in versions, we wouldn't need this hack.

That been said. I don't know if:
1. It is justifiable for paulimer to release a new version in crates.io with just a Cargo.toml change.
2. If we are willing to wait until the next public release of paulimer to merge this PR.

joao-boechat (approved):

PR #3467, opened 2026-07-14
Add Python type annotations

fedimser:
This PR:
* Requires that all Python functions are annotated with types.
* Removes all pyright exclusions except `qdk/qre` (QRE will be annotated in another PR, it currently has 45 errors).
* Adds all missing type annotations.
* Adds missing imports to `check_requirements.txt` (azure-quantum, cirq-core, qiskit).

### DIFF
--- a/source/qdk_package/qdk/_adaptive_pass.py
+++ b/source/qdk_package/qdk/_adaptive_pass.py
@@ -244,13 +244,23 @@ class Reg:
     val: int  # index in the registers table
 
 
-def is_immediate(arg) -> bool:
+OperandLike: TypeAlias = int | IntOperand | FloatOperand | Reg
+
+
+def is_immediate(arg: OperandLike | None) -> bool:
     return isinstance(arg, (IntOperand, FloatOperand))
 
 
 def prepare_immediate_flags(
-    *, dst=None, src0=None, src1=None, aux0=None, aux1=None, aux2=None, aux3=None
-):
+    *,
+    dst: OperandLike | None = None,
+    src0: OperandLike | None = None,
+    src1: OperandLike | None = None,
+    aux0: OperandLike | None = None,
+    aux1: OperandLike | None = None,
+    aux2: OperandLike | None = None,
+    aux3: OperandLike | None = None,
+) -> int:
     flags = 0
     if is_immediate(dst):
         flags |= FLAG_DST_IMM
@@ -270,7 +280,13 @@ def prepare_immediate_flags(
 
 
 def unwrap_operands(
-    dst, src0, src1, aux0, aux1, aux2, aux3
+    dst: OperandLike,
+    src0: OperandLike,
+    src1: OperandLike,
+    aux0: OperandLike,
+    aux1: OperandLike,
+    aux2: OperandLike,
+    aux3: OperandLike,
 ) -> Tuple[int, int, int, int, int, int, int]:
     if not isinstance(dst, int):
         dst = dst.val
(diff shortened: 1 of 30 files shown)


source/qdk_package/qdk/cirq/_neutral_atom.py
```diff
@@ -112,11 +114,7 @@ def run_sweep(
         :return: A list of :class:`NeutralAtomCirqResult` objects, one per resolver.
         :rtype: List[NeutralAtomCirqResult]
         """
-        resolvers = (
-            list(cirq.to_sweep(params))
-            if params is not None
-            else [cirq.ParamResolver()]
-        )
+        resolvers = list(cirq.to_resolvers(params))
```
fedimser, line 117: I checked, this is equivalent to what was here before.

source/qdk_package/qdk/qiskit/backends/backend_base.py
```diff
@@ -199,12 +200,13 @@ def __init__(
     def _build_target(self) -> Target:
         supports_barrier = self._qiskit_pass_options["supports_barrier"]
         supports_delay = self._qiskit_pass_options["supports_delay"]
+        assert self.options is not None
 
         # explicitly set ``num_qubits`` to ``None`` to indicate a :class:`Target` representing a
         # simulator or other abstract machine that imposes no limits on the number of qubits.
         return QirTarget.build_target(
             num_qubits=None,
-            target_profile=self._options["target_profile"],
```
joao-boechat, line 207: is it ok to remove the _ from _options here?

source/qdk_package/qdk/simulation/_simulation.py
```diff
@@ -64,9 +64,11 @@ def run(self, mod: pyqir.Module) -> Tuple[List[QirInstruction | Tuple], int, int
         self.required_num_results = pyqir.required_num_results(func)
 
         super().run(mod)
+        assert self.required_num_qubits is not None
+        assert self.required_num_results is not None
         return (self.gates, self.required_num_qubits, self.required_num_results)
 
-    def _on_block(self, block):
+    def _on_block(self, block: pyqir.BasicBlock) -> None:
```
joao-boechat: Can you follow the standard from other files and import BasicBlock from pyqir instead of referencing pyqir every time?

source/qdk_package/qdk/_adaptive_pass.py
```diff
@@ -244,13 +244,23 @@ class Reg:
...
+
+def is_immediate(arg: OperandLike | None) -> bool:
     return isinstance(arg, (IntOperand, FloatOperand))
 
 
 def prepare_immediate_flags(
-    *, dst=None, src0=None, src1=None, aux0=None, aux1=None, aux2=None, aux3=None
-):
+    *,
+    dst: OperandLike | None = None,
```
joao-boechat: Could you use Optional[OperandLike] here?

joao-boechat (approved):

source/qdk_package/qdk/qiskit/backends/backend_base.py
fedimser: Yes, because this is how BackendV2.options is defined in qiskit:

```
    @property
    def options(self):
        return self._options
```

But this is not required for type checking, so I removed this change (kept assert not None).

source/qdk_package/qdk/simulation/_simulation.py
fedimser: Done

PR #3470, opened 2026-07-15
Handle cross-package callable fields during QIR codegen

idavis:
Summary

- Fixes defunctionalization of callable fields flowing through constructors and factories, including across package boundaries.
- Associates FIR transform diagnostics with their owning package so errors point to the correct source file.
- Adds Rust, language service, and Python regression coverage for adaptive QIR generation and cross-package callables.


source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -143,6 +143,72 @@ fn compile_source_to_qir_with_library_result(
...
+        .flatten()
+        .next()
+        .expect("diagnostic should have a source label");
+    let (source, relative_span) = error.resolve_span(label.inner());
+    let span_start = relative_span.offset();
+    let span_end = span_start + relative_span.len();
+
+    assert_eq!(source.name.as_ref(), "lib.qs");
+    assert_eq!(&source.contents[span_start..span_end], "config.Apply(q)");
+    assert_ne!(source.name.as_ref(), "OutOfBounds");
```
swernli, line 209: This check is technically redudant, since the name is already verified to be equal to "lib.qs" above.

source/compiler/qsc/src/compile.rs
```diff
@@ -14,6 +14,24 @@ use thiserror::Error;
...
+) -> WithSource<qsc_fir_transforms::PipelineError> {
+    let package_id = qsc_lowerer::map_fir_package_to_hir(diagnostic.package);
+    if let Some(unit) = store.get(package_id) {
+        WithSource::from_map(&unit.sources, diagnostic.error)
+    } else {
+        WithSource::from_map(
+            &SourceMap::default(),
+            qsc_fir_transforms::PipelineError::MissingDiagnosticSourcePackage(diagnostic.package),
+        )
+    }
```
swernli: while this is nicely defensive, I don't think this error case is needed. If the package id for the source of a transform error is not in the store, how did it get transformed in the first place? Seems like this can be changed to an expect and the `MissingDiagnosticSourcePackage` error removed entirely.

swernli (approved):

PR #3242, opened 2026-05-20
Learning panel follow-up

minestarks:
**DON'T REVIEW** will split into separate PRs.

Addresses remaining feedback from #3228.

### Structured hint markers

- Katas exercise hints previously used `<details><summary>Need a hint?</summary>...</details>` HTML blocks, extracted by a fragile regex in generate_katas_content.js.
- Replaced with a `:::hint` / `:::` fenced block syntax across all 49 exercise `index.md` files. The generator now matches these structured markers instead.

### AI file consolidation

- `agents/`, `prompts/`, `skills/`, and `resources/chat-instructions/` were scattered across vscode. Moved all Copilot-related files into a single ai directory.
- Updated package.json contribution paths (`chatInstructions`, `chatSkills`, `chatPromptFiles`, `chatAgents`) accordingly.
- The QDK Learning agent model was hardcoded to `"Claude Haiku 4.5 (copilot)"`. Now specifies `["Claude Haiku 4.5 (copilot)", "auto"]` so VS Code falls back gracefully if the model is unavailable.

### Progress file error handling

- `loadProgress()` previously caught all errors (including corrupt JSON) silently and reset progress to defaults, destroying user data.
- Now distinguishes file-not-found (use defaults) from invalid JSON or unexpected structure (throw with a descriptive message). The corrupt file is preserved — the user must fix or delete it.
- `saveProgress()` wraps write failures with a clear error message. Errors propagate through existing paths:
  - **Panel UI**: caught by `handleAction` → shown in the webview.
  - **LLM tools**: caught by `invoke()` → surfaced as `CopilotToolError` to the model. `get-state` also returns the error string when initialization fails due to a corrupt file.
  - **Dispose**: shows a `showWarningMessage` notification instead of silently swallowing.

PR #3472, opened 2026-07-15
Handle corrupt kata progress

amcasey:
From #3242

When qdk-learning.json is invalid, we need to do something about it before the user can actually use the katas.  Enrich the error messages and give the agent instructions for how to handle errors.  It now does a pretty good job of making simple repairs to the file.

### DIFF
--- a/source/vscode/ai/qdk-learning.agent.md
+++ b/source/vscode/ai/qdk-learning.agent.md
@@ -81,6 +81,12 @@ Call `get-state` first. If the user is asking to navigate, run, check, reset, et
 
 Render tool results in chat. Keep responses short and tutor-like.
 
+### Error Handling
+
+When you call `get-state`, you may receive an error indicating a problem loading the user's progress.  When this occurs, you have two options:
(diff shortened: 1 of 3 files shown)


source/vscode/ai/qdk-learning.agent.md
```diff
@@ -81,6 +81,12 @@ Call `get-state` first. If the user is asking to navigate, run, check, reset, et
 
 Render tool results in chat. Keep responses short and tutor-like.
 
+### Error Handling
+
+When you call `get-state`, you may receive an error indicating a problem loading the user's progress.  When this occurs, you have two options:
+1. **Preferred** Read the error message and the contents of `qdk-learning.json`.  Together, they should indicate what's wrong.  Explain to the user what's wrong and offer to fix the problem.
```
amcasey: Telling it to read the file was the secret sauce.  Before that, it would give up on interpreting the error and immediately try to delete the file.

billti (approved):

PR #3473, opened 2026-07-15
Katas use Auto if Haiku 4.5 unavailable

amcasey:
From #3242


amcasey:
Validated by making Haiku 4._1_ the preferred option so I could see it fall back.  Also confirmed in the docs that the array syntax is supported and works as you would expect (descending order of preference).

swernli (approved):

PR #3476, opened 2026-07-15
Redact accessToken and idToken in debug log

amcasey:
Reduces log spew and protects user tokens.


amcasey:
We're getting the token from a VS Code API anyway, so anything that programmatically read our output window can also request a token.

PR #3477, opened 2026-07-16
[QRE] Add type annotations in QRE

fedimser:


source/qdk_package/qdk/qre/models/factories/_cultivation.py
```diff
@@ -693,12 +693,13 @@ def _extract_data_from_simulations(filename: Path, zenodo_path: Path):
     # Extracting and processing simulation data #
     #############################################
     def select_rep_stats(
-        stats: list[sinter.TaskStats], circuit: stim.Circuit
+        stats: list[sinter.TaskStats],
+        circuit: stim.Circuit,  # pyright: ignore[reportAttributeAccessIssue]
```
joao-boechat: why is this needed? leave a comment so that it's more clear and we know when we can remove it

source/qdk_package/qdk/qre/interop/_cirq.py
```diff
@@ -736,7 +760,12 @@ def _decompose_(self, qubits: Sequence[cirq.Qid]) -> Iterator[cirq.Operation]:
             yield cirq.reset(comp_qs[i])
             yield cirq.SWAP(mem_qs[i], comp_qs[i])
 
-    def _to_trace(self, context: _CirqTraceBuilder, op: cirq.Operation, **_kwargs):
+    def _to_trace(
+        self,
+        context: _CirqTraceBuilder,
+        op: cirq.Operation,
+        **_kwargs: object,
```
joao-boechat: you use "object" for these kwargs but "Any" for the other ones. Is there any way you can be more specific as to avoid Any or is it a particularity of this function that you can type it as "object"?

source/qdk_package/qdk/qre/interop/_qsharp.py
```diff
@@ -54,7 +54,9 @@ def _bucketize_rotation_counts(
     return result
 
 
-def trace_from_entry_expr(entry_expr: str | Callable | LogicalCounts, *args) -> Trace:
+def trace_from_entry_expr(
+    entry_expr: str | Callable | LogicalCounts, *args: object
```
joao-boechat: same question here (as above)

joao-boechat (approved):

source/qdk_package/qdk/qre/models/factories/_cultivation.py
fedimser: This is needed to avoid installing stim library (whcih takes unsually long time) just for this single check.
I added a short comment.

source/qdk_package/qdk/qre/interop/_qsharp.py
fedimser: Changed to Any.

source/qdk_package/qdk/qre/interop/_cirq.py
fedimser: Changed to Any.

PR #3483, opened 2026-07-17
Unified GPU shaders

billti:
This PR merges the base and adaptive shaders back into one shader (WGSL) file. This removes a bunch of duplication that had to be kept in sync. (The bulk of the lines changed are merging code from the old wgsl files into the new one).

It also enables using tools such as the wgsl-analyzer VS Code extension <https://github.com/wgsl-analyzer/wgsl-analyzer> to reason about the file wholistically. I added regions the extension recognizes so the common areas can be collapsed for easy navigation...

image

And when the extension can see all the code it enables type inlay hints, goto def, etc.

image

This unification also puts in place the structure for some further optimizations I want to experiement with in the near future.


billti:
Note, there's a few wgsl-analyzer bugs on the backlog to be fixed before this code will show error free, specifically the below (that I know of right now), so expect some spurious errors in the editor still in the meantime

https://github.com/wgsl-analyzer/wgsl-analyzer/issues/703
https://github.com/wgsl-analyzer/wgsl-analyzer/issues/1317
https://github.com/wgsl-analyzer/wgsl-analyzer/issues/1318

orpuente-MS (approved):
Looks good, thanks!

My only concern is that we now seem to compile all the machinery of both profiles (base and adaptive) when running base or adaptive programs; but I didn't notice a perf difference, so I guess it's fine.

Also, the logic for computing buffer sizes for the adaptive path now seems to be redundant, since both paths share the same data structures. But this is non-blocking.

PR #3486, opened 2026-07-18
Std.Core.ConfigValue

fedimser:
* Store key-value configuration map in compiler state (actually, in PassContext).
* Add `Std.Core.ConfigValue` to access config value stored inside compiler state (actually it's stored in PassContext.
  * It must be in `Std.Core` so we can refer to it in HIR pass. 
* Add HIR transformation pass `ConfigInline` to relace`ConfigValue(key, default)` with a literal
  * If key is found, replaces call with found value.
  * Otherwise replaces with default.
  * Only supports 4 types (string, int, bool, double).
  * Error if stored or default value has type other than these 4 types, or when stored value has different type than default. This guarantees, for example, that `ConfigValue("key", 1)` will not be replaced with a String value if config has `{'key": "1"}`. 
  * Also require that both key and default value are literals.
* Add Python API to set config:
  * The config must be set by passing Python dict to Context constructor. If not passed, empty dict is used.
  * Deliberately not supported in `qsharp.init`. 
  * Once `Context` is created, it's immutable. To use different config value, new `Context` must be created. This makes it valid for `GetConfig` to be a function, not operation.
  * This is propagated from Python to Rust by passing dict to `Interpreter::new`. 
  * It's impossible to mutate config after `Context` is created using Python API.
* Tests:
  * Unit tests for rule.
  * Interpreter tests (in compiler/qsc/src/interpret/tests.rs) using simulator, including tests for all possible errors.
  * One QIR generator test to make sure the GetConfig is inlined in QIR. This single test takes 3 seconds to run. However, We we have 24 qirgen tests, each taking 3-6 seconds to run - meaning qirgen tests are slow.

Example:
```python
import qdk
context = qdk.Context(qsharp_config={"experiment_name": "baseline", "shots": 1000})
assert context.eval('Std.Core.ConfigValue("experiment_name", "")') == "baseline"
assert context.eval('Std.Core.ConfigValue("shots", 100)') == 1000
assert context.eval('Std.Core.ConfigValue("noise_level", 0.01)') == 0.01
```


source/qdk_package/README.md
```diff
@@ -103,6 +103,32 @@ For convenience, the following helpers and types are also importable directly fr
...
+after initialization. To use different configuration values, create a new `Context`.
+
+In Q#, read values with `Std.Core.GetConfig(name, defaultValue)`. In Q# code, config
+values are immutable: in the same program, repeated calls with the same 
+`(name, defaultValue)` produce the same result.
+
+Supported types: `int`, `float`, `str`, and `bool` (corresponding to `Int`, `Double`, 
+`String` and `Bool` in Q#). The type of each value in `qsharp_config` must match the 
+type of its corresponding default value.
+
```
swernli: nitpick: the trailing spaces cause warnings in my local markdown linting, and can be removed:

```suggestion
In Q#, read values with `Std.Core.GetConfig(name, defaultValue)`. In Q# code, config
values are immutable: in the same program, repeated calls with the same
`(name, defaultValue)` produce the same result.

Supported types: `int`, `float`, `str`, and `bool` (corresponding to `Int`, `Double`,
`String` and `Bool` in Q#). The type of each value in `qsharp_config` must match the
type of its corresponding default value.

```

source/compiler/qsc/src/interpret/tests.rs
```diff
@@ -122,6 +122,165 @@ mod given_interpreter {
...
+                "Std.Core.GetConfig(\"int_config\", 20, 30)",
+            );
+            is_only_error(
+                &result,
+                &output,
+                &expect![[r#"
+                    type error: expected (String, ?), found (String, Int, Int)
+                       [line_4] [("int_config", 20, 30)]
+                "#]],
+            );
```
swernli: These two test cases aren't really needed here, as this is part of the expected handling for compiling Q# functions and not specific to `GetConfig`.

source/compiler/qsc_passes/src/config_inline.rs
```diff
@@ -0,0 +1,173 @@
...
+    core: &global::Table,
+    package: &mut Package,
+    config: &FxHashMap<Rc<str>, Value>,
+) -> Vec<Error> {
+    let mut pass = ConfigInline::new(core, config);
+    pass.visit_package(package);
+    pass.errors
+}
+
+pub(crate) struct ConfigInline<'a> {
```
swernli: nitpick: this doesn't need to be `pub(crate)` as nothing outside of this file uses it.

```suggestion
struct ConfigInline<'a> {
```

fedimser: Removed pub

source/compiler/qsc/src/interpret/tests.rs
fedimser: Removed these tests.
I added them because it wasn't obvious to me what exactly will happen on these inputs.

swernli (approved):

PR #3489, opened 2026-07-20
allow repeat-until loops in functions

### LINKED ISSUE #3462: Allow RUF loops in functions
This program gets a warning saying that Main doesn't contain any quantum operations and could be a function.
```qsharp
@EntryPoint(Adaptive)
operation Main() : Unit {
    repeat {

    }
    until 1 < 2
}
```
```
main.qs@(2,11): operation does not contain any quantum operations
help: this callable can be declared as a function instead
```

If you take that advice and make it a function, you get an error:
```
main.qs@(3,5): functions cannot use repeat-loop expressions
```

There doesn't seem to be any particular reason RUF loops aren't allowed in functions.

HABER7789:
## Allow repeat-until loops in functions

Functions previously rejected `repeat`/`until` loops with the error *"functions cannot use repeat-loop expressions"*, even though `for` and `while` loops were already allowed. A repeat-until loop is ordinary classical control flow, so there was no reason to forbid it in functions.

This also produced contradictory tooling: the `NeedlessOperation` lint suggests turning a non-quantum operation into a function, but doing so failed to compile if the body used a repeat loop.

This removes the restriction in the `callable_limits` pass so functions may use repeat-until loops like any other classical loop. Operations are unaffected.

Fixes #3462

Example:

```qsharp
function CountDown() : Unit {
    mutable x = 3;
    repeat {
        set x -= 1;
    } until x == 0;
}

// before → error: functions cannot use repeat-loop expressions
// after  → compiles
```

### DIFF
--- a/source/compiler/qsc_passes/src/callable_limits/tests.rs
+++ b/source/compiler/qsc_passes/src/callable_limits/tests.rs
@@ -121,7 +121,7 @@ fn funcs_cannot_allocate_qubits() {
 }
 
 #[test]
-fn funcs_cannot_use_repeat() {
+fn funcs_can_use_repeat() {
     check(
         indoc! {"
             namespace Test {
@@ -131,14 +131,7 @@ fn funcs_cannot_use_repeat() {
             }
         "},
         &expect![[r#"
-            [
-                Repeat(
-                    Span {
-                        lo: 51,
-                        hi: 71,
-                    },
-                ),
-            ]
+            []
         "#]],
     );
 }
--- a/source/compiler/qsc_passes/src/callable_limits.rs
+++ b/source/compiler/qsc_passes/src/callable_limits.rs
@@ -31,10 +31,6 @@ pub enum Error {
     #[diagnostic(code("Qdk.Qsc.CallableLimits.QubitAlloc"))]
     QubitAlloc(#[label] Span),
 
-    #[error("functions cannot use repeat-loop expressions")]
-    #[diagnostic(code("Qdk.Qsc.CallableLimits.Repeat"))]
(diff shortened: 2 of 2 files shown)


source/compiler/qsc_passes/src/callable_limits/tests.rs
```diff
@@ -121,7 +121,7 @@ fn funcs_cannot_allocate_qubits() {
 }
 
 #[test]
-fn funcs_cannot_use_repeat() {
+fn funcs_can_use_repeat() {
```
billti, line 124: Is it worth a test that such a construct in source code can generate correct IR?

HABER7789: > Is it worth a test that such a construct in source code can generate correct IR?

Good point, that's a different layer than the `callable_limits` test (which I believe just checks the restriction is lifted). I'll add a partial-evaluator test in `loops.rs` that lowers a function containing a repeat-until loop to RIR, to confirm it generates correct IR. Will push shortly.

swernli: Since later passes convert `repeat` loops into `while` loops, there isn't really anything special about them after HIR part of the compiler. The additional test doesn't hurt, but isn't really needed, I think.

swernli (approved):

source/compiler/qsc_passes/src/callable_limits/tests.rs
HABER7789: Alright, I will drop the test and merge it since adding the test would require another review round. Thank you for the review!

PR #3491, opened 2026-07-20
Propagate static values through Partial Eval

swernli:
This change updates how variables with known static values propagate across callable invocations in Partial Evaluation. By recognizing the static values and setting the corresponding argument `ComputeKind` in the call scope, we can ensure the computed application generator set from RCA is invoked properly and unnecessary dynamism is not propagated.

### DIFF
--- a/source/compiler/qsc_partial_eval/src/lib.rs
+++ b/source/compiler/qsc_partial_eval/src/lib.rs
@@ -4740,8 +4740,8 @@ impl<'a> PartialEvaluator<'a> {
 // to full evaluation without requiring any emission of RIR instructions.
 fn is_static_value(args_value: &Value) -> bool {
     match args_value {
-        // Qubit/Result ids and variables values cannot be treated as static.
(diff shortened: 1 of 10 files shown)


swernli (commented):
Of note for reviewers: the only product code changes are in `source/compiler/qsc_partial_eval/src` in `lib.rs` and `evaluation_context.rs`. The rest of the updates are for tests and test snapshots.

idavis (approved):

source/compiler/qsc_partial_eval/src/lib.rs
```diff
@@ -4740,8 +4740,8 @@ impl<'a> PartialEvaluator<'a> {
 // to full evaluation without requiring any emission of RIR instructions.
 fn is_static_value(args_value: &Value) -> bool {
     match args_value {
-        // Qubit/Result ids and variables values cannot be treated as static.
```
fedimser, line 4743: nit: variables->variable

fedimser (approved):

PR #3492, opened 2026-07-21
Add performance logging for qdk native wheel build diagnostics

domorale:
## Summary
- Adds best-effort, non-blocking performance logging for qdk native wheel builds.
- Moves logging helpers out of build.py into a dedicated module.
- Logs pre-build, post-build, and failure snapshots (platform, Python version, sanitized command, system memory state).
- Adds in-build peak indicators using OS-native tooling where available.

## Scope
- Logging is limited to the qdk native wheel build path.
- Build outputs, flags, and packaging behavior are unchanged.
- No new runtime dependencies were added.

## Platform behavior
- Linux: uses /usr/bin/time -v for peak indicator output.
- macOS: uses /usr/bin/time -l for peak indicator output.
- Windows: keeps snapshot logging; no built-in peak wrapper parity yet.

## Validation
- python3 -m py_compile performance_logging.py
- python3 build.py --no-check-prereqs --no-check --no-test --qdk
- python3 ./build.py

## Risk
- Low.
- Logging is best-effort and non-blocking.
- If logging collection fails, the build continues.

## Current iteration scope
- Rename runtime diagnostics wording to performance logging.
- Move logging helpers out of the main build orchestration path into a dedicated module.
- Add during-build peak indicators on Linux and macOS using OS-native tooling.
- Keep diagnostics best-effort, non-blocking, and scoped to qdk native wheel builds.

## Potential follow-up (if needed)
- If current logs are insufficient, evaluate optional cross-platform library-based sampling (for example psutil) to complement OS-native logging.
- If cost/benefit is favorable, add process-tree tracking during the build window to capture peak root-process RSS and peak aggregate child-process RSS.
- If attribution remains unclear, report Top-N subprocess memory offenders near peak.
- For reproducible failures, consider optional profiler-assisted deep dives outside the default CI path.

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,6 +15,10 @@ import functools
 import tomllib
 
 from prereqs import check_prereqs, add_wasm_tools_to_path
+from performance_logging import (
+    performance_logging_enabled,
+    run_native_wheel_build_with_logging,
+)
 
 # Disable buffered output so that the log statements and subprocess output get interleaved in proper order
 print = functools.partial(print, flush=True)
@@ -219,7 +223,13 @@ def build_wheel(python_bin, src, env=None, outdir=None, maturin=False):
         if platform.system() == "Linux":
             args.append("--config-setting=build-args='--compatibility'")
     args.extend(["--outdir", outdir, src])
-    run(args, cwd=src, env=env)
+    target_is_qdk_native = maturin and os.path.abspath(src) == os.path.abspath(
+        qdk_python_src
+    )
+    if performance_logging_enabled() and target_is_qdk_native:
+        run_native_wheel_build_with_logging(args, cwd=src, env=env, repo_root=root_dir)
+    else:
+        run(args, cwd=src, env=env)
 
 
 def install_from_wheels(python_bin, *positional_args, cwd, env=None):
--- a/build.py
+++ b/build.py
@@ -14,11 +14,7 @@ import subprocess
 import functools
 import tomllib
 
-from prereqs import check_prereqs, add_wasm_tools_to_path
-from performance_logging import (
-    performance_logging_enabled,
-    run_native_wheel_build_with_logging,
-)
+from prereqs import add_wasm_tools_to_path, check_prereqs, psutil_checks
 
 # Disable buffered output so that the log statements and subprocess output get interleaved in proper order
 print = functools.partial(print, flush=True)
@@ -192,8 +188,35 @@ QISKIT_VERSION_MATRIX = [
 ]
 
 
-def run(cmd, cwd, env=None):
-    subprocess.run(cmd, check=True, text=True, cwd=cwd, env=env)
+def run(cmd, cwd, env=None, perf_log_config=None):
+    # Build one execution policy and reuse it across plain and instrumented
+    # subprocess paths so behavior cannot drift.
+    run_kwargs={ "cwd": cwd, "env": env, "text": True }
+    check = True
+
+    if perf_log_config is None:
+        subprocess.run(cmd, check=check, **run_kwargs)
+    else:
+        run_with_logging_lazy_import(cmd, check, run_kwargs, perf_log_config) 
+
+
+def run_with_logging_lazy_import(cmd, check, run_kwargs, perf_log_config):
+    # Delay importing perf logging so prereq checks can run first.
+    from performance_logging import ExecutionPolicy
+    from performance_logging import run_with_logging
+
+    execution_policy = ExecutionPolicy(run_kwargs, check)
+    run_with_logging(args=cmd, execution_policy=execution_policy, config=perf_log_config)
+
+
+def qdk_perf_log_config_lazy_import(src, maturin):
+    target_is_qdk_native = maturin and os.path.abspath(src) == os.path.abspath(qdk_python_src)
+    if not target_is_qdk_native:
+        return None
+
+    from performance_logging import PerfLogConfig
+
+    return PerfLogConfig(repo_root=root_dir, log_context_label="qdk native wheel")
 
 
 def pip_install(python_bin, *positional_args, cwd, env=None, extra_args=()):
@@ -223,13 +246,8 @@ def build_wheel(python_bin, src, env=None, outdir=None, maturin=False):
         if platform.system() == "Linux":
             args.append("--config-setting=build-args='--compatibility'")
     args.extend(["--outdir", outdir, src])
-    target_is_qdk_native = maturin and os.path.abspath(src) == os.path.abspath(
-        qdk_python_src
-    )
-    if performance_logging_enabled() and target_is_qdk_native:
-        run_native_wheel_build_with_logging(args, cwd=src, env=env, repo_root=root_dir)
-    else:
-        run(args, cwd=src, env=env)
+    perf_log_config = qdk_perf_log_config_lazy_import(src=src, maturin=maturin)
+    run(args, cwd=src, env=env, perf_log_config=perf_log_config)
 
 
 def install_from_wheels(python_bin, *positional_args, cwd, env=None):
@@ -326,6 +344,10 @@ def run_python_checks():
 if args.check_py:
     run_python_checks()
     exit(0)
+
+def use_repo_python_env():
+    # Repo-level build-system tests run with the same interpreter as build.py.
+    return sys.executable
 
 
 if npm_install_needed:
@@ -454,6 +476,19 @@ def run_python_integration_tests(cwd, interpreter):
     subprocess.run(command_args, check=True, text=True, cwd=cwd)
 
 
+def run_build_system_tests(repo_root, tests_dir, interpreter):
+    # Keep build-system tests on pytest for consistency with other Python suites.
+    # Run from repo root so imports like `performance_logging` resolve normally.
+    command_args = [
+        interpreter,
+        "-m",
+        "pytest",
+        "-q",
+        os.path.relpath(tests_dir, repo_root),
+    ]
+    subprocess.run(command_args, check=True, text=True, cwd=repo_root)
+
+
 def run_ci_historic_benchmark():
     branch = "main"
     output = subprocess.check_output(
@@ -500,6 +535,12 @@ def run_ci_historic_benchmark():
 if build_qdk:
     step_start("Building the qdk python package")
 
+    # Intentionally local to the qdk path: psutil gates perf logging for
+    # qdk native wheel builds only. Keeping this scoped avoids imposing
+    # psutil on unrelated build targets. If we want global enforcement,
+    # that should be an explicit team decision in check_prereqs().
+    psutil_checks(install=True)
+
     # Reuse (or create) the pip environment so qdk wheel can be built/installed consistently.
     python_bin, pip_env = use_python_env(qdk_python_src)
 
@@ -532,6 +573,16 @@ if build_qdk:
         install_python_test_requirements(qdk_python_src, python_bin)
         if not args.editable:
             install_from_wheels(python_bin, "qdk", cwd=qdk_python_src)
+
+    if run_tests:
+        step_start("Running build system tests")
+        repo_python_bin = use_repo_python_env()
+        run_build_system_tests(
+            root_dir,
+            os.path.join(root_dir, "tests", "build"),
+            repo_python_bin,
+        )
+        step_end()
 
     if args.check:
         step_start("Checking qdk public API surface for private type leakage")
--- a/performance_logging.py
+++ b/performance_logging.py
@@ -1,154 +1,580 @@
 #!/usr/bin/env python3
+"""Performance logging helpers for build subprocess execution.
+
+Public API:
+- PerfLogConfig: runtime configuration for instrumentation.
+- SubprocessRunKwargs: supported process-launch kwargs.
+- ExecutionPolicy: caller-owned execution behavior.
+- run_with_logging: instrumented subprocess execution entrypoint.
+"""
 
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
 import ctypes
-import functools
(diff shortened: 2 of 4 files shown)


build.py
```diff
@@ -187,11 +181,111 @@ def step_end():
     },
 ]
 
+TELEMETRY_PREFIX = "build.py: [telemetry]"
```
billti: The term "Telemetry" in our general usage means data sent to a service and stored in a database for analysis. This is really just logging/tracing via printing to the console if I'm reading this correctly. I'd avoid using the term 'telemetry' for that.

build.py
```diff
@@ -187,11 +181,111 @@ def step_end():
...
+
+
+def log_native_build_telemetry(prefix, args):
+    log_telemetry_header(prefix)
+    print(f"{TELEMETRY_PREFIX} platform={platform.platform()}")
+    print(f"{TELEMETRY_PREFIX} python={sys.version.split()[0]}")
+    print(f"{TELEMETRY_PREFIX} command={' '.join(args)}")
+    log_memory_snapshot()
+
+
```
billti, line 221: `build.py` is getting pretty long already. If we were to add this (more on that below), I'd probably want to move it to a separate module outside of build.py (e.g. see prereqs.py) as its not core to the build process.

build.py
```diff
@@ -219,7 +313,24 @@ def build_wheel(python_bin, src, env=None, outdir=None, maturin=False):
...
+        try:
+            run(args, cwd=src, env=env)
+        except Exception:
+            elapsed = time.time() - start
+            print(f"{TELEMETRY_PREFIX} qdk native wheel failed after {elapsed:.3f}s")
+            log_native_build_telemetry("qdk native wheel failure snapshot", args)
+            raise
+        elapsed = time.time() - start
+        print(f"{TELEMETRY_PREFIX} qdk native wheel completed in {elapsed:.3f}s")
+        log_native_build_telemetry("qdk native wheel post-build", args)
```
billti: What do we see as the value of adding this? Especially as this is logged before the build has started, and then after the build is finished, I'm not sure if it adds a huge amount of value. (How much memory is being used _during_ the build process is where we typically see memory pressure, but once the build process is done I expect most of that has been reclaimed).

build.py
amcasey: Definitely had me nervous for a second

domorale: Got it, I'll update it.

build.py
domorale: Makes sense, I'll move it in the next revision.

build.py
domorale: Indeed, this PR was just the first iteration to see general data in the three platforms. Besides addressing the above comments, I'm planning a small refinement to the macOS memory reporting semantics so that the pre/post snapshots are more comparable with Linux and Windows.

Now, to address your comment, the current plan discussed with Stefan is to start with peak memory tracking during compilation in a separate follow-up PR and then iterate further if the additional data indicates that more detailed diagnostics would be useful.

domorale: Discussed with Stefan, I'll be adding the peak memory tracking to this PR to align with the team's culture :).

build.py
domorale: Addressed in 0cc83c76b: logging wording is now consistently performance logging (not telemetry), including emitted log prefixes and related helper names.

build.py
domorale: Addressed in 0cc83c76b: extracted performance logging into a dedicated module and kept the main build script focused on orchestration for the qdk native wheel path.

build.py
domorale: Addressed in 0cc83c76b: this iteration adds during-build peak indicators using OS-native wrappers (/usr/bin/time -v on Linux and /usr/bin/time -l on macOS), while keeping diagnostics best-effort and non-blocking. Windows remains snapshot-based in this PR due to no built-in equivalent wrapper. As a follow-up, if needed based on signal quality and cost/benefit, we can evaluate optional cross-platform process-tree sampling and top-offender reporting for improved parity and attribution.

domorale:
@microsoft-github-policy-service agree company="Microsoft"

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+import sys
+import time
+
+# Keep logs interleaved with subprocess output.
+print = functools.partial(print, flush=True)
+
+PERF_LOG_PREFIX = "build.py: [perf-log]"
+
+
+def performance_logging_enabled():
```
amcasey: Without having thought too deeply about it, I would have thought this would stay in build.py since it's effectively the caller's prerogative to log or not.

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+        try:
+            interesting = {
+                "MemTotal",
+                "MemFree",
+                "MemAvailable",
+                "SwapTotal",
+                "SwapFree",
+            }
+            with open("/proc/meminfo", encoding="utf-8") as file:
+                for line in file:
```
amcasey, line 468: Is this order stable across runs?  If not, would it make sense to use the order from `interesting`?

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+                    if line.split(":", 1)[0] in interesting:
+                        print(f"{PERF_LOG_PREFIX} {line.strip()}")
+        except Exception as exc:  # best effort diagnostics
+            print(
```
amcasey: Will subsequent attempts be skipped or will this appear in the log repeatedly?

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+                    + f"memory_load={status.dwMemoryLoad}% "
+                    + f"total_phys={status.ullTotalPhys} "
+                    + f"avail_phys={status.ullAvailPhys} "
+                    + f"total_pagefile={status.ullTotalPageFile} "
+                    + f"avail_pagefile={status.ullAvailPageFile}"
+                )
+            else:
+                print(f"{PERF_LOG_PREFIX} GlobalMemoryStatusEx failed")
+        except Exception as exc:
+            print(
```
amcasey: Is this possible?  The code above makes it seem like it returns an hresult.

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+    _log_build_context("qdk native wheel pre-build", sanitized_args)
+    start = time.time()
+
+    wrapper = _time_wrapper_for_platform()
+    if wrapper is not None:
+        print(
+            f"{PERF_LOG_PREFIX} peak indicator source={' '.join(wrapper)}"
+            + " (check command output for max RSS)"
+        )
+    elif platform.system() == "Windows":
```
amcasey: Why not just `else`?  Doesn't the lack of a wrapper imply that it's unsupported?

performance_logging.py
```diff
@@ -0,0 +1,255 @@
...
+    for token in args:
+        token = str(token)
+        if os.path.isabs(token):
+            token = _sanitize_path(token, repo_root)
+        sanitized.append(token)
+    return sanitized
+
+
+def _time_wrapper_for_platform():
+    if platform.system() == "Linux" and os.path.exists("/usr/bin/time"):
```
amcasey: Do these cover child processes too?

performance_logging.py
domorale: In offline discussion with Stefan, we agreed on keeping logging enabled by default for the qdk native wheel path. The call site keeps the scope decision (only that path), while the helper module encapsulates logging mechanics so build orchestration stays focused. If we want runtime toggles later, we can add explicit controls (for example env flag or CLI option).

performance_logging.py
domorale: It is stable as long as proc meminfo keeps the same order, which is usually the case, but not guaranteed as a strict contract. If preferred, we can enforce a fixed output order for these fields.

performance_logging.py
domorale: There are no retries or skip logic tied to that error. It is best-effort logging only.
Within one build attempt, this warning can appear once per snapshot invocation (currently pre-build and post-build, or pre-build and failure snapshot), so up to twice per attempt. It may appear again on subsequent independent attempts/reruns if the underlying issue persists.

performance_logging.py
domorale: Good question. This API returns a BOOL (nonzero on success, zero on failure), not an HRESULT, so the current boolean check is intentional and correct. On failure, we log and continue as best effort.
Reference: https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex

performance_logging.py
domorale: Great catch. `wrapper is None` means we run unwrapped, not unsupported. In the current matrix, Windows is the only platform without a configured wrapper, which is why this branch was written as Windows-specific. That said, your point about future platforms is valid, and I’ll update this to add a generic fallback message when no wrapper is configured (for example: “peak indicator wrapper unavailable on <platform>; running unwrapped”). This preserves current behavior and makes intent explicit for new platforms.

domorale:
## Optional Next Iteration Proposal (qdk native wheel diagnostics)

Current iteration objective was to improve diagnostics for the qdk native wheel step using a dependency-free, OS-native approach.  
That objective is satisfied in this PR.

A potential next iteration is to evaluate optional `psutil`-based process metrics, mainly to improve Windows peak-memory parity and add top offender attribution.

### Why consider this next iteration
- Current gap: Windows does not have OS-native peak-wrapper parity in the current implementation.
- `psutil` can provide:
  - peak largest-process RSS
  - peak process-tree RSS
  - top offenders at peak

### Local experiment summary (same qdk native wheel path)
- OS-native time max RSS: 3906428 kB (~3814.9 MiB)
- psutil peak largest-process RSS: 3753.7 MiB (rustc)
- psutil peak process-tree RSS: 4524.5 MiB
- top offenders were identified with crate-level context

Interpretation:
- largest-process peak from `psutil` was close to OS-native max RSS (good parity signal),
- process-tree peak added extra attribution signal.

### Cost / tradeoff
- Adds one native dependency (`psutil`, ~1.1 MB installed in local test), which increases dependency/maintenance surface.

### Decision
No change is proposed to current iteration scope in this PR by this comment.  

If the team wants to pursue this next iteration, we can do either:

1. A new revision on this same PR, or
2. A separate follow-up PR.

If the team is satisfied with current status, we can keep this PR as-is and merge the dependency-free implementation.

performance_logging.py
domorale: Thanks for raising this. In the current PR scope, child-process aggregate peaks are not explicitly computed; the OS-native wrapper reports for the wrapped command. I added a top-level [PR comment](https://github.com/microsoft/qdk/pull/3492#issuecomment-5041204995) with an optional next-iteration proposal and experiment data (including process-tree metrics and offender attribution via psutil), so the team can decide whether to keep this PR as-is or extend it in a new revision/follow-up PR.

amcasey: I wouldn't have guessed the parent process was using much memory - doesn't it fork a bunch of children to do the work?

domorale:
Perf logging for the qdk native wheel build is now complete and validated.

This PR:
- adds best-effort, non-blocking perf logging for the qdk build path
- captures structured build context and memory diagnostics
- uses psutil process-tree sampling for peak memory tracking
- keeps psutil scoped to the qdk build flow only
- adds build-system test coverage for the perf logging behavior
- hardens path sanitization and normalization across Linux, macOS, and Windows

Validation:
- targeted perf-logging tests pass locally
- full `python build.py` passes locally
- GitHub CI passed on all platforms

performance_logging.py
domorale: Addressed by the current shape of the implementation. The scope decision still lives at the build.py call site and remains limited to the qdk native wheel path, while the logging mechanics moved into a helper module so the build orchestration stays focused. If we want an explicit runtime toggle later, we can add that separately.

performance_logging.py
domorale: Addressed in the latest revision. The Linux snapshot now emits only a fixed set of keys in an explicit order rather than relying on `/proc/meminfo` iteration order.

performance_logging.py
domorale: This concern is now largely superseded by the current implementation, but the behavior remains best-effort and per-run. Any warning is local to a given build attempt and does not affect execution; it can appear again on a later independent rerun if the underlying environment issue still exists.

performance_logging.py
domorale: Yes. This path uses a BOOL-style success/failure result rather than an HRESULT, so the boolean check is intentional. The implementation keeps that path best-effort and non-blocking.

performance_logging.py
domorale: The current implementation no longer relies on the earlier wrapper-specific split in the same way. The fallback behavior remains best-effort and explicit, but the final version moved to psutil-based process-tree sampling for peak tracking rather than leaning on platform wrapper parity.

performance_logging.py
domorale: Addressed by the final implementation. Peak tracking now uses psutil process-tree sampling, so it covers the build process together with its children rather than only a single wrapped process view.

tests/build/test_performance_logging.py
```diff
@@ -0,0 +1,393 @@
+import subprocess
+import unittest
```
fedimser, line 2: We use pytest for Python tests everywhere else in QDK. Would be good to use only one testing framework for all Python tests.
Is there a reason to prefer unittest over pytest for this test?

domorale: No reason at all. I agree we should use pytest.

domorale: I have the update in my local branch just using pytest. On monday we'll discuss with the team about the pros/cons of adding functionality and dependencies (psutil in this PR) to the QDK build.

PR #3495, opened 2026-07-21
Workaround SBOM issue

swernli:
Manually installs .NET on all platforms, sets SBOM generation task to fail rather than continue on error.


amcasey (approved):

amcasey:
Consider adding a comment that this is a workaround

PR #3496, opened 2026-07-21
Raise Python type checking level to Standard

fedimser:
This PR changes Pyright type checking mode from "basic" to "standard" and fixes 44 new errors found by Pyright.

For reference, [here](https://github.com/microsoft/pyright/blob/main/docs/configuration.md) you can find the list of checks corresponding to "standard" mode.


source/qdk_package/qdk/qiskit/backends/backend_base.py
```diff
@@ -212,17 +212,22 @@ def _build_target(self) -> Target:
         )
 
     @property
-    def target(self) -> Target:
+    def target(self) -> Target:  # pyright: ignore[reportIncompatibleMethodOverride]
```
amcasey, line 215: Why were these suppressions needed?

source/qdk_package/qdk/qre/application/_openqasm.py
```diff
@@ -17,6 +17,15 @@
 from ..interop import trace_from_entry_expr
 
 
+def _find_unique_openqasm_name() -> str:
+    """Find an unused module name for importing an OpenQASM program."""
+    for _ in range(1_000):
```
amcasey, line 22: Not your code, but choosing a thousand random numbers seems like a strange way to find an unused name.

amcasey (approved):

source/qdk_package/qdk/qiskit/backends/backend_base.py
fedimser: Because these functions are not annotated in qiskit, which is understood as returning None.

For example: https://github.com/Qiskit/qiskit/blob/ccd029f4cab63c2540becc55b96aec4f7c128c11/qiskit/providers/backend.py#L149-L153

As you can see, most of methods on Qiskit are annotated with return type, but these 3 (target, max_circuits and _default_options) are not.

source/qdk_package/qdk/qre/application/_openqasm.py
fedimser: I'd like to keep this PR functionally no-op.

This logic makes sense to me. It's "pick random name, and retry if it's already used" but with protection from infinite loop. I expect in practice this to perform O(1) iterations when called << 1M times.

I am only worried if `code` is long-living object (e.g. someone runs it on a server) and it's eventually called close to 1M times, then this will start failing.

Maybe it would be better to create new Context here rather that using global `code` object...

amcasey: Agreed, definitely don't change it in this PR.

PR #3499, opened 2026-07-21
[STIM] SELECT block improvements + support for REPEAT

joao-boechat:
This PR makes the `SELECT` block's scoping rules more flexible, and adds support for two new instructions: `NOTLEAKED` (an analogue of `REQUIRE` that only checks for loss) and `REPEAT k`.

## Support for NOTLEAKED
`NOTLEAKED` is a QDK-only instruction that checks whether any of the listed measurement records correspond to a lost qubit. If any of them indicate loss, the enclosing `SELECT` block restarts, exactly like `REQUIRE` does when its condition fails.
Example:
```stim
SELECT {
    M 0
    NOTLEAKED rec[-1]
}
```

## New SELECT scoping rules

Previously the scoping rules were very restrictive: a `SELECT` block could only reference measurements made inside that same block. Now, for each `REQUIRE` / `NOTLEAKED` instruction, we only require that at least one of the referenced measurement records is in scope (i.e. produced in the same or an inner scope).

This still rules out programs that could only ever loop forever, while allowing much more flexibility.

For example, this is an error, since the only referenced record comes from outside the block, so restarting could never change it:

```stim
M 0
SELECT {
    REQUIRE rec[-1]
}
```

...while this is allowed, because `rec[-1]` refers to an in-scope measurement:

```stim
M 0
SELECT {
    M 1
    REQUIRE rec[-1] rec[-2]
}
```

The same rule applies to `NOTLEAKED`. For more examples, see `samples/notebooks/stim_select.ipynb`.

## Support for REPEAT blocks

Finally, we now support `REPEAT` blocks, as described in the official [Stim docs](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#REPEAT). These are implemented by unrolling the block.


source/compiler/stim_compiler/src/qir/tests/select_block.rs
fedimser, line 150: This jumps to next line, so this jump looks redundant, is it necessary?

fedimser, line 166: The "else" part of this branch jumps to next line, so it seems redundant.
Does QIR have version of "br" taking one condition and one label so we can do `br i1 %restart_1, label %select_0`. Then we wouldn't need to generate "continue" labels.

source/vscode/syntaxes/stim.tmLanguage.json
```diff
@@ -52,7 +52,7 @@
       "name": "comment.line.number-sign.stim"
     },
     "control-flow": {
-      "match": "\\b(REPEAT|PREPARE|REQUIRE)\\b",
```
fedimser, line 55: Why removed PREPARE?

fedimser (approved):

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1230,16 +1261,46 @@ impl<'noise> Compiler<'noise> {
...
+            return;
+        };
+
+        let mut loss_registers = Vec::new();
+        for &(result_id, negated) in &record_metadata {
+            if negated {
+                self.push_error(Error::NegatedTarget {
+                    instruction: instruction.name.clone(),
+                    span: instruction.span,
+                });
```
orpuente-MS: This error is using the span for the entire instruction. Would be nice to use the span for the specific target; but not a blocker if it requires a large change.

orpuente-MS: If you do that change, it be a good idea to push errors for all negated targets before returning from this method.

orpuente-MS (approved):

source/vscode/syntaxes/stim.tmLanguage.json
joao-boechat: Now we're using "SELECT" instead of "PREPARE"

source/compiler/stim_compiler/src/qir/tests/select_block.rs
joao-boechat: Unfortunately, in QIR there's no concept of "if this is true go here, other just keep going with the program counter". So we need to specify both labels

joao-boechat: In QIR you have to add those statements before adding a label

PR #3503, opened 2026-07-22
Allow blocks in interpolated string expressions

### LINKED ISSUE #3463: Allow blocks in interpolated string expressions
```qsharp
@EntryPoint(Adaptive)
operation Main() : Int {
    Message($"{for j in 1..i-1 { Message() }} ");
}
```

[Playground](https://microsoft.github.io/qdk/?code=H4sIAAAAAAAACj3JsQqAIBQF0N2vuESDDgmutdTQ0CD0C0IWr%252BEZ9ghC%252FPe2znrGmSW%252FayIWPW3hEnqiUemKOQglhg%252FE2qDHwoKiAMDH%252Bw5H1G1T9pRxghjOWuocyp8GtaIxg6ofpG9nqGIAAAA%253D)

It would be nice if we at least parsed it as balanced brackets and then produced an error, but once we've parsed it, we can probably just evaluate it.

HABER7789:
## Allow blocks in interpolated string expressions

Previously the raw lexer tracked interpolation depth with a single `u8` counter, which only balanced one level of braces and couldn't handle a `{ ... }` block or a nested interpolation inside `${ ... }`. As a result, expressions such as a `for` loop or a block inside an interpolated string failed to parse.

This change replaces that counter with a `Vec<u32>` stack that tracks brace depth per interpolation level: `{` increments the innermost level, `}` at depth greater than zero closes a block brace, and `}` at depth zero ends the interpolation. Blocks and nested interpolations now balance correctly, and the old 255-level `u8` overflow panic risk is gone.

I also added lexer, parser, and evaluator tests covering blocks, nested blocks, `for` loops, and an unclosed block.

This PR Fixes #3463.

Example:

```qsharp
function Main() : String {
    let x = 5;
    $"result is { { let y = x + 1; y * 2 } }"   // -> "result is 12"
}
```

### DIFF
--- a/source/compiler/qsc_parse/src/lex/raw.rs
+++ b/source/compiler/qsc_parse/src/lex/raw.rs
@@ -172,7 +172,7 @@ pub enum CommentKind {
 #[derive(Clone)]
 pub struct Lexer<'a> {
     chars: Peekable<CharIndices<'a>>,
-    interpolation: u8,
+    interp_brace_depths: Vec<u32>,
     starting_offset: u32,
 }
 
@@ -181,7 +181,7 @@ impl<'a> Lexer<'a> {
     pub fn new(input: &'a str) -> Self {
         Self {
             chars: input.char_indices().peekable(),
-            interpolation: 0,
+            interp_brace_depths: Vec::new(),
             starting_offset: 0,
         }
     }
@@ -190,7 +190,7 @@ impl<'a> Lexer<'a> {
     pub fn new_with_starting_offset(input: &'a str, starting_offset: u32) -> Self {
         Self {
             chars: input.char_indices().peekable(),
-            interpolation: 0,
+            interp_brace_depths: Vec::new(),
             starting_offset,
         }
     }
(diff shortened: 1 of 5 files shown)


fedimser:
It would be good to add more non-trivial tests, maybe when there is a for-loop that affects result:
```
$"result is { mutable r=0; for i in 0..2 {set r+=1}; r }"
```

fedimser:
As discussed, please also add a test where vector would have more than one element (interpolated string inside interpolated string, and both using blocks). And it should return something non-trivial so we can see expression was correctly evaluated.

fedimser (approved):

swernli (approved):

source/compiler/qsc_parse/src/lex/raw.rs
```diff
@@ -172,7 +172,7 @@ pub enum CommentKind {
 #[derive(Clone)]
 pub struct Lexer<'a> {
     chars: Peekable<CharIndices<'a>>,
-    interpolation: u8,
+    interp_brace_depths: Vec<u32>,
```
amcasey, line 175: If we wanted this to be strictly pay-for-play, we could stick with a `u8` until we hit an extra lbrace and then promote to `vec`.  I don't think that's worthwhile right now.

amcasey (approved):

source/compiler/qsc_parse/src/lex/raw.rs
HABER7789: Agreed. Keeping `Vec` and updating the tests. Happy to revisit if nested interpolation shows up on a hot path.

fedimser: I wanted to add similar suggestion - use `smallvec` with default size 1, so it does memory allocation only if we encounter nested interpolated strings (which should be rare). I don't think it could hurt and while we don't use `smallvec` in codebase it's used in transitive dependencies. But then, the gain of this is negligible.

PR #3505, opened 2026-07-23
Add arithmetic library

fedimser:
See https://github.com/microsoft/qdk/tree/fedimser/arith-lib/library/arithmetic

This is tested by Python tests that use qdk.Context (wrapped in a pytest ficture) and ArithmeticTestHelper. Tests are in library/arithmetic/test and I manually added them to paths for integration tests in build.py. I am not sure if it's the best approach.


joao-boechat (commented):
I think the configuration should use more standard jargon for what we are trying to optimize rather than defining the specifics of it. So something like "optimize_space" vs "optimize_time" makes more sense imo. And you should make the default behavior more explicit (the else)

fedimser:
See https://github.com/microsoft/qdk/pull/3523 instead.

PR #3507, opened 2026-07-23
Prevent divergent types from flowing out of loop and conjugate exprs

swernli:
This is a follow up from #2700, which turns out didn't fully fix the problem. All three loop constructs are known to be of type `Unit` and should not have their type marked as divergent based on any containing expressions. This matches behavior of other languages like Rust that do not propagate the divergent/unknown/`!` type from within loop structures. Notably this makes for a change to both Q# and OpenQASM where a loop with a `fail` expr inside will no longer satisfy type checking on it's own, requiring some explicit return values to match an outer callable scope or binding. The OpenQASM lowerer detection of missing returns is updated to provide feedback to the user rather than assuming a loop with a return satisfies requirements.

### DIFF
--- a/katas/content/teleportation/entanglement_swapping/Verification.qs
+++ b/katas/content/teleportation/entanglement_swapping/Verification.qs
@@ -35,10 +35,11 @@ namespace Kata.Verification {
                 return false;
             }
 
-            Message($"Correct.");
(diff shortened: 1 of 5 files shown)


ScottCarda-MS (approved):
This seems like a simpler approach. Thanks!

PR #3523, opened 2026-07-27
Add arithmetic library

fedimser:
See https://github.com/microsoft/qdk/tree/fedimser/arith-lib-2/library/arithmetic

This is tested by Python tests that use qdk.Context and ArithmeticTestHelper. Tests are in library/arithmetic/test and I manually added them to paths for integration tests in build.py.

### DIFF
--- a/library/arithmetic/qsharp.json
+++ b/library/arithmetic/qsharp.json
@@ -1,0 +1,19 @@
+{
+    "license": "MIT",
+    "author": "Microsoft Quantum",
+    "files": [
+        "src/Add.qs",
+        "src/AddConst.qs",
+        "src/AddLookup.qs",
+        "src/ClassicalMath.qs",
+        "src/Compare.qs",
+        "src/ModAdd.qs",
+        "src/ModDiv.qs",
+        "src/ModExp.qs",
+        "src/ModMul.qs",
+        "src/ModNegate.qs",
+        "src/MultiControl.qs",
+        "src/ResourceEstimation.qs",
+        "src/Utils.qs"
+    ]
+}
--- a/library/arithmetic/test/test_mod_add.py
+++ b/library/arithmetic/test/test_mod_add.py
@@ -1,0 +1,74 @@
+import random
+
+import pytest
+from qdk.test_utils import ArithmeticOpTester
+
+from test.test_utils import get_qdk_context
+
+
+@pytest.mark.parametrize("optimize", ["space", "time"])
+@pytest.mark.parametrize("n", [2, 5, 20])
+def test_mod_add(n: int, optimize: str):
+    context = get_qdk_context(optimize=optimize)
+    for _ in range(10):
+        modulus = random.randint(2, 2**n - 1)
+        x = random.randint(0, modulus - 1)
+        y = random.randint(0, modulus - 1)
+        op = f"ModAdd.ModAdd(_,_,{modulus}L)"
+        result = ArithmeticOpTester.run_op(op, [n, n], [x, y], context)
+        assert result == [x, (x + y) % modulus]
+
+
+@pytest.mark.parametrize("optimize", ["space", "time"])
+@pytest.mark.parametrize("n", [2, 5, 20])
+def test_mod_add_controlled(n: int, optimize: str):
+    context = get_qdk_context(optimize=optimize)
+    for _ in range(10):
+        modulus = random.randint(2, 2**n - 1)
+        x = random.randint(0, modulus - 1)
+        y = random.randint(0, modulus - 1)
+
+        # Test with control = 1 (should apply ModAdd)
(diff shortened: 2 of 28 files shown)


library/arithmetic/test/test_mod_add.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pytest.mark.parametrize("optimize", ["space", "time"])
+@pytest.mark.parametrize("n", [2, 5, 20])
+def test_mod_add(n: int, optimize: str):
+    context = get_qdk_context(optimize=optimize)
+    for _ in range(10):
+        modulus = random.randint(2, 2**n - 1)
+        x = random.randint(0, modulus - 1)
+        y = random.randint(0, modulus - 1)
+        op = f"ModAdd.ModAdd(_,_,{modulus}L)"
```
billti, line 17: I still find it a little odd to have exported operations in modules with the same name, e.g. `ModAdd.ModAdd` and `ModDiv.ModDiv`, etc. 

Have you considered exporting all names (if they are all unique) from the top level of the module? For example, see table_lookup at https://github.com/microsoft/qdk/blob/main/library/table_lookup/src/Main.qs . Anything you reexport from `Main.qs` in a library becomes available at the top level (e.g. see https://github.com/microsoft/qdk/wiki/Q%23-External-Dependencies-%28Libraries%29#the-mainqs-convention ). So for example if you referenced this library as `arithmetic`, then from the consuming code you would just have `import arithmetic.ModAdd` etc.

library/arithmetic/qsharp.json
```diff
@@ -0,0 +1,19 @@
...
+        "src/ModAdd.qs",
+        "src/ModDiv.qs",
+        "src/ModExp.qs",
+        "src/ModMul.qs",
+        "src/ModNegate.qs",
+        "src/MultiControl.qs",
+        "src/ResourceEstimation.qs",
+        "src/Utils.qs"
+    ]
+}
```
billti, line 19: Add this library to the manifest also (https://github.com/microsoft/qdk/blob/main/source/vscode/src/registry.json) so that it shows up in the pre-populated list when you right-click on a qsharp.json file and say "Add Q# project reference"

library/arithmetic/test/test_mod_add.py
fedimser: Having file name match a name of entity (e.g. class) defined in that class is a standard practice in programming languages. In Java, for example, usually a class is put in a file with exactly the same name.

For me, there is nothing wrong in writing `import Artihmetic.ModDiv.ModDiv`. In most files, there is more than one operation exported, and at most one of them matches the file name. The intention is to group similar operations, or operations doing the same thing differently. For example, in `AddConst` there are 3 operations `AddConstantSanders`, `AddConstantUsingCGAdd` and `AddConstant`.

I don't mind re-exporting everything in Main, and I will do this, but I don't see it adding any value. Also, this pattern will break as soon as operation with the same name is defined in 2 different files.

fedimser: Actually, there already is a conflict. There is `ModMul.ModMul` and `ModDiv.ModMul`. Both are modular multiplication, but of different kind. `ModMul.ModMul` is out-of-place and based on modular doubling, and supports 0. `ModDiv.ModMul` is in-place, based on extended euclidean algorithm, and doesn't support multiplication by 0.

I think the problem you are seeing is that structure and naming of algorithms is not well thought through. In general-purpose library like this, you might expect to have similar but different operations, coming from different research papers and having different purpose.

There are few ways to organize it:
* Add a file for every research paper. Inside the file, add all helpers, building blocks and circuits defined in that paper. This is how I structured a library for arithmetic I worked on earlier (https://github.com/fedimser/quant-arith-re/tree/main/lib/src/QuantumArithmetic). This approach is convenient for library authors - for new paper you just add a new file and you don't bother if very similar functionality is already implemented in other file. This, however, is not user-friendly - when a library user reads file tree, names like `LYY2021.qs` don't say anything about what algorithms are inside.
* Add each file for a kind of arithmetic operation, and have multiple implementations of that operation inside that file - this is approach I did here. It makes it easier for the user to find the file they want to use, but it naturally leads to imports like `ModAdd.ModAdd`.

fedimser: And `export ModAdd.ModAdd` doesn't work anyway.

If I do `export ModAdd.ModAdd as ModAdd1` in `Main.qs` in this library and then `import Arithmetic.ModAdd1 as ModAdd;` at call site, it works. However, when I do  `export ModAdd.ModAdd` in `Main.qs` and   `import Arithmetic.ModAdd` at call site, I get `ModAdd not found` error. 

This makes sense, because how the compiler can distinguish between `Arithmetic.ModAdd` namespace and `Arithmetic.ModAdd` operation?

So I'm going to keep everything as is, without re-exporting in `Main.qs`, so for each operation there is exactly one way to import it.

billti (approved):

PR #3526, opened 2026-07-29
Introduce Python Notebook learning experience prototype

amcasey:
This change generalizes the QDK Learning feature to support multiple courses of different kinds, including author-provided Python Jupyter notebook courses, while keeping the Quantum Katas as the default.

Builds on #3394.

- Removes a lot of the metadata files
- Drops the learning panel for notebook courses
- Adopts the Python Environments API
- Explicitly creates a Python project for Jupyter to pick up as a kernel


source/vscode/src/learning/commands.ts
```diff
@@ -138,3 +428,135 @@ function nodeToLocation(
     }
   }
 }
+
+/**
+ * Resolve a target course id from a tree node, or prompt the user with a
+ * quick pick when invoked without one (e.g. from the command palette).
+ */
+async function resolveCourseId(
+  service: LearningService,
```
HABER7789: Take a note to review this carefully

source/vscode/src/learning/commands.ts
```diff
@@ -138,3 +428,135 @@ function nodeToLocation(
...
+    modal: false,
+  });
+}
+
+/**
+ * Run environment diagnostics for a course and present a rich, readable
+ * report, offering the fixes the report surfaces (e.g. one-click
+ * environment setup, install extensions).
+ */
+async function runEnvironmentCheckCommand(
```
HABER7789: Check this, there's supposed to be two buttons

source/vscode/src/learning/commands.ts
```diff
@@ -138,3 +428,135 @@ function nodeToLocation(
...
+  }
+
+  const report = await vscode.window.withProgress(
+    {
+      location: vscode.ProgressLocation.Notification,
+      title: "Running course diagnostics…",
+    },
+    () => service.runEnvironmentCheck(),
+  );
+
```
HABER7789: Check this part

source/vscode/src/learning/notebookCourseProvider.ts
```diff
@@ -0,0 +1,368 @@
...
+        );
+        return undefined;
+      }
+      return parsed;
+    } catch (e) {
+      log.warn(`Failed to parse ${manifestUri.toString()}: ${String(e)}`);
+      return undefined;
+    }
+  }
+
```
HABER7789, line 173: validation

source/vscode/src/learning/notebookCellStatusBar.ts
```diff
@@ -0,0 +1,83 @@
...
+class LearningCellStatusBarProvider
+  implements vscode.NotebookCellStatusBarItemProvider, vscode.Disposable
+{
+  private readonly _onDidChangeCellStatusBarItems =
+    new vscode.EventEmitter<void>();
+  readonly onDidChangeCellStatusBarItems =
+    this._onDidChangeCellStatusBarItems.event;
+
+  private readonly subscription: vscode.Disposable;
+
```
HABER7789, line 27: Review

source/vscode/src/learning/notebookSync.ts
```diff
@@ -0,0 +1,104 @@
...
+} from "./constants.js";
+import type { LearningService } from "./service.js";
+
+/**
+ * Keep the learning service and {@link LEARNING_NOTEBOOK_ACTIVE_CONTEXT} in
+ * sync with the active notebook editor.
+ *
+ * VS Code activates this extension for any Jupyter notebook, so a session
+ * can be restored with a course workbook in the editor and the learning
+ * views never shown. Nothing else would initialize the service in that
```
HABER7789, line 19: Review

source/vscode/src/learning/progressTreeView.ts
```diff
@@ -53,12 +56,49 @@ class LearningProgressTreeProvider implements vscode.TreeDataProvider<LearningPr
...
 
+  constructor(private readonly service: LearningService) {}
+
   update(snapshot: OverallProgress | undefined): void {
     this.snapshot = snapshot;
     this.emitter.fire(undefined);
   }
 
   getTreeItem(node: LearningProgressNode): vscode.TreeItem {
+    if (node.kind === "course") {
```
HABER7789, line 68: Review

source/vscode/src/learning/service.ts
```diff
@@ -1,18 +1,31 @@
...
 import { FullProgramConfig, getProgramForDocument } from "../programConfig.js";
 import { ProgramRunStatus, runProgram } from "../run.js";
 import { EventType, sendTelemetryEvent } from "../telemetry.js";
-import { loadKatasCourse } from "./catalog.js";
+import { createCourseProvider, toDescriptor } from "./courseProvider.js";
+import { courseRootUri, workbookUri } from "./courseLayout.js";
+import { EnvironmentManager } from "./python/environment.js";
 import {
+  checkPythonExtensions,
+  promptInstallPythonExtensions,
```
HABER7789: Important review

source/vscode/src/telemetry.ts
```diff
@@ -344,7 +344,7 @@ type EventTypes = {
     properties: {
```
HABER7789, line 345: might make more changes to telemetry

source/vscode/package.json
```diff
@@ -341,6 +345,26 @@
...
+        {
+          "command": "qsharp-vscode.learningCheckEnvironment",
+          "when": "qsharp-vscode.learningWorkspaceDetected"
+        },
+        {
+          "command": "qsharp-vscode.learningNotebookHint",
+          "when": "false"
+        },
+        {
+          "command": "qsharp-vscode.learningNotebookExplain",
```
HABER7789, line 362: Maybe a few commands and tools could be removed in case unnecessary

source/vscode/package.json
```diff
@@ -1331,6 +1436,94 @@
...
+      },
+      {
+        "name": "qdk-learning-check-environment",
+        "tags": [
+          "qdk",
+          "qdk-learning",
+          "quantum-katas"
+        ],
+        "toolReferenceName": "qdkLearningCheckEnvironment",
+        "displayName": "QDK Learning: Check Environment",
```
HABER7789: Switch course review

package-lock.json
```diff
@@ -2267,6 +2268,17 @@
         "katex": "^0.16.4"
       }
     },
+    "node_modules/@vscode/python-environments": {
+      "version": "1.0.0",
+      "resolved": "https://ms-feed-12.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/@vscode/python-environments/-/python-environments-1.0.0.tgz",
```
amcasey: Make sure we don't merge with `ms-feed-25`

amcasey:
TODO: sanitize trial imports

amcasey:
TODO: non-exercise code cells in nav tree

.vscode/launch.shared.json
```diff
@@ -10,7 +10,7 @@
       "args": [
         "--profile=dev",
         "--extensionDevelopmentPath=${workspaceFolder}/source/vscode",
-        "${workspaceFolder}/samples/"
+        "${workspaceFolder}/source/vscode/test/suites/learning/test-workspace"
```
billti: I assume this is temporary for testing this PR and we don't want to change this be default when merged?

source/vscode/ai/qdk-learning.agent.md
```diff
@@ -28,13 +30,30 @@ The tools refer to each kata as a "unit." Each unit contains ordered activities
...
+| Intent                                | Tool                | Notes                                                                     |
+| ------------------------------------- | ------------------- | ------------------------------------------------------------------------- |
+| "What courses are available?"         | `list-courses`      | Returns the available courses and the active course id.                   |
+| "Switch to …" / "Open the … course"   | `switch-course`     | Pass the `courseId`. Switching changes the active course and position.    |
+| "Tell me about this course"           | `course-info`       | Returns the course descriptor and README (defaults to the active course). |
+| "Diagnose" / "Set up the environment" | `check-environment` | Runs environment diagnostics for the active course (Python courses).      |
+
+**Handling guidance:**
+
+- When the user asks to change courses, call `list-courses` first if you're unsure of the exact `courseId`, match the user's request to a course, then call `switch-course`. After switching, call `show` to surface the new course's current activity and briefly tell the user where they landed.
```
billti, line 53: Doesn't the user see all the courses in the tree view? Why does the user need to take and action or indicate to 'switch courses'? Shouldn't it just be whatever node they have selected (and thus the content is open in VS Code) is the active course?

source/vscode/src/learning/commands.ts
```diff
@@ -47,6 +48,37 @@ export function registerLearningCommands(
...
+          "Reset this unit to the original notebook? Your current work will be lost.",
+          { modal: true },
+          "Reset",
+        );
+        if (confirmed !== "Reset") {
+          return;
+        }
+
+        await service.resetExercise();
+        await openCourseNotebook(service);
```
billti, line 78: I'm not sure if resetting a notebook implies I want it open. (And if I did, I likely already have it open)

source/vscode/src/learning/index.ts
```diff
@@ -30,15 +32,65 @@ export function initLearning(
...
+    vscode.workspace.onDidChangeNotebookDocument((e) => {
+      // TODO (acasey): move to notebookSync.ts?
+
+      // When a cell finishes executing (executionSummary changes), auto-save
+      // the notebook, check if it corresponds to an exercise in the active
+      // python-notebook course and update focus. If execution succeeded,
+      // mark complete.
+      if (
+        !learningService.initialized ||
+        learningService.getActiveCourseInfo().kind !== "python-notebook"
```
billti: Just because the learning service is initialized and the active course is a Python notebook, that doesn't mean the notebook they just executed a cell in is one of the learning notebooks. They could also have other notebooks open. This code looks like it risks auto-saving notebooks outside of the learning experience whenever a cell is executed.

billti: (i.e. this early return should check if this is actually a learning notebook, and return if not)

source/vscode/ai/qdk-learning.agent.md
amcasey: You can ask the agent to list available courses and switch on your behalf.  Those seem at least marginally useful.

I believe the reason that it's an explicit gesture is that it causes a bunch of files to be created (and formerly triggered package installation).  Right now, when you click the button, the other courses get collapsed, which you probably wouldn't want if you were just clicking around a tree.

source/vscode/src/learning/commands.ts
amcasey: I expect it to already be open, since it's generally (only?) triggered from within an open notebook.  Opening it while it's open causes it to refresh (scroll to the top and possibly pick up a new virtual environment you've created).  It also triggers auto-save, since the buffer generally (always?) opens in a dirty state.

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,235 @@
...
+        return;
+      }
+
+      // Register the course folder as a Python project. This creates a workspace
+      // setting, which causes Jupyter to pick up the venv.
+      const courseName = courseRoot.path.split("/").pop();
+      await api.addPythonProject({
+        name: `QDK Course: ${courseName}`,
+        uri: courseRoot,
+      });
```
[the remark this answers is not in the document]
amcasey: Removed this file

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,235 @@
...
+    if (!env) {
+      return modules.map((module) => ({ module, ok: false }));
+    }
+
+    const results: { module: string; ok: boolean }[] = [];
+    for (const module of modules) {
+      const code = await runPython(api, env, ["-c", `import ${module}`]);
+      results.push({ module, ok: code === 0 });
+    }
+    return results;
```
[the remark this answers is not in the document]
amcasey: Removed this file

source/vscode/test/suites/learning/test-workspace/qdk-learning/courses/circuit-diagrams-new/_check_env.py
```diff
@@ -0,0 +1,208 @@
...
+    if missing:
+        results.append(
+            ("Packages", ", ".join(f"<code>{m}</code> missing" for m in missing), False)
+        )
+        errors.append(
+            "Install missing packages by running this in a new cell, then re-run this one:"
+            f"<pre>  %pip install -r ../requirements.txt</pre>"
+            "Or click the <b>Run Course Diagnostics</b> button in the notebook toolbar "
+            "to set up the full environment."
+        )
```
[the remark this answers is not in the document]
amcasey: Removed this file

source/vscode/src/learning/python/environment.ts
```diff
@@ -0,0 +1,235 @@
...
+    switch (envs.length) {
+      case 0:
+        return undefined;
+      case 1:
+        return envs[0];
+      default:
+        log.warn(
+          `Found multiple virtual environments, using first: ${envs.join(", ")}`,
+        );
+        return envs[0];
```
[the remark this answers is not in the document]
amcasey: Removed this file

source/vscode/src/learning/commands.ts
amcasey: Deleted this functionality

source/vscode/src/learning/commands.ts
amcasey: Deleted this functionality

source/vscode/src/learning/notebookCourseProvider.ts
amcasey: If this was about input validation, it wouldn't hurt but we're broadly covered by the fact that the user has to trust the workspace before our extension will even run.

source/vscode/src/learning/service.ts
amcasey: I think this was just flagging the file as key to the learning service.

source/vscode/src/telemetry.ts
amcasey: For now, I've just added the course kind.  That will give us a vague sense of how often python courses are used and keep our existing telemetry intact/separate.

source/vscode/package.json
amcasey: I got rid of browse.  The rest seem fine, though it sounds like we might consider dropping switch as well.

source/vscode/src/learning/commands.ts
amcasey: It turns out you can reset a unit you're not working on from the tree.  As written, the code does switch to that unit after resetting it.  I can't say that I remember adding that, so I'm guessing it was Mine, but it may have been copilot.  Personally, I find that behavior reasonable, since interacting with it in the tree does normally open it.

source/vscode/package.json
```diff
@@ -421,7 +450,21 @@
...
+          "command": "qsharp-vscode.learningResetUnit",
+          "when": "qsharp-vscode.learningNotebookActive",
+          "group": "navigation@110"
+        }
+      ],
+      "notebook/cell/title": [
+        {
+          "command": "qsharp-vscode.learningNotebookExplain",
+          "when": "notebookType == 'jupyter-notebook' && qsharp-vscode.learningWorkspaceDetected",
+          "group": "inline/cell@50"
```
billti, line 467: Where does `navigation@110` and `inline/cell@50` come from? Is there any significance to these names?

source/vscode/src/learning/notebookCourseProvider.ts
```diff
@@ -0,0 +1,362 @@
...
+    return courses;
+  }
+
+  // ─── Discovery ───
+
+  /** Enumerate candidate course folders and parse their manifests. */
+  private async discover(): Promise<CourseLocation[]> {
+    const dirs: vscode.Uri[] = [];
+
+    // The well-known in-workspace courses folder.
```
billti, line 125: The user workspace should contain the working copy, but the 'authored' content should come from a location in the extension (e.g. `resources/qdk-learning/courses` or similar).

It might be good to also have a VS Code setting for additional locations to look. That would make it easy for folks to author/test new content by having another location they can easily author or download content into and just have it show up as learning content (without having to open a particular folder as a workspace in VS Code)

source/vscode/src/learning/courseLayout.ts
```diff
@@ -0,0 +1,44 @@
...
+export function isNotebookCourse<Course extends CourseWithKind>(
+  course: Course,
+): course is Course & { kind: "python-notebook" };
+export function isNotebookCourse(course: CourseWithKind): boolean {
+  return course.kind === "python-notebook";
+}
+
+/** The authored notebook that a unit's workbook is derived from. */
+export function sourceNotebookUri(unit: NotebookCatalogUnit): vscode.Uri {
+  return unit.sourceNotebookUri;
```
billti, line 30: Does this function add any value? It just returns a property of the same name.

source/vscode/package.json
amcasey: My understanding is that before the @ indicates where the button will go and after is a precedence score (e.g. lower scores to the left)

source/vscode/src/learning/courseLayout.ts
amcasey: It used to.  I ended up leaving it to parallel `workbookUri` to make it clearer that those were your two options.

source/vscode/src/learning/notebookCourseProvider.ts
amcasey: In the current design, the author and learner copies are deliberately side-by-side so that relative paths (mainly to python scripts) will resolve the same way.

We're discussing offline what it would mean to include course content directly in the extension.

amcasey: Right now, we search all workspace roots for the first one that contains `qdk-learning.json`.  That's borderline inadequate.  If we added additional locations, we'd either need to handle multiple or provide some sort of collision detection.  Personally, I'd rather hold off on adding that complexity until there's a clear need for it.

If I were authoring a new course, I'd probably create and open an empty folder, initialize it with bundled courses, and then copy-paste a notebook course to hack up.  (To which end, maybe we want to synthesize a .gitignore for the user?)

amcasey: Anecdotally, the katas are quite a pain during authoring.  When I'm focusing on my own notebook course, I really don't want a bunch of kata files sprayed into the directory and having to wait for the katas (and the learning agent) to finish initializing before I can switch to my own course for a test run is tedious.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -325,4 +392,28 @@ export class LearningTools {
       },
     };
   }
+
+  /**
+   * For notebook courses, build CurrentActivity from the selected cell
+   * rather than from the service's stored position.
+   */
+  private notebookCurrentActivity(): CurrentActivity | undefined {
```
amcasey, line 400: @billti I still have mixed feelings about this part.  If we use the current progress, it returns info about the "current" code cell from the "state" perspective.  This will respect backwards navigation, but can't handle non-code cells.  That's fine for hints, which are only offered for code cells, but it makes "explain this" behave poorly.

On the other hand, looking at the selected cell is also flaky because the agent is so slow that the selection can change while it's working and there's a risk of it getting different information from consecutive tool calls.

Separately, we have similar logic in readCode.  readCode is only really intended to be called when retrieving exercise hints, but it's easy to imagine a false positive, so it feels like it has to behave sensibly in more scenarios.

All of which to say, I'd appreciate a second set of eyes on the tool changes.

source/vscode/src/learning/notebookCourseProvider.ts
```diff
@@ -0,0 +1,381 @@
...
+ */
+export class NotebookCourseProvider implements CourseProvider {
+  readonly id = "notebook-provider";
+
+  constructor(
+    private readonly workspaceRoot: vscode.Uri,
+    private readonly extensionUri: vscode.Uri,
+  ) {}
+
+  async listCourses(): Promise<NotebookCatalogCourse[]> {
```
billti, line 70: This is not just 'listing' the courses, it ends up copying a bunch of files into the users' workspace. Maybe rename (or make side-effect free).

amcasey: Renamed to "copyAndParseCourses"

billti (approved):

PR #3534, opened 2026-07-30
Drop redundant document updates in the LS

amcasey:
If the user types five characters, we get five document update events.  If no language service requests are made in the interim, only the last one actually matters.  At ~120ms each, it's worth dropping the ones we don't need.

We already had code for detecting that an update superseded a previous update (same file, higher version), but it never actually kicked in because of node's event loop: processing the update blocks the event loop so keystrokes that happen during that processing get queued up in VS Code without turning into events that would queue up (and get merged) in our language service.

Mine and I figured out where to put a setTimeout to yield so that more events could join the queue, but it was very fussy trying to tune the timeout duration.  It also fought with a prior fix (#1682) with its own timeout.

A better fix (guided by me but implemented by Copilot) makes completion wait until the version it wants is available (instead of for a fixed amount of time) and uses a flexible amount of yielding in the update handler.

Benefits:
- No longer sleep for >= 50ms before returning completions
- Drop document updates that would serve no purpose

Sadly, there's no good way to measure the improvement since we don't have the ability to detect when the keystroke comes in - we can only measure from the start of our processing to its end, skipping over the time spent waiting for a turn in the node event loop.


amcasey:
This needs both review and buddy testing since it's fairly subtle and largely Copilot's work.  Particular areas to focus on:

1. Rust memory management.  I've noticed it's a little clone-happy.
2. Testing.  Probably overkill.
3. VS Code vs Playground, since they use different yield mechanisms.

In local testing, debug builds are actually tolerable - you have to wait 1s for a single document update, but anything that happens in the interim gets merged.  This makes it a lot easier to see the merging behavior in action.

Edit: At present, the playground is (intended to be) unaffected - having a direct correspondence between edits and updates seemed desirable in a test environment.  So the testing that would be valuable is confirming that it's _actually_ unaffected.  If there's a way to test vscode.dev, that would be interesting too.

amcasey:
Edit [amcasey]: I got rid of MAX_YIELDS_PER_BATCH since it didn't empirically have any effect.  Now it just yields once.  The rest of the explanation should still apply.

Claude's summary of its own work:

## Adaptive yielding

```rust
let mut updates = vec![update];

for _ in 0..MAX_YIELDS_PER_BATCH {
    yield_to_host().await;

    let batch_before = updates.len();
    let drained = self.drain_pending(&mut updates);
    if drained == 0 {
        break;
    }
    let dropped = drained - (updates.len() - batch_before);
    trace!("drained {drained} update(s), merging dropped {dropped} as redundant");
}

self.apply(updates).await;
```

The loop asks the host "got anything else?" and stops as soon as the answer is no.

**Why not a fixed delay.** A fixed `setTimeout(20)` pays 20ms on *every* update, including a single keystroke after a long pause — pure added diagnostics latency for zero benefit. And the right value depends on typing speed, which is exactly the thing you said varies too much to tune against. The adaptive loop sidesteps the question: it's not waiting for *time*, it's waiting for the host to run out of queued input.

**Termination is structural, not timed.** When you pause, the first yield comes back with an empty channel and the loop exits after ~1 tick. While you're typing fast, the backlog drains in one or two passes and then the next keystroke is still ~100ms away, so the following yield comes back empty. The cap of 4 only bounds a pathological case; normally it exits earlier. No clock is involved, which is why the unit test is deterministic.

**One subtlety:** `drained` counts `try_recv` successes, not batch growth, because `push_update` merges in place. Using `updates.len()` would read "nothing arrived" when in fact five updates arrived and all merged — and the loop would stop absorbing exactly when it shouldn't.

## Node vs. browser

The yield has to be a *macrotask*. The channel wake from `unbounded_send` is a microtask (wasm-bindgen-futures polls via promises), and microtasks drain before the host can dispatch anything — so without an explicit macrotask hop the loop resumes having seen nothing.

But not all macrotasks are equal.

**Node** runs phases in a fixed order per iteration:

| Phase | What runs |
|---|---|
| timers | `setTimeout` / `setInterval` |
| poll | **I/O — the extension host reads queued IPC messages here** |
| check | `setImmediate` |

`setTimeout(0)` lands in **timers**, which precedes **poll**. Schedule one and it fires at the top of the next iteration — *before* that iteration reads the socket. The pending keystrokes are still sitting unread. This is why the original fixed delay needed 20ms: it wasn't waiting for a duration, it was waiting long enough that a poll phase happened to occur in some intervening iteration.

`setImmediate` lands in **check**, immediately *after* poll in the same iteration. One hop guarantees everything currently readable has been read and dispatched. That's precisely the semantics we want, and it's why the delay could drop to zero.

**Browser** (web extension host / worker) has no `setImmediate`. Its task queue is FIFO, and host `message` events are tasks, so a task posted via `MessageChannel` queues *behind* the already-pending message tasks — same guarantee by a different mechanism.

`setTimeout(0)` is also actively worse here: per the HTML spec, once nesting depth exceeds 5 the timeout is clamped to a 4ms minimum. Our yields chain from within timer callbacks, so they'd hit that clamp. `MessageChannel` isn't clamped.

Hence the runtime pick in `createHostYield()`: `setImmediate` → `MessageChannel` → `setTimeout(0)` as a last resort.

**Why inject it from TypeScript** rather than detect in Rust: the host knows its own environment, and it means the primitive can be swapped or tuned without a wasm rebuild — which matters given a release wasm build is ~60s.

**Measured:** 20 rapid edits under a 100ms simulated compile → **1 compilation on Node, 2 on web**. The web host taking one extra pass is consistent with `MessageChannel` being a slightly coarser approximation of "after all pending input" than `setImmediate`'s exact phase placement.

amcasey:
There's a hypothetical problem where you ask for completion at version N and N gets merged into N+1.  My expectation was that the next update would cancel the completion session anyway, so this (presumably rare) problem was acceptable.  If we find we're hitting it in practice, the solution is probably to communicate to the queue that a particular update has a caller waiting on it and is unmergeable.

The current mitigation is to mark results for affected requests as incomplete so VS Code will re-query.

Claude adds:

Two things worth remembering:

The whole design rests on that VS Code contract, so completion-retrigger.test.ts exists purely to guard it. It asserts both directions — complete lists are not re-requested, incomplete ones are. If VS Code ever changed this, that test fails loudly instead of the feature silently degrading into missing completion lists.

Returning a bare array from provideCompletionItems (the original behavior) means "complete", so the vscode.CompletionList wrapper is load-bearing, not cosmetic. Same for the 2s timeout path — it also returns isIncomplete: true, so a genuinely stuck update degrades to "ask again" rather than "no completions ever".

source/language_service/src/lib.rs
```diff
@@ -183,6 +200,49 @@ impl LanguageService {
...
+    /// caller's `position` was computed against `version`, so against newer text it may
+    /// point somewhere else entirely, or not exist at all.
+    ///
+    /// The returned future is independent of `&self` so that callers can hold it across
+    /// await points without keeping the language service borrowed.
+    pub fn wait_for_document_version(
+        &self,
+        uri: &str,
+        version: u32,
+    ) -> impl std::future::Future<Output = VersionWait> + 'static + use<> {
```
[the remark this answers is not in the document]
amcasey: @copilot What is `+ use<>`?  Don't change anything, just explain

source/language_service/src/tests.rs
```diff
@@ -367,6 +371,195 @@ fn assert_compilation(ls: &LanguageService, uri: &str, expected: &Expect) {
     expected.assert_debug_eq(&compilation.user_unit().sources);
 }
 
+/// Drives the real `run()` loop to verify that updates delivered by the host while an
+/// update is in flight get coalesced into a single compilation.
+///
+/// The host event loop is simulated by the yield closure: it pushes updates on the
+/// iteration where a real host would have delivered queued keystrokes. No timers are
+/// involved, so this is deterministic.
+#[tokio::test]
```
amcasey, line 380: These are probably not all that valuable, insofar as we already had tokio tests for the merging logic and they didn't catch the fact that it wasn't actually happening.

source/npm/qsharp/src/language-service/language-service.ts
```diff
@@ -192,16 +245,32 @@ export class QSharpLanguageService implements ILanguageService {
...
+      version,
+      completionWaitTimeoutMs,
+    );
+
+    const completions: CompletionListResult =
+      this.languageService.get_completions(documentUri, position);
+
+    if (status !== "ready") {
+      log.info(
+        `Providing completions for ${documentUri} from a ${status === "timeout" ? "older" : "newer"} version than requested`,
```
amcasey, line 266: I don't love this, but it's what we would have done before this change and it seems to be rare in practice.  In addition, I suspect the editor drops the results as stale.

source/language_service/src/tests.rs
```diff
@@ -367,6 +371,268 @@ fn assert_compilation(ls: &LanguageService, uri: &str, expected: &Expect) {
...
+    let yield_to_host = || {
+        if yields.replace(yields.get() + 1) == 0 {
+            let mut ls = ls.borrow_mut();
+            ls.update_document("bar.qs", 1, "namespace Bar { ", "qsharp");
+            ls.stop_updates();
+        }
+        std::future::ready(())
+    };
+
+    worker.run(yield_to_host).await;
```
amcasey: Should be called handler

source/language_service/src/tests.rs
```diff
@@ -367,6 +371,268 @@ fn assert_compilation(ls: &LanguageService, uri: &str, expected: &Expect) {
...
+    let mut ls = LanguageService::new(Encoding::Utf8);
+    let mut worker = create_update_handler(&mut ls, &received_errors, &test_cases);
+
+    ls.update_document("foo.qs", 1, "namespace Foo { ", "qsharp");
+
+    let ls = RefCell::new(ls);
+    let yields = Cell::new(0);
+    let yield_to_host = || {
+        if yields.replace(yields.get() + 1) == 0 {
+            let mut ls = ls.borrow_mut();
```
amcasey, line 443: This needs a comment explaining why there's an update document in the mock yield implementation

source/language_service/src/state.rs
```diff
@@ -677,6 +677,12 @@ fn is_openqasm_file(language_id: &str) -> bool {
 }
 
 impl CompilationState {
+    /// The version of `uri` that the client last told us about and that has since been
+    /// applied. `None` if we haven't processed any update for the document yet.
+    pub(crate) fn get_open_document_version(&self, uri: &str) -> Option<u32> {
+        self.open_documents.get(uri).map(|doc| doc.version)
+    }
```
orpuente-MS, line 684: What are some scenarios where we haven't processed any updates of the documents yet? Is this the case for opened files that haven't been modified? Does it happen for read-only views of a file?

amcasey: I'd expect it to happen for a document that is open but hasn't been changed since it was opened (possibly due to being read-only).  It seems like it would be unlikely (or possibly even impossible) in practice.

orpuente-MS (approved):

amcasey:
Reopening to try to unstick status check.

ScottCarda-MS (approved):

PR #3536, opened 2026-07-31
Use dependencies on local libraries in library tests

fedimser:
* Patches `qsharp.json` to depend on local library (via `path`) rather than on the same library on GitHub (via `github`).
* This fixes integration test failure that we get after updating dependencies to next version but before creating a tag for that version.
* Other benefits:
  * We don't need to make requests to GitHub while running integration tests (these requests sometimes fail).
  * We now test all libraries with latest version of their dependencies, so if there is a change that breaks integration between 2 libraries, we will discover it immediately and not after next release.
* We can't statically update libraries to depend on local paths because (1) these paths depend on environment, (2) this will make libraries unusable if user downloads the library separately, as dependencies would become invalid.

### DIFF
--- a/build.py
+++ b/build.py
@@ -444,7 +444,7 @@ def run_python_integration_tests(cwd, interpreter):
         cwd,
         os.path.join(root_dir, "library", "arithmetic", "test"),
     ]
-    command_args = [interpreter, "-m", "pytest", *paths]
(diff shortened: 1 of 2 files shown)


fedimser:
Will do https://github.com/microsoft/qdk/pull/3541 instead

PR #3537, opened 2026-07-31
Support emission of locally constant arrays with dynamic contents

swernli:
This change adds additional handling in partial eval for arrays on entry to a new scope during function calls. Specificaly, any arguments that are arrays are recursively checked for contents that are `Value::Var` which indicates that the contents of the array are dynamic at runtime. However, since Q# callable arguments are read-only, these arrays can be treated as "locally constant" and emitted as stored values on scope entry. This then allows later instructions that might need to refer to those arrays (for now, just index instructions) to use those stored values in the same manner that globally constant arrays are used from the data section of the program. Since the existing RIR passes are already capable of eliminating unused store instructions, these stored arrays only persist into the final emitted program when they are actually needed for program functionality. Notably, this unblocks the emission of `Adjoint Std.TableLookup.Select` callable when using the `Adaptive` profile.

### DIFF
--- a/source/compiler/qsc/src/codegen/tests.rs
+++ b/source/compiler/qsc/src/codegen/tests.rs
@@ -5773,6 +5773,332 @@ fn operand_continue_in_controlled_call_skips_consumer_and_following_statement_in
 }
 
 #[test]
+fn array_with_dynamic_contents_passed_as_argument_and_dynamically_indexed_emits_locally_constant_array()
+ {
+    let source = indoc::indoc! {r#"
+        namespace Test {
+            function ValueAt(arr : Int[], idx : Int) : Int {
+                arr[idx]
+            }
+            @EntryPoint()
+            operation Main() : Int {
+                use q = Qubit();
+                let arr = [if M(q) == One { 1 } else { 0 }, if M(q) == One { 1 } else { 0 }];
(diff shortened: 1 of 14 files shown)


source/compiler/qsc/src/codegen/tests.rs
```diff
@@ -5779,10 +5779,7 @@ fn foreign_table_lookup_callable_generates_qir() {
             use address = Qubit[1];
             use output = Qubit[1];
             Std.TableLookup.Select([[false], [true]], address, output);
-
-            // Adjoint of Select does not work until support for static sized, dynamic content arrays is added.
-            // See related issue: https://github.com/microsoft/qdk/issues/3388
-            // Adjoint Std.TableLookup.Select([[false], [true]], address, output);
+            Adjoint Std.TableLookup.Select([[false], [true]], address, output);
```
idavis, line 6108: Can you add a test that shows what the QIR looks like for the newly added behavior?

idavis (approved):

source/compiler/qsc/src/codegen/tests.rs
swernli: Added QIR snapshot unit tests to show the structure of the emitted code.

PR #3538, opened 2026-07-31
validate import checks in python course environment setup

HABER7789:
Module names from course.json flowed into a `python -c` command line without validation, so a drop-in course could run random Python when someone ran the environment check. They now have to be dotted identifiers, and anything else is reported as a failed check rather than silently dropped.

Also fixes the multiple-environments warning, which joined the environment objects directly and so logged `[object Object]` instead of their names.


HABER7789:
Closing this PR since it was only being used as a reference for the environment-check work. That work is now part of #3568, so this PR is no longer needed and won’t be merged.

PR #3540, opened 2026-07-31
Prevent code meant only for simulation from leaking into generated quantum programs and analysis

idavis:
## Summary

This change prevents code meant only for simulation from leaking into generated quantum programs and FIR transform analysis.

## What changed

- Removed simulation-only implementations before code generation.
- Treated diagnostic helpers such as `DumpOperation` and `Fact` as operations that produce no generated code.
- Preserved measurements and other side effects in their arguments.
- Added one shared list of operations that should not appear in generated output.
- Added earlier, clearer errors for unsupported generic or callable parameters.
- Updated compiler tests for restricted target profiles and cross-package code.

## Why

Simulation helpers may contain fallback code that works in the simulator but cannot be emitted as part of a target program. Processing that code caused compiler failures or references to operations that did not exist at runtime.

The compiler now removes those fallback bodies at the correct boundary while leaving normal simulation behavior unchanged.


source/compiler/qsc_data_structures/src/intrinsic_names.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/// Intrinsic names that partial evaluation handles as code generation no-ops.
+pub const CODEGEN_NOOP_INTRINSIC_NAMES: &[&str] = &[
+    "DumpRegister",
```
amcasey, line 7: Without really understanding what this list refers to, it surprises me that DumpRegister and DumpOperation would be present, but DumpMachine would be absent.

swernli: `DumpMachine` is missing for a rather subtle reason: it takes no arguments. Both the other operations may have dynamic expressions in their arguments, making the whole call expression dynamic and requiring that it gets handled by partial evaluation. Since `DumpMachine()` is a static expression so we expect it to always go to the full evaluator.

That said, thinking through this answer made me realize there is nothing stopping someone from writing something like `DumpMachine(X(q))` which is nonsensical but valid from a type inference perspective... so `DumpMachine` probably should be added to this list even though we've never encountered code like that.

swernli (commented):
One minor comment in response to Andrew's comment that can fix an existing bug, otherwise looks good!

swernli (approved):

source/compiler/qsc_data_structures/src/intrinsic_names.rs
amcasey: This would also make a great comment

swernli: Not really. The list just needs `DumpMachine` added. There's nothing special to comment on here.

idavis: Done

PR #3541, opened 2026-07-31
Allow overriding github deps with local deps using env variables

fedimser:
* Checked that integration tests pass after updating dependencies to non-existing tag v1.32.0.


.github/workflows/ci.yml
```diff
@@ -174,6 +174,9 @@ jobs:
     name: Integration tests
     timeout-minutes: 40
     runs-on: ubuntu-latest
+    env:
+      QDK_LIB_OVERRIDE_MICROSOFT_QDK_LIBRARY_QTEST: ${{ github.workspace }}/library/qtest
+      QDK_LIB_OVERRIDE_MICROSOFT_QDK_LIBRARY_SIGNED: ${{ github.workspace }}/library/signed
```
billti: this isn't the design we discussed in the chat. Is there a reason for making it per library rather than per repo?

billti: With this design if we ever added another (transitive) library dependency, we'd need to remember to go add a new environment var here for it. Why not just use the repo redirection rather than each library individually?

.github/workflows/ci.yml
```diff
@@ -174,6 +174,9 @@ jobs:
     name: Integration tests
```
billti, line 174: We use ADO to publish, which also runs the integration tests, so you'll need to update `.ado/publish.yml` also.

.github/workflows/ci.yml
fedimser: Done.

fedimser: Done.

.github/workflows/ci.yml
fedimser: That is not necessary, because by the time we publish package, the new tag already exists.

fedimser: Reason to do it in publish.yml - reduce traffic to GitHub, make pipeline less flaky.
Reason to not do it - more realistic setup, i.e. using GitHub deps gives us a chance to test libraries depending on each other in exactly the way how the end user would use it.

source/compiler/qsc_project/src/tests.rs
```diff
@@ -625,3 +625,33 @@ fn single_file_sets_has_manifest_false() {
...
+        repo: "qdk".into(),
+        r#ref: "main".into(),
+        path: None,
+    };
+
+    assert_eq!(
+        local_override_env_var_name(&github),
+        "QDK_LIB_OVERRIDE_MICROSOFT_QDK"
+    );
+}
```
domorale, line 642: Nice coverage for naming logic. It feels like we're still missing an easy way to test the "no override configured" behavior, since that path currently depends on process-wide environment state. Would it make sense to decouple the override logic from the environment lookup (e.g. via an injected lookup function)? That would make the override-present and override-absent cases much easier to unit test.

.github/workflows/ci.yml
billti: > That is not necessary, because by the time we publish package, the new tag already exists

No it doesn't. I usually don't create the tag and release on GitHub until after the build and publish succeeds (as that can sometimes fail requiring more time and/or commits before the release is live).

billti: (For example, here's a commit we had to push after we'd already done the version bump and the GitHub CI was passing, but ADO failed on the same commit due to having stricter Python version requirements - so we had to push further commits before the ADO build/publish succeeded)

https://github.com/microsoft/qdk/pull/1996

billti (approved):

PR #3542, opened 2026-07-31
Enable source maps in source/npm/qsharp/src

amcasey:
Lets us set breakpoints in those TS files.

### DIFF
--- a/.vscode/launch.shared.json
+++ b/.vscode/launch.shared.json
@@ -12,7 +12,10 @@
         "--extensionDevelopmentPath=${workspaceFolder}/source/vscode",
         "${workspaceFolder}/samples/"
       ],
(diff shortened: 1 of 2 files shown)


ScottCarda-MS (approved):

billti:
I'm not sure I follow what this enables. VS Code doesn't reference any compiled .js files in the npm folder does it? I though in only references the .ts files, and thus the bundling and map files are all already in the vscode/out directory.

amcasey:
> I'm not sure I follow what this enables. VS Code doesn't reference any compiled .js files in the npm folder does it? I though in only references the .ts files, and thus the bundling and map files are all already in the vscode/out directory.

I'm not sure how things were expected to work but my breakpoints in `source\npm\qsharp\src\language-service\language-service.ts` don't bind without this.

billti (approved):

PR #3544, opened 2026-07-31
[QRE] Dollar cost calculator

fedimser:


source/qdk_package/qdk/qre/_json_specs.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+
+    name: str
+    qubits_per_node: int
+    cnot_time: int
+    cnot_error_rate: float
+    one_qubit_time: int
+    one_qubit_error_rate: float
+    measurement_time: int
+    measurement_error_rate: float
```
brad-lackey: I don't think we want to have cnot, one_qubit, and measurement time/error_rate as part of the System spec, and this refers to a specific type of qubit architecture of the qubit plane (namely some platforms may not have these as their fundamental instructions). So, we may want to replace this with an Architecture instance, which is the reference qubit plane the system can refer to.

fedimser: I removed these 6 fields from the spec.

msoeken: The lines are still in the example_cost.json file.

fedimser:
I have another solution in https://github.com/microsoft/qdk/pull/3672
It requires passing cost per hour in Architecture object. We can add extracting it from some text file (like JSON) later.

PR #3550, opened 2026-08-04
[STIM] Support peek_loss

joao-boechat:
## Overview

This change adds end-to-end support for STIM's `PEEK_LOSS` instruction, allowing adaptive programs to report whether a qubit is lost without collapsing its state. The instruction produces a measurement-record result that can drive classical control while preserving the qubit for subsequent operations.

## STIM compiler

- Parses `PEEK_LOSS`, broadcasts it across qubit targets, allocates its result records, and lowers it to the `__quantum__qis__peek_loss__body` QIR intrinsic.
- Tracks records produced by `PEEK_LOSS` so they can be used by normal classical controls while rejecting unsupported `NOTLEAKED` references to those records. The reason why we don't allow `NOTLEAKED` to reference a measurement record generated by `PEEK_LOSS` is simply because combining both instructions is ambiguous and doesn't make semantical sense.

## Adaptive execution

- Adds the `PEEK_LOSS` adaptive bytecode opcode and emits it from the Python adaptive-profile pass.
- Extends the simulator interface and bytecode runtime with non-destructive loss inspection.
- Implements the operation in the CPU full-state, stabilizer, and GPU full-state simulators, recording `One` for a lost qubit and `Zero` otherwise.


source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1297,6 +1305,11 @@ impl<'noise> Compiler<'noise> {
             "REQUIRE" => self.compile_require(instruction),
             "NOTLEAKED" => self.compile_notleaked(instruction),
 
+            // Miscellaneous
+            "PEEK_LOSS" => self.broadcast(instruction, |s, q| {
```
billti: Are you only adding this for Stim? Should we not have similar functionality in Q# and/or QASM too?

source/simulators/src/gpu_full_state_simulator/gpu_statevector_shaders.wgsl
```diff
@@ -2883,6 +2884,17 @@ fn interpret_classical(@builtin(global_invocation_id) gid: vec3<u32>) {
                 pc++;
             }
 
+            // PEEK_LOSS: Reports whether a qubit was lost, but doesn't
+            // collapse the state.
+            case OP_PEEK_LOSS {
```
billti, line 2909: Being this is in the adaptive kernel, does this mean you can't use PEEK_LOSS in base profile programs? (Makes sense that you can't, we should just ensure that only adaptive programs are able to use it and give a useful error otherwise.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: This PR only adds it for stim. It will probably take more time to agree on a final design for Q# / QASM, so we can leave that for another PR.

joao-boechat: https://github.com/microsoft/qdk/issues/3610

Created this issue to track that.

source/simulators/src/gpu_full_state_simulator/gpu_statevector_shaders.wgsl
joao-boechat: The user currently can only run this instruction via qdk-stim (unless they type QIR directly, which is unusual), and by default the qdk stim compiler only generates adaptive profile QIR. In the future, we might also generate base profile, and will look into this again.

source/simulators/src/lib.rs
```diff
@@ -129,6 +129,9 @@ pub trait Simulator {
     /// for `PartialEq` up to a global phase. This is meant to be used for testing.
     fn state_dump(&self) -> &Self::StateDumpData;
 
+    /// Measures loss but does not collapse the state. The result is stored in `result_id`.
+    fn peek_loss(&mut self, target: QubitID, result_id: QubitID);
```
billti: Isn't the result_id of type `ResultID`, not `QubitID`?

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -710,6 +710,15 @@ impl Simulator for StabilizerSimulator {
         self.state.clifford()
     }
 
+    fn peek_loss(&mut self, qubit: QubitID, result_id: QubitID) {
```
billti: Same here. result_id is the wrong type.

source/simulators/src/bytecode/runtime.rs
```diff
@@ -475,6 +476,13 @@ pub fn run_shot<S: Simulator>(program: &AdaptiveProgram<u64>, sim: &mut S) -> Ve
                 rt.pc += 1;
             }
 
+            OP_PEEK_LOSS => {
+                let target = rt.resolve_u64(instr.aux0, flags, 3) as QubitID;
+                let result_id = rt.resolve_u64(instr.aux1, flags, 4) as QubitID;
```
billti: It's everywhere....

fedimser (approved):

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1433,23 +1474,24 @@ impl<'noise> Compiler<'noise> {
...
-            };
-            if factors.is_empty() {
-                continue;
-            }
-            self.decompose_pauli_product_operation(&factors, negated, &mut f);
-        }
+        self.for_each_pauli_product(instruction, operation);
+    }
+
+    fn broadcast_pauli_product_measure(
```
billti, line 1483: Is this related to peek_loss? Why are the changes to 'broadcast_pauli_product' in here?

source/simulators/src/lib.rs
joao-boechat: Fixed it!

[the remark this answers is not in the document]
joao-boechat: fixed.

source/simulators/src/stabilizer_simulator.rs
joao-boechat: Fixed.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -710,6 +710,15 @@ impl Simulator for StabilizerSimulator {
...
+        let is_lost = self.loss[qubit];
+        self.measurements[result_id] = if is_lost {
```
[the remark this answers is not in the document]
joao-boechat: Fixed.

source/simulators/src/bytecode/runtime.rs
joao-boechat: If you look around in this file (and the other ones), result_id was already using "QubitID" as the type. I had simply followed the pattern to avoid disruptions in other parts of the code. Now I'm modifying all of them to make sure they're consistent.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: Not directly, but upon writing the code for peek_loss I realized there was some code cleanup I didn't do in regards to pauli product operations (since the readout_noise PR came afterwards, and that one contained a bunch of the cleanup), so I took the opportunity to do it here rather than in a new PR.

source/qdk_package/tests/test_adaptive_cpu_bytecode.py
```diff
@@ -618,6 +618,50 @@ def test_read_loss(sim_type):
...
+
+
+@pytest.mark.parametrize("sim_type", SIM_TYPES)
+def test_peek_loss(sim_type):
+    """s (with 100% loss) → peek_loss records loss into a result without collapsing.
+
+    Result 0 is the peek: ``One`` ('1') because qubit 0 is lost. Result 1 is a
+    later mz on the same qubit: ``Loss`` ('L'), showing peek_loss did not clear
+    the loss — it behaves like a measurement but does not collapse state.
+    """
```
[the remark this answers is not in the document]
joao-boechat: Added a test case for that.

PR #3553, opened 2026-08-06
Fix spec & perf issues, rename and decouple the OpenQASM crate (`qdk_openqasm_parser` -> `qdk_openqasm`)

idavis:
## Summary

Renames `source/qdk_openqasm_parser` to `source/qdk_openqasm`, severs its remaining dependency on the compiler workspace, and gives it a documented, standalone public API fully vendoring repo deps. Consumers in `qsc`, `qsc_openqasm_compiler`, `qsc_project`, `qsls`, and `wasm` are updated to the new crate name, the new entry points, and the crate-local `Span`/`SourceMap` types.

The lexer now emits structured tokens for pragma and annotation directives instead of one whole-line token, and semantic analysis now keeps one gate call per source statement with broadcast width recorded as metadata, expanded on demand by consumers.

## New public surface

* Root entry points: `parse`, `parse_source`, `parse_sources`, `analyze`, `analyze_source`, `analyze_sources`, `analyze_parse_result`, all doctested.
* `source/qdk_openqasm/src/unparse.rs`: checked canonical serialization of the syntactic AST with a typed `UnparseError`.
* `source/qdk_openqasm/src/source/line_column.rs`: strict UTF-8 and UTF-16 position and range conversion returning `PositionError` rather than clamping. Line/column is to support a different type of span for the Python API to ease migration from other parsers.
* `source/qdk_openqasm/src/semantic/broadcast.rs`: `expand_gate_call` and `GateCallExpansion` for scalar expansion of compact gate calls.
* `tokens` module exposing lossless raw tokenization.
* `SourceSnapshot` and `SourceFileSnapshot` on parse and analysis results, carrying per-file id, offset, aliases, and resolved or unresolved status for diagnostic rendering. This will likely change with OpenQASM spec updates in flight.

## Behavior changes

* Broadcast is no longer expanded during lowering. One semantic gate call per source statement, with validated width metadata. Consumers expand explicitly. This is what preserves source-statement identity for breakpoints and is covered by the new debugger test. This allows us to emit loops in the future instead of eagerly generating a flat list of calls, and speeds up semantic pass.
* Directive lexing. New `DirectiveCommand`, `DirectiveValue`, and `DirectiveEnd` tokens replace whole-line pragma and annotation tokens. This removes the line-sniffing heuristics and the second semantic parse from `source/language_service/src/completion/openqasm.rs`; only box-target completion still re-analyzes.
* New diagnostics: `Qdk.Qasm.Parser.EmptyPragma`, `Qdk.Qasm.Lex.UnterminatedBlockComment`, `Qdk.Qasm.Lex.InvalidStringLiteral`, `Qdk.Qasm.Lowerer.IncludeNotResolved`, `Qdk.Qasm.Lowerer.VersionInIncludedSource`.
* Version policy. `OPENQASM 2;` without a minor version is now an error, and `OPENQASM` declarations in included sources are rejected at any depth. This will change once the OpenQASM include loading spec changes have merged.
* `real()` and `imag()` complex accessors added to the const-eval builtin table. These were missed spec built-in functions.
* `SourceMap::find_by_offset` tightened to an explicit containment check in `source/compiler/qsc_data_structures/src/source.rs`, and `longest_common_prefix` rewritten from an O(n·m) `chars().nth()` scan to a bytewise fold.

## Reviewer notes

The 245-file count is misleading; roughly 180 are pure renames. That said, there are some files that are marked as renames that also have changes that need to be reviewed. There are many type renames which were chosen to either simplify the API or make them more aligned with the OpenQASM grammar and reference parser. Concentrate on:

1. `source/qdk_openqasm/src/lib.rs`, the whole new public contract.
2. `source/qdk_openqasm/src/semantic/broadcast.rs` and `source/qdk_openqasm/src/lex/cooked.rs`, the two real behavior changes.
3. `source/compiler/qsc_openqasm_compiler/src/parser_types.rs`, the span and source-map bridge. Everything in `compiler.rs` follows from it and is mechanical.
4. `source/language_service/src/completion/openqasm.rs`, specifically `incomplete_directive_context`, the deliberately narrow heuristic retained for editor prefixes such as `@` and `#p` that are too short for the parser to classify.
5. `source/qdk_openqasm/src/unparse.rs`, is for dumps functionality in the Python API


amcasey:
> semantic analysis now keeps one gate call per source statement with broadcast width recorded as metadata, expanded on demand by consumers.

What did it do before?

> * Root entry points: `parse`, `parse_source`, `parse_sources`, `analyze`, `analyze_source`, `analyze_sources`, `analyze_parse_result`, all doctested.

Probably clear from the doc comments, but what are the three flavors of each?  String vs file vs files?

> * `source/qdk_openqasm/src/unparse.rs`: checked canonical serialization of the syntactic AST with a typed `UnparseError`.

What is unparsing?

> * `source/qdk_openqasm/src/source/line_column.rs`: strict UTF-8 and UTF-16 position and range conversion returning `PositionError` rather than clamping.

"Clamping" here means "moving it to the nearest valid position"?

> * `tokens` module exposing lossless raw tokenization.

What does lossless mean here?  Preserving comments and whitespace?

> * Version policy. `OPENQASM 2;` without a minor version is now an error, and `OPENQASM` declarations in included sources are rejected at any depth. This will change once the OpenQASM include loading spec changes have merged.

This sounds like a breaking change.  But the next sentence implies it might be a temporary state during development?

> `longest_common_prefix` rewritten from an O(n·m) `chars().nth()` scan to a bytewise fold.

What's a "bytewise fold"?

source/compiler/qsc/src/location.rs
```diff
@@ -31,11 +31,13 @@ impl Location {
...
     ) -> Self {
-        let source = package_store
+        let sources = &package_store
             .get(package_id)
             .expect("package id must exist in store")
-            .sources
+            .sources;
+        let source = sources
             .find_by_offset(span.lo)
+            .or_else(|| sources.iter().last())
```
amcasey: I don't remember this from the description.  If you ask for a position that's out of range, we just use the last file?

source/compiler/qsc_data_structures/src/source/tests.rs
```diff
@@ -0,0 +1,63 @@
...
+        (
+            &["/project/src/a.qasm", "/project/src/b.qasm"][..],
+            "/project/src/",
+        ),
+        (
+            &[r"C:\project\src\a.qasm", r"C:\project\src\b.qasm"][..],
+            r"C:\project\src\",
+        ),
+        (
+            &["file:///project/a.qasm", "file:///project/b.qasm"][..],
```
amcasey, line 23: Are we worried about the case where the longest common prefix is `"file://"`, which seems more or less equivalent to `""`?

source/compiler/qsc/src/location.rs
amcasey: Copilot plausibly guesses this is compensating for the stricter `contains_offset` check.  In which case, it suggests reviewing

The author already hit this and patched it in location.rs:33-42 by adding .or_else(|| sources.iter().last()) before the .expect(...). However, ~10 other call sites still do .find_by_offset(...).expect(...) with no such fallback and would now panic instead of silently mis-resolving, if they ever hit the same edge case:

qsc_utils.rs:9-16 (into_range, both span.lo and span.hi)
state.rs:743-746 (add_unnecessary_code_diagnostics)
error.rs:44-46 (WithSource::from_map)
debug.rs (get_source)
interpret.rs:2093-2105 (get_source_code, indentation_at_offset)
linter.rs:122-138
plus qsc_circuit/src/builder.rs, language_service/src/completion.rs, text_edits.rs

source/compiler/qsc_data_structures/src/source.rs
```diff
@@ -128,37 +131,79 @@ pub struct Source {
...
+impl Source {
+    #[must_use]
+    pub fn contains_offset(&self, offset: u32) -> bool {
+        let end = self
+            .offset
+            .checked_add(
+                u32::try_from(self.contents.len()).expect("contents length should fit into u32"),
+            )
+            .expect("source end should fit into u32");
+        (self.offset..=end).contains(&offset)
```
amcasey, line 143: Seems like another way to avoid the overflow might be to compare `offset - self.offset` to `self.contents.len()`.  But this may be so idiomatic that more experienced rust devs just read right through it.

source/compiler/qsc_data_structures/src/source.rs
```diff
@@ -128,37 +131,79 @@ pub struct Source {
...
+/// source returns its containing path, and an empty slice returns an empty
+/// string.
+///
+/// Comparison is bytewise and linear in the compared input. This is UTF-8 safe
+/// because returned slices end only after an ASCII path separator or at the end
+/// of the first source name.
 #[must_use]
 pub fn longest_common_prefix<'a>(strs: &'a [&'a str]) -> &'a str {
     if strs.len() == 1 {
         return truncate_to_path_separator(strs[0]);
```
amcasey, line 172: Why is this truncated?  The comment suggests that that only happens when the inputs aren't identical.

source/qdk_openqasm/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+    io::InMemorySourceResolver::from_iter([(path.clone(), source.clone())])
+}
+
+/// Parses a self-contained `OpenQASM` program.
+///
+/// Only the entry source and the built-in `stdgates.inc`, `qelib1.inc`, and
+/// `qdk.inc` includes are available. Any other `include` is reported as
+/// unresolved and recorded as an unresolved entry in the source snapshot.
+///
+/// Use [`parse_sources`] when the program's dependencies are already loaded, or
```
amcasey: I don't find this distinction between `sources` and `source` clear from the names.

amcasey: Maybe `parse`, `parse_all`, and `parse_and_resolve`?

source/qdk_openqasm/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+/// use qdk_openqasm::parse_sources;
+///
+/// let result = parse_sources(&[
+///     ("main.qasm".into(), "OPENQASM 3.0; include \"gates.inc\"; qubit q; my_h q;".into()),
+///     ("gates.inc".into(), "gate my_h q { h q; }".into()),
+/// ]);
+/// assert!(!result.has_errors());
+/// ```
+#[must_use]
+pub fn parse_sources(sources: &[(Arc<str>, Arc<str>)]) -> ParseResult {
```
amcasey: Are tuples idiomatic here?  I would have thought you'd want to name the members when they're both strings.

source/qdk_openqasm/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+///
+/// ```
+/// use qdk_openqasm::parse;
+///
+/// let result = parse("OPENQASM 3.0; qubit q; h q;");
+/// assert!(!result.has_errors());
+/// assert_eq!(result.program().statements.len(), 2);
+/// ```
+#[must_use]
+pub fn parse(source: impl Into<Arc<str>>) -> ParseResult {
```
amcasey, line 87: If you want this source to have a path, you have to use `parse_source` with an empty resolver instead?

source/qdk_openqasm/src/semantic/broadcast.rs
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) Microsoft Corporation.
```
amcasey, line 1: I'm not sure I can review this without knowing what a broadcast is

source/compiler/qsc_data_structures/src/source.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
```
amcasey, line 1: Does this need to stay in sync with the vendored copy and, if so, would a warning comment (or simple test) be helpful?

source/qdk_openqasm/src/lex.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// One lossless raw token with a source-local UTF-8 byte span.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct RawToken {
+    pub kind: RawTokenKind,
+    pub span: Span,
+    pub text: String,
+    pub is_trivia: bool,
+    pub detail: Option<&'static str>,
+    pub is_complete: bool,
```
amcasey, line 53: I take it this only applies to delimited tokens like strings and multi-line comments?

amcasey: Is an `Unknown` token complete?

amcasey (commented):
It's a big PR.  Here's what I have so far.

source/qdk_openqasm/src/lib.rs
idavis: Yes

source/qdk_openqasm/src/lex.rs
idavis: `is_complete` is `false` only for unterminated bitstrings/strings/invalid strings/unterminated block comments.

`Unknown` is complete.

source/qdk_openqasm/src/semantic/broadcast.rs
idavis: [Gate broadcasting](https://openqasm.com/language/gates.html#broadcasting) is used when any arguments of a gate are quantum registers instead of qubits and is a shorthand application of the gate call across the register.


```qasm
// Given:
qubit[1] qr0;
qubit[3] qr1;
qubit[2] qr2;
qubit[3] qr3;

// then
g4 qr0[0], qr1, qr2[0], qr3;

// is equivalent to:
g4 qr0[0], qr1[0], qr2[0], qr3[0];
g4 qr0[0], qr1[1], qr2[0], qr3[1];
g4 qr0[0], qr1[2], qr2[0], qr3[2];
```

source/language_service/src/completion/openqasm.rs
```diff
@@ -20,29 +20,37 @@ use crate::{
 
 use super::{Completion, Locals, into_completion_list};
 
+/// Gives parser-classified directive positions exclusive candidate sets so
```
amcasey, line 23: I think this is saying that there are (now) directive-specific completions within the span of a directive?

source/language_service/src/completion/openqasm.rs
```diff
@@ -57,166 +65,94 @@ pub(super) fn completions(
...
-/// keeps those completions out of ordinary statement positions.
-fn classify_pragma_or_annotation_line(
+/// Preserves completion for editor prefixes too incomplete for parser context.
+/// The narrow boundary keeps the parser authoritative for all other input.
+fn incomplete_directive_context(
     source_contents: &str,
     cursor_offset: u32,
-) -> Option<DirectiveLine> {
+) -> Option<qsc::openqasm::completion::CompletionContext> {
     let offset = (cursor_offset as usize).min(source_contents.len());
```
amcasey, line 74: Is the `min` because of virtual space or something?  Why would the cursor be beyond the content?

source/qdk_openqasm/src/lex/cooked.rs
```diff
@@ -68,34 +79,27 @@ impl Error {
...
-    ///  or
-    /// `#pragma REST_OF_LINE`
+    /// `pragma` or `#pragma` when it introduces a pragma.
     Pragma,
+    /// The opaque value following an annotation path.
+    DirectiveValue,
+    /// The complete opaque command following a pragma introducer.
+    DirectiveCommand,
+    /// The zero-width end of a physical-line directive.
+    DirectiveEnd,
```
amcasey, line 102: Why are there `DirectiveEnd` tokens?  I'm guessing the parser doesn't see new lines and a directive can contain arbitrary text, so only the lexer knows where it ends?

source/qdk_openqasm/src/lex/cooked.rs
```diff
@@ -758,3 +820,206 @@ impl Iterator for Lexer<'_> {
...
+                    span: Span {
+                        lo: start,
+                        hi: *cursor,
+                    },
+                },
+                false,
+            )
+        }
+        AnnotationState::Separator => {
+            if input.as_bytes().get(*cursor as usize) == Some(&b'.') && *cursor < line_end {
```
amcasey, line 953: I'm not sure I understand the line_end check.  It's not a dot if it's the last character in a line?

source/qdk_openqasm/src/lex/cooked.rs
```diff
@@ -758,3 +820,206 @@ impl Iterator for Lexer<'_> {
...
+    cursor: &mut u32,
+    line_end: u32,
+    state: &mut AnnotationState,
+) -> (Token, bool) {
+    match state {
+        AnnotationState::Segment => {
+            let start = *cursor;
+            let Some(first) = input[start as usize..line_end as usize].chars().next() else {
+                *state = AnnotationState::End;
+                return next_annotation_token(input, cursor, line_end, state);
```
amcasey, line 921: Why is this recursive, rather than iterative?  Does rust have tail call optimization?

source/qdk_openqasm/src/lex/cooked.rs
amcasey, line 953: Can you have whitespace around a dot? `a  .  b`?

source/qdk_openqasm/src/lex/cooked.rs
```diff
@@ -758,3 +820,206 @@ impl Iterator for Lexer<'_> {
...
+}
+
+fn skip_horizontal_space(input: &str, mut cursor: u32, line_end: u32) -> u32 {
+    while cursor < line_end && matches!(input.as_bytes()[cursor as usize], b' ' | b'\t') {
+        cursor += 1;
+    }
+    cursor
+}
+
+fn is_identifier_start(character: char) -> bool {
```
amcasey: I'm kind of surprised these helpers are new.  We didn't need them for identifier lexing?

source/qdk_openqasm/src/lex/raw.rs
```diff
@@ -30,6 +30,8 @@ enum NumberLexError {
     EndsInUnderscore,
     /// An incomplete binary, octal, or hex number.
     Incomplete,
+    /// A numeric literal that does not conform to the OpenQASM grammar.
+    Invalid,
```
amcasey: I guess `EndsInUnderscore` and `Incomplete` are special cases of `Invalid`?

amcasey: Actually, it looks like `Invalid` (presently?) means "has repeated separator"?

source/qdk_openqasm/src/lex/raw.rs
```diff
@@ -595,6 +643,14 @@ impl Iterator for Lexer<'_> {
     }
 }
 
+fn is_identifier_start(c: char) -> bool {
```
amcasey: Duplicated from cooked?

amcasey (commented):
Round 2.  Personally, I would have found it easier to review and test a bunch of smaller, independent PRs.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: `@a. b . c 2` used to report `unexpected annotation: @a.b.c`, but now reports `unexpected annotation: @a` and `expected identifier, found annotation value`.

source/compiler/qsc_openqasm_compiler/src/tests/statement/gate_call/functor_constraints.rs
```diff
@@ -112,6 +112,25 @@ fn inverted_gate_implements_adj_functor() {
...
+    gate test_gate q {}
+    qubit[2] q;
+    inv @ test_gate q;
+    ";
+    check_qasm_to_qsharp(
+        src,
+        &expect![[r#"
+            import Std.OpenQASM.Intrinsic.*;
+            operation test_gate(q : Qubit) : Unit is Adj {}
+            borrow q = Qubit[2];
```
amcasey, line 127: I thought the `borrow` syntax was dead?

source/compiler/qsc_openqasm_compiler/src/tests/statement/pragma/profile.rs
```diff
@@ -1,9 +1,59 @@
...
+use qsc::target::Profile;
+
+fn assert_broadcast_operation_multiplicity(profile: &str) -> miette::Result<(), Vec<Report>> {
+    let source = format!(
+        r#"
+        include "stdgates.inc";
+        #pragma qdk.qir.profile {profile}
+        qubit[3] qs;
+        output bit[3] results;
+        h qs;
```
amcasey, line 16: Would it be interesting to test a two-qubit gate?

source/language_service/src/references/openqasm_tests.rs
```diff
@@ -0,0 +1,51 @@
...
+    for target in target_spans {
+        assert!(
+            actual.contains(&target),
+            "expected {actual:?} to contain {target:?}"
+        );
+    }
+}
+
+#[test]
+fn broadcast_register_reference_occurs_once_per_source_token() {
```
amcasey, line 31: This was a concern because of the way we convert to Q# and use that LS?

source/qdk_openqasm/src/lex/cooked/tests.rs
```diff
@@ -241,6 +244,26 @@ fn int() {
             ]
         "#]],
     );
+    check(
+        "0O7",
```
amcasey, line 252: Is it expected that `0o_7` is accepted?

source/qdk_openqasm/src/lex/cooked/tests.rs
```diff
@@ -1123,6 +1190,51 @@ fn annotation() {
...
+                        kind: DirectiveValue,
+                        span: Span {
+                            lo: 9,
+                            hi: 15,
+                        },
+                    },
+                ),
+                Ok(
+                    Token {
+                        kind: DirectiveEnd,
```
amcasey, line 1294: I still don't quite understand what this is for.  It seems like the parser can already tell where the directive ends.

source/qdk_openqasm/src/parser/ast.rs
```diff
@@ -1252,6 +1256,110 @@ impl Display for Pragma {
...
+            lo: command_span.lo,
+            hi: command_span.lo + name_end,
+        }),
+        value: value_start.map(|start| &command[start..]),
+        value_span,
+    }
+}
+
+fn pragma_name_end(command: &str) -> Option<(usize, bool)> {
+    let mut characters = command.char_indices().peekable();
```
amcasey, line 1318: This is just my ignorance, but why would the parser need to decompose a token into characters?  Surely, that's the lexer's job?

source/qdk_openqasm/src/parser/expr.rs
```diff
@@ -606,8 +606,8 @@ fn infix_op(name: OpName) -> Option<InfixOp> {
             ClosedBinOp::LtLt => left_assoc(BinOp::Shl, 8),
             ClosedBinOp::GtGt => left_assoc(BinOp::Shr, 8),
             ClosedBinOp::Amp => left_assoc(BinOp::AndB, 5),
-            ClosedBinOp::Bar => left_assoc(BinOp::OrB, 4),
-            ClosedBinOp::Caret => left_assoc(BinOp::XorB, 3),
+            ClosedBinOp::Caret => left_assoc(BinOp::XorB, 4),
```
amcasey, line 609: This seems breaking (even though it's a fix)?

source/qdk_openqasm/src/parser/tests.rs
```diff
@@ -0,0 +1,486 @@
...
+        Ok(value) if errors.is_empty() => expect.assert_eq(&f(&value)),
+        Ok(value) => expect.assert_eq(&format!("{}\n\n{errors:#?}", f(&value))),
+        Err(error) if errors.is_empty() => expect.assert_debug_eq(&error),
+        Err(error) => expect.assert_eq(&format!("{error:#?}\n\n{errors:#?}")),
+    }
+}
+
+#[test]
+fn int_version_can_be_parsed() -> miette::Result<(), Vec<Report>> {
+    let source = r#"OPENQASM 3;"#;
```
amcasey, line 106: I assume it's intentional that this doesn't follow the usual int lexing rules.

amcasey (commented):
Found a bit more willpower.

source/compiler/qsc_openqasm_compiler/src/compiler.rs
```diff
@@ -722,14 +712,13 @@ impl QasmCompiler {
             semast::StmtKind::ExternDecl(stmt) => self.compile_extern_stmt(stmt),
             semast::StmtKind::For(stmt) => self.compile_for_stmt(stmt),
             semast::StmtKind::If(stmt) => self.compile_if_stmt(stmt),
-            semast::StmtKind::GateCall(stmt) => self.compile_gate_call_stmt(stmt),
-            semast::StmtKind::Include(stmt) => self.compile_include_stmt(stmt),
-            semast::StmtKind::IndexedClassicalTypeAssign(stmt) => {
-                self.compile_indexed_classical_type_assign_stmt(stmt)
+            semast::StmtKind::GateCall(stmt) => {
+                self.compile_gate_call_stmt(stmt, output);
```
amcasey, line 718: I would find a comment helpful here:
```
// Since broadcasting might result in multiple statements being generated, we update `output` ourselves and return early
```

source/qdk_openqasm/src/parser/completion/collector.rs
```diff
@@ -97,12 +106,38 @@ impl ValidWordCollector {
         }
     }
 
+    /// The parser is interpreting the cursor as a directive-specific position.
```
amcasey: I'm not sure I understand this comment as the function doesn't seem tied to directives.

source/qdk_openqasm/src/parser/expr.rs
idavis: This is a bug fix and there is an associated test for it.

source/compiler/qsc_data_structures/src/source.rs
idavis: I've created vendoring scripts to automate and test

source/qdk_openqasm/src/parser/tests.rs
idavis: The version isn't an int

source/compiler/qsc_openqasm_compiler/src/tests/statement/pragma/profile.rs
idavis: Using a two qubit gate wouldn't effect anything.

source/compiler/qsc_openqasm_compiler/src/tests/statement/gate_call/functor_constraints.rs
idavis: It is dead from a Q# perspective but we leverage it internally to control release of qubits.

source/qdk_openqasm/src/parser/expr.rs
amcasey: But it's going to change the behavior or existing code, right?

source/compiler/qsc/src/location.rs
idavis: The tightening changes exactly one case. An offset past the end of the last source used to resolve to that last source, and now returns `None`. Everything inside a source resolves identically.

Two classes of caller can reach that case. Editor-supplied positions are clamped upstream to the inclusive bound before they get here, so they cannot exceed it. The other class is compiler-internal spans, and the one instance I found is root-caused by the evaluator callee-span fix in this PR.

I considered deleting the fallback outright once the root cause was fixed, but removing it would turn any unproven production path from a wrong answer into a panic. We can consider using debug asserts for any bad span checks in the future.

source/compiler/qsc_data_structures/src/source/tests.rs
idavis: When two `file://` URIs share only the scheme there is nothing to shorten, so the display name stays essentially the whole URI.
 
 Your comment did turn up a real defect when I reviewed the code again. It searched for the last `\` only when the name contained no `/` anywhere. A name mixing both truncated at the wrong one, so `src/dir\a.qasm` and `src/dir\b.qasm` produced the prefix `src/` and left `dir\` in both display names. This is now fixed.

source/compiler/qsc_data_structures/src/source.rs
idavis: This is idiomatic for an inclusive range. `offset - self.offset` will lead to value underflows for most calls to `contains_offset`.

source/compiler/qsc_data_structures/src/source.rs
idavis: I updated the docs. It is now stated as three cases:

* An empty slice returns an empty string.
* A single source name returns its containing path.
* Two or more source names return the text they all share, truncated the same way, unless the first name is a prefix of every other name, in which case the whole first name is returned. Identical names are the common case of that exception.

Truncating is what keeps the result from ending in a partial path component, and that sentence is now in the doc too.

source/language_service/src/completion/openqasm.rs
idavis: Yes

source/language_service/src/completion/openqasm.rs
idavis: Safety for live document vs snapshot offset making sure we aren't passed the end of the line.

source/qdk_openqasm/src/lex/cooked.rs
idavis: Yes, the cooked stream discards trivia and directives are terminated by a physical line end

source/qdk_openqasm/src/lex/cooked.rs
idavis: I think you're misreading the boundary. It's `<`, not `<=`

source/qdk_openqasm/src/lex/cooked.rs
idavis: Rust doesn't guaranteed tail call elimination. That doesn't matter here, because the recursion is bounded at three frames as a state transtion. The state machine only ever moves forward, and the two transitions that could repeat don't recurse:

- Segment emits an identifier and returns, leaving state on Separator.
- Separator sees a dot, emits it, and returns, leaving state on Segment.

So `@a.b.c.d.e` doesn't nest at all. Each token is a fresh call from the iterator. Recursion only happens on the transitions that produce no token, and those strictly descend: `Segment` => `Value` => `End`, or `Separator` => `Value` => `End`. Three frames is the cap regardless of input.

source/qdk_openqasm/src/lex/cooked.rs
idavis: You [cannot have spaces](https://openqasm.com/language/directives.html#pragma-and-annotation-namespacing). The new behavior is better IMHO.

source/qdk_openqasm/src/lex/cooked.rs
idavis: It is because we weren't processing the namespaced directive correctly before.

source/qdk_openqasm/src/lex/raw.rs
idavis: Good call. I renamed `Incomplete` => `IncompleteRadix`, `Invalid` => `IncompleteSeparator`, `None` => `NotApplicable`

source/qdk_openqasm/src/lex/raw.rs
idavis: Yup, unified now.

source/language_service/src/references/openqasm_tests.rs
idavis: When we process the broadcast behind the scenes we need to make sure it works with the debugger.

source/qdk_openqasm/src/lex/cooked/tests.rs
idavis: Fixed test cases that got merged during cleanup. Upper and lower tests are in there now.

source/qdk_openqasm/src/lex/cooked/tests.rs
idavis: Answered elsewhere

source/compiler/qsc_data_structures/src/source/tests.rs
amcasey: I think you're saying that the existing caller(s) will handle that case correctly?

source/compiler/qsc_data_structures/src/source.rs
amcasey: Sorry, I'm used to dotnet where no one ever uses unsigned integers.

source/compiler/qsc_data_structures/src/source.rs
amcasey: Maybe it should be called something like `longest_common_folder_prefix` if cutting off the file component is a key part of the functionality?

source/qdk_openqasm/src/lex/cooked/tests.rs
amcasey: No harm in testing case sensitivity, but I was actually asking about the leading underscore.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: So it's for the cooking lexer, rather than for the parser?

source/qdk_openqasm/src/parser/completion/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+    CompletionContext, CompletionDirective, completion_at_offset_in_source,
+    possible_words_at_offset_in_source,
+};
+use expect_test::expect;
+
+fn get_source_and_cursor(input: &str) -> (String, u32) {
+    let mut cursor = -1;
+    let mut source = String::new();
+    for c in input.chars() {
+        if c == '|' {
```
amcasey: Nit: I think we use an arrow symbol for the cursor in other tests

source/qdk_openqasm/src/parser/ast.rs
idavis: This is a spec extension derivation of an 'opaque' value that we try to interpret for our UI extension. This is called from `derive_pragma_command` which isn't part of the spec but part of a 'guideline'.

source/qdk_openqasm/vendor-sync/manifest.json
```diff
@@ -0,0 +1,65 @@
...
+      "vendored": "source/qdk_openqasm/src/vendor/source/tests.rs",
+      "origin": "source/compiler/qsc_data_structures/src/source/tests.rs",
+      "tier": "copy",
+      "public": false
+    },
+    {
+      "name": "error.rs",
+      "vendored": "source/qdk_openqasm/src/vendor/error.rs",
+      "origin": "source/compiler/qsc_data_structures/src/error.rs",
+      "origin_sha256": "e1ed05600d31e7836e8aa55b3783b80e5e8714a12925276b0a51b38a191bf554",
```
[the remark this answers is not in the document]
amcasey: This is clearly not a token, but I'm curious why this file has a sha when the others (e.g. source/test.rs) appear not to.

source/qdk_openqasm/src/lib.rs
amcasey: Sounds like the path is only useful for `#include` and the file isn't going to include itself.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: I had assumed `line_end` was the first character not included - is that not the case?  The intended reading is that there must be at least one character after the dot?

source/qdk_openqasm/src/lex/cooked.rs
amcasey: I would find iteration clearer, but I don't feel strongly about it.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: But we did have identifiers before.  I would have thought the existing identifier lexing would need these.

source/compiler/qsc_openqasm_compiler/src/tests/statement/pragma/profile.rs
amcasey: You know that, but the person who's learning how this works from the tests won't.  I'm fine with omitting it.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: I'm fine with disallowing spaces, but this seems like it will break programs that work now.

source/qdk_openqasm/src/lex/cooked/tests.rs
idavis: `OctalIntegerLiteral: '0o' ([0-7] '_'?)* [0-7];`

Underscores are only permitted between digits

source/compiler/qsc_data_structures/src/source/tests.rs
idavis: In your example they would always share `/`. The only way to share just the scheme is if they have different drive letters which can't happen with the way we load files. Changing the code to be fully URI aware is a big change.

source/qdk_openqasm/src/lex/cooked/tests.rs
amcasey: I was asking because I tried it and didn't get an error, but I may well have done something silly.

amcasey: On my box, this test passes, which surprises me:
```rust
fn octal_int_uppercase_prefix_leading_underscore() {
    check(
        "0O_7",
        &expect![[r#"
            [
                Ok(
                    Token {
                        kind: Literal(
                            Integer(
                                Octal,
                            ),
                        ),
                        span: Span {
                            lo: 0,
                            hi: 4,
                        },
                    },
                ),
            ]
        "#]],
    );
}
```

amcasey: I don't see anything in [eat_number_characters](https://github.com/microsoft/qdk/pull/3553/changes#diff-e451f55fbbc1661d3943e09d5050defe06552ebc2bda45fbf90c5d898f6cf9b9R520) that would prevent a leading underscore.

amcasey (approved):
The language service and directive parsing changes seem fine.

source/qdk_openqasm/src/lex/cooked.rs
amcasey: I'm told there aren't enough consumers for this to matter.

source/qdk_openqasm/src/parser/completion/tests.rs
idavis: Updated. I was keeping it simple since all of those tests are in the language service, but it makes sense for uniformity.

source/qdk_openqasm/src/lex/cooked.rs
idavis: Both, the lexer creates the tokens consumed by the parser

source/qdk_openqasm/src/lex/cooked.rs
idavis: `line_end` is the offset of the newline character

source/qdk_openqasm/src/parser/completion/collector.rs
idavis: Good catch, fixed.

source/compiler/qsc_openqasm_compiler/src/tests/statement/const_eval.rs
```diff
@@ -491,53 +491,6 @@ fn binary_op_shl_bitarray() -> miette::Result<(), Vec<Report>> {
     Ok(())
 }
 
-#[test]
-fn binary_op_shl_creg_fails() {
-    let source = r#"
-        const creg a[3] = "101";
-        const creg b[3] = a << 2;
-        bit[b] r;
-    "#;
```
orpuente-MS, line 500: Are we allowing this operation now? Is there a positive test for it?

idavis: This code was never valid and I moved the test, it isn't because of the shift, it was syntax as const creg isn't allowed.

source/qdk_openqasm/src/semantic/tests/expression/binary.rs
```diff
@@ -0,0 +1,28 @@
...
+mod comparison;
+mod complex;
+mod duration;
+mod ident;
+
+use crate::semantic::{ast::StmtKind, tests::check_map};
+use expect_test::expect;
+
+#[test]
+fn bitwise_xor_binds_more_tightly_than_or() {
```
orpuente-MS, line 14: Should there be a similar test for bitwise `^` and `&`?

source/qdk_openqasm/src/semantic/tests/expression/builtin_functions/complex_accessors.rs
```diff
@@ -0,0 +1,319 @@
...
+           ,-[test:1:1]
+         1 | imag(1im, 2im);
+           : ^^^^^^^^^^^^^^
+           `----
+        ]"#]],
+    );
+}
+
+#[test]
+fn real_rejects_nonconstant_input() {
```
orpuente-MS: I see you have these tests for `real`:
```
real_rejects_non_complex_input
real_rejects_nonconstant_input
```

and these tests for `imag`:
```
imag_rejects_zero_arguments
imag_rejects_multiple_arguments
```

They are disjoint tests. Should we have 8 tests in total? So that `real` and `imag` have tests for all 4 of those properties.

idavis: Done

source/qdk_openqasm/src/semantic/tests/expression/binary.rs
idavis: This was the original test that showed the bug. There is a larger test suite for precedence now, so we can keep or remove this.

source/qdk_openqasm/src/semantic/tests/expression/builtin_functions/complex_accessors.rs
```diff
@@ -0,0 +1,445 @@
...
+        ]"#]],
+    );
+}
+
+#[test]
+fn imag_rejects_nonconstant_input() {
+    let source = "
+        complex value = 1.0 + 2.0 im;
+        imag(value);
+    ";
```
orpuente-MS, line 358: The spec allows accessing the `real` and `imag` components of non-const values:

https://openqasm.com/language/types.html#complex-numbers

idavis: Great catch

idavis: Fixed

orpuente-MS (approved):

PR #3554, opened 2026-08-07
Introduce command and tools for venv creation

amcasey:
Creates or updates a venv in the workspace root with "qdk[jupyter]", "ipympl", and "ipykernel".

The tool behaves a bit differently from the command:
- signals failures with exceptions
- uses a heuristic to choose the appropriate workspace root folder
- hard-codes the package list since there's no inline prompt

The command version lets you tweak the list of packages to include "qdk[azure]" and/or "qdk-chemistry".


amcasey:
Jupyter discovers the venv automatically (since it's in a workspace root) but only if the notebook is in the workspace (i.e. not for an unsaved temp file).

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,192 @@
...
+
+function isUnderWorkspaceRoot(
+  env: PythonEnvironment,
+  root: vscode.Uri,
+): boolean {
+  const envStr = env.environmentPath.toString();
+  const rootStr = root.toString();
+  return envStr === rootStr || envStr.startsWith(rootStr + "/");
+}
+
```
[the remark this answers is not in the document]
amcasey: Since both values come from VS Code-provide uris, I would expect them to match the casing on disk.  They do have forward slashes.  `toLowerCase` isn't safe because of the Turkish-I problem.

source/vscode/src/gh-copilot/tools.ts
```diff
@@ -11,6 +11,7 @@ import { QSharpTools } from "./qsharpTools";
 import { CopilotToolError } from "./types";
 import { ToolState } from "./azureQuantumTools";
 import type { LearningService } from "../learning/index";
+import { createQuantumVenv, findWorkspaceVenv } from "../pythonEnvs.js";
```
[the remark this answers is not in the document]
amcasey: `@vscode/python-environments` is a type-only import.

source/vscode/package.json
```diff
@@ -638,6 +642,11 @@
         "category": "QDK",
         "title": "Create a Microsoft Quantum notebook"
       },
+      {
+        "command": "qsharp-vscode.createQuantumNotebookVenv",
+        "category": "QDK",
+        "title": "Create a Microsoft Quantum notebook virtual environment"
+      },
```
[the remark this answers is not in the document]
amcasey: I don't see any other uses of `onCommand` and the command seems to work in practice.

source/vscode/src/extension.ts
```diff
@@ -38,6 +38,7 @@ import {
   registerChangelogCommand,
 } from "./changelog.js";
 import { getPlatformEnv, toVsCodeRange } from "./common.js";
+import { createQuantumVenvCommand } from "./pythonEnvs.js";
```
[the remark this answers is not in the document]
amcasey: `@vscode/python-environments` is a type-only import.

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,192 @@
...
+// Fixed set for the Copilot tool.
+const toolPackages = ["qdk[jupyter]", "ipympl", "ipykernel"];
+
+// All packages offered in the command palette picker.
+const packagePickItems: vscode.QuickPickItem[] = [
+  { label: "qdk", picked: true },
+  { label: "qdk[azure]", picked: false },
+  { label: "qdk[jupyter]", picked: true },
+  { label: "qdk-chemistry", picked: false },
```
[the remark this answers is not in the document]
amcasey: There's no support for nested lists, so there's no way to express that azure and jupyter are extras (other than an additional layer of prompting, which seemed excessive).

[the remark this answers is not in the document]
amcasey: One could argue that we should coalesce them before handing them to the package manager, but it seems work.

amcasey:
The fix for the unrelated integration test failure is here: https://github.com/microsoft/qdk/pull/3555

amcasey:
Can we include descriptive text in the multi-select menu?

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+} from "@vscode/python-environments";
+import { CopilotToolError } from "./gh-copilot/types.js";
+
+const pythonEnvsNotInstalledMsg = `The Python Environments extension (${EXTENSION_ID}) is not installed or is disabled.`;
+
+// Fixed set for the Copilot tool.
+const toolPackages = ["qdk[jupyter]", "ipympl", "ipykernel"];
+
+// All packages offered in the command palette picker.
+const packagePickItems: vscode.QuickPickItem[] = [
```
amcasey, line 16: @billti Please have a look at these package descriptions and let me know if they're what you had in mind.

source/vscode/package.json
```diff
@@ -638,6 +642,11 @@
         "category": "QDK",
         "title": "Create a Microsoft Quantum notebook"
       },
+      {
+        "command": "qsharp-vscode.createQuantumVenv",
+        "category": "QDK",
+        "title": "Create a Microsoft Quantum virtual environment"
```
billti: Should probably mention Python in here (e.g. `Create a Microsoft Quantum Python virtual environment`)

source/vscode/package.json
```diff
@@ -1522,6 +1531,26 @@
...
+      {
+        "name": "qdk-create-notebook-venv",
+        "tags": [
+          "qdk",
+          "python",
+          "jupyter",
+          "venv"
+        ],
+        "toolReferenceName": "qdkCreateNotebookVenv",
+        "displayName": "Create Quantum Notebook Virtual Environment",
```
billti: Again, should probably mention "Python"

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+    const picked = await vscode.window.showWorkspaceFolderPick({
+      placeHolder: "Select the workspace folder for the virtual environment",
+    });
+    if (!picked) return;
+    root = picked.uri;
+  }
+
+  await api.refreshEnvironments(root);
+  const existingEnvs = await api.getEnvironments(root);
+  const existingEnv = existingEnvs.find((env) =>
```
billti: I often have several Python venvs at the root of my workspace, e.g. for testing different Python versions or installed packages. Does this make the assumption there will only be one? (Seems like it will just return the first if there are multiple. It should probably ask the user to choose)

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+  {
+    label: "qdk[jupyter]",
+    description: "QDK optional support for Jupyter notebooks",
+    detail: "Enable magic commands and rich output in Jupyter notebooks",
+    picked: true,
+  },
+  {
+    label: "qdk[qiskit]",
+    description: "QDK optional support for Qiskit",
+    detail: "Interop with Qiskit via qdk.cirq",
```
billti: Cut/paste error with cirq instead of qiskit in the detail text.

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+    picked: false,
+  },
+  {
+    label: "qdk[cirq]",
+    description: "QDK optional support for Cirq",
+    detail: "Interop with Cirq via qdk.cirq",
+    picked: false,
+  },
+  {
+    label: "qdk[jupyter]",
```
billti, line 36: Are these listed in the order shown? Feels like we should have the most likely/useful first (e.g. `qdk[jupyter]`). We may also want the `qre` extra in the list, being that will be a bigger focus going forward.

billti: Note: I think the packages you've defaulted to 'true' make sense, and the default selected packages should probably be all at the top of the list.

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+    description: "QDK optional support for Jupyter notebooks",
+    detail: "Enable magic commands and rich output in Jupyter notebooks",
```
billti: Maybe "Enable Q# code cells and interactive quantum widgets in Jupyter notebooks" is more useful and compelling

source/vscode/src/gh-copilot/tools.ts
```diff
@@ -176,6 +177,24 @@ const toolDefinitions: {
       },
     }),
   },
+  // ─── Python environment tools ───
+  {
+    name: "qdk-create-notebook-venv",
+    tool: async () => await createQuantumVenv(),
+    confirm: async () => {
+      const exists = await findWorkspaceVenv();
```
billti: The logic here seems a little disjointed. `findWorkspaceVenv` looks to see if there is a venv for the active document's workspace. But `createQuantumVenv` doesn't necessarily use that workspace - if there are multiple it asks the user to choose which workspace. So asking the user if they want to update or create the environment here doesn't make sense, as we don't know which workspace (or venv if they have multiple in the workspace) to ask about.

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (existingEnv) {
+      await api.managePackages(existingEnv, { install, upgrade: true });
+      vscode.window.showInformationMessage(
+        "Quantum notebook packages updated in existing environment.",
+      );
+      return;
+    }
+
+    // Quick create so the user isn't prompted
+    const env = await api.createEnvironment(root, { quickCreate: true });
```
billti: You can just pass the packages to be installed in the options too as `additionalPackages`, rather than the separate step below. (Though maybe cleaner or more reliable to do separately - not sure)

billti: Also, does this just automatically call it `.venv`? Looking through thier code, I'm not seeing where you can name the venv that gets created.

source/vscode/src/pythonEnvs.ts
```diff
@@ -0,0 +1,255 @@
...
+  );
+
+  if (existingEnv) {
+    const choice = await vscode.window.showQuickPick(
+      ["Update existing environment", "Cancel"],
+      {
+        placeHolder: "A virtual environment already exists at this workspace.",
+      },
+    );
+    if (choice !== "Update existing environment") return;
```
billti: How does this code work if I have an existing venv, but I don't want to use that, I want to create a new venv with the necessary packages?

source/vscode/src/pythonEnvs.ts
amcasey: I agree that it shouldn't silently ignore the fact that there are multiple, but the tool isn't really aimed at people who can already juggle multiple virtual environments.

Having said that, if someone already has multiple environments, it seems more likely to me that they want a fresh one for this than that they want to select an existing one to update.

source/vscode/src/pythonEnvs.ts
amcasey: They're listed in display order.  My reasoning was QDK core, followed by its extras, followed by QDK Chemistry, with notebook support last.

I'm not sure I agree with putting the True ones first since that would split up the list of QDK extras.  I guess we could have 

1. [x] ipykernel
2. [x] ipympl
3. [x] qdk
4. [x] qdk[jupyter]
5. [ ] qdk[azure]
6. [ ] qdk[cirq]
7. [ ] qdk[qiskit]
8. [ ] qdk-chemistry

I don't love it, but it does put all the true ones at the top.

source/vscode/src/gh-copilot/tools.ts
amcasey: Note that `createQuantumVenv` and `createQuantumVenvCommand` have different strategies.  In the absence of the ability to prompt for a selection within the chat (and I thought showing the palette picker disrupted the flow), the tool version uses the first workspace root.

I agree that the tool experience isn't great in advanced cases and I'd be okay with removing that support entirely (though there's something to be said for the 90% case of a user with one folder and no virtual environment who just wants to ask copilot how they should proceed).

source/vscode/src/pythonEnvs.ts
amcasey: I'll test `additionalPackages`, but the API isn't very mature and I'm pretty sure this is the way they recommended doing it.

amcasey: They call it .venv or .venv-N, according to whether there's an existing one.  That's why we search first.  I'm also not aware of a way to name it (and tooling support for other names isn't great).

source/vscode/src/pythonEnvs.ts
amcasey: There isn't presently a way to do that.

source/vscode/src/pythonEnvs.ts
amcasey: If you pass `quickCreate: false`, they show their own wizard experience, which does include a way to name the venv, but it conflicts with our package selection mechanism (they try to direct users to pypi and/or discover requirements files in the workspace).

amcasey: Here's the naming logic: https://github.com/microsoft/vscode-python-environments/blob/dbb10ecbe1c09e4068dfc917ccaa6bfd309b1023/src/managers/builtin/venvUtils.ts#L481

source/vscode/src/pythonEnvs.ts
amcasey: I've added more logic and handling for the case where there are multiple existing environments, but I'm not really interested in building out a custom creation story (subject to change if the API support gets better) since there's already a VS Code extension (not to mention a fleet of command line tools) for doing that.  This is intended to be a quick way to get the basic packages if you don't really know python and just want to focus on quantum programming / chemistry.

source/vscode/src/pythonEnvs.ts
amcasey: Added "active environment" as a tie breaker and more logging around just picking one.

billti (approved):

source/vscode/src/pythonEnvs.ts
amcasey: Leaving it for now.

PR #3556, opened 2026-08-07
[STIM] Support Generalized Pauli Product Gates

joao-boechat:
This PR adds QIR compiler support for Stim's generalized Pauli-product gates: `MPP`, `SPP`, and `SPP_DAG`. For more info on how they work, I recommend the official [stim docs](https://github.com/quantumlib/Stim/blob/main/doc/gates.md#MPP).

## High-level changes

- **Lexer:** Starts treating Pauli targets (e.g. `X1`, `Y1`, `Z1`) and loss targets (e.g. `L1`) as separate tokens, rather than the general "instruction_name".
- **Parser:** Groups `*`-connected Pauli targets into a single `PauliProduct` AST node. It combines factor negations with XOR and represents each term as a `PauliFactor`.
- **Compiler:** Adds shared validation, canonicalization, and emission for generalized Pauli products.

## Canonicalization

Before emission, the compiler canonicalizes each product by sorting factors by qubit, multiplying repeated same-qubit Paulis, and tracking the resulting phase `i^k`. Odd phases are rejected because they produce an anti-Hermitian product. A `-1` phase negates the canonical product, while factors reducing to identity are removed.

## Decomposition

To emit the product, the compiler:

1. Rotates each Pauli factor into the Z basis.
2. Accumulates the product parity onto one focus qubit using CNOTs.
3. Performs the central operation:
   - `MPP`: measure the focus qubit.
   - `SPP`: apply `S`.
   - `SPP_DAG`: apply `S_DAG`.
4. Reverses the CNOTs and basis rotations.

## Inversion logic

For `MPP`, negation flips the measurement result. For phase gates, negation reverses the phase direction:

```rust
let adjoint = (instruction.name == "SPP_DAG") ^ negated;
```

## Identity products

An `MPP` target whose factors all cancel measures the identity, which is deterministic but still has to append a result to the measurement record. We don't have an instruction that appends to the measurement record without acting on a qubit, so these targets are rejected for now. Once that instruction exists, `MPP` will be able to support them.

`SPP` and `SPP_DAG` don't have this limitation: phasing the identity is a global phase, so those targets are emitted as no-ops.


source/compiler/stim_compiler/src/parser.rs
```diff
@@ -539,43 +580,58 @@ impl<'a> Parser<'a> {
         })
     }
 
-    fn is_target_start(&self, token: &Token) -> bool {
+    fn is_target_start(token: &Token) -> bool {
```
fedimser: This name of this function is incorrect, because it returns `true` for tokens that are not target starts. 
I would suggest renaming it to something like `should_parse_target`. I would also suggest adding short documentation-style comments (even for private functions) explaining in plain English what this function is doing.

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -515,7 +556,7 @@ impl<'a> Parser<'a> {
         }
 
         while let Some(token) = self.peek() {
-            if !self.is_target_start(&token) {
+            if !Self::is_target_start(&token) {
```
fedimser: Can it be made simpler? We already check whether token is valid start of target in `parse_target`. So here it would be sufficient to loop until the end of the instruction. Can we just loop until newline here?
(This is just a suggestion, the code here is correct modulo misleading `is_target_start` name)

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1544,6 +1583,106 @@ impl<'noise> Compiler<'noise> {
...
+            }
+        }
+
+        if phase % 2 != 0 {
+            // a phase of i or -i makes the product anti-Hermitian, so it has no measurable eigenvalues
+            self.push_error(Error::AntiHermitianPauliProduct { span: target.span });
+            return None;
+        }
+        if canonical_factors.is_empty() && instruction.name == "MPP" {
+            // TODO: an empty product measures the identity, which needs support for appending to measurement records
```
fedimser, line 1630: Do you expect this TODO will be addressed in the future, or should we just forbid MPP with empty product as this doesn't make much sense?
Would it be better to raise a specific error rather than generic UnsupportedTarget?

fedimser (approved):
In general, I would suggest adding a comment for each function explaining what it's doing (when it's not obvious from the name), which would make it easier to review. For example, `decompose_pauli_product` doesn't literally decompose Pauli product, but decomposes an instruction whose target is Pauli product.

source/compiler/stim_compiler/src/lex.rs
```diff
@@ -216,12 +220,9 @@ impl<'a> Lexer<'a> {
 
     fn scan_identifier(&mut self, lo: usize) -> TokenKind {
         self.eat_while(|c| c.is_alphanumeric() || c == '_');
-        let hi: usize = self
-            .chars
-            .peek()
-            .map_or(self.input_len as usize, |(i, _)| *i);
-        // TODO: What if some identifier starts with "rec" but is not a rec token?
```
ScottCarda-MS, line 223: Did this TODO get addressed? Or is it not a concern anymore?

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -161,7 +164,16 @@ impl Display for TargetKind {
                 }
             }
             TargetKind::Loss { value } => write!(f, "Loss({value})"),
-            TargetKind::Combiner => write!(f, "Combiner"),
+            TargetKind::PauliProduct { negated, factors } => {
```
ScottCarda-MS: So this will end up looking like
```
PauliProduct(-Pauli(X 3)*Pauli(Y 4)*Pauli(Z 2))
```
Is that correct?

Is it every possible to have the negative attached to multiple terms?

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -173,6 +185,23 @@ pub enum Pauli {
...
+    /// Multiplies two Paulis acting on the same qubit. Returns the resulting Pauli (`None` when
+    /// they cancel to the identity) and the exponent `k` of the accompanying phase `i^k`.
+    pub fn multiply(self, other: Pauli) -> (Option<Pauli>, u8) {
+        use Pauli::{X, Y, Z};
+        match (self, other) {
+            (X, X) | (Y, Y) | (Z, Z) => (None, 0),
+            (X, Y) => (Some(Z), 1),
+            (Y, Z) => (Some(X), 1),
+            (Z, X) => (Some(Y), 1),
+            (Y, X) => (Some(Z), 3),
```
ScottCarda-MS, line 171: Is the `3` meant to represent a negation?

source/compiler/stim_compiler/src/parser/tests/targets.rs
```diff
@@ -166,23 +166,271 @@ fn negated_pauli_target() {
...
+                    tag: <none>
+                    args: <empty>
+                    targets:
+                        Target [4-10]:
+                            kind: PauliProduct(-Pauli(X 0)*Pauli(Y 1))"#]],
+    );
+}
+
+#[test]
+fn pauli_product_negated_second_factor() {
```
ScottCarda-MS, line 237: Something about this seems off to me. The negation is on the Y1 target, and if this was a naked Pauli the negative sign would be _inside_ the Pauli string, not in front of it: `Pauli(-Y 1)`.

ScottCarda-MS: Shouldn't the `PauliProduct` be displayed in a way that reflects this hierarchy?
i.e. `PauliProduct(Pauli(X 0)*Pauli(-Y 1))`

ScottCarda-MS: Processing the negations during the parsing (i.e. cancelling out even negations) seems like we are doing evaluation at the wrong step of the process.

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
```diff
@@ -0,0 +1,1569 @@
...
+        !5 = !{i32 5, !"float_computations", !{!"double"}}
+        !6 = !{i32 7, !"backwards_branching", i2 3}
+        !7 = !{i32 1, !"arrays", i1 true}
+    "#]],
+    );
+}
+
+#[test]
+fn mpp_negated_single_pauli_yields_expected_qir() {
+    // same as MZ !5
```
ScottCarda-MS, line 134: Should the "5" in "!Z5" appear somewhere in the qir?

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
```diff
@@ -0,0 +1,1569 @@
...
+fn mpp_negated_product_yields_expected_qir() {
+    // order of qubit indices in the output is not guaranteed to match the order of qubit indices in the input
+    check(
+        "MPP !Z3*Z4*Z5",
+        &expect![[r#"
+        define i64 @ENTRYPOINT__main() #0 {
+          call void @__quantum__rt__initialize(ptr null)
+          call void @__quantum__qis__cx__body(ptr inttoptr (i64 0 to ptr), ptr inttoptr (i64 1 to ptr))
+          call void @__quantum__qis__cx__body(ptr inttoptr (i64 2 to ptr), ptr inttoptr (i64 1 to ptr))
+          call void @__quantum__qis__x__body(ptr inttoptr (i64 1 to ptr))
```
ScottCarda-MS, line 326: Does the order of `CX` qubits matter? And does it matter which qubit the negating `X` is performed on? It doesn't seem to be related to the term that specified the negation.

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
```diff
@@ -0,0 +1,1569 @@
...
+        !7 = !{i32 1, !"arrays", i1 true}
+    "#]],
+    );
+}
+
+#[test]
+fn mpp_repeated_qubit_folding_to_minus_one_negates_result() {
+    // X0*Y0 = iZ0 and X1*Y1 = iZ1, so the product is i^2 Z0*Z1 = -Z0*Z1.
+    check(
+        "MPP X0*Y0*X1*Y1",
```
ScottCarda-MS, line 508: What about a test for `MPP X0*Y0`? I'm curious how it would handle the `iZ0` resolution.

source/compiler/stim_compiler/src/parser.rs
joao-boechat: Good point, I was able to remove this function which simplifies things

source/compiler/stim_compiler/src/parser.rs
joao-boechat: There were a couple of cases where the error message was slightly improved by having the code this way, but upon reviewing them again, I realized that the improvements were minor and not really that clear. For example, `H 0 )` would say "expected newline" and now it says "expected valid target". In the end, both error messages are correct, so we can just simplify this code.

Since instructions can also start blocks (e.g. `REPEAT {`), I loop until I see a newline or an open brace.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: To clarify a possible misunderstanding: this check isn't for the case where MPP has no targets, as that's already rejected by the parser. This is for checking when the factor canonicalization results to an identity (e.g. `MPP X0*X0`). In that case, I'll just be replicating stim behavior, which isn't throwing an error but adding a deterministic value to measurement record. Currently, our simulators don't support that, that's why I left the TODO, but I plan on implementing it very soon.

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
```diff
@@ -0,0 +1,1569 @@
...
+        !5 = !{i32 5, !"float_computations", !{!"double"}}
+        !6 = !{i32 7, !"backwards_branching", i2 3}
+        !7 = !{i32 1, !"arrays", i1 true}
+    "#]],
+    );
+}
+
+#[test]
+fn mpp_product_folding_to_identity_yields_error() {
+    // this is temporary
```
ScottCarda-MS, line 793: Is this comment left over from something?

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
ScottCarda-MS: Never mind, I found it: mpp_product_with_imaginary_phase_yields_anti_hermitian_error

source/compiler/stim_compiler/src/lex.rs
joao-boechat: I realized it's not a concern because we can treat "rec" as a reserved keyword. But any longer identifier starting with rec (such as "record") will be tagged as an identifier, not a rec, which is expected behavior.

source/compiler/stim_compiler/src/parser.rs
joao-boechat: The parser will consolidate negation in different terms into a single general negation.

source/compiler/stim_compiler/src/parser.rs
joao-boechat: That number represents k in i^k, as per the comment above the function. So 3 would be -i

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
joao-boechat: The compiler "allocates" a separate qubit id for each new qubit, but it doesn't guarantee that those ids will be the same as the ones in stim, nor that the order will be the same. Regardless, correctness is guaranteed that way. In my other PR I make this difference clearer by using the types QubitId and StimQubitId, but essentially that's how it works.

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
joao-boechat: Answered that above!

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
joao-boechat: I have a test named "mpp_product_with_imaginary_phase_yields_anti_hermitian_error" for exactly that case

source/compiler/stim_compiler/src/qir/tests/generalized_pauli_product_gates.rs
joao-boechat: No, this is a temporary behavior that will disappear in a PR that's coming soon. I just left the comment there to remind myself of removing this test case.

source/compiler/stim_compiler/src/parser/tests/targets.rs
ScottCarda-MS: Looks like you've changed this to keep the negative of the individuals Pauli targets, and combine negatives in the `canonicalize_pauli_product` step.

ScottCarda-MS (approved):

PR #3560, opened 2026-08-10
Bump cargo crate versions

swernli:
This change is just updates to cargo crate versions, both with direct updates to our cargo.toml listed versions of `bitflags`, `nalgebra`, `paulimer`, and `rand`, as well as via `cargo update` to pull in minor version updates to the lockfile. Interestingly, this leads to a nice performance bump across some simulation benchmarks, most notably "Bench 5x5" (~25% improvement) and "Array literal evaluation" (~55% improvement). The sparse simulator in particular shows ~60% perf improvement in gate benchmarks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,8 +67,6 @@ js-sys = "0.3"
 libfuzzer-sys = "0.4"
 log = "0.4"
 miette = { version = "7.6" }
-# Explicitly use matrixmultiply transitive dependency to disable avx512 default feature
-matrixmultiply = { version = "0.3", default-features = false, features = ["std"] }
 thiserror = "2.0"
 nalgebra = { version = "0.35" }
 ndarray = "0.17"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -68,13 +68,13 @@ libfuzzer-sys = "0.4"
...
 thiserror = "2.0"
-nalgebra = { version = "0.34" }
+nalgebra = { version = "0.35" }
 ndarray = "0.17"
 num-bigint = "0.4"
 num-complex = "0.4"
 num-derive = "0.4"
 num-traits = "0.2"
-paulimer = "0.1.2"
+paulimer = "0.2.3"
```
billti, line 77: Did you chat with Oscar on this? He's been doing some Paulimer revs recently (and we also had a chat re `rand` not that long ago)

swernli: Yup, Oscar said it would be fine to update paulimer. And the reason I ran the tests and benchmarks was to confirm the behavior of rand, so at least on my machine this all works.  The multiplat build (and ADO) should alert us if it affects behavior on other platforms.

billti (approved):

PR #3581, opened 2026-08-12
Support readout noise in simulators and stim compiler

joao-boechat:
## Overview

Adds end-to-end support for **readout noise** (classical bit-flip noise applied
to a measurement result _after_ it is sampled). A measured `0` can be reported
as `1` with probability `p_zero_as_one`, and a measured `1` as `0` with
probability `p_one_as_zero`. The feature spans the Stim compiler, the QIR/adaptive
runtime representation, and every simulator back end.

## Stim compiler

- Measurement instructions (`M`/`MZ`/`MX`/`MY`, `MR`/`MRZ`/`MRX`/`MRY`, and the
  pair measurements `MXX`/`MYY`/`MZZ`) now accept an optional parenthesized
  probability, e.g. `M(0.1) 0`, and emit a `__quantum__rt__readout_noise` call
  per produced result. Previously a parenthesized argument was rejected.
- Probability is validated (`0 <= p <= 1`, at most one argument); a probability
  of `0` emits no noise call.
- Introduced `ResultId` / `QubitId` / `StimQubitId` type aliases to make the
  qubit-vs-result index distinction explicit throughout the QIR writer.

## Runtime / QIR representation

- New runtime intrinsic:
  `__quantum__rt__readout_noise(double p_zero_as_one, double p_one_as_zero, ptr result)`.
- New adaptive-bytecode opcode (`OP_READOUT_NOISE`) and a new
  `QirInstruction::ReadoutNoise` / `QirInstructionId::ReadoutNoise`.
- The Python adaptive-profile pass and the aggregate-gates pass lower the
  intrinsic into the new instruction/opcode.

## Simulators

Readout noise is supported in all simulators, base profile and adaptive profile:

- CPU full-state simulator
- CPU stabilizer (Clifford) simulator
- GPU full-state simulator


source/simulators/src/lib.rs
```diff
@@ -125,4 +125,8 @@ pub trait Simulator {
     /// Dumps the current state of the simulator in some representation that can be compared
     /// for `PartialEq` up to a global phase. This is meant to be used for testing.
     fn state_dump(&self) -> &Self::StateDumpData;
+
+    /// Applies readout noise to the measurement result with the given `result_id`.
+    /// The probabilities of flipping a 0 to a 1 and a 1 to a 0 are given by `p_zero_as_one` and `p_one_as_zero`, respectively.
+    fn apply_readout_noise(&mut self, p_zero_as_one: f64, p_one_as_zero: f64, result_id: QubitID);
```
fedimser: result_id is not QubitID (it's not number of a qubit).
Introduce new alias ResultID (or use usize).

source/simulators/src/gpu_full_state_simulator/gpu_statevector_shaders.wgsl
```diff
@@ -2389,6 +2391,24 @@ fn prepare_op_base_impl(shot_idx: u32) {
...
+        let result_id = op.q1;
+        let p_zero_as_one = op.unitary[0].x;
+        let p_one_as_zero = op.unitary[0].y;
+        let result_idx = shot_idx * RESULT_COUNT + result_id;
+        let result = atomicLoad(&results[result_idx]);
+        let sample = shot.rand_measure;
+        if (result == 0u && sample < p_zero_as_one) {
+            atomicStore(&results[result_idx], 1u);
+        } else if (result == 1u && sample < p_one_as_zero) {
+            atomicStore(&results[result_idx], 0u);
```
billti, line 2404: Why are you using atomic loads and stores here and a couple of times above? I assume this is the only core/thread that will be updating this specific result?

source/simulators/src/gpu_full_state_simulator/gpu_statevector_shaders.wgsl
```diff
@@ -2883,6 +2903,27 @@ fn interpret_classical(@builtin(global_invocation_id) gid: vec3<u32>) {
...
+                let p_one_as_zero = resolve_f32(shot_idx, instr.aux1, flags, 4u);
+                let result_id = resolve_u32(shot_idx, instr.aux2, flags, 5u);
+                let result_idx = shot_idx * RESULT_COUNT + result_id;
+                let result = atomicLoad(&results[result_idx]);
+                let sample = next_rand_f32(shot_idx);
+
+                if ((result == 0u && sample < p_zero_as_one)) {
+                    atomicStore(&results[result_idx], 1u);
+                } else if ((result == 1u && sample < p_one_as_zero)) {
+                    atomicStore(&results[result_idx], 0u);
```
billti, line 2921: Same comments here. I'm not sure why you are using atomic operations, versus regular read/writes.

billti: Ah, I see. The buffer is marked atomic, so they need to be. Ignore me!

fedimser (approved):
Looks correct, but I have to call out few things:
1. You are adding new QIR intrinsic `__quantum__rt__readout_noise`. Can we just do that without seeking approval from QIR alliance?
2. result_id argument is not validated in simulator. If it's out of bounds, this will result in indexing outside of array which may result in segfault or some other undefined behavior. I understand that in Clifford/CPU simulators there is no way to validate arguments. Should we at least panic if it's out of bounds? For GPU simulator, there is a mechanism for raising errors (by setting exit code for shot) - should we use that? 
3. Probabilities are not validated as well.

source/simulators/src/lib.rs
joao-boechat: Answering the questions up there:

1. We can do that because this is just for simulation purposes. Since it's a noise instruction, we only support it for simulators, and ignore it if emitting QIR for an actual machine. So it should be ok. (feel free to correct me @billti )
2. In Rust, an out of bounds index would result in panic already, instead of a segfault. In any case, for the stim compiler, I apply these validations, and I suppose that is the case for other compilers as well. And since most of the time we are the ones generating QIR to the simulators, these errors shouldn't happen (unless there is a bug in QIR generation).
3. Same as above. We should expect the compilers to validate it, it's not the simulator's job necessarily.

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1210,39 +1250,43 @@ impl<'noise> Compiler<'noise> {
             }
 
             // Collapsing Gates
-            "M" | "MZ" => self.broadcast_measure(instruction, |s, q, invert| {
-                s.op_measure("m", q, invert);
-            }),
-            "MR" | "MRZ" => self.broadcast_measure(instruction, |s, q, invert| {
-                s.op_measure_reset("mresetz", q, invert);
+            "M" | "MZ" => {
+                self.broadcast_measure(instruction, |s, q, negated| s.op_measure("m", q, negated))
```
billti, line 1262: Why did you change a bunch of parameter naming and error messages as part of this PR that seem unrelated to readout noise? It makes trying to review for the readout noise logic quite difficult.

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1305,51 +1352,64 @@ impl<'noise> Compiler<'noise> {
...
     fn broadcast_measure(
         &mut self,
         instruction: &Instruction,
-        f: impl FnMut(&mut Self, u32, bool),
+        mut measure: impl FnMut(&mut Self, StimQubitId, bool) -> ResultId,
     ) {
-        self.unsupported_args(instruction);
-        self.for_each_negated_qubit(instruction, f);
+        let Some(readout_noise) = self.expect_readout_noise(instruction) else {
+            return;
```
billti, line 1414: I don't understand the logic here. So if there is no readout noise, it just returns silently without doing any measurement?

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1305,51 +1352,64 @@ impl<'noise> Compiler<'noise> {
...
+        };
+        self.for_each_negated_qubit(instruction, |s, q, negated| {
```
billti: So it only does measurement on negated qubits? I'm not following...

source/compiler/stim_compiler/src/qir.rs
```diff
@@ -1425,40 +1489,49 @@ impl<'noise> Compiler<'noise> {
...
     fn broadcast_pair_measure(
         &mut self,
         instruction: &Instruction,
-        f: impl FnMut(&mut Self, u32, u32, bool),
+        mut measure: impl FnMut(&mut Self, StimQubitId, StimQubitId, bool) -> ResultId,
     ) {
-        self.unsupported_args(instruction);
-        self.for_each_negated_pair(instruction, f);
+        let Some(readout_noise) = self.expect_readout_noise(instruction) else {
+            return;
```
billti, line 1551: Again, this seems like an error condition to me, not something you should silently ignore. Add comments to clarify if that isn't the case.

source/qdk_package/tests/test_simulators_gates_noisy.py
```diff
@@ -841,3 +841,54 @@ def test_noise_intrinsic_combined_with_gate_noise(sim_type):
...
+%Result = type opaque
+%Qubit = type opaque
+
+define i64 @ENTRYPOINT__main() #0 {
+    call void @__quantum__rt__initialize(i8* null)
+    call void @__quantum__qis__m__body(%Qubit* inttoptr (i64 0 to %Qubit*), %Result* inttoptr (i64 0 to %Result*))
+    call void @__quantum__qis__x__body(%Qubit* inttoptr (i64 1 to %Qubit*))
+    call void @__quantum__qis__m__body(%Qubit* inttoptr (i64 1 to %Qubit*), %Result* inttoptr (i64 1 to %Result*))
+    call void @__quantum__rt__readout_noise(double 1.0, double 0.0, %Result* inttoptr (i64 0 to %Result*))
+    call void @__quantum__rt__readout_noise(double 0.0, double 1.0, %Result* inttoptr (i64 1 to %Result*))
```
billti, line 837: It would be good to have some tests where the noise probability is something other than 0.0 or 1.0 to verify that statistically the probability and rng is actually being used correctly.

source/qdk_package/qdk/_native.pyi
```diff
@@ -892,6 +892,7 @@ class QirInstructionId(Enum):
     TupleRecordOutput: QirInstructionId
     ArrayRecordOutput: QirInstructionId
     CorrelatedNoise: QirInstructionId
+    ReadoutNoise: QirInstructionId
```
billti, line 914: Seems like maybe something you should be able to set on the NoiseConfig (i.e. NoiseTable for measurement). Thoughts @orpuente-MS ?

source/simulators/src/gpu_full_state_simulator/gpu_statevector_shaders.wgsl
```diff
@@ -2389,6 +2391,24 @@ fn prepare_op_base_impl(shot_idx: u32) {
         return;
     }
 
+    if (op.id == OPID_READOUT_NOISE) {
```
billti, line 2394: As you've added code for both the base profile and adaptive simulation here, you probably want some tests for both also. Looks like you only added tests in adaptive profile on the GPU currently.

joao-boechat: `test_simulators_gates_noisy.py` covers clifford, cpu full state, and gpu full state, all in base profile. I've already added tests there.

source/qdk_package/qdk/_native.pyi
joao-boechat: I had discussed about this with Oscar and we agreed to start with implementing it as an intrinsic, and then we would work on adding it to NoiseConfig in the future if we see there's a need.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: It doesn't return silently. If you go into "expect_readout_noise", you can see what happens:

```
fn expect_readout_noise(&mut self, instruction: &Instruction) -> Option<f64> {
        if instruction.args.len() > 1 {
            self.push_error(Error::TooManyArgs {
                instruction: instruction.name.clone(),
                expected: 1,
                span: instruction.span,
            });
            return None;
        }
        if instruction.args.is_empty() {
            return Some(0.0);
        }
        let arg = instruction.args[0];
        if !(0.0..=1.0).contains(&arg) {
            self.push_error(Error::InvalidReadoutNoiseProbability {
                instruction: instruction.name.clone(),
                probability: arg,
                span: instruction.span,
            });
            return None;
        }
        Some(arg)
    }
```

Essentially, if there is no readout_noise, it will treat it as having one with the value of 0. If there are any errors, just like the other "expect_..." functions, it will push into the compiler's state and return None. 

The reason I don't return the errors here is because there are cases where I want to accumulate as many errors as possible to have a more complete error message to the user, and if I just returned a Result<> at every step, the code would get very verbose and wouldn't achieve that.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: It will apply the closure for qubits that can be negated (but aren't necessarily). This covers instructions such as `MZ !0 1 2 !3`, where we can have both cases. 

The correct name for this function should've been "for_each_negatable_qubit", or "for_each_qubit_that_could_possibly_be_negated" but none of the options I could think of sounded any better. Do you have any suggestions?

source/compiler/stim_compiler/src/qir.rs
joao-boechat: As explained above, this isn't silently ignored.

source/compiler/stim_compiler/src/qir.rs
joao-boechat: Apologies, I found a bunch of small optimizations I could do in code readability that wouldn't warrant a separate PR, so I just pushed into this one. What would you recommend I do instead?

source/compiler/stim_compiler/src/qir.rs
billti: Not critical now, but maybe better naming or comments in the future. The way it reads is that it _expects_ to find readout noise and returns silently if it doesn't. Where possible, code should be intuitive by reading locally, you shouldn't have to read the called functions to make sense of it.

source/compiler/stim_compiler/src/qir.rs
billti: `negatable` would be a better name, else it reads like its only iterating over negated qubits.

source/qdk_package/qdk/_native.pyi
billti: Open an issue to track. This would be easy to 'forget'.

source/qdk_package/tests/test_simulators_gates_noisy.py
joao-boechat: Just added those!

source/qdk_package/qdk/_native.pyi
joao-boechat: Done!
https://github.com/microsoft/qdk/issues/3606

billti (approved):

PR #3585, opened 2026-08-13
Fix vendored file checks and Windows OpenQASM paths

### LINKED ISSUE #3583: CI failure: windows-latest (31651193979) (OpenQASM parser path failures on Windows)
See the corresponding [**workflow run**](https://github.com/microsoft/qdk/actions/runs/31651193979) for failure details, update the issue with any relevant information, and assign to the appropriate team member(s).

idavis:
- Added a dedicated PR vendor-sync CI job and re-enabled the `build.py`
- Synchronized both stale vendored source files and updated the renamed public re-export
- Corrected repair commands to include python3 and full script paths
- Normalized resolver paths to forward slashes, fixing Windows alias resolution, with regression coverage

### DIFF
--- a/source/qdk_openqasm/src/parser.rs
+++ b/source/qdk_openqasm/src/parser.rs
@@ -413,6 +413,12 @@ fn resolve_path(base: &Path, path: &Path) -> miette::Result<PathBuf> {
     Ok(normalized)
 }
 
+/// Converts a platform-native path into a logical source resolver path.
+/// Logical paths always use forward slashes so resolver keys are portable.
+fn logical_path(path: &Path) -> String {
+    path.to_string_lossy().replace('\\', "/")
+}
+
 /// Parse a QASM file and return the parse result using the provided resolver.
 /// Returns `Err` if the resolver cannot resolve the file.
 /// Returns `Ok` otherwise. Any parse errors will be included in the result.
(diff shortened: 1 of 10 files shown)


amcasey (approved):

source/qdk_openqasm/src/parser.rs
```diff
@@ -413,6 +413,12 @@ fn resolve_path(base: &Path, path: &Path) -> miette::Result<PathBuf> {
     Ok(normalized)
 }
 
+/// Converts a platform-native path into a logical source resolver path.
+/// Logical paths always use forward slashes so resolver keys are portable.
+fn logical_path(path: &Path) -> String {
+    path.to_string_lossy().replace('\\', "/")
```
swernli, line 419: definitely a corner case, but any concern this would cause bad behavior if someone is using unix paths with escaped spaces? As in, `/path/folder\ with\ spaces/file.txt`

swernli: that comes up exceedingly rarely, so not a blocker, just worth considering.

swernli (approved):

source/qdk_openqasm/src/parser.rs
idavis: It is just a key, not the real path.

swernli: Ah, makes sense, no concern then.

PR #3588, opened 2026-08-13
Add language samples for `parallel`, update notebooks

swernli:
This adds a language sample explaining the use of the new `parallel` expresison. It also updates some of the neutral atom notebooks that used manual parallelism to take advantage of the language feature for simpler code.


samples/language/Parallel.qs
```diff
@@ -0,0 +1,64 @@
...
+
+    // If instead of full parallelism only limited parallelism is desired,
+    // use the `parallel within <int> <expr>` expression instead. This allows for
+    // specifying an upper limit on the number of new allocations that are performed
+    // before falling back to qubit reuse. In this case, by specifying a limit of
+    // 2, the call to `JointMeasure` will use two distinct ancilla qubits for the
+    // first two iterations of the loop and then reuse the first ancilla during
+    // the third iteration.
+    parallel within 2 JointMeasure(qs);
+}
```
ScottCarda-MS, line 41: nit: Can we modify the example to avoid confusion on whether the `within 2` is promoting the "**two** distinct ancilla qubits" or the "first **two** iterations" of the loop?

ScottCarda-MS (approved):

samples/language/Parallel.qs
swernli: Ah, good point. That distinction, two ancilla qubits vs two loop iterations, is one that could use cleaning up. I'll reword the comment to clarify.

idavis (approved):

PR #3594, opened 2026-08-16
Add stabilizer branching for non-Clifford simulation

billti:
WIP: Prototype.

Support T gates and arbitrary Pauli rotations while preserving adaptive measurement, noise, and loss behavior. Add differential and integration coverage plus an OpenQASM notebook demonstrating simulation at 111 qubits.


source/simulators/src/stabilizer_simulator/branching_state.rs
```diff
@@ -0,0 +1,362 @@
...
+
+            let clifford = self.clifford_with_basis_state(basis);
+            return stabilizer_outcome_probability(&clifford, observable, outcome);
+        }
+
+        let mut projected = self.clone();
+        projected.project(observable, outcome)
+    }
+
+    fn project_single_stabilizer(&mut self, observable: &SparsePauli, outcome: bool) -> f64 {
```
fedimser, line 116: Add assertion that self.amplitudes.len()==1, to make it clear that this is correct only when state has single amplitude.

source/simulators/src/stabilizer_simulator/branching_state.rs
```diff
@@ -0,0 +1,362 @@
...
+/// stabilizer state, while keeping the Clifford frame common makes Clifford gates
+/// independent of the number of branches.
+#[must_use]
+#[derive(Clone)]
+pub struct BranchingState {
+    clifford: CliffordUnitary,
+    amplitudes: FxHashMap<BigUint, Complex64>,
+}
+
+impl BranchingState {
```
fedimser, line 29: It would be helpful to add more comments to functions on this class (AI can do that).

source/simulators/src/stabilizer_simulator/branching_state.rs
```diff
@@ -0,0 +1,362 @@
...
+            amplitudes: FxHashMap::from_iter([(BigUint::ZERO, Complex64::new(1.0, 0.0))]),
+        }
+    }
+
+    #[cfg(test)]
+    fn branch_count(&self) -> usize {
+        self.amplitudes.len()
+    }
+
+    pub fn unitary_op(&mut self, operation: UnitaryOp, support: &[usize]) {
```
fedimser, line 48: This doesn't apply arbitrary unitary, but only UnitaryOp, which are Clifford unitaries. The name is misleading, should it be apply_clifford_op?
I understand it's called like this to minimize diff. If you want to keep ti unitary_op, please state in a comment that it applies a Clifford op.

source/simulators/src/stabilizer_simulator/branching_state.rs
```diff
@@ -0,0 +1,362 @@
...
+            };
+            add_amplitude(
+                &mut result,
+                basis ^ &x_mask,
+                pauli_coefficient * pauli_phase * z_phase * amplitude,
+            );
+        }
+
+        result.retain(|_, amplitude| amplitude.norm_sqr() > AMPLITUDE_EPSILON.powi(2));
+        assert!(
```
fedimser, line 207: Instead of panic, can we return Err here? Thei is called from CliffordSim, and it already returns Result from its methods, so propagating Result will be a small change.

source/simulators/src/stabilizer_simulator.rs
```diff
@@ -31,8 +34,10 @@ pub struct StabilizerSimulator {
     noise_config: Arc<CumulativeNoiseConfig>,
     /// Random number generator used to sample from [`Self::noise_config`].
     rng: StdRng,
-    /// The current inverse state of the simulation.
-    state: OutcomeSpecificSimulation,
+    /// Random number generator used to sample measurement outcomes.
+    measurement_rng: StdRng,
```
fedimser, line 37: It seems that OutcomeSpecificSimulation is no longer needed and can be removed?

source/simulators/src/stabilizer_simulator/branching_state.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn arbitrary_rxx_correlates_measurement_outcomes() {
+        let angle = 0.417;
+        let mut state = BranchingState::new(2);
+        state.rotate(angle, &[paulimer::core::x(0), paulimer::core::x(1)].into());
+
+        let expected = (angle / 2.0).sin().powi(2);
+        assert!((state.project(&z(0), true) - expected).abs() < 1e-12);
+        assert!((state.project(&z(1), true) - 1.0).abs() < 1e-12);
+    }
+
```
fedimser, line 371: I would add more tests for rotation around Y, Z, YY, ZZ.

fedimser (approved):

source/simulators/src/stabilizer_simulator/branching_state.rs
billti: Wiring this through is not as simple as it looks. I'll have to leave this for now and add a follow up issue to address (and prioritize it if we see feedback). We'll doc it and treat it as a resource limit for now that will panic if exceeded (kind of like an OOM).

PR #3595, opened 2026-08-17
Add python widget for Bloch Sphere

ScottCarda-MS:
Adds python support for the Bloch Sphere visualizer by making it a python widget, which is then added under the qsharp_widgets package, and additionally made available under the QDK python package.

Example python usage:
```python
import qdk.widgets as w
bloch = w.BlochSphere('H T H') # string is initial gate sequence, can be empty
bloch
```

To minimize the impact of KaTeX on the Python package:

- KaTeX CSS and fonts are isolated in a Bloch-specific stylesheet, so existing widgets continue loading the original lightweight stylesheet.
- The production build retains only the four .woff2 font faces used by the Bloch sphere’s matrices and math expressions, reducing the Bloch stylesheet from approximately 383 KB to 109 KB raw (264 KB to 60 KB compressed).


billti (commented):
> To minimize the impact of KaTeX on the Python package:
>
> KaTeX CSS and fonts are isolated in a Bloch-specific stylesheet, so existing widgets continue loading the original lightweight stylesheet.
> The production build retains only the four .woff2 font faces used by the Bloch sphere’s matrices and math expressions, reducing the Bloch stylesheet from approximately 383 KB to 109 KB raw (264 KB to 60 KB compressed).

The index.js file for the widget however is 600kb larger after this PR (which is already too large as we still need to move the molecule visualizer out of it). This loads in every cell that just wants to show a histogram or a circuit or whatever.

As discussed, we should figure out how to make this a separate entry point so it only gets loaded when the user wants to see a Bloch sphere (which will prbably be pretty rarely). I can give you some guidance if needed.

source/widgets/js/bloch.css
```diff
@@ -0,0 +1,13 @@
...
+/*
+ * The Bloch sphere renders matrices that require KaTeX's fonts for correct
+ * spacing and scalable delimiters. Keep these fonts out of the shared widget
+ * stylesheet so widgets that predate the Bloch sphere retain their original
+ * CSS payload.
+ */
+@import "../../npm/qsharp/ux/qsharp-ux.css";
+@import "../../npm/qsharp/ux/bloch/bloch.css";
+@import "./widgets.css";
+@import "katex/dist/katex.min.css";
```
domorale, line 13: This import causes the build to embed KaTeX CSS and WOFF2 font data into `qsharp_widgets/static/bloch.css`, but the resulting wheel does not appear to include KaTeX's MIT license notice.

I verified both sides separately:

```bash
# Confirms that KaTeX assets are embedded in the wheel.
unzip -p target/wheels/qsharp_widgets-*.whl \
  qsharp_widgets/static/bloch.css |
  grep -o 'KaTeX_Main'

# Searches every file in the wheel for distinctive text from KaTeX's MIT notice.
zipgrep -i \
  'Khan Academy\|Permission is hereby granted' \
  target/wheels/qsharp_widgets-*.whl
```

The first command finds multiple KaTeX_Main declarations. The second finds no matches and exits with status 1.

Therefore, the KaTeX assets are included, but their copyright and permission notice is not. Since KaTeX's MIT license requires that notice to accompany copies or substantial portions, could we include the KaTeX license in the `qsharp-widgets` wheel and add a packaging check for it?

source/widgets/js/index.tsx
```diff
@@ -289,6 +293,25 @@ function renderCircuit({ model, el }: RenderArgs) {
...
+          model.set("gates", gates);
+          model.save_changes();
+        }}
+      ></BlochSphere>,
+      el,
+    );
+  };
+
+  onChange();
+  model.on("change:initial_gates", onChange);
```
domorale: This listener rerenders the wrapper when Python changes `initial_gates`, but the underlying `BlochSphere` reads `initialGates` only inside its mount-only `useEffect([])`. Since Preact preserves the mounted component when rendering it again, the updated prop is not applied.

I reproduced this with the wheel built from this PR:

```python
from IPython.display import display
from qsharp_widgets import BlochSphere

widget = BlochSphere("X")
display(widget)
```

The widget's gate-program text field initially shows `X`. Then run:

```python
widget.initial_gates = "H"

print("initial_gates:", widget.initial_gates)
print("gates:", widget.gates)
```

Observed:

```text
initial_gates: H
gates: X
```

The rendered gate-program field also remains `X`, rather than changing to `H`.

Could the core component respond to subsequent `initialGates` prop changes, or should `initial_gates` not be exposed as a synchronized mutable trait?


BlochSphere remains X after initial_gates changes to H

domorale (commented):
I built and tested the PR locally with the repository's Rust 1.95 toolchain. I left two inline comments:

- The generated widget wheel embeds KaTeX assets but appears not to include KaTeX's MIT license notice.
- Updating the synchronized `initial_gates` trait after display does not update the mounted Bloch Sphere; I included a minimal notebook reproducer and screenshot.

The widget otherwise built successfully and worked as expected in JupyterLab.

source/widgets/js/bloch.css
ScottCarda-MS: Hmm, I'm not really sure what the policy is here. We haven't been including copies of the licenses in the python packages we've shipped.

ScottCarda-MS: I've added the licenses to the package just in case. I'm not really sure it is needed, but we can remove them later if we find out they are unnecessary.

source/widgets/js/bloch.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import type { AnyModel } from "@anywidget/types";
+import mk from "@vscode/markdown-it-katex";
+import markdownIt from "markdown-it";
+import { render as prender } from "preact";
+
+import { BlochSphere } from "../../npm/qsharp/ux/bloch/bloch.js";
```
billti: You should never import from the relative path the source like this. Import from `qsharp-lang/ux` just like it did in index.tsx.

source/widgets/js/index.tsx
ScottCarda-MS: I'm not really sure what to do about this. The initial_gates are meant to be a memory of how the bloch sphere was initialized, so maybe it should be a read-only field. But not sure what the behavior of the `gates` field then. It updates as the user interacts with an adds/removes gates via the GUI, but I don't think the GUI should update if the user edits the `gates` field. So should they both be read-only?

I also don't know what the desired behavior would be if we have multiple instances of a bloch object rendered.

domorale: Me neither :). Probably something to discuss with the team?

source/widgets/js/bloch.tsx
ScottCarda-MS: fixed

source/widgets/js/index.tsx
ScottCarda-MS: To simplify things, I've dropped the `gates` field so that there isn't a dynamic state for the GUI to try to manage that can escape the GUI. That way multiple instances of the GUI don't fight over what the "correct" gate sequence is.
The `initial_gates` field is left, but is made read-only as it doesn't makes sense to edit something that is meant to tell you about how the object was initialized.

source/widgets/src/qsharp_widgets/__init__.py
```diff
@@ -213,6 +212,35 @@ def __init__(self, circuit):
...
 
+class BlochSphere(anywidget.AnyWidget):
+    _esm = pathlib.Path(__file__).parent / "static" / "bloch.js"
+    _css = pathlib.Path(__file__).parent / "static" / "bloch.css"
+
+    comp = traitlets.Unicode("BlochSphere").tag(sync=True)
+    _initial_gates = traitlets.Unicode("").tag(sync=True)
+
+    @property
+    def initial_gates(self):
```
billti: Why is this property here?

ScottCarda-MS: This makes the `initial_gates` read-only by letting me define a getter without a setter. This helps avoid the object having a state that can escape the GUI, reducing the complexity and avoiding situations where the GUI and the python object are out of sync.

source/widgets/js/index.tsx
billti: What do you mean by "multiple instances of the GUI" fighting over state? Does that related to your comment about "if we have multiple instances of a bloch object rendered" above? What are the concerns here?

ScottCarda-MS: This is the case in a notebook where a user has multiple cells that are used to render the GUI for a single bloch object. The user will then have access to multiple GUI's that can each independently be used to edit the gate sequence of the object. If the state of a GUI is just internal, then these multiple GUIs can act like independent experiences, but if there is a shared state that is accessible outside of them, like a `gate` field that is meant to update when users interact a GUI, then the experiences get tangled to gather in an awkward way. And gets even more confusing if the user is allowed to edit the `gates` field in a way to try to tell the GUIs what they should have for state. This is all avoided by simplifying things and not having this state external to the GUIs, i.e. not having `gates`.

source/widgets/src/qsharp_widgets/__init__.py
billti: Would anyone want to read it (especially if it should only ever have the value they used to create it)? Just leave it as an _ prefixed property to indicate it is private and isn't intended to be accessed.

billti (approved):

source/widgets/src/qsharp_widgets/__init__.py
ScottCarda-MS: No problem. It makes it even simpler to remove the public-facing read-only property and just leave the private backing field.

PR #3599, opened 2026-08-18
Enable user friendly angles in circuits

idavis:
This adds friendly angle rendering to circuits in both the circuit editor and the generated circuits view.

For the VS Code changelog on the next release:

```markdown
### Symbolic angles in circuit diagrams

Circuit diagrams now label a rotation with the angle you recognize instead of a decimal. A gate written as `rz(pi / 4) q;` in OpenQASM, or `Rz(Std.Math.PI() / 4.0, q)` in Q#, is drawn as `Rz(π / 4)` rather than `Rz(0.7854)`.

```qasm
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
rz(pi / 4) q[0];
rx(2 * pi / 3) q[1];
```

```text
q_0    ──── Rz(π / 4) ────
q_1    ── Rx(2 * π / 3) ──
```

This applies everywhere a circuit is shown: the circuit view, the text diagram, Python output, and the circuit JSON. An angle that is not a whole multiple or a simple fraction of π keeps the four-decimal form it has always had, so `rz(0.5) q;` is still drawn as `Rz(0.5000)`. Angles are recognized from their computed value, so a rotation built at run time is labeled too. Note that OpenQASM reduces an angle modulo 2π before the circuit is drawn, so `rz(-pi/4) q;` appears as `Rz(7 * π / 4)`.

If you read circuit JSON programmatically, note that a gate's `args` entry is an angle expression rather than always a decimal number, so code that converts it straight to a float needs to handle forms such as `π / 4`.
```

### DIFF
--- a/source/compiler/qsc_circuit/src/angle_format.rs
+++ b/source/compiler/qsc_circuit/src/angle_format.rs
@@ -49,7 +49,7 @@ pub(crate) fn format_angle(value: f64) -> String {
     if multiple >= 1.0 - RATIO_EPS && (multiple - multiple.round()).abs() < RATIO_EPS {
         let count = multiple.round();
         // A single turn reads better unqualified than as "1 * π".
-        return if count == 1.0 {
+        return if (count - 1.0).abs() < RATIO_EPS {
             format!("{sign}π")
         } else {
             format!("{sign}{count:.0} * π")
--- a/source/compiler/qsc_circuit/src/angle_format.rs
+++ b/source/compiler/qsc_circuit/src/angle_format.rs
@@ -1,0 +1,113 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#[cfg(test)]
+mod tests;
+
+use std::f64::consts::PI;
+
+/// Tolerance in radians for matching an angle against a candidate fraction of π.
+const RADIAN_EPS: f64 = 1e-9;
+
+/// Tolerance on the dimensionless ratio `angle / π` used by the whole-multiple rule.
+/// Numerically equal to `RADIAN_EPS`, but it bounds a ratio rather than an angle, so
+/// that rule's effective angular tolerance is this value times π.
+const RATIO_EPS: f64 = 1e-9;
+
+/// Largest numerator and denominator considered for a reduced fraction of π.
+const MAX_FRAC: u32 = 64;
+
+/// Largest denominator considered for a simple `π / d` fraction.
+const MAX_PI_FRAC: u32 = 128;
+
+/// Formats a gate angle for display, preferring a symbolic multiple or fraction
+/// of π and otherwise falling back to four decimal places.
+///
+/// Emitted symbolic forms always use explicit operators (`2 * π / 3`, never
+/// `2π/3`) so they remain valid input to the circuit editor's angle expression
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: source/compiler/qsc_circuit/src/angle_format.rs [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+    let sign = if value < 0.0 { "-" } else { "" };
+
+    // Whole multiples of π. The first test rejects a magnitude below π: its ratio
+    // rounds to zero, which would pass the second test and print "0 * π". The
+    // second test is the actual rule, that the ratio is an integer within tolerance.
+    let multiple = magnitude / PI;
+    if multiple >= 1.0 - RATIO_EPS && (multiple - multiple.round()).abs() < RATIO_EPS {
+        let count = multiple.round();
+        // A single turn reads better unqualified than as "1 * π".
+        return if (count - 1.0).abs() < RATIO_EPS {
```
ScottCarda-MS: `count` should be a positive integer, so you should just be able to check if `count == 1` here.

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
```diff
@@ -158,6 +159,57 @@ test("isValidAngleExpression: adjacent factors without an operator are invalid",
...
+    ["-π", -Math.PI],
+    ["3 * π", 3 * Math.PI],
+    ["π / 4", Math.PI / 4],
+    ["-π / 4", -Math.PI / 4],
+    ["2 * π / 3", (2 * Math.PI) / 3],
+  ];
+  for (const [expr, expected] of cases) {
+    const actual = evaluateAngleExpression(expr);
+    assert.ok(
+      actual !== undefined && Math.abs(actual - expected) < 1e-12,
```
ScottCarda-MS: Does the tolerance here depend on the `RATIO_EPS` or `RADIAN_EPS`?

source/compiler/qsc_circuit/src/angle_format/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    assert_eq!(format_angle(0.5), "0.5000");
+    assert_eq!(format_angle(1.0), "1.0000");
+    assert_eq!(format_angle(1.2345), "1.2345");
+    assert_eq!(format_angle(-0.3), "-0.3000");
+}
+
+#[test]
+fn tolerance_boundary() {
+    // Inside the tolerance, still recognized.
+    assert_eq!(format_angle(PI / 4.0 + 1e-11), "π / 4");
```
ScottCarda-MS: I guess it depends on if those epsilon constants ever change, but maybe these tolerances should reference those constants.

ScottCarda-MS (approved):
Just a few comments

source/compiler/qsc_circuit/src/angle_format.rs
idavis: Yeah, that should be safe for 1.0

source/npm/qsharp/test/circuit-editor/angleExpression.test.mjs
idavis: Neither. Refactored and documented.

PR #3601, opened 2026-08-18
Fix learning notebook tree navigation

amcasey:
It was only working for type `exercise`, but I introduced type `code-cell` at the last minute and it should work for those too.


HABER7789 (approved):
Looks good to me. This matches the navigation behavior we discussed.

PR #3602, opened 2026-08-19
Add `qdk.ec` module

orpuente-MS:
This PR added `qdk.ec` for authoring, profiling, and auditing quantum error-correction schemes described by [qodecs](https://github.com/microsoft/qdk-ec/tree/main/qodec).

The `qodec` package owns the declarative data model and persistence, including loading and saving artifacts. `qdk.ec` provides construction, algebraic analysis, exact-simulation-based profiling, fault analysis, and consistency checks through a small, flat public API.

## Public API

| API | Purpose |
| --- | --- |
| `qdk.ec.build_qodec` | Build a two-layer qodec from a supported stabilizer code, using flagged or bare CSS construction strategies. |
| `qdk.ec.derive` | Discover checks and readout bindings for a gadget or complete qodec, returning a new artifact without modifying the input. |
| `qdk.ec.audit` | Check a qodec’s structural and semantic consistency and return a structured report. |
| `qdk.ec.CodeProfile` | Analyze a code’s stabilizer, gauge, and logical operators; compute syndromes, logical effects, equivalence, encoding Cliffords, and distance. |
| `qdk.ec.GadgetProfile` | Profile a gadget or bare circuit: realized action, declared objective where applicable, checks, readouts, fault effects, equivalence, and fault distance. |
| `qdk.ec.ChannelAction` | Represent and compare logical channel actions without exposing their internal representation. |
| `qdk.ec.Pauli` | Represent Pauli operators used in code and fault analysis. |
| `qdk.ec.FaultEvent` | Describe and compose circuit faults, including Pauli errors and recorded-readout bit flips. |
| `qdk.ec.FaultEffect` | Describe a fault’s effect on check syndromes, logical readouts, and output errors. |
| `qdk.ec.Distance` | Return certified distance bounds with witness factors establishing a finite upper bound. |
| `qdk.ec.Diagnostic` | Represent an audit finding with severity, artifact context, details, and source location when available. Severity is exposed as `Diagnostic.Severity`. |
| `qdk.ec.Report` | Collect audit diagnostics, expose overall validity, and group findings by severity, rule, or artifact. |

Install the optional dependencies with:

```sh
pip install "qdk[ec]"
```

The optional HiGHS distance backend is available through `qdk[ec,ec-highs]`.


source/qdk_package/qdk/ec/README.md
```diff
@@ -0,0 +1,154 @@
+---
+description: Develop and test quantum error correction schemes with qdk.ec
+---
```
apaetz: I notice that, for example, qre does not have corresponding readme like this.  Is that on purpose? Do the other qdk modules have readmes?

source/qdk_package/qdk/ec/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    "checks",
+    "code",
+    "complete_gadget",
+    "complete_qodec",
+    "distance",
+    "equivalence",
+    "faults",
+    "from_yaml",
+    "lint",
+    "load_yaml",
```
apaetz: I suggest we push the `yaml` part into an optional paramter.  The type yaml, json, etc could by default be inferred from the filename or file contents.  Same also for `save_yaml`.  Also, should the name be `load_qodec`? That is should it say `qodec` directly? For now `Qodec` is the only thing to load or save. But maybe there will be other artifacts in the future?

source/qdk_package/qdk/ec/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Develop
+
+`qdk.ec` moves qodecs between disk, memory, and YAML text, and finishes drafts
+that a human should not have to finish by hand.
+
+```python
+import qdk.ec as ec
+
+qodec = ec.load_yaml("protocol.qodec.yaml")
```
apaetz: Something tricky here is that we also plan on having a package called `qodec`.  So a variable with the same name will shadow it.  Better to use the actual name of the qodec, in this case `protocol`
```suggestion
protocol = ec.load_yaml("protocol.qodec.yaml")
```

source/qdk_package/qdk/ec/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    "memory_program",
+    "qodec_from_code",
+    "readouts",
+    "save_yaml",
+    "synthesis_notes",
+    "to_yaml",
```
apaetz: This is a mix of functions and modules. No data structures.  I suggest either flattening out more, or offering a very small number of principled sub-modules (or both).

source/qdk_package/qdk/ec/action.py
```diff
@@ -0,0 +1,37 @@
...
+the codes on its boundaries. Predicates comparing two already-computed actions
+live in :mod:`qdk.ec.equivalence`.
+"""
+
+from ._analysis.circuit_action import (
+    CircuitAction,
+    action_of,
+    declared_action_of,
+    gadget_action_mismatch,
+    input_qubits_of,
```
apaetz: Not clear what this has to do with action. Also, does it make sense? Do the input qubits need to be integers?

source/qdk_package/qdk/ec/action.py
```diff
@@ -0,0 +1,37 @@
...
+
+:func:`action_of` computes the action of any program, optionally with respect to
+the codes on its boundaries. Predicates comparing two already-computed actions
+live in :mod:`qdk.ec.equivalence`.
+"""
+
+from ._analysis.circuit_action import (
+    CircuitAction,
+    action_of,
+    declared_action_of,
```
apaetz: Is this one needed? Really we want the `CircuitAction` from `Gadget.action`, right?

source/qdk_package/qdk/ec/action.py
```diff
@@ -0,0 +1,37 @@
...
+    gadget_action_mismatch,
```
apaetz: I'm not sure this function carries enough weight.  From the call site it's just as clear and easy to say something like `gadget.action = action_of(gadget.circuit)`

source/qdk_package/qdk/ec/checks.py
```diff
@@ -0,0 +1,109 @@
...
+    if not rows:
+        return OutcomeCode(BitMatrix.zeros(0, measurement_count))
+    return OutcomeCode(BitMatrix(rows))
+
+
+__all__ = [
+    "OutcomeCode",
+    "Profile",
+    "checks_of",
+    "essential_checks_of",
```
apaetz: We might consider making this private, at least at first.  It keeps the API surface smaller.

source/qdk_package/qdk/ec/_analysis/propagation/pauli.py
```diff
@@ -0,0 +1,139 @@
...
+    Literal,
+    Mapping,
+    cast,
+    get_args,
+)
+
+from more_itertools import nth_combination, nth_product
+from paulimer import SparsePauli
+
+Pauli = SparsePauli
```
apaetz: `Pauli` is leaked into some public functions, but isn't publicly exported itself. It probably should be.

source/qdk_package/qdk/ec/_analysis/check_discovery.py
```diff
@@ -0,0 +1,356 @@
...
+        tuple(input_outcomes),
+        program_result.observe_outcomes,
+        tuple(output_outcomes),
+        declared_outcomes,
+        input_refs,
+        output_refs,
+    )
+
+
+def checks_of(gadget: qc.Gadget) -> list[Equation]:
```
apaetz: `Equation` isn't exported publicly, but `checks_of` is.  Should we return `Equation` here, or somethign else?

apaetz: There are also other public checks functions like `OutcomeProfile.checks` that have a different data structure.

source/qdk_package/qdk/ec/_synthesis.py
```diff
@@ -0,0 +1,726 @@
...
+            "verification",
+            type(error).__name__,
+            str(error),
+        )
+    if mismatch is not None:
+        return _SynthesisFailure("verification", "ActionMismatch", mismatch)
+    return gadget
+
+
+def memory_program(qodec: qc.Qodec, *, rounds: int = 1) -> "Program":
```
apaetz: These mnemonics aren't gauranteed to be included in any particular ISA.  Is this kind of function well defined for all qodecs?

source/qdk_package/qdk/ec/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    "equivalence",
+    "faults",
+    "from_yaml",
+    "lint",
+    "load_yaml",
+    "memory_program",
+    "qodec_from_code",
+    "readouts",
+    "save_yaml",
+    "synthesis_notes",
```
apaetz: This function seems out a place for a top-level export.  I'm not clear on what it does.

source/qdk_package/qdk/ec/code.py
```diff
@@ -0,0 +1,75 @@
...
+def encoding_clifford_of(
+    code: qc.Code,
+    *,
+    supported_by: Sequence[int] | None = None,
+) -> CliffordUnitary:
+    """Return a Clifford encoder for ``code``."""
+    return _encoding_clifford_of(_view(code), supported_by=supported_by)
+
+
+__all__ = [
```
apaetz: These functions could instead be exported as methods on a richer code type, i.e., `SubystemCode`.

source/qdk_package/qdk/ec/equivalence.py
```diff
@@ -0,0 +1,30 @@
...
+from ._analysis.circuit_action import (
+    are_equivalent_mod_paulis as actions_equivalent_mod_pauli,
+    are_outcome_equivalent as actions_outcome_equivalent,
+)
+from ._analysis.equivalence import gadgets_equivalent, why_not_equivalent
+from .code import codes_equivalent
+
+__all__ = [
+    "actions_equivalent_mod_pauli",
+    "actions_outcome_equivalent",
```
apaetz: Let's consider consolidating into one (public) function, with a `modulo_pauli=False` parameter.

source/qdk_package/qdk/ec/equivalence.py
```diff
@@ -0,0 +1,30 @@
...
+    "codes_equivalent",
+    "gadgets_equivalent",
```
apaetz: Let's consider consolidating these all into one `are_equivalent` function, with union-type parameters.

source/qdk_package/qdk/ec/readouts.py
```diff
@@ -0,0 +1,56 @@
...
+        for index, outcomes in enumerate(observables_as_xor_map(gadget).values())
+    )
+    return OutcomeProfile(checks=checks, observables=observables)
+
+
+__all__ = [
+    "OutcomeProfile",
+    "Profile",
+    "outcome_profile_of",
+    "outcomes_flipped_by_anti_observables_of",
```
apaetz: Consider removing from the API surface.

source/qdk_package/qdk/ec/lint/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+__all__ = [
+    "Auditor",
+    "Diagnostic",
+    "Phase",
+    "Report",
+    "Rule",
+    "Severity",
+    "diagnose",
+    "why_not_valid",
+]
```
apaetz: I'd prefer to keep these out of the public API initially. Eventually we could expose them to allow for extensions.

source/qdk_package/qdk/ec/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+__all__ = [
+    "action",
+    "checks",
+    "code",
+    "complete_gadget",
+    "complete_qodec",
+    "distance",
+    "equivalence",
+    "faults",
+    "from_yaml",
```
apaetz: I think we can drop the yaml functions. The qodec package already covers this.

qdk-ec-review/audit-diagnostics.html
```diff
@@ -0,0 +1,2175 @@
+<!doctype html>
```
billti: Why is this file checked in?

orpuente-MS: Removed

joao-boechat (approved):

PR #3604, opened 2026-08-19
Billt/vsce locked

billti:
Move from a global install of VSCE (which isn't locked or repeatable) to having it as a dependency and installing via the lock file.


.ado/publish.yml
```diff
@@ -527,7 +522,7 @@ extends:
                   workingFile: .npmrc
 
               - script: |
-                  npm install -g @vscode/vsce --userconfig=.npmrc
+                  npm ci --userconfig=.npmrc
```
amcasey, line 530: Seems like this will install a lot of packages that weren't previously installed.  We could create separate package and lock files, but that would probably be overkill.

amcasey (approved):

.ado/publish.yml
billti: I'm not seeing any impact on build times, and this would only add time (if any) to the publishing step, which we only do a couple times a month, so I'm not overly worried.

PR #3607, opened 2026-08-19
Attempt to specify a venv for jupyter

amcasey:
Use some undocumented APIs to detect changes to, read, and set notebook python environments.

On open, look for a per-course value persisted in qdk-learning.json. If there isn't one, look for a virtual environment in the learning workspace root (typically, the workspace root).

When the user sets an environment, persist it per-course in qdk-learning.json.

The code assumes the undocumented APIs might change or disappear. If we lose reading or events, we'll still try to specify the heuristically chosen environment. If we lose writing, then reading is kind of moot.
The ultimate fallback is the behavior from before this PR: let jupyter and/or the user choose an environment for each notebook.

### DIFF
--- a/source/vscode/src/learning/service.ts
+++ b/source/vscode/src/learning/service.ts
@@ -5,6 +5,7 @@ import { log } from "qsharp-lang";
 import { getExerciseSources } from "qsharp-lang/katas-md";
 import * as vscode from "vscode";
 import { FullProgramConfig, getProgramForDocument } from "../programConfig.js";
+import { getVenvInFolder } from "../pythonEnvs.js";
 import { ProgramRunStatus, runProgram } from "../run.js";
 import { EventType, sendTelemetryEvent } from "../telemetry.js";
 import { createCourseProvider, toDescriptor } from "./courseProvider.js";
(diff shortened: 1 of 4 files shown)


source/vscode/src/learning/service.ts
```diff
@@ -163,6 +166,22 @@ export class LearningService {
...
+    courseId: string,
+  ): Promise<{ id: string; path: string } | undefined> {
+    const ws = this.requireWorkspace();
+    const saved = ws.progressData.pythonEnvironments[courseId];
+    if (saved) {
+      if (await uriExists(vscode.Uri.file(saved.path))) {
+        return saved;
+      }
+      log.info(
+        `Persisted venv at ${saved.path} no longer exists; falling back to discovery.`,
```
billti, line 179: If I'm reading this correctly, a potential issue may be if a user is using a venv that is NOT the global venv, but is also not in the workspace. It's quite common folks folks to create venvs under their home directory and use those across projects (e.g. `~/.venv`). As that is not under the workspace, and uriExists uses workspace APIs to try and find it, would this fail in that scenario? (i.e. not honor the saved kernel)

amcasey: I wasn't aware VS Code's file APIs restricted themselves to workspace files.  If that's the case, then yes, it should probably do a raw stat, assuming such an API is available.

amcasey: I chose a virtual environment outside (and not above) the workspace and it was successfully persisted to qdk-learning.json and reused when I closed and reopened the workspace.

billti (approved):

PR #3611, opened 2026-08-19
Add pyscf

amcasey:
...as an option when creating a venv and to the chemistry course requirements.txt.

### DIFF
--- a/source/vscode/resources/qdk-learning/courses/chemistry-qpe/requirements.txt
+++ b/source/vscode/resources/qdk-learning/courses/chemistry-qpe/requirements.txt
@@ -1,2 +1,3 @@
 qdk-chemistry[jupyter]>=2.1.0
 ipykernel>=6.0
+pyscf>=2.14.0
--- a/source/vscode/src/pythonEnvs.ts
+++ b/source/vscode/src/pythonEnvs.ts
@@ -73,8 +73,10 @@ const packagePickItems: vscode.QuickPickItem[] = [
 ];
 
 // pyscf doesn't support Windows
+const isWindows =
+  typeof process !== "undefined" && process.platform === "win32";
 const availablePackagePickItems = packagePickItems.filter(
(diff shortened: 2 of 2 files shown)


source/vscode/resources/qdk-learning/courses/chemistry-qpe/requirements.txt
```diff
@@ -1,2 +1,3 @@
 qdk-chemistry[jupyter]>=2.1.0
 ipykernel>=6.0
+pyscf>=2.14.0
```
amcasey, line 3: This is latest.  There's probably an older one that would work.

source/vscode/src/pythonEnvs.ts
```diff
@@ -65,6 +72,11 @@ const packagePickItems: vscode.QuickPickItem[] = [
   },
 ];
 
+// pyscf doesn't support Windows
+const availablePackagePickItems = packagePickItems.filter(
+  (item) => item.label !== "pyscf" || process.platform !== "win32",
```
billti: Does this distinguish between Python for Windows and Python running in WSL? The latter should work fine.

billti: (i.e. if they are running in a WSL environment, we should still offer PySCF)

billti (approved):

source/vscode/src/pythonEnvs.ts
amcasey: That was the goal and what I saw when I tested on both native windows and WSL.

source/vscode/resources/qdk-learning/courses/chemistry-qpe/requirements.txt
amcasey: Sounds like it won't be in the course requirements for long enough for fine-tuning the min version to be worthwhile.

PR #3615, opened 2026-08-20
[QRE] Show errors in case when resource estimation has no results.

fedimser:
* Add mechanism to propagate resource estimation errors from Rust to Python (new `Vec<String>` in EstimationCollection).
* Add errors there for failed estimates.
* Add error for the case when all traces have error higher then max_error.
* In Python, when estimation produced zero results, print a message containing:
  * All errors.
  * Suggestion to increase max_error.

Example:
image

### DIFF
--- a/source/qdk_package/qdk/qre/_estimation.py
+++ b/source/qdk_package/qdk/qre/_estimation.py
@@ -211,6 +211,22 @@ def estimate(
         total_jobs = collection.total_jobs
         successful = collection.successful_estimates
 
+    if len(collection) == 0:
+        warning_message = "Resource estimation produced zero results.\n"
+        if any(
+            "resource estimation exceeded the maximum allowed error" in err
+            for err in collection.errors
+        ):
+            warning_message += (
+                "This was caused by error exceeding max_error for every trace. "
+                "Try increasing max_error to get results.\n"
+            )
+        if len(collection.errors) > 0:
+            warning_message += "Resource estimation produced the following errors:\n"
+            for err in collection.errors[:10]:
+                warning_message += "  " + err + "\n"
+        print(warning_message)
+
     # Post-process the results and add them to a results table
     table = EstimationTable()
 
--- a/source/qre/src/trace/estimation.rs
+++ b/source/qre/src/trace/estimation.rs
@@ -443,7 +443,7 @@ pub fn estimate_with_graph(
             let runtime_affecting_ids = &runtime_affecting_ids;
             let isa_index = Arc::clone(&isa_index);
             scope.spawn(move || {
-                let mut local_results: Vec<Result<EstimationResult, Error>> = Vec::new();
+                let mut local_results = Vec::new();
                 loop {
                     let job_idx = next_job.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
                     if job_idx >= total_jobs {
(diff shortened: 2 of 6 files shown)


source/qdk_package/qdk/qre/_estimation.py
```diff
@@ -211,6 +211,22 @@ def estimate(
...
 
+    if len(collection) == 0:
+        warning_message = "Resource estimation produced zero results.\n"
+        if any(
+            "resource estimation exceeded the maximum allowed error" in err
+            for err in collection.errors
+        ):
+            warning_message += (
+                "This was caused by error exceeding max_error for every trace. "
+                "Try increasing max_error to get results.\n"
```
msoeken, line 222: This is not the right way to fix the problem. The `max_error` is application specific and not something one can change. Increasing the error budget for an algorithm may render the result of running the quantum algorithm meaningless. Having computed multiple estimates successfully but they all exceeded the error budget (max_error) means that the parameter space for the ISA transforms has been too tight and need to be relaxed.

source/qre/src/trace/estimation.rs
```diff
@@ -423,7 +443,7 @@ pub fn estimate_with_graph(
             let runtime_affecting_ids = &runtime_affecting_ids;
             let isa_index = Arc::clone(&isa_index);
             scope.spawn(move || {
-                let mut local_results = Vec::new();
+                let mut local_results: Vec<Result<EstimationResult, Error>> = Vec::new();
```
swernli: The explicit type annotation here is not strictly needed, though it doesn't hurt.
```suggestion
                let mut local_results = Vec::new();
```

source/qre/src/trace/estimation.rs
```diff
@@ -458,8 +477,10 @@ pub fn estimate_with_graph(
 
                         result.set_trace_index(*trace_idx);
 
-                        local_results.push(result);
+                        local_results.push(Ok(result));
                         record_success(combination, &pruning_witnesses[*trace_idx]);
+                    } else {
+                        local_results.push(result);
                     }
```
swernli, line 484: nitpick: the result gets pushed into `local_results` regardless of the path in this if-statement, so it could be simplified a bit.

```suggestion
                    let result = traces[*trace_idx].estimate(&isa, Some(max_error), composition);
                    if let Ok(mut result) = result {
                        let isa_idx = isa_index
                            .write()
                            .expect("RwLock should not be poisoned")
                            .push(combination, &isa);
                        result.set_isa_index(isa_idx);

                        result.set_trace_index(*trace_idx);

                        record_success(combination, &pruning_witnesses[*trace_idx]);
                    }
                    local_results.push(result);
```

swernli (approved):
Changes look good, just a couple minor nitpicks!

source/qre/src/trace/estimation.rs
fedimser: This cannot be done, because of this error: use of partially moved value: `result`

source/qdk_package/qdk/qre/_estimation.py
```diff
@@ -211,6 +211,22 @@ def estimate(
...
+            )
+        if len(collection.errors) > 0:
+            warning_message += "Resource estimation produced the following errors:\n"
+            for err in collection.errors[:10]:
```
msoeken, line 226: This is not what I had in mind for an error analysis in the case of no results. At minimum, I suggest to do the following:

1. Check if there are some successful estimates but with a too high error rate; in that case suggest to review the parameters in the transforms of the ISA query to have a larger search space
2. Extract some useful message from the values in the stats, e.g., it could be that non of the traces was compatible with any of the ISAs. This means that would need additional trace transforms or additional ISA transforms.

PR #3616, opened 2026-08-20
Keep track of whether the user has chosen a course

amcasey:
Until then, leave the tree collapsed to give them all equal visual weight.  Have the agent simply list available courses, rather than kicking one off.


HABER7789 (approved):
I have tested locally with a new workspace. Courses remained unselected until I chose one, and the selection persisted after reload. Looks good!

billti (approved):

PR #3620, opened 2026-08-21
RCA Perf: Reduce `RuntimeFeatureFlags` size

swernli:
This change removes `QubitAllocation` and `MustBeInlined` feature flags in favor of tracking those via out-of-band hash sets. `QubitAllocation` was only ever used to determine callables whose entire call chain must be inlined, while `MustBeInlined` was about tracking call expressions that must be inlined. By tracking these separately, the `RuntimeFeatureFlags` enum can be reduced back to `u32` and significantly improve RCA perf.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/source/compiler/qsc_fir/src/fir.rs
+++ b/source/compiler/qsc_fir/src/fir.rs
@@ -283,7 +283,7 @@ impl From<(PackageId, BlockId)> for StoreBlockId {
 }
 
 /// A unique identifier for an expression within a package store.
-#[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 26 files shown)


swernli:
Going to close this for now. The complexity trade-off relative to the perf gain makes this not a high priority at the moment. Will return to this idea if/when we need to squeeze any other performance out of the RCA stage.

PR #3627, opened 2026-08-21
Make getCurrentActivityForCell not depend on current position

amcasey:
It's fundamentally passing in a URI and cellID, so resolve those instead of trying to compare them to the active position.

Bonus: tidy up the hint tool so it uses a single state, rather than using two and trying to detect conflicts.


source/vscode/src/learning/service.ts
```diff
@@ -582,7 +577,7 @@ export class LearningService {
     }
 
     const location: ActivityLocation = {
-      courseId: this.activeCourse.id,
+      courseId: course.id,
```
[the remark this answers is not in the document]
amcasey: Because of our editor event listener, I think it's basically impossible to have notebooks open for two courses at once, but I find the code cleaner after addressing this.

source/vscode/src/learning/service.ts
```diff
@@ -766,27 +770,24 @@ export class LearningService {
...
+  getHintContext(
+    location: ActivityLocation,
+    source?: TelemetrySource,
+  ): HintContext | undefined {
     if (source) {
       this.sendActivityActionTelemetry("hint", source);
     }
 
-    const exercise = this.resolveExercise();
+    const exercise = this.resolveExerciseAt(location);
```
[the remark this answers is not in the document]
amcasey: Fixed

source/vscode/src/learning/service.ts
```diff
@@ -582,17 +577,26 @@ export class LearningService {
...
       location,
       unitTitle: unit.title,
       activityTitle: activity?.title ?? cellId,
       content,
     };
+
+    const progress = this.getCourseProgress(course.id);
+
+    return {
+      course,
```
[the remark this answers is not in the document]
amcasey: Fixed

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -386,21 +371,20 @@ export class LearningTools {
...
+    // The notebook learning state will be undefined if we're not in a notebook
+    // or if we couldn't identify the current cell for some reason.
+    // In that case, we fall back to the current state.
+    const state = this.notebookLearningState() ?? this.service.getState();
+
     const progress = state.progress;
-    const cur = progress.currentPosition?.unitId;
-    const currentUnit = cur
-      ? progress.units.find((u) => u.id === cur)
+    const unitId = progress.currentPosition?.unitId;
```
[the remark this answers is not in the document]
amcasey: Fixed

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -195,15 +195,13 @@ export class LearningTools {
     await this.ensureInitialized();
     return this.invoke(async () => {
       const uri = this.getCurrentFileUri();
-      if (isNotebookCourse(this.service.getActiveCourseInfo())) {
+      const editor = vscode.window.activeNotebookEditor;
+      if (editor && editor.notebook.uri.toString() === uri.toString()) {
```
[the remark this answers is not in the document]
amcasey: Fixed

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -167,7 +167,7 @@ export class LearningTools {
     return this.invoke(async () => {
       await this.service.switchCourse(input.courseId, "chat");
       await this.showActivity();
-      return { state: this.serializeState() };
+      return { state: this.serializeState(false) }; // workspace state is correct after switch
```
amcasey, line 174: There's a race between opening the new editor and capturing the editor state, so we now explicitly favor the position-to-be, rather than the current position.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -195,17 +195,17 @@ export class LearningTools {
     await this.ensureInitialized();
     return this.invoke(async () => {
       const uri = this.getCurrentFileUri();
-      if (isNotebookCourse(this.service.getActiveCourseInfo())) {
+      const editor = vscode.window.activeNotebookEditor;
```
amcasey, line 202: Read the current code cell, regardless of the active course.  Shouldn't make a difference in practice.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -217,28 +217,15 @@ export class LearningTools {
   /**
    * Return all built-in hints for the current exercise.
    */
-  async hint(): Promise<{ result: HintContext | null }> {
+  async hint(): Promise<{ result: HintContext | undefined } & StateSnapshot> {
```
amcasey, line 224: No longer throws in the very narrow scenario where it used to.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -260,10 +247,11 @@ export class LearningTools {
    */
   async next(): Promise<{ moved: boolean } & StateSnapshot> {
     await this.ensureInitialized();
+    this.requireQSharpCourse();
```
amcasey: It wasn't happening, but be explicit.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -289,9 +278,13 @@ export class LearningTools {
   }): Promise<StateSnapshot> {
     await this.ensureInitialized();
     return this.invoke(async () => {
+      const courseId = input.courseId;
+      if (courseId && courseId !== this.service.getActiveCourseId()) {
```
amcasey, line 286: The absence of switchCourse was an unrelated bug.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -327,10 +322,11 @@ export class LearningTools {
    */
   async resetExercise(): Promise<StateSnapshot> {
     await this.ensureInitialized();
+    this.requireQSharpCourse();
```
amcasey: Obviously, resetUnit will behave differently, @HABER7789

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -340,9 +336,13 @@ export class LearningTools {
   async solution(): Promise<{ solutions: string[] } & StateSnapshot> {
     await this.ensureInitialized();
     return this.invoke(async () => {
-      const solutions = this.service.getAllSolutions("chat");
+      const state = this.serializeState(true); // use editor for solutions
```
amcasey, line 343: Now consistent with hint behavior.  Not sure why someone would be asking for a solution on an arbitrary code cell (whereas they might ask for a hint/help).

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -385,22 +394,23 @@ export class LearningTools {
    * Build a compact snapshot of position and progress to attach to
    * every tool response.
    */
-  private serializeState(): SerializedLearningState {
-    const state = this.service.getState();
+  private serializeState(considerEditor: boolean): SerializedLearningState {
```
amcasey, line 405: Now caller's choice.

source/vscode/src/gh-copilot/learningTools.ts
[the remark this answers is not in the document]
amcasey: Separate work is planned to make a reset tool for notebooks.

HABER7789: Agreed. Thanks for flagging.

source/vscode/src/gh-copilot/learningTools.ts
billti: I'm still not clear why this call is here. Can you add a comment to explain it.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -367,6 +367,15 @@ export class LearningTools {
     }
   }
 
+  private requireQSharpCourse(): void {
+    const { kind } = this.service.getActiveCourseInfo();
+    if (kind !== "qsharp") {
+      throw new CopilotToolError(
+        "This operation is only supported for Q# courses.",
```
HABER7789, line 382: The guard works, just one small thing when I say next in the chemistry course Copilot told me to "use the Next button in the Learning panel", which doesn't exist for our notebook courses. Could this say "you may navigate from the learning tree"?

Image

HABER7789 (approved):
I have tested in the dev host. The state now correctly follows the visible notebook tab across units. No issues found.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -367,6 +367,15 @@ export class LearningTools {
     }
   }
 
+  private requireQSharpCourse(): void {
```
billti: Ah, I see now why it's called in all the other places. Maybe "checkIsQsharpCourse" would be a better name.

billti (approved):

source/vscode/src/gh-copilot/learningTools.ts
amcasey: "require*" appears to be an existing pattern for throwing checks (e.g. requireWorkspace).  I don't feel strongly though.  I would also be fine with "assertIsQSharpCourse" or "throwIfNotQSharpCourse".

source/vscode/src/gh-copilot/learningTools.ts
amcasey: We'd have to play whack-a-mole with the instructions - we don't directly control that text

amcasey: Took a shot.  No promises.

source/vscode/src/gh-copilot/learningTools.ts
amcasey: Went with "throwIfNotQSharpCourse" and added a comment.

source/vscode/src/gh-copilot/learningTools.ts
amcasey: Renamed as discussed below

PR #3630, opened 2026-08-22
[Circuit renderer] Add attribute to hide wrapper operations from circuit diagrams

fedimser:
* Adds new @CircuitRenderingOptions attribute for Q# operations.
* Parse it as comma-separated key-value list. For now, support only 'hideBox' argument, ignore all others.
* Operations annotated with `@CircuitRenderingOptions("hideBox=true")` are displayed as initially expanded group without a box and group name (that is, they are inlined).
* This was requested in https://github.com/microsoft/qdk/issues/3471
* This requires lookup of callable from RIR in FIR store, for that I had to store callable ID in RIR (otherwise we would need to do inefficient linear search by offset and name).

### Example

Code:

```
operation Main() : Unit {
	use qubits = Qubit[3];

	PrepareRegister(qubits);
	HiddenEntanglingLayer(qubits);

	for qubit in qubits {
		Reset(qubit);
	}
}

operation PrepareRegister(qubits : Qubit[]) : Unit {
	H(qubits[0]);
	X(qubits[2]);
}

@CircuitRenderingOptions("hideBox=true")
operation HiddenEntanglingLayer(qubits : Qubit[]) : Unit {
	EntanglePair(qubits[0], qubits[1]);
	EntanglePair(qubits[1], qubits[2]);
	Rz(0.25, qubits[2]);
}

operation EntanglePair(control : Qubit, target : Qubit) : Unit {
	CNOT(control, target);
}
```

With `hideBox=false` (or without annotation), having box manually expanded:
image

With `hideBox=true`:
image

### DIFF
--- a/source/compiler/qsc/src/interpret/circuit_tests.rs
+++ b/source/compiler/qsc/src/interpret/circuit_tests.rs
@@ -467,6 +467,137 @@ fn grouping_nested_callables() {
 }
 
 #[test]
+fn invisible_callable_is_flattened_in_circuit_json() {
+    let circuit = circuit_with_options_success(
+        r"
+            namespace Test {
+                @EntryPoint()
+                operation Main() : Unit {
+                    use q = Qubit();
+                    Visible(q);
+                    Invisible(q);
+                }
+
+                operation Visible(q : Qubit) : Unit {
+                    H(q);
+                }
+
+                @InvisibleInCircuit()
+                operation Invisible(q : Qubit) : Unit {
+                    X(q);
+                    Y(q);
+                }
+            }
+        ",
+        Profile::AdaptiveRIF,
+        CircuitEntryPoint::EntryPoint,
+        CircuitGenerationMethod::Static,
+        TracerConfig {
+            source_locations: false,
+            group_by_scope: true,
+            ..default_test_tracer_config()
+        },
+    );
+    let json = serde_json::to_string_pretty(&circuit).expect("circuit should serialize");
+
+    expect![[r#"
+        {
+          "qubits": [
+            {
+              "id": 0,
+              "numResults": 0
+            }
+          ],
+          "componentGrid": [
+            {
+              "components": [
+                {
+                  "kind": "unitary",
+                  "gate": "Main",
+                  "children": [
+                    {
+                      "components": [
+                        {
+                          "kind": "unitary",
+                          "gate": "Visible",
+                          "children": [
+                            {
+                              "components": [
+                                {
+                                  "kind": "unitary",
+                                  "gate": "H",
+                                  "targets": [
+                                    {
+                                      "qubit": 0
+                                    }
+                                  ]
+                                }
+                              ]
+                            }
+                          ],
+                          "targets": [
+                            {
+                              "qubit": 0
+                            }
+                          ],
+                          "metadata": {
+                            "scopeLocation": {
+                              "file": "test.qs",
+                              "line": 9,
+                              "column": 16
+                            }
+                          }
+                        }
+                      ]
+                    },
+                    {
+                      "components": [
+                        {
+                          "kind": "unitary",
+                          "gate": "X",
+                          "targets": [
+                            {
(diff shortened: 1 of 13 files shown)


source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -466,6 +466,56 @@ fn grouping_nested_callables() {
...
+
+                @InvisibleInCircuit()
+                operation Invisible(q : Qubit) : Unit {
+                    X(q);
+                    Y(q);
+                }
+            }
+        ",
+        Profile::AdaptiveRIF,
+        CircuitEntryPoint::EntryPoint,
```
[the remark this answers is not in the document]
fedimser: Ran the same test for all 3 circuit generation methods.

source/compiler/qsc/src/interpret/circuit_tests.rs
```diff
@@ -466,6 +466,65 @@ fn grouping_nested_callables() {
...
+                    Visible(q);
+                    Invisible(q);
+                    Reset(q);
+                }
+
+                operation Visible(q : Qubit) : Unit {
+                    H(q);
+                }
+
+                @CircuitRenderingOptions("unknownOption=value, hideBox=true")
```
ScottCarda-MS, line 485: What is "unknownOption"?

source/compiler/qsc_rir/src/debug.rs
```diff
@@ -51,12 +51,22 @@ impl Display for DbgLocation {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct DbgCallableId {
```
ScottCarda-MS, line 55: How does the debug-callable id play into the attribute or the circuit generation?

source/compiler/qsc_frontend/src/lower.rs
```diff
@@ -382,6 +382,63 @@ impl With<'_> {
...
+                    };
+                    let (key, value) = (key.trim(), value.trim());
+                    if key.is_empty() || value.is_empty() {
+                        self.lowerer.errors.push(Error::InvalidAttrArgs(
+                            "non-empty keys and values".to_string(),
+                            attr.arg.span,
+                        ));
+                        return None;
+                    }
+                    if key == "hideBox" {
```
billti: Could we also support the casing `HideBox`? CamelCase is pretty common in Q#, and with no IntelliSense in the editor it's annoying to have to remember the exact casing.

source/compiler/qsc_rir/src/debug.rs
fedimser: This is in PR description.
We generate circuit from RIR, and there we need to resolve Scope to CallableDecl (see builder.rs:873), so we can check what attributes the callable has. When Scope contains CallableId::Id, we can look up callable by id. When Scope contains CallableId::Source, it's just name and offset - and looking it up would require iterating over all callables in package (so quadratic complexity for building circuit). 
But turns out we can just store callable id in RIR, so then we can later look it up in fir::PackageStore.

source/compiler/qsc/src/interpret/circuit_tests.rs
fedimser: This is to demonstrate the case when there are multiple key-value pairs in rendering options. The intended behavior is to allow unrecognized keys but ignore them.

source/compiler/qsc_rir/src/debug.rs
ScottCarda-MS: I think I understand now. We need the callable id to be passed all the way down to the circuit builder to do a lookup for the original callable, then to check the attributes on that callable for the new attribute.

source/compiler/qsc_frontend/src/lower.rs
fedimser: Agreed, made key case insensitive

ScottCarda-MS (approved):

PR #3635, opened 2026-08-24
Split Learning reset into exercise and unit

HABER7789:
`resetExercise` reset one `.qs` file for Q# courses but wiped the entire notebook for notebook courses, while telling Copilot it "only resets an exercise." It now restores just the current activity one file, or one cell.

Adds `resetUnit` and a `qdk-learning-reset-unit` tool so chat can reset a whole unit, including by name.

Reset was picking the wrong notebook cell, because the stored position only moves when you run a cell, not when you click one. Now that #3627 has merged, reset resolves the selected cell by location the same way the hint and solution tools do. I also hardened the reset paths so a cancelled notebook close, rejected edit, or failed save aborts and keeps the user's progress instead of reporting success.


source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -332,12 +332,90 @@ export class LearningTools {
...
+   * approve a reset that will fail.
+   *
+   * **Must be free of side-effects** — only reads state and the editor.
+   */
+  confirmReset(): vscode.PreparedToolInvocation | undefined {
+    const confirmation: vscode.PreparedToolInvocation = {
+      confirmationMessages: {
+        title: "Reset Exercise",
+        message:
+          "Reset the current exercise to the original placeholder? Your code will be lost.",
```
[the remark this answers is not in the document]
HABER7789: Addressed the feedback in the latest commit

source/vscode/src/learning/service.ts
```diff
@@ -963,12 +1005,71 @@ export class LearningService {
...
+    input?: { unitId?: string },
+    source?: TelemetrySource,
+  ): Promise<{ unitId: string; unitTitle: string }> {
+    const course = this.activeCourse;
+    const unitId = input?.unitId ?? this.position.unitId;
+
+    if (isNotebookCourse(course)) {
+      const unit = this.findCourseUnit(course, unitId);
+      // Close any open notebook tabs for this unit before overwriting it.
+      await this.closeNotebookTab(workbookUri(unit));
```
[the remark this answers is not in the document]
HABER7789: Fixed. The closeTabs and closeNotebookTab now report whether the notebook close happened.

source/vscode/src/learning/service.ts
```diff
@@ -1013,12 +1055,79 @@ export class LearningService {
...
+      } else if (activity.type === "lesson" && activity.example) {
+        uri = this.exampleFileUri(unit.id, activity.example.id);
+        code = activity.example.code;
+      } else {
+        continue;
+      }
+      // Save any unsaved edits first so the editor is clean, then overwrite
+      // the file on disk. The editor will pick up the change automatically
+      // because it's no longer dirty.
+      await this.saveOpenDocument(uri);
```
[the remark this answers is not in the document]
HABER7789: Fixed. `saveOpenDocument` now reports whether the save landed, and reset aborts before overwriting so a dirty editor can't clobber it back.

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +31,116 @@ export async function materializeCourseWorkbooks(
...
+  edit.set(notebook.uri, [
+    vscode.NotebookEdit.replaceCells(
+      new vscode.NotebookRange(index, index + 1),
+      [data],
+    ),
+  ]);
+  if (!(await vscode.workspace.applyEdit(edit))) {
+    return false;
+  }
+  return notebook.save();
```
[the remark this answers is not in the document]
HABER7789: Went with the best effort save approach, the in editor cell edit is the reset.

source/vscode/src/learning/service.ts
```diff
@@ -979,46 +1002,148 @@ export class LearningService {
...
+      } else if (activity.type === "lesson" && activity.example) {
+        uri = this.exampleFileUri(unit.id, activity.example.id);
+        code = activity.example.code;
+      } else {
+        continue;
+      }
+      // Save any unsaved edits first so the editor is clean, then overwrite
+      // the file on disk. Abort the whole unit reset if a save didn't take,
+      // rather than clearing progress on a file we couldn't actually reset.
+      if (!(await this.saveOpenDocument(uri))) {
```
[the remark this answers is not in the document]
HABER7789: Resolved, dropped Q# unit reset; it's notebook-only now, so the loop is gone.

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +327,109 @@ export class LearningTools {
...
+    if (!isNotebookCourse(this.service.getActiveCourseInfo())) {
+      return false;
+    }
+    const editor = vscode.window.activeNotebookEditor;
+    const workbook = this.service.getCurrentCodeFileUri();
+    const onWorkbook =
+      !!editor &&
+      !!workbook &&
+      editor.notebook.uri.toString() === workbook.toString();
+    return onWorkbook && this.selectedNotebookCellId() === undefined;
```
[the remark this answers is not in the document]
HABER7789: I have fixed this, the guard now checks the active editor via isCourseWorkbook instead of stored position

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,69 @@ export class LearningTools {
...
+  async resetUnit(input?: {
+    unitId?: string;
+  }): Promise<{ unitId: string; unitTitle: string } & StateSnapshot> {
+    await this.ensureInitialized();
+    return this.invoke(async () => {
+      // Unit reset is notebook-only; the service rejects Q# courses.
+      const { unitId, unitTitle } = await this.service.resetUnit(
+        { unitId: input?.unitId },
+        "chat",
+      );
```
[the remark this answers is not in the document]
HABER7789: Addressed this in the latest commit. `reset-unit` now syncs from the active notebook before resolving the current unit, matching how the other chat tools resolve from the editor. Falls back to stored position when no notebook is focused.

HABER7789:
> ### 🔵 Needs a closer look
> Closing a workbook can change the active course or unit, causing reset callers to reopen or navigate to the wrong notebook.
> 
> Review details
> ### Suppressed comments (2)
> **Previously missed (2)** — in code that hasn't changed since the last review.
> 
> **source/vscode/src/gh-copilot/learningTools.ts:398**
> 
> * If closing this workbook activates a notebook from another course, notebook synchronization can change the active course before this line runs. `goTo` deliberately resolves only within the active course, so it may throw after the reset already succeeded or navigate to a same-named unit in the wrong course. Capture the target course before resetting and restore it before navigating.
>   **source/vscode/src/learning/commands.ts:98**
> * Closing the target workbook can activate another open course workbook, whose `onDidChangeActiveNotebookEditor` handler asynchronously changes the service position. This call then resolves the notebook URI from that new position and can reopen the wrong unit. Preserve the target course/unit and navigate back to it after `resetUnit` returns, before calling `openCourseNotebook`.
> 
> * **Files reviewed:** 9/9 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Balanced

The learning experience keeps a single course notebook open at a time focusing or opening another closes the previous via closeStaleEditorTabs. So when reset closes the current notebook there's no other course workbook to activate, and the position can't shift under the reopen. The precondition here isn't reachable through normal use. Leaving as-is

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,81 @@ export class LearningTools {
...
+    }
+    const cellId = editor.notebook.cellAt(selection.start).metadata?.id;
+    return typeof cellId !== "string";
+  }
+
+  /**
+   * Reset an entire unit, clearing completion for all of its activities.
+   * Defaults to the current unit.
+   */
+  async resetUnit(input?: {
```
amcasey, line 371: Why parameterize it?  Reset exercise targets the current exercise or nothing.

source/vscode/src/learning/service.ts
```diff
@@ -979,46 +997,126 @@ export class LearningService {
...
-   * completion status.
+   * Reset the current exercise to its original state and clear its
+   * completion status. For python-notebook courses this restores just the
+   * current cell.
    */
   async resetExercise(source?: TelemetrySource): Promise<void> {
-    // Python-notebook courses: close the notebook, re-copy the entire unit
-    // from source, and clear completion.
-    const course = this.activeCourse;
+    await this.resetExerciseAt(
```
amcasey, line 1005: It seems like the VS Code command will call this one, regardless of whether or not the current unit is a notebook.  What will it do in that case?

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,81 @@ export class LearningTools {
...
+      // A destructive reset must never silently act on a different cell, so
+      // if a workbook is focused but its selected cell can't be identified,
+      // fail loudly instead of falling back to the stored position.
+      if (this.notebookSelectionUnidentified()) {
+        throw new CopilotToolError(
+          "I couldn't tell which cell is selected — it has no stable id yet. " +
+            "Click into the exercise cell you want to reset and try again, or reset the whole unit.",
+        );
+      }
+      const state = this.serializeState(true);
```
amcasey, line 341: Existing calls to `serializeState` explain why they're passing true or false.  (This isn't the only instance.)

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,81 @@ export class LearningTools {
...
-    this.throwIfNotQSharpCourse();
     return this.invoke(async () => {
-      await this.service.resetExercise("chat");
+      // Resolve the target from the editor — the selected cell for notebook
+      // courses — rather than the stored position, matching hint/solution.
+      // A destructive reset must never silently act on a different cell, so
+      // if a workbook is focused but its selected cell can't be identified,
+      // fail loudly instead of falling back to the stored position.
+      if (this.notebookSelectionUnidentified()) {
+        throw new CopilotToolError(
```
amcasey, line 336: I think this message is an explanation to copilot, rather than to the user.  I'm also not sure what "no stable id yet" means.  I might go with something closer to "the current activity isn't an exercise".

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,81 @@ export class LearningTools {
...
+    });
+  }
+
+  /**
+   * True when the learner is on a course workbook but the selected cell has no
+   * stable id, so {@link serializeState} would fall back to the stored
+   * position — unsafe for reset. Judged from the active editor's URI so a
+   * different unit's workbook is still covered.
+   */
+  private notebookSelectionUnidentified(): boolean {
```
amcasey, line 354: I'm not sure I understand why we needed a new method.  Does this do something meaningfully different than `serializeState`/`notebookLearningState`?

source/vscode/src/gh-copilot/learningTools.ts
```diff
@@ -326,11 +326,81 @@ export class LearningTools {
...
+        { unitId: input?.unitId },
+        "chat",
+      );
+
+      // The reset closed the workbook and notebook courses don't use the
+      // lesson panel, so re-open the fresh copy. The open command resolves the
+      // notebook from the current position, so move there first — the reset
+      // unit isn't necessarily the one the learner was on.
+      await this.service.goTo({ unitId }, "chat");
+      await vscode.commands.executeCommand(
```
amcasey, line 399: We can't just ask the service to open the notebook?

source/vscode/src/gh-copilot/learningTools.ts
amcasey: Actually, I'd probably eliminate this whole check and have resetExerciseAt throw if it's unhappy with the state.  `this.invoke` exists to translate service errors into copilot errors.

source/vscode/src/learning/service.ts
```diff
@@ -979,46 +997,126 @@ export class LearningService {
...
+    }
+  }
+
+  /**
+   * Reset an entire unit: recopy its notebook from the course original and
+   * clear completion for all of its activities. Notebook courses only — a Q#
+   * unit is a folder of separate `.qs` files that are reset individually via
+   * {@link resetExercise}. Defaults to the current unit.
+   */
+  async resetUnit(
```
amcasey, line 1079: This functionality already existed, so I'm confused about why there's so much new code.

source/vscode/src/gh-copilot/learningTools.ts
amcasey: Do we even want to reopen the notebook if it wasn't the current unit?

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +32,147 @@ export async function materializeCourseWorkbooks(
...
+
+    const dest = workbookUri(unit);
+    const open = vscode.workspace.notebookDocuments.find(
+      (n) => n.uri.toString() === dest.toString(),
+    );
+    if (open) {
+      return replaceOpenCell(open, cellId, authored);
+    }
+
+    const destText = new TextDecoder().decode(
```
amcasey, line 79: Why would we be operating on a cell in a closed notebook?

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +32,147 @@ export async function materializeCourseWorkbooks(
...
+  }
+
+  const existing = notebook.cellAt(index);
+  // Restore the authored kind and language, not the learner's current ones: if
+  // they converted the exercise cell to Markdown, reset must bring back a
+  // runnable code cell. Fall back to Python — these are Python-notebook
+  // courses — when the notebook declares no kernel language.
+  const isCode = authored.kind === "code";
+  const data = new vscode.NotebookCellData(
+    isCode ? vscode.NotebookCellKind.Code : vscode.NotebookCellKind.Markup,
```
amcasey, line 135: Is this just `authored.kind`?

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +32,147 @@ export async function materializeCourseWorkbooks(
...
+    authored.source,
+    isCode ? (authored.language ?? "python") : "markdown",
```
amcasey, line 137: `existing.language`?

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +32,147 @@ export async function materializeCourseWorkbooks(
...
+
+  const edit = new vscode.WorkspaceEdit();
+  edit.set(notebook.uri, [
+    vscode.NotebookEdit.replaceCells(
+      new vscode.NotebookRange(index, index + 1),
+      [data],
+    ),
+  ]);
+  if (!(await vscode.workspace.applyEdit(edit))) {
+    return false;
```
amcasey, line 152: Should we log something?

source/vscode/src/learning/python/materialization.ts
```diff
@@ -27,21 +32,147 @@ export async function materializeCourseWorkbooks(
...
+  }
+
+  // Persist the reset best-effort. The edit already updated the editor, so a
+  // save that doesn't land just leaves the cell unsaved, not un-reset; log it
+  // for diagnostics but still report the cell as reset.
+  try {
```
amcasey, line 158: Why save?

source/vscode/src/learning/python/materialization.ts
amcasey: This file is really just for file operations - it probably shouldn't be talking to the editor at all.  It might be cleaner to just expose a helper that returns a cell and its metadata.

amcasey (changes requested):

PR #3637, opened 2026-08-25
[Circuit renderer] Compact rendering of classically controlled gate

fedimser:
* In JSON model (circuit.rs/circuit.ts) added classical_controls to represent classical controls of a gate.
* In rust, if there is a then/else branch with condition "resullt==One" or "result==Zero" and all gates inside have single target and no controls, instead of generating a group, emit gates without grouping them, but with added classical control.
* In typescript, add a code to render classically controlled gate.
* Add circuit-to-Q# conversion for classically controlled gate.
* Add circuit-to-string rendering for classically controlled gate.

### Example 1 - teleportation

Before:
image

After:
image

### Example 2

```
operation Tmp1() : Unit {
    use q = Qubit[4];

    H(q[0]);
    let m0 = M(q[0]);
    if (m0 == One) {
        X(q[1]);
    } else {
        X(q[2]);
        Z(q[2]);
    }
    let m1 = M(q[1]);
    if (m0 == Zero) {
        Rx(0.5, q[3]);
    } else {
        Ry(0.5, q[3]);
        if (m1==Zero) {
            X(q[1]);
        }
    }
}
```

image

### Example 3 - negative cases (controlled gate, complex condition, 2-qubit gate).

```
operation Tmp2() : Unit {
    use q = Qubit[4];
    
    H(q[0]);
    H(q[1]);
    let m0 = M(q[0]);
    let m1 = M(q[1]);

    if (m0 == Zero) {
        CNOT(q[2], q[3]);
    }
    if (m0 == One and m1 == One) {
        X(q[1]);
    }
    if (m0 == Zero) {
        Rxx(0.5, q[2], q[3]);
    }
}

```

image


source/npm/qsharp/src/data-structures/circuit.ts
```diff
@@ -175,6 +194,8 @@ export interface Unitary extends BaseOperation {
   targets: Register[];
   /** Control registers the gate acts on. */
   controls?: Register[];
+  /** Classical controls that determine whether the gate is applied. */
+  classicalControls?: ClassicalControl[];
```
ScottCarda-MS: I wonder if we could have a more general approach and have all controls fall under the `controls` field. The type of that field would be modified to be `(Register | ControlValue)[]` with `ControlValue` being the rename for `ClassicalControl`. That would leave open the ability to specify inverted quantum controls also.

ScottCarda-MS: Or better yet, for the type, it would be a `Control[]` with
```
export interface Control extends Register {
  inverted?: boolean;
}
```

source/npm/qsharp/src/data-structures/circuit.ts
```diff
@@ -165,6 +165,25 @@ export interface Measurement extends BaseOperation {
   results: Register[];
```
ScottCarda-MS, line 165: Since this is a change to the data contract for circuits, we should consider bumping the circuit version number and making sure we have a plan to auto-update old circuit files for backwards compatibility.

PR #3643, opened 2026-08-25
[STIM] Introduce non-Clifford gates to qdk-stim

joao-boechat:
This PR introduces support for `T`, `T_DAG`, `R_X`, `R_Y`, and `R_Z` to qdk-stim.

These gates aren't supported by official stim, so we used the [clifft docs](https://github.com/unitaryfoundation/clifft/blob/main/docs/reference/gates.md#supported-gates) as a basis for these changes. For example, while QIR expresses angles in radians, all of our rotation gates will be representing them in half turns (pi radians), similar to clifft.

### DIFF
--- a/source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
+++ b/source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
@@ -1,0 +1,362 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::check;
+use expect_test::expect;
+use indoc::indoc;
+
+#[test]
+fn t_gate_yields_expected_qir() {
+    check(
+        "T 0",
+        &expect![[r#"
+            define i64 @ENTRYPOINT__main() #0 {
+              call void @__quantum__rt__initialize(ptr null)
+              call void @__quantum__qis__t__body(ptr inttoptr (i64 0 to ptr))
+              call void @__quantum__rt__array_record_output(i64 0, ptr null)
+              ret i64 0
+            }
+
+            declare void @__quantum__rt__result_record_output(ptr, ptr)
+            declare void @__quantum__rt__array_record_output(i64, ptr)
+            declare void @__quantum__qis__t__body(ptr)
+            declare void @__quantum__rt__initialize(ptr)
+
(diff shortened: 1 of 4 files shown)


source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
```diff
@@ -0,0 +1,362 @@
...
+            !6 = !{i32 7, !"backwards_branching", i2 3}
+            !7 = !{i32 1, !"arrays", i1 true}
+        "#]],
+    );
+}
+
+#[test]
+fn r_z_yields_expected_qir() {
+    check(
+        "R_Z(0.25) 0",
```
billti: instead of just repeating the same angle in every test, maybe broaden the coverage a little (e.g. what does an `R_Z(123.432) 0` do?)

billti: Do we support pi to -pi? Or only 0 to 2*pi?

source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
```diff
@@ -0,0 +1,362 @@
...
+              | 00000000000000000000000000000000000000000000000000000000000000000000000000
+              | 00000000000000000000000000000000000000000000000000000000000000000000000000
+              | 0000000000000 half turns
+               ,----
+             1 | R_X(1e308) 0
+               : ^^^^^^^^^^^^
+               `----
+        "#]],
+    );
+}
```
billti, line 362: Can you add some Python tests too that actually run the Clifford simulator with Stim with non-Cliffords and verify the outcome probabilities? (At least one or two tests to verify it works end-to-end).

source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
joao-boechat: Good point, will add more coverage. And we do support (-inf, inf), i.e. all real numbers.

billti (approved):

source/compiler/stim_compiler/src/qir/tests/non_clifford_gates.rs
joao-boechat: Added those.

PR #3646, opened 2026-08-25
Add unit to unit navigation links to Chemistry QPE notebooks

HABER7789:
Each unit notebook now ends with links to the previous and next units, so a
learner can move through the course without going back to the Learning view.

### DIFF
--- a/source/vscode/resources/qdk-learning/utils/chemistry-qpe/README.md
+++ b/source/vscode/resources/qdk-learning/utils/chemistry-qpe/README.md
@@ -25,6 +25,12 @@ theme. `tutorial_qpe_atomic_basis_functions.png` and
 `tutorial_qpe_example_molecular_orbitals.png` stay PNG attachments. Every
 figure needs an `:alt:`, which becomes the SVG's accessible name.
 
+Each notebook ends with links to the neighbouring units. Unit order and titles
+come from `course.json` and the notebook names from `RECIPES`, so the two must
+list the same units; the converter stops if they disagree. The links point at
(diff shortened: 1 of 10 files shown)


source/vscode/resources/qdk-learning/utils/chemistry-qpe/README.md
```diff
@@ -25,6 +25,12 @@ theme. `tutorial_qpe_atomic_basis_functions.png` and
 `tutorial_qpe_example_molecular_orbitals.png` stay PNG attachments. Every
 figure needs an `:alt:`, which becomes the SVG's accessible name.
 
+Each notebook ends with links to the neighbouring units. Unit order and titles
```
amcasey, line 28: I wonder if this is something we want to insert as part of constructing the working copy?  It seems pretty tedious to keep up to date manually.

amcasey (approved):
It's not immediately clear to me why, on reaching the end of a unit, I'd want to navigate to the preceding one.  Otherwise, the buttons look nice and work well.

source/vscode/resources/qdk-learning/utils/chemistry-qpe/README.md
HABER7789: So it is already generated our converter reads the unit list from course.json, so adding or reordering units updates the links automatically. Nothing to maintain by hand.

Inserting it while constructing the working copy would work too, and would give every notebook course navigation for free instead of just this one course. Maybe we can add it to materialization which already strips author cells. I'll keep this in mind.

PR #3647, opened 2026-08-25
[STIM] Finalize non-Clifford support

joao-boechat:
This PR finalizes non-Clifford support in the STIM-to-QIR pipeline by adding the remaining gate lowerings, explicit radian angles, and CCX execution in the Clifford simulator.

## STIM compiler

- Adds QIR lowering for `TPP`/`TPP_DAG`, `CH`, `CCX`, `CCZ`, `U`/`U3`, `R_XX`/`R_YY`/`R_ZZ`, and `R_PAULI`.
- Splits number scanning into integer, fraction, exponent, and rad suffix helpers, making the lexer easier to follow.
- Makes `Double` tokens unit-aware through `DoubleUnit`, distinguishing default values from radians before parsing.
- Represents parsed arguments as `Arg { value, span }`, preserving their units and source locations for compiler validation.
- Centralizes double extraction, strips the `rad` suffix before conversion, and reports floating-point overflow at the argument span.
- Converts default angles from half turns to radians, passes explicit radians through unchanged, and rejects radians for probability arguments.
- Improves argument-count errors by reporting the expected and actual counts for too few or too many arguments.
- Reorders compiler helpers by purpose, making argument validation, target broadcasting, and QIR emission easier to navigate.
- Handles broadcasting and validation for single-, pair-, triple-, and Pauli-product targets, including negated Pauli products.

## Sample notebooks
- Consolidated all of the three stim notebooks we had into one: `qdk_stim.ipynb`.

## Tests

- Covers number parsing, radian units, overflow, gate lowering, broadcasting, argument counts, and invalid targets.

## Simulation

- Decomposes CCX into Clifford+T operations so CCX circuits can run in the Clifford simulator.

## Editor support

- Highlights the new gates and `rad`-suffixed numbers in STIM files.


source/compiler/stim_compiler/src/lex.rs
```diff
@@ -61,28 +61,28 @@ impl Display for Token {
...
+    Comma,              // ,
 }
 
 impl Display for TokenKind {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
             TokenKind::Newline => f.write_str("newline"),
             TokenKind::Uint => f.write_str("uint"),
-            TokenKind::Double => f.write_str("double"),
+            TokenKind::Double(_) => f.write_str("double"),
```
ScottCarda-MS, line 85: Since this is for debugging primarily, it would be useful to know when dealing with a double that is in radians. The default should still be "double", but the radians variant might read "double(rads)" or something.

joao-boechat: That still works fine because the content of "double" will contain the "rad" suffix. So, for example, 1rad will be lexed as `double(1rad)`

source/compiler/stim_compiler/src/qir/tests/noise_channels.rs
```diff
@@ -906,17 +937,46 @@ fn pauli_channel_1_with_probabilities_summing_to_exactly_one_is_valid() {
...
+           ,----
+         1 | PAULI_CHANNEL_1(1.5, 0.0, 0.0) 0
+           : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+           `----
+    "#]],
+    );
+}
+
+#[test]
+fn pauli_channel_1_with_probability_in_radians_yields_error() {
```
ScottCarda-MS, line 970: Could you add a test where more than one argument is expressed in radians? Should multiple errors surface, or will it just be one error for the whole call?

source/compiler/stim_compiler/src/qir/tests/peek_loss.rs
```diff
@@ -124,6 +124,50 @@ fn peek_loss_with_readout_noise_yields_expected_qir() {
...
+           ,----
+         1 | PEEK_LOSS(-0.1) 0
+           : ^^^^^^^^^^^^^^^^^
+           `----
+    "#]],
+    );
+}
+
+#[test]
+fn peek_loss_with_readout_noise_in_radians_yields_error() {
```
ScottCarda-MS, line 156: What about a test for an argument `-0.1rad`? That way we can test for both errors surfacing instead of one "beating" the other.

source/compiler/stim_compiler/src/parser/tests/targets.rs
```diff
@@ -439,16 +439,18 @@ fn loss_target() {
...
-                            kind: Loss(0)"#]],
+            Circuit [0-10]:
+                items:
+                    Instruction [0-10]:
+                        name: E
+                        tag: <none>
+                        args:
+                            Arg [2-6]:
+                                value: 0.01
+
```
ScottCarda-MS: Can we avoid having this extra line space here? It breaks off the "targets" field and makes it look like it is unrelated to the rest of the fields of Instruction.

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -94,6 +94,34 @@ impl Display for Instruction {
...
 
+#[derive(Debug, Clone, Copy)]
+pub struct Arg {
+    pub span: Span,
+    pub value: ArgValue,
+}
+
+impl Display for Arg {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln_header_with_span(f, "Arg", self.span)?;
```
ScottCarda-MS: The `value` field is pretty compact in its display. Consider making the Arg display a single line:
`Arg [x-y]: 1.2 rad`

source/compiler/stim_compiler/src/parser.rs
```diff
@@ -94,6 +94,34 @@ impl Display for Instruction {
...
+        writeln_field(f, "value", &self.value)
```
ScottCarda-MS:
```suggestion
        write_field(f, "value", &self.value)
```
I think using `write_field` like `Target` does will avoid that extra line space in the output.

samples/notebooks/non_clifford_stim.ipynb
```diff
@@ -0,0 +1,155 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "aee7f877",
+   "metadata": {},
+   "source": [
+    "## Controlled gates\n",
+    "\n",
+    "`CH` applies a Hadamard when its control is `1`. `CCX` flips its target, and `CCZ` applies a phase flip, when both controls are `1`."
```
ScottCarda-MS: It is a bit more nuanced than this makes it sound. How deep you want to describe this depends on the audience, but doing a quantum operation with a control means that the execution of the operation is entangled to the part of the quantum state of the control that is 1. There's probably a better way to word that then what I gave, but I assume there are nice descriptions of what quantum controls do out there.

domorale (changes requested):
Requesting changes based on the angle-unit design concern in my review comment. I consider this a user-facing compatibility and predictability issue: matching Clifft’s half-turn model avoids a QDK-specific Stim dialect.

samples/notebooks/non_clifford_stim.ipynb
joao-boechat: I have merged all of the stim notebooks into a single, more complete one. The idea isn't for it to be a full documentation of how all of the gates are supposed to work. For that, one can simply go to stim's docs. For now, this is supposed to be used by people who already use stim / clifft but aren't familiar with qdk.stim.

In this particular case, the description is explaining what the example is supposed to do, rather than making a general statement about those gates.

domorale (approved):
I requested changes earlier, but the discussion about opt-in rad support and compatibility clarified the design intent. I approve the approach.

ScottCarda-MS (approved):

PR #3655, opened 2026-08-26
Course authoring guide for QDK Learning courses

HABER7789:
Updates the previous course authoring guide to match the current codebase.


source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+Setting up a Python environment can be tricky for new users and we want the focus to be on the course content, so we've added some helper functionality around installing and validating dependencies.
+
+- `requirements.txt` lists course dependencies; each unit's first cell has a commented-out `%pip install -r ../requirements.txt` for the learner to run
+- The `QDK: Create a Microsoft Quantum Python virtual environment` command sets up an environment and prompts for the packages to install
+- `course.json` lets you list imports you expect to work so they can be checked before the student starts the unit (e.g. in case they've selected the wrong notebook kernel), which the course's `_check_env.py` reads
+- The course infrastructure depends on the Python and Jupyter VS Code extensions, so they'll be prompted if those are absent
+
+## Trying it out
+
+Open a workspace folder in VS Code and navigate to the Microsoft Quantum extension panel (indicated by a moebius strip).
```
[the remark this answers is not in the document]
HABER7789: Changed to "Mobius"

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+    "importChecks": ["qdk_chemistry", "pyscf"]
+  }
+}
+```
+
+- `schemaVersion`: set to `1`.
+- `title`: Display name of the course.
+- `shortDescription`: Simple help/alt text about the course.
+- `units`: An ordered list of units and where to find each.
+- `environment.importChecks`: Optional. Imports the course's `_check_env.py` verifies before a unit starts.
```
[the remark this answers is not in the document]
HABER7789: Fixed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+ANSWER = register_value_exercise("forty_two", expected=42)
+```
+
+The `sys.path` preamble is what lets the notebook import `_course_lib.py` from the course root.
+`register_value_exercise` covers the common case of checking the return value.
+Use `register_exercise(name, validate, ...)` when a unit needs its own checking; `validate` returns an error message when the answer is wrong and `None` when it's right.
+
+The validation function should raise an exception if the solution is incorrect so that running the cell fails.
+Once the cell runs successfully, the exercise will be considered to be complete.
```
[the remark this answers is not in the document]
HABER7789: Fixed, the library raises the error when validation fails and authors write return "message" or None

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+```
+resources/qdk-learning/courses/
+└── chemistry-qpe/           folder name is the course ID
+    ├── course.json          course metadata
+    ├── requirements.txt     python requirements for the notebooks
+    ├── _course_lib.py       utilities for validating exercises
+    ├── _check_env.py        checks the learner's setup against `environment.importChecks`
+    ├── 00-overview/
+    │   ├── _unit.py         unit-specific helpers (mostly exercise registration)
```
amcasey, line 20: I think we were considering renaming these files.  We might want to revisit that before we start creating a bunch of new courses. #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+    │   └── overview.ipynb   unit content (author copy)
+    └── 01-energy-and-accuracy/
+        ├── _unit.py
+        └── energy_and_accuracy.ipynb
+```
+
+Bundled courses live under the extension's `resources` folder.
+A course can also be authored in a workspace under `qdk-learning/courses/`, which is where the extension copies bundled courses the first time a learner opens one.
+
+`_course_lib.py` and `_check_env.py` aren't supplied by the extension, so copy them from an existing course when starting a new one.
```
amcasey:
```suggestion
`_course_lib.py` and `_check_env.py` contain utilities that you may find helpful, but aren't required.  You can use the ones from an existing course as models.
``` #Closed

File: source/vscode/authoring-courses.md [hunk 6]
```diff
@@ -0,0 +1,157 @@
...
+
+_Note_: The folders in this example are numbered, but that's not required - the order comes from `course.json`
```
amcasey:
```suggestion
_Note_: The folders in this example are numbered, but that's not required - the display order comes from `course.json`
``` #Closed

File: source/vscode/authoring-courses.md [hunk 7]
```diff
@@ -0,0 +1,157 @@
...
+}
+```
+
+- `schemaVersion`: set to `1`.
+- `title`: Display name of the course.
+- `shortDescription`: Simple help/alt text about the course.
+- `units`: An ordered list of units and where to find each.
+- `environment.importChecks`: Optional. The imports the course's `_check_env.py` checks before a unit starts.
+
+The course ID comes from the folder name, so there is no `id` field.
```
amcasey: Unnecessary #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+    "importChecks": ["qdk_chemistry", "pyscf"]
+  }
+}
+```
+
+- `schemaVersion`: set to `1`.
+- `title`: Display name of the course.
+- `shortDescription`: Simple help/alt text about the course.
+- `units`: An ordered list of units and where to find each.
+- `environment.importChecks`: Optional. The imports the course's `_check_env.py` checks before a unit starts.
```
amcasey: There's an extra word or something - please rephrase. #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+The course ID comes from the folder name, so there is no `id` field.
+
+## Exercises
+
+Exercises are one of the main value-adds of the QDK Learning courses.
+You can tag individual code cells as exercises and associate hints, solutions, and explanations with them.
+Each exercise also has associated validation logic to tell a learner whether their solution is correct and course progress is updated as they complete exercises successfully.
```
amcasey: Nit: we changed this and now progress reflects all code cell executions #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+An exercise code cell is preceded by a markdown cell giving the name of the exercise (as a header) and a brief description of what's to be accomplished.
+The exercise code cell itself is tagged `exercise` using the `Add Cell Tag` functionality in VS Code (or by editing the JSON directly).
+
+<div align="center">
+  <img src="../../media/add-cell-tag.png" alt="Adding a cell tag from the notebook cell context menu">
+</div>
+
+The exercise code cell may optionally be followed by additional cells tagged `hint`, `solution`, or `explanation`.
```
amcasey: There can be multiple of each #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+`solution` goes on a code cell; `hint` and `explanation` go on markdown cells.
+A tag on the wrong kind of cell is ignored with only a log warning, so the cell simply never reaches Copilot.
```
amcasey: Don't need this sentence #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+from _course_lib import exercise, register_value_exercise  # noqa: E402, F401
+
+ANSWER = register_value_exercise("forty_two", expected=42)
+```
+
+The `sys.path` preamble is what lets the notebook import `_course_lib.py` from the course root.
+`register_value_exercise` covers the common case of checking the return value.
+Use `register_exercise(name, validate, ...)` when a unit needs its own checking; `validate` returns an error message when the answer is wrong and `None` when it's right.
+
+When validation fails, `_course_lib` shows the message and raises, so the cell errors out.
```
amcasey, line 118: I'm used to "throws", but I take it "raises" is idiomatic in python? #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+Once the cell runs successfully, the exercise will be considered to be complete.
+
+## Editing a published course
```
amcasey, line 121: Probably need to mention that, once a course is copied into a learning workspace, it won't be updated from the extension's bundled copy, even if the extension is updated.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+Progress is tracked per cell, using the notebook's nbformat cell IDs.
+Those IDs are opaque to the learning system, so editing a cell's text is safe, but deleting a cell and adding a replacement gives it a new ID and the old completion no longer matches it.
```
amcasey: I'm not sure what "opaque to the learning system" means.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+Existing learners are insulated from this: a `*.workbook.ipynb` is never overwritten once it exists, so they keep working in the copy they already have.
```
amcasey: I think I would rephrase this in terms of my comment above - the course won't be updated.  The workbook _will_ be overwritten if they reset the unit.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+The change reaches them only when they reset the unit, and it reaches anyone starting the course after the update.
```
amcasey: Is that true?  I wasn't expecting resets to pick up updates.  I was pretty sure I'd copied the author-copy into the workspace as well and it was refreshed from that.

amcasey: You may be referring to the scenario where the course isn't bundled.  That scenario is probably worth describing, but I think it would help to be explicit about which statements apply to which scenario.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+Images are worth a note too.
```
amcasey: I might reframe this as a recommendation: where possible, prefer SVGs to embedded images because SVGs can adapt to the learner's UI theme.

amcasey: Also, it seems like it was hard to figure out how to do that, so an example might help.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+An attachment is rendered as an `<img>`, which cannot read the notebook's stylesheet, so a light-background diagram stays light in a dark theme.
+Putting the SVG markup directly in a Markdown cell keeps it in the notebook's own DOM, where it can pick up `var(--vscode-*)` colours and follow the active theme.
+Attachments that stay attachments must be base64-encoded.
+
+## Environment
+
+Setting up a Python environment can be tricky for new users and we want the focus to be on the course content, so we've added some helper functionality around installing and validating dependencies.
+
+- `requirements.txt` lists course dependencies; each unit's first cell has a commented-out `%pip install -r ../requirements.txt` for the learner to run
```
amcasey: Whether the first cell contains that text would seem to be up to the course author.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+- The `QDK: Create a Microsoft Quantum Python virtual environment` command sets up an environment and prompts for the packages to install
```
amcasey: That's true, but it's worth noting that the package list is specific to things QDK courses might want.  I'm not sure it even includes QDK EC.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+- `course.json` lets you list imports you expect to work so they can be checked before the student starts the unit (e.g. in case they've selected the wrong notebook kernel), which the course's `_check_env.py` reads
+- The course infrastructure depends on the Python and Jupyter VS Code extensions, so they'll be prompted if those are absent
```
amcasey: Nit: I find "they" ambiguous here - consider "the learner" #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Trying it out
+
+Open a workspace folder in VS Code and navigate to the Microsoft Quantum extension panel (indicated by a Mobius strip).
+If you've never used it before, it'll offer you a `Start Learning` button.
```
amcasey, line 161: "To return to this state, you can delete qdk-learning.json" (with more details, probably)

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+The Microsoft Quantum Katas are the default course, so you'll need to explicitly Switch Course to your new content in the tree view.
```
amcasey: Not as true as it used to be #Closed

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,157 @@
...
+
+When you switch to your course, temporary working copies of all the notebooks will be created (indicated by the `.workbook.ipynb` file extension).
+These copies omit all the exercise hints, solutions, and explanations and give the learner a notebook they can edit freely without worrying about overwriting anything important.
+An existing working copy is never overwritten, so if you based your course on a sample you'd previously run, `git clean` it first to clear out any working copies left behind.
```
amcasey: Personally, I find "never overwritten" too strong. #Closed

amcasey: `git clean` is a dangerous thing to recommend if you're not confident the author has their course files checked in.

source/vscode/authoring-courses.md
HABER7789: Alright, left the current names in for now, will update the guide when we rename them

source/vscode/authoring-courses.md
HABER7789: Fixed.

source/vscode/authoring-courses.md
HABER7789: Applied your suggestions.

source/vscode/authoring-courses.md
HABER7789: Removed it.

source/vscode/authoring-courses.md
HABER7789: I have rephrased it as "A list of imports that `_check_env.py` verifies before a unit starts."

source/vscode/authoring-courses.md
HABER7789: Fixed.

source/vscode/authoring-courses.md
HABER7789: Right, I have made the change.

source/vscode/authoring-courses.md
HABER7789: Removed.

source/vscode/authoring-courses.md
HABER7789: That's right, both mean the same thing

source/vscode/authoring-courses.md
HABER7789: Mentioned that in the latest changes

source/vscode/authoring-courses.md
HABER7789: My bad for being too vague, removed it, I meant that we match ID string and never look at cell content, but progress is still there.

source/vscode/authoring-courses.md
HABER7789: Fixed, reframed around the course not updating. Also noted that resetting a unit does overwrite the working copy.

source/vscode/authoring-courses.md
HABER7789: Yes you are right, a bundled course is only copied into the workspace once, so an existing learner never gets updates not on extension update, not on reset. Corrected myself here

source/vscode/authoring-courses.md
HABER7789: Fixed it as a recommendation to prefer the SVGs and also added a small example snippet

source/vscode/authoring-courses.md
HABER7789: Fixed.

source/vscode/authoring-courses.md
HABER7789: Dropped it, now it just says SwitchCourse.

source/vscode/authoring-courses.md
HABER7789: Oops, thanks for the catch. Instead mentioned about deleting the leftover `*.workbook.ipynb` files

HABER7789: Okay, changed it with "isn't replaced when the course is copied"

source/vscode/authoring-courses.md
HABER7789: Agreed I had written it like a platform rule. Reworded to attribute it to the chemistry qpe course as a pattern you can copy, rather than something the system expects

source/vscode/authoring-courses.md
HABER7789: Applied it

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+
+When the course isn't bundled - say, one the learner cloned from a repo - they update the course copy themselves, so they'd pull your changes and then reset the unit to rebuild the working copy from them.
+
+Where possible, prefer inline SVG over embedded images, because an SVG can adapt to the learner's UI theme.
+An attachment is rendered as an `<img>`, which cannot read the notebook's stylesheet, so a light-background diagram stays light in a dark theme.
+Putting the SVG markup directly in a markdown cell keeps it in the notebook's own DOM, where it can pick up `var(--vscode-*)` colours:
+
+<!-- prettier-ignore -->
+```html
+<svg viewBox="0 0 120 40" role="img" aria-label="Example"><rect x="1" y="1" width="118" height="38" fill="none" stroke="var(--vscode-editor-foreground)"/><text x="60" y="25" text-anchor="middle" fill="var(--vscode-editor-foreground)">Example</text></svg>
```
amcasey, line 138: Consider including the output in light and dark themes.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+Keep the markup on a single line - if it spans multiple lines, markdown splits it into separate paragraphs and it won't render.
+Attachments that stay attachments must be base64-encoded.
```
amcasey:
```suggestion
Non-SVG images must be embedded as base64-encoded text.
```

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Environment
+
+Setting up a Python environment can be tricky for new users and we want the focus to be on the course content, so we've added some helper functionality around installing and validating dependencies.
+
+- `requirements.txt` lists course dependencies; the Chemistry QPE course starts each unit with a commented-out `%pip install -r ../requirements.txt` for the learner to run, which is a pattern you may want to copy
```
amcasey: This is more of a recovery step if we discover that the learner's environment is inadequate.  I'm not sure we want to include it (or recommend including it) as code that run-all will trigger in each unit.  Also, I personally find it quite easy to accidentally run it against my global python installation.

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+- The `QDK: Create a Microsoft Quantum Python virtual environment` command sets up an environment and prompts for the packages to install; the list it offers is specific to what QDK courses tend to need, so check that it covers your dependencies
```
amcasey:
```suggestion
- The `QDK: Create a Microsoft Quantum Python virtual environment` command sets up an environment and prompts for the packages to install; the list it offers is specific to what QDK courses tend to need.  If it already includes the packages you need, you can direct learners to the command palette for virtual environment setup.
```

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+- `course.json` lets you list imports you expect to work so they can be checked before the learner starts the unit (e.g. in case the learner selected the wrong notebook kernel), which the course's `_check_env.py` reads
+- The course infrastructure depends on the Python and Jupyter VS Code extensions, so they'll be prompted if those are absent
+
+## Trying it out
+
+Open a workspace folder in VS Code and navigate to the Microsoft Quantum extension panel (indicated by a Mobius strip).
+If you've never used it before, it'll offer you a `Start Learning` button.
+Use `Switch Course` in the tree view to select your new content.
+Progress lives in `qdk-learning.json` at the workspace root: the current position, which activities are complete, the Python environment chosen per course, and whether a course has been picked yet.
+Delete it to get back to the first-run state.
```
amcasey: It's not really the first-run state if the workbook files are already there with the exercises filled in.  Maybe "To return to the first-run state, delete `qdk-learning.json` and all `*.workbook.ipynb` files."

amcasey: And maybe put this after we introduce working copies in the next paragraph.

source/vscode/authoring-courses.md
amcasey:
```suggestion
- The course infrastructure depends on the Python and Jupyter VS Code extensions, so the learner will be prompted if those are absent
```

source/vscode/authoring-courses.md
```diff
@@ -0,0 +1,169 @@
...
+That resets progress only - the `*.workbook.ipynb` files are left alone, so delete those too for a completely clean run.
+
+When you switch to your course, temporary working copies of all the notebooks will be created (indicated by the `.workbook.ipynb` file extension).
+These copies omit all the exercise hints, solutions, and explanations and give the learner a notebook they can edit freely without worrying about overwriting anything important.
+An existing working copy isn't replaced when the course is copied, so if you based your course on a sample you'd previously run, delete any leftover `*.workbook.ipynb` files first.
```
amcasey: I find the multiple uses of "copy" confusing.
```suggestion
If the working copy is already present, it won't be overwritten the next time the course is loaded, so if you based your course on a sample you'd previously run, delete any leftover `*.workbook.ipynb` files first.
```

amcasey (approved):

source/vscode/authoring-courses.md
HABER7789: Added a side-by-side render beneath the svg example.

source/vscode/authoring-courses.md
HABER7789: Applied, thanks!!

source/vscode/authoring-courses.md
HABER7789: Right, thanks for catching this, I have dropped it.

source/vscode/authoring-courses.md
HABER7789: Applied your wording

source/vscode/authoring-courses.md
HABER7789: Applied your rewrite.

source/vscode/authoring-courses.md
HABER7789: I have changed it and placed it in the next paragraph.

source/vscode/authoring-courses.md
HABER7789: Fixed

source/vscode/authoring-courses.md
HABER7789: Agreed, I have mentioned that

